Issue #1, opened 2018-05-03
Change name in .lando.yml


genjohnson:
The name of this docroot in .lando.yml is currently `uiowa`. As we will have additional D8 applications in the future, it would be helpful to match the name in lando with the application name. So I propose we change the name in .lando.yml to `uiowa801`.

richardbporter:
I think the better approach is to leave the application name as uiowa since that is the machine name in AC. Lots of things rely on that including multisite, drupal-manifest, etc.

I've renamed the repository. Users should get a message when interacting with the remote to update the URL. We should notify them to rename their local repository directory.

PR #2, opened 2018-05-03
Change name to uiowa801

genjohnson:
This PR addresses issue #1.


richardbporter:
This works but causes a regression in the appserver URLs. I think the better option is match the application name in Acquia: uiowa.

PR #159, opened 2019-05-06
Standardize multisite naming conventions.

### LINKED ISSUE #151: Clean up databases once BLT artifact is deployed to stage, prod
- [ ] Delete eight_uiowa_edu

- [ ] Delete grad_uiowa_edu

- [ ] Delete sitenow

- [ ] Delete uiowa_edu

- [ ]  Delete www_dentistry_uiowa_edu

- [ ] Delete hr (see #155)

### LINKED ISSUE #155: Standardize multisite machine names.
The machine name is used for:

- the site directory
- the alias name
- the database name

We have more flexibility which is good but do we want to standardize to what multisite generated automatically? We could add that in `blt/src/Blt/Plugin/Commands/MultisiteCommands.php`. The big change here is the site directory. We're used to having those be the fully qualified URL. It needs to be the machine name for BLT to include the database automagically. We could manually set that include but it might be nice to stick with the BLT Way.

For reference: https://github.com/uiowa/multisite/blob/master/src/Site.php#L45

richardbporter:
Resolves #155 

I can resolve #151 once this is deployed to dev.


richardbporter:
@joewhitsitt I know we talked in person but can you glance over this and approve?

joewhitsitt:
👍

PR #163, opened 2019-05-08
Remove uiowa_env in favor of contrib.

richardbporter:
One less module we have to maintain.


richardbporter:
Adminimal does not look great with this. We should investigate Claro.

joewhitsitt (changes requested):
Ideally, accessible background colors should be used to handle the white foreground text as long as we are using adminimal.

#C3561A test
#31873E local

joewhitsitt (approved):

joewhitsitt:
I think I am getting a chicken before the egg issue: `The module uiowa_env does not exist.` on CIM. Should we uninstall and then remove it in a separate deployment?

PR #165, opened 2019-05-10
Move uiowa_bootstrap to docroot/themes/custom.

richardbporter:
So it is tracked in the repo.


richardbporter:
I'm looking into frontend asset compilation with BLT.

richardbporter:
Nevermind. I'll follow up later with that.

joewhitsitt:
the updb didn't do the trick. HR says `Base theme uiowa_bootstrap has not been installed.`

richardbporter:
HR does not seem to have any problems with the theme living in docroot/themes/custom.  

@joewhitsitt You good with this?

We may want to take a similar approach to modules, eventually. #174

joewhitsitt (approved):
Changed HR uiowa_bootstrap paths back. I think this is good to go.

PR #168, opened 2019-05-10
Improve recipes:multisite:init and document.

richardbporter:
This should allow developers to create new multisites and install locally. 

Check this branch out locally and run through the instructions in the readme. Feel free to make improvements!

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,8 @@ Acquia Cloud UI. Once merged and deployed, the database can be synced.
 Example: `blt recipes:multisite:init --site-uri mysite.com`.
 
 ## Databases
-Use [SequelPro](https://www.sequelpro.com/) to manage your local databases.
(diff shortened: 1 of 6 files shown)


richardbporter:
I think two more things would be helpful here:

1. Add some docs on creating databases and domains in AC.
2. Change the localhost database names to match those in AC, i.e. an underscored site directory (URI).

The second will require people to rename their existing local databases and then regenerate each site's local.settings.php files.

richardbporter:
I've made those changes. Apologies if that messes up your local setup but I think its for the best to be consistent.

joewhitsitt:
Added a bit about upping php resources as my 128M default memory exhausted during site install.

joewhitsitt:
Otherwise, this appears to have worked as expected. See my result here: https://github.com/uiowa/uiowa/tree/add-gis-sites

joewhitsitt (approved):

PR #176, opened 2019-05-13
Claro + frontend requirements.

richardbporter:
To test:

- Check the README and follow step 6.
- Run `blt dsa`.


joewhitsitt:
We should get the nvmrc files to be all the same (docroot, uiowa_bootstrap, hr). Is there are particular reason why Claro needs node v8 versus a more recent version (v10 used in the themes)?

richardbporter:
No, I don't think so. Good idea.

richardbporter:
Could we only have one nvmrc and run compiles from the app root with a CWD option?

joewhitsitt:
Chronic wasting disease (CWD)? I don't follow.

richardbporter:
CWD = current working directory.

Nevermind. I don't think that is possible with NPM. Looks like a Yarn-only option.

joewhitsitt:
After looking at the [issue queue](https://github.com/jsdrupal/drupal-admin-ui/issues), this theme looks farther off than I have previously heard. Their goal was 8.7 but it just now has an alpha release with several elements still needing to be touched. I think we should wait for at least a beta version.

richardbporter:
@joewhitsitt That is the issue queue for the Drupal Admin UI, a separate project. The Claro issue queue is here: https://www.drupal.org/project/issues/claro?categories=All They're targeting 8.8 for Claro as an experimental theme. 

Are you seeing any problems with this theme + our page building tools?

joewhitsitt (approved):
still don't like the dropdown button styles but overall, i like claro over adminimal

richardbporter:
After further testing, I don't think is improves the usability of the admin UI. There are a too many unpolished aspects of it. Most problematic is the loss of margin/padding when expanding paragraphs items. Also, the negative space is nice but the expanded pages makes it harder to work with our page building tools. A more compact UI is better for us.

PR #192, opened 2019-05-17
Add command to clean Acquia remote.

richardbporter:
Its difficult to deploy an artifact with so many refs cluttering up the interface. This command can be run locally to delete all refs except master/develop in the Acquia remote. Requires adding the remote locally, i.e. `git remote add acquia URL`.


joewhitsitt (approved):
i will blindly approve!

PR #199, opened 2019-05-20
Misc fixes/additions

joewhitsitt:


richardbporter (approved):
The xDebug extension path gave me an error. I'm not sure why or if everyone will have that problem. Added some additional docs.

PR #200, opened 2019-05-20
provision www.dentistry.uiowa.edu

briand44:


richardbporter:
The internal domains in sites.php won't work since they're already taken up by the dentistry application. We'll need to remove one or more to add them to the uiowa app or use different domains.

richardbporter (commented):
The internal domains in sites.php won't work since they're already taken up by the dentistry application. We'll need to remove one or more to add them to the uiowa app or use different domains.

briand44:
thanks for the reminder. I'll start with just moving the dev domain.

richardbporter (approved):

Issue #219, opened 2019-05-24
Related content is not working on newer sites (post hr, dentistrycallctr)


joewhitsitt:
Blocks in general do not seem sustainable for our setup. It was removed with https://github.com/uiowa/uiowa/commit/5e741fa6644389cf7dce96269231ce82615aced9

We should consider making it an HR only feature and working on a "featured content" swiss-army-like paragraph bundle for the service.

joewhitsitt:
I am going to work on removing this functionality but leaving it for HR (forks basic page...)

Issue #220, opened 2019-05-24
Featured Content Paragraph Bundle


joewhitsitt:
Related to #219 and the need/want to pull in entity content from elsewhere in the site (people, etc.)

I suggest a multi reference field that uses an entity browser to select one or more items based on content filtering. Filtering based on tag or other taxonomy (unit, person type?) could make things difficult or force multiple entity browser tabs.

entity browser does not have a select all based on filter option though it is on entity browser enhanced's roadmap: https://www.drupal.org/project/entity_browser_enhanced/issues/2972699

Display should have options like articles, events with an additional listgroup option.

joewhitsitt:
This is being worked on again here https://github.com/uiowa/uiowa/pull/337

dnkemper:
- [ ] - Make entity references backend display more table like (entity browser) (done)
- [ ] - Select all results
- [x] - Remove tags from entity browser view
- [x] - Remove ‘read more’ link from person teaser
- [x] - field_teaser (short bio) needs to be displayed in the person teaser if not empty

dnkemper:
**Testing Instructions**

1. Go to content -> add content -> person
2. Add Info (first & last name, credentials, title, email, phone, bio, *on right sidebar* add short bio and person photo **make sure to add a short and long bio*
3. Save as ‘Published’
4. Please add a second person too
5. Go to content -> add content -> and edit an existing page
6. On the right hand menu, ‘Tags’ should not exist
7. Add a featured content paragraph to the existing section 
8. Click on ‘Select Content’
9. In the popup window, check the two people you created*
10. Should be grouped by CT (not required just thought it helped visually)
11. The two people should be displayed vertically, **NOT** horizontally
**FACT**: entity reference backend display is not feasible at the moment (See #[414](https://github.com/uiowa/uiowa/issues/414))
12. When you save the page, you should see the person’s short bio
13. There should NOT be a ‘Read More’ option on the ‘Person’ teaser
14. When you click on the person, you should see the full view/bio

joewhitsitt:
This was merged in.

PR #222, opened 2019-05-28
Split related content

joewhitsitt:
#219


joewhitsitt:
This removes related content (entity reference sidebar view block) from the "service" and makes it custom for HR. Would love another eye on this before merge. It doesn't seem to cause issue on sites that were using it (callctr.dentistry) and hr seems to still work (node/713)

richardbporter:
I'm confused - this was already removed from the sync directory for shared sites. It still exists in the HR config and seems to be working on HR right now.

richardbporter:
Ah, I see. The field still exists for all sites.

richardbporter:
Sync worked. No config discrepancies. I think this is fine so that folks do not continue to use the shared field since it does not work. However, splitting basic page for HR is probably not a good idea long term and we'll want to get #220 implemented soon.

richardbporter (approved):

PR #224, opened 2019-05-29
Update label publishing options to page options, resolves #216

### LINKED ISSUE #216: Publishing options title is misleading
Makes it sound like more of a workflow/status fieldset but it has visual/design fields.

image

ethanstjohn:
Resolves #216


richardbporter (approved):
Thanks!

PR #225, opened 2019-05-29
Person Additions

joewhitsitt:
- position, credentials, biography, image, short bio (teaser) fields
- credential field renders as a subtitle on node full by utilizing a new variable in an overridden block template.
- reduce extends in article and person CSS
- content type specific teaser labels
- start standardizing teaser markup

This is work towards feature parity.


joewhitsitt:
Apparently there are some decisions on campus that would affect all of this. Waiting for more info.

joewhitsitt:
Moving forward with local people until APR or whatever else comes down the pipeline.

joewhitsitt:
I think this is ready to review. It would be nice to get another set of eyes on this.

richardbporter:
A few things:

- The WYSIWYG is missing the padding we added in #232. Maybe we should scope that differently.
- Why is the image field in the collapsible?
- Why is the content description in the collapsible?
  - Also, is summary not a thing anymore?

The image was gigantic on the full content display until I compiled SASS locally. Gulp.

joewhitsitt:
The image and the teaser are in the collapsible for consistency. Articles (now) and basic page (soon) have the same setup.

The decision to use field_teaser instead of a body summary was a long time ago, but I feel the language is different depending on the content. Article Summary, Short Biography, Content Description

richardbporter:
03bfafb exports the credential field to be hidden on the minimal entity form display.

richardbporter:
I can't review because I have not been requested but 👍

richardbporter (approved):

PR #228, opened 2019-05-30
Compile SASS in CI and remove unused code.

richardbporter:
The biggest change here is the removal of the hawkeye dependency and reorganizing the SASS partials.


joewhitsitt:
I think we can get this working with node 10. I ran into an issue with node-sass and had to downgrade it to work with node 10. I figured node 10 was the preferred upgrade of the two.

https://github.com/uiowa/uiowa_bootstrap/commit/22264a05b89f5c9386d30e3fc382c70d200b62a9#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

richardbporter:
@joewhitsitt Ping!

joewhitsitt:
This initialized correctly once I installed NVM the preferred way (not via Homebrew).

Some styles are missing. The hawkeye brand colors are gone. primary is blue, etc.

joewhitsitt (approved):
Not crazy about all themes being compiled per site during blt dsa or blt ds, but this does lead to more consistency and reduces css merge conflicts.

It also adopts Yarn, which from what I've heard is used in other parts of ITS.

Issue #221, opened 2019-05-24
Remove media_entity_browser


richardbporter:
No longer used.

richardbporter:
Feel free to open the PR when ready for review.

PR #230, opened 2019-05-30
Delete media_entity_browser, resolve issue #221

ethanstjohn:


richardbporter:
I think something may have gotten goofed up here. The only file changes should be in the composer.json and composer.lock files.

Would you mind creating a new feature branch off the latest master, making the change and then opening another PR?

PR #235, opened 2019-05-31
Adminimal 2

ethanstjohn:


joewhitsitt:
I think `.toolbar-tab` is what we are concerned about. It is grey text on our environment bg-color that is inaccessible. The icon is decorative and doesn't have to meet contrast guidelines.

I wouldn't worry about the sub tool bar.

joewhitsitt (changes requested):
See comment about only focusing toolbar-tab

ethanstjohn:
I committed the change.

ethanstjohn (commented):
Hopefully, fixed.

joewhitsitt (approved):
looks good after https://github.com/uiowa/uiowa/pull/235/commits/e3a56d7f03eee625d30d4d3685d4545f2d24a12b'

tour is a button and wasn't getting the 'a' specific style

Issue #208, opened 2019-05-23
uiowa_bootstrap CSS is massive


richardbporter:
With absurd selectors. This is causing performance issues and needs to be reduced.

richardbporter:
This is due to the use of `@extend` everywhere. I'm not sure if there is a quick fix since BS partials don't ship with placeholders.

richardbporter:
Probably the best course of action is to implement the BS classes at the template/preprocess level and then remove the `@extend`s and avoid using them going forward.

richardbporter:
Example in #225: https://github.com/uiowa/uiowa/pull/225/commits/23d3d720ffda89e005d54c131156dd089ef83467

joewhitsitt:
If the activity by OSC and CLAS morphs this issue into a re-write with components we should look into some CSS via libraries performance: https://www.previousnext.com.au/blog/performance-improvements-drupal-8-libraries

The main key being to load the 90% used CSS as aggregated/compressed and the one off components separately. Otherwise a one-time component being loaded would throw out the entire aggregated/compressed/cached css file.

joewhitsitt:
As of right now, it sounds like uiowa_bootstrap is not going away any time soon (HR), but majority of theming resources will be going towards a new HDS/UDS based theme and adapting uiowa_bootstrap to work with the result of #260. Will anymore work be done on optimizing/reducing size of this theme?

richardbporter:
Not if its going away eventually.

PR #241, opened 2019-06-04
initial work on removing unnecessary sass extends

joewhitsitt:
Related to #208


joewhitsitt:
This is a step in the right direction. Reduces compiled size. More work could be put into preprocessing in other bootstrap classes or writing bootstrap specific templates.

richardbporter (approved):

PR #242, opened 2019-06-04
update media settings so we can link directly to files

briand44:
This allows directly linking to files when using linkit rather than linking to the file entity page.


briand44:
When adding new links to files it seemed to add the path of media/[fid]/edit. I am not sure if this could cause issues down the road if we changed the linkit matcher to something different (link to canonical url).

joewhitsitt:
This fixes HR's issue. Ideally, I'd like this to be pushed sooner than later. Not super happy about the media/{id}/edit alias within Linkit, but this fix appears to work for existing and new. I haven't been able to find an issue that mentions the edit alias change.

joewhitsitt (approved):

richardbporter (commented):
Is the iframe_domain change related?

richardbporter:
Otherwise, works for me.

PR #244, opened 2019-06-04
Multisite initialization fixes.

### LINKED ISSUE #158: Multisite initialization issues.
- [x] Order of operations is not clear.
- [x]  Site install not ran locally.
- [x]  Poor user 1 credentials
- [x]  No sitename or differentiation between sites
- [x]  Database and domain creation is a manual process.
- [ ]  ~~Cloud hook failure during initial code deployment.~~
- [x]  No local Drush alias.

richardbporter:
Somewhat resolves #158. Not much we can do about the task failure for now.

Two big changes:

- Remote Drush aliases changed to production
  - Content entry _should_ happen there.
- HR domains changed to match all others.
  - There should be no one-off domain names.

This will require:
- changing the existing HR domain names
- syncing content and files for HR to prod
- instructing customers to work in prod domain from now on


richardbporter:
We should add an update hook for existing sites that sets the user 1 name, email and role.

joewhitsitt:
I have notified HR of the change and said that we plan to deploy this before the end of the week.

richardbporter:
I've been sidetracked with some D7 work if anyone wants to take a shot at writing that update hook.

richardbporter:
Thanks!

PR #246, opened 2019-06-04
Move profile modules to custom directory.

### LINKED ISSUE #174: Modules not placed where BLT expects.
We are Git-ignoring docroot/themes/custom and docroot/modules/custom because we require some internal modules through Composer. With #165, we will stop ignoring themes so we may want to move modules as well.

That would require changing the package type of our internal modules to contrib.

richardbporter:
Resolves #174.

Also moves all templates to theme.

Requires a `composer install` to move things around.


richardbporter:
Update: I got sick of searching for templates in lots of different places.

joewhitsitt (approved):
Depending on which one gets merged first (this or #244), the update number will have to change.

joewhitsitt:
I am getting errors with this now. Approval revoked...

I got farther with `blt sme cr` before a `blt dsa` but it keeps choking on the default site. Warnings for missing modules but then:
`Error: Class '\Drupal\uiowa_seo\ConfigOverrides\GoogleAnalyticsOverride' not found`

PR #257, opened 2019-06-10
update entity embed to rc1

### LINKED ISSUE #104: Investigate linking image assets from WYSIWYG
In Drupal 7, editors were able to wrap an image in a link by selecting the image and then using the Link button in the WYSIWYG toolbar. Similar functionality is currently not available or working in our Drupal 8 WYSIWYG.

Looks like there may be a few issues out on D.O. for this:
- https://www.drupal.org/project/entity_embed/issues/2511404

briand44:
resolves #104

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "drupal/default_content": "^1.0@alpha",
         "drupal/editor_advanced_link": "^1.4",
         "drupal/entity_browser": "2.x-dev",
-        "drupal/entity_embed": "^1.0@beta",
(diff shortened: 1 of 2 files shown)


joewhitsitt:
Appears to resolve #104, but there is a lot more stuff with this: https://git.drupalcode.org/project/entity_embed/compare/8.x-1.0-beta2...8.x-1.0-rc1

Should we start slowing down with media/entity_embed updates like we did in Drupal 7 or continue to dive in head first?

joewhitsitt (approved):
i think there is more potential good than bad with this update.

PR #264, opened 2019-06-13
Exif Orientation correction

### LINKED ISSUE #194: EXIF data iPhone orientation
Photos taken from an iPhone with orientation data get rotated incorrectly, and need to be stripped of their orientation EXIF data prior to being uploaded to Drupal.

cory-skeers:
Resolves #194 

Adds a file_presave hook for JPEGs, checks for orientation EXIF and if found, rotates image and resets orientation metadata.

Does not currently handle EXIF orientations which also mirror the image, but this can be expanded in the switch statement. Working on that, though they should be less common. Currently testing it against different camera/phone models, but File Metadata Manager seems to handle them pretty consistently.


joewhitsitt:
Is this still a work in progress or is it ready for review? can you explain why you are implementing a custom module for this versus using a drupal contrib module(s)?

cory-skeers:
Ready for review now that the branch has been updated with master.
I couldn't find a Drupal contrib module that worked consistently. The closest was ImageEffects (https://www.drupal.org/project/image_effects) that works off of File Metadata Manager and ImageMagick, but had issues when different styles were applied (it might be correctly oriented when embedded in body content, but not as an image for a card).

Don't think this is the only way to go about it, but it was the most consistent I could get to work. Open to suggestions, even if it's to do another review of contrib modules.

joewhitsitt:
Thanks for working on this.

At first I thought this wasn't working, but then I realized the module wasn't enabled. 🤦‍♂Assuming those test images are the real deal, this appears to work as expected.

I'd argue that this code goes into uiowa_media_wysiwyg. And correct me if I am wrong, but I don't see the need for the .libraries.yml file (duplicate of uiowa_media_wysiwyg?), but if if this is all in the same module, that will resolve itself.

I think any existing images that have this issue can use the replace file method once this is in place to update their content.

cory-skeers:
I agree it doesn't need to be a standalone module--moved it into uiowa_media_wysiwyg. Also got rid of the excess files, including the test images, as they can be tested locally and don't need to ship out for new builds.

joewhitsitt (approved):

PR #268, opened 2019-06-14
add safe tokens to pathauto settings, resolves #267

### LINKED ISSUE #267: [node:menu-link:parents:join-path]/[node:title] is not working as expected
On the HR site, slashes are getting removed:

/developmentreward-and-recognitionimproving-our-workplace-award-iowa/writing-nomination

which should be:
development/reward-and-recognition/improving-our-workplace-award-iowa/writing-nomination

solution from what I have read is to add safe tokens:
`alias, path, join-path, login-url, url, url-brief` to the config.

joewhitsitt:


richardbporter (approved):
An update hook would be nice... :)

PR #272, opened 2019-06-17
Webform theming

GaryRidgway:
Theme the webform Implementation and add a few other small tidbits of functionality.


docroot/profiles/custom/sitenow/css/admin-overrides.css
```diff
@@ -45,12 +45,16 @@
   background: #0071b8;
 }
 
+.adminimal .webform-ui-element-container .button--small {
```
joewhitsitt, line 45: Do you think that this and `.adminimal .webform-ui-element-type-table .button--small` could be consolidated?

GaryRidgway: Yes definitely I would say this should be done.

GaryRidgway: this has been done with https://github.com/uiowa/uiowa/pull/272/commits/304d2a827364a5718adc0b67fc55c6936296ba7b

joewhitsitt (approved):

PR #281, opened 2019-06-19
Better menu management.

### LINKED ISSUE #135: Better menu management
In d7 we used a collection on contrib to allow users to place nodes into the menu system more intuitively. It would be nice to improve this for d8

## Projects
- https://www.drupal.org/project/bigmenu
- https://www.drupal.org/project/collapsible_dnd
- [Menu Link Weight](https://www.drupal.org/project/menu_link_weight) has a D8 alpha version that relies on several other modules (beta, alpha versions).

richardbporter:
Do we need bigmenu now? Or should we cross that bridge when we get there?

Resolves #135


joewhitsitt (approved):
Just tested this on HR. Seems to handle it (menu link form, node form), but it is a little slow to load and update. Not as slick as D7.

Re: Big Menu - max-input seems to be working for now. the big menu module solves a problem that 80% of our sites won't have and in turn sacrifices some UX.

richardbporter:
Works4me.

PR #284, opened 2019-06-21
Fix env check for GA settings.

### LINKED ISSUE #278: Google Analytics does not appear to work
The tracking js that does seem to load is the uiowa_bar stuff, not the Google Analytics module tracking code. gis.sites has both for test purposes and I am the owner of the site-specific tracking id.

joewhitsitt:
Resolves #278.


richardbporter:
@joewhitsitt Can you pull and run a `drush cim` to see if GA is uninstalled for you? May require a `cr` first....

richardbporter:
Actually I don't think it worked...

richardbporter:
@joewhitsitt That might have done it.

joewhitsitt:
```
  Configuration <em class="placeholder">google_analytics.settings</em> depend  
  s on the <em class="placeholder">Google Analytics</em> module that will not  
   be installed after import.
```


`blt ds --site=gis.sites.uiowa.edu` after `cr`

richardbporter:
We can't use Config Split to handle installing this module because it is listed for Config Ignore and that will aways take precedence.

richardbporter:
Another option would be to put this in the global settings file, but having test coverage for this is not a bad thing.

richardbporter (approved):

Issue #289, opened 2019-06-24
Reduce events cache time and remove configuration option


richardbporter:
Why we let people set this is beyond me.

PR #293, opened 2019-06-24
Set cache time to constant, removed config option

cory-skeers:
- Removed the config form option (and its default profile setting)
- Set cache creation to 15 minutes

I believe this is all that is required, and that the cache value isn't referenced elsewhere.

Chose 15 minute cache as a practical minimum, as it seems a short enough time that updates will be available in a short time, while keeping most of the benefits of caching. Let me know if this should be changed, or if there's a more specific approach to choosing cache time.


cory-skeers:
Resolves #289

richardbporter (approved):
Perfect. Thanks!

Issue #134, opened 2019-02-20
No feedback form built into site.


adamdelaney:
It would be nice to have a form to collect feedback.

briand44:
I think the thought behind this was an easy way for customers to provide feedback on the service directly through their site. So this would be some kind of form that is only available on the admin side that would email its-web. Other options could be just to have a feedback form on the sitenow site and direct customers to that.

richardbporter:
This could be as simple as a link to a workflow form.

richardbporter:
But we should evaluate all our methods of communication so that we are not making life harder for ourselves.

joewhitsitt:
We talked about this a bit today. Without knowing what the current Standard form is doing or any usage stats here are my thoughts.

For this form if implemented, we need to make sure it is going to one place and even for custom sites, removing the responsibility of the project developer. If that means it goes to the helpdesk/sharewell or workflow I don't know. Ultimately, this won't solve the "i can't login" issue or prevent customers from reaching out to a developer they know directly.

If we go the helpdesk/sccm/sharewell method, we should revisit how we are using that system. pinging slack about taking on a group issue has not be very efficient/effective. We should explore how other groups are handling it and/or experimenting with assigning tasks individually.

Should we coordinate with the rest of CSI in where these are routed/picked up so that we can continue to integrate with the rest CSI where it makes sense?

In the past, Romy has stressed the importance of individual contacts for custom sites and we can keep that going as we work on sites but support after go-live should go through this method. Customers are redirected around after someone leaves, is on leave or is no longer responsible for the project. It also allows students or folks with free time to jump in and assist when they can without forwarding emails/attachments/contact-info.

We should capture as much information as we can from the start (contact info, role, website URL, screenshots, affected URL(s)).

I am not very familiar with Workflow but it sounds like a good tool for site requests. A site request is a multi-step process that may need to go backwards (more info, denied, approved) before it branches out in several directions (hostmaster, developer). Support requests are often more linear and often require a more informal approach.

cory-skeers:
The current D7 SiteNow Standard contact form sends its email to its-hd-sm@uiowa.edu. It also creates a log in the watchdog, which could be used to find overall usage numbers--not sure if/how the email might get logged, to see what kinds of issues are being raised through it.

    watchdog('SiteNow Support', 'Support request sent by %hawkid', array('%hawkid' => $form_state['values']['hawkid']));

richardbporter:
[...]

PR #297, opened 2019-06-26
Support docs

### LINKED ISSUE #288: Add link to documentation in toolbar
Much like the D7 SiteNow service

cory-skeers:
Resolves #288 

Note:#134 is related but needs more discussion first.

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.profile
+++ b/docroot/profiles/custom/sitenow/sitenow.profile
@@ -589,7 +589,7 @@ function sitenow_page_attachments(array &$attachments) {
  */
 function sitenow_toolbar() {
 
-  $url = Url::fromUri('https://docs.sitenow.uiowa.edu');
+  $url = Url::fromUri('//docs.sitenow.uiowa.edu');
 
   $items = [];
   $items['support'] = [
@@ -601,13 +601,15 @@ function sitenow_toolbar() {
       '#options' => [
         'attributes' => [
           'title' => t('Opens help documentation in a new window'),
+          'id' => 'toolbar-item-sitenow-help',
+          'class' => [
+            'toolbar-item',
+          ],
+          'role' => 'button',
           'target' => '_blank',
         ],
       ],
     ],
-    // Devel is weighted 999. This can be altered later
-    // if additional toolbar options are needed.
-    '#weight' => 1100,
   ];
 
   return $items;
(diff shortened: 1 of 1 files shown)


joewhitsitt:
Just looking at the code: I don't know if the weight needs to be that high. Devel is not a production module and is only seen by administrators. I'd also be interested in knowing whether we can get away with passing the URL as `//:docs.sitenow...` instead of explicitly stating https.

Will try and test locally soon.

docroot/profiles/custom/sitenow/sitenow.profile
```diff
@@ -583,3 +583,32 @@ function sitenow_page_attachments(array &$attachments) {
...
+      '#options' => [
+        'attributes' => [
+          'title' => t('Opens help documentation in a new window'),
+          'target' => '_blank',
+        ],
+      ],
+    ],
+    // Devel is weighted 999. This can be altered later
+    // if additional toolbar options are needed.
+    '#weight' => 1100,
```
joewhitsitt: I don't this is needed. Setting weights is slippy slope similar to z-index. I think it is fine if this is before the user, devel, and tour buttons.

joewhitsitt: If near-last on the toolbar is desired, the tour module has specific class alongside toolbar-tab to target and adds a float:right bit of css.

docroot/profiles/custom/sitenow/sitenow.profile
```diff
@@ -583,3 +583,32 @@ function sitenow_page_attachments(array &$attachments) {
...
+  $items['support'] = [
+    '#type' => 'toolbar_item',
+    'tab' => [
+      '#type' => 'link',
+      '#url' => $url,
+      '#title' => 'SiteNow Help',
+      '#options' => [
+        'attributes' => [
+          'title' => t('Opens help documentation in a new window'),
+          'target' => '_blank',
```
joewhitsitt, line 609: Similar to other toolbar items I inspected, it might be good to add `role=button` and `id="toolbar-item-sitenow-support"`

joewhitsitt (changes requested):
Sorry to nitpick.

docroot/profiles/custom/sitenow/sitenow.profile
cory-skeers: Looking at it with fresh eyes, I don't think the weight is needed--the class option is a good one to keep in mind for future reference though.

Removed the weight.

docroot/profiles/custom/sitenow/sitenow.profile
cory-skeers: Added the role and id attributes. Also added 'tool-bar' item class for further consistency.

Small detail--The button text is listed as "SiteNow Help" so created an id of "toolbar-item-sitenow-help." Minor, but is there a preference between 'Help' vs. 'Support?'

joewhitsitt (approved):

PR #300, opened 2019-07-01
Rename modules.

richardbporter:
@joewhitsitt Here is what I have started. Not working yet.


joewhitsitt:
with uiowa_paragraphs (and submodules) to sitenow_paragraphs, I was able to do a successful import while leaving old placeholder modules in the directory, but it looks like field data was either corrupted or deleted in the process. i haven't had time to investigate further.

joewhitsitt:
For modules with config, I think this needs to be done in multiple releases:

~~Step 1: pre-rename #303 - need to downgrade all of the "enforced" module dependencies related to these old module so that configuration is not removed when the module is uninstalled. This was ideal back when we thought we could "turn stuff off", but now it is creating a headache.~~
Step 2: move forward with this PR with the old modules still enabled in core.extension but leave placeholder modules in the directory with info files. create copy modules "sitenow_*" and enable them. update the config dependencies to the sitenow_* versions
Step 3: update core.extension to remove the placeholder modules
Step 4: remove modules from application

I want to test a bit more with step 1 and 2 before moving forward but just wanted to write out what I have been successful with so far

joewhitsitt:
i think uiowa_paragraphs => sitenow_paragraphs is done with regards to step 1/2

richardbporter:
Looking at uiowa_events - it seems like there is some cruft code that could be deleted during this rename.

richardbporter:
Still need to change the field dependencies and enable sitenow_events.

richardbporter:
Left the uievents cache bin as-is because I see that going away with JS eventually.

richardbporter:
I broke it. :( Switching to workflow form.

richardbporter:
Working for me locally now. 👍

dnkemper (commented):
Looked at SiteNow Events module
- sitenow module installed
- uiowa module still installed
- sitenow event has same content in prod (where using uiowa event module) yay!
- can create new events
- can update existing events

Issue #254, opened 2019-06-10
Kaltura Support


joewhitsitt:
Ideally we would go the oEmbed route for this, but requires MediaSpace (coming this fall for uiowa)
https://knowledge.kaltura.com/mediaspace-oembed-integration

Kalutra/MediaSpace are not currently in the [oembed providers list](https://oembed.com/providers.json). Probably because the domains are unique to the customer (e.g. https://media.unh.edu/media/myUNH+Portal+Overview.mp4/1_azlfikak). To implement as a remote video provider we would need: https://www.drupal.org/project/drupal/issues/3008119

Below is some untested google brain dumpage:
[GATech](https://drupal.iac.gatech.edu/project/iac_mediaspace) and [UNH](https://www.unh.edu/it/kb/article/mediaspace-embedding-media-in-a-web-page-or-powerpoint.html) both have Drupal solutions but not necessarily the remote video method.

Here is a sandbox project to create a separate media entity type: https://git.drupalcode.org/project/media_entity_kaltura

joewhitsitt:
Update: The service owner provided me with a mediaspace URL to test which I played around with in D7. The oEmbed is coming over with details but without any markup to actually display the video. A ticket with Kaltura has been created I'm told.

joewhitsitt:
Marking as blocked. Dave is talking with Kalutra support about this. I have setup a D7 site for him to use to demonstrate that other college's endpoints are working and that the uiowa endpoint is not.

joewhitsitt:
This should now be working. Kaltura changed configuration on their end.

PR #302, opened 2019-07-02
Kaltura Support

joewhitsitt:
Related to #254

Blocked as it doesn't work yet because of something technical on the UI/Kalutra side of things. Swapping out the UofI details for another MediaSpace customer worked.

Relies on an uncommitted core patch that Nebraska also needs otherwise we will have to maintain our [own list of providers](https://oembed.com/providers.json)

Screen Shot 2019-07-02 at 2 31 02 PM


joewhitsitt:
This should now be working. Test using videos from: https://mediaspace.uiowa.edu/

joewhitsitt:
This also re-fixes the EXIF issue. The function didn't get updated to "sitenow_media_wysiwyg"

dnkemper (approved):
Works across all browsers in all sizes and if a user wants to start video at certain time, it does it correctly. Awesome!

PR #303, opened 2019-07-02
downgrade enforced dependencies

joewhitsitt:
I think this needs to happen first before we try to rename/replace modules that have configuration associated with them.


richardbporter:
What do I test here?

joewhitsitt:
basically just make sure paragraphs data is not lost and that you can create new paragraph content

aka CIM should not be deleting/creating these files, just updating them

PR #304, opened 2019-07-02
Hook preSync command to unset multisite route

### LINKED ISSUE #301: Syncing a site with BLT while server is running routes incorrectly
The RS proc should be stopped prior to syncing much like https://github.com/uiowa/uiowa/blob/master/blt/src/Blt/Plugin/Commands/MultisiteCommands.php#L271.

richardbporter:
Resolves #301


joewhitsitt (approved):
This magic appears to resolve my issue.

Issue #290, opened 2019-06-24
Port sitenow_dcv to D8


richardbporter:
I'm sure we will need this sooner than later. Currently on D7 SiteNow platform.

joewhitsitt:
https://github.com/ITS-UofIowa/sitenow_dcv

PR #338, opened 2019-07-18
Dcv module

cory-skeers:
Adds DCV file option resolving #290, based on D7 Sitenow DCV module.


richardbporter:
Nice work! I renamed this to fit our new naming conventions and refactored it to use Drupal best practices (ex. DI) as well as the temporary file system. Thanks!

richardbporter (approved):

richardbporter:
I also switched from the config API to the state API since this file should not be tracked, i.e. it should not be deployed to other environments.

PR #373, opened 2019-07-25
add pathologic

joewhitsitt:
This ignores site specific config (base urls).

HR is about to launch with 40-80 absolute urls that point to hr.prod.

This needs testing. On the surface, it still does what we need it to do, however this module is in alpha and has some bugs in the issue queue: https://www.drupal.org/project/issues/pathologic?categories=1&version=8.x


joewhitsitt:
This task doesn't have an issue, so I am assigning it to the project for review.

joewhitsitt:
I am not necessarily against pathologic if it tests well, but I did see this article which would allow us to do a go-live checklist and just automatically update any absolute URLs before launch and thus not rely on the pathologic module:
https://kevinquillen.com/mass-updating-drupal-fields-mysql-stored-procedure-pt-2

PR #374, opened 2019-07-25
Brand Lockup Feature

joewhitsitt:


richardbporter (commented):
The brand core controller "should" be using dependency injection to access services. Similar to form docs: https://www.drupal.org/docs/8/api/services-and-dependency-injection/dependency-injection-for-a-form

It should also return a response object instead of doing output buffering and returning markup.

Issue #366, opened 2019-07-25
Allow <span> element in WYSIWYG


joewhitsitt:
Add the span element with class `<span class>` as an available filtered_html html element. Update configuration.

### Test
New site install and existing site make sure the configuration gets added an you are able to place a span tag in the HTML source of a WYSIWYG, save it and not have it stripped from the rendered markup.

PR #377, opened 2019-07-26
Allow <span class> in config

cory-skeers:
<span class> (and <span>) added to the allowed html for filtered_html text format.

Tested locally and the change worked for a new site provision as well as an existing site (after a config synchronize).


cory-skeers:
Resolves #366

joewhitsitt (approved):
#works4me

Issue #359, opened 2019-07-25
Person - Hide from listing


joewhitsitt:
### Problem
There might be people added to the site that should not for whatever reason be listed in the /people listing (#358). Blocked until 358 is farther along so you can modify the view.

### Proposed Solution
Create a boolean field or "Person Options (e.g. Page Options)" that is off by default that toggled on will filter that person from the people listing view page. The view page will have to be modified to include this filter.

### Test
Create a person that initially shows up in the people listing. Then check the new boolean to hide person on people page and see if that worked. Uncheck the boolean to revert. Check again.

joewhitsitt:
Now that the view is a bit further along on #358, i've started work on this as well.

joewhitsitt:
This is more or less done, but tied up as part of #358.

dnkemper:
This works correctly! Can hide/display people as you please.

joewhitsitt:
this was merged in #405

Issue #358, opened 2019-07-25
People View/Configuration Pages


joewhitsitt:
### Problem
Though we are planning to provide a way for editors to display an abitrary list of people (featured content #247) it has also been requested that there be a default alphabetical listing of people that works similar to the SiteNow Articles view page.

### Proposed solution
- [x] Create a view page that default sorts people alphabetically by last name that lives at /people.
- [x] Create a configuration page similar to SiteNow Articles that allows webmasters the ability to enable/disable the page, change the path/title, provide header content, reorder the global weight of people (similar to Standard service's configuration using draggableviews).
- [x] Teaser listing for MVP, possibly card grid and table later on.

Draggableviews is a mess... Ideally, this module would fix some things (all of these have issues i believe but development is stalled):

- Show weight values
- Add people to bottom of the list
- Allow separate views contextual filtering sorts

_Considerations should be made with regard to future APR integrations._

### Test
On site install, the /people page should render if person content has been created with a default sort by last name. Webmasters should be able to access a configuration screen that allows them to configure the options listed above, include the ability to change the order globally for people.

joewhitsitt:
I tried out entityqueue in combination with auto_entityqueue. Didn't love it for the use-case of a global management of people order. Going to look at draggable views again with the idea of adding a "reset to alphabetical" button or something.

After attempting to change sort order/change sort options on one view page display I am not thinking there there will be three people page displays (sticky/last/first order, sticky/first/last order and the arbitrary sort option based on draggable views or whatever) that will enable/disable from the configuration form.

dnkemper:
1. After creating people, the page automatically rendered
2. All three people page displays work
      i. Can drag and drop people manually when using customized sorting
3. Can change path/header of people page
4. Can enable/disable people listing
5. Customized order, sticky/first/last, and sticky/last/first displays all look similar
6. Looks good in large mobile, tablet, and laptop views
      i. **ONLY ISSUE** small mobile view: 
![Screen Shot 2019-08-15 at 12 53 21 PM](https://user-images.githubusercontent.com/16872949/63114931-b8827f80-bf5b-11e9-984f-02e6d8a9a67f.png)
      ii. medium mobile view works:
![Screen Shot 2019-08-15 at 12 54 08 PM](https://user-images.githubusercontent.com/16872949/63114975-d18b3080-bf5b-11e9-9065-17a44c9ab1af.png)

GaryRidgway:
[...]

PR #405, opened 2019-08-06
SiteNow People View/Config

joewhitsitt:
related to #358 and #359


joewhitsitt:
See #358 and #359 for testing instructions.

PR #416, opened 2019-08-06
Throw 404 if event instance not set.

### LINKED ISSUE #145: uiowa_events - single event chokes on cancelled event
uiowa_events single instance is choking on cancelled events with no date.

```TypeError: Argument 1 passed to uiowa_events_format_instance() must be of the type array, null given, called in /app/docroot/profiles/custom/sitenow/modules/uiowa_events/includes/uiowa_events.single_event.inc on line 51 in uiowa_events_format_instance() ```

Example:
/event/8286/0

Original event: content.uiowa.edu/node/8286

This affects D7 as well so a backport would be nice.

richardbporter:
Resolves #145


joewhitsitt (approved):
Testing locally and production, this works for the example: event/20231/15

Do you think this is backportable? https://github.com/ITS-UofIowa/uiowa_events

PR #438, opened 2019-08-09
Implement carousel paragraph bundle.

### LINKED ISSUE #357: Carousel Paragraph Bundle
### Problem
Site owners are expecting us to have similar slideshow and image gallery functionality out of the box. We more or less need a port of the snp_image_gallery within our paragraphs implementation.

### Proposed Solution
Create a paragraph bundle that allows the upload of multiple images (bulk upload would be a bonus). The bundle settings can then default as a fancybox/lighbox thumbnail gallery or be turned into a slideshow.

We should evaluate if the libraries we were using in snp_image_gallery are still what we should be using in Drupal 8 (slick slider, fancybox).
- [x] Images should be able to be reordered by the editor.
- [x] Slideshow
- [ ] ~~Thumbnail with modal displays.~~

Additionally, we should consider supporting captions/title text that users can toggle on to display as part of the modal or slide.

richardbporter:
Resolves #357


richardbporter:
A summary:

- Created a Minimal HTML text format to use with the carousel item caption field
- Added allowed_formats module to restrict usage of minimal format
- Created a carousel -> carousel item nested paragraph bundle to use within sections

I did not see much value in DropzoneJS for this feature since it didn't work in field context and the bulk upload form did not let users set the caption field one-by-one. Could be useful later on.

I did not go the fielded media type route since I thought that might cause more confusion with embedding images in the WYSIWYG. 

I did not use the existing plain text caption (field?) on the image media type since that seems to be only used in WYSIWYG context. I'm also not sure where it is coming from.

richardbporter:
New image style usage should resolve "jump" between slides. However, a responsive image would be ideal so it looks better across different viewports.

Also disabled autoplay.

richardbporter:
If the benefit of structuring widgets with paragraphs is controlling output and being able to migrate field data, then we should merge this as-is. Then, when we adopt the HDS, we can migrate the field data to Slick/other fields and get rid of Bootstrap.

joewhitsitt:
- Renamed carousel item to carousel image (future proofs for other carousel item types)
- Recreated responsive image styles, view mode (plan to use for non-carousel)
- Added styles to carousel image for slide by slide styles (set sane defaults bg-dark, semi-transparent)
- Fixed issues with responsive images always falling back to large image style.
- Styles to account for classes.
- Full-width slideshow option for power users (section must be large, carousel must be full)

dnkemper (approved):
Everything looks awesome! All browsers are consistent and full, half, third, fluid sizes look great. Quarter carousel can look weird if user has longer caption, but doubt that will ever be used. Great job!

Issue #173, opened 2019-05-13
Set default favicons in metatag config


joewhitsitt:
https://github.com/ITS-UofIowa/uiowa_favicon

see if campus is interested in the module or rename it sitenow_favicon

would be nice to extend to include some SEO image defaults (opengraph, etc.)

richardbporter:
Check if we do this with theme settings.

joewhitsitt:
For now, HR will be launch with a single .ico file of the uiowa dome. This can be a go-live task until we come out with a better solution.

richardbporter:
Metatag module can do this for Apple/Android. Windows mobile is dead. Closing in favor of re-scoped PR #440.

Issue #287, opened 2019-06-24
Allow webmasters to toggle display of certain theme settings


richardbporter:
This is currently possible on the D7 SiteNow service.

joewhitsitt:
Need to audit current options we provide and whether or not we should continue do so.

richardbporter:
They should at least be able to toggle slogan. Need to hide/remove:

-  BS JS/CSS
- Favicon b/c metadata overrides

Audit other D7 options.

joewhitsitt:
This is what webmasters currently see on the standard service:

<img width="975" alt="Screen Shot 2019-08-20 at 1 21 39 PM" src="https://user-images.githubusercontent.com/4663676/63373413-830bd680-c34d-11e9-8fc9-7d3e9136a5d4.png">

richardbporter:
There is no slogan theme setting. Nothing else makes sense for a webmaster to change. Closing.

PR #440, opened 2019-08-12
Update metatag configuration for default favicons, user access and improved SEO

### LINKED ISSUE #351: Finalize SEO Configuration
### Problem
We have thrown some modules at the content types hoping to check off SEO expectations, but I don't think we are there.

### Proposed Solution
Revisit what global and content specific options are available to the webmaster role. Run a page, article, person and the articles view page through SEO audit tools like Google lighthouse, Twitter cards, open graph to see where we need to improve.

Is it proper to have both schema article options available on the basic page content type?

- [x] View pages (articles, people) need meta description.

Establish SEO defaults for things like twitter cards (image, organization, etc.).

### Test
Above tools and maybe consult with Michael from Tippie to get a MVP thumbs up.

### LINKED ISSUE #451: Metatag configuration not restricted for webmasters
**Describe the bug**
There seems to be a lot more configuration available on the metatag admin pages.

**Expected behavior**
I would expect this extra configuration to be hidden for webmasters.

**Screenshots**
image
 
**Additional context**
I think we updated metatag recently and this could be new config that we need to hide.

GaryRidgway:
base work for #173

Resolves #351 
Resolves #451


richardbporter:
This configures views to use the default meta description which is now the site slogan. Not great but better than nothing. 

The individual views could be updated to provider more accurate descriptions.

richardbporter:
A follow up should be to update the form description of site slogan and allow webmasters to hide it via #287.

richardbporter:
@GaryRidgway I can't request a review from you but would appreciate one.

GaryRidgway:
I have tested this and determined it does indeed make the favicons work. No idea why i can't approve though.

richardbporter:
This does override a theme's favicon. I think if sites want a custom favicon, they will need to split metatags config.

richardbporter:
Got an UPDB to set existing article page meta descriptions but I'm not going to expose this on the config form now. I think that could be a future enhancement.

I think we should deploy this, review structured data with cloud tools and iteratively improve our SEO config.

richardbporter:
Sites with permission splits are not getting the config changes. 😢

PR #459, opened 2019-08-15
i413 Added audio media type

dnkemper:
i 413 Added audio media type


joewhitsitt (approved):
added custom site permissions and adjust some audio settings. update entity browser.

audio _should_ get a generic thumbnail, but i am having issues getting those locally.

PR #510, opened 2019-08-27
make some sick nasty updates making the animation better with better …

GaryRidgway:
…spacing


GaryRidgway:
This essentially fixes accordion styles that were not sitting well on thinner viewports. the +/x should not take up less space and be always in a reasonable position and not covering the label text.

joewhitsitt (approved):

PR #513, opened 2019-08-27
Fix ajax paging for sitenow articles paragraph

### LINKED ISSUE #456: Article AJAX pager not working and zero-based
**Describe the bug**
The article block view AJAX pager does not work and is zero-based.

**To Reproduce**
1. Add 2 articles.
2. Add article listing to page showing 1 items but enable pager.
3. View page and click page 2.
4. Page reloads with zero-based index.

**Expected behavior**
I would expect the page not to reload since the view is configured to use AJAX. I think this is the preferred behavior.

joewhitsitt:
fixes #456

Oh dear this is hacky, but it was the only way I could figure out how to do it. I couldn't figure out a way to programmatically change items per page and have ajax paging on the fly without it messing up the count from page to page.

The "work around" is to expose paging options to the user and then set those on the fly followed by hiding the exposed form with core's hidden class.


richardbporter (approved):
Yes, this works.

PR #538, opened 2019-09-09
added proper scoping for caption styling

GaryRidgway:


docroot/themes/custom/uiowa_bootstrap/scss/partials/03_components/media/_embedded-entity.scss
```diff
@@ -2,25 +2,26 @@
...
   @include media-breakpoint-up(sm) {
     margin-left: 1rem;
   }
 }
-figure.caption.embedded-entity figcaption {
+figure.caption .embedded-entity + figcaption {
   margin-top: 0.5rem;
-  @extend .figure-caption;
+  font-size:90%;
+  color:#737373
```
joewhitsitt: Thanks for trying to reduce extends. In doing so though, can we use color variables? https://github.com/uiowa/uiowa/blob/master/docroot/themes/custom/uiowa_bootstrap/scss/_variables.scss#L19

joewhitsitt (changes requested):

docroot/themes/custom/uiowa_bootstrap/scss/partials/03_components/media/_embedded-entity.scss
GaryRidgway: Yes of course, that makes a lot of sense. I just replaced it with what the extend had, but I will go ahead and change that.

joewhitsitt (approved):
thx!

PR #540, opened 2019-09-11
add config for articles more link/path fields

### LINKED ISSUE #171: Articles paragraph read more link option
Similar to events. If articles view enabled in site-wide settings use that path as default. Otherwise provide an other link option.

joewhitsitt:
fixes #171


dnkemper:
# Testing

- Style is same as events 'more' link
- Links to 'news' correctly
- If change link within paragraph, it goes to that link correctly

### Looks good!

dnkemper (approved):

PR #546, opened 2019-09-11
i496 - Default Content

dnkemper:


joewhitsitt (approved):
I made a few tweaks:

- The old paragraphs were still in the db during export.
- The about page, I combined the texts into one and modified the layout to compact things a bit.
- Section background image was using a low-res image (adding that file in the process). I changed the media ref back to the high-res and removed the low-res from even being installed.
- A few language tweaks (we don't have an auto feed of articles for example)
- Simplified event filters to just be Everybody and ArtsIowa to make sure we have a good flow of events.

I think this is ready.

Issue #551, opened 2019-09-16
Deprovision rdmevents.sites.uiowa.edu


joewhitsitt:
D8 Deprovision

richardbporter:
Why is this blocked?

joewhitsitt:
no longer blocked. was waiting on D8 deprovision steps #556

PR #577, opened 2019-09-23
delete_businsvc_printmail

ethanstjohn:


richardbporter (changes requested):
You can finish #551 as part of this also. Just delete that site too.

The drush alias should be removed: https://github.com/uiowa/uiowa/blob/master/drush/sites/businsvcprintmail.site.yml

PR #582, opened 2019-09-24
Full-Width Featured Images

joewhitsitt:
- Reduced some preprocesses
- Full-width featured images if provided
  - Should be mobile friendly regardless title length, breadcrumb length, other header crap. Famous last words.
  - Should account for hidden titles.
- Caption display and aria-describedby if provided on article.
- Re-styled primary tabs


joewhitsitt:
HR has some different regions setup, so they don't get the full-width functionality. They had article featured image so i changed that back without splitting more config.

dnkemper (approved):
The Featured Images are displaying correctly! Great job! Looks great!

PR #583, opened 2019-09-25
Finalize sitenow.uiowa.edu webform.

richardbporter:
This PR includes:

- A fix for authenticated user permissions in profile.
- A custom requester role for sitenow.uiowa.edu which has access to request form.
- Config split settings for sitenow.uiowa.edu for blocks, auth, requester role.
- Updated role weights for profile.
- Working request form.

Of note is that the URL form field key had to be renamed to uri to fix an uknown bug which prevented the form from submitting. There are some hidden fields that are computed off the URL value for use in emails or development.

A possible workflow might be to "assign" yourself via submission admin notes, flag for follow up if necessary and lock when site is provisioned.


richardbporter:
There is a bug that prevents non-administrators from submitting the form. Any value in the URL field gets cleared and the form errors out saying it is required. 🤷‍♂

richardbporter:
@joewhitsitt Can you test this?

dnkemper (approved):

joewhitsitt (approved):
I am now able to see the form as a requestOr ;)

richardbporter:
- [x] Validate site does not already exist
- [x] #585

richardbporter:
👍

PR #600, opened 2019-09-30
oops forgot to add this stuff

joewhitsitt:
featured image full-width is "broken" without this stuff...


joewhitsitt:
Screen Shot 2019-09-30 at 11 28 01 AM

Should look like this if featured image is supplied and page title is not hidden.

PR #604, opened 2019-10-01
I256

GaryRidgway:
This is a start for the webmaster API. do consider before merging that this is allowing anonymous users to see webmaster emails so lets be sure that we want to do this or have authentication put in place before it is merged


joewhitsitt:
Removed the webmaster_page_test doggy kint stuff, which fixed coding standards as well.

First glance, it seems to be pulling the last email address for all entries.

joewhitsitt:
I wished I had time to look at this earlier. I know time has been spent trying to get emails. I talked to Rick about this a bit yesterday. It sounds like getting the mail value back is tricky. hawkid@iowa.uiowa.edu is supposed to get routed and might be the easier path forward if we are still looking to move forward with the API. I see the benefit when paired with auditing a site for certain configuration, etc. I have reached out to the email team to see if we can comfortably rely on sending emails to the hawkid addresses instead.

joewhitsitt:
>Emailing to hawkid@uiowa.edu will deliver messages to any Office 365 user.   If they are routing to a personal address, that will not work however, you would need to use the alias (first-last@uiowa.edu).

richardbporter:
HawkID is all we need: https://apps.its.uiowa.edu/dispatch/help/populations#manualid

richardbporter (approved):
I removed the email alter hook and renamed the API to be more generic in case want to add more admin-type endpoints. I also changed the serializer to only accept JSON as I was getting a client error locally without that change.

Let me know what you think.

Issue #371, opened 2019-07-25
UIDS SiteNow Sub-Theme


joewhitsitt:
Leveraging the new UIDS (https://github.com/uiowa/uids), create a new component base theme for the SiteNow service that will replace uiowa_bootstrap for all sites but HR (for now).

More details will need to be flushed out for this or separated in to separate tasks. Should be addressed in conjunction with #260

- [x] #1304 Audit differences between paragraphs and components.
- [ ] Enable uids_base on SiteNow Demo and start adopting global regions
  - [ ] Move away from uiowa_footer by programmatically adopting uids_base equivalents.
  - [ ] Move away from uiowa_bar by programmatically adopting uids_base equivalents.
  - [ ] Superfish
- [ ] Initialize new uids sub theme
- [ ] Mimic uids where we need to for faster adoption (sections, column widths, carousel, featured content)
- [ ] See if we can reduce unused CSS: https://www.previousnext.com.au/blog/performance-improvements-drupal-8-libraries
- [ ] Contribute back to uids_base regarding site wide theme settings (e.g. #587, #94)
- [ ] Address #260 before going to far with design.
- [ ] Contribute back as much a11y work that has already been done (landmarks, aria, html5 semantics, headings)
- [ ] Fix SEO issues from our current theme:
  - [ ] https://developers.google.com/web/fundamentals/design-and-ux/responsive#optimize_text_for_reading
  - [ ] https://developers.google.com/web/fundamentals/accessibility/accessible-styles#multi-device_responsive_design

joewhitsitt:
Work has started to compare the differences between SiteNow paragraph bundles and existing UIDS components #1304

joewhitsitt:
Marking as blocked. Good info, but work is being done to just use uids_base instead of sub-theming

joewhitsitt:
Much of this is either done or will be handled later through their respective tasks. Marking for reference but closing. Further work is being done towards this in #1324 and #1660

PR #640, opened 2019-10-09
i587 - Scroll to top functionality

### LINKED ISSUE #587: Scroll to top option for D8 Sites

dnkemper:
Resolves #587  - Scroll to top functionality


richardbporter:
This should be configurable as a theme setting but we'll need #371 to expose theme settings to webmasters.

richardbporter:
Unfortunately, I think we should close this in favor of adding it to a UIDS-based SiteNow theme. I don't think adding more features to uiowa_bootstrap is where we are headed. Some work has started on a UIDS base theme already. As that progresses, we can return to this issue.

PR #643, opened 2019-10-10
Redirect veterans.org.uiowa.edu to veterans.uiowa.edu/uiva

richardbporter:
This PR also updates our www/https redirects as per the Acquia docs. Mostly just looking for a review of typos... :)

Deployed to dev without issue.


richardbporter:
Also, I'd prefer to not add like "Date added" in comments since that is available Git. We can just run `git log --before="6 months ago" --grep="redirect" -- docroot/.htaccess` periodically or something.

briand44 (approved):
looks good. Did you look into other options outside of htaccess? Did you try the Redirect Domain submodule?

richardbporter:
Nah, this was the simplest. Thanks!

PR #675, opened 2019-10-24
Append version to sitenow.info.yml post-artifact:build

### LINKED ISSUE #645: Status report does not show install profile version
It would be nice to see what version of sitenow is deployed.

image

richardbporter:
Resolves #645


richardbporter:
This should be squashed when finished.

PR #737, opened 2019-11-12
Article source/organization fields

### LINKED ISSUE #365: Article - Source/Organization Label, Source Link
### Problem
Not all articles are written by individuals, some are written by an group/org/etc. Additionally, not all articles originate from the site, some are lifted from other sites like IowaNow. There needs to be a way to capture and display this.

### Proposed Solution
Create two fields, text field and a link field (link only), as both fields are optional and can be used with or without each other. Display the text field with no label. If a source link is provided, wire it to the text field.

Conditional states should display the source link if the source/organization field is filled out.

This should be incorporated into the byline area that already exists.

The help text should also encourage that if this is an external source that the metatag canonical URL should be specified.

### Test
An editor should be able to populate the organization and source URL and have it render appropriately on the node teaser and full node.

cory-skeers:
Resolves #365 

Don't like how much of the display logic is handled in the twig templates rather than a preprocess, but the fields and groundwork are there.


joewhitsitt:
Looks like this is failing because not all the config got packaged out: field.storage.node.* for each new field instance. Hopefully you have these local @cory-skeers or we might have to recreate them from scratch.

cory-skeers:
Yeah, nope--I'll have to recreate them.

joewhitsitt:
I think this is ready for review.

This PR adds three separate fields (Source Organization, Source Link and Link directly to source) that are all optional and work in conjunction with each other.

HR had article full template split so it was updated to work with the new functionality.

Organization filled = Added plaintext to byline.
Source Link filled = Added link to byline, Link directly to source option is displayed.
Org/ Link filled = Linked org added to byline.
Source Link/Link direct = Link is displayed on full byline, article title path is replaced on teaser and link is hidden.
Source Link/Org/Link direct = Linked org on full byline, org plain text on teaser, article title path is replaced on teaser.

This code should also account for a checked boolean but empty link, but could probably use more testing or be handled better.

dnkemper (approved):
Went down the list, and everything works accordingly! Great job!

joewhitsitt:
@dnkemper brought up a good question about the direct link option. how there isn't a way to click on to the full node.

> The use case for the direct link option is to avoid plagiarism. If used correctly, the article would really only contain the title, article summary, authoring information and any SEO meta tags that support the use case. It was a common pattern on some of our D7 custom sites that wanted to display Iowa Now articles or reference a local paper article mixed in their news listings

dnkemper:
> @dnkemper brought up a good question about the direct link option. how there isn't a way to click on to the full node.
> 
> > The use case for the direct link option is to avoid plagiarism. If used correctly, the article would really only contain the title, article summary, authoring information and any SEO meta tags that support the use case. It was a common pattern on some of our D7 custom sites that wanted to display Iowa Now articles or reference a local paper article mixed in their news listings

Should we add this to the documentation?

joewhitsitt:
Yes, probably, in other words.

PR #738, opened 2019-11-12
Allow SVG image uploads

### LINKED ISSUE #130: Allow svg image upload
There appears to be a core issue: https://www.drupal.org/project/drupal/issues/2220381.

cory-skeers:
Resolves #130

### Notes

1. Permissions are not restricted. I believe any embedded scripts will not run during the upload process, or the render as it is converted to a JPG, but a more concrete answer on whether this could be a security risk is needed.
2. All images are converted to JPG with upon render with this (through an ImageMagick style)--I did not run into issues with this, but more testing may be needed.


richardbporter:
This is really cool! My only concern is that we will get requests for SVGs to be uploaded without conversion so that end users can download the SVGs.

richardbporter:
@joewhitsitt Do we allow zip file uploads like we did in D7? For svg, makerbot files and the like.

richardbporter:
I'd probably be okay with this if it was documented to the end user.

joewhitsitt:
zip files should be possible. brand requested it.

I don't know if I like the idea of everything (not just SVGs) being converted to jpeg. transparent pngs and animated gifs off the top of my head. haven't tested this to see how these are handled.

cory-skeers:
Starting work on limiting the JPEG conversion to SVGs on render, with the intention of allowing ZIP uploads should be possible in the case of needing to download the original SVG.

joewhitsitt:
just to clarify zip should only be for the document/file media type. not image

richardbporter:
I think the ideal scenario is SVGs are uploaded and then rendered as SVGs. 

Zip files are already allowed for document media types.

joewhitsitt:
Thanks for trying to do this @cory-skeers but I think this issue was ill-defined from the start and the current solution, though accomplishes the issue, does it in a way that might not fit the original use-case.

I believe the original issue stemmed from the HR project and the customer wanted to upload fontawesome icon svgs as the card image. We ultimately worked around it by telling them to convert the svg icons to png and added a card style to handle the icon aspect ratio. Even if this was accomplished without affecting other image formats (transparent png, wysiwyg original image animated gifs), we still have a potential security concern with the permanent file URI where imagemagick or wrapping in an img tag is not involved.

If anything, we should revisit uiowa_footer for the same security concern. The d.o issues I have read seem to say that while possible, should be limited in use (small circles of folks that can actually upload and where).

My vote is to stop work on this issue until we have a clear use-case.

cory-skeers:
Right, @joewhitsitt -- from what I'd seen as well, there's very little that can be done with ImageMagick to make it compatible with rendering SVGs apart from some form of conversion, and even the per-filetype basis has its issues in D8.

Thanks for the clarification on original use-case: closing pull request.

PR #739, opened 2019-11-12
Added moderation_state messaging

### LINKED ISSUE #331: Better Visual Indicator for Content Workflow States
### Problem
There is currently very little indication as to the state of content (published draft, etc..).

### Proposed Solution
Implement drupal messages that display when the content is in another state besides Published.

### Test
Using the new viewer role which does not have workflow transition permissions view a node in a draft/needs review/archived state and be able to tell that it is not published.

If using drupal messages, make sure the messages don't cache/follow you to other pages where the context is incorrect. This was an issue in D7.

cory-skeers:
Resolves #331 

Message displays for all users (that the page would render for--anonymous won't see an archived page, for instance). Felt the message wasn't intrusive enough to need to remove it even when the transition options are displayed.

### Note
From the issue: "If using drupal messages, make sure the messages don't cache/follow you to other pages where the context is incorrect. This was an issue in D7." I was not able to recreate this caching/following behavior.


joewhitsitt:
I haven't tested this, but `drupal_set_message()` is a deprecated function: https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/drupal_set_message/8.7.x

here is an example of the addmessage method:
https://github.com/uiowa/uiowa/blob/b30c22ef5384efdda823e9b87509765a9ce1582a/docroot/sites/sitenow.uiowa.edu/modules/sitenow_core/sitenow_core.module#L42

docroot/profiles/custom/sitenow/sitenow.profile
```diff
@@ -515,6 +515,28 @@ function sitenow_preprocess_page(&$variables) {
   if (!$admin_context->isAdminRoute()) {
     $node = \Drupal::routeMatch()->getParameter('node');
     $node = (isset($node) ? $node : \Drupal::routeMatch()->getParameter('node_preview'));
+    if (isset($node)) {
+      // Get moderation state of node.
+      $revision_id = $node->getRevisionId();
+      $revision = \Drupal::entityTypeManager()->getStorage('node')->loadRevision($revision_id);
+      $moderation_state = $revision->get('moderation_state')->getString();
+      switch ($moderation_state) {
```
joewhitsitt: Can I propose an alternative:
```
      // If not published (status = 1), display a message indicating the current workflow state.
      $status = $revision->get('status')->value;
      if ($status == 0) {
        $warning_text = t('This content is currently in a <em>"@moderation_state"</em> state.', [
          '@moderation_state' => $moderation_state,
        ]);
        \Drupal::messenger()->addWarning($warning_text);
      }
```

joewhitsitt: This keeps the language consistent across states, less lines and allows for additional moderation states in the future. Even moderation states that transition the node to status 1.

cory-skeers: Updated to check on status rather than moderation_state, and updated language (I see now there's a nice example right above this in _sitenow_prevent_front_delete_message...)

Made a minor tweak: seeing '...a "archived"...' made my skin itch. Adds two extra lines, but keeps in line with allowing additional moderation states in the future without requiring updates.

joewhitsitt: haha, i didn't think about that. creative workaround.

joewhitsitt (approved):

richardbporter (approved):

PR #743, opened 2019-11-13
Implement config_split_ignore.

### LINKED ISSUE #736: AC config overridden in production
This causes code deployments to fail.

https://www.drupal.org/project/acquia_connector/issues/2864697

Possible solution:
- Move AC config out of splits and enable in sync
- Ignore AC config
- Move hide_signup_messages out of override and into default config
- ~~Re-enable BLT toggle modules to uninstall locally~~
- ~~Document process for ignoring config from modules and enabling/disabling with toggle~~

This should also be done for google analytics.

richardbporter:
Resolves #736.

This leverages config_split_ignore to ignore AC and GA configuration while allowing them to be installed/uninstalled according to the environment splits.

I'd like to leave the uiowa_seo config override in place as a fail safe in case GA gets exported incorrectly. https://github.com/uiowa/uiowa/blob/master/docroot/modules/custom/sitenow_seo/src/ConfigOverrides/GoogleAnalyticsOverride.php


richardbporter:
@joewhitsitt https://github.com/uiowa/uiowa/pull/743/files#diff-b49a60a26091987afebf6eb46b02cf8a

PR #759, opened 2019-11-18
Implement node_revision_delete.

### LINKED ISSUE #110: Implement revision clean-up with custom or contrib module
There are issues in D7 with nodes that have "too many" revisions, and so we've been using node_revision_delete to clean-up old revisions automatically. We anticipate that there will be similar issues with D8 sites which have "too many" revisions for an entity.

We should investigate possible solutions for cleaning up revisions in D8, either by number of revisions or by time since the revision was created.

richardbporter:
Resolves #110


joewhitsitt:
I tested this on HR. With running node_revision_delete and following with the paragraphs orphans process, it reduced the database by 72%. Afterward, I was able to work with nodes where I had reverted to three-month old version.

Things to note
- reverting creates a copy of the old version. Using the copy as the latest, the old version can get deleted. There is a reference to the old version (by date) but the copy doesn't depended on it and doesn't break when the old is deleted.
- No matter how many transitions a workflow has, there are only two that are important to the process. The published live version and the latest version regardless of what non-published state it is in. You can't have more than one non-published state.

joewhitsitt (approved):

joewhitsitt:
A little worried about the patches, but seeing the 72% reduction makes me thing we need this sooner than later.

richardbporter:
Yea not ideal to rely on so many. The ERP patch is fine since it is just a Drush command. We don't have to do anything with that yet.

PR #768, opened 2019-11-20
Remove config_split_ignore.

### LINKED ISSUE #763: config_split_ignore deployment resulted in data loss
As a follow up to #743, the deployment caused data loss for at least Google Analytics configuration in PROD.

We need to poke around to find out why this happened. All releases should be on hold until we have a solution in place to prevent this from happening again.

I don't know if this is related:
https://www.drupal.org/project/config_split_ignore/issues/3084022

Setting $split = 'prod' doesn't cause the problem locally, but setting the split to prod and then running csim prod after a sync does result in config loss.

richardbporter:
Resolves #763.


joewhitsitt (approved):
😰

PR #775, opened 2019-11-22
i720 - Created Taxonomy Term Reference

dnkemper:
-Added a TX Term Browser and view so users can call existing TX terms from all existing Taxonomies on their site and create new Tag TX terms in Basic Page CT and Article CTs


joewhitsitt:
I haven't tested this yet, but glancing at the code, this will result in data loss for existing sites because the field_reference field is removed and replaced without moving content.

I would be willing to work with you on a more sustainable solution.

joewhitsitt:
trying to clean up pulls. #778 is the current branch of this right? can we close this one @dnkemper ?

PR #777, opened 2019-11-22
Add full_html text format

### LINKED ISSUE #776: Add full_html text format
## Problem
We tried to live without, but let's face it, there are use-cases to override and place "dangerous" html.

#478 , dispatch sign-up, an 11th hour UStream player placement for the Prez.

At least this time around we have more than just the body field to override and place stuff so we don't have to worry about locking people completely out of a page.

joewhitsitt:
Resolves #776


richardbporter:
Tested adding my signup form locally. Webmaster cannot edit text but edit the section, text settings, etc. 👍

richardbporter (approved):

PR #778, opened 2019-11-25
i720b - Updated Relationship tags 

dnkemper:
Added taxonomy browser for relationships so users can add new and existing tag terms


joewhitsitt:
Okay, I am working on this, trying to get HR split working in a sustainable way.

joewhitsitt:
Yikes, I went crazy... Sorry.

The big change with this was to allow sites to split this without too many configuration files and then get HR "fixed" and updated when it came to FAQ Categories.

joewhitsitt:
Requesting other reviewers since i am now biased.

## Test
Users on a standard site should be able to add tags and select existing from the page and article content types. On HR, users should be able to add/select existing Units.

**Note:** You can add terms of the same name so there is the possibility of duplicates, however the same thing can happen when visiting the taxonomy listing page and clicking add term without reviewing all the existing terms...

IEF is considering options to disable edit/remove functionality, but it isn't ready yet.

joewhitsitt:
We will need to update documentation for this

joewhitsitt:
Need to account for no results in "Add existing." it causes an error.

joewhitsitt:
Applied a patch which seems to work. Would be nice if people tested again.

dnkemper:
Joe's patch worked great! Added a message that will display when there are no existing taxonomy terms to select. So users know why the 'Select References' button sends them back to the 'Add new Taxonomy Term' 'Add existing Taxonomy Term' options.

joewhitsitt:
We are going in a different direction

PR #779, opened 2019-11-25
i721 - updated content types and paragraphs to handle sticky

dnkemper:
i721 - updated content types and paragraphs to handle sticky


joewhitsitt:
trying to clean up pulls. #780 is the current branch of this right? can we close this one @dnkemper ?

PR #780, opened 2019-11-25
i721b

### LINKED ISSUE #721: Consistent use of "Sticky"
## Problem
Late in the game, we used sticky for the person content type to assist with people order. We should revisit article and basic page to offer the same kind of thing. For featured content, we should be able to then filter by sticky as well.

For events, we should see if a "featured" or something similar can be added/used.

dnkemper:
Added sticky permissions for page and articles and added sticky filter option for featured content and articles

resolves #721


joewhitsitt:
Something is wrong with `core.entity_view_display.paragraph.articles.preview` which is why the pipelines job is failing.

joewhitsitt:
taking a look at this

joewhitsitt (approved):
added permission splits for custom sites

PR #792, opened 2019-12-03
Provision doublebass.studio.uiowa.edu

dnkemper:


joewhitsitt (approved):
batching these in the same PR will prevent merge conflicts with sites.php. let me know if you need help resolving the conflict

PR #807, opened 2019-12-04
add hide breadcrumb option

### LINKED ISSUE #805: Hide breadcrumbs option
## Problem
There is a page option to hide the title, but this leaves the breadcrumb. For small sites the breadcrumb is a visual disruption and serves little purpose when the site title and top navigation already provide everything that is needed.

We have received this request a few times now.

## Proposed Solution
Similar to the hide title option, we should provide a hide breadcrumbs option on the page content type. Breadcrumbs will still render to the DOM for non-visual information and navigation but will be hidden visually.

joewhitsitt:
This consolidates the hide page title code as well by adding the sr-only class to a block level element.

Resolves #805


joewhitsitt:
Closing PR for now. Leaving branch. I had some thoughts on why we didn't previously do it this way and will comment in #805

PR #810, opened 2019-12-05
Updated Article tour to include new fields

dnkemper:
Added Organization, Source Link, and the 'Link directly to source' option to the Article tour


joewhitsitt (approved):
opted for 'display' instead of 'view' and remove /news stuff since the SiteNow Articles path is configurable and paragraph listings can be created.

PR #812, opened 2019-12-05
Redirect after login to tour.

### LINKED ISSUE #747: Demo content requires authentication but does not indicate so
Possible solutions:

- Add some basic text
- Adjust URLs to use destination parameter
  - Ex. `saml/login?destination=/admin/content/?tour=1`

joewhitsitt:
resolves #747

This will only work in dev/stage/prod because SAML is not local.


richardbporter (changes requested):
There are a few more links to admin pages in the demo content.

richardbporter (approved):

PR #826, opened 2019-12-10
added print statement to the console to tell a user if the site is a …

### LINKED ISSUE #641: Add sitenow identification for Help Desk
This would help for site identification with HelpDesk tickets and doesn't seem like too intrusive. Could just be a link to the new SiteNow site.

Don't think this belongs in the uiowa_footer module.

GaryRidgway:
…sitenow v2 site

resolves #641


docroot/profiles/custom/sitenow/js/sitenow_detector.behaviors.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @file
+ * Chosen.
+ */
+(function ($, Drupal) {
+    Drupal.behaviors.sitenowDetector = {
+      attach: function (context) {
+        $('meta[content="SiteNow v2 (https://sitenow.uiowa.edu)"]', context).once('sitenowDetector').each(function () {
```
richardbporter: My only concern is that if we ever remove that metatag, possibly unintentionally, this will stop working. Could we just attach it to the document?

GaryRidgway: yes of course, my mind must have gotten stuck in needing the metatag for some reason. I'll change that.

richardbporter:
Tested locally. Looks nice.

richardbporter (approved):

PR #827, opened 2019-12-10
SiteNow Site Request - changes based on feedback

joewhitsitt:
Screen Shot 2019-12-10 at 4 50 16 PM

Screen Shot 2019-12-10 at 4 50 51 PM
Screen Shot 2019-12-10 at 4 50 56 PM
Screen Shot 2019-12-10 at 4 51 01 PM


briand44:
From what I can tell this looks good. I am curious as to what the help text says for the url field when existing domain is checked but as long as that help text is different and says something along the lines of enter your existing url, then I think this should be good to go.

briand44 (approved):

briand44:
We will want to communicate with Gayle so she can update her communication.

PR #830, opened 2019-12-11
Redirect legacy login paths.

richardbporter:
Including smart trim update in this PR.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -12915,11 +12915,11 @@
         },
         {
             "name": "uiowa/uiowa_auth",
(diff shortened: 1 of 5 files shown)


richardbporter:
@joewhitsitt More concerned with the smart trim update. I've tested the redirects.

joewhitsitt (approved):
i am not concerned about the smart trim update. it was needed for paragraph previews to strip html (full_html social embeds in particular). if you are okay with the redirects then go forth, you have my blessing

PR #831, opened 2019-12-12
Events followup.

richardbporter:
Removes the DI from the events controller in favor of the existing config method on the parent class. Adds sitenow_events.settings to config ignore.


joewhitsitt:
Good catches! Sry.

joewhitsitt (approved):

PR #849, opened 2019-12-19
Preprocess check for view enabled status

### LINKED ISSUE #847: Articles/People - More Link - Check for route
## Problem
If the articles or people view is disabled through the configuration interface, but there is an existing paragraph item with the more link set to use the default path, there is going to be a problem because when the view is disabled, the path does not exist (WSOD!).

## Proposed solution
Preprocess the paragraph to check that the path exists before trying to write a render for it.

## Test
With sitenow articles view enabled, create a articles paragraph item somewhere and configure it to display a more link but leave the path blank which will default to the view route. save the page and render it out. now disable the sitenow articles view (via SiteNow Articles configuration) and revisit the node that had the article paragraph. It should not WSOD or log an error/warning/notice.

cory-skeers:
Resolves #847 

If field_articles_more and defaulting to articles view for the URL, checks that view is enabled, else skips over creating template's $more variable.


joewhitsitt (approved):
I think this resolves the issue, but we may want to create a follow up issue to update help text and documentation. As a user, I may wonder why my more link doesn't show since I checked the box.

Issue #860, opened 2019-12-20
field_reference clean-up


joewhitsitt:
Blocked until #841 has been deployed.

- [x] Remove field_reference from sitenow profile. keep it in hr.uiowa.edu for now.
- [x] Remove References view.
- [x] Move remaining field_reference logic out of sitenow into hr_core.
- Update docs
  - [x] https://sitenow.uiowa.edu/node/121
  - [x] https://sitenow.uiowa.edu/node/126

joewhitsitt:
Needs testing once #841/#863 have been closed/merged.

joewhitsitt:
No tours mention the field_reference field.

Issue #841, opened 2019-12-16
Update Tags Functionality


joewhitsitt:
## Problem
#720 created a complex UX for mainly one site (HR) and #748 on the horizon, we are going to move things around a bit to create a better UX for internal tagging of content.

Blocked until #598 and #839 have been deployed to production.

## Proposed Solution

- [x] Recreate field_tags (core field) on person (#598), page and article (#839)
- [x] Update articles content block render to pull tids from field_tags instead
- [x] Update hook to move existing field_reference (node) tags values to field_tags
- [x] Change field_reference (node) to #access false until successful migration where it can be cleaned up from all sites except HR.
- [x] Limit field_reference (paragraph) to tags only, investigate what needs to happen for the same field on (HR's FAQs)
- [x] HR - Change references labeling to Unit

Several more steps involved, but that is currently what I think this entails. All workbench access stuff has been removed from this as we don't yet know the full requirements of dentistry.

joewhitsitt:
Follow up issue #860

joewhitsitt:
I ran a DSA on all sites over the holiday. looks like four sites used the field_reference field and all were successfully updated to use field_tags. moving forward.

PR #863, opened 2019-12-20
Update Tagging Functionality

joewhitsitt:
## Description
Initial step at removing field_reference on SiteNow websites in favor of a simpler approach.

For all sites except for HR, this will move all field_reference tags to field_tags, writes a revision message and saves the node (page, article, person). This also updates the articles paragraph functionality to use field_tags instead and limits/restricts access to field_reference until it can be deleted/split only for HR.

Four sites at the time of writing this use tags: hr, mnh, oldcap and sitenow.

There is a risk in validation errors if a node with field_refrerence containing tags is reverted, but there is limited use and a follow-up issue #860 to eliminate the possible validation/access errors.

This change should be communicated with HR.

## Related Issue
#841 

## Follow-Up
#860


richardbporter:
Tested `dsa`-ing all the sites mentioned with no problems.

Possibly unrelated but I noticed the article view filtering by tags acting wonky. Also, the people view seems to be weird as well.

richardbporter:
mnh.uiowa.edu
- Article "foo" tagged with "bar"
- Page with article listing filtering on tag "bar", configured to show 3 articles
- Article "foo" displays along with two others not tagged with "bar"

Is this expected?

joewhitsitt:
i will try to reproduce this locally, thanks for testing

richardbporter (changes requested):
I think the issues with articles is related to this PR. Not sure exactly what yet but master works fine.

richardbporter:
The people issue seems unrelated as it is currently an issue in prod. I have to re-save person nodes to get them to show up in the people view.

richardbporter:
A simpler article test case for the default site:

- Create two terms foo and bar
- Tag "Keep Moving Forward" with foo
- Filter articles on front page by tag bar
- The article "Keep Moving Forward" still displays

richardbporter:
Same issue in https://github.com/uiowa/uiowa/pull/863#issuecomment-569793722 still.

joewhitsitt:
@richardbporter it won't let me re-re-request a review... but i think i fixed the issue

richardbporter (approved):

PR #877, opened 2019-12-27
SiteNow Image Gallery

### LINKED ISSUE #724: Image Gallery Paragraph Bundle
I think Drupal 8.8 fixes multi-media-uploading (alt text) so we could move forward with this paragraph bundle.

Single paragraph bundle with similar global fields (column-width, unique id, styles) and a multiple media upload reference field.

for MVP, I think we need to pick a lightbox type solution to render square thumbs similar to https://github.com/ITS-UofIowa/sitenow_paragraphs/tree/master/snp_image_gallery .

If media image doesn't already, we will need a title or caption field.

joewhitsitt:
# Description
Resolves #724


joewhitsitt:
I think the current styles for this aren't working because it need the height: 100vh which impossible to specify in the section/content block the gallery resides in. When you set a specific height (reasonable), it overflows out of the container.

I didn't want to overwrite your styles, but I got the following styles to work though it reads top down in columns instead of left to write wrapping. Compare that to the editor interface and it could easily be confusing to folks who want a specific order:
```
.paragraph--type--image-gallery {
  .field--name-field-image {
      column-count: 2;
      column-gap: 1rem;
      @include media-breakpoint-up(md) {
        column-count: 3;
      }
      @include media-breakpoint-up(lg) {
        column-count: 4;
      }

    .field__item {
      display: inline-block;
      margin: 0 0 0.5rem;
      width: 100%;

      &:hover{
        img {
          opacity:0.28;
        }
      }
    }
  }
}
```
Screen Shot 2020-02-14 at 4 28 18 PM

_Note: Those are the original aspect ratios. That dark blue is really that vertical_

For the SiteNow service, I think the default should be a wrapping grid of focal point cropped squares and to meet the OSC request, provide a masonry style option that presents this un-cropped top down then right approach. Maybe later we can revisit with a complete JS solution/fallback.

pyrello:
@joewhitsitt No problem with overwriting my css. It was just meant as a starting point.

By having all the images crop to a standard width and then be the appropriate height for that is just going to cause confusion no matter what.

I've added numbering to the image you provided above and the ordering is not very straightforward if you could somehow get it to lay out as desired, but order left to right.

74574031-f5015900-4f47-11ea-9f14-5b6ca0f9a837

joewhitsitt:
Seeking a11y review here: #1104

joewhitsitt:
Added documentation (drafts):
https://sitenow.uiowa.edu/documentation/content-blocks/image-gallery
https://sitenow.uiowa.edu/node/156/latest

joewhitsitt:
Added known issue:
https://sitenow.uiowa.edu/node/196/latest related to https://github.com/uiowa/uiowa/issues/1157

briand44 (approved):

PR #918, opened 2020-01-13
Refactor BLT commands in preparation for combining.

richardbporter:
# Description
Moves GitCommands and ReplaceCommands to Uiowa namespace and refactors for shared usage.

We are currently [replacing two BLT commands](https://docs.acquia.com/blt/extending-blt/#replacing-or-overriding-a-robo-command) to extend their functionality. I moved these since I don't think we can replace the same BLT command multiple times. We'll want to put all those in one command file.

The `uiowa:git:clean` command deletes artifact branches from an Acquia remote since those never get cleaned up. We'll want to run that occasionally to reduce the size of our Acquia remotes. 

GitCommands also includes a post-command artifact build to write the repo tag/branch/sha to the info file so it is visible on the status report page. 

# Checklist
- [ ] Add appropriate PR labels.
- [ ] Request approval from multiple developers across units.
- [ ] Dependencies have been properly updated.
  - https://github.com/uiowa/uiowa#updating-dependencies
- [ ] Site config splits have been accounted for.
  - Go to https://github.com/uiowa/uiowa/find/master to find split config entities affected by this PR.
- [ ] Test PR locally with default site.
- [ ] Test PR locally with custom site that might be affected by these changes.
- [ ] Update documentation.
  - Document changes introduced by this PR in application, profile or site README.


pyrello (approved):
I reviewed the files but have not tested. Everything looks good to me.

PR #932, opened 2020-01-14
Move acquia credentials to custom uiowa key.

richardbporter:
All custom BLT configuration has been moved under an arbitrary uiowa key to avoid conflicts with other core BLT config. I forgot to move this config the first time around. This will require developers to update their local.blt.yml file.


pyrello (approved):
Looks good.

PR #935, opened 2020-01-15
Adopt DrupalVM.

richardbporter:
Initial work to adopt DrupalVM for SiteNow sites.


richardbporter:
Trying to get xDebug working...

pyrello:
I added a setting that sets up a wildcard ssl certificate. I think it will be necessary to run local sites against https.

richardbporter:
I think I'm going to start a new feature branch for this and re-run the updates b/c we have some new sites created.

PR #963, opened 2020-01-17
Feature: Updates settings files for sites on uiowa01.

### LINKED ISSUE #962: Fix db include statements for uiowa01 sites
When uiowa01 sites were migrated into this project as part of #916, I copied the database include statement verbatim from the other sites and `uiowa` is hardcoded in that path, so it cause the sites not to work on the uiowa01 container.

pyrello:
# Description
Resolves #962.

This resolves an issue specific to uiowa01 sites when they were added to this project. They were added with a database include statement that had `uiowa` hardcoded into the path and this causes the sites to not work because it should be `uiowa01` for those sites.


richardbporter (approved):

richardbporter:
We should refactor that test to use regex. Its not the greatest test but if that include gets deleted, the site goes down.

PR #965, opened 2020-01-21
Updated honeypot version

dnkemper:
Updated the honeypot version (Step 1 of 3 potential steps to prevent webform spam)

I think this is what was meant in meeting, not entirely sure


joewhitsitt:
I think this needed an updated composer.lock file.

joewhitsitt (approved):

PR #975, opened 2020-01-22
Make database include statement application-aware.

### LINKED ISSUE #961: Command should write database include specific to app container
This line in the Multisite command file hard codes the uiowa application container when writing the database include line into the settings file. This would ideally either use the appropriate application or (even better) it could be dynamically set based on some environmental variable so that we wouldn't have to change it when moving a site from one container to another container.

richardbporter:
# Description
Resolves #961

The large number of changes was just a find and replace for existing site settings files. Included some other minor fixes while working in these files.


pyrello (approved):
Looks good to me.

richardbporter:
I forgot the uiowa01 sites. Doing now.

pyrello:
My approval was a little premature as it turns out. This looks good to me now. I was able to deploy in dev on uiowa01 and checking a site there didn't result in an error because the include statement. I think this is good to go, unless you are thinking about additional updates.

richardbporter:
I think its ready. Thanks for testing.

Issue #974, opened 2020-01-21
Command sql-sanitize was not found


pyrello:
```
Command sql-sanitize was not found. Drush was unable to query the database.  
   As a result, many commands are unavailable. Re-run your command with --deb  
  ug to see relevant log messages. 
```

Ran into this when attempting to copy a database from prod to dev. Any ideas @richardbporter?

pyrello:
To add a little more context around this, I copied a database from prod to dev using the AC UI. It appears that copying a db triggers a cloud hook, which in turn runs the above command, which fails.

Here's the full output:
```
[20:17:51] [20:17:51] Started
[20:18:03] [2020-01-21 20:17:51] Active region not set for site uiowa01, using region of lowest numbered db.
[2020-01-21 20:17:51] Using dns server dns-768 (dns-768.prod.hosting.acquia.com) as resolver.
[2020-01-21 20:17:51] Active region not set for site uiowa01, using region of lowest numbered db.
[2020-01-21 20:17:52] Using dns server dns-768 (dns-768.prod.hosting.acquia.com) as resolver.
[2020-01-21 20:17:52] Active region not set for site uiowa01dev, using region of lowest numbered db.
[2020-01-21 20:17:52] Using dns server dns-768 (dns-768.prod.hosting.acquia.com) as resolver.
[2020-01-21 20:17:52] Executing: ssh ahbot@staging-15039 "mysql -h staging-15039 -u s58259 -pm6fRZHakBUTiKWe -e 'DROP DATABASE IF EXISTS `uiowa01dedb325910`; CREATE DATABASE `uiowa01dedb325910`'"
[2020-01-21 20:17:52] Executing: ssh ahbot@web-33201 "mysqldump -h dbmaster-33200 -u s58257 -pYmvqpbuz2665Yqx -qcK --ignore-create-error --single-transaction uiowa01db325911 | gzip > /mnt/tmp/backup-uiowa01db325911-20200121201752.sql.gz
[2020-01-21 20:17:57] Executing: ssh ahbot@web-33201 "cat /mnt/tmp/backup-uiowa01db325911-20200121201752.sql.gz; rm -f /mnt/tmp/backup-uiowa01db325911-20200121201752.sql.gz" | ssh ahbot@staging-15039 "gunzip | mysql -h staging-15039 -u s58259 -pm6fRZHakBUTiKWe uiowa01dedb325910"
[2020-01-21 20:18:02] Starting hook: post-db-copy
[...]

PR #977, opened 2020-01-22
Add Drush alias for uiowa01 application.

richardbporter:
I use these application aliases to quickly SSH in and poke around, e.g. `drush @uiowa01.dev ssh`. Also cleaned up some unused cruft from uiowa alias.


pyrello:
To test this, I checked out the pipelines-build-aliases branch on dev in uiowa01, then attempted to do the database copy again. I ended up getting the same error. So either the error is not because the site doesn't exist or it is not working.

richardbporter:
This is unrelated to #974. More of a heads up.

PR #982, opened 2020-01-28
Feature: Multisite command updates

### LINKED ISSUE #988: BLT Sync Not Importing Local Split Config
```
> drupal:config:import
[warning] BLT will NOT import configuration, /var/www/uiowa/docroot/../config/default/core.extension.yml was not found.
[warning] The drupal:toggle:modules command is disabled.
```
I was expecting stage file proxy to be on locally and had to run a `cim` on the site to import the already "active" local split.

pyrello:
Update: resolves #988

# Checklist
- [x] Add appropriate PR labels.
- [x] Request approval from multiple developers across units.
- [x] Test PR locally.
- [x] Update documentation.

# Changes
This PR moves the remaining Multisite commands into the `uiowa:multisite:*` namespace and generalizes them to allow them to be used for both the collegiate and sitenow profiles. In the `uiowa:multisite:create` command the `requester` parameter has been replaced with a `--requester` option and `profile` is now a required parameter. Additionally, a `--no-commit` option has been added - that is disabled by default - that allows the user to avoid have a commit created.

```
Usage:
  uiowa:multisite:create [options] [--] <host> <profile>
  umc

Arguments:
  host                           The multisite URI host. Will be used as the site directory.
  profile                        The profile that will be used when creating the site.

Options:
      --simulate                 Simulate database creation and filesystem operations.
      --no-commit                Do not create a git commit.
      --no-db                    Do not create a cloud database.
      --requester[=REQUESTER]    The HawkID of the original requester. Will be granted webmaster access.
```

No functional changes have been made to the `delete` or `execute` commands.


pyrello:
I went ahead and provisioned the site I need to start building as a way of testing this out.

pyrello:
@mark-bennett-uiowa and @bspeare In the meeting we had yesterday afternoon, we discussed changes to the ReplaceCommand, which effectively swaps out new logic for existing blt commands. The uiowa app is using this to replace the `artifact:update:drupal:all-sites` command with logic that runs a site install if the site has not already been setup. As it is currently set up, this logic only happens for sites that are set up using the sitenow profile. We discussed whether it would make sense to include the collegiate profile in this process as well. I believe that implementing this for collegiate makes sense for the reason that the alternative is to manually run the `drush @example.remote si collegiate` command for each remote environment.

Setting this up does not preclude being able to start working on a site locally before that happens, although this does require some precautions to prevent loss of work. If you start working locally on a site that has been provisioned but not merged via PR into the master branch, you'll want to make sure to exclude the site from the multisites array in your `blt/local.blt.yml` file. This will prevent any blt syncing commands that use that array from overwriting the database with the prod database if you happen to run a command like `blt dsa`. Also, you will have to manually sync the local database to the remote databases after the provisioned site has been merged and is present on the remote servers.

Any questions or concerns, please reply.

richardbporter:
I'll test this as soon as I can. One thing I know our group will need to update is the sitenow site request form. It prints the BLT command to run in the form submission. We'll just need to change it to the new syntax.

richardbporter (changes requested):
Overall, I think this looks great. One thing I noticed is that the request option should probably use the value `InputOption::VALUE_REQUIRED` otherwise a user can run `blt umc foo.com --requester` which doesn't make sense.

richardbporter:
I like the `no-commit` option. That is useful for testing. Do you think we should remove the check for the `simulate` option during database creation? It seems redundant now.

I agree with the `@todo` comments. Adding profile-named includes seems logical. The include file already has a check to see if the file exists. Ex.
https://github.com/uiowa/uiowa/blob/8e63ceb40e6351d3bb5d8d1406265a221ab13401/docroot/sites/billing.uiowa.edu/settings/includes.settings.php#L18-L22

pyrello:
@richardbporter I'm not sure I understand your point about the `simulate` option. 

Since the check to add additional settings file checks existence, maybe we should just change this to write it no matter what and it will only get included if it exists? Would it be a concern that it might create unnecessary additional overhead in some cases?

If the requester can be added to any site regardless of profile, then do we want to still store that value under `uiowa.profiles.$profile.requester`?

richardbporter:
What the options do seems to overlay a little bit. Should we remove the check for simulate here: https://github.com/uiowa/uiowa/blob/feature_multisite_command_updates/blt/src/Blt/Plugin/Commands/MultisiteCommands.php#L288 ?

That way its clear what each one does and people can use a combination of them for testing.

pyrello:
If we did that, then I think we would want to manually set the `--no-db` option to `FALSE` when `simulate` is selected. Otherwise, if someone selected `simulate` on its own, wouldn't a database end up getting created?

richardbporter:
Yes. I tested with `no-db` and `no-commit`. Do you think simulate is still valuable? if so, lets just leave it as is.

richardbporter:
I'm going to push a fix related to #988.

pyrello:
I guess `simulate` is of limited value, since it cannot truly simulate the provision because of the issue where the option doesn't get passed to the `blt rmi` command. However, it is a thing I always wanted because testing these commands is so annoying. Maybe it matters less with a fully functional `delete` command.

richardbporter:
Yea its annoying rmi doesn't respect that. Maybe we can figure that out later. Lets just leave it for now.

richardbporter:
> If the requester can be added to any site regardless of profile, then do we want to still store that value under uiowa.profiles.$profile.requester?

I think thats fine.

richardbporter:
I'm fine with removing this switch to just check if the requester config is set: https://github.com/uiowa/uiowa/blob/feature_multisite_command_updates/blt/src/Blt/Plugin/Commands/ReplaceCommands.php#L75

Do you want to invoke BLT `drupal:update` when a database is synced as well? If so, we should remove this switch as well: https://github.com/uiowa/uiowa/blob/feature_multisite_command_updates/blt/src/Blt/Plugin/Commands/ReplaceCommands.php#L117 This is useful when environment splits are set up.

pyrello:
I'm cool with both of those changes.

richardbporter:
@joewhitsitt I'm sure we have some Wiki docs that need to be updated after this?

richardbporter (approved):

pyrello:
Good validation catch!

PR #994, opened 2020-01-29
Refactor sitenow_migrate into feature split for use with DrupalVM with improved link handling.

cory-skeers:
Updates to sitenow_migrate to fix internal links during D7-> D8 migration.

Needs some debugging, as well as refactoring for all of sitenow_migrate with changes to local development.


docroot/modules/custom/sitenow_migrate/sitenow_migrate.install
```diff
@@ -25,3 +25,4 @@ function sitenow_migrate_uninstall() {
   }
   return TRUE;
 }
+*/
```
joewhitsitt: Should we just remove this. Git will have it if we need to add it back in later.

joewhitsitt:
I fixed some coding standards with codesniffer, but there are still some errors to fix before this passes checks:
```
FILE: /Users/jwhitsit/acquia/uiowa/docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
FOUND 11 ERRORS AND 8 WARNINGS AFFECTING 15 LINES
  25 | ERROR   | Class property $source_to_dest_ids should use lowerCamel naming without underscores
  25 | ERROR   | Missing member variable doc comment
  26 | ERROR   | Class property $d7_aliases should use lowerCamel naming without underscores
  26 | ERROR   | Missing member variable doc comment
  27 | ERROR   | Class property $d8_aliases should use lowerCamel naming without underscores
  27 | ERROR   | Missing member variable doc comment
  29 | ERROR   | Class property $base_path should use lowerCamel naming without underscores
  29 | ERROR   | Missing member variable doc comment
 237 | ERROR   | Variable "DRUPAL_7" starts with a capital letter, but only $lowerCamelCase or $snake_case is allowed
 238 | ERROR   | Variable "DRUPAL_7" starts with a capital letter, but only $lowerCamelCase or $snake_case is allowed
 258 | ERROR   | Variable "DRUPAL_7" starts with a capital letter, but only $lowerCamelCase or $snake_case is allowed
```

cory-skeers:
Access to the local D7 database copy needs to be updated to adopt DrupalVM.

See #1010

richardbporter:
I was able to connect to a localhost server, not DevDesktop, but had to remove the quotes from the port number config otherwise it seemed to default back to 33067. I'm guessing that is the DD default port number? I can test again with DD.

I also had to uninstall/install to fix a nondescript error about a missing plugin.

richardbporter:
This keeps cropping up when running any Drush migrate commands:
```
 The "" plugin does not exist. Valid plugin IDs for Drupal\migrate\Plugin\MigrateSourcePluginManager are: embedded_data,  
   empty, url, files, articles, people, pages      
```

richardbporter:
I can get around that uninstalling/re-installing all migrate modules but when attempting to run the d7_page migration, I get:
```
 SQLSTATE[42S02]: Base table or view not found: 1146 Table 'demo_sitenow_uiowa_edu.migrate_map_d7_arti  
  cle' doesn't exist   
```

Seems like some schema/config is missing.

cory-skeers:
Good news! (maybe). If it's erroring after pulling content, it's getting snagged at the fetchMapping() post-migrate function after the page migration is complete.

Added a dependency on d7_article from d7_page. If it is getting caught on fetchMapping(), running d7_article before d7_page should take care of it (the link checker requires all node mappings be in place prior to running).

cory-skeers:
Tried running it with the latest updates. First getting a `[2006] MySQL server has gone away` error, followed by later attempts getting the `The "" plugin does not exist...` error.

richardbporter:
It doesn't seem like configuration is being imported correctly. Everything works fine on the master branch with DrupalVM and the Vagrant gateway IP. The issue might 7154fb1 - was that change necessary?

cory-skeers:
`sitenow_migrate` working with DrupalVM; #1010 resolved.

richardbporter:
Is this ready to be reviewed and merged? Or are there additional changes coming related to the SQL regex?

cory-skeers:
This is ready to review. Everything seems to be working as expected in DrupalVM.

## Notes
The drush output lists nodes suspected to have broken links (has at least one link with the base path or 'node' within the href), as well as the old (and if a replacement was found) new links on a per-node basis as a way to double check its results.

Nodes which receive an updated content display full HTML on the page until the page is edited/saved. The content itself is unaffected by this. Think it's a result of setting a paragraph field directly, and part of the presave/revisioning process being bypassed, but not sure.

richardbporter:
I think I realized what I did wrong while attempting to move this config to a feature split earlier. I'm going to give it another shot.

richardbporter:
I moved this to a feature split since not every site, sitenow or otherwise, needs this enabled locally. What this means is:

- to enable, activate the split and run `drush config:import` for the site.
- to disable, deactivate the split and run `drush config:import` for the site.
- the import process on AC will no longer uninstall this config so it needs to be disabled before sql syncing.

Forgetting to import configuration after deactivating can lead to problems exporting configuration later. For example, deactivating the split and running config export will move the split config back to the sync.

I tested locally and it works as before.

richardbporter (approved):

PR #999, opened 2020-01-30
Only use existing-config option if using sitenow profile.

### LINKED ISSUE #998: Only use existing-config flag if using sitenow profile for auto site installs

pyrello:
Resolves #998, fixing a bug that was introduced in #982.


joewhitsitt (approved):
not tested, but looks fine

pyrello:
Tested on dev in uiowa01 and this fixed the issue that was preventing the site install of sites that didn't exist yet using the collegiate profile.

PR #1002, opened 2020-02-03
Provisions Feb 03

dnkemper:
Provisioned sites


briand44:
i haven't been following real close but I didn't think hostmaster approved cognitivecontrol.uiowa.edu

dnkemper:
@briand44 I owe you BIG time! Needed to refresh inbox and got a bunch of emails from hostmaster

briand44 (changes requested):
top level domains should be approved before provisioning

PR #1020, opened 2020-02-06
Added sidebar menu configuration for businessmanager.fo.uiowa.edu

dnkemper:
Apparently you have to hardcode changes in blocks. Darn. Added Sidebar menu to display all of their links like they requested.

### DIFF
--- a/config/businessmanager.fo.uiowa.edu/.htaccess
+++ b/config/businessmanager.fo.uiowa.edu/.htaccess
@@ -1,0 +1,27 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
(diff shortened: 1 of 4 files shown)


briand44 (commented):
I would like Joe and/or Rick to weigh in on this. My thought is we should be config ignoring this so all customers have this ability. You mentioned they didn't have permissions to do this as well. I think that would be another service wide config change.

briand44:
I don't think we want to be going custom with this site. We want to investigate if we can add this functionality to the entire SiteNow v2 service.

dnkemper:
@briand44 if we remove the config split file, it will affect SiteNow sites

richardbporter:
Yea, we probably should not be config-splitting standard SiteNow sites. If the behavior is desired for all sites, then adding it to the sitenow sync directory would be fine. However, I think @briand44 tested this and was unable to get working on another site.

joewhitsitt:
Without testing this, I don't know what it changes exactly. Is there an issue that describes the change? All I can say is that the current configuration was decided by Adam and I based on the offerings available in the D8 version (not as good as D7). It is supposed to show the first level of children for the current page. https://sitenow.uiowa.edu/documentation/media with the breadcrumb or top-nav being used as the "backtrack"

I don't care if it changes for the service. I would resist changing it for this one site. I think blocks are ignored, so an update hook is probably necessary to update existing sites.

richardbporter:
Let's close this and potentially follow up with other issues that describe what the problems are.

PR #1021, opened 2020-02-07
Provision uiowa.edu

pyrello:


pyrello:
So, I guess this is a special case that breaks our previous assumptions. For now, I renamed the alias file to `home.site.yml`, since there is already a drush alias file called `uiowa.site.yml`.

richardbporter (changes requested):
Can we move that special case up just in case another ID returns empty? Ex:
```
public static function getIdentifier($uri) {
    if ($parsed = parse_url($uri)) {
      if ($parsed['host'] === 'uiowa.edu') {
        $id = 'uiowa';
      }
      elseif (substr($parsed['host'], -9) === 'uiowa.edu') {
        // Don't use the suffix if the host equals uiowa.edu.
        $id = substr($parsed['host'], 0, -10);
...
```

richardbporter:
I moved the uiowa.dev.drupal.uiowa.edu domain over from uiowa701.

richardbporter (approved):

PR #1024, opened 2020-02-07
Provisions Feb 7

dnkemper:
Nothing like some Friday site provisions


briand44:
Can you add cbig.iibi.uiowa.edu to this? It will need an SSL cert so we should create a task for that as well.

briand44:
the original request was for cbig.engineering so there is not a submission for cbig.iibi.uiowa.edu. We will need to sql-sync so the requestor doesn't matter

briand44 (approved):
Approving

PR #1030, opened 2020-02-10
Remove SSH options from local Drush aliases.

### LINKED ISSUE #1026: Local drush aliases do not work when executed on VM
The SSH options set in each multisite local Drush alias file are nice for use outside the VM but do not work when executed on it. For example, commands like sql:sync and rsync fail when working with a remote.

richardbporter:
Resolves #1026 

Change notice: All Drush commands should execute on the VM, i.e. after running `vagrant ssh`.


pyrello (approved):
This all looks good to me.

PR #1038, opened 2020-02-11
Write custom asset info file versions dynamically.

richardbporter:
This PR invokes a [post-deploy-build hook](https://docs.acquia.com/blt/extending-blt/#overriding-variables-at-runtime) to write an [unannotated tag](https://git-scm.com/docs/git-describe#_examples) to our custom asset info files. This is to fix a pre-merge, pre-Travis regression where the profile version is empty on the status report page. I expanded it to work with other custom assets.

This will not update info files that already contain a version in the case that manual versioning is preferred. This will also not search for info files in nested modules.

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -49,6 +49,9 @@ command-hooks:
   frontend-assets:
     dir: ${docroot}
     command: 'yarn workspaces run gulp'
+  post-deploy-build:
+    dir: ${repo.root}
+    command: './vendor/bin/blt uiowa:git:version'
 sync:
   commands:
     - 'blt:init:settings'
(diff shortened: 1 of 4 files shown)


pyrello:
@richardbporter Have you merged this into the develop branch? If so, I'm assuming I can check the logs on uiowa01 to see if there were any issues?

richardbporter:
Yes, I did push it to dev. Examples:

Screen Shot 2020-02-11 at 7 45 04 PM

---

Screen Shot 2020-02-11 at 7 45 41 PM

The version on prod will just be the tag, e.g. 2.1.7.

pyrello (approved):
I was trying to make sure that dev build was able to be deployed without problems on uiowa01. There are issues but I don't think they are related to this.

Issue #1040, opened 2020-02-12
blt dsa failing on Collegiate sites


pyrello:
During the `drupal:sync:db` step, it is failing with the following output:
```
  Error Output:                                                                
  ================                                                             
                                                                               
  In EntityTypeManager.php line 150:                                           
                                                                               
    The "config_split" entity type does not exist. 
```

I'm assuming this is related to collegiate since those sites are not using the config_split module yet.

PR #1041, opened 2020-02-12
Feature: Add config to collegiate sites

pyrello:
Addresses the issue of sites not having config.


pyrello:
@mark-bennett-uiowa We'll need to add config for your two sites. We can go over that and add them to this PR when we get together later today.

richardbporter:
This resolves #1040 - collegiate should use `core-only` strategy until ready to adopt config_split which is the default BLT strategy. Otherwise, config_split gets enabled during the drupal:update process.

richardbporter:
I am able to `blt dsa` all collegiate sites without error on the CLI. Was #1040 showing up on the CLI or in the Drupal admin interface?

pyrello:
@richardbporter In the CLI. I think that uninstalling those modules on `@gradcogscilang` fixed that issue.

richardbporter (approved):
Assuming CI passes, I think this is good to go. 👍

richardbporter:
We can check out master-build in stage on uiowa01 once this is merged.

PR #1042, opened 2020-02-12
Site wants people sorted manually

dnkemper:
When I change the SiteNow People config (ie change to manual sort, sticky first/last) and do a config export, it changes the config file…. so whenever someone makes a change, it will eventually get overwritten **Which is why I'm doing this for the site.** We really need to look into this further...


joewhitsitt:
This should be fixed by #1044

PR #1044, opened 2020-02-12
Ignore the SiteNow people view

joewhitsitt:
## Problem
SiteNow People configuration wasn't being ignored. Title, path, header content, sort customizations are being wiped out each deployment...

## Solution
Add the view to the config ignore list similar to SiteNow articles view


dnkemper (approved):

joewhitsitt:
Waiting to hear from @richardbporter about whether additional steps need to be done to register this ignore on existing sites

richardbporter:
Run a sync locally and if things are as expected afterwards then I'd say there are no additional steps.

joewhitsitt:
Once the ignore is established, this will work... unfortunately, all sites will be starting over with their people configurations. The good news is that manual sort weights are saved so if they want a manual sort they can just switch it back

richardbporter:
Merging this into develop locally and pushing may have caught the config error before being deployed to stage.

PR #1045, opened 2020-02-12
Provisions Feb 12

dnkemper:
Provisioned hostmaster approved sites


richardbporter (approved):
I can create a new release this afternoon. I'd like to double check things before deploying.

briand44:
Was the customer ok with accel.education.uiowa.edu? I know that is not what they wanted and would hate to have to change it after the fact.

PR #1047, opened 2020-02-12
add php version specific zip package to drupalvm

### LINKED ISSUE #1046: Add PHP Zip to DrupalVM
## Problem
Webform and brand.uiowa.edu need the zip utility that is currently available on Acquia. Testing locally results in errors since it isn't included.

## Proposed Solution
Add zip to php_packages_extra in the box config

joewhitsitt:
Resolves #1046 

Would require a `vagrant provision` to take effect.

### DIFF
--- a/box/config.yml
+++ b/box/config.yml
@@ -725,6 +725,7 @@ php_packages_extra:
   - 'php{{ php_version }}-bz2'
   - 'php{{ php_version }}-imagick'
   - imagemagick
+  - 'php{{ php_version }}-zip'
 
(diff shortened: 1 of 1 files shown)


richardbporter (approved):

richardbporter:
Why do we have both `php{{ php_version }}-imagick` and `imagemagick`?

pyrello:
I'm assuming that `imagemagick` is the base library and the `php-*` version is the php implementation on top of that.

Issue #1039, opened 2020-02-12
uiowa01 Deploy Issues


pyrello:
After fixing the config issue for cogscilang, we are still left with a series of errors when attempting to run a deployment. This ultimately causes the `post-code-deploy` hook to exit prematurely.

```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/html/uiowa01.dev/vendor/bin/drush @self config:set system.site uuid [site_uuid] --uri=cogscilang.grad.uiowa.edu --no-interaction -v &&
 /var/www/html/uiowa01.dev/vendor/bin/drush @self pm-enable config_split --uri=cogscilang.grad.uiowa.edu --no-interaction -v &&
 /var/www/html/uiowa01.dev/vendor/bin/drush @self config-import sync --uri=cogscilang.grad.uiowa.edu --no-interaction -v &&
 /var/www/html/uiowa01.dev/vendor/bin/drush @self config-import sync --uri=cogscilang.grad.uiowa.edu --no-interaction -v &&
 /var/www/html/uiowa01.dev/vendor/bin/drush @self cache-rebuild --uri=cogscilang.grad.uiowa.edu --no-interaction -v in /var/www/html/uiowa01.dev/docroot
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_actions.Commands.ConfigActionsCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_distro.Commands.ConfigDistroCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_sync.Commands.ConfigSyncCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_actions.Commands.ConfigActionsCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_distro.Commands.ConfigDistroCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [warning] file_put_contents(/tmp/.drush/dev/factory/Drupal.config_sync.Commands.ConfigSyncCommands.cache): failed to open stream: Permission denied JSONCache.php:30
 [notice] Already enabled: config_split
[...]

PR #1048, opened 2020-02-12
Remove DRUSH_PATHS_CACHE_DIRECTORY export in cloud hooks.

richardbporter:
Resovles #1039


richardbporter:
Upstream BLT cloud hooks no longer have these export statements in them. It seems to be causing permissions issues when multiple apps try to write to the same directory.

pyrello (approved):

PR #1057, opened 2020-02-13
Set collegiate sites cm.strategy to none for now.

richardbporter:
Follow up from #1041. I think more work needs to be done for collegiate sites to adopt the automated BLT config workflow. This change simply stops BLT from importing config on deployments.


pyrello (approved):
Looks good.

PR #1063, opened 2020-02-14
Initialize intermediatechinesereader.sites.uiowa.edu multisite

dnkemper:
Customer agreed to the URL hostmaster suggested


briand44:
Let's hold off for a minute. I believe someone is filling out the request form as we speak for another site. I think we could throw a couple others in here as well

richardbporter:
Can we re-run these on uiowa01? Just delete the databases on uiowa and start over with another PR.

PR #1065, opened 2020-02-14
Return exit code 1 on Gulp task failures.

### LINKED ISSUE #1061: Some scss files are not getting built in dev
Example: https://uiowa.dev.drupal.uiowa.edu/themes/custom/uids_base/assets/css/components/global/menus/off-canvas/off-canvas.css

I ran `blt artifact:deploy --dry-run` locally to see if it was showing up in the deployment package and the file gets built locally.

However, when @bspeare was testing #1043, we noticed the same problem on his local machine even after running `blt frontend` and `yarn workspace uids_base build`.

The .scss file that generates the file listed above exists in the develop branch: https://github.com/uiowa/uiowa/blob/develop/docroot/themes/custom/uids_base/scss/components/global/menus/off-canvas/off-canvas.scss. I also confirmed that the .scss file exists in the correct location in dev on uiowa01.

richardbporter:
Resolves #1061 

Our Gulp tasks do not return exit codes > 0 on error. This can lead to corrupt/missing frontend dependencies in the build artifact.

I _think_ we should change `this.emit('end');` to `process.exit(1);` or have both. I'm not sure what `this.emit('end');` does.

### DIFF
--- a/docroot/themes/custom/collegiate_theme/gulpfile.js
+++ b/docroot/themes/custom/collegiate_theme/gulpfile.js
@@ -46,7 +46,7 @@ function css() {
     .pipe(plumber({
       handleError: function (err) {
         console.log(err);
-        this.emit('end');
+        process.exit(1);
       }
     }))
     .pipe(sass({
@@ -54,8 +54,7 @@ function css() {
         includePaths: config.css.includePaths
       }).on('error', function (err) {
         console.log(err.message);
-        // sass.logError.
-        this.emit('end');
+        process.exit(1);
       }))
     .pipe(prefix(['last 2 versions', '> 1%', 'ie 9', 'ie 10'], {
       cascade: true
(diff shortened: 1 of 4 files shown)


pyrello (approved):

richardbporter:
I think we still need to change other tasks to match this in all themes.

I don't have a great understanding of Gulp so anyone who does please chime in if this will cause other regressions.

richardbporter:
That did fail the build successfully. 👍

pyrello:
I didn't write the original gulpfile, but I have modified it a little bit over time. I believe this fix should be okay.

richardbporter:
Also, we should probably fix #1061 in this PR so we can merge it. :)

joewhitsitt:
I am no gulp expert but i have reason to believe that if there is an error during Watch, it will stop until the issue is fixed. IMO, I think that is a small hiccup for a developer versus a potentially style breaking deployment.

pyrello:
The watch command did not appear to be working the last time I tried it anyways, so I would say that we need to fix that regardless.

pyrello:
Still approved. Now with passing test!

richardbporter:
Yea it doesn't look like collegiate_theme or uids_base gulp watch tasks work. Also, uiowa_bootstrap does not have a watch task.

PR #1066, opened 2020-02-14
Set drush cache dir to random before executing cloud hooks.

richardbporter:
Related to https://github.com/acquia/blt/issues/2957#issuecomment-505942336. We may still have edge cases when cron is running simultaneously.


pyrello (approved):
Looks good to me.

PR #1073, opened 2020-02-17
Added subscription link to sitenow webform creation email

### LINKED ISSUE #927: Add link to our request delivery emails to promote subscription list
We are linking to our documentation in the site created emails to customers but we should also highlight our marketing dispatch list so they will opt-in and receive notifications about new features, upcoming trainings and events.

I know we have a form on the homepage, but we should find a permalink to use so that if the form moves, the link will still be valid

dnkemper:
Added subscription link to sitenow webform creation email

Resolves https://github.com/uiowa/uiowa/issues/927


joewhitsitt:
Should these be linked similar to the contact info?

richardbporter:
Yea we can use markup here to create friendlier links.

joewhitsitt (approved):

PR #1076, opened 2020-02-18
Fixed #1075: Update .htaccess to handle iconsortium.subst-abuse.uiowa…

### LINKED ISSUE #1075: Update .htaccess to handle iconsortium.subst-abuse.uiowa.edu redirect
Update the .htaccess file to redirect all traffic from iconsortium.subst-abuse.uiowa.edu to icsa.uiowa.edu as of Wednesday, February 19th, 2020.

mark-bennett-uiowa:
….edu redirect.


mark-bennett-uiowa:
This doesn't need to be in place until Wednesday, February 18, 2020.

docroot/.htaccess
```diff
@@ -84,6 +84,10 @@ AddEncoding gzip svgz
   RewriteCond %{HTTP_HOST} veterans.org.uiowa.edu [NC]
   RewriteRule ^ https://veterans.uiowa.edu/uiva%{REQUEST_URI} [L,R=301]
 
+  # Rrdirect iconsortium.subst-abuse.uiowa.edu to icsa.uiowa.edu
```
pyrello: 'Rrdirect' => 'Redirect'

pyrello:
This looks okay to me, written narrowly enough that it shouldn't affect other sites. However, I'm not totally familiar with the syntax of .htaccess rules and this one looks slightly different than the one above it. Would love to have @richardbporter take a look just to make sure.

mark-bennett-uiowa:
Since I didnt have to match page URLs directly to corresponding pages, I just told it to take all traffic from the original URL iconsortium.subst-abuse.uiowa.edu and redirect it to the icsa.uiowa.edu homepage. The 301 says its a permanent redirect and the L tag means to stop processing rules if this condition is met and redirected. There are definitely more complex versions of these redirects, but this one is pretty straightforward.

richardbporter:
The RewriteCond for veterans.org.uiowa.edu should probably escape the dot characters to be technically correct. Although, the docs say it accepts plain text. I think both work.

richardbporter (approved):

PR #1082, opened 2020-02-19
i1078 - Site Request Migration Email Handler

### LINKED ISSUE #1078: Add migration email handler
For SiteNow site requests, add a migration email handler for sites that are created but still need content migrated over

dnkemper:
Added migration email handler for sites that plan on migrating their content


joewhitsitt:
I'd like to see @briand44 's approval on this in terms of how the sub-directory project is being run, but otherwise, looks fine to me.

briand44:
I didn't review this from a technical perspective but I am fine with the idea of this. Some customers tell us they want us to migrate the content prior to requesting the site so I think it makes sense that when we tell them the site is ready we are clear in that we will work on migrating the content and let them know when it is ready for review. Thank you.

joewhitsitt (approved):

PR #1083, opened 2020-02-19
i990 - Changed FontAwesome iconpicker location and added URL to view FA icons

### LINKED ISSUE #990: Disable autocomplete and add link to font awesome icons
When picking icons for social media links, the icon picker starts off connected, but as the autocomplete filters, it shrinks the results.

May need to change the fly out position of the picker results.

dnkemper:
Changed FontAwesome icon picker location and added link to view all icons

 * When you filter icons, it stays in the same place - it's the only location that works (unless we want to display it inline where it's always visible)

Resolves #990


richardbporter (changes requested):
This results in horizontal scroll on mobile and the icon picker is off-canvas. I would recommend moving it to bottom and altering that field to disable browser autocomplete so it doesn't cover that.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

dnkemper:
@richardbporter so should I change the issue gist? 'Fontawesome iconpicker floats away from the field it is attached to' because it would float away when using autocomplete... Then just say 'disable autocomplete and add link to font awesome icons'

richardbporter:
Yea that sounds good.

joewhitsitt (approved):

PR #1091, opened 2020-02-20
Feature: UIDS theme spacing

bspeare:
Updated layout margins and container margins in uids_base to match gutter widths from uids mockups.


pyrello (approved):
Looks good to me. Only reviewed the changes, haven't tested.

PR #1096, opened 2020-02-24
Automate homepage config workflow and configure metatag.

richardbporter:
This PR switches the homepage config strategy back to the default, config-split. Since all config will be exported for this site, I've removed some unnecessary config-related modules as well. The less overhead, the better.

The test case for these config changes was metatag. I configured global tokens, permissions and added a field to the basic page bundle. I configured masquerade to print a block on the admin/people page to test the permissions locally. This is similar to how sitenow uses masquerade.

Still need to setup config_ignore settings. We just need to decide what config webmasters should be able to modify.


richardbporter:
Successfully deployed to dev.

richardbporter:
Looks like I'll have to re-do this without uninstalling config_sync et al.

Issue #891, opened 2020-01-06
Iowa Global Health Network Website/Customizations


joewhitsitt:
ighn.international.uiowa.edu

Currently: https://international.uiowa.edu/partnerships-and-research/academic/ighn
Inspiration: https://sustainability.uiowa.edu/sustainability-scientists-and-scholars

A SiteNow v2 website that adopts OSC or International theme later that lives at *.international.uiowa.edu with a separate content type, taxonomies and view page that accomplishes the following:

- Crowdsourced content type (open to any that login) with moderation.
- First, Last, Credentials, College (taxonomy single), Department (taxonomy single), primary focus areas (taxonomy multi 3), Global Regions (taxonomy multi), Language Expertise (taxonomy single), Biography, CV/Website Link, Photo.
- Select or other type functionality or help text/extra non-display field to capture content that isn't an existing option.
- Structured in a way that either another type of network could be captured with the same content type or that another similar content type doesn't conflict.
- View must filter by name and focus areas. uiowa_bootstrap grid display similar to sustainability inspiration.
- Could eventually be migrated into future International D8 website.
- Separate from person content type in case person content type is being used on the main International website.

See @briand44 or @joewhitsitt for more details.

dnkemper:
- [x] Comma separated field
- [x] Linked card
- [x] Read more/Expand to read bio (collapse) - uiowa bootstrap functionality
- [x] Reduce uiowa_bootstrap styles from custom override and use template class names instead (e.g. row, col-md-4)
- [x] Register permission splits (new permissions based on content type, so we need to split for ighn any roles that can work with the content type)

joewhitsitt:
I looked over all of this yesterday and had a couple more thoughts:

- CV/Website Link (unless changed in the MOU) should just be one field (link, title required if link provided). Setup like this, we can copy/paste the functionality around linkable card content blocks.
- Credentials field since it is just text should be a single value similar to the SiteNow person content type. We can then add the credentials as they want them formatted in a string appended to their name like it is typically done.
- I think the full node should be public but not linked to. I added the sitemap settings for it and we should put together basic output of the content so that if a human comes across it it will provide meaningful info. Trying to hide content is harder than it is worth.
- The teaser should be constructed like a bootstrap card to work well within the theme and to recreate the sustainability site's example. See what classes are being applied to teasers for SiteNow articles grid and apply those to the teaser template.

dnkemper:
- I'm confused.. isn't it just one field? With the link/title?
[...]

PR #1107, opened 2020-02-25
ighn.international development round 2

joewhitsitt:
related to #891


richardbporter (dismissed):
It looks like this merge conflict was not resolved correctly. @joewhitsitt is going to hard reset back to last known good state.

dnkemper (approved):

PR #1109, opened 2020-02-26
Set up config_ignore for uiowa.edu.

richardbporter:
Editors and webmasters currently have access to change active configuration which will be imported on code deployments and therefore lost. This PR sets that changeable configuration in the config_ignore settings which will prevent it from being imported. Currently, this is only uiowa_bar.settings but we'll want to add/remove things as development progresses.

Additionally, editors and webmasters had permission to administer menus. Instead of config_ignore'ing that I swapped that permission for a more granular approach with menu_admin_per_menu. This way, editors/webmasters can only manage links within menus.


richardbporter:
I only gave webmasters access to the CTA menu but that can changed based on feedback.

pyrello (approved):
Looks good.

PR #1111, opened 2020-02-26
hpproject - set front and prevent deletion

### LINKED ISSUE #1110: set home to page and prevent delete
## Issue
Took a sticky on the sprint board. The homepage site's homepage is the drupal default. set it to node/1 (Michael said he would move content to repurpose) and prevent users from deleting the front page until a new path has been set.

## Proposed solution
- Set the front page. Done :wink:
- Reuse sitenow code to prevent deletion.

joewhitsitt:
Resolves #1110


richardbporter:
It would be nice if we could move this code to a shared module or something so we don't have to duplicate across profiles.

pyrello:
I am also cool with that idea.

joewhitsitt:
@richardbporter is there an existing module you recommend or are we talking like uiowa_core?

richardbporter:
I'm not sure. Maybe we could wait on that and create one if we have more shared functionality?

joewhitsitt:
i like putting it off for now and revisiting. can i get an approval for the short term solution?

richardbporter (approved):

PR #1112, opened 2020-02-26
Update Composer dependencies as per docs and remove unused.

richardbporter:
This PR removes the unused dependency config_split_ignore which caused us issues in the past. It also move some dependencies around based on their docs:

- https://github.com/hirak/prestissimo#install
- https://github.com/zaporylie/composer-drupal-optimizations#installation


joewhitsitt (approved):
GOODBYE CONFIG_SPLIT_IGNORE!

PR #1113, opened 2020-02-26
Set up and configure config_split for homepage.

richardbporter:
The current environment splits have some sitenow-specific configuration in them. This makes them unusable for other profiles. This PR moves configuration to feature splits and allows for easier sharing of environment splits between profiles.


richardbporter:
This is probably not the way we should go with this. We should look into sharing default configuration as per the BLT docs.

PR #1122, opened 2020-02-28
Feature: uids_base uiowa_blocks/superfish updates

bspeare:
In this PR:

- Added CTA block to uiowa_blocks that included gold, black, and brain pattern background image styles. 
- Added padding/margin/background options to text area block in uiowa_blocks.
- Added a layout builder style to section container for extended with container that goes beyond page container. 
- Updated uiowa_blocks styles so that padding/margin options are only available within text block.
- Updated superfish spacing and styles for <nolink>, active links, and links without dropdowns.


pyrello (approved):
Reviewed, not tested.

PR #1126, opened 2020-03-02
i985 - Featured Content (Entity Browser) view changes

### LINKED ISSUE #985: Adding unpublished content as featured content and clicking preview results in an error

dnkemper:
- Instead of viewing all content whether it's published/draft/review/archived in the Featured Content entity browser, I made it so only published content is available to be selected from users when adding Featured Content to a page (to avoid the errors we witnessed in Joe's D8 training session)
- Removed the status column from the Featured Content (Entity Browser) View


joewhitsitt (approved):
Tested and works as expected. I was curious about what would happen if something that was already selected/saved became unpublished. The field allows it to still be selected and saved. It shows the unpublished content (not noted as such) to logged in users. It resolves the bug which is important, but I am going to create some follow up tasks to make this stuff more clear to the editor.

PR #1127, opened 2020-03-02
Feature: Improvements based on investigation.

pyrello:
* Moved block definitions from collegite to uids_base - Necessary to install uids_base outside of collegiate.
* Updated ReplaceCommands to check if has 'cm.core.install_from_config' set before setting the '--existing-config' flag for site install - All SiteNow sites are currently created with this setting as TRUE by default, so this should not impact current functionality.
* Updated sitenow profile and sitenow_paragraphs module to add checks to see if  has a bundle() method - I was getting errors when the site would attempt to load the field_body_content_block paragraphs into Layout Builder.
* Added check for uids_base to ensure ['element']['#name'] is set before checking it - resolves a bug in uids_base when using Layout Builder.


pyrello:
These are changes I had to make in my attempts to get the uiowa.edu site running on SiteNow with enhancements from Collegiate. I added descriptions of why they are included after each bullet point.

joewhitsitt (approved):
reviewed, not tested. ;)

PR #1132, opened 2020-03-03
Feature: uiowa.edu on SiteNow, take 2

### LINKED ISSUE #957: Implement patch to fix webkit form control rendering issue
# The Problem 

Screen Shot 2019-11-08 at 1 26 19 PM

### LINKED ISSUE #968: "You have unsaved changes" Error is back on prod
This time, I was editing the content inside of a UIowa block on the frontpage of the policy site: [https://policy.clas.uiowa.edu/](https://policy.clas.uiowa.edu/). I removed the title and body of the block, but kept the Link title and Link, in order to see if we could just have an image and link. Immediately upon the save attempt, Im locked in the "You have unsaved changes" error again.

The following did NOT fix the issue:
drush cr.
Clear cache from Configuration > Performance.
Discard changes button in LB.
Revert to default button in LB.
Returning all deleted content back to the block and re-saving.

pyrello:
Resolves #968 
Resolves #957 

This version uses the config split workflow and SiteNow as intended. It includes most or all of the updates that we had made to the original uiowa.edu homepage site.


pyrello:
@richardbporter Asssuming this all make it onto dev without any further issues, then I'd say the next step is to figure out swapping this out for the original site. Happy to make those changes - or let you do it ;) - just need some help figuring out what all changes need to be made.

joewhitsitt:
Do you want the body field/uiowa_blocks text area block to be filtered_html by default? Looks like they are currently set to minimal.

I see other clean-up tasks to do, but for a POC that can move forward I think this is good.

Brain dump for future tasks
- Alter page options to remove the hide title, sidebar options
- Decide how to handle body trimmed/summary versus field_teaser (Content Description) for metatags and teasers.

richardbporter:
One thing I noticed is that the Masquerade block is not on the /admin/people page.

richardbporter (approved):

richardbporter:
💯

PR #1134, opened 2020-03-04
i1129

dnkemper:
This is for many HD related questions/issues. Full width settings were impacting IE carousel images. The styling needs to be applied to all carousel widths, not just the full width carousel. Not sure if you want to implement this a specific way or not.


joewhitsitt:
This looks like it has file changes associated with https://github.com/uiowa/uiowa/pull/1126 included. I will try to test both of them today, but we should keep the changes separate in terms of the git log

dnkemper:
@joewhitsitt WHY HAS THIS BEEN HAPPENING TO ME THIS WEEK? Very sorry about that. Would you like me to create a new branch and start over with a clean slate? P.S. Thank you!

joewhitsitt (approved):
removed the featured content change. i don't have IE here at home so this is a guess and check deployment. just made sure it didn't effect non-ie functionality. also snuck in previous css that got removed accidentally that solves the small carousel/large caption issue.

PR #1139, opened 2020-03-04
Provisions March 4

dnkemper:


briand44 (approved):
I am assuming the centerforconferences are on uiowa and the other on uiowa01

dnkemper:
@briand44 yes they are!

PR #1142, opened 2020-03-04
Feature: uids_base sprint 8

bspeare:
Changes in this feature:

- Add css grid class to section
- Add background patterns to section 
- Add hero class to one-col section 
- Heading sizes resized
- Added page_title_hero block
- Sass variable for image path  
- Hero section block with background image/video
- Removed Headline block from uiowa_blocks
- Padding/Margin updates to layout builder styles


richardbporter:
Nothing seems awry on the sitenow demo site. Syncing HR now as it has slightly different block config.

One potential issue is the layout dependency - future exports may add it back if core config detects it. Was that a manual addition to uiowa_blocks or was features auto-detecting it?

pyrello:
It was detected as a dependency because we had tried to set up one of the blocks using layout builder to control its layout. I removed that section of the config from the relevant block.

richardbporter:
HR seems fine. 

I exported some bogus custom block config from the demo site just fine. However, I could not import into the homepage since it has been split off: https://github.com/uiowa/uiowa/blob/master/config/uiowa.edu/config_split.config_split.uiowa_edu.yml#L52

1. Edit blocks on sitenowdemo.local.uiowa.edu
2. `drush @sitenowdemo.local config:export`
3. `drush @home.local config:import`
4.  [notice] There are no changes to import.

Should we remove some of those graylisted items?

pyrello:
I think that all those items had to be graylisted to prevent config changes being exported to SiteNow. Feel free to double check my work. I was figuring it out as I went.

Sean Adams-Hiett via mobile

pyrello:
@richardbporter Okay. I think I see what you are saying now. Since the config is added to SiteNow, we need to remove it from the homepage split.

pyrello:
I removed all required uiowa_blocks config from the uiowa.edu site split. Layout builder styles are optional config in that module and will get enabled if the layout builder styles module gets turned on, so I left those in the site split.

joewhitsitt (approved):
poked around on sitesgis.dev. no biggies

PR #1149, opened 2020-03-06
Place masquerade block in claro theme on admin/people

richardbporter:


config/uiowa.edu/block.block.masquerade_2.yml
```diff
@@ -0,0 +1,26 @@
+uuid: 31ac6ef8-4069-4b41-85e5-b21755cc344a
+langcode: en
+status: true
+dependencies:
+  module:
+    - masquerade
+    - system
+  theme:
+    - claro
+id: masquerade_2
```
pyrello: Is there a reason this is `masquerade_2`? I think we might want it to be something like `claro_masquerade`.

richardbporter: Don't know. The adminimal block is named similarly. I just re-exported and set the machine name myself.

pyrello (approved):

PR #1151, opened 2020-03-06
Force add missing UIDS JS.

richardbporter:
Because custom theme asset directories are ignored in the top level .gitignore, we have to either compile JS during build or force add them combined with a local gitignore that negates it.

https://github.com/uiowa/uiowa/blob/master/.gitignore#L32
https://github.com/uiowa/uiowa/blob/master/docroot/themes/custom/uids_base/.gitignore#L1

Long term, compiling all assets might be preferable.


pyrello (approved):
Looks good.

PR #1156, opened 2020-03-06
Feature: uids base header/footer updates

bspeare:
In this PR:

- Fixed sticky header
- Created fixed footer
- Updated breadcrumbs and page title spacing to work with top header
- Updated admin.scss toolbar styles to work when header is sticky and set to IOWA header type


bspeare:
@pyrello I think this is ready.  At some point we will need to address #1158, but for the sake of getting this up for the demo we should move forward with it.

pyrello:
@bspeare It looks like this branch has some conflicts with the master branch. Do you want to resolve those and I'll get it reviewed?

pyrello (approved):
Looks good!

PR #1162, opened 2020-03-10
Copy global Drush commands to build artifact.

richardbporter:
The post-artifact build hook documented at https://docs.acquia.com/blt/extending-blt/#overriding-variables-at-runtime does not seem to be working. As a result, our UiowaCommands are not copied into the build artifact. This refactors the post-build step into one command that account for additional functionality.

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -53,7 +53,7 @@ command-hooks:
     command: 'yarn workspaces run gulp'
   post-deploy-build:
     dir: ${repo.root}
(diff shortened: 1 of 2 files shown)


richardbporter:
The global Drush commands we have now give us the ability to check database sizes as part of the status command as well as additional command to list the largest SQL tables per site. It also invokes the BLT update process post-sql sync.

pyrello (approved):
I have not tested this, but I have reviewed the file changes. Everything looks good to me.

PR #1163, opened 2020-03-10
Allow classes on additional html elements

### LINKED ISSUE #1159: Allow for classes to be added to the filtered and full HTML formats
# The Problem

- Editors building the new home page design need access to add classes to html elements such as p, h2, h3, h4, h5, h5, ul, ol, table, etc. 
- There are currently workarounds in place to use a `<div>` or a `<span>` on an element, but that doesn't work for a list-item like https://uiowa.github.io/uids/components/detail/lists.html or if you want to preserve the HTML document outline by adding classes to heading tags. 

# Proposed Solution

Allow classes to be added to HTML tags in the filtered HTML format.

joewhitsitt:
Resolves #1159 for SiteNow and uiowa.edu.

### DIFF
--- a/config/uiowa.edu/filter.format.filtered_html.yml
+++ b/config/uiowa.edu/filter.format.filtered_html.yml
@@ -56,7 +56,7 @@ filters:
     status: true
     weight: -49
     settings:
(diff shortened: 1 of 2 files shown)


joewhitsitt:
The following sites have split or have their own configuration for filtered_html:

- config/uiowa-old.uiowa.edu/filter.format.filtered_html.yml
- config/cogscilang.grad.uiowa.edu/filter.format.filtered_html.yml
- config/informatics.grad.uiowa.edu/filter.format.filtered_html.yml
- docroot/profiles/custom/collegiate/config/install/filter.format.filtered_html.yml

pyrello:
@joewhitsitt It is fine to not include these sites in this update. Cogscilang and informatics are going to need to be transitioned to SiteNow to allow for the deprecation of collegiate and they will pick up these changes then.

pyrello (approved):
I have reviewed this, but not tested it.

PR #1166, opened 2020-03-11
Scope layout builder specific styles

### LINKED ISSUE #1164: Visually hidden title block is showing up outside of Layout Builder
The recent changes to the page title block that allow it to be hidden visually seem to be not working quite right. If you visit https://uiowa.stage.drupal.uiowa.edu while logged out, you can still see the following:
Screen Shot 2020-03-10 at 11 06 12 PM
I believe this should only be showing when viewing the block inside the Layout Builder UI.

joewhitsitt:
Resolves #1164


pyrello (approved):
Approved, not tested.

PR #1179, opened 2020-03-12
Refactor install command to require manual confirmation.

richardbporter:
Running this command without confirmation could be disastrous. This is unfortunately just another step in the provisioning process.

Deploy a release, SSH into the application, CD to the app root and run `./vendor/bin/blt uiowa:multisite:install`.

### DIFF
--- a/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
+++ b/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
@@ -98,21 +98,20 @@ class MultisiteCommands extends BltTasks {
    * This command invokes the BLT drupal:install command which handles passing
    * arguments and options to Drush site:install.
    *
-   * @see: Acquia\Blt\Robo\Commands\Drupal\InstallCommand
-   *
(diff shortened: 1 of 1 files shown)


joewhitsitt (approved):
This method makes me feel better. Less concern about off-hour errors. Reviewed but not tested

pyrello:
Looks good. Also reviewed, but not tested.

richardbporter:
I think this gets us closer to allowing sites to be built locally first then synced up. We would need some way to exclude sites from this command. A custom config option set on creation seems like the best approach.

richardbporter:
> I think this gets us closer to allowing sites to be built locally first then synced up.

Just to clarify, that is a nice-to-have option but not necessarily the route "standard" sites will take.

richardbporter:
The slack notifications might be unnecessary if this is run manually.

richardbporter:
> The slack notifications might be unnecessary if this is run manually.

They probably don't hurt in case someone starts the process and has to walk away.

PR #1189, opened 2020-03-13
Updated breadcrumbs block setting to remove the layout builder active…

pyrello:
… visibility condition.


bspeare (approved):
Tested locally and it works.  Thanks!

PR #1193, opened 2020-03-16
Prevent accidental multisite install in dev and test environments.

richardbporter:
This PR prevents UMI from running in undesired environments like dev/test where DB storage is limited.

Also cleaned up lots of PHPCS warnings. We should try to clean those up occasionally otherwise `blt phpcs` is hard to parse during development.


joewhitsitt (approved):
reviewed, not tested

PR #1194, opened 2020-03-16
Feature: footer menus and uiowa.edu block placement of them

pyrello:


joewhitsitt (approved):
Looks fine, but this might cause us to migrate SiteNow off of footer and footer2 because of permissions. uiowa.edu can split permissions for now, but SiteNow webmasters will eventually be confused with access to five different footer menus...

PR #1196, opened 2020-03-16
Feature: Secondary navigation LB styles

pyrello:
* Adds to layout builder styles to apply to menu blocks, one for horizontal, one for vertical
* Adds a patch to only show one set of blocks in layout builder when Menu Block module is enabled.
* Updates to uids 1.5.0 (which includes improvements to support internal navigation updates).
* Unrelated update to add `<div>` with classes around footer menu region.
* Lower the weight of the the `minimal` filter format so it doesn't show up first.


bspeare:
@pyrello Which menu block do you suggest adding to the page to test?

bspeare (approved):
Works great! Thanks!

PR #1197, opened 2020-03-16
Add login link to footer.

### LINKED ISSUE #1186: Add login link back to footer.
Probably in uids_base theme.

richardbporter:
Resolves #1186 

I'm confused where the best place to put this is. Started implementing based off the styles in in footer.scss:
https://github.com/uiowa/uiowa/blob/ba2337a215a464e535af2d8ad7fb0bfdf728648e/docroot/themes/custom/uids_base/scss/components/global/footer/footer.scss#L59-L76


pyrello:
@richardbporter That seems like as good a place as any. At this point, it is kind of Drupal-specific, so putting in the Drupal footer scss file makes sense to me.

bspeare:
@richardbporter I think all the custom styles after the import statements can be deleted.

richardbporter:
@bspeare I can't get that FA icon to show up. What am I doing wrong?

bspeare (approved):
Thanks for adding this!   I think the fa unicode might have been from 4 instead of 5.   I updated the unicode to the free sign-in icon from 5.

richardbporter:
Looks good. 👍

PR #1229, opened 2020-03-19
Feature: uiowa.edu menu updates

pyrello:
* Created Top links menu
* Created External quick links menu
* Placed Top links menu block in the Top links region
* Placed External quick links menu
* Added preprocessing logic to add classes to the above blocks to apply uids styles


bspeare (approved):
Tested it locally and it works great.  Thanks!

Issue #798, opened 2019-12-03
Look into webform spam issues


joewhitsitt:
## Problem

Honeypot, if setup correctly, is not protecting our webforms effectively by the look of our own email inbox as it relates to our sitenow.uiowa.edu contact form.

We need make sure honeypot is configured correctly and if so, consider other preventative options.

joewhitsitt:
This seems to be helping folks: https://www.drupal.org/project/honeypot/issues/2811189

dnkemper:
According to users, the patch made the honeypot field visible BUT didn't prevent spam. Researching other options

dnkemper:
Waiting for next Webform release

joewhitsitt:
Next version of Honeypot: https://www.drupal.org/project/honeypot/issues/3081238

joewhitsitt:
In looking at #924, I noticed we haven't configured and enabled Honeypot time limit. Do we also want to try that?

joewhitsitt:
Also, https://www.drupal.org/project/recaptcha v2 checkbox was given the semi-thumbs up by Todd. We could ask Bill how it is working on Engineering

joewhitsitt:
i think we updated honeypot but the release didn’t include that patch. opinions about whether the patch was effective or not aside. so we could try to apply that patch or move forward with removing honeypot and adding recaptcha v2 checkbox and establishing an its-web recaptcha account to use by default.

joewhitsitt:
Marking this as a bug, because honeypot is not doing what it is meant to do.

joewhitsitt:
From: https://www.drupal.org/project/honeypot/issues/3009566, it sounds like the defaults are ineffective in this machine learning age. geerlingguy, maintainer has resorted to reCAPTCHA for a few forms. Some are seeing success with the timeout set higher than 0.

joewhitsitt:
Evaluation (sitenow.uiowa.edu) of the timeout cannot happen until a successful deployment of 2.4.6 on uiowa occurs.

joewhitsitt:
Closing for now. The spam has been reduced utilizing the timeout option

joewhitsitt:
_if_ we need to investigate another solution. https://www.hcaptcha.com/ is a new service that prioritizes privacy more than reCAPTCHA. I don't think there is a contrib project yet for Drupal.

PR #1242, opened 2020-03-23
Test honeypot patch

joewhitsitt:
Relates to #798

This is a small step to test before trying things like cache busting timeouts or reCAPTCHA.


briand44:
added Alan as a reviewer, he was looking for something to work on.

joewhitsitt:
it can't really be reviewed locally besides confirming that there is a honeypot field at the top of webforms (hidden). to test, we need to go where the spambots are

briand44:
I realize that but I thought you had asked for someone to confirm that during our meeting. If you just want me to approve, tell me.

joewhitsitt:
of course i would love a blind approval, but given that there is a composer update, it would probably be safer to have at least one other successfully update their local with this branch.

GaryRidgway (approved):
I have tested this and confirmed the field is now at the top of the page

PR #1220, opened 2020-03-18
A-Z List for uiowa.edu

joewhitsitt:
Related to #1217

Everything but styles and import update hook (commented out). I think it would be best to get this established on stage first as-is then follow up with the import and styles.

Issue #1217, opened 2020-03-18
As a site visitor I can see a list of A-Z links to quickly navigate to where I need to go.


joewhitsitt:
- [x] #1230 Port A-Z Vocabulary and Taxonomy fields
- [x] #1250 Export/Import existing terms/links
- [x] Recreate view page/attachment
- [x] Breadcrumb (See SiteNow Articles for contextual filters)
- [ ] Autocomplete filter?

Issue related to views contextual filters and breadcrumbs: https://www.drupal.org/project/drupal/issues/2787051

PR #1250, opened 2020-03-24
Import A-Z Links into uiowa.edu

joewhitsitt:
Uncommented an existing update hook to import links into new functionality rolled out with #1220

Relates to #1217


pyrello (approved):
Nice work!

PR #1251, opened 2020-03-24
Feature: Collegiate feature split and Grad College updates

pyrello:
- Removes and re-creates iowasuperfund.uiowa.edu.
- Removes and re-creates uipda.grad.uiowa.edu.
- Creates a temporary home for current informatics.grad.uiowa.edu
- Creates a temporary home for current cogscilang.grad.uiowa.edu
- Creates a feature split called 'Collegiate' that is a clone of the uiowa.edu site split

To apply the Collegiate feature split, you run the following command on a SiteNow site:
```
drush @site.env config:import --source ../config/features/collegiate --partial
```


joewhitsitt (approved):
Cleanly applied collegiate split to an out of the box site. Originally tried informatics but it choked on existing entities.

PR #1260, opened 2020-03-25
Feature: UIowa Alerts module, breadcrumb region, theme refactoring

pyrello:
* Added uiowa_alerts module (adapted from 7.x version).
* Added a breadcrumb region to the page template.
* Refactored page template to simplify and eliminate redundancy.
* Added new generic hero template. 
* Both existing block-specific hero templates now use that via 'embed'. 
* Updated uids_base_form_alter to skip certain forms when adding 'uids-content' class. 
* Refactored the way that the page title hero template pulls in the breadcrumb so that it doesn't trigger the condition plugin on the block.
* Re-added Layout Builder visibility setting on breadcrumb block.
* Enabled media_library_theme_reset module for uiowa.edu split.

# Testing breadcrumb updates
* Observe that a breadcrumb block shows up on the A-Z page: https://uiowa.local.drupal.uiowa.edu/a-z
* Observe both that the breadcrumb shows up inside the hero page title block and does not show up at the top of the page on https://uiowa.local.drupal.uiowa.edu/admissions/undergraduate-admissions

# Testing uiowa_alerts module
* Enable uiowa_alert module
* Place a block in the Alerts region on the blocks page: https://uiowa.local.drupal.uiowa.edu/admin/structure/block/add/uiowa_alerts_block/uids_base?region=alert
* Visit the config at https://uiowa.local.drupal.uiowa.edu/admin/config/system/uiowa-alerts to test two scenarios:
  1. Production + No alerts message - this will allow you to see the "No alerts message"
  2. Test - this will show you what the alerts themselves look like
* To view the alerts block, you'll need to visit a non-admin page

# Other testing
* Observe when visiting the content pages above that the hero looks the same as it did before (no changes because of hero templates refactoring)
* Observe on any content page that the page looks the same as it did before (no changes because of page template refactoring)
* Click the Layout tab on a content page and add a block that includes a media item. Note that the media library modal is styled like the Seven theme.


pyrello:
Is this also something that we want to turn on for uiowa.edu? If so, should the Hawk Alert block be placed in the newly created Alert region?

@richardbporter I'm open to ideas about how to set the config to be the correct feed URL in prod and test/dev/local. Would that be something that we handled via config_split?

joewhitsitt (approved):
Nice work. Works as expected. I did notice squished media thumbs in the settings tray that doesn't happen in stage, but compared to the improvements I wondered if it was worth mentioning.

Screen Shot 2020-03-25 at 8 12 24 PM

pyrello:
@joewhitsitt Thanks for checking it out. Nice catch. I think we still have an issue open back in uiowa/uiowa01 for that.

pyrello:
Here it is: #1264

Issue #697, opened 2019-10-31
Lazy Load Images Using Blazy


joewhitsitt:
## Problem
Our web performance score is low. I don't know what that means but off-screen images on page load have something to do with it and with basic page we are indirectly encouraging long pages and lots of images.

https://www.drupal.org/project/blazy

https://www.symphonythemes.com/drupal-blog/speed-your-drupal-8-sites-image-lazy-loading

Probably not for section images, but everything else makes sense to me.

bspeare:
@joewhitsitt, having this or https://www.drupal.org/project/lazy could greatly help the page load speed for the new image heavy home page.   Do you have any thoughts on whether we could implement this?

joewhitsitt:
@bspeare I think this is past whether we _can_ and more about _how and when_ we implement it. We postponed it for SiteNow because we didn't have time to think about it. Blazy definitely has the usage compared to Lazy. I don't have experience with either and don't know how it works with everything else (responsive images, etc.).

bspeare:
Yes, blazy has some good numbers behind it.  I'll add it to the parking lot for further discussion.

pyrello:
Some notes about using Blazy:
- It does integrate with responsive images.
- It appears that the aforementioned integration requires using a Blazy formatter, which means that it would become a dependency of every image field and require swapping all of those out.

joewhitsitt:
Prioritizing for the homepage project with #1262 . Attempted to do SiteNow as well and things broke. The way we are rendering images programmatically is no longer ideal for all the things we want to do going forward.

PR #1262, opened 2020-03-25
uiowa.edu - Lazy Loading Images

joewhitsitt:
Related to #697 but only for uiowa.edu

block image gallery and block vertical video need work. getting these others in for now


bspeare:
I'm having a quirk in my local environment.   Can someone test updating the banner block on https://uiowa.local.drupal.uiowa.edu/node/41/layout with a new photograph and then save the block and layout?  The image won't show up unless I manually clear the cache on the site. Is this just a local caching issue?

bspeare:
@joewhitsitt Is there anything visual that you noticed that Blazy does to the images or does it just lazy load them in the background?

joewhitsitt:
Darn. I noticed the same thing with updating images... I will search the issue queue

No visual was set (didn't want to mess with styles), I think it reserves the space for the image though. The noticable difference is the markup changes and light house doesn't complain about the images loaded by blazy.

joewhitsitt:
Going in a different direction.

PR #1265, opened 2020-03-26
i858 Fixed the Display More Link verbiage that was confusing users

### LINKED ISSUE #858: Doc/help text update for invalid Articles "more" link
Followup to issue #847 

### Issue
Articles paragraph has the more link option enabled with no link entered (defaults to Articles view page) and the view is disabled through SiteNow Articles config, the "View more articles" button does not render. May be confusing to a user, as currently there is no explanation for this behavior.

Documentation/help text on the SiteNow Articles config and/or articles paragraph should be updated.

dnkemper:
i858 Fixed the Display More Link verbiage that was confusing users. Resolves #858


richardbporter (approved):
Thanks. Can you add "resolves #858" to the PR description so that that issue is closed automatically when this PR is merged into master?

PR #1287, opened 2020-03-27
Remove docs on vagrant/virtualbox issue.

richardbporter:
While working locally, I noticed the following message:
```
==> uiowa: A newer version of the box 'geerlingguy/drupal-vm' is available and already
==> uiowa: installed, but your Vagrant machine is running against
==> uiowa: version '2.0.6'. To update to version '2.0.8',
==> uiowa: destroy and recreate your machine.
```

This worked for me without issue but requires a `vagrant destroy` which drops all local databases. Make sure you're at a stopping point before doing this, e.g. all config exported.

While updating local dev tools, I updating Vagrant to the latest release which resolves an issue with VirtualBox 6.1. I probably did not follow the proper update/uninstall methods but everything seems to work for me even after a restart. I can't guarantee it will work for anyone else.

I don't think any testing/review needs to happen here. The point of this PR is just to remove our documented caveat on installing Vagrant/VirtualBox so that the BLT docs can be followed as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This project is based on BLT, an open-source project template and tool that enab
     ```
 ----
 # Local Environment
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Looks good.

PR #1292, opened 2020-03-30
Lazy Load Images

### LINKED ISSUE #697: Lazy Load Images Using Blazy

joewhitsitt:
Resolves #697 

This implements lazy load images for both SiteNow and uiowa.edu.

### Editors
Editors should be able to add, edit images like before. Need to test all image fields, view modes and wysiwygs. For both layout and paragraphs, caches should be invalidated and editors should see new and replacement images right away. Inspecting the DOM, editors should see lazy load markup around the image markup.

### Visitors
Visitors should see a fade-in effect for images on page load. The network tab should show that the majority of image assets are loaded later than other assets.

**Note:** SiteNow Carousel in my testing could be one instance where we don't want lazy load. The load doesn't seem to happen until the carousel navigation is clicked. The carousel uses a unique view mode so other images should not be affected if removed. There is lazy load compatibility with Slick slider if we adopt that after Bootstrap.


bspeare (approved):
Love it! It works great on my local version of uiowa.edu.

joewhitsitt:
Cool! I'd like to make sure that it is working as expected in most instances (paragraph bundles, view modes)

richardbporter:
I tested the demo site locally but adding a bunch of widgets with images. 👍

joewhitsitt:
The other issue I had to setup config for is related to the entity embed, entity browser and media library to make sure images loaded correctly for editing interfaces.

I'd like to also get thoughts on SiteNow Carousel. Is it a problem to load on slide change?

joewhitsitt:
@dnkemper spotted an issue with the global caption displaying when it isn't supposed to. working on a fix

dnkemper:
@joewhitsitt at first, the second image lagged a bit with loading and only showed the caption... but it eventually showed after a couple seconds then loaded fine. Might just be me though.

dnkemper:
> @dnkemper spotted an issue with the global caption displaying when it isn't supposed to. working on a fix

discovered that the global caption only appears in wysiwyg when you display the original picture. Works correctly with displaying large, small and medium images

dnkemper:
Tested w/ updated code using the original image setting in the wyswyg and works correctly now! Great job, Joe!

richardbporter:
I don't think the carousel lazy loading is an issue.

richardbporter (approved):

PR #1294, opened 2020-03-30
Document local logging setup and patch entity_embed to prevent log flooding.

richardbporter:
While monitoring server logs, I noticed entity_embed flooded entries and looked to fix it locally. Documented how to monitor Drupal watchdog locally since dblog is never enabled.


pyrello (approved):
Reviewed, not tested.

PR #1296, opened 2020-03-30
Enable bigpipe and sessionless bigpipe in default config.

richardbporter:
Bigpipe will improve TTI for authenticated sessions while sessionless bigpipe will improve TTI for the first load for unauthenticated sessions.


richardbporter:
This is in dev now: https://uiowa.dev.drupal.uiowa.edu

I still ran into that issue where the initial page load takes a long time after a cache rebuild. I didn't see any noticeable difference with sessionless bigpipe enabled, even though it sounded like that was the point of it. I think that might be a dev/stage hardware issue and testing prod would be more accurate.

I don't think this change will hurt anything and will probably enhance authenticated TTI.

joewhitsitt (approved):

PR #1298, opened 2020-03-30
Restrict locally-hosted image, video upload sizes to 10mb or less.

### LINKED ISSUE #1297: Locally-hosted assets are not restricted to under the cacheable limit
Problem: any locally-hosted asset like file, image, video can be up to 50mb in size. This is above the 10mb recommendation https://docs.acquia.com/acquia-cloud/performance/varnish/files/#varnish-10mb. This has two negative consequences: disk space is eaten up faster (currently a 83% usage) and requests for these files bypass Varnish. 

Solution: limit all locally-hosted assets to a max of 10mb. Obviously, this is not a large amount. Some sites have legitimate use cases for larger assets. We should allow for better integration for embedding remote assets from third-party sources.

- [x] Images #1298
- [x] Video #1298
- [ ] Audio
- [ ] Document

richardbporter:
See #1297 for description.


joewhitsitt:
i don't like this for file, audio until we have an alternative workflow or documentation for people to use.

richardbporter:
I don't either but I don't like this more:
>To preserve website uptime and availability, either use a content delivery network (such as Acquia Cloud Edge) or another external storage system for serving large assets.

https://docs.acquia.com/acquia-cloud/performance/varnish/files/#varnish-10mb

joewhitsitt:
I think there should be a larger discussion with ITS about providing a web accessible large storage service that we can integrate with: https://its.uiowa.edu/categories/file-storage-and-sharing

I will approve a PR that just focused on images as a step in the right direction.

richardbporter:
Do we not support remote embeds? Isn't that it for integration?

richardbporter:
It sounds like we need alternatives for uploading 11-50MB files either with remote embeds and/or documentation.

joewhitsitt:
Just added documentation about image compression: https://sitenow.uiowa.edu/documentation/media/images-and-files

richardbporter:
Paging @joewhitsitt...

joewhitsitt (approved):

PR #1299, opened 2020-03-30
Honeypot/Webform Timeout

joewhitsitt:
Related to #798 

Attempt no. 2 at preventing webform spam. This sets a minimum 15 second count on submissions. If the form is submitted before 15 seconds, there is an error and the user must wait so many seconds before trying again. This disables page cache so if we stick with this, we should update docs to recommend not placing webforms all over the place.

The human part of this PR can be tested, it needs to go to PROD to be tested by the bots.


briand44 (approved):

richardbporter:
Not a huge fan of this considering webforms can be placed anywhere.

>recommend not placing webforms all over the place.

Might not be enough as opposed to directing people to one webform page.

PR #1301, opened 2020-03-30
Feature: Collegiate split updates + Grad College site provisions

pyrello:
# Base profile
 * Updated webmaster role permissions to allow editing the "authored on" field for basic page and person content types. This change was applied to base profile and collegiate split
* Added uids_base.settings config to ignored config for sitenow profile to allow these settings to be managed per site without requiring every site to have a split.

# Collegiate split
* Updated UUID, since it was created using the same UUID as the uiowa.edu split, which was preventing them both from being installed.
* Updated to reflect changes to uiowa.edu split since this was created as a copy of that.

# uiowa.edu split
* Removed items from site split that were represented in collegiate split.

# Grad College sites
* Recreated informatics.grad.uiowa.edu using sitenow profile.
* Updated cogscilang-temp.grad.uiowa.edu to use sitenow settings.
* Initialized immuno.grad.uiowa.edu, grad.uiowa.edu, neuroscience.grad.uiowa.edu, genetics.grad.uiowa.edu, gss.grad.uiowa.edu, and slis.uiowa.edu.


richardbporter:
Its hard to review large scale config changes but it seems like config is just shifting around. If I wanted to test this, should I try to sync these collegiate sites from test or prod?

richardbporter (approved):

PR #1316, opened 2020-04-01
Import config ignore and collegiate split settings before full import runs.

richardbporter:
Hotfix for the homepage in stage.


richardbporter:
Sounds like we're back to where we were and this can be closed. This may have prevented any config from being lost in between imports, although I'm still seeing some wonkiness locally. I think additional config entities like uids_base.settings would have need to be imported manually as well.

PR #1317, opened 2020-04-01
Scope sitenow_media_wysiwyg library to admin pages.

### LINKED ISSUE #1305: Investigate why ckeditor.js loads for public facing site
# Problem 
`<script src="/core/assets/vendor/ckeditor/ckeditor.js?v=4.14.0"></script>` loads for all sites even when logged out of the site. 

# Proposed Solution

So far, I have enabled Stark as the theme and the file still appears.

richardbporter:
Resolves #1305 

The sitenow_media_wysiwyg override library has a dependency on core/ckeditor. Its being attached for all pages which seems odd. This changes it to admin-only pages although there may be a better way. It'd be nice if we could attach by preprocessing/altering a WYSIWYG-enabled field.

@joewhitsitt @briand44 Do you know why we scoped this so widely to begin with?


joewhitsitt (approved):
Looks like my fine work in stripping down the table wizard. Probably inexperience with D8 libraries.

richardbporter:
Anyone have a better solution here? Even adding on every admin page is overkill.

pyrello:
It looks like there is a libraries-extend property that can be added to a theme. Not sure how to solve at the module level.

https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/themes/claro/claro.info.yml#L129-131

richardbporter:
I have another way of doing this.

richardbporter:
I combined this with another CKEditor config override we had in the profile. Seems less of a core/profile type functionality so the module might be more appropriate. I also added some comments with what I think these overrides are trying to do.

joewhitsitt:
Approved in the current state 👍

Issue #1275, opened 2020-03-27
As a site owner, I can turn on hawk alerts and/or specify a custom hawk alert message.


mjoneill:
- [x] #1277 Ability to select message options: Feed/Feed + Custom Message/Custom Only
- [x] #1278 Color selection? for alert levels
- [x] #1280 Meaningful a11y experience

richardbporter:
Do we want active Hawk Alerts to display while logged in? I would think, yes?

mjoneill:
I vote yes..  that feels like the right answer, provided it doesn't expose a whole dimension of pain to make it do that. 

^^^ also, not incredibly passionate about this.  So I'd defer to someone with a strong opinion, even if it's in the opposite direction.

PR #1321, opened 2020-04-02
Fix typo in config_ignore config.

richardbporter:
~~Once #1275 is done, I would imagine this module would be enabled by default and the config would be ignored.~~


pyrello (approved):
D'oh! Two typos in this file in as many days. This just isn't my week :|

joewhitsitt (approved):

Issue #1200, opened 2020-03-17
Add Colorbox to frontend build process


joewhitsitt:
## Problem
As raised in https://github.com/uiowa/uiowa/pull/1177, the colorbox implementation could be cleaned up.

Since the current implementation is a patched, library overridden version, style/js hybrid of Colorbox, I am not sure to what extent this can be done.

richardbporter:
Probably the same as the easysvg package - fork and require.

joewhitsitt:
This refactor would apply to both since they have the assets local.

joewhitsitt:
So, I started to do this in https://github.com/uiowa/uiowa/pull/2521 and have created a fork with patch of the repo https://github.com/joewhitsitt/colorbox, but I am getting a composer error when I try to use it/update the lock file.

`No valid composer.json was found in any branch or tag of git@github.com:joewhitsitt/colorbox, could not load a package from it.`

I revisited because I didn't want to start leaflet off on the wrong foot.

joewhitsitt:
@richardbporter since you helped with my easysvg fork, I was wondering if you knew how to get around this one?

richardbporter:
Yea - I'm curious what exactly was patched with it though as there may be another option. Was it only jackmoore#753 and jackmoore#765?

joewhitsitt:
This is the pull that has everything I think: https://github.com/jackmoore/colorbox/pull/832

richardbporter:
I gave this a shot in https://github.com/uiowa/uiowa/pull/2521/commits/8d26e81d90ab0caf2fbc90525d36d8f935bc1d65. This approach would not require you to fork the repo. That was goofy anyway as this package is not available on Packagist. 

Can you verify this works as expected? Any other thoughts?

Also, it would seem we may want to start thinking about a better supported lightbox solution for the future.

joewhitsitt:
I will give this a shot in a bit. Consumed by another bug on that branch.

joewhitsitt:
I think it worked, unfortunately, I don't think the a11y fixes in that branch are doing everything I want them to now that I am revisiting it...

PR #1330, opened 2020-04-03
Feature: uiowa_core module

### LINKED ISSUE #1273: Create module to collect additional site information
Adds the following fields:
- [x] Has a parent organization - checkbox which shows the next 2 fields
- [x] Organization name - textfield
- [x] Organization link - URL

### LINKED ISSUE #1328: [Story] As a site owner, I see both the campus-wide GTM and a local GTM loaded
Additionally, if I, the site owner, abuse the functionality, an administrator can turn it off without having access to the account.

- [x] Implement Campus-wide GTM in a uiowa_core module that all SiteNow profile have enabled.
- [x] Local GTM https://github.com/uiowa/uiowa/issues/1128
- [x] Feature branch update uiowa_bar to remove gtag stuff.
- [x] New release of uiowa_bar that is deployed at the same time that uiowa_core is turned on.

pyrello:
# Site Information form updates
Resolves #1273 
In addition to adding fields to the Site Information form, this also changes the way that uids_base theme choose which version of the header to display. First it checks to see if the site path matches 'sites/uiowa.edu' and if so uses the homepage header. Second it checks to see if the `has_parent` value is true and if so, it uses the parent org name and url to show the "tertiary" header. If neither of the previous two conditions are met, then it uses the "secondary" header.

## To test
1. `blt ds --site=uiowa.edu`
2. Check that https://uiowa.local.drupal.uiowa.edu/ is still showing the correct header.
3. `blt ds --site=immuno.grad.uiowa.edu`
4. `drush @gradimmuno.local uli` and follow link.
5. Visit http://gradimmuno.local.drupal.uiowa.edu/admin/appearance/settings/uids_base and save with the default settings.
6. Check that https://gradimmuno.local.drupal.uiowa.edu/ is showing the "inline" header:
    Screen Shot 2020-04-10 at 6 12 58 PM
7. Visit http://gradimmuno.local.drupal.uiowa.edu/admin/appearance/settings/uids_base and change the Site name display to "Display below the IOWA bar" and save.
8. Check that https://gradimmuno.local.drupal.uiowa.edu/ is showing the "below" header:
    Screen Shot 2020-04-15 at 4 26 50 PM
9. Don't worry about the fact that nothing is showing in the yellow bar - its because of the 'legacy' setting.
10. Visit https://gradimmuno.local.drupal.uiowa.edu/admin/config/system/site-information and add some parent organization information.
11. Check that https://gradimmuno.local.drupal.uiowa.edu/ is showing the "below" header and has the correct parent name and link in the yellow bar:
    Screen Shot 2020-04-10 at 6 12 27 PM
12. Visit http://gradimmuno.local.drupal.uiowa.edu/admin/appearance/settings/uids_base and verify that the settings to change the Site name display are disabled.
13. Verify that when you change the Header navigation to something other than 'toggle', the option to turn on Sticky is no longer visible and click save.
14. Visit https://gradimmuno.local.drupal.uiowa.edu/ and use the Inspector to verify that the body has the class 'header-not-sticky' and does not have 'header-sticky'.

# GTM updates

Resolves #1328 

This handles both campus-wide GTM and local site-specific GTM. The campus-wide GTM is loaded as it was previously in uiowa_bar (production only). The local site-specific GTM utilizes the contrib GTM module to allow multiple containers with separate configuration. Though a webmaster can go ahead and configure settings for the container, each container displays a warning notifiying the webmaster that to be inserted and used within the site, they must contact an administrator to enable the functionality. Additionally, non-admins are prevented from configuring the global GTM settings through the contrib module.

For the admins, the local GTM functionality is controlled through a boolean on the uiowa_core settings form. This toggle is checked on GTM insert to decide at the last moment whether or not to insert the configured containers. This uses the contrib module api hook. Additionally, in case GTM does something to the UI interface, the boolean can also be toggled through a custom drush command which will also clear caches and attempt to purge the domain.

Though I believe this resolves the story, deployment must be done in coordination with https://github.com/uiowa/uiowa_bar/pull/55

## To Test

Due to the production assumptions of this module. Testing is kind of complicated. Since this functionality is off by default, I would suggest the following to approve it for PROD where additional testing can occur:

- Temporarily disable the AH_PRODUCTION check in uiowa_core.module locally around line 19 to see whether the campus-wide GTM is loaded similar to the uiowa_bar method.
- Confirm that non-admins can't access:
 - https://sitenowdemo.local.drupal.uiowa.edu/admin/config/system/google-tag/settings
 - https://sitenowdemo.local.drupal.uiowa.edu/admin/config/system/uiowa-core
- Toggle the boolean with the custom drush command: `drush uiowa_core:toggle-gtag`


richardbporter:
~~What is the difference between this and the uiowa_edu_core module?~~

Nevermind!

pyrello:
@richardbporter The difference is that this is intended to be used on all sites, while the uiowa_edu_core module is specifically scoped to uiowa.edu.

pyrello:
In case anyone is helping me troubleshoot the issue I have encountered, here is the error message:

https://gist.github.com/pyrello/0242e3b3b9fd107226194461a4c804c8

joewhitsitt:
switching gears for moment, but wanted to put this here: changing the module and all of the references to `iowa_core` instead resulted in the bug going away and allowing me to proceed to the basic site settings form:
Screen Shot 2020-04-06 at 1 03 44 PM

I know changing the module name is not the path forward but might help in pinpointing the overall issue.

richardbporter:
I looked at this today but did not have any luck. I did notice that the `SiteInformationForm` is marked as `@internal` and maybe should not be extended? https://www.drupal.org/project/drupal/issues/2873696

Would an old-school form alter work for this? The uiowa_footer module is already doing this for the site information form.

pyrello:
@joewhitsitt Just to make sure I understand. You were not able to get to the form at all until you changed the module name? Or just unable to submit the form?

@richardbporter I'm open to that as an option. Generally speaking, I think that there is a move away from form_alters and similar sorts of .module code, so I was trying to embrace that.

joewhitsitt:
i couldn't navigate around the /admin/config/* section of the backend without error

pyrello:
Huh... that is different than the issue I was having. I could view the Site Information form without issue (as in your image above), but when I would click the button to save the form, I would get the error indicating that it couldn't find the `system.site` config.

pyrello:
@joewhitsitt As a webmaster, I can view this page: https://sitenowdemo.local.drupal.uiowa.edu/admin/config/system/uiowa-core

It seems like your instructions indicate that I should not be able to do that. Can you confirm?

joewhitsitt:
@pyrello have you recently pulled and cim'd? i made the permission change this morning: https://github.com/uiowa/uiowa/pull/1330/commits/807914a495ff5f9def0df0b3aaed7a757b63e30c#diff-137378a30894e649172b26407e15a09aL58

pyrello:
@joewhitsitt I think I ran the `blt ds` then git pulled, so that makes sense. I've tested your part of the PR and it looks good.

bspeare:
@pyrello I love the functionality that you built in for the header, but I'm not sure if the header will work for all sites.   I think we need the ability to switch from the `secondary` and `tertiary` headers without it being based on a parent link in the top bar.     The `tertiary` version was designed specifically for sites that also didn't need a super large Block Iowa logo as well as having a long site name.   Maybe that means an additional checkbox that says "Use slim version of Block Iowa logo" or something like that?

pyrello:
So, @bspeare, just to be clear, you want someone to be able to have a tertiary header with no name in the bar? Like this?
Screen Shot 2020-04-13 at 12 16 53 PM

bspeare:
@pyrello Yes, that is correct.

joewhitsitt:
Added draft to document the new "parent org" fields for standard sitenow v2 sites. Will publish when this is merged/deployed. https://sitenow.uiowa.edu/node/56/latest

joewhitsitt (commented):
My `dsa` doesn't trigger a `blt frontend`, but after that, the parent org functionality worked as expected.

richardbporter:
> My dsa doesn't trigger a blt frontend, but after that, the parent org functionality worked as expected.

It doesn't trigger one - that is a manual step.

joewhitsitt:
- Created follow-up issue: https://github.com/uiowa/uiowa/issues/1375
- Merged https://github.com/uiowa/uiowa_bar/pull/55
- [x] A release of uiowa_bar and a composer update still need to be done as part of this

pyrello:
So, as I'm understanding it, we have the following header options:
1. uiowa.edu:
    Screen Shot 2020-04-13 at 3 34 01 PM
2. Thick bar with site name in it:
    Screen Shot 2020-04-13 at 3 34 32 PM
3. Thin bar with parent organization:
    Screen Shot 2020-04-13 at 3 34 57 PM
4. Thin bar with site name below it:
    Screen Shot 2020-04-13 at 3 35 31 PM

The rules, as far as I can figure them out are:
1. Only the hompage.
2 & 4. Sites can opt for either option if they don't have a parent organization.
3. Sites must uses this if they have a parent organization.

Or am I thinking about this wrong? Is it a separate setting in the theme that they can choose to display a parent organization if one exists? And they can choose 2 or 4 if they either don't have a parent org or choose not to display it?

pyrello:
Here is the proposed functionality:

The site information form doesn't change and allows the user to choose to add a parent org an URL.

The uids_base theme settings for would replace the current header type dropdown field with the 2 radio options:
* Display name in the IOWA bar
* Display name below the IOWA bar
The user could only select one of the above options. If the user has already entered a parent (or later enters a parent), the field above would be disabled (it still shows but can't be changed) and the "Display name below the IOWA bar" option is selected. There could be additional help text explaining why the selection is locked and provide a link to the site information form for them to remove the parent org info, if they wanted to change it.

Please provide feedback!

joewhitsitt:
I think for the most part, the proposed solution makes sense. I am thinking ahead with regard to https://github.com/uiowa/uiowa/issues/1348 and how the type of site could determine header placement and further collection of info of things so that there is consistency across the university.

Wearing the SiteNow hat, if a site is supposed to follow the brand and the student org isn't allowed a certain header placement, it would be nice if a site-provisioner could just select the type of site and decide placement off that instead of going back in forth with client/osc about politics.

pyrello:
@bspeare Will you address @joewhitsitt's point about student orgs? Are they allowed to use either header 2 or 4?

@joewhitsitt I think that developing a list of site types is potentially problematic for a few reasons. It has to be a thing that makes sense to a site requester, so the list should encompass all the possible options so people aren't confused and are forced to choose a type of site that isn't correct for them. But those options are still going to map back to just a few possible configurations. That is why I'm trying to figure out what the boundaries are here. I wasn't even aware that #4 was a possibility until today...

bspeare:
@joewhitsitt @pyrello I think the Student Org is a good question.  I would prefer to see a student org use option 4 by default.   I think we can bring up those types of governance decisions in the design meeting on Wednesday morning for feedback.

pyrello:
@bspeare Your preference is noted. When surfacing this in the design meeting, let's keep in mind that we have to have a way to either make this choice without the user knowing it was available (forcing student orgs to 4 by default) or make the choice available to the user in a way that is straightforward and makes sense.

docroot/themes/custom/uids_base/theme-settings.php
```diff
@@ -21,6 +24,24 @@ function uids_base_form_system_theme_settings_alter(&$form, FormStateInterface $
...
+    '#title' => t('Site name display'),
+    '#description' => t('Select an option'),
+    '#options' => [
+      'inline' => t('Display inline with the IOWA bar'),
+      'below' => t('Display below the IOWA bar'),
+    ],
+    '#default_value' => theme_get_setting('header.type'),
+  ];
+  if ($has_parent) {
+    $test = Url::fromUri('internal:/admin/config/system/site-information');
```
joewhitsitt: unused var

docroot/themes/custom/uids_base/theme-settings.php
```diff
@@ -21,6 +24,24 @@ function uids_base_form_system_theme_settings_alter(&$form, FormStateInterface $
...
+    $form['header']['type']['#disabled'] = TRUE;
+    $form['header']['type']['#default_value'] = 'below';
+    $form['header']['type']['#description'] = t('This option is disabled because a parent organization was set on the <a href=":site-settings-page">site settings page</a>. When you have a parent organization, your site name will <em>always</em> display on the line below. You will need to remove the parent organization information to select another option.', [
+      ':site-settings-page' => Url::fromUri('internal:/admin/config/system/site-information')->toString(),
```
joewhitsitt: Better to use route than string path `Url::fromRoute('system.site_information_settings'...`

joewhitsitt (approved):
I see improvements, not blockers. Also biased related the GTM stuff

pyrello:
I made updates per @joewhitsitt's suggestions. Also fixed some coding standards warnings and errors. Also, put in a temporary workaround for the fact that every site has outdated config for theme settings, which needs to be corrected in a follow-up issue.

richardbporter (commented):
Can we not hard-code JS tag manager (https://github.com/uiowa/uiowa/pull/1330/files/050eae6b3a6805105a9a1c89855cf94c29b440e9#diff-6fbeff958f2ca4054b860e34486ea914)? i.e. download it through the build process somehow?

pyrello:
@richardbporter Would it be sufficient to do this in a follow-up issue? Or do you consider this blocking?

joewhitsitt:
the tag manager stuff was a port from the uiowa_bar 8.x which was a port from the 7.x uiowa_bar. i'd vote that this is not a blocker but a follow-up

richardbporter:
Yea thats fine. Probably similar to #1200.

mark-bennett-uiowa:
This all looks good to me. Tested as described above. No issues.

mark-bennett-uiowa (approved):
Tested as instructed. Verified changes were successful. Approved with no issues.

PR #1346, opened 2020-04-07
Change uiowa_alerts XHR data type to JSON so cache is used.

richardbporter:
Related to the e.uiowa.edu D7 caching issue.

The default value for the `$.ajax()` cache setting is false when `jsonp` is set for the `dataType`. This PR changes it to JSON now that the e.uiowa.edu API has an Access-Control-Allow-Origin header set which makes JSONP unnecessary for subverting CORS.

https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings


pyrello (approved):
Reviewed not tested.

PR #1360, opened 2020-04-08
Feature: Update webmaster permission and add iowasuperfund.uiowa.edu split

### LINKED ISSUE #1355: Allow webmaster role to choose unpublished node as parent menu item
What the webmaster sees:
Screen Shot 2020-04-08 at 11 08 49 AM

What the administrator sees:
Screen Shot 2020-04-08 at 11 07 11 AM

The administrator role is able to do this by virtue of the fact that it has the "Bypass content access control" permission: 
https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/lib/Drupal/Core/Menu/DefaultMenuLinkTreeManipulators.php#L141-148

But it is not clear to me by what mechanism access is being disallowed to the Webmaster role. The webmaster role has the following permissions, which would seem to allow access:
* Basic page: Create new content
* Basic page: Edit any content
* Basic page: View revisions
* View published content
* View own unpublished content
* View any unpublished content
* View the latest version

The "Administer content" role does not accomplish the desired outcome. The only thing that I've found so far that works is adding the "Bypass content access control" permission to the webmaster role.

Are there concerns about allowing this permission for the webmaster role? Is there another way to allow this access to webmasters?

pyrello:
Resolves #1355


bspeare (approved):
I tested this locally and while masquerading as a webmaster I was able to view an unpublished node as an option in the menu settings.

joewhitsitt:
don't merge this in yet. i will explain in standup

pyrello:
@joewhitsitt Just in the nick of time

joewhitsitt (approved):
Thanks for taking the extra step regarding custom sites

PR #1362, opened 2020-04-09
Require drupal/webp:1.x-dev to hide config page from webmasters.

richardbporter:
WebP 1.0-beta2 uses an odd permission for its config page that results in webmasters being able to access. It has been "fixed" in the latest dev but there is no stable release yet. Its still an odd permission but does hide the config page from webmasters.

https://www.drupal.org/project/webp/issues/3050939
https://git.drupalcode.org/project/webp/-/compare/8.x-1.0-beta2...8.x-1.x


joewhitsitt (approved):

richardbporter:
There are no database updates or config changes. I tested locally and everything seemed to work. I had originally set this module up using the docs on the project page which were not reflective of the version but now they are.

richardbporter:
I think this adds ImageMagick support for derivative generation. 👍

PR #1380, opened 2020-04-14
Fix preview error caused by moderation check

### LINKED ISSUE #1378: Preview link when editing returns server error 500
## Problem
Customer reporting that they cannot use Preview link when editing content… it returns a 500 error.

## Notes
This might be another issue related to https://github.com/uiowa/uiowa/pull/1268

joewhitsitt:
Resolves #1378 where if revision id is null, the following moderation state get string fails resulting in a 500 error on preview.

This now checks for a revision id before moving forward with the moderation check/status message.

## To Test
Editors can successfully view published/needs review/draft states of their node content including previewing their current changes and looking back at old revisions.


pyrello (approved):
Reviewed, not tested. Looks sound

PR #1385, opened 2020-04-15
Clean up frontend dependencies.

richardbporter:
This PR removes fractal as a build dependency in combination with https://github.com/uiowa/uids/pull/73 and https://github.com/uiowa/hds/pull/85. It also adds some additional options to `blt frontend` which makes it explicitly a production install. It is possible to install devDependencies by manually running `yarn install` if necessary. I added jquery and popper.js to uiowa_bootstrap which resolves a warning during the install process.

I do not have fractal installed in node_modules after this. I did `rm -rf node_modules` beforehand though. I'm not if everyone will have to do that as well.

We'll want to change the required versions before merging this.

- [x] Update uids and hds versions in uids_base and collegiate_theme, respectively.

### Testing
- Check out branch locally and run `blt frontend`.
- Make sure fractal is not in node_modules (see note above)
- Make sure sites look as expected for uiowa_bootstrap (hr, brand), collegiate_theme and uids_base themes.

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -47,7 +47,7 @@ validate:
 command-hooks:
   frontend-reqs:
     dir: ${docroot}
-    command: 'yarn install'
+    command: 'yarn install --production --ignore-optional --non-interactive'
   frontend-assets:
     dir: ${docroot}
     command: 'yarn workspaces run build'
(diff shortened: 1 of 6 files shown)


richardbporter:
I hit HR, brand, demo, home locally and everything looked normal.

joewhitsitt (approved):
nuked node_modules locally and was successful with a `blt frontend`. no fractal. pinged gis.sites and uiowa.edu with no visual issues

PR #1387, opened 2020-04-15
Require SSH agent to be running during vagrant up.

### LINKED ISSUE #937: How to add ssh-key automatically on computer start?
How do we make sure that this command is run when computers get restarted?

```
ssh-add -K ~/.ssh/id_rsa
```

richardbporter:
Resolves #937 

I just stumbled upon this and it seemed like a quick "fix": https://www.phase2technology.com/blog/sometimes-vagrant

I don't think we can or want to be starting the agent for anyone. That seems intrusive and it is possible to have multiple keys for different services. 

This doesn't really resolve the issue but I'm not sure what else we can do. Adding some docs to the Wiki seems appropriate.


pyrello:
@richardbporter When you tested this, is there a command you used to clear keys?

richardbporter:
`ssh-add -D` should remove them all.

pyrello (approved):
Looks good.

Issue #1281, opened 2020-03-27
[Story] As a web developer, I can see that modules and other dependencies have been updated before the homepage go-live.


joewhitsitt:
~~Require https://www.drupal.org/project/lb_claro~~ #1233 
- [x] drupal/acquia_connector
- [x] drupal/address
- [x] ~~drupal/administerusersbyrole~~ #141 
- [x] drupal/coder
- [x] drupal/config_ignore
- [x] drupal/crop
- [x] drupal/entity_browser
- [x] drupal/entity_embed
- [x] drupal/entity_reference_revisions
- [x] drupal/entity_usage
- [x] drupal/field_group
- [x] drupal/file_mdm
- [x] drupal/file_mdm_exif
- [x] drupal/focal_point
- [x] drupal/google_analytics
- [x] drupal/imagecache_external
- [x] drupal/imagemagick
- [x] drupal/inline_entity_form
- [x] #1427 drupal/layout_builder_restrictions @pyrello 
- [x] drupal/layout_builder_styles
- [x] drupal/lb_direct_add
- [x] drupal/masquerade
- [x] drupal/media_entity_twitter
- [x] drupal/media_library_theme_reset
~~drupal/menu_block~~ https://www.drupal.org/project/menu_block/issues/3066521
- [x] drupal/menu_breadcrumb
- [x] drupal/metatag
- [x] drupal/node_revision_delete
- [x] drupal/paragraphs
- [x] drupal/robotstxt
- [x] drupal/simplei
- [x] drupal/stage_file_proxy
- [x] drupal/twig_tweak
- [x] drupal/webform
- [x] drupal/workbench_email
---
- [x] drupal/colorbox               
- [x] drupal/config_update                           
- [x] drupal/ctools                 
- [x] drupal/date_recur_modular     
- [x] drupal/externalauth Requires uiowa_auth update.
- [x] drupal/mailsystem               
~~drupal/migrate_plus~~ 5.x not yet recommended.
- [x] drupal/search_api             
- [x] drupal/search_api_autocomplete
- [x] drupal/simple_sitemap         
---
- [x] drupal/core
---
- [x] acquia/blt --with-dependencies
--- 
- [x] PHP 7.3

richardbporter:
I think dependencies should be updated one at a time, if possible, and tested thoroughly. The BLT update process makes this difficult as it involves an open ended `composer update` so that should be updated last.

richardbporter:
[...]

PR #1419, opened 2020-04-17
Update drupal/acquia_connector to 1.21.0.

richardbporter:
Related to #1281 

Acquia Connector settings were config_ignored-ed due to https://www.drupal.org/project/acquia_connector/issues/2864697. This PR removes that ignored config. It also:

- cleans up some cruft when we tried to split the module per environment
- lowers the reporting time to every 12 hours
  - Seems overkill to report this to frequently
- Removes a config override to hide signup messages locally since the config in sync is now used.


richardbporter:
I don't think this can be tested beyond a `blt dsa` and making sure config is not overridden. It wasn't for few sites I synced. Its also in dev.

pyrello (approved):
Reviewed not tested

PR #1424, opened 2020-04-17
Creation of policy.clas.uiowa.edu config split

mark-bennett-uiowa:
Created policy.clas.uiowa.edu config split and populated with initial settings.


pyrello (approved):
No impact on other sites because it has to be enabled. Reviewed not tested.

PR #1426, opened 2020-04-20
Feature: Contact Info (content) block

### LINKED ISSUE #1293: Include "Contact info" block in default content
Screen Shot 2020-03-30 at 8 51 08 AM

As part of providing a set of footer components that are preconfigured out of the box when a new SiteNow site is setup in uids_base, I think we should provide a "Contact info" block. The default content could be pretty minimal or empty and since the block is content once it is added to the site, any changes made on the site would not be affected by the config update process.

### LINKED ISSUE #1425: Basic page LB template doesn't account for all default fields
And this causes a situation where LB "generates" a template which is a combination of the the defined template + every field which is not already defined getting dropped into the first section. Some even get defined twice, inexplicably.

This is observable in any new SiteNow site after the collegiate split is enabled and configuration is imported.

pyrello:
Resolves: #1293 
Also resolves: #1425 

# To test
* Add the following sites to your multisites array in your `local.blt.yml` file and run `blt dsa`.
* `uiowa.edu` # This site should be unaffected by changes - the contact block should match stage and there should only be one of them listed on custom blocks page (https://uiowa.local.drupal.uiowa.edu/admin/structure/block/block-content)
* `brand.uiowa.edu` # This site should now have the default contact info block showing up in the footer (https://brand.local.drupal.uiowa.edu/).
* `coronavirus.uiowa.edu` # This site should now have the default contact info block showing up on the block page (https://coronavirus.local.drupal.uiowa.edu/admin/structure/block/block-content) but not showing up anywhere since it still uses uiowa_bootstrap.
  Screen Shot 2020-04-23 at 11 33 54 AM

# Description and request for feedback
As part of the homepage project, a contact info block was created and a block definition was included with UIDS base. When a website is created with SiteNow + the Collegiate split, this block definition is imported, but it refers to block content that doesn't exist.

This PR would fix that.

Is this the right approach? Relying on content being instantiated for a block to be defined seems awkward. Additionally, should this include an update for sites that don't already have this block defined to install it, so that when every site eventually moves over to using uids_base, it would exist already?

Additionally, this also includes a fix for a problem that stems from the fact that the current page LB template doesn't include all fields and this causes LB to generate a new template that duplicates some fields.

Looking for feedback!


mark-bennett-uiowa (commented):
Looked through the code, not tested yet. Looks okay on the first file. Do we want references to a site.local.drupal.uiowa.edu though, or I may be looking at that second file wrong. Ill test if I can shortly.

pyrello:
@mark-bennett-uiowa I don't think that the references matter because I don't think they get imported. I switched them to localhost because that is consistent with what SiteNow does.

richardbporter:
We talked out of band about creating this block for every site but not placing it since that is handled by uids_base. I think that makes sense since every site will need this block content.

bspeare (approved):
I tested all 3 sites locally and they worked as expected.

PR #1427, opened 2020-04-20
Feature: Updated layout_builder_restrictions to latest version.

pyrello:


bspeare (approved):
I tested this locally and all restrictions on the blocks remain in place and I was able to add new block restrictions as well.

PR #1433, opened 2020-04-20
Add core patch which sets img dimensions with fallback info

### LINKED ISSUE #1406: Console warning about lazy loaded image with no dimensions
`[Intervention] An  element was lazyloaded with loading=lazy, but had no dimensions specified. Specifying dimensions improves performance.`

https://uiowa.stage.drupal.uiowa.edu/academics

joewhitsitt:
Resolves #1406 

## To Test
- Patch successfully applies to core.
- After `cr` all responsive view mode rendering images should have image dimensions for the `` tag.
- Console warning message from #1406 should not output.


bspeare (approved):
Looks good.  I like that it creates the space for the image to lazy-load on elements as you scroll down the page.

PR #1435, opened 2020-04-20
Fix References to Unpublished

### LINKED ISSUE #1428: Article authors can't reference and save with non-published people
## Problem
https://github.com/uiowa/uiowa/pull/1334#issuecomment-616568122

## Proposed Solution
Only allow published people references. Check to see if unpublished people can be added/saved on the fly. If not, set the published checkbox to true and disable it from being switched.

joewhitsitt:
Resolves #1428 

## To Test
- Module should be enabled for all sites.
- As a editor/publisher, you are able to create article content with draft authors without the error: `This entity (node: N) cannot be referenced`


dnkemper (approved):
This worked perfectly! Great job, Joe!

PR #1436, opened 2020-04-21
Feature: Banner and banner block template updates

### LINKED ISSUE #1438: Banner block contextual links not showing up - regression

pyrello:
Resolves #1438 

* Re-adds contextual links to the banner block template.
* Updates the link URL empty check in the banner template so that it isn't triggered unintentionally.


bspeare (approved):
I tested it out locally and it works great!  I love that the edit link is now on the far right as it should be - thanks!

PR #1446, opened 2020-04-21
SiteNow Blockquote for UIDS_Base

joewhitsitt:
I don't know what I am doing... criticism welcomed!


bspeare:
This looks like a great start.  Do you want to merge this in or wait until we have a better design for the blockquote?

joewhitsitt:
i am fine with whatever, i just wanted to get my toes wet. thanks for removing my hello test class. usually it is foo and i search for those before committing

docroot/themes/custom/uids_base/templates/uids/blockquote.html.twig
```diff
@@ -0,0 +1,24 @@
...
+ # * blockquote_attributes
+ # * blockquote_cite
+ # * blockquote_cite_url
+ # * blockquote_footer
+ # * blockquote_content
+ #
+ #}
+
+{% if (blockquote_cite_url is not empty) %}
+  <blockquote {{ blockquote_attributes }} cite="{{ blockquote_cite_url }}">
```
pyrello: I wonder if the `cite` attribute can be added to `blockquote_attributes`?

docroot/themes/custom/uids_base/templates/uids/blockquote.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{% else %}
+  <blockquote {{ blockquote_attributes }}>
+{% endif %}
+    {{ blockquote_content }}
+    {% if blockquote_footer is not empty or (blockquote_cite is not empty) %}
```
pyrello, line 18: I think this `or` condition should be changed to `and`. Otherwise, I think the `elseif` will never run.

bspeare (approved):
Looks good!

joewhitsitt:
I assume

>Removed blockquote styles to make way for update from UIDS.

is going to get pushed around the same time.

Thanks for the help!

bspeare:
@joewhitsitt It will be in https://github.com/uiowa/uiowa/pull/1597 if you are ok in a short delay after this gets merged in.

PR #1453, opened 2020-04-22
Updated exported config for policy.clas.uiowa.edu

pyrello:
This should fix the issue where config import is failing for this site.

# To test
`blt ds --site=policy.clas.uiowa.edu`. If this works without issue, then the problem has been resolved.


richardbporter (approved):
Tested. 👍

mark-bennett-uiowa (approved):
Tested and approved. Looks great, thanks so much.

Issue #1391, opened 2020-04-16
[Story] As a site editor, I can see a similar Drupal 8 backend theme across campus


joewhitsitt:
## Problem
The backend experience is fragmented across D8 sites and relying on Adminimal to be ready for D9 is one more thing we have to think about later.

## Proposed Solution
Revisit adopting Claro for the base profile and switch all sites over to use it once we have ironed out a few issues that remain with the experimental theme as it relates to Paragraphs.

- [x] Port and adapt the space saving drag handle tweaks we have in Adminimal. https://github.com/uiowa/uiowa/blob/master/docroot/profiles/custom/sitenow/css/admin-overrides.css
- [x] Update paragraphs to take advantage of the recent Claro fixes, most notably: https://www.drupal.org/project/paragraphs/issues/3079128
- [x] Chosen looks skinny/dated and doesn't match the green focus style
- [x] experimental dropbutton spacing
- [ ] fields extending to the border of parent
- [x] Duplicate drag handles
- [ ] Consider tightening up the white-space when it relates to paragraphs.
- [x] Set claro as admin theme.
- ~~Remove adminimal~~ https://github.com/uiowa/uiowa/issues/1471

<img width="1077" alt="Screen Shot 2020-04-15 at 2 31 03 PM" src="https://user-images.githubusercontent.com/4663676/79477115-6bb1e580-7fcf-11ea-91be-68ca32f0e04d.png">
<img width="224" alt="Screen Shot 2020-04-15 at 2 19 27 PM" src="https://user-images.githubusercontent.com/4663676/79477128-70769980-7fcf-11ea-89f8-8a3377d50054.png">

>Most problematic is the loss of margin/padding when expanding paragraphs items. Also, the negative space is nice but the expanded pages makes it harder to work with our page building tools. A more compact UI is better for us.

joewhitsitt:
>Consider tightening up the white-space when it relates to paragraphs.

This is part of their meta issue: https://www.drupal.org/project/paragraphs/issues/3108997

joewhitsitt:
> fields extending to the border of parent

I don't know if we want to create a bunch of overrides again and just see what improvements come with later releases.

PR #1445, opened 2020-04-21
Update dependencies.

richardbporter:
Resolves #1281 
Resolves #1391 
Resolves #921
Resolves #1238
Resolves #853 
Resolves #141 

Configuration changes related to the following exist but were not updated:
- uiowa_lb_blocks (layout_builder_styles)
- entity_types_seminar (field_group, paragraphs)

Testing (dev)
- Log in to a dev site and test out any and all functionality.
  - https://uiowa.dev.drupal.uiowa.edu

Testing (local)
- Run `vagrant provision` locally to get PHP 7.3.
- Sync a few sites via `blt dsa` and ensure functionality (paragraphs, layout, default, custom) is as expected.

PR #1461, opened 2020-04-22
Switch from Adminimal to Claro

joewhitsitt:
Related to #1391


joewhitsitt:
Moving this into #1445

PR #1466, opened 2020-04-24
Extend bs classes in carousel picture img.

### LINKED ISSUE #1464: Carousel paragraphs component broken
It seems like the responsive img element is missing the required classes d-block and w-100 based on the docs: https://getbootstrap.com/docs/4.0/components/carousel/

richardbporter:
Resolves #1464


briand44 (approved):
Didn't test but looks pretty harmless.

joewhitsitt:
I think this is better than nothing for now. I am concerned with broken images, but this seems to solve lazy loading of non-broken images.

PR #1467, opened 2020-04-24
Hotfix webp image style derivatives.

richardbporter:
The patch fixes the issue but I added a couple config changes for good measure. I'm surprised core doesn't transliterate file extension by default...


richardbporter:
The webp module leverages Imagemagick (if available) regardless of 0f033c0 but it seems like a good idea to enable it anyway in case that changes in the future.

joewhitsitt (approved):
fixes my local webp 404 issues

PR #1505, opened 2020-05-01
Updated icsa-temp config split with content type and views.

mark-bennett-uiowa:
Added new content type for projects.
Added two views for displaying projects and reports.
Added new menus for Projects and Reports.
Added new data format to remove day name.
Enabled the datetime_range module.


pyrello (approved):
Reviewed, not tested. Scoped to site split.

Issue #1511, opened 2019-10-17
Investigate Fragments module


pyrello:
# Problem
There are use cases where we may want to have fragments of content that are not created using custom block types. The de facto standard for this type of thing is currently Paragraphs, but Paragraphs has some limitations and drawbacks. Notably, paragraphs cannot be created outside the context of a parent node.

# Proposed Solution
Investigate the Fragments module to see whether it is a better fit for the use cases where we need this type of content fragments.
https://www.drupal.org/project/fragments

joewhitsitt:
I took a look at this module briefly last week and I don't see a whole lot of a difference between using this and creating our own custom content entity that allows different bundles.

The one thing I see beneficial is that they figured out what to call it. Alternatively, I thought "Bits" works.

I am starting to see a lot of use-cases for something like this and will mention them over on https://github.com/uiowa/uiowa/issues/2096

pyrello:
The main value of this module is that it has already done the work of putting all the admin UI in place for creating stuff like this, if we roll our own, we'll have to create and maintain that.

pyrello:
I think the work of this issue is done and follow-up can be continued at #2096

joewhitsitt:
With only two sites using the Fragments module, we would probably end up maintaining it anyway. Views makes admin ui's pretty easy in D8/9 and allows us to build it how we want:

> Weirdly, the description field is required for a fragment type, which is annoying

`drush generate` does most of the initial creation work.

pyrello:
I look forward to your demo on this subject, since I know not of what you speak! :)

joewhitsitt:
<img width="434" alt="Screen Shot 2020-11-23 at 2 59 34 PM" src="https://user-images.githubusercontent.com/4663676/100014802-8f76ae80-2d9c-11eb-9cc6-e33bbe9da31b.png">

😜

pyrello:
Demos speak louder than pictures :)

PR #1529, opened 2020-05-04
Feature: Collection block type with view mode switching

### LINKED ISSUE #1555: As a site builder, I can add collections of items via Layout Builder
- [ ] Add block type for collection
- [ ] Add paragraph template for collection item
- [ ] Add block content template for accordion view mode
- [ ] Add block content template for tab view mode
- [ ] Clean up IEF embedded form in LB

pyrello:
Resolves #1555 

This adds a Collection custom block type that can be used to create sets of items in Layout Builder. It has the ability to to switch between different presentations based on view mode. This allows it to potentially be used as Accordions, Carousels, or Tabs.

So far, only some theming has been done for the Accordion view mode.


pyrello:
@bspeare noticed an issue when attempting to revert to a revision after the name of a block has been changed. I have not been able to figure it out yet, but I left a comment on the thread that the patch to IEF came from: https://www.drupal.org/project/inline_entity_form/issues/3002175#comment-13596759

pyrello:
Previous issue is resolved. Now I am hung up on a core issue where custom block entities don't invoke their theme hook and template when rendered as part of a list. There are a number of issues that touch on the subject, but this one probably best sums up the issue we are having: https://www.drupal.org/project/drupal/issues/2704331

joewhitsitt:
So, I haven't pulled this down yet, but a long time ago I was experimenting with a custom block type that referenced a custom content entity. Is there a strong reason you are trying to stick to blocks for these inner items?

I saw a demo of Twig Tweak (one of the workarounds) at DrupalCon Nashville. I would like to be cautious in adopting it and use it minimally if at all.

pyrello:
There may be some cases where we are using Twig Tweak still, but generally, I have tried to remove it wherever possible.

The rationale for using blocks as the nested component is just that the block seems to be the core "standard" for embedded content entities. I'm not opposed to structuring it another way, but had already mostly gotten this to work.

joewhitsitt:
So custom content entities are core too (less out of the box though) and it sounds like this idea ventures away from the standard by nesting. That linked issue seems to have a lot of opinions and has been kicked to D9 as feature request. I know it is backtracking, but a custom entity might require the least amount of fighting while still being revisionable, etc. You can still utilize IEF and all of that parent stuff.

pyrello:
@joewhitsitt Creating a custom content entity is not a small lift. You have to replicate a lot of infrastructure to get the admin pages to be able to utilize the field UI. The [ECK module](https://www.drupal.org/project/eck) exists to solve that problem to some degree, but I have used it before and not been very impressed with it.

I also opened an issue a long time ago to look into using the Fragments module (#1511), as a possible way to handle content items that are meant to be embedded, but should be manageable on their own.

I may end up trying to recreate this with Paragraphs, but last I checked, the ERR module was not handling Layout Builder stuff correctly. FWIW, when we made the decision to go with LB + Blocks, it was not a decision that was made lightly or without a LOT of consideration: https://github.com/uiowa/uiowa01/issues/15#issuecomment-523152766

My gut tells me that if I pursue another path on this, rather than trying to figure out the core patch that would be necessary to get the templates to work as expected, that I will end up facing a totally different set of issues that are unique to the paragraphs or custom entity subsystems that will be just as annoying and difficult to solve.

joewhitsitt:
I was just presenting an idea for the nested level of content beyond layout+blocks, not trying to question the original decision. I didn't know that a site-builder experience was part of the requirement and I was only thinking about the basic collection item structure I had previously talked about that forms a lot of the content in this PR (media item, title, content, meta info).

>will end up facing a totally different set of issues that are unique

You are working in Drupal. Issues always come up and they are almost always unique.

pyrello:
@joewhitsitt I appreciate your feedback and I hope I didn't give you the impression that I didn't.

I understand there are always going to be issues. My point is that I already spent a lot of time working through the issues to get to where I'm at currently with this PR and I have gotten to the point where I understand the lay of the land reasonably well. If I introduce a whole new subsystem of Drupal into the equation (e.g. Paragraphs) it will introduce a whole new site of corresponding issues and learning that will have to take place to solve them. Maybe it will "just work" but more likely there will be particular quirks of that system that I end up having to burn a day learning about.

pyrello:
A few notes for my own reference:
* Going to split carousel into its own block type and leave accordion and tabs together. This is because the way that accordion/tab function is similar and switching between them is logical. I don't think the same can be said for carousels.
* `hook_block_view_alter` doesn't get invoked when the view for a block via entity reference is built.
* I still need to try using the [block_content_template](https://github.com/swentel/block_content_template) module to see how it works.
* Instead of targeting block-related hooks, this might be solvable through field-related hooks. Here is one possibility for experimenting with: https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/template_preprocess_field/8.8.x
* Specifically, if there is a way to override the existing entity reference formatter class or add a new entity reference formatter class. Here is an article describing how to do the latter: https://www.lullabot.com/articles/extending-a-field-type-in-drupal-8
* I attempted setting up a container block that referenced paragraph items. I have not tried using the IEF widget yet, but the paragraphs widgets are unusable within Layout Builder - literally not functional in some parts. I was not able to get to the point of actually adding something to the page that used a paragraph to see what my theming options would be. I'll try switching the field widget to IEF just to get something to the page to see if it presents a different theme option we could hook into.

pyrello:
Related to #1555

pyrello:
A few thoughts that I could use some feedback on:
* I think that a carousel is a fundamentally different thing that an accordion or tabs. Unlike the latter 2 items, where you could potentially imagine a variety of blocks combined to form the content, a carousel should be pretty strictly defined. Our current banner or hero components make pretty good carousel slides.
* Therefore, I think it makes sense to define at least 2 separate block types: Carousel and Accordion/Tab.
* The Carousel block type would make use of the same kind of entity reference field where individual Banners/Heros can be added.
* The Accordion/Tab block type would reference a new block type called a Pane or Panel, which could in turn consist of a variety of blocks to form its content.

Thoughts?

pyrello:
@bspeare When you have a minute, could you take a look at two things for me?
* There is a javascript being used in our D8 platform that swaps out font awesome spans for an svg. This affects the way that the slider looks. I added an override style to try to fix it but it still doesn't look right: https://github.com/uiowa/uiowa/blob/b5567dc0270eab42b5f2f8e475e36e6753797e20/docroot/themes/custom/uids_base/scss/components/slider.scss#L6-L20
* The accordion seems to be rendering not full width. I cannot figure out why this is happening:
  Screen Shot 2020-07-15 at 4 24 35 PM

bspeare:
The slider looks good!  I noticed the accordion is outputting duplicate `<h2>` tags for the heading.  Are we able to strip that markup out here? https://github.com/uiowa/uiowa/blob/feature_lb_carousel/docroot/themes/custom/uids_base/templates/field/field--block-content--field-uiowa-accoridontabs-panels--uiowa-accordiontabs--accordion.html.twig#L41

bspeare:
Or maybe the accordion heading field needs to be changed to a text field since `<h2>` is coded into the template? https://github.com/uiowa/uiowa/blob/feature_lb_carousel/docroot/themes/custom/uids_base/templates/uids/accordion.html.twig#L14

pyrello:
That's a good question. I was wondering about how an accordion should render inside the document flow. Is the user able to change the `h*` value of the titles? For example, if you have a section heading, then a set of accordions, then they would probably use `h3`, but if there was no section heading, then they would likely use `h2`, right?

bspeare:
I think @joewhitsitt was going to test out the use of https://www.drupal.org/project/heading with the custom blocks in his layout builder tests, perhaps he will have some additional feedback on this.

pyrello:
I just mean, philosophically, do we want to allow people to change this or not. If we are going to allow people to set an `h*` for the accordion, we'll have to do it at the level of the accordion itself rather than for each panel, meaning we won't be using the heading field. I don't think we actually want people to be able to set each heading field individually, as that would be a mess from a document organization standpoint.

I think I just answered your earlier question earlier about whether that field should be a text field instead of a heading field.

joewhitsitt:
> I think @joewhitsitt was going to test out the use of https://www.drupal.org/project/heading with the custom blocks in his layout builder tests, perhaps he will have some additional feedback on this.

I was not looking at this. I was doing a fix for node teasers based on where they are placed (paragraphs view, entity reference, view page) #1900 

I am looking to make use of the card heading in teasers and create a variable/default that will allow the proper heading to be passed in.

For the most part the paragraphs were structured enough that we could assume what the header should be. All of our "collections" require a header though they can visually hide and assumes that it is an h2 below a h1 (page title) making all the node teasers and such an h3. teasers are limited to plain text so that there won't be bad headings after that. cards, accordion items, text areas is where we lose control.

pyrello:
The way that I have solved this, for now, is to add a field to the Accordion/Tabs block type for collecting the heading size. When rendering, this gets passed through to the individual panels and determines which `h` tag is used. This ensures that all accordions or tabs will have the same size `h` tag and makes it configurable.

Eventually, we may want to revisit this and attempt to determine programmatically what size each heading should be, but this gets us through for now.

pyrello:
The last major thing I am contemplating before moving this along is whether to swap out using paragraph for the nested field collections instead of custom block types. The rationale for doing this would be that paragraphs might be lighter weight in terms of their schema/config than blocks (because they are not meant to serve so many purposes).

I weighing this against the effort in going through replacing the already defined nested block elements with paragraph replacements.

I had also been thinking that this might mean that we are able to remove Paragraphs (and Entity Reference Revisions) from the project entirely, but now that I am thinking about it, I think that they still have valid use cases within more structured content types.

What do you think?

bspeare:
@pyrello What would the Paragraphs version look like?  Do you add it through the node edit screen or through layout builder?

pyrello:
@bspeare The experience from the user's perspective would be exactly the same.

pyrello:
I moved the slider into its own PR and I am going to start a different branch so I can start over with naming things. I think that the names I choose for the collections are not good considering how they might be used in the future.

PR #1535, opened 2020-05-05
Added link to parent org in the footer, if it exists.

pyrello:
Re-factors the footer template in uids_base to use the parent org information if it exists instead of slogan.


bspeare (approved):
I tested this locally with the UIDS version bump and it works!

Issue #1533, opened 2020-05-05
Investigate fixing lazy loading image jank / aspect ratio on cards


bspeare:
# The Problem

@quamsta mentioned:

```
I wanted to add a comment/idea on the Uiowa homepage card styles that Ben showed briefly. I recall that you landed on the lazysizes JS module for lazy loading images in the template. We also use this library and think it's great. Ive found the aspect ratio plugin for lazysizes to be very helpful in preventing page "jumps" while the image loads and changes the height of the container-- especially noticeable on mobile. Dunno if you guys have checked it out or if it's on your radar at all, but thought I'd drop a link in case it wasn't:
https://github.com/aFarkas/lazysizes/tree/gh-pages/plugins/aspectratio

We use this on After Class in order for our Masonry grid to work properly with lazysizes and not jump around and go completely bonkers as images load in since we're letting the heights of the images be completely arbitrary. 
https://afterclass.uiowa.edu/events/

You do need to be able to calculate the aspect ratio of the image and place it in the <img> or <picture> elements  before the javascript runs though (aka via the templating system, serverside). But I'm guessing that Drupal should have the height/width of the uploaded assets somewhere and you might be able to access them and calculate it that way!
```

PR #1538, opened 2020-05-05
turn on lazy aspect ratio plugin

joewhitsitt:
Relates to #1533 

So the lazy module has a plugin checkbox for aspect ratio in the configuration you can toggle on. With prefer native lazy turned off, this will be loaded. From what I am seeing, the over-extended jank up is gone (heros), but the jank down still occurs.


bspeare (approved):
Thanks! It's great that the module has that built-in.  It's tough to notice the difference but it might be worth trying out for now.

PR #1544, opened 2020-05-06
Updated icsa.uiowa.edu views with minor changes.

mark-bennett-uiowa:
Minor updates to the Projects and Reports views. These are site specific views for the icsa.uiowa.edu site and will not be used on other CLAS sites.

- Fixed menu setting in /projects/finished to show properly in the main navigation menu.
- Fixed line sharing for titles and dates in Projects views so that dates are on the same line as the titles.
- Attempted to fix menu item weights to correct order.


pyrello (approved):
Reviewed, not tested. icsa.uiowa.edu only.

PR #1546, opened 2020-05-07
Feature: Move cogscilang back to original domain

pyrello:
This PR transitions the cogscilang.grad.uiowa.edu site to be the same as the cogscilang-temp.grad.uiowa.edu as part of the process of transitioning that site.

After this deploys, I'll be syncing the database and files from cogscilang-temp > cogscilang.


joewhitsitt (approved):
reviewed, not tested

PR #1530, opened 2020-05-04
Adopt responsive images part 1

joewhitsitt:
Relates to #1499

For page, article, person teasers as well as the paragraphs card, this PR switches out the programatic single image styles (sitenow_card and sitenow_square_m) for close equivalent responsive image styles using drupal view modes.

Originally scoped to just touch brand.uiowa.edu for adoption of uids_base, with the preprocessing logic in SiteNow profile and custom modules, it was easier to just refactor it.

I don't think an image flush will be necessary for this update. I don't like it, but I am seeing the need for a double refresh on pages with several images.

Follow-up issues will be needed to set a replacement image style and to delete the unused image styles.

## To Test

- I recommend testing a site where you have all the files local and that stage file proxy is not turned on.
- `blt sync --site mysite.com --sync-public-files` and then `pmu` or uninstall stage file proxy locally before rendering front end pages.
- Start viewing pages with content type teaser images and paragraph cards implemented. The images should eventually load via lazy and you should see responsive image information instead of a single image tag. Refreshing the page a couple times might be necessary on pages with a lot of images.

## Split config that was not updated:
**core.entity_view_display.node.article.teaser.yml**
-  docroot/modules/custom/entity_types/entity_types_article/config/install/core.entity_view_display.node.article.teaser.yml

**core.entity_view_display.node.page.teaser.yml**

- config/features/collegiate/core.entity_view_display.node.page.teaser.yml
-  config/policy.clas.uiowa.edu/core.entity_view_display.node.page.teaser.yml
-  config/cogscilang.grad.uiowa.edu/core.entity_view_display.node.page.teaser.yml
- docroot/modules/custom/entity_types/entity_types_page/config/install/core.entity_view_display.node.page.teaser.yml

**core.entity_view_display.node.person.default.yml**

-  docroot/modules/custom/entity_types/entity_types_person/config/install/core.entity_view_display.node.person.default.yml

**core.entity_view_display.node.person.teaser.yml**

-  docroot/modules/custom/entity_types/entity_types_person/config/install/core.entity_view_display.node.person.teaser.yml

PR #1547, opened 2020-05-07
Updated policy_clas_uiowa_edu site split to contain entity_view_displ…

mark-bennett-uiowa:
Added wildcard core.entity_view_display.* to partial split to handle core entity updates in the future and fix current issue.


pyrello (changes requested):
This should be closed in favor of being solved as part of #1530

mark-bennett-uiowa:
So just close this PR?

joewhitsitt:
@pyrello or should it stay here and go out first?

joewhitsitt:
then the other PR can be tested again

pyrello:
The config exports will still need to happen in the other PR, so I don't see the point of doing them separately.

I applied this change to the other PR and ran `drush @claspolicy.local csex policy_clas_uiowa_edu`, which resulted in the same config files being added to the split. That does not happen in this PR.

joewhitsitt:
closing since duplicated in https://github.com/uiowa/uiowa/pull/1530

Issue #1541, opened 2020-05-06
Accordian items are not rendering unique id value.


briand44:
I tried creating an anchor to an accordian item and it is not working. Upon further inspection the id is not rendering at all.

briand44:
Modei is the customer who was having issues with this in case we roll out a fix, we should communicate that to him.

GaryRidgway:
this should be solved in https://github.com/uiowa/uiowa/pull/1548, waiting on pr approval and then merging when we can.

PR #1548, opened 2020-05-07
adding the attributes to the accordion and putting the card in the at…

GaryRidgway:
…tributes


joewhitsitt:
what we need is the value from this (if exists and is not empty):
https://github.com/uiowa/uiowa/blob/master/docroot/modules/custom/sitenow_paragraphs/sitenow_paragraphs.module#L160

to swap in for label here:
https://github.com/uiowa/uiowa/blob/master/docroot/themes/custom/uiowa_bootstrap/templates/paragraphs/accordion/paragraph--accordion-item--default.html.twig#L42

and now here:
https://github.com/uiowa/uiowa/blob/master/docroot/themes/custom/uids_base/templates/paragraphs/accordion/paragraph--accordion-item--default.html.twig#L42

That way, if I type "hello" in the accordion item's unique id field, i should be able to navigate to that item via #hello URL anchor

GaryRidgway:
I was trying to mirror the twig templates of other paragraph items that have it on the outer most element. is that supposed to be different here? I felt it would be inconsistent if we changed that because other paragraphs, when given a unique id, put it on the outer most element. besides, the label and id stuff is programmatically generated for use by the js, so i don't know if it is correct to change that.

it also seemed that this usage of attributes was the way to properly apply the attributes of the default configuration of the data [passed to the twig template](https://www.drupal.org/docs/8/theming-drupal-8/using-attributes-in-templates), so i was trying to mirror that practice.

this also does jump to the given id with #unique-id, I have tested and verified that that works.

GaryRidgway:
I will add this to uids_base when we get the details worked out though, I did not think of having to do that, so I will go add it there afterwords as well.

joewhitsitt:
Fair enough. I can't remember off hand why the accordion item didn't get the same attributes treatment other than more than that id could potentially get added. The js won't be affected because the label variable in the template gets applied to where it needs to and the js isn't doing any string matches.

Update the uids version with your method and I will approve.

joewhitsitt (approved):

GaryRidgway:
resolves #1541

PR #1567, opened 2020-05-08
Accessible button labels for uiowa.edu vertical videos

### LINKED ISSUE #1447: button.vidbttn does not have an accessible name
## Problem

The homepage vertical video buttons don't have an accessible name to identify it beyond "button."

## Proposed Solution
Set an aria-label based on the video being displayed: "Play/pause video about ___"?

joewhitsitt:
Resolves #1447 

Turns the video title into a variable for use throughout the template and adds it as part of an aria-label for the play button.

Looks like title is not required based on other template logic so there is an id fallback label.


bspeare (approved):
Looks Good!  I added this issue to fix after running another Lighthouse report: https://github.com/uiowa/uids/issues/177.

bspeare:
@joewhitsitt I added detailed steps on https://github.com/uiowa/uids/issues/177 if you want to fold those changes in.

joewhitsitt:
blocking as https://github.com/uiowa/uids/pull/179 and a uids version bump need to happen.

PR #1568, opened 2020-05-08
Delete cogscilang-temp.grad.uiowa.edu multisite on uiowa01

pyrello:
Final step after moving cogscilang.grad.uiowa.edu from collegiate to sitenow profile.


joewhitsitt (approved):
reviewed, not tested

PR #1576, opened 2020-05-11
Updated policy.clas config to reflect updated carousel config.

pyrello:
This should fix the issue that is causing the the post deploy hook on uiowa01 to fail on policy.clas.uiowa.edu.


mark-bennett-uiowa (approved):
Reviewed not tested. This updates the config on the single policy.clas.uiowa.edu site. Approved.

Issue #1477, opened 2020-04-28
Post Collegiate clean-up


pyrello:
- [x] Remove `docroot/profiles/custom/collegiate`.
- [x] Remove `docroot/themes/custom/collegiate_theme`.
- [x] Remove `collegiate_theme` entry from package.json workspaces.
- [x] Remove `Uiowa\Blt\Plugin\Commands\ReplaceCommand::replaceSetup` (https://github.com/uiowa/uiowa/blob/fc7fe6af7c60024677380f0da72e9b74257c6fac/blt/src/Blt/Plugin/Commands/ReplaceCommands.php#L74) to remove looping through profiles.
- [x] Compare collegiate.info.yml with composer.json to see if any dependencies can be removed.
- [ ] Archive https://github.com/uiowa/hds
- [x] Remove profile argument from `uiowa:multisite:create`.
- [x] Remove uiowa.profiles custom configuration from blt.yml. Move requester config up for new and existing sites.
- [x] Set cm.core.install_from_config to true in blt.yml.
- [x] Remove default profile configuration from site creation process.
- [x] Update all existing site blt.yml files to remove profile specific config.
- [x] Move sitenow.settings.php to global.settings.php and remove profile specific include chain.
- [x] #1248 
- [x] #1247

richardbporter:
I added some additional cleanup steps. This would greatly reduce the complexity of our setup. If sites need to use a different profile or export all configuration, that would still be possible.

PR #1587, opened 2020-05-12
Change default install profile to sitenow so it runs in CI.

richardbporter:
Currently, the standard profile is being installed which isn't testing configuration integrity. This just changes the default profile to sitenow so that it is used during CI.


richardbporter:
I need to change a few more things around before this will work...

blt/blt.yml
```diff
@@ -54,6 +54,12 @@ command-hooks:
   post-deploy-build:
     dir: ${repo.root}
     command: './vendor/bin/blt uiowa:post:build'
+cm:
+  core:
+    install_from_config: true
+    dirs:
+      sync:
+        path: profiles/custom/sitenow/config/sync
```
richardbporter, line 62: This should be removed after #1477.

richardbporter:
This is a quick fix with #1477 being the long term goal. The default site might need to be re-installed when deploying this. I can handle that.

pyrello (approved):
Not tested, but reviewed the output from CI, which is what this should affect.

PR #1589, opened 2020-05-12
Simplify env splits and prevent indexing of all drupal.uiowa.edu domains.

### LINKED ISSUE #1300: Internal production domains are indexed by search engines.
Because our robotstxt split only denies indexing for dev/test, sites that are being "staged" in production are indexed. We should stage sites in stage or find a way to dynamically override the configuration based on the request URL. That should be possible based on how we are handling other config overrides like GA, AC, etc.

Example:
https://github.com/uiowa/uiowa/blob/8ebdb2594a472bb7f9d2f390557181fc0e386246/docroot/profiles/custom/sitenow/src/ConfigOverride/SimpleSitemapOverride.php#L38-L42

### LINKED ISSUE #1554: Simplify environment split blacklists
The environment split blacklists were calculated based on our old local environment where many extensions like memcache wouldn't work. This isn't necessary anymore. We should simplify these for easier management and faster BLT syncs.

richardbporter:
Resolves #1300

Resolves #1554, kinda. Purge complains without a Varnish instance running locally as does Acquia Purge. I don't think we want to set that up in DrupalVM right now, though it is possible.

### Changes
- Move memcache, robotstxt to default config enabled in all envs
- Remove unused config from env splits
- Add additional purger log channels to dev/test/prod splits
  - Potentially resolve an issue with purger config cim/cex discrepenancy.
- ~~Split robotstxt for uiowa.edu and deny all temporarily~~
  - Now handled dynamically for all sites.
- Hopefully fix acquia_connector.settings discrepancy which apparently was not resolved with the last module update.

### Testing
- Sync a site via `blt sync`.
- Ensure no config changes via `blt config:status`
- Optional: Manually send SPI data on a dev site and ensure no config changes.
- Ensure robots.txt is deny all locally and on dev
  - https://uiowa.dev.drupal.uiowa.edu/robots.txt


richardbporter:
Had to do this to split robotstxt for the homepage so that we can move it to the prod env. Also included that here.

richardbporter:
I briefly tried to get Varnish working locally so we could leave the purge modules enabled, except acquia_purge, but I was getting an error running `vagrant provision`.

richardbporter:
This seems to be working in dev if you compare the sitenow demo site config status:
`drush @sitenowdemo.prod cst`
`drush @sitenowdemo.dev cst`

richardbporter:
I had to run `drush @mysite.env p:queue-add url https://mysite.com/robots.txt` for some reason to purge that file after changing domains.

joewhitsitt (approved):
Tested a sync of multiple sites locally. icsa.uiowa.edu broke, but I wonder if that is because of the current state of the site.

Locally I get a `User-agent: * Disallow: /`

Did not test the optional step. Not sure how to do that.

richardbporter:
What broke with icsa?

richardbporter:
Syncing icsa/icsa_temp worked from prod to dev.

Issue #1553, opened 2020-05-07
As a content editor/webmaster, I can create different categories of people content


pyrello:
- [x] Add person type selection field
- [x] Add additional fields for students and faculty
- [x] Add logic to conditionally show fields based on person type selection.

pyrello:
> Add person type selection field

_Should this field be a static list? Or should it be a taxonomical list that can be augmented with additional types of people as the need arises?_

The problem with using a taxonomical list is that the taxonomy items would be content, we would have to provision default content for them, and then the conditional field logic would depend on content. Either the terms could be deleted or we would have to protect them from deletion somehow. At that point, it seems like it might have just been more logical to define a static list.

_Should we allow multiple selections? Are there any cases where we can imagine a person that fits into 2 categories?_

If we don't allow the selection of multiple categories, then if there is someone who should be listed in multiple categories, are we going to require they are entered twice?

On the other hand, making this a multiple selection widget, when that is almost certainly an edge case, is going to mean there is a lot of potential for misuse.

> Add additional fields for students and faculty

Additional faculty fields that have been requested for immuno.grad.uiowa.edu:
* Office Room
* Office Building
* Office Phone
* Lab Room
* Lab Building
* Lab Phone
* Department
* Faculty Type - Examples in immuno: Program, Resource
* Research Area

All of these, with the possible exception of Faculty Type, seem like they might be useful campus-wide.

Additional student fields that have been requested for immuno.grad.uiowa.edu:
* Lab Address
* Lab Phone
* __* Advisor__
* __* Program Start Date__
* Dissertation
* __* Former Student__ (Yes/No) – this indicates if they are alumni

I have highlighted the items in the student fields list that seem broadly relevant. The rest seem like they may be more specific to this site or program sites where lab work is a part of what the student is involved in.

> Add logic to conditionally show fields based on person type selection.

I'm thinking this can maybe be accomplished generically, where a field is named something like `field_person_faculty_department` and the form is altered to add #states for any fields with **`field_person_faculty`** to show when the type selection matches **`faculty`**.

joewhitsitt:
First off, just to pass on Todd's requests from the last week or so until a better issue is created to address the service wide additions: 
- pronouns
- additional roles/[positions]

I had the same debate with myself when we were considering adding the person type back in for our feature parity sprint. I was leaning towards a taxonomy with protected defaults that align with the APR person types along with prevent deletion functionality allowing site owners to add to the list and make it their own.

[...]

PR #1605, opened 2020-05-14
Feature: Extended person features and fields

pyrello:
Addresses #1553 

Adds a person_extended config split that adds additional fields for the person content type.

* Adds person type taxonomy and entity reference field.
* Adds logic in sitenow_people that will show/hide fields when the person type field is present. It looks for the presence of person type (lowercased) in the field name.
* Provides a Research Areas field which only shows when the faculty type is selected (`field_faculty_research_areas`).

I tested this against the default site to make sure that the logic related to the person type field gracefully degrades when the split isn't turned on.

This meets the minimum need for me to continue with migrating content on the gradinformatics.prod.drupal.uiowa.edu site and it does so in a way that doesn't require any other site to adopt the changes. If we decide that we want to adopt any of these fields into the base profile, then we can move them from the split into the profile.


config/default/config_split.config_split.person_extended.yml
```diff
@@ -0,0 +1,20 @@
+uuid: 41268a21-7cce-49ad-ab5c-192dcb64e597
+langcode: en
+status: true
```
joewhitsitt: Status should be `false` for default correct? I just synced gis.sites and saw the fields. person type label was visible but blank (no terms) and required.
Screen Shot 2020-05-21 at 8 28 03 AM

pyrello:
@joewhitsitt Currently, you have to manually add the person types. This is something that I expect to revisit.

joewhitsitt:
Now when i try to enable the split through the UI on a standard site, nothing happens in terms of vocab, fields. i am new to the feature split functionality

richardbporter:
I'm guessing you need to import config after activating the split: https://github.com/uiowa/uiowa/blob/master/config/features/README.md

pyrello:
Now I am running into the issue where, if there are no person types defined, then I have nothing to check against for the additional fields that are added, and they will all show up. I think this means that I need to have a common fragment in the field name that I can check against to hide all of them if there is no person type. I made the decision previously to name the "Research areas" field `field_faculty_research_areas` because adding `person` in it would have made it too long. I would be tempted to switch to a naming convention that doesn't include `field_` except that all the fields have already been named that way.

I'm open to suggestions about how to proceed.

joewhitsitt (approved):
I could see website coming back to the default config soon. I'd like to identify who is supposed to guide us on what a UofI person should look like in terms of fields. Thanks for making changes quickly

joewhitsitt:
Darn, didn't see your comment before i approved...

pyrello:
Adjusted the naming scheme for conditional fields to start with `person_pt_`. This also allows for me to have a more stringent check when checking for a match against a taxonomy term. Since the terms can be managed on site, there was the possibility of someone creating a match for an existing field by accident. This is unlikely now since they would have to include `pt_` in their term.

Also noting that this doesn't have proper handling for taxonomy terms that are multiple words, like "Student worker". I'll create a follow-up issue since this isn't getting deployed universally yet.

I tested against a site that had terms set for person type and a site that did not.

joewhitsitt (approved):

PR #1621, opened 2020-05-15
Handle individual multisite exceptions thrown during the update process.

### LINKED ISSUE #1575: Allow post deployment to proceed after failing config status check
Should post deploy fail on the "config is different" check? This shows as an error both locally and in production, but is really a warning that config needs to be fixed. The only way to resolve it is to merge another PR and in the meantime, no sites that come after will get database updates or config updates as part  that release. It seems like this could potentially be just a notification instead that allows the post deploy to move on to the next site.

richardbporter:
Resolves #1575 

With this change, it will no longer be necessary to re-deploy the same tag to production to "finish" deploying updates. All multisites will be processed and all errors logged in one job. 

Note that this does not change the behavior of `blt dsa`. That is a separate command that only runs locally.

### Testing
- Configure local multisites such that one fails before one that succeeds.
  - icsa.uiowa.edu
  - uiowa.edu
- Run `blt auda` which is what happens during a deployment.
- See that the first site fails but the process continues to second.
- See that the entire process is marked as a failure.


joewhitsitt:
Maybe I tested wrong, but I didn't see any output for `blt auda` and it only lasted a couple seconds before returning to a terminal prompt.

I configured blt multisites to try on demo.sitenow, icsa, uiowa.edu and hr

joewhitsitt:
I tried it again and it only got through the first site in my list.

richardbporter:
I was able to reproduce this. The `blt auda` command will only update sites that are installed. It skips sites that it cannot bootstrap. I ran it with the below list and verbose logging via `blt auda -vvv`:
```
multisites:
  #  - brand.uiowa.edu
  #  - careers.uiowa.edu
  #  - cogscilang.grad.uiowa.edu
  - demo.sitenow.uiowa.edu
  #  - dentistry.uiowa.edu
  #  - gis.sites.uiowa.edu
  - hr.uiowa.edu
  - icsa.uiowa.edu
  #  - ighn.international.uiowa.edu
  #  - iowasuperfund.uiowa.edu
  #  - policy.clas.uiowa.edu
  #  - sitenow.uiowa.edu
  - uiowa.edu
```

The sites it skipped have empty databases. So I guess in order to test this, a `blt dsa` needs to happen first.

pyrello (approved):
Works as stated. Nice work!

joewhitsitt:
i think that makes sense, i did a ume sql-drop originally and have only sync'd one site since

PR #1622, opened 2020-05-16
Feature: Featured image backgrounds and sidebar menus for pages

### LINKED ISSUE #1484: Change custom block type labels to hidden for uiowa_blocks
# Problem

The labels for the custom block type fields used on uiowa.edu have not be set to  `-Hidden-` yet. 

Screen Shot 2020-04-29 at 1 31 56 PM

# Proposed Solution

Set all custom block labels to `-Hidden-`.

### LINKED ISSUE #1617: Integrate menu-vertical style from uids with sidebar menus in uids_base
# Problem

The sidebar menu in SiteNow is is unstyled in uids_base. 

# Proposed Solution

Screen Shot 2020-05-14 at 4 07 25 PM

Add vertical menu styles from https://uiowa.github.io/uids/components/detail/main-menu--main-menu-secondary-vertical.html

bspeare:
Resolves #1484. 
Resolves #1617.
# Updates
* Changed base profile config for page entity view to use layout builder for managing layout.
* Changed base profile config for article entity view to use layout builder for managing layout.
* Add a layout template for page content with sidebar.
* Add a layout template for content over a background image.
* Add styles to accommodate new layouts.
* Update article and page template to remove featured image, breadcrumb, and title.
* Adds Layout Builder Direct Add module to base profile.

# How to test
Ideally, this should be checked against several sites. We want to make sure it doesn't break anything for uiowa.edu, while also seeing that the functionality is updated for brand.uiowa.edu, hawkeyemarchingband.uiowa.edu, and any other sites you care to check against.

- `blt frontend` 
- `blt dsa` for any sites you are checking.
- Check a page that does not have a featured image. Make sure that the title and breadcrumb show as normal.
- Add a featured image to the same page and save. Confirm that it now show the featured image with the breadcrumb and title on top of it and that the regular breadcrumb and title don't appear.
- Check an article that does not have a featured image for the same as above.
- Repeat adding a featured image to the article and confirm the same as above.
- Attach a page to a parent menu item, verify that sidebar menu appears on left-hand sidebar. 
- Set "hide page title" in Page Options and verify that page title is hidden when not using a featured image. 
- Set "Remove sidebar regions" in Page Options on a page that is attached to a parent item in the main menu.


joewhitsitt:
Is this affected by the re-organization of #1757?

bspeare:
@joewhitsitt Yes, the partial being used for the page node template will be moving in the restructure.

config/default/core.extension.yml
```diff
@@ -121,6 +121,7 @@ module:
   token: 0
   toolbar: 0
   tour: 0
+  twig_tweak: 0
```
pyrello: We will probably not want to enable this for all sites unless it is actually necessary with the updates we are making.

config/default/system.theme.yml
```diff
@@ -1,4 +1,4 @@
 admin: claro
-default: uiowa_bootstrap
+default: uids_base
```
pyrello: We will probably not want to make this update quite yet, unless there is no way to roll this out until we update all sites to use uids_base.

docroot/themes/custom/uids_base/templates/field/field.html.twig
```diff
@@ -48,53 +48,29 @@
   label_display == 'visually_hidden' ? 'visually-hidden',
 ] %}
 
-{% if entity_type == 'block_content' %}
```
pyrello, line 51: @bspeare Looking back, it looks like you added this section as part of the brand.uiowa.edu updates. I removed it and did a little checking and it seems like maybe it may not be necessary anymore. Can you review this and check to see that I haven't broken whatever you changed this to solve?

docroot/themes/custom/uids_base/templates/layouts/page--paragraph/layout--page--paragraph.html.twig
```diff
@@ -0,0 +1,62 @@
...
+
+{% set sidebar = content.sidebar|render %}
+
+{%
+  set classes = [
+  'layout',
+  'layout--page--paragraph',
+  'layout__container',
+  settings.extra_classes,
+  sidebar ? 'uids-menu-enabled' : 'uids-menu-disabled',
```
pyrello: @bspeare I think we should change this class name to something less specific to the menu, since it is actually controlling whether the sidebar appears. Maybe something like `.has-sidebar` or `.content--has-sidebar`.

docroot/themes/custom/uids_base/templates/field/field.html.twig
bspeare: Yes, it is still necessary until this happens: https://github.com/uiowa/uiowa/issues/1484

config/default/system.theme.yml
bspeare: Agreed!

pyrello:
I updated the config for custom blocks to have the labels be visually hidden. I'm not sure if that is the right setting in every case, since I believe that means that the labels will be announced to screen readers, which, for better or worse, will be a change in behavior. Previously they were not being rendered at all.

bspeare:
@pyrello It looks like the "Link Directly to Source" functionality of the Article is not working currently. 

Screen Shot 2020-06-16 at 4 38 16 PM

pyrello:
@bspeare You'll have to provide more info to replicate. I tested this out on the HMB site and the articles on the homepage link directly to the source article when one is present.

pyrello:
We don't necessarily have to figure this out in this PR, but the way this is currently set up, some of the features are only going to roll out to non-collegiate sites because the updates to the article and page entity view display are overridden by collegiate. Do we want to try to roll this out to collegiate? I think there is a chance that it will have no effect on uiowa.edu because every page has the template overridden for the most part. Not sure about sites like icsa.uiowa.edu or policy.clas.uiowa.edu.

@mark-bennett-uiowa Are you using the featured image on any pages on those sites? Or do you have some kind of header set up that might be in danger of getting changed on this update?

bspeare:
@pyrello Here are some screenshots of the missing functionality that I thought was related to the link checkbox.  I have styled the page to work without this functionality so it can wait for a future PR. 

The issue is that the Org name is linked up if there as a url in the Source Link:
Screen Shot 2020-06-16 at 8 27 57 PM

Here is how it displays in `uiowa_bootstrap`:
Screen Shot 2020-06-16 at 8 27 48 PM

I think this is ready for review.

cory-skeers:
Checked it on cnm.uiowa.edu, a large SiteNow site, and everything worked well and looked great.

Still needs review/testing on the primary sites listed above.

pyrello:
@cory-skeers Thanks!

richardbporter:
Synced default, hr, sitenow, uiowa. Tested default. Noticed some wonkiness on the menu link in the off-canvas nav:

image

Looks like this is also changing the base theme? Sorry if I'm out of the loop here.

 I can try to test additional sites later.

pyrello:
@richardbporter Thanks for testing! The theme update is going to be reverted. It is convenient for testing because you don't have to go through that step. This PR is part of the broader effort to get uids_base working with current SiteNow sites, so doesn't represent all the work in that regard.

pyrello (approved):
I have tested this on uiowa.edu, brand.uiowa.edu, the HMB site, icsa.uiowa.edu. That combined with all the other testing makes me feel pretty confident we have covered our bases for this to merge.

bspeare:
@richardbporter Thanks for pointing out the menu issues.  I have created an issue to address it: https://github.com/uiowa/uiowa/issues/1801.

pyrello (changes requested):
In looking at this on HMB site again, there is still some work that needs to be done to get the styling correct, probably in uiowa_bootstrap.

pyrello (approved):
The changes that have been applied to this PR fixed the issues that were of concern.

pyrello:
@joewhitsitt I'm pulling down HR to check it out against this PR.

pyrello:
Actually, looking at the config for HR, it has split config for both the page and article entity_view_display, so these updates won't have any impact here. I'm still going to look it over quick, but it shouldn't be affected.

PR #1629, opened 2020-05-18
Add status message JS to close dismissable alert.

### LINKED ISSUE #1620: Alert close button is not working in uids_base
# Problem 

You cannot close an alert in uids_base. 

Screen Shot 2020-05-15 at 12 15 42 PM

# Proposed Solution

Investigate fixing the JS so that alerts can be dismissed.

richardbporter:
Resolves #1620 

### Testing
- Check out branch.
- cache:rebuild
- save page or something
- dismiss alert


bspeare (approved):
Thanks for setting this up!

richardbporter:
Looking at this again, do we want to keep the data-* attributes? Is that from BS?

richardbporter:
We can revisit later if needed.

Issue #1588, opened 2020-05-12
Create unique id for card block component a11y


bspeare:
# Problem

<img width="1202" alt="Screen Shot 2020-05-12 at 11 06 41 AM" src="https://user-images.githubusercontent.com/1036433/81717899-bf81e400-9440-11ea-8a5d-35cd10661d70.png">

# Proposed Solution

Create unique id variable based on card title

PR #1631, opened 2020-05-18
Card Block Unique ID

joewhitsitt:
Relates to #1588


joewhitsitt:
Moved this into #1635

PR #1635, opened 2020-05-19
Update card template for linking with no title or link text

### LINKED ISSUE #1588: Create unique id for card block component a11y

### LINKED ISSUE #1615: Adjust paragraph card in uids_base to match fallback link strategy used in UIDS
# Problem

A link will only be added to the paragraph card if the `card_title` exists. 

# Proposed Solution

Adjust template to add a link to the card if `card_title` is empty: https://uiowa.github.io/uids/components/detail/card--default.html.

pyrello:
Resolves #1615
Resolves #1588

Testing instructions coming soon. Please feel free to review in the meantime.


bspeare:
I'm adding this as a reminder once the UIDS release comes through:

This `card__media` wrapper: https://github.com/uiowa/uiowa/blob/ee61d836a8549853ebe7d87060296955a7a086a3/docroot/themes/custom/uids_base/templates/paragraphs/card/paragraph--card--default.html.twig#L91-L99

Also needs to be added to the card block template: 

https://github.com/uiowa/uiowa/blob/feature_fix_card_link/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-card.html.twig#L38-L40

joewhitsitt:
Snuck my change in, but I am still not loving it. This will create an aria-describedby of the card title that points to a link which may not exist... Also, `clean_id` is based on `drupal_clean_id_identifier` (DEPRECATED) and so clean_id doesn't ensure the id is unique...

Since this just needs to be unique and not readable, can we just use something like `card-link-{{ id }}`

```
* - attributes: array of HTML attributes populated by modules, intended to
*   be added to the main container tag of this template.
*   - id: A valid HTML ID and guaranteed unique.
```

pyrello:
@joewhitsitt Thanks for adding that!

After reviewing this a bit more, here is my understanding about cards with links
1. In the case where both link text and a title are present, then the `aria-describedby` attribute is added to the `<a>` in the heading, which points to an ID that is then added to the pseudo button `<div>`.
2. Unless the above case is true, the `aria-describedby` attribute is not necessary - Correct me if I am wrong about this. This means that either there is just a title and no button, or no title and just a button, which in that case is an actual `<a>`.
3. Under the current way that paragraph cards are set up - where we are not able to rely on the link text as button text - the case outlined in (1) can never be true, so we don't need to include it on that template for now.
4. If we eventually migrate the current paragraph card link text into an aria label field and add default "Learn more" link text to every card with a link, then we'll need to revisit this.

joewhitsitt:
1. correct
2. correct, didn't want to duplicate the heading include just to omit the heading_aria but i imagine you know how to better this.
3. you _could_ rely on the card link text and call out content editors on it, but yes, paragraphs support is not needed now.
4. that sounds like a good option to consider if or when that happens.

pyrello:
@joewhitsitt I did try to use the `attributes.id` described in the templates docblock, but it doesn't seem to exist. I went with the title, if set. This isn't ideal, but workable for now.

docroot/themes/custom/uids_base/templates/paragraphs/card/paragraph--card--default.html.twig
```diff
@@ -88,7 +89,13 @@
       {% block card_img %}
         {% if not paragraph.field_card_image.isEmpty() %}
           <div class="card__media">
-            {{ card_image }}
+            {% if link_element == 'image' %}
+            <a href="{{ card_link_url }}" aria-describedby="desc-a-card">
```
joewhitsitt: desc-a-card is still here

docroot/themes/custom/uids_base/templates/uids/card.html.twig
```diff
@@ -1,32 +1,68 @@
...
+  {{ title_suffix }}
   {% block card_img %}
     {% if card_image is not empty %}
       {# @todo Add image template #}
       <div class="card__media">
-        <img class="card__img" src="{{ card_image }}" alt="Alt Text">
+        {% if link_element == 'image' %}
+        <a href="{{ card_link_url }}">
+          {% endif %}
+          <img class="card__img" src="{{ card_image }}" alt="Alt Text">
```
joewhitsitt: alt text is hardcoded here

docroot/themes/custom/uids_base/templates/paragraphs/card/paragraph--card--default.html.twig
pyrello: I removed this, since, as previously discussed, the paragraph card won't ever meet the criteria for adding the `aria-describedby` attribute.

docroot/themes/custom/uids_base/templates/uids/card.html.twig
pyrello: I changed this so that it just prints the `card_image` value in this template, since that is what happens in all our instance template and it allowed me to remove the `card_img` block from those templates.

bspeare (approved):
Looks good! I tested it on uiowa.local and brand.local and tried out a number of link/nolink combinations.

PR #1651, opened 2020-05-20
Route internal domains without ID to default site.

richardbporter:
Drupal will go though a subdomain stripping process that will bootstrap a different site if the URL does not have a directory alias. If a multisite is attempted to be accessed via an internal domain that does not exist, it could bootstrap another site. Ex. foo.prod.drupal.uiowa.edu on uiowa01 could bootstrap uiowa.edu if foo.prod.drupal.uiowa.edu does not exist. This PR forces internal domains to route to the default site, stopping that stripping process high up the chain.

### Testing
- Run `blt sync` on the master branch. 
- Notice that the sql:sync steps says "You will destroy data in **uiowa_edu** and replace with..."
- Check out this branch.
- Run `blt sync` again and it should say "You will destroy data in **uiowa** and replace with..."
- Hit other sites like brand.local.drupal.uiowa.edu and ensure the correct site is bootstrapped.

This test also depends on not having a docroot/default/local.drush.yml file with a correct uri option set which I just instructed everyone to do.

### DIFF
--- a/docroot/sites/default/blt.yml
+++ b/docroot/sites/default/blt.yml
@@ -3,11 +3,11 @@ project:
     hostname: default.local.drupal.uiowa.edu
     protocol: https
   machine_name: default
-  human_name: uiowa
+  human_name: default
 drush:
   aliases:
     local: self
     remote: default.prod
 drupal:
   db:
-    database: uiowa
+    database: default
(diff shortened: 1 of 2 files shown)


richardbporter:
https://api.drupal.org/api/drupal/sites%21default%21default.settings.php/8.7.x

joewhitsitt (approved):

PR #1675, opened 2020-05-22
Limit webform upload file size and allowed extensions.

### LINKED ISSUE #1664: Confirm Webform Settings regarding file uploads
We are seeing some odd default extensions and file size defaults.

I thought we addressed this in https://github.com/uiowa/uiowa/issues/353

Wonder if there was a configuration regression when we update webform.

richardbporter:
Resolves #1664 

The max upload config set limits individual elements but the allowed extensions do not. This PR limits file uploads to 50 MB, reduces the allowed upload extensions to a limited set and removes access from webmasters to change that per element.

### Testing
- Check out this branch and `blt dsa`.
- Masquerade as a webmaster.
- Create or edit a webform with a file upload field.
- Try to increase file upload size and add additional extensions. (exe, dmg, etc.)

### DIFF
--- a/config/default/webform.settings.yml
+++ b/config/default/webform.settings.yml
@@ -133,13 +133,12 @@ element:
   default_description_display: ''
   default_more_title: More
   default_section_title_tag: h2
+  default_algolia_places_app_id: ''
+  default_algolia_places_api_key: ''
   default_empty_option: true
   default_empty_option_required: ''
   default_empty_option_optional: ''
-  default_algolia_places_app_id: ''
-  default_algolia_places_api_key: ''
   excluded_elements:
-    captcha: captcha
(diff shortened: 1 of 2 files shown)


config/default/webform.settings.yml
```diff
@@ -139,7 +139,6 @@ element:
   default_algolia_places_app_id: ''
   default_algolia_places_api_key: ''
   excluded_elements:
-    captcha: captcha
```
richardbporter, line 142: Not sure why this exported or what it means.

joewhitsitt: makes me think that captcha is now an allowed element.

joewhitsitt (approved):

PR #1700, opened 2020-05-27
Brand - SVG Fonts Missing Smart Quote Glyphs

joewhitsitt:
Several glyphs were gutted from the original roboto font files when the iowa-roboto versions were created.

Copy pasting the desired smart quotes for review. Kern pair values for these glyphs don't exist and I assume they will just take a default value.


joewhitsitt:
Closing in favor of #1711

PR #1707, opened 2020-05-29
Fixed styling of people featured content to display correctly in all browsers

dnkemper:
Fixed styling of people featured content to display correctly in all browsers

Figured out that for some reason, extending the 'card' functionality was creating some very wonky results and conflicting with the styling. Then because of that, the "flex-direction: column-reverse" resulted in the image overlapping the text. I tried everything but this was the only way I could get it to display correctly in all browsers.


joewhitsitt:
I am trying to review this but with IE testing and the diff of this PR, I am having troubles.

PR #1710, opened 2020-05-29
Setup uiowa02 and improve uiowa:multisite:create command.

richardbporter:
This PR sets up the uiowa02 application for deployment and use within our monorepo system. Additionally, there are two changes to the `uiowa:multisite:create` command:

1. The command tries to inform the user of what application a multisite should be created on by outputting number of databases, SSL SANs and a potential SAN match.
2. The `checkIfRemoteDatabaseExists` method was removed so that the command throws an exception if a database already exists when it tries to create it. I think this is the desired behavior.

### Testing
First, looking for feedback on the SAN search logic in https://github.com/uiowa/uiowa/blob/8d5c418018ae589202ade7c21fa763bc4648d2f3/blt/src/Blt/Plugin/Commands/MultisiteCommands.php#L466-L481

- Check out this branch.
- Run some URLs through the command, e.g. `blt uiowa:multisite:create foo.research.uiowa.edu --no-db --no-commit`.
- Verify that the application information is correct specifically, the SAN match field.


richardbporter:
@briand44 This is the PR I mentioned earlier.

joewhitsitt (approved):
Can't give valuable feedback on the SAN logic, but it picked up my test case:

Screen Shot 2020-06-03 at 3 10 06 PM

PR #1711, opened 2020-05-29
Brand - Attempt to fix custom svg fonts

### LINKED ISSUE #1706: Brand - Loss of key glyph/kern pair data from custom font files
## Problem

Over the course of testing the lockup generator, we have had several different font files.
The original roboto files had glyphs for smart quotes, along with hundreds of other glyphs that are not in the current files. As recent as Mar 5, we had the original roboto files out there for testing. I think we currently have the Mar 11 version out there which had a slight kern tweak from the Mar 3 version.

The last known iowa-roboto versions to have all of the glyphs that were exported from Glyphr Studio was Feb 21. Now that isn’t saying the kerning was correct in the Feb 21st version, but it appears Glyphr Studio exported all the glyphs that were in the original roboto files.

joewhitsitt:
Hopefully resolves #1706 

Need to get this to DEV for testing/review.


joewhitsitt:
This was approved by Sondra

PR #1681, opened 2020-05-26
Feature: UIDS updates

pyrello:
This update follows a massive update to UIDS, so every path that references that library has changed. This means that testing this requires checking as many frontend pieces as possible on sites that are using uids_base. If you need more information to get started, please ping me.

# How to test

`blt frontend` 

**Test on Home Page (https://uiowa.local.drupal.uiowa.edu)**

Test Home Page Video https://github.com/uiowa/uiowa/issues/1680
- Check "Reduce Motion" in your Accessibility settings and refresh the home page. 
Screen Shot 2020-05-29 at 3 38 32 PM

**Test home page scroll changes for (https://github.com/uiowa/uiowa/issues/1665)**
- Home page header should disappear after scrolling down the height of the header bar. 

**Test all pages on uiowa.local.drupal.uiowa.edu**
- Click through and make sure that all of the components and background images are loading correctly including the top toggle menu and secondary page menu. 
- Test form elements on https://uiowa.local.drupal.uiowa.edu/feedback

**Test on Brand Site (https://brand.local.drupal.uiowa.edu/)**

- Test lockup generator
- Test mobile/desktop header variations in (https://brand.local.drupal.uiowa.edu/admin/appearance/settings/uids_base)
- Test mobile/desktop appearance of site after parent organization checkbox is set on (https://brand.local.drupal.uiowa.edu/admin/config/system/site-information)

PR #1712, opened 2020-05-29
Include stylesheets for WYSIWYG ckeditor

### LINKED ISSUE #1662: UIDS_Base WYSIWYG Styles
## Problem
WYSIWYG text areas are currently unstyled and don't offer a "preview" of uids_base rendering.

## Proposed Solution
Load the theme styles similarly to how uiowa_bootstrap did.

cory-skeers:
Resolves #1662 and #1714

Current SiteNow theming styles includes Lead and multiple Alert levels--these are available in the editor, but do not render within the editor. Additional styles and classes should also be available.

## Test
SiteNow: Edit a text body, check that WYSIWYG properly renders styling options, including allowable HTML, series of WYSIWYG editor buttons, and style/format dropdowns.

Collegiate: Edit a text body, check that appropriate uids styles are available, and appear correctly.


joewhitsitt:
Not sure if these are the stylesheets to use. I think we want to capture WYSIWYG related styles in a separate aggregated file and reference that. We need to create a list of the styles associated with the wysiwyg buttons and allowed html... This turned into a bit of a rabbit hole since there isn't a "master" css file.

I created a separate issue that can be part of this PR: https://github.com/uiowa/uiowa/issues/1714

joewhitsitt:
I think this will need to be updated after #1681 is merged in as paths have changed

joewhitsitt:
Tested this a bit. Not sure if it is pulling the same alert styles as rendered. Also I think some typography is missing.
Screen Shot 2020-06-04 at 4 22 53 PM
Screen Shot 2020-06-04 at 4 22 38 PM

cory-skeers:
For the alerts, alerts.scss has a few adjustments for pages that aren't using layout builder--these adjustments aren't pulled in at the wysiwyg editor-level, so the non-adjusted styling is being rendered in the editor.

For the typography, looks like the wysiwyg editor is overriding the body styling for fonts, and needs to either be prevented from doing so, or ordering of how the styling gets pulled in adjusted.

Working on it...

cory-skeers:
Removed the non-layout-builder adjustments to alerts as no longer needed--UIDS alerts theming is now rendered on the page as well as in the editor.

body{font-family:...} in ckeditor-iframe.css in core ckeditor module is the culprit for the typography issues for styling that does not indicate specific font-family to override. Needs to be targeted.

cory-skeers:
Added the Roboto sans-serif font-family to the ckeditor css through js addition. Takes care of typography mismatches when a font-family isn't explicitly declared in a class. There's still a small difference in typography of button text

cory-skeers:
Will need slight adjustments following #1756

joewhitsitt:
Because of #1756 / #1757, marking as blocked for now

joewhitsitt (approved):
Nice progress. I think this can go for now. Had an issue with font-weight related to the .lead class. Not sure how that can be resolved without making sure all font variants are loaded (heavy).

Issue #1680, opened 2020-05-26
A11y adjustments for autoplay video loop


joewhitsitt:
# Proposed Solution

- [x]  Set `prefers-reduced-motion` in JavaScript. 
- [x]  Set a cookie to remember whether pause was pressed? https://github.com/uiowa/uiowa/pull/1719
- [x]  Add aria label to play/pause button https://github.com/uiowa/uids/pull/228
- [x]  Move play/pause button top the bottom left of the video so that it can be paused https://github.com/uiowa/uids/pull/225

prefers-reduced-motion (https://web.dev/prefers-reduced-motion/)

https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#audiovideo_elements

https://support.mozilla.org/en-US/kb/block-autoplay

bspeare:
# Proposed Solution

1.  Set `prefers-reduced-motion` in JavaScript. 
2.  Set a cookie to remember whether pause was pressed?
3.  Anything else?

pyrello:
To test this, there is an option in Google Chrome console to emulate this setting: https://developers.google.com/web/updates/2019/10/devtools#userpreferences

PR #1719, opened 2020-06-02
Hero Video Cookie

joewhitsitt:
Relates to #1680 

## To Test

- Go to https://uiowa.local.drupal.uiowa.edu/ and try pausing the video. Reload or revisit the page after navigating somewhere else. Video should stay paused and button should be showing the play icon. Click the play button to resume video. Go somewhere and come back. Video should be playing.

- Check the cookie via Chrome's Dev Tools > Application > Cookies (refresh)

- Make sure Safari doesn't pull any funny business...


richardbporter:
I think jquery.cookie is still available in D8: https://www.drupal.org/node/3104677 It might make sense to use that or the D9 equivalent instead of rolling our own.

joewhitsitt:
"Doh!" Nice catch. Assumed there wouldn't be anything helpful from Drupal for JS. I will give it a go.

richardbporter:
Tested on Safari, FF and Chrome on OSX. Seems to work as intended. I think this is a nice feature.

Did not test IE because I haven't set up Browser Stack yet. Does the video play on mobile? Do we need to push this to dev to test mobile?

One thing about `document.getElementById(...` - as far as I know, we have a hard dependency on jQuery with how BigPipe delivers content. We actually _should_ be using jQuery.once. So I think its fine to use jQuery selectors for the time being, if desired.

joewhitsitt:
I do have the .once() in there but I am not a javascript ninja so I am probably not being efficient by additionally using the elementbyid. If you know the quick change, feel free to adjust.

richardbporter:
Yep, not saying anything is wrong. Just confirming.

I think using jQuery selectors (`$('#video-container')`) or not is just preference.

richardbporter:
I would merge this into dev to let other folks take a look.

joewhitsitt:
I will be doing demo in local, but this should be available in DEV for testing.

richardbporter:
30 days seems low - how do we feel about upping it to 90 or 180?

joewhitsitt:
I think it depends on what all we are trying to accomplish with this functionality. The first bit of feedback related to this a while back seemed like session-only related. User just wanted it to remember paused while they were navigating away and back from the homepage. Next came the hardware acceleration and accessibility concerns. As the video loop is swapped out for new, do we want people to experience it. If the idea is to remember the preference as long as possible, why stop at 180?

pyrello:
The way this is currently written, the cookie serves as a catch-all for any videos of this type on the site. If I'm understanding the code correctly, pausing the video on the homepage means that you have effectively paused any hero video on the site.

I'm also wondering about naming the js file with `hero-block` in it. The component is no longer called `hero-block`. Also, should this be in uids?

mjoneill:
> If the idea is to remember the preference as long as possible, why stop at 180?

I'd be fine setting it to never expire.  Philosophically, this puts the maximal amount of control in the visitor's hands.  I'm not sure I want to be in the space of saying, "It's ok for them to make this decision, but we're going to override it in 30 days."

^^ Please challenge my thinking here.  

I think we could do some interesting stuff in the future about making it clearer in the UX what is happening and what the consequence is when a visitor clicks pause/play.  I suspect in the future we'll want more feedback or scent of information than just pause and play...but I don't think perfection needs to be the enemy of this.

richardbporter:
I'm fine with it never expiring.

joewhitsitt:
hero-block was problem a result of when i started the PR and when the re-organization got merged in. I will update that. I think the cookie is path specific so it would pause on ALL hero videos within a page (hopefully there is just one). Haven't tested this use-case.

I remember hearing someone to say to put this in uids_base.

joewhitsitt:
- If no date is given, it becomes a session only cookie so I set the date to 99 years per @mjoneill request.
  - Looks like browsers are starting to set their own limits: https://www.cookiestatus.com/
- Minor cleanup, use existing variable.
- Renamed file to remove "block"
- Attempt to use documented path config but I don't think it is working. Added a video to /admissions and it was the same cookie. Otherwise it looks like it is subdomain specific cookie so you will have to pause on each site you encounter it. https://github.com/carhartl/jquery-cookie#path

richardbporter (approved):
I'll go out on a limb and approve this. I think its a nice feature for users that don't want want to mess with OS config. I'll probably use when developing locally. If I leave the front page for 20 minutes or so, my machine heats up.

The 99 year thing is weird but doesn't sound possible to create indefinite cookie.

joewhitsitt:
From stand-up: "Good for now"

PR #1733, opened 2020-06-03
fix image center align problems

GaryRidgway:
fixes images that are not center aligned properly.


joewhitsitt (approved):
better than broken!

PR #1734, opened 2020-06-03
Added update hook to invalidate twig cache for theme updates.

pyrello:


richardbporter (approved):
No fatal errors locally after syncing.

PR #1751, opened 2020-06-05
Improve SSL coverage validation in UMC.

richardbporter:
The previous logic was misleading as it was only a potential match. This reports whether or not the domain is covered. If not, it logs an error but still allows for site creation. We could take it further and look for "related SANs" but right now that is a manual process.

### Testing
Run URLs through `blt umc` and see if it the output is as expected. 

Ex. 
`blt umc foo.lab.uiowa.edu --no-db --no-commit` should log coverage on every app.
`blt umc foo.clas.uiowa.edu --no-db --no-commit` should log coverage on uiowa01.
`blt umc foo.hr.uiowa.edu --no-db --no-commit` should log an error. This would require manual discovery to determine where to create site. It would go on uiowa after regenerating the cert to cover *.hr.uiowa.edu.


richardbporter:
Ping.

PR #1753, opened 2020-06-08
Initialize emfisis.sites.uiowa.edu multisite on uiowa02

dnkemper:
Initialize emfisis.sites.uiowa.edu multisite on uiowa02


briand44 (changes requested):
I would first follow up with the customer. They said they don't want to abandon their existing url (https://emfisis.physics.uiowa.edu) and were hoping to have it forward to the new url they requested. It may be worth checking if they would prefer to continue using their current url for the new site. They may not be aware that we could provision it out using an internal production url.

PR #1754, opened 2020-06-08
Initialize emfisis.physics.uiowa.edu multisite on uiowa02

dnkemper:
Initialize emfisis.physics.uiowa.edu multisite on uiowa02

They decided to create internal domain to migrate content then switch once done migrating


briand44:
We probably want to put this on uiowa01 since that is where tracers.physics.uiowa.edu exists. Did the work Rick did with the validation in the UMC command catch this? I see there is still an PR out there, maybe that will address this. It would be nice to get @richardbporter thoughts. If it does go on uiowa01 then we will want to update the cert and add *.physics.uiowa.edu and remove tracers.physics.uiowa.edu.

dnkemper:
@briand44 I mean, third time's the charm? Haha. That makes sense. Should I redo it on uiowa01?

dnkemper:
@briand44 also, if we move forward with this, would I be able to change the cert? Or is that something you/Rick/Joe do?

richardbporter:
I changed #1751 to tell the user if the domain is covered or not. Not where to put it. Some conversation would need to happen around that. So in this scenario, it would output:
```
$ blt umc emfisis.physics.uiowa.edu --no-db --no-commit
[Acquia\Blt\Robo\Tasks\GitTask] Running git commands...
[Acquia\Blt\Robo\Tasks\GitTask] git rev-parse --abbrev-ref HEAD
Note: Multisites should be grouped on applications by domain since SSL certificates are limited to ~100 SANs. Otherwise, the application with the least amount of databases should be used.
+-------------+-----+------+--------------+
| Application | DBs | SANs | SSL Coverage |
+-------------+-----+------+--------------+
| uiowa       | 150 | 62   |              |
| uiowa01     | 89  | 31   |              |
| uiowa02     | 7   | 8    |              |
+-------------+-----+------+--------------+
[error]  No SSL coverage found on any application for emfisis.physics.uiowa.edu. Be sure to install new SSL certificate before updating DNS. 
 Which cloud application should be used? 
  [0] uiowa
  [1] uiowa01
  [2] uiowa02
 > 
```

Should we throw an error there to prevent creating the site until it is covered?

dnkemper:
@richardbporter @briand44 going to close this and redo it on uiowa01

briand44:
> @briand44 also, if we move forward with this, would I be able to change the cert? Or is that something you/Rick/Joe do?

As of now, not everyone has access to cert manager so you would have to just create an issue and someone with access would have to do this (Rick, Joe and I have access on our team).

briand44:
>Should we throw an error there to prevent creating the site until it is covered?

I guess my initial thought is no but I am open to discussions on this. If the site is going to use an internal domain to stage content then we probably shouldn't hold up the provision waiting for the SSL cert.

PR #1755, opened 2020-06-08
I1611 migrate footer menus

### LINKED ISSUE #1611: Migrate old footer menus to new footer menus
## Problem
There are two sets of footer menus. SiteNow uiowa_bootstrap uses footer and footer-second while SiteNow uids_base uses footer-primary, footer-secondary, footer-tertiary.

## Proposed solution
Update hook to move footer to footer-primary and footer-second to footer-secondary. The deployment of such an update hook will be pretty intensive and slow things down.

If this is done before uids_base adoption then, footer-primary and footer-secondary needs to be wired up to uiowa_bootstrap.

## Related ToDo
- [ ] #1763

cory-skeers:
Moves SiteNow footer links to match UIDS footer menus. Resolves #1611.


cory-skeers:
Consolidates footer menus and updates config to (hopefully) allow use with basic site settings config, and both uiowa_bootstrap and uids_base themes for SiteNow sites.

## SiteNow Testing
Pull the branch and run update hook; footer menus should still be present and functional in the uiowa_bootstrap theme. Changing themes to uids_base should retain these menus.

In admin/config/system/site-information, language and links for primary/secondary footer menus should function (again in both themes).

* Options such as 'enabled' should be retained; a disabled menu link should still be disabled.
* Sites in uiowa_bootstrap shouldn't have menu items in footer-primary and footer-secondary, but if links exist, footer and footer-second links should only add to existing links

joewhitsitt:
I haven't had a chance to test this yet, but are you planning to include the "related todo items" in this PR or should we create separate issues?

cory-skeers:
I put the related todo items into a separate issue. Mostly any followup could be handled separately from this PR, assuming the update hook works as intended and/or depending on the timing of its merge and the rollout of the uids theme.

joewhitsitt:
I tested on Brand which has already adopted uids_base with new footer links but still had old links in the old menus. That seemed to work.

HR worked after I updated some split config.

gis.sites (no footer links) ~~worked~~ didn't error as well.

So, if I understand this correctly, this PR could potentially go before uids_base switch as long as we do the other todos like permissions.

cory-skeers:
Updated references, and this PR now resolves #1763.  

In addition to moving footers to the new footer-primary and footer-secondary: 
- webmaster permissions are updated to administer these
  - updated for default, hr, and ighn.international
- basic page menu settings allow for adding to these menus
- menu and site settings tours are updated (links and label)

cory-skeers:
@joewhitsitt @briand44 Circling back to this from the earlier discussion.

In addition to footer-tertiary not being wired up to site settings, tours, etc., users (non-admin) still won't have access permissions for the footer-tertiary menu from this PR. Should be able to move forward with this without the case of adding something to a menu which isn't rendered.

joewhitsitt:
I noticed that after check-in. I was planning to review again next week. We should plan when this gets merged in as the update hooks on all SiteNow v2 sites will take a while.

joewhitsitt (approved):

joewhitsitt:
Getting this ready again. Got lost in a Slack thread: https://iowaweb.slack.com/archives/C8SM82ZFZ/p1592584867049900?thread_ts=1592584657.048600&cid=C8SM82ZFZ

Morning deployment with maintenance mode?

richardbporter:
I reordered the update hook to 8021 as 8020 had already run in main.

PR #1764, opened 2020-06-10
Feature: Block content templates in uids_base

### LINKED ISSUE #1740: Custom blocks created through block interface do not use inline block templates
# Problem 

Custom blocks created through block interface do not use inline block templates

For example, add a custom CTA block to the pre-footer block region in `uids_base`.

# Proposed Solution

Investigate way to map custom blocks placed through the block interface to use the inline block templates: https://www.drupal.org/forum/support/theme-development/2016-05-12/theming-custom-block-types.

bspeare:
This resolves #1740. The issue is that a block placed outside of the layout builder doesn't pick up our block templates where we have implemented UIDS components. This PR adds a module that causes block content to get its own preprocess function and template.

# How to test
- `composer install`
- Add Card to a block region like `pre-footer`. 
- Inspect the HTML of the card and verify that it is picking up the `block-content--uiowa-card.html.twig` template.
- Verify that markup on uiowa.local.drupal.uiowa.edu for blocks placed by Layout Builder is not broken.


pyrello:
I'm not sure how to go about testing whether this has a significant performance impact or not. Every inline block now has 2 templates applied to it and goes through a second preprocess function. This process only seems to happen when the cache is empty, so maybe it doesn't matter very much.

bspeare:
I tested this locally with uiowa.local.drupal.uiowa.edu and hawkeyemarchingband.local.drupal.uiowa.edu and I was able to add custom CTA and Card blocks to the `pre-footer` region and they were themed with the correct block template.

pyrello:
I have tested this on both uiowa.edu and brand.uiowa.edu. The main impact right now is that it *doesn't* impact inline blocks on uiowa.edu and the footer contact info block is rendered in the new template on both uiowa.edu and brand.uiowa.edu.

richardbporter:
Is this the same project? https://www.drupal.org/project/block_content_template If so, we'll want to require it from the drupal repo instead of GitHub, e.g. `composer require drupal/block_content_template`.

pyrello:
@richardbporter It is similar, but I could not get that version to work. I had to make changes to the module to get it to set up the templates correctly.

richardbporter:
Yes, sorry I forgot.

Looks like the original repo was archived and may no longer be maintained. Could/should this functionality just be put in uids_base?

pyrello:
I had intended to eventually open issues with the drupal.org module maintainer to make these changes so that we could use the contrib module instead.

cory-skeers (approved):
Tested with uiowa.edu, and works as intended.

As for performance, I don't think it should have a significant impact. but not sure of a way to get an accurate test of it outside of deploying it in the wild and doing a pre-/post-change comparison.

Issue #1471, opened 2020-04-27
Remove Adminimal Theme


joewhitsitt:
## Problem
Once Claro is adopted, no sites will be using the theme so there is no reason to keep it in the codebase.

Blocked until #1391 has been released.

- [x] uninstall/remove config
- [x] remove composer dependency

PR #1772, opened 2020-06-10
Remove Adminimal Step 1

joewhitsitt:
Relates to #1471

Follow-up task to resolve, remove from composer.

Quick note. the masquerade block is adminimal specific. there is a separate one for claro.


pyrello (approved):
Reviewed not tested.

PR #1790, opened 2020-06-16
Update uiowa_blocks banner and card excerpt text formats

### LINKED ISSUE #1340: Change Text (plain, long) in uiowa_blocks to Text (formatted, long)
# Problem 

A few of the fields used in the uiowa_blocks are set to Text (plain, long) when they need to be Text (formatted, long). 

**Card**
https://github.com/uiowa/uiowa/blob/master/docroot/profiles/custom/sitenow/config/sync/field.field.block_content.uiowa_card.field_uiowa_card_excerpt.yml 

**Banner**
https://github.com/uiowa/uiowa/blob/master/docroot/profiles/custom/sitenow/config/sync/field.storage.block_content.field_uiowa_banner_excerpt.yml

# Proposed Solution

- Create new fields that are Text (formatted, long)
- Create an update hook to migrate data from old Text (plain, long)

cory-skeers:
Initial update hook to resolve #1340.

Stores data in the field and revision tables, deletes the old field storage and recreates with the corrected type, restores the data/revisions, and re-sets view and form display weights in order to re-enable the fields.

Needs general testing, and checks that no other settings in related configs (such as view and form displays) get lost in the process.

Also, uiowa_banner_excerpt and uiowa_card_excerpt were identified as needing to be updated, but there may be more fields as well.


joewhitsitt:
Haven't tested yet, but this should not be in uiowa_edu_core as it affects other sites. Probably sitenow profile. Is that correct @bspeare or @pyrello ? I think #1340 was mislabeled.

bspeare:
Yes, I would think that it should be in the sitenow profile.

cory-skeers:
Thanks @joewhitsitt and @bspeare. Got it moved over.

joewhitsitt:
I tested this locally, didn't dig much, but I noticed the text was missing from https://uiowa.edu/academics "Create your path" banner.

cory-skeers:
If there wasn't a database error, it's probably related to updating the view display. In the initial field update, the field becomes disabled on the form and on the display because the weights are empty. The weights _should_ get pulled and re-set after update, enabling the field display.

I'll take a look.

Looks like the updates are sticking for the card excerpts, but not the banners.

cory-skeers:
Display is updating correctly, and the text is correctly placed in the database. Editing the page after, and saving a newly written excerpt, replaces what was inserted in the database as it should, tied to the same entity_id, etc.

Bug search continues...

cory-skeers:
uiowa banner excerpt revisions table needed updating to "block_content_r__a8c3187e9f".

Tested on uiowa.edu and igec.uiowa.edu (which does not currently have content in blocks) and fields appear to update correctly, and retain content.

In addition to testing, need to consider adding sanitation/validation--plain text field being inserted directly into the database field and rendered as html. I _believe_ this shouldn't be an issue, as the field is taken from the db (so no field/insertion errors should result), and the filtered_html format is  inserted directly, so no errant html snippets should be able to sneak in.

pyrello:
I tested this against uiowa.edu locally and I did not see any issues.

@mark-bennett-uiowa You may want to test this against icsa and/or policy.clas.

mark-bennett-uiowa:
Tested locally on policy.clas.uiowa.edu and saw no errors or issues. Thanks! Approved.

mark-bennett-uiowa (approved):
It appears there are no issues with this PR, approving for merge.

cory-skeers:
Thanks for the previous testing @mark-bennett-uiowa and @pyrello .

There was an issue with how Drupal shortens the revision table names using a partial hash, so redid part of it to retrieve field tables and revision tables dynamically to ensure they're correct.

Should work the same way for testing--sync a site, run the update hook, check that the uiowa_banner_excerpt and uiowa_card_excerpt fields are formatted text, and if the site had values in these fields prior, those fields should maintain their content.

PR #1792, opened 2020-06-16
I1723 migrate text format

### LINKED ISSUE #1723: SiteNow Migrate - paragraph text filter not applying
## Problem
Paragraph text fields aren't recognizing the 'format' attribute and defaulting to NULL when they are created as part of the Basic page migration (d7_page).
Screen Shot 2020-05-29 at 12 32 49 PM

## Proposed Solution
Root out the bug causing the format attribute to not be properly applied in order to fix on migrate.

Alternative would be a post-migrate event to programmatically update body text formats in all paragraphs, either directly or indirectly with an additional save to trigger the automatic filter update (the latter may not work if the wysiwyg editor isn't rendered).

## Workaround
Once the editor is opened, it fills the value to the proper filtered_html and gets saved. Currently, needs to be done for all pages.

cory-skeers:
Resolves #1723. 

Includes some code cleanup to use dependency injections, avoid using t() and \Drupal calls within classes.


joewhitsitt (approved):
reviewed, not tested

PR #1798, opened 2020-06-17
Unshift uiowa.edu to top of multisite iteration so it runs first.

richardbporter:
On deployment - this does not affect `blt dsa`. You can set that order yourself in blt.local.yml.

### Testing
Run `blt auda` before and after this PR. See uiowa.edu run first and only once on this branch.


joewhitsitt (approved):
playing favorites are we? slippery slope...

worked as described

richardbporter:
Born out of technical requirements due to traffic.

PR #1799, opened 2020-06-17
Update drupal/core from 8.8.7 to 8.8.8.

richardbporter:


pyrello (approved):
This applied cleanly except for an issue that seems to be unrelated and I reported in slack. I have not tested this thoroughly other than running `blt ds --site=uiowa.edu` and visiting the site locally to see that nothing appears to be broken.

PR #1818, opened 2020-06-22
Updated policy.clas.uiowa.edu permissions for absent book node rights.

mark-bennett-uiowa:
Updated permissions on the policy.clas.uiowa.edu site in order to fill in missing book node permissions. Limited scope of policy site, only updated three files in the site split.


pyrello (approved):
Reviewed not tested.

PR #1806, opened 2020-06-18
Change default theme to uids_base.

richardbporter:
Resolves #1660 
Resolves #1749
Resolves #1817 

### Notes
- ~~uiowa_bootstrap cannot be uninstalled yet since it is a dependency of other config entities.~~
- ~~Do we need seven installed? It is currently the maintenance theme.~~

### Testing
- The usual `git pull`, `composer install`, `blt frontend`.
- Run `blt dsa` for sites you are familiar with. Test everything!?
- Comment here with issues found.
- Check sites against this list: #1901 

## Tested:

- [x] virtualdance.studio.uiowa.edu - @joewhitsitt 
- [x] coronavirus.uiowa.edu @pyrello 
- [x] protostudios.uiowa.edu @cory-skeers 
- [x] hawkeyemarchingband.uiowa.edu @pyrello 
- [x] cnm.uiowa.edu @cory-skeers 
- [x] diversity.uiowa.edu @joewhitsitt 
- [x] itaccessibility.uiowa.edu @joewhitsitt
- [x] iti.uiowa.edu @joewhitsitt 
- [x] sitenow.uiowa.edu @richardbporter 
- [x] planning-and-development.fo.uiowa.edu @joewhitsitt 

## Other potential sites to test.
- ap-purchasing.fo.uiowa.edu
- bands.uiowa.edu
- businsvc.fo.uiowa.edu
- covidresponse.research.uiowa.edu
- dasblatt.sites.uiowa.edu
- honors.uiowa.edu
- iti.uiowa.edu
- miles.lab.uiowa.edu
- pac.org.uiowa.edu
- seru.uiowa.edu
- uibio.org.uiowa.edu
- vanallenprobes.centerforconferences.uiowa.edu

PR #1819, opened 2020-06-22
Copy quick links to top links uiowa_bootstrap -> uids_base

### LINKED ISSUE #1817: Migrate uiowa_bar quick links
Relates to #1660 

As part of the move to uids_base and from uiowa_bar to uiowa_search, look into programmatically migrating links currently in quick links, possibly to the Top Links menu.

cory-skeers:
Resolves #1817 

If quick links are present (set in /admin/config/system/uiowa-bar in uiowa_bootrap using uiowa_bar), they are added to the Top Links menu, which roughly translates to the same link positioning/usage after transitioning to uids_base.

## Testing
Run the update, switch to uids_base. If quick links had been included on the site, they should be included in the top bar with the same title and link. If no links were included, or if they were incomplete (i.e. no link or title), they should not be brought over.

If there are links already present in the Top Links menu, quick links should be appended to the end of the menu links.

## Notes
Update hook does not check for broken links, or links that could be duplicated by being moved to the Top Links menu. Assumptions are that if they were broken, they will remain broken for general maintenance work to fix, and that the sites which may have links appearing in both quick links and top links is limited enough, and the possibility low enough, that this does not need to be accounted for.

Depending on order of updates, an additional check to run only on sites using uiowa_bootstrap, or something similar could be added. It also leaves the links in the uiowa_bar settings, so this could run prior to uids_base rollout, with the understanding that there could be a data loss for changes made between this and uids_base.


joewhitsitt:
worked as expected. i am resistant to approve this PR because I think it should be merged and rolled out as part of #1806

cory-skeers:
Thanks for the review, @joewhitsitt -- I'd agree it can be merged and rolled out as part of #1806. Since this update isn't doing anything with config, there shouldn't be any update time concerns like with #1755.

joewhitsitt:
Alright, can you attempt the merge into #1806 modify the description to resolve the issue automagically, and close this PR? I know the update hooks are conflicting so let me know if you need help.

cory-skeers:
Yep! I'll get on it shortly, and let you know if I run into trouble.

cory-skeers:
This has been rolled into #1806.

PR #1832, opened 2020-06-24
change import format and set allowed formats 

### LINKED ISSUE #1831: Set import to minimal and set default to minimal for new banner/card excerpts
## Problem
Follow-up to https://github.com/uiowa/uiowa/pull/1790 we decided to import to minimal instead of filtered_html to avoid mis-use/display problems with rendering rich markup in these components.

## Proposed Solution
Update the import and set format default

joewhitsitt:
Resolves #1831

As a follow up to https://github.com/uiowa/uiowa/pull/1790. All previous test steps apply, but the main thing is that content is imported as minimal instead of filtered_html and the allowed formats for card and banner excerpts are restricted to minimal and plain text.

This also needs to be changed before the next release. No pressure...

### DIFF
--- a/config/default/field.field.block_content.uiowa_banner.field_uiowa_banner_excerpt.yml
+++ b/config/default/field.field.block_content.uiowa_banner.field_uiowa_banner_excerpt.yml
@@ -6,7 +6,14 @@ dependencies:
     - block_content.type.uiowa_banner
     - field.storage.block_content.field_uiowa_banner_excerpt
   module:
+    - allowed_formats
     - text
+third_party_settings:
+  allowed_formats:
+    minimal: minimal
+    plain_text: plain_text
+    filtered_html: '0'
+    full_html: '0'
 _core:
   default_config_hash: R12wgVOfQanvzxQM2QVSOwZE86dXdnY1K6gGkV-7KlE
 id: block_content.uiowa_banner.field_uiowa_banner_excerpt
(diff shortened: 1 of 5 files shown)


bspeare:
Pulling this down to test now.

bspeare (approved):
I tested this locally on uiowa.local.drupal.uiowa.edu and I didn't notice any issues or regressions with the banners or cards.   I did create an issue for future work to make these components work with the new text formats: https://github.com/uiowa/uids/issues/255.  Thanks!

joewhitsitt:
@bspeare is this a uiowa/uids#255 for release? editors could potentially start creating markup that isn't styled

bspeare:
@joewhitsitt I could create a kitchen sink fix and add it to this PR until it can be addressed in UIDS.  What do you think?

joewhitsitt:
That would be great because we with #1790 we can't create a release until this change goes out with it.

bspeare:
@joewhitsitt sounds good.  I'll add it to this PR later this evening.

bspeare:
@joewhitsitt I added some fixes for `banner--reversed` as well as an adjustment to the banner template changing the `banner__summary` to a `<div>` instead of a `<p>` tag.   I believe this fix is all that is necessary for now.

pyrello:
Looks like the spacing is a little tighter:
Updated:
Screen Shot 2020-06-24 at 9 22 03 PM
Original:
Screen Shot 2020-06-24 at 9 21 45 PM

I don't know that that matters.

bspeare:
I'm not too worried about that spacing.  As soon as the block text is updated the spacing will return.  I can update the banners on uiowa.edu manually after this is deployed.

bspeare:
@joewhitsitt I'm done making changes to this unless you see anything else that needs to be addressed.

PR #1846, opened 2020-06-25
Restrict top links

### LINKED ISSUE #1822: Restrict non-homepage sites to two rendered top links
## Problem

Top links are now controlled by a separate menu that has no restrictions. The brand team decided the max of two links still applies however except for the homepage.

## Proposed solution

Add logic to only print two menu links for non-homepage sites in the template.

Add menu guidelines explaining the logic and point to documentation where it is reinforced.

joewhitsitt:
Resolves #1822 

Restricts top links to 2 on all sites except for the homepage where the limit is 3.

## Test

- pull/cr to register code and template
- sync a couple sites (including uiowa.edu) and confirm update hooks run. uiowa_edu_core should run after the sitenow update on uiowa.edu
- turn on uids_base if not already
- on those sites confirm that only two top links show for all sites except the homepage (3) by creating a few more top links.
- go to the top links menu admin form and confirm a message is indicating how many links are displayed.
- switch the theme off uids_base and see that no message is displayed on the top links form.

### DIFF
--- a/config/default/uids_base.settings.yml
+++ b/config/default/uids_base.settings.yml
@@ -13,6 +13,7 @@ header:
   type: inline
   nav_style: toggle
   sticky: 0
+  top_links_limit: 2
 layout:
   container: page__container
 branding:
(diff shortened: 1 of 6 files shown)


cory-skeers (approved):
Tested on uiowa.edu and igec.uiowa.edu/protostudios.uiowa.edu with the quicklinks->toplinks migrate, and everything worked well.

Issue #1550, opened 2020-05-07
Program/Degree/Certificate block component for departmental sites


mark-bennett-uiowa:
My goal is to have a reusable component for adding program/degree/certificate to departmental sites under Undergraduate and Graduate Program pages. Majors, minors and certificates will be entered on these pages using this component.

<img width="1364" alt="Screen Shot 2020-05-07 at 1 42 51 PM" src="https://user-images.githubusercontent.com/3067017/81333283-f7b4ab80-9069-11ea-9088-13aa6bea148b.png">

As the above image shows, there are three sections of this component with separation borders that contain the following information:
**Top:**

- The title of the program/degree/certificate.
- A brief description of the program/degree/certificate.

**Bottom Left:**

- A heading to this section.
- An unordered list of various features of this program/degree/certificate.

**Bottom Right:**

- A heading to this section.
- A text area.
- A button with link attached.

mark-bennett-uiowa:
This is to be a template for each program on the site. And these will only be used on the Undergraduate Programs and Graduate Programs pages. They will be stacked on top of each other in a “list” of sorts, one for each Major, one for each Minor, and one for each Certificate. So on the UG Programs page, it will be B.S. in X, B.S. in Y, B.A. in Z, Minor in X, Minor in Y, Certificate in Z, with each one having its own block like this. Its like a splash screen for degrees/certificates, that will then allow them to go to an interior major/minor/certificate specific page or directly to the General Catalog.

pyrello:
Just adding some notes here:
@mark-bennett-uiowa and I talked about the possibility of using the Layout builder library module (https://www.drupal.org/project/layout_library), which allows admins to create multiple layout templates for a given entity/bundle. A content editor can select the layout type and it is applied to that node.

The way in which this module fails to be a good solution for the problem is that there is no way to use the selected layout to override the node's layout. If you attempt to modify the node's layout, you'll get a generated layout based on the default layout for the node type. This is documented in this issue: https://www.drupal.org/project/layout_library/issues/3082434

[...]

PR #1855, opened 2020-06-26
Added custom block type layout for uiowa_program block.

mark-bennett-uiowa:
Placed this here for eventual testing, not to be merged yet.

# To test
* `blt ds`
* `drush @default.local uli admin/config/development/configuration/config-split`
* Turn on the CLAS Program Block split.
* `drush @default.local cim`
* You should now be able to add the program block via LB.


bspeare:
@mark-bennett-uiowa do you have a screenshot of what this looks like with your site content?

mark-bennett-uiowa:
This is the structure:
Screen Shot 2020-06-30 at 12 04 09 PM
This is what it looks like (totally unformatted and styled):
Screen Shot 2020-06-30 at 11 58 34 AM
Here is the layout builder frame (3 images):
Screen Shot 2020-06-30 at 11 59 07 AM
Screen Shot 2020-06-30 at 11 59 25 AM
Screen Shot 2020-06-30 at 11 59 38 AM

mark-bennett-uiowa:
program_block_design

This is the plan with the location and styling. Ned to change subtitle to excerpt title in block config. @bspeare does this look like something that OSC could approve as far as meeting design standard?

mark-bennett-uiowa:
TODO: Rename field_uiowa_program_link to field_uiowa_program_cta_link and Link to CTA Link. Rename field_uiowa_program_subtitle to field_uiowa_program_excerpt_title and Subtitle to Excerpt Title.

DONE: This was updated in https://github.com/uiowa/uiowa/pull/1855/commits/a65626cffe03c2af408797ddfb7348449f375d8b.

pyrello:
@mark-bennett-uiowa My feeling on this is that I'm still not sure if it is the right approach. I'm not convinced that a custom block is the right way to create a *list* of items on a page. I'm also not sure if we want to create custom blocks in the default for a use case that you've indicated is fairly limited. As I indicated in my previous comment (https://github.com/uiowa/uiowa/issues/1550#issuecomment-647796403), I think that a better approach for this might be to use fragments and a view listing them. Either way, it might be better to put this into a config split for now so that you don't have to continue to wait to be able to use it and then we can continue to evaluate how to make this into a feature that would work for a broader set of uses.

pyrello:
@uiowa/developer I would love to get some feedback from other developers on this. My feeling is that this feature should be scoped to a split to be rolled out onto individual sites, since the program block is not something we would want on just any site. Also, I'm not sure that creating a custom block is the right approach for something that is ultimately a list. My gut feeling is that something like Fragments or even a content type would be more appropriate as you could then build a list of items in views and possibly expose it to LB via a block type as was done for articles and people.

All that being said, if I am in the minority here, I'll accept that.

mark-bennett-uiowa:
On the CLAS sites, these are going to be locked down from editing, as I would prefer that unscrupulous or inexperienced content editors not change these to meet their own personal views. Changing these require approval at the DEO or Admin level. This was built as an easy way to display department degrees or programs, without having to build out each individual piece separately. For CLAS, once these are entered, they will be left alone until a curriculum change at the department level.

Would this be more accepted if this was renamed to something like "Showcase Block" or something similar? Alex is already creating something similar on [https://coronavirus.uiowa.edu/returning-campus/showing-symptoms](https://coronavirus.uiowa.edu/returning-campus/showing-symptoms), [https://coronavirus.uiowa.edu/returning-campus/masks-distancing-and-health](https://coronavirus.uiowa.edu/returning-campus/masks-distancing-and-health), and [https://coronavirus.uiowa.edu/returning-campus/buildings](https://coronavirus.uiowa.edu/returning-campus/buildings). Im using this multiple times on every site basically.

joewhitsitt:
>My feeling is that this feature should be scoped to a split to be rolled out onto individual sites, since the program block is not something we would want on just any site.

Agreed. If "collegiate" is becoming the unrestricted layer split eventually, I see the need for an "academic" feature split to handle program, department, degree stuff and the "drupal" relationships between them. I can think of a couple units who would be interested in a share-cost sponsorship of such a split if the functionality can be agreed on.

I haven't been following the requirements of this PR, but why isn't this a content type? I would hate to re-type all this content per block if I need to show it on multiple places.

Additionally, if Alex is already building this on sites, why is this needed?

mark-bennett-uiowa:
Well, Alex isnt building this specifically, he is creating something in LB that is similar. Im building these out on every site right now, but using existing blocks to simulate what this would do. These are used in CLAS on the Undergraduate and Graduate pages to "showcase" an individual degree from a very high level. It doesnt get into the details, it just shows what the screenshots above are displaying. They then link to pages that contain all the details. At best they are fancier cards with a couple extra fields. Overall, I just want to avoid splits on splits on splits.

pyrello:
@mark-bennett-uiowa This page - https://coronavirus.uiowa.edu/returning-campus/showing-symptoms - is actually a great example of why I would not be in favor of a "showcase" block. On that page, you have 3 different versions of a two column layout. In addition to being different in terms of what is placed in each column, even the column content that is similar has slight variations: button under a list, no button under a list, button inside a text area with a background color, same thing with no button.

I'm not saying that there is no use for this type of complex "mini-layout", but we haven't really introduced a way to handle layouts inside of layouts yet, and until we do, I think we should just embrace the flexibility that the blocks and layouts we have provide for us to make all these variations.

mark-bennett-uiowa:
> @mark-bennett-uiowa This page - https://coronavirus.uiowa.edu/returning-campus/showing-symptoms - is actually a great example of why I would not be in favor of a "showcase" block. On that page, you have 3 different versions of a two column layout. In addition to being different in terms of what is placed in each column, even the column content that is similar has slight variations: button under a list, no button under a list, button inside a text area with a background color, same thing with no button.

@pyrello This is all the more reason to create a showcase block. If this is "grouped" content, keep in grouped. Simply add a link field into the "left" column, and then rename the fields in the two lower columns to be a more generic "left" vs "right" gives you this (I changed them to excerpt_title/excerpt_description and cta_title/cta_description/cta_link based on earlier discussion on them being too restrictive). The two lower columns have a title field (h3 vs the h2 of the main title/description in the parent section above it), a text area for whatever you want to put in there, and then the right column has a link field.

The only downfall is the borders that I styled into the program block. I would like them, where as others probably wont. Im happy to add the changes above, and then move this whole project to a CLAS split only if that would take this card/issue out of the project queue.

richardbporter:
I ran `blt drupal:install` locally to reproduce the error then exported default configuration with `drush config:export` to add that search API config change.

mark-bennett-uiowa:
@richardbporter Thanks so much Rick! May need to pick your brain on that fix process, just in case I need to utilize it later.

richardbporter:
Is this PR ready to go or still a draft?

pyrello:
@mark-bennett-uiowa Would you like some help putting this into a feature split?

mark-bennett-uiowa:
@pyrello Yes Sean that would be great. I feel I do need to update this though so that it can provide some of the styling from the cards though, such as the white and gray backgrounds, so that the fonts, borders, etc make and change like the cars do.

pyrello:
I did the initial work of scoping what had already been created in a feature split that can be turned on. I updated the OP with testing instructions that also cover how to turn it on.

pyrello:
@bspeare and @mark-bennett-uiowa I think that we should create a custom module (either `clas_custom` or `clas_program_block`) and put templates and styles there similar to `admissions_core` or `uiowa_alerts`. This helps with maintaining the theme, which already includes a lot of "extra" styles.

mark-bennett-uiowa:
@pyrello Im all for that too, whatever is easier. Just as a heads-up though; Sam in Nursing is interested in using these blocks for their degrees also, so can we come up with a name without clas in it?

Would this mean that it goes from a feature back to a module?

config/default/config_ignore.settings.yml
```diff
@@ -13,6 +13,7 @@ ignored_config_entities:
   - 'config_split.config_split.scholarship:status'
   - 'config_split.config_split.sitenow_migrate:status'
   - 'config_split.config_split.sitenow_v2:status'
+  - 'config_split.config_split.*:status'
```
richardbporter: I think this would ignore site split statuses which would be undesired. Maybe we should prefix feature splits with feature_ or something.

pyrello: Good call.

mark-bennett-uiowa: Is this causing the issue in the conflicting file however?

pyrello: I changed this and fixed the conflict.

pyrello:
It seems like maybe we can close this based on https://github.com/uiowa/uiowa/issues/1550#issuecomment-1276778349, @mark-bennett-uiowa?

pyrello:
Closing this based on previous comment and lack of response.

mark-bennett-uiowa:
Yes @pyrello it can be closed. Thanks for doing so.

PR #1856, opened 2020-06-26
Create media entity for brand lockup files

### LINKED ISSUE #1829: Migrate brand lockup image to managed files
## Problem

Use of brand lockup images on websites is being removed. Move to uids_base will include the discontinued use of these images in the footer, and SiteNow users will not have access to their lockup image files.

## Proposed solution

Migrate brand lockup image into the managed files.
Potentially update the file updated date to place the file at the end of the list (with the intention of if a user asks about the lockup image removal, can direct them to it.

cory-skeers:
Resolves #1829 

## Testing
Run update hook.
- Site which does not have a brand lockup image in the footer
  - Update hook should finish, no changes to site.
- Site with a brand lockup image in footer
  - Brand lockup image should still be present in footer (if in uiowa_bootstrap theme)
  - In Content > Media, the brand lockup entity should be available
    - When sorted by updated date, it should appear at the end of the list

### Note
If testing locally without syncing files, several `[error] File at...does not exist` will be thrown. Other than these error flags, update hook should run normally and media entity should still be created.


joewhitsitt:
pulling this down for testing.

joewhitsitt:
Is there an issue if we set the managed media file owner to user 1?

cory-skeers:
Shouldn't be a problem--setting to anon was just a carryover of mine from working on other sites.

Updated it to set the uid to 1.

joewhitsitt (approved):
Thanks for making the change

cory-skeers:
Added into #1806--should roll out with update to uids_base to avoid possible data loss if someone added a brand lockup after the update hook.

PR #1861, opened 2020-06-29
Add parent accordion id. Add accordion item heading id.

### LINKED ISSUE #1857: uids_base accordion ids in markup
Relates to #1775 

## Problem

- [x] uids-accordion__heading id missing from markup. used for aria-labeledby.
- [x] accordion-No1 id doesn't line up with data-parent value.

cory-skeers:
Resolves #1857

## Problems addressed
Accordions weren't receiving unique ids to be referenced by data-parent attributes on accordion items.
Accordion item headings weren't receiving unique ids to be referenced by aria-labelledby attributes on their expanded content.

## Test
Sync a paragraphs site. Change theme to uids_base. `blt frontend`.
Create/check at least two accordions. 
Check that they receive unique ids which match their items' data-parent attributes.
Check that each accordion item's heading has a unique heading id which matches its respective expanded content aria-labelledby attribute.

### DIFF
--- a/docroot/themes/custom/uids_base/templates/field/paragraphs/field--paragraph--field-accordion-item.html.twig
+++ b/docroot/themes/custom/uids_base/templates/field/paragraphs/field--paragraph--field-accordion-item.html.twig
@@ -38,11 +38,9 @@
  */
 #}
 
-{#
-{{ 'accordion-' ~ parent_id }}
-#}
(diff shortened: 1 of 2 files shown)


GaryRidgway (approved):
I have checked this and can confirm that it is working as stated.

PR #1864, opened 2020-06-29
makes slick dots smaller when there are too many for the width of the…

### LINKED ISSUE #1848: Carousel slide markers overflow
## Problem

Too many slides and not enough width cause overflow. Also, it doesn't seem centered at larger widths.

Screen Shot 2020-06-25 at 10 48 21 AM

## Proposed solution

Responsive sizing of markers

GaryRidgway:
…  container

This is made to fix #1848. 

to test: create a carousel with many slides, and make its container a small size (a third or a quarter should work).
Then, test whether the slick dots shrink down to fit the container, like so:

Screen Shot 2020-06-29 at 2 42 56 PM


joewhitsitt (approved):
much better. thanks!

Issue #1775, opened 2020-06-12
uids_base - SiteImprove A11y Fixes


joewhitsitt:
## Problem

SiteImprove is starting to scan for issues. Here are some of the issues from the results.

- [ ] https://github.com/uiowa/uiowa/issues/1320 - FontAwesome Icon Role
- [ ] https://github.com/uiowa/uiowa/issues/1119 - Reduce use of aria attributes that duplicate semantics of the HTML tag
- [x] https://github.com/uiowa/uids/issues/249 block iowa wordmark id is not unique (the same id is used for header and footer)
- [x] https://github.com/uiowa/uiowa/issues/1152 - underlined links where ever possible (_not a 1660 blocker_)
- [x] Increase footer link contrast next to normal text
- [x] dark gray bg with yellow links does not meet 4.5:1 wcag color contrast. (_might be part of #1728_)
- [x] #1875 role=search needs a unique aria label
- [x] https://github.com/uiowa/uids/issues/249 - unique wordmark id based on placement
- [x] https://github.com/uiowa/uiowa/issues/1853 desc-a-card is still lingering around on some cards
- [x] #1857 uids-accordion__heading id missing from markup. used for aria-labeledby.
- [x] #1857 accordion-No1 id doesn't line up with data-parent value
- [x] focus styles needed on slick carousel navigation
- [ ] #1898 hawk alert needs a landmark
- [ ] #1899 hawk alert needs `aria-atomic="true"` on role=alert element.
- [x] #1896 Horizontal `<nav>` element needs an aria-label: `<nav class="menu-horizontal nav--horizontal header--secondary horizontal">`
- [x] #1900 Nested heading issues
- [ ] #2072 
- [ ] #2073 
- [ ] #2074

richardbporter:
I can take a shot at the alerts tasks if nobody has started on those.

joewhitsitt:
Quick note about "hawk alert needs a landmark" there are two. One is controlled by our JS, the other is the drupal announce region. There is an issue about that.

I would suggest role=region for the JS one and an aria-label to make it unique

joewhitsitt:
Closing this as https://github.com/uiowa/uiowa/labels/a11y lists all a11y related issues and this story is no longer a sprint priority currently.

PR #1875, opened 2020-06-30
remove extra role search

GaryRidgway:
This is a solution for #1775 `role=search` Landmark being in the markup twice.

To test: use an accessibility tool to check that this role does not have any conflicting landmark problems.


joewhitsitt (approved):
This resolves the initial issue and from my testing does not introduce a new issue (e.g. absence of the landmark), so I think we should deploy and then see how it goes.

Issue #1114, opened 2020-02-27
Layout Builder Safari off-canvas dialog bug


bspeare:
As reported by @mjoneill, here is the Drupal issue and patch: https://www.drupal.org/project/drupal/issues/2958588.

joewhitsitt:
This was merged in.

PR #1882, opened 2020-07-01
fix for #1114

GaryRidgway:
this patch applies to #1114, running composer install should fix the bug.


bspeare (approved):
I tested this locally with https://uiowa.local.drupal.uiowa.edu and I can confirm that the svg layout icons appear in the editing tray.   Does this need to be tested anywhere else?

joewhitsitt (approved):
> Does this need to be tested anywhere else?

This patch seems straight-forward. Based on scoping I don't see where else this would need to be checked. We can easily remove if there is a regression

Issue #1879, opened 2020-07-01
Update footer contact info text area in uids_base footer to be minimal html format


bspeare:
# Problem

The Additional Information area for SiteNow sites in the `uids_base` footer is currently set to allow filtered html.    

# Proposed Solution

Change text format to minimal html and allow `<span>` tags for font icons.

joewhitsitt:
Footer Contact Info is a text area block. Do we want to restrict the text area block to just minimal? Otherwise just restricting block/1 to minimal is a bit more complicated.

pyrello:
We do not want to restrict text area block to minimal, since it is our catch-all block type that allows us to cover cases we can't cover with other components yet.

This would require an update hook instead of changing config.

joewhitsitt:
We added span/class functionality to minimal and decided it wasn't worth the squeeze in trying to regulate the textarea format for a single block.

Should we close this one for now and wait to see what kind of policy comes from brand regarding footers?

pyrello:
I concur.

PR #1887, opened 2020-07-01
add span with class functionality to minimal format

joewhitsitt:
Relates to #1879 

Makes it possible to add fontawesome icons and such to the minimal text areas.


cory-skeers (approved):
Works as expected 👍

PR #1888, opened 2020-07-02
Remove image background media type

### LINKED ISSUE #596: Consider migrating all image_background media types back to image.
## Problem
The original reasoning for using this media type has become unclear and has only created unnecessary complications.

Poor media file management/selection UX
Migration issues #379 

## Proposed solution
Migrate image_background entities back to image, use the sitenow_16_9 responsive image style, and update section background code/field.

cory-skeers:
Resolves #596 

## Problem
Image_background is available in SiteNow(P) (and as a side effect, collegiate sites as well)--used for section background images in paragraphs, but not used in layout builder.

## Resolution

- [x] Move existing image_background to standard image.
- [x] Remove image_background
- [x] Update paragraphs theming to apply proper style (ui_edgy_no_crop) using regular image for section backgrounds

## Testing
- Sync site. 
  - (If you haven't rsynced files, 'file does not exist' errors will be thrown. This is to be expected, and syncing or stage proxy after syncing should still work appropriately.)
- Check that image:background is no longer available as an option in media.
- If testing a SiteNow(P) site, check that any previous image:background media now appear as regular image media.
- Add a section background image using standard image media and check that it renders correctly, with the appropriate styling.
- Check that non-background images haven't changed, and are appropriately styled (dimensions appear normal, resolution appears normal)


pyrello:
@cory-skeers I think the build on this is failing because of the default content, which has references to the `image_background` media type: https://github.com/uiowa/uiowa/blob/i596-remove-image_background-mediatype/docroot/profiles/custom/sitenow/content/media/fc23ffc0-95e1-45ff-ab07-b92464068d66.json

I think this needs to be re-exported or possibly it could just be updated via find and replace.

richardbporter (approved):
Tested uiowa.edu but it only has one background image which is broken - probably not used.
Tested a fresh install on default. 👍 
Tested on brand. 👍

joewhitsitt:
The policy.clas site is affected by this change:

```
Synchronized configuration: delete field.field.media.image_background.field_media_image_1.
 [error]  The field_section_image entity reference field (entity_type: paragraph, bundle: section) no longer has any valid bundle it can reference. The field is not working correctly anymore and has to be adjusted. 
 [notice] Synchronized configuration: delete media.type.image_background.
```

PR #1847, opened 2020-06-25
Mirroring recent updates to UIDS

pyrello:
Resolves #1780 
Resolves #1804 

# How to test

- Verify that `blt frontend` can run without errors.
- Initial testing should be done against the following sites to ensure there are no visual or functional regressions:
  - [ ] uiowa.edu
  - [ ] brand.uiowa.edu
  - [ ] cogscilang.grad.uiowa.edu
  - [ ] icsa.uiowa.edu
  - [ ] policy.clas.uiowa.edu

## Site Header
- Verify that site header markup is similar to: https://uids.brand.uiowa.edu/components/detail/iowa-bar--site_name_below.html.
- Test at all breakpoints. 
- Test on BrowserStack ie11 if you have access. 
- Test IOWA bar stickiness (or lack thereof). Test the scroll functionality down the page, scroll up and test the menu and search options mid-scroll and verify that there are no regressions. 

## Site Header Toggle
- In a private browser, verify that new `iowa-bar.js` is being applied to `.o-canvas__drawer` to set top position based on scroll position in browser.

## Site Footer
- Test that footer and footer logo appear the same without visual regressions (https://github.com/uiowa/uiowa/issues/1811)  
- Test footer with parent organization and without. 

## In-page menus
Verify that the in-page menu on https://uiowa.local.drupal.uiowa.edu/student-life have no visual regressions. Check a mobile breakpoint.

## Button Styles 
- Verify that #1780 has been implemented.  

## Contact form
Check that there are no visual regression on https://uiowa.local.drupal.uiowa.edu/contact

## Visual Regression Fix
- Verify that extra spacing seen in the stat grid on https://uiowa.edu/ is gone (caused by https://github.com/uiowa/uiowa/issues/1810)

## Layout Builder Style Changes
- Check that the background pattern for the events header and the stats section of the home page are in place: https://uiowa.local.drupal.uiowa.edu/
- Inspect the HTML of those areas to confirm that the background is being applied using a `bg--*` class instead of a `uids-*` class.
- Check that the stats blocks have `element--flex-center` alignment class applied to them and that they look like they do in production.

PR #1902, opened 2020-07-06
Padding for section background color (layout builder)

### LINKED ISSUE #1688: Add default padding if background color is used in layout builder
# Problem

No top or bottom padding is added when color is applied to a section. 

Screen Shot 2020-05-22 at 11 25 49 AM

cory-skeers:
Resolves #1688 

# Problem

No top or bottom padding is added when color is applied to a section. 

Screen Shot 2020-05-22 at 11 25 49 AM

## Test
Sync, blt frontend.

- Check section w/background styles applied and see that there is appropriate padding above and below its elements.
- Sections w/out background styles should not be affected.
- If a section has a background style and is the last section on the page, the bottom default margin should be removed.


cory-skeers:
Needs to be updated following #1847, and could better target bg--* and bg-pattern--* patterns to account for future style additions.

bspeare:
Pulling this down for testing now.  As mentioned above, we will want to change these styles to be something like:

```
.layout__container {
  &[class*="bg-pattern--"],
  &[class*="bg--"] {
    padding-top: $desktop-width-gutter;
    padding-bottom: $desktop-width-gutter;
    margin-bottom: 0;
  }
}
```

docroot/themes/custom/uids_base/scss/global.scss
```diff
@@ -304,6 +305,10 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled
 
   .layout__container {
     margin-bottom: $desktop-width-gutter;
+
+    &div[class*="uids-component--"]:last-child {
```
bspeare: @cory-skeers A default margin is set on all the layout builder sections to create better design spacing by default.  I added this to remove the default bottom margin for the last layout container on each page if there is a background color/pattern.

docroot/themes/custom/uids_base/scss/global.scss
```diff
@@ -441,6 +446,32 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled
...
+    padding-top: $mobile-width-gutter;
+    padding-bottom: $mobile-width-gutter;
+
+    @include breakpoint(sm) {
+      padding-top: $desktop-width-gutter;
+      padding-bottom: $desktop-width-gutter;
+    }
+  }
+
+  &.section-padding__removed {
```
bspeare, line 435: I added a new `section-padding__removed` layout builder style to remove the default padding for sections with a background color.  There are two sections that will need this option on the home page, the stats section and the events section.

docroot/themes/custom/uids_base/scss/global.scss
```diff
@@ -441,6 +446,32 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled
   }
 }
 
+
+
+// Section w/ color background padding
+
+div[class*="uids-component--"] {
+  .lb__container {
```
bspeare, line 425: I moved this to target the`lb__container` because we were trying to add all padding to this container instead of the parent container.  Let me know if you see the parent container as being used for padding elsewhere.

bspeare:
To recap, the current layout builder margin/padding setup: 
- Each layout section has a default bottom margin to create better spacing between each section
- This margin can be removed through the "Remove default margins" option in the layout builder style overrides for each section. 
- The last layout builder section on each page with a background color applied will have its margin removed by default. 
- With this PR, If a section uses a background color it automatically has top and bottom padding applied to the section.  This padding can be removed through the layout builder style included in this PR with the "Section Padding Removed" option.

bspeare:
Also, a number of sections on uiowa.edu will need to have the "Section padding removed" class applied to them when this deploys.   Should we deploy a separate PR first with the new layout builder styles option so that we can set those before this padding css is implemented?

bspeare:
Added blocked label until https://github.com/uiowa/uiowa/pull/1917 is deployed to uiowa.edu.

bspeare:
I've gone through and set the sections on uiowa.edu with the new class.  This can be tested at any time.

richardbporter (approved):
Minimally tested with my limited LB knowledge. Padding looks better. 👍

PR #1903, opened 2020-07-07
Update SiteNow footer settings.

### LINKED ISSUE #1824: Remove Additional Info, Brand Lockup, and Slogan options from Site Settings
Site slogan, brand lockup, and additional info options are available in basic site settings of SiteNow sites. In uids_base, content from these fields is not used, and needs to be removed.

## Proposed solution

-  ~~Move content in Slogan to a suitable spot within uids_base (if necessary)~~
- [x] Remove Site Slogan from site settings options for uids_base theme, or add notice that it is deprecated/won't show
- [x] Move brand lockup image to media system (#1829)
- [x] Move Additional Info to uids_base footer (#1820)
- [x] Remove brand lockup and additional info options from site settings in uids_base theme (both can be removed by removing the uiowa/uiowa_footer module)
  - [x] ~~Move uiowa_footer.settings config, and update references~~ Remove uiowa_footer.settings
  - [ ] Replace login link addition outside of uiowa_footer _This should be resolved in uids_base_
- [x] Update tour info
- [ ] Update documentation

cory-skeers:
Resolves #1824 

Blocked by #1755 (footer migration--doesn't impact testing)
Blocked by #1806 (additional info migration, brand lockup media creation, uids_base rollout. This PR can be rolled into that one, or come after).
Blocked by HR and/or other sites on uiowa_bootstrap after #1806.

## Issue
Site slogan isn't used in uids_base; brand lockup and additional info fields are being deprecated. Adjustments to the Basic Site Settings to account for these, along with the removal of the uiowa_footer module.

## Test
Sync site and run update hook.
Check that uiowa_footer module is disabled.
In /admin/config/system/site-information, check that site slogan, footer variant, brand lockup, and additional info fields are no longer available.
Footer menu links should remain available.
Check that the Site Settings tour doesn't error/doesn't reference non-present fields (/admin/config/system/site-information?tour).

### Notes
`uiowa_footer` is referenced in the following, and they need to be checked:
- [x] docroot/sites/brand.uiowa.edu/modules/brand_core/brand_core.module
- [x] docroot/themes/custom/uids_base/uids_base.theme


docroot/profiles/custom/sitenow/sitenow.profile
```diff
@@ -97,25 +97,6 @@ function sitenow_query_administerusersbyrole_edit_access_alter(AlterableInterfac
   }
 }
 
-/**
```
joewhitsitt: Unless this is getting moved somewhere else I think we still need this

cory-skeers: Yeah, nope--definitely still need it, and it wasn't getting moved. Committed changes to an old version of sitenow.profile. Got it updated.

cory-skeers:
@joewhitsitt I've updated this to account for HR, if we were otherwise prepared to move forward with updating site settings and deprecating uiowa_footer.

And question in general: is there a preferred way to determine what site we're on/custom target exceptions in update hooks and modules? Base path would be more consistent than site title, that still felt like it wouldn't be quite right.

richardbporter:
>And question in general: is there a preferred way to determine what site we're on/custom target exceptions in update hooks and modules? Base path would be more consistent than site title, that still felt like it wouldn't be quite right.

The RequestStack service can get the host which returns the site directory via Drush. So:
```
$host = \Drupal::requestStack()->getCurrentRequest()->getHost();

  if ($host != 'hr.uiowa.edu') {
...
```

If its run through the UI (which it won't be during deployment) it returns the host accessed through the browser which could be different.

P.S. - I didn't test that snippet. :)

richardbporter:
Why is this blocked by #1806? It resolves #1915 and a PHP warning on the basic site settings form. ~~Also, since #1755 was merged, the site settings form is confusing since it still points people to the old menus. They could theoretically edit links there between now and Monday.~~

joewhitsitt:
>And question in general: is there a preferred way to determine what site we're on/custom target exceptions in update hooks and modules?

For theme related (not site specific), we have been using:

`$active_theme = \Drupal::service('theme.manager')->getActiveTheme();`

I think if this does what it needs to do and doesn't cripple HR, it can be rolled into 1806. I will try and test this soon.

richardbporter:
>Also, since #1755 was merged, the site settings form is confusing since it still points people to the old menus. They could theoretically edit links there between now and Monday.

Scratch that.

cory-skeers:
@richardbporter It takes away webmaster access to edit additional info, brand lockup, and site slogan. In uids_base, this isn't an issue, but if it rolls out before July 13, it'll just be taking away those features while they're still present on the site.

@joewhitsitt If it gets rolled into 1806, I think the active_theme would be a better way to handle those checks. There's one in the update_hook, and one in the form_hook_alter in sitenow.profile where pulling the host might not work.

cory-skeers:
I believe if this rolls with or after #1806, the login link is taken care of by
https://github.com/uiowa/uiowa/blob/18391a3a20cfb3305033574070c64e4112819239/docroot/themes/custom/uids_base/uids_base.theme#L132-L143
otherwise below needs to be accounted for somewhere:
https://github.com/uiowa/uiowa_footer/blob/3482228a3a49e6965fa0bee74c0c85fa55aac997/uiowa_footer.module#L228-L241
(current HR workaround isn't effected by this)

joewhitsitt:
Tested on a couple sites. No apparent issue other than what has been mentioned.

Removed the brand code that was placing the uiowa_bar and footer on backend pages (e.g. create lockup).

cory-skeers:
Updates have been added to #1806 as part of the full uids_base rollout.

PR #1909, opened 2020-07-07
Run git gc in post deploy build

richardbporter:
Run garbage collection before pushing build artifact as per Acquia Support recommendation. I tested by running `blt artifact:deploy` on the feature branch. I don't think any additional testing is needed.


pyrello (approved):
Reviewed, not tested.

Issue #1889, opened 2020-07-02
Investigate limiting available options in layout builder editing tray


bspeare:
# Problem

The layout builder tray offers too many options when adding a new block when you select the "More..." option. 

<img width="1663" alt="Screen Shot 2020-07-02 at 9 12 15 AM" src="https://user-images.githubusercontent.com/1036433/86369751-7d0ca480-bc44-11ea-9fa1-075054407b66.png">
<img width="323" alt="Screen Shot 2020-07-02 at 9 12 01 AM" src="https://user-images.githubusercontent.com/1036433/86369762-80079500-bc44-11ea-9a65-323af205cdfd.png">

# Proposed Solution

Investigate methods to limit available fields and blocks available in the tray.

pyrello:
You can set this in the content type manage display page by implementing block restrictions.

bspeare:
To set it on uiowa.edu, you would go to https://uiowa.edu/admin/structure/types/manage/page/display#edit-layout-layout-builder-restrictions-allowed-blocks.

Issue #1789, opened 2020-06-16
Create Articles/People functionality in layout builder using custom block types


bspeare:
# Problem 

Provide a way to add view of People, Articles.

# Proposed Solution

Recreate the paragraph bundles in v2 as custom block types in v3, keeping things to a list style only for now.

Create a separate issue offer more layout/style options to display articles, people and possibly events in the future.

pyrello:
So, it looks like both of the listed module address how to display views in the formats listed, but not how to go about making them available to attach in LB. I think just creating a view blocks will make them show up?

joewhitsitt:
You folks don't like my programmatic view renders?!

bspeare:
I do like them! I just added the items above to start the conversation.

bspeare:
@GaryRidgway  It looks like these are the current views available via layout builder:

<img width="372" alt="Screen Shot 2020-07-02 at 10 01 20 AM" src="https://user-images.githubusercontent.com/1036433/86375446-4dad6600-bc4b-11ea-82bd-c6ba94b9147b.png">


- Is this the correct list of views? 
- Do we need to add more or remove any?  
- Do we want to provide style variants of these views?  
- If so, what are the style variants? Grid, Masonry, List?
- Are the views currently styled if they are placed within a layout section?
- Are there layout sections where we would want to limit the placement of these views?

GaryRidgway:
are there any sites with layout builder that actually have people or articles for me to test with? the three I pulled don't have any.

In addition, are the "who's online" and "recent Content" views useful? I don't think i have seen people use those. recent content maybe if we can filter with it but definitely not "Who's online"

joewhitsitt:
Restricting blocks is in https://github.com/uiowa/uiowa/issues/1889

I am not sure we want to use these same views for articles/people unless we implement them via custom block types. These were crafted with paragraphs and view pages in mind.

pyrello:
Here are my notes from investigating this:

* I think that an approach similar to what was done with the paragraph implementation (spaghetti code) is likely our best bet at this point.
* I also looked at the [EVA](https://www.drupal.org/project/eva) module as a possible way to solve this. EVA allows you to create a view that can be attached to an entity type. However when attached view is added, it doesn't provide any options that get passed through to the block entity, so it doesn't really seem like it provides much advantage over just manually defining the block that gets attached. 
[...]

PR #1910, opened 2020-07-07
updating allowed blocks for #1889

GaryRidgway:
This is for #1889 and it is to limit the number of possible blocks a user is presented with when building a page. I do not think this list is definitive, but at least it is a start to what we could work towards. This does not account for [this](https://github.com/uiowa/uiowa/issues/1789#issuecomment-653075625) as my reasoning is that if paragraphs goes away, those might go away and then we might have to rethink our strategy? Either way, this should start the ball rolling.

To test:
1. blt dsa a layout builder site, composer install
2. check that you can only add a limited number of blocks
3. additionally, you can check at this location `/admin/structure/types/manage/page/display#edit-layout-layout-builder-restrictions-allowed-blocks` to see if there is configuration being set for every option.


pyrello (approved):
I re-added the fields that we are already using:
* Links
* Moderation control

We can revisit these later.

PR #1911, opened 2020-07-07
Removed option to use AJAX for IGHN person view

dnkemper:
Removed option to use AJAX for IGHN person view - was interfering with filter results


joewhitsitt (changes requested):
_If_ we disable ajax (i'd like to figure out why it isn't working), we should do this as a singe commit to sitenow-uids #1806

joewhitsitt:
This was fixed here: https://github.com/uiowa/uiowa/pull/1806/commits/3b5db3b3f3ccfe86532e9eb21314ad606a04cfc2

PR #1912, opened 2020-07-08
Feature: Section alignment updates in uids_base

### LINKED ISSUE #1457: Adjust layout builder flexbox styles so that components will stack by default
# Problem

Cards will not wrap if placed into the same layout column. 

Screen Shot 2020-04-22 at 4 02 37 PM

# Proposed Solution

Adjust container settings on layout for `flex-wrap`.

bspeare:
This PR:

- Resolves https://github.com/uiowa/uiowa/issues/1457 by adding `flex-wrap: wrap` to all layout builder columns and updates the card default width to `flex: 1 1 100%`.  
- Adds a layout builder style for section alignment called **Alignment: Start** in order to shift the column alignment to the top of each section.  By default content is distributed to fill the entire section so that card rows all appear at equal heights by default. 

# How to test

`blt frontend`
`blt ds --site=uiowa.edu` 

## How to test stacking issue

Using a site with layout builder enabled, try adding block components to columns within the two, three, and four column layout options.   
- Verify that they visually stack within the layout builder editor preview and remain stacked when the page is saved and does not look like https://github.com/uiowa/uiowa/issues/1457.    
- Stack two cards on top of each other
- Stack a text area with embedded image, a CTA
- Stack any other combinations and verify that they stack on top of each other

## How to test new alignment style

The new **Alignment: Start** style is meant to override the default alignment of each section. 

Example section with default alignment using a text area block on top of a cta block:

Screen Shot 2020-07-09 at 10 26 47 AM

Same section with **Alignment: Start**  applied to it:

Screen Shot 2020-07-09 at 10 26 56 AM


joewhitsitt:
pulling this down for testing

joewhitsitt (approved):
The stacking issue looks to be fixed. You almost lost me with the alignment stuff as it didn't occur to me that is what made up that image:

>using a text area block on top of a cta block

bspeare:
Thanks @joewhitsitt! Yes, I should have mentioned uploading an image there!

config/default/layout_builder_styles.style.section_alignment_start.yml
```diff
@@ -0,0 +1,10 @@
+uuid: 03cf5420-de43-4f89-9f93-8f37e819c0ae
+langcode: en
+status: true
+dependencies: {  }
+id: section_alignment_start
+label: 'Alignment: Start'
```
pyrello: Should this be 'Vertical alignment:'?

bspeare: I think that's a good suggestion since there could eventually be Horizontal alignment.   @joewhitsitt What's your take on the horizontal and vertical alignment options that you built into the paragraph sections?

joewhitsitt: Yeah, I would prepare for a horizontal in the future by naming this one Vertical. We prefixed "content" because people were confused when adding this setting on a section what it actually did.

We also used labels that more users identify with (e.g. top, bottom, middle, left, center, right) instead of flex terms. "none" was an after thought to achieve equal-heights.

Screen Shot 2020-07-10 at 8 54 23 AM

Moving the layouts world, all of our stuff was on a row though, not able to stack like you have done here, so I am not sure what would make the most sense to users.

bspeare: @joewhitsitt thanks for chiming in.   I changed the name to "Vertical alignment: Start" for now.

PR #1916, opened 2020-07-08
Temporarily update default footer config with new menus.

### LINKED ISSUE #1915: WSOD on newly created site
`Error: Call to a member function label() on null in template_preprocess_uiowa_footer() (line 274 of modules/uiowa/uiowa_footer/uiowa_footer.module).`

### Steps to reproduce locally on the default site (simulating site creation):
- `drush sql:drop`
- `blt drupal:install`
- Visit the front page.

I'm not sure if this belongs here or in the uiowa_footer repo. First reaction is we might be missing default config.

richardbporter:
Resolves #1915 and a warning on the site settings form.

@joewhitsitt and @cory-skeers are you ok with this short-lived change?

### DIFF
--- a/config/default/uiowa_footer.settings.yml
+++ b/config/default/uiowa_footer.settings.yml
@@ -4,8 +4,8 @@ lockup:
   alt: ''
   url: ''
 social_media_menu: social
-custom_menu: footer
-custom_menu_2: footer-second
+custom_menu: footer-primary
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Good fix. I thought this had been updated as part of #1755, but better late than never!

PR #1917, opened 2020-07-08
Feature: Layout builder style for removing default section padding

bspeare:
This needs to be deployed and configured on uiowa.edu before https://github.com/uiowa/uiowa/pull/1902 is deployed. 

# How to test

The only thing that needs to be verified is whether the class of `section-padding__removed` is applied to a layout builder section if this style is applied.  The weight of each style has been adjusted to make room for this new style.


pyrello:
I updated the labels for all the LB styles to remove "Section," "Block," and "Style", since the styles are all restricted to appropriate scopes where those labels should be unnecessary. Additionally, I changed the capitalization of the label from "Style Category: Specific Style Applied" to "Style category: Specific style applied". I think this is more consistent with the way other things in the system are labeled both in core and in our custom modules.

pyrello (approved):
Tested on homepage site. Nice work!

PR #1923, opened 2020-07-08
Image background hotfix

cory-skeers:
Hotfix for #1888 

## Problem
Resultant errors in policy.clas.uiowa.edu, icsa.uiowa.edu, and possibly others:
`Synchronized configuration: delete field.field.media.image_background.field_media_image_1.
 [error]  The field_section_image entity reference field (entity_type: paragraph, bundle: section) no longer has any valid bundle it can reference. The field is not working correctly anymore and has to be adjusted. 
 [notice] Synchronized configuration: delete media.type.image_background.`

A config update was being missed for sites which did not have any image_background media entities, resulting in config incompatibilities. This update should fix the oversight, and prevent any order-of-operations problems with the config sync.

## Testing
Sync site.
(If you haven't rsynced files, 'file does not exist' errors will be thrown. This is to be expected, and syncing or stage proxy after syncing should still work appropriately.)
Check that image:background is no longer available as an option in media.
If testing a SiteNow(P) site, check that any previous image:background media now appear as regular image media.
Add a section background image using standard image media and check that it renders correctly, with the appropriate styling.
Check that non-background images haven't changed, and are appropriately styled (dimensions appear normal, resolution appears normal)

### DIFF
--- a/config/policy.clas.uiowa.edu/core.entity_view_display.media.image.edgy.yml
+++ b/config/policy.clas.uiowa.edu/core.entity_view_display.media.image.edgy.yml
@@ -1,0 +1,37 @@
+uuid: 127e16b8-9193-4609-bc2e-40a38765d9e8
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.media.edgy
+    - field.field.media.image.field_media_image
+    - media.type.image
+    - responsive_image.styles.ui_edgy_no_crop
+  module:
+    - lazy
+    - responsive_image
+_core:
+  default_config_hash: HW52IsyizTd9psyOjhjteunTeM7jjttYXkwPB0YAbcU
+id: media.image.edgy
+targetEntityType: media
+bundle: image
+mode: edgy
+content:
+  field_media_image:
+    weight: 0
+    label: hidden
+    settings:
+      responsive_image_style: ui_edgy_no_crop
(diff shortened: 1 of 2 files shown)


pyrello:
I tested this and encountered no issues. However, I also didn't see that update hook 1822 ever ran, which seems wrong.
```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/uiowa/vendor/bin/drush @self updb --uri=icsa.uiowa.edu --no-interaction --ansi in /var/www/uiowa/docroot
 --------- ----------- --------------- --------------------------------------- 
  Module    Update ID   Type            Description                            
 --------- ----------- --------------- --------------------------------------- 
  sitenow   8021        hook_update_n   Move SiteNow footer menus to the UIDS  
                                        footer naming scheme.                  
 --------- ----------- --------------- --------------------------------------- 
```

pyrello:
Whoops. I'm dumb. I never actually pulled the feature branch down.

pyrello:
Seems to have gotten through all the sites without a problem, but it will cause a config difference with policy.clas, as was indicated in the previous issue.

pyrello:
I went ahead and ran the command to re-export the config for the policy.clas site split:
```
drush @claspolicy.local csex policy_clas_uiowa_edu
```

pyrello (approved):

Issue #1778, opened 2020-06-12
Layout Builder feature maturity


pyrello:
# Layout Builder feature maturity
- [x] #1994 Slider (carousel) component.
- [x] #1529 Layout Builder collections: Accordions, Tabs.
- [x] #1523 Provide a way to lock the header section.
- [x] #1789 View blocks for articles/people with a list display. 
- [x] #2071 Custom events block
- [x] #1925 Replace homepage default content with Layout Builder version.
 
# Issues
- [x] #1783 Unable to edit custom blocks on reverted LB node with custom blocks.
- [x] #1890 Granular block permissions
- [x] https://github.com/uiowa/uiowa/issues/1114 Safari icons in tray.
- [x] https://github.com/uiowa/uiowa/issues/596 - deprecate/remove image background media type - appears to be available to layout/collegiate media library
- [x] https://github.com/uiowa/uids/issues/245 Hero block centering bug
- [x] https://github.com/uiowa/uiowa/issues/1457 Flexbox stacking issue in preview mode.
- [x] https://github.com/uiowa/uiowa/pull/1912 Layout builder style to set container alignment.
- [x] https://github.com/uiowa/uiowa/issues/1889 Limit custom fields available in layout builder tray. 
- [x] #1922 Hide Site Improve widget when editing a layout builder page. 
- [x] display title/title on block types does nothing but is visible to webmasters?  (We are currently removing this title from display in twig template: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/field/field.html.twig.  Do we want to have it available for display?)
- [x] https://github.com/uiowa/uiowa/pull/1917 Layout builder style to remove default padding
- [x] https://github.com/uiowa/uiowa/issues/1688 Container padding when background color is used.
- [x] #960 Remove image button is not very usable in LB tray.
- [x] #1926 Add body field to Page content type.
- [x] #2038 Change component text areas to use `minimal_html` by default or limit toolbar buttons to reduce kitchen sink.  
- [x] https://github.com/uiowa/uiowa/issues/1957 Stat component regressions. 
- [x] https://github.com/uiowa/uiowa/issues/1962 One-col layout fix.
- [x] #1973 Update entity_usage to work with layouts/media library.
- [x] https://github.com/uiowa/uiowa/issues/1987 Card Horizontal formatting and permissions within one-col layout
- [x] Remove Spacer/Separator block or did that already happen in https://github.com/uiowa/uiowa/pull/1979?
[...]

PR #1927, opened 2020-07-09
I1922 hide si widget

### LINKED ISSUE #1922: Hide Site Improve widget when editing a layout builder page.
When editing a Layout builder page and pulling in blocks from a side menu, the Site improve widget is by default open and blocks some of your view. This should be collapsed by default when editing a layout page.

GaryRidgway:
This resolves #1922 and should collapse the Site Improve widget when editing the layout of a layout builder page.

To test:
1. Pull down branch and set up a Layout builder site.
2. Go to any layout builder page and edit its layout.
3. Check to make sure the SI widget is collapsed, and that a double click opens it.


richardbporter:
Found the offending PHPCS line by running `blt phpcs` locally.

One question I have here is if this SI overlay is needed period. I find it annoying on other pages too.

richardbporter:
Ah, I see what it does now. However, I don't see any per-page information when editing a layout. It doesn't seem to know what page it is like it does when just editing a node.

joewhitsitt:
So just to understand the problem, I went to the PROD homepage site. By default, yeah it sticks out, but you can click or drag it close and it remembers with a browser cookie it looks like, so I don't know if additional code is needed here.

Screen Shot 2020-07-17 at 12 18 34 PM

richardbporter:
@bspeare I think you added this to #1778 - any thoughts here? I guess I could go either way. This works as advertised.

bspeare:
I like the idea of having it collapsed because it helps reduce the visual clutter and does not require every editor to know that it will disappear if set on other pages.   If you guys think we should avoid the additional code, I can see it go that way too.  I just want to reduce the visual noise on the layout edit screen.

richardbporter (approved):

PR #1943, opened 2020-07-13
Color contrast fixes for card markup

joewhitsitt:
Fixes color contrast ratio issue on https://diversity.local.drupal.uiowa.edu/about/dei-across-campus where headers are used in card body and buttons have spans nested...

Our kitchen sink didn't account for these things https://default.dev.drupal.uiowa.edu/kitchen-sink

Checked accordion and it looks like the uids- prefix was removed from uids but not from the theme. Accounted for the same header and span markup

Should be able to see card/accordion (headers, buttons, etc.) markup on the bg-black sections.


richardbporter (approved):
Tested diversity. 👍

Issue #1900, opened 2020-07-06
Nested heading issues


joewhitsitt:
## Problem

There are a few situations where the programmatic render of content is causing improper heading combinations.

## Proposed Solution

Fix by creating specific templates or setting heading level variable that is passed to template.

- [x] Articles (articles, featured content paragraphs = h3, view page  = h2)
- [x] People (people, featured content paragraphs = h3, view page  = h2)
- [x] Basic Page (featured content paragraphs = h3)
- [x] Pagination should be switched to a div instead of a heading (classy view template) as we don't always know the heading content before.

joewhitsitt:
Seeing as the basic page teaser template is not using the card component, I have created https://github.com/uiowa/uiowa/issues/1955. Basic page is already an h3 where it is being called currently and is not an issue.

Issue #1878, opened 2020-07-01
Set heading style for footer links in uids_base


bspeare:
# Problem 

Footer link headings appear unstyled and in a flex column next to the footer links.

<img width="670" alt="Screen Shot 2020-07-01 at 9 54 14 AM" src="https://user-images.githubusercontent.com/1036433/86259601-d3fd7580-bb81-11ea-8757-7d581583fa06.png">

# Proposed Solution

Add a style to `footer.scss` component to set this style.

joewhitsitt:
@bspeare not sure if you have additional plans for this beyond color contrast which was fixed in #1947

bspeare:
@joewhitsitt, yes I need to set some sizing styles and flexbox adjustments for the headings.   Thanks for the reminder.

joewhitsitt:
Sorry, I mis-understood what this issue was about..

bspeare:
@joewhitsitt my description could have spelled it out better!

PR #1947, opened 2020-07-13
Site Improve A11y Fixes

### LINKED ISSUE #1896: Aria-labels on <nav> elements
## Problem

Horizontal <nav> element needs an aria-label. This may affect other menu blocks as well in uids_base.

`<nav class="menu-horizontal nav--horizontal header--secondary horizontal">`

joewhitsitt:
fixes #1896
relates to #1900 as it doesn't fix basic page or pagination.

relates to #1878 as it just ensures proper contrast colors for headings.

## To Test

- 1896 - With the site improve extension locally, view sites with different primary navigations and confirm that no errors related to navigation are thrown for "Non-distinguishable landmarks" (e.g. brand.local.drupal.uiowa.edu, uiowa.local.drupal.uiowa.edu)
- 1900 - On default.local.drupal.uiowa.edu after a `cr` confirm that the article teasers on the homepage have `<h3 class="card__title">` while the article teasers on /news have `<h2 class="card__title">` which provides the proper header nesting order.
- 1878 Visit a site like brand.local.drupal.uiowa.edu which has a header in their footer. It should now show up with proper contrast ratio.

### DIFF
--- a/docroot/modules/custom/sitenow_articles/sitenow_articles.module
+++ b/docroot/modules/custom/sitenow_articles/sitenow_articles.module
@@ -188,7 +188,7 @@ function sitenow_articles_preprocess_node(&$variables) {
           $route_name = $current_route->getRouteName();
           $variables['#cache']['contexts'][] = 'route.name';
           if (strpos($route_name, 'view.articles.') === 0) {
-            $variables['content_heading'] = "h2";
+            $variables['content_heading'] = 'h2';
           }
 
           if ($node->hasField('field_article_source_link_direct')) {
--- a/docroot/modules/custom/sitenow_articles/sitenow_articles.module
+++ b/docroot/modules/custom/sitenow_articles/sitenow_articles.module
@@ -186,6 +186,7 @@ function sitenow_articles_preprocess_node(&$variables) {
           $current_route = \Drupal::routeMatch();
           $route_name = $current_route->getRouteName();
           if (strpos($route_name, 'view.articles.') === 0) {
+            $variables['#cache']['contexts'][] = 'route.name';
             $variables['content_heading'] = "h2";
           }
           if ($node->hasField('field_article_source_link_direct')) {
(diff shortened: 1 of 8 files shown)


richardbporter:
Tested the footer text color issue - fixed. 👍 

Currently getting an error on articles view page on gis.sites:
```
The website encountered an unexpected error. Please try again later.
AssertionError: assert(\Drupal::service('cache_contexts_manager')->assertValidTokens($cache_contexts)) in assert() (line 33 of core/lib/Drupal/Core/Cache/Cache.php).
```

richardbporter:
I think the cache context should be set to `route.name` which is one of the available core cache contexts: https://www.drupal.org/docs/drupal-apis/cache-api/cache-contexts#core-contexts Try that maybe? Ex. `$variables['#cache']['contexts'][] = 'route.name';`

Its not ideal because each route where this is used will be cached separately but I'm not sure what else to use.

joewhitsitt:
I just tried that. It was working while logged in, but then didn't when logged out.

- cr
- visit paragraphs articles
- visit /articles

richardbporter:
I think you'll want to set the cache context for every implementation of the teaser view mode but only modify the content heading if on the view route. Ex.
```
if ($variables['view_mode'] == 'teaser') {
$variables['#cache']['contexts'][] = 'route.name';

  $route_name = $current_route->getRouteName();

  if (strpos($route_name, 'view.articles.') === 0) {
    $variables['content_heading'] = "h2";
  }
```

Otherwise the cache context is only added for the view route.

richardbporter (approved):
I think I already tested everything. If I was nitpicking, I prefer single quotes unless double are required for a specific reason. Ex. `$variables['content_heading'] = "h2";` :)

SI checked out for nav on gis.sites.uiowa.edu.

joewhitsitt:
Last min addition is to pull templates local to uids_base to switch the heading on the pagination to a div. These are already visually hidden so the change is transparent.

PR #1932, opened 2020-07-10
Feature: Sitenow v2 split

pyrello:
Resolves #1807. 
Resolves #1926. 
Resolves #1973

This update moves configuration related to the Paragraphs-based layout system into a split called `sitenow_v2`. Additionally, it runs an update hook that removes paragraph instances from sites that have the `collegiate` split enabled and adds the `sitenow_v2` split to all other sites.

A couple things to note:
- I have not moved any of the non-paragraphs configuration to the SiteNow V2 split.
- Paragraphs is not being turned off in the default profile because it will be needed for #1529.  
- I updated the page type to be called just "Page" instead of "Basic page" because I think "Basic page" is confusing in the context of being able to create a complex layout.

# To test

## Install new site
* `blt drupal:install --site=cogscilang.grad.uiowa.edu` or another site locally. You are just temporarily installing over top of an existing site.
* Check that fresh site install does not have `sitenow_v2` split enabled.
* Check that you are able to customize the layout of new and existing pages.

### V2 split on new site
* drush @gradcogscilang.local cim --source ../config/features/sitenow_v2 --partial
* Check that site is consistent with experience of current SiteNow sites.

## Update `collegiate` sites
* `blt ds --site=cogscilang.grad.uiowa.edu` or the `collegiate` site of your choice.
* Check to make sure that existing paragraph types (and therefore instances also) have been removed.

## Update non-`collegiate` sites
* `blt ds --site=brand.uiowa.edu` or the non-`collegiate` site of your choice.
* Check that `sitenow_v2` split is enabled.
* Check that content blocks field still functions as expected.

PR #1954, opened 2020-07-17
Roll back footer content removal

### LINKED ISSUE #1949: Footer addresses have disappeared from existing sites after latest release
if you look at the footer on the [https://americanstudies.prod.drupal.uiowa.edu/](https://americanstudies.prod.drupal.uiowa.edu/) site, youll see that the footer address is now missing. I did not remove this, but I also havent updated the site to use a new solution if the footer address block was removed.

cory-skeers:
Resolves #1949 

# Problem
uids_base rollout led to the removal of content in the Footer Contact Info Block (due to Additional Info migrate update hook), resulting in missing contact information for sites. Particularly an issue with collegiate sites, sitenow sites which did not have Additional Info will have lost default contact content in this block.

# Test
sync and run update hook
If prod site _has_ content in this block, no change
If prod site has no content in this block currently, it should now be visible with the last updated content


cory-skeers:
As it's set, it doesn't check for time, but only if the content has been nulled out--if so, revert to last revision, and if not, no change.

This should account for not changing blocks if anyone has updated this block since the issue, and the check for collegiate split could potentially be removed to restore default content to sitenow sites as well.

joewhitsitt:
@cory-skeers is this ready to review? drafts can't be merged until it is switched.

joewhitsitt:
pulling down for testing

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -964,3 +964,38 @@ function sitenow_update_8029() {
...
+
+/**
+ * Roll back contact block revision on collegiate sites.
+ */
+function sitenow_update_8030() {
+  /** @var \Drupal\Core\Plugin\DefaultPluginManager $filters */
+  $filters = \Drupal::service('plugin.manager.config_filter')->getDefinitions();
+  $collegiate_split = 'config_split:collegiate';
+  // This site has the 'collegiate' split enabled.
+  if (isset($filters[$collegiate_split]) && $filters[$collegiate_split]['status']) {
```
richardbporter, line 976: FYI `\Drupal::configFactory()->getEditable('config_split.config_split.collegiate')->get('status');` also seems to work here. I wasn't sure it would since status is ignored.

https://github.com/uiowa/uiowa/blob/deb4c8c179a37fd553ad54cb7da230339ff54769/config/default/config_ignore.settings.yml#L6

richardbporter:
Seemed to work for homepage:
image

bspeare:
Will this remove the site name that is appearing in the footer on https://brand.uiowa.edu/?

joewhitsitt:
I don't think it will change brand. this is only for collegiate split sites.

That said, I don't know what content to expect on https://gradcogscilang.local.drupal.uiowa.edu/ -  I am not seeing a change but during stand up @pyrello made it sound like content was missing.

joewhitsitt:
It did work on icsa:
Screen Shot 2020-07-17 at 1 27 12 PM

bspeare:
OK, do we know why there are `<h2>` showing up with <`h2 class="footer__site--name">UI Non-Discrimination Statement</h2>` Like on:

Screen Shot 2020-07-17 at 1 26 21 PM

or 

Screen Shot 2020-07-17 at 1 27 53 PM

joewhitsitt:
HR was unaffected

joewhitsitt:
@bspeare this is footer content that was migrated from uiowa_footer. It technically worked. Just doesn't look great. 
https://github.com/uiowa/uiowa/pull/1806/files#diff-86202f85a83d188875ad3349908b71c4R866

Contrast fix will go out in the next release

brand was technically a v2 site that was adopting uids_base in this case

bspeare:
@joewhitsitt So I should just manually delete it from the brand footer then?

richardbporter:
brand was unaffected

pyrello:
@joewhitsitt I checked it against the D8 Grad site and it was gone there. I don't think cogscilang has anything there.

richardbporter:
> @joewhitsitt So I should just manually delete it from the brand footer then?

Thats one option but also https://github.com/uiowa/uiowa/commit/deb4c8c179a37fd553ad54cb7da230339ff54769 should fix it everywhere.

richardbporter (approved):
americanstudies unaffected

joewhitsitt:
@bspeare if you can justify editing content on the brand site, then yes, I would remove it.

pyrello:
Technically, brand shouldn't have been updated before because they had adopted to the new contact info block with the transition to UIDS Base. Maybe the check for whether to run should have just been for sites that didn't have UIDS Base set as the default theme.

docroot/profiles/custom/sitenow/sitenow.install
pyrello: Thanks for this, @richardbporter. This might help me resolve the issues I had run into in #1932.

Issue #1908, opened 2020-07-07
Fix page title / featured image preview within layout builder edit screen


bspeare:
# Problem

There are a few issues that need to be addressed:

- The layout builder edit screen for pages contains a page title / hero section that is vertically condensed due to the `vh` property set on the hero container. 
- There are also multiple "Add Banner" options within the page title hero section - do we need to remove the ability to add blocks within this section?
- There is also no page title/breadcrumbs preview set if the page doesn't use a featured image. 

Screenshots:

<img width="1826" alt="Screen Shot 2020-07-07 at 10 59 30 AM" src="https://user-images.githubusercontent.com/1036433/86814049-82a02b00-c046-11ea-9550-08540f58d549.png">
<img width="1827" alt="Screen Shot 2020-07-07 at 10 58 55 AM" src="https://user-images.githubusercontent.com/1036433/86814078-8b90fc80-c046-11ea-84e7-f8e0522d5af4.png">

# Proposed Solution(s)

1. Adjust styles for hero component within `admin.scss` in uids_base 
2. Investigate method to show page title block by default and only show page title/hero component if a feature image is set. 
3. Remove option to add custom blocks within the Header Layout with Background Media

richardbporter:
Should the featured image be showing in the layout builder edit screen when that field is set on the node? I'm not seeing it for some reason.

richardbporter:
> Should the featured image be showing in the layout builder edit screen when that field is set on the node? I'm not seeing it for some reason.

I think this is working. Might be a local issue with lazy loading and/or stage file proxy.

richardbporter:
> There is also no page title/breadcrumbs preview set if the page doesn't use a featured image.

I'm not seeing that. Seems to appear regardless. The only time it doesn't show is on the frontpage but thats by design.

PR #1965, opened 2020-07-20
Default LB off-canvas dialog to 500px and style drag handle.

### LINKED ISSUE #1366: Improve Layout Builder drag handle
Layout builder has a hidden drag handle:
Screen Shot 2020-04-09 at 4 21 02 PM

It can be a very useful tool when it works right, but sometimes, when using it, the settings tray develops a gap on the right hand side that gets larger each subsequent time it gets moved.
Screen Shot 2020-04-09 at 3 57 19 PM
Screen Shot 2020-04-09 at 4 30 13 PM

It would be nice to fix this so that we could let site owners and content editors make use of it.

richardbporter:
Resolves #1366 

Attempting this, I definitely came around on #1927 - in fact I went further and just hid it entirely on layout builder forms because it **does not** show node information at all. Maybe we should do this for all admin paths so it only displays on the front end?

I tried to set the dialog width using the Drupal API but that was proving difficult: https://www.drupal.org/project/drupal/issues/2951547 Eventually went with JS since the existing inline styles can't be removed from the element.

I ripped off the lb_ux tray exactly. If someone has style ideas for this, I would say go for it in this PR.

### Testing
- Check out this branch.
- Edit a layout on a collegiate site.
- See the wider tray and drag handle.
- See that the form elements do not have extra space on the right after making the tray more narrow.
- Test multiple browsers, viewport widths, etc.


richardbporter:
Also, I was unable to reproduce the extra padding on the right of form elements as described in #1366.

bspeare (approved):
This works great!  I love the wider tray and drag handle.  I tested this locally on safari, chrome, and firefox.  One thing I wonder about is if there is a way to only set the default width to `500px` for adding blocks and keep it at the system default for adding layout sections?  There is quite a lot of extra space for the section layouts since they are just icons. 

Screen Shot 2020-07-20 at 3 45 20 PM

richardbporter:
Possibly. I can investigate.

richardbporter:
Ugh. Now I see the extra space on the right. It only happens after making the dialog more narrow.

richardbporter:
Something I noticed when working on this is the header section styles is borked when editing the layout:
image

Does anyone else see this? Do we have an issue for it?

bspeare:
@richardbporter Yes, I created one here https://github.com/uiowa/uiowa/issues/1908.

richardbporter:
Great, thanks. I think I've addressed your previous comment if you want to give it another look. I also addressed that extra space to the right of form elements on redraw and an overflow issue with the text format link icon.

pyrello:
@richardbporter There is a CSS min-height fix that needs to happen there. It is using a container that depends on having an image inside of it.

I'm looking forward to testing this out. Will take a look after 3 PM.

richardbporter:
Cool thanks. I should say my CSS chops are dated so there may be a better approach to some of these things.

pyrello:
Probably not with this PR, but I think it would be great if there was a cookie that remembered the width that you set the off canvas container to and rendered it at that width.

bspeare:
@richardbporter the updates look good.   I am seeing a weird jump with the tray when I try to add a custom block.  Do you see that? I can post a video capture to Slack if that helps.

richardbporter:
Yes, I think that is from adding the styles after the dialog is rendered. Maybe its possible to do all this in CSS.

pyrello:
@richardbporter I'm still seeing the behavior reported in #1366.

bspeare:
@pyrello which browser or block component?  It looks good in chrome and firefox.

pyrello:
Chrome, on the body field.

pyrello:
It happens immediately when I drag the handle out to a wider width.

pyrello:
Nevermind... I forgot to `git pull` :|

richardbporter:
@bspeare Can you look at that again? Seemed better after the last change.

richardbporter:
Also, we're not able to set this via CSS because it overrides the adjustments that happen on drag. Kinda silly this can't be set more easily with Drupal API. :/

bspeare:
@richardbporter changing it to 

``` 
$(this).css('width', '500px');
$(this).css('left', 'auto');
$(this).css('right', '0');
```
Seemed to fix it for me.  Does that mess with anything for you?

GaryRidgway:
trying to get my local to work because it is busted again, but to make sure we dont make too many DOM calls with jquery you can allocate one css call to $(this) by calling the css property as such:

`$(this).css({
    'width':'500px',
    'left':'auto',
    'right':'0'
});`

bspeare (approved):
Looks great, I don't see the jump anymore - thanks!

GaryRidgway:
I gave the handle a more modern feel to it, hope everyone likes it! here is a quick screenshot:

Screen Shot 2020-07-21 at 5 02 11 PM

pyrello:
That looks great, @GaryRidgway!

pyrello:
I'm going to try something with this quick.

pyrello:
I refactored the javascript a little bit (based on looking at other core examples) and added a cookie for remembering the drag handle setting from page to page.

Also, it sets the cookie the when the `dialogContentResize.off-canvas` event fires, which happens right away when it is first opened. I'm not sure if there is a better way to do it so that the cookie only fires on an event that is triggered by the user or not. The cookie is set to last for 99 years, which seems like a long time for something that people didn't necessarily choose to set.

pyrello:
Okay, I think I figured out how to prevent it from responding to non-interaction. If it is the first time the `dialogContentResize.off-canvas` event fires, there is a flag set to true indicating that and the process of setting the cookie is skipped and the flag is set to false.

docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @file
+ */
+(function ($, Drupal) {
+    Drupal.behaviors.layoutBuilderCustomOverrides = {
+        attach: function (context, settings) {
+          $(window).once('off-canvas-overrides').on({
+            'dialog:aftercreate': function (event, dialog, $element) {
+              let justCreated = true;
+              console.log('event', event);
```
richardbporter: Leftover console.log here.

bspeare:
@pyrello The cookie works great for setting the width.  It's a great improvement!  Any chance it can remember the size of the "Add Section" tray? :smiley:

docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
```diff
@@ -0,0 +1,36 @@
...
+              if (Drupal.offCanvas.isOffCanvas($element) && $element.find('.layout-selection').length === 0) {
+                let offCanvasWidth;
+                const offCanvasCookie = $.cookie('ui_off_canvas_width');
+                if (offCanvasCookie === undefined) {
+                  offCanvasWidth = 500;
+                } else {
+                  offCanvasWidth = offCanvasCookie;
+                }
+                $($element).parent().attr('style', `position: fixed; width: ${offCanvasWidth}px; right: 0; left: auto;`);
```
richardbporter: I love template literals but they're not supported in IE11 (which I think we support?) https://caniuse.com/#feat=template-literals 

I guess we can go old-school concatenation or set up a Yarn workspace. The former is probably faster/easier.

docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
pyrello: Fixed.

pyrello:
@bspeare You want it to remember the 'Add section' tray as a separate option? Per the request earlier, that tray is excluded from the width setting.

GaryRidgway:
@pyrello  will that cookie be able to work cross sites? like if i set the cookie on one site, and edit another site, will it remember my desired tray width?

pyrello:
@GaryRidgway I think we might run into cross-site issues with that, but maybe it would be possible to set a *.uiowa.edu cookie? I'm honestly not that familiar with the subject.

GaryRidgway:
I know you can just save things to a persons browser and it should just be able to be accessed like any other json data, if that is useful, but im not sure of the ethicality of that versus using cookies, or how it differs mechanically

richardbporter (commented):
Whoops. Never submitted this.

docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
richardbporter: Went ahead and replaced it.

pyrello: Sounds good. Eventually would like to set up a modern workflow for all js, but until then, old-school is probably best.

richardbporter: Agreed. Created a follow up https://github.com/uiowa/uiowa/issues/1977

richardbporter:
Tested in IE11. 👍

richardbporter:
Sounds like there two questions left here: 

- Whether to save the 'add section' tray width in what I'm assuming would be a different cookie, otherwise I think it'd be quite wide in most cases.
- Whether to expand the cookie domain to our other sites.

richardbporter:
We can expand the cookie to all of uiowa.edu but not vanity domains.

I'm a little unclear on the desired behavior for 'Add section' tray width.

richardbporter:
I don't see much benefit to expanding the add section tray because the list options don't fill the space: 
image


@bspeare Is the issue that 300px is too wide?

bspeare:
@richardbporter There's no need to address the "Add Section" tray.  I think the solution that is there works great.  I just liked the idea of it remembering the tray width if I made it smaller than the default size.

richardbporter:
That would be nice. :)

I created a follow up issue for that: https://github.com/uiowa/uiowa/issues/1981.

pyrello:
Nice work on this @richardbporter, @bspeare, and @GaryRidgway! This is a solid enhancement.

PR #1976, opened 2020-07-22
Add P2LB  module and feature split

### LINKED ISSUE #1803: Option for automated content move from Paragraphs to Layout Builder
## Problem
Depending on timelines and methodology, possibility of an automated update to move content from paragraphs to layout builder could ease the transition.

### Update
Migration option is being built out in a custom module (sitenow_p2lb). Current functionality works (with some bugs) with text and card content blocks.

### Content types to account for from paragraphs
- [x] Text
- [x] Card
- [x] Accordion/accordion item
- [x] Blockquote
- [x] Image gallery
- [x] Webform
- [x] Featured content
- [ ] Articles/Events/People
  - [ ] Need updating, specifically People
- [x] Carousel/carousel image

### Other to-dos
- [x] Revisioning bug leading to migration of previously deleted content blocks
- [x] Section overrun bug (some cases lead to 5 or more column sections, which are invalid)
- [x] Section background images
- [x] Optional inclusion of section/content styling choices like background colors
- [ ] Improved options
  - [ ] Review of moved content?
  - [ ] Auto-removal of old paragraphs after migrate/review?
  - [ ] Rollback or revision option?
  - [ ] On-page conversion option rather than/in addition to separate menu?
  - [x] Bulk node updates?
  - [ ] Update perms split?
  - [ ] Revisions?
    - [ ] Remove previous node revisions to avoid rolling back pre-update? (low cost)
    - [ ] Create block revisions based on paragraph revisions, and update node revisions? (questionable...high cost)
  - [ ] Content moderation? Check that we have published and possible drafts updated.
- [ ] Lots of testing...
- [ ] Documentation and/or decision on methodology of rollout--user-run, dev-run, automated in an update hook, etc.?
- [ ] Remove sitenow_v2 flags in sites' blt.yml file
- [ ] Other things
- [ ] Handle cards with videos somehow
- [ ] Prevent deleting the rollback revision: https://github.com/uiowa/uiowa/issues/1803#issuecomment-876835327

### Migrate modules
Some work may be repurposed from SiteNow Migrate for the paragraphs end--SiteNow Migrate may also need to be updated if any automated D7->D8 migrations into layout builder would take place.

### Update hook
As multiple databases won't need to be handled as in a typical site migration, it's possible to include the entirety of the update in an update hook.

cory-skeers:
Building block for an automated move from paragraphs to layout builder. Resolves #1803.

# To test
- `ddev blt ds --site brand.uiowa.edu`
- Observe there are no issues with site sync.
- `ddev drush @brand.local uli`
- Ensure that no P2LB functionality is present on the brand site.

<details>
<summary>Original post</summary>
Process is encapsulated in the sitenow_p2lb module, which adds a config menu at /admin/config/system/sitenow-p2lb. Currently select nodes (ids populated based on presence of paragraphs content on the page), and the option to delete all paragraphs content (tester, or to use following migrate) or update, which runs through the various functions within the main module to move content over. The moved content remains on the same page, so that internal links are preserved.

List of functionality considerations started on #1803.

Couple additional notes:
- Depending on resolution of #1555, some refactoring will be needed. Current build assumes all paragraphs on a site should be removed, which may not be the case.
</details>


docroot/modules/custom/sitenow_p2lb/src/Form/P2LbSettingsForm.php
```diff
@@ -0,0 +1,105 @@
...
+   */
+  public function updateButton(array &$form, FormStateInterface $form_state) {
+    // Grab nids for all boxes that were checked (0s are filtered out).
+    $nids = array_filter(array_values($form_state->getValue('nodes_w_paragraphs')));
+    foreach ($nids as $nid) {
+      sitenow_p2lb_node_p2lb($nid);
+    }
+    // @todo Option to remove paragraphs after migrate, or review first?
+    return $form_state;
+  }
```
pyrello: We might want to look at using the Queue API for processing a large amount of nodes, if that situation exists: https://www.sitepoint.com/drupal-8-queue-api-powerful-manual-and-cron-queueing/

cory-skeers:
## To do:
All newly created blocks here need to have their usage/access dependency applied.

pyrello:
I think this is now blocked by #2259. Also, probably #2358.

pyrello:
Adding a note about refactoring to set headlines where possible.

richardbporter:
I'm playing around with this to get familiar with it but running into a lot of errors/warnings when using the magic button. Are there specific steps needed to set up a site locally for this? Should I document those errors/warnings in #3568 ?

cory-skeers:
> I'm playing around with this to get familiar with it but running into a lot of errors/warnings when using the magic button. Are there specific steps needed to set up a site locally for this? Should I document those errors/warnings in #3568 ?

There shouldn't be any specific steps to set it up, so most/all of the warnings and errors you're getting will need to be addressed. Documenting in #3568 sounds good.

pyrello:
Putting some thoughts down about how to handle the site owner admin UI for the period of time where we are allowing people to convert their content on their own.

I was able to successfully wire up the existing code as an action that can be triggered via ~Views Bulk Operations~ core's bulk update system, which allows us to just use Views to build the page for the user to manage conversions. ~We don't currently use Views Bulk Operations, since some of this content was moved into core in D8. If we don't want to start using it going forward, we can just use it until we are done with this process and only need it turned on for V2 sites.~

For the V2 split, we should add a field to the page content type that we can use to track the revision ID that was created when we ran the conversion operation. This would allow us to easily tell the difference between pages that have been converted or not, which would be useful for the admin UI. It would also allow an easy way to know which revision to revert back to, if that were necessary.

Having the above field would also potentially allow us to manipulate what options are shown to the user depending on what "state" the page is in. If it has not been converted, then it could show the Paragraphs UI to allow the user to continue to edit the page. If it has been converted, then we could hide the Paragraphs UI to prevent confusion about where the content should go.

cory-skeers:
New method for generating the header section is working well!

Also did light testing on the image gallery migrate, and it appears to be working. I wasn't very thorough in the testing, but everything functioned as expected in the limited tests.

pyrello:
I adjusted the that we are pulling in the config for the content moderation and header sections, since it is already defined in our default config. I just tested it and that approach works.

I worked on getting Featured Content converting, but last check I did of that, it wasn't working. I'm going to test it again now that I can see the rest of the process seems to be functioning as expected again.

pyrello:
I tested Featured Content again, but it is still not working. I'm not sure why, because it looks like everything is set up correctly. However, I'm going to step back from this for a little bit so I can focus on some other things.

pyrello:
Per conversation today, I moved the P2LB module and config to a separate split from the `sitenow_v2` split so that it can be enabled separately.

pyrello:
I believe that we should go ahead and merge this branch after it is reviewed without being too concerned about the current state of its functionality (although if @cory-skeers or anyone else wants to sneak in some quick fixes, I'm not opposed to that). At this point the P2LB stuff should not be functional unless turned on, so I think it is safe to merge it in its current state. Then we can use separate PRs to incrementally fix or improve it.

joewhitsitt:
taking a look see

joewhitsitt (approved):
It is not on, and I don't see any of it as a webmaster or admin

cory-skeers:
Oh gosh. I've waited so long to merge this.

Gonna do it. Gonna press the button!

pyrello:
Yay!! Thanks, @joewhitsitt! Congrats, @cory-skeers!

Issue #955, opened 2019-12-03
Restrict ability to add block (types) to layout regions


pyrello:
Tracking issue from https://www.drupal.org/project/layout_builder_restrictions/issues/2982452#comment-13367752.

Once a patch is created, test and add to the project.

bspeare:
Also related: https://www.drupal.org/docs/8/modules/layout-builder-restrictions/restricting-blocks-to-specific-layouts-regions

pyrello:
This is actually being used in #1622

richardbporter:
Seems like this was resolved in #1979. Reopen if not!

joewhitsitt:
Is this "done?" I was going to restrict more general non-layout specific blocks here: https://github.com/uiowa/uiowa/issues/1920

richardbporter:
I'm not sure. I thought it was initially but then read https://github.com/uiowa/uiowa/pull/1979#issuecomment-670245133 which maybe indicated it was not completely done?

richardbporter:
I would be in favor of closing this since #1920 represents the further investigation.

pyrello:
Sounds good to me.

PR #1979, opened 2020-07-22
Feature layout block restrictions

pyrello:
Relates to #955

# Updates
Removes the ability to add the page title hero and the spacer anywhere. Restricts banner/hero in multi-column layouts.

# To test
* `composer install` - There is an update to `lb_direct_add` module that is necessary for restricted options to not show up.
* View the instructions at #1932 about how to pull down/install various layout builder test sites.
* Visit a page and click the layout tab (e.g. https://gradcogscilang.local.drupal.uiowa.edu/node/1/layout)
* Add 1, 2, 3, and 4 column layouts and check the following:
  * Banner/Hero should not be available on 2-, 3-, and 4-column layouts.
  * Page title hero and separator/spacer should not be available.

## Policy site
Policy splits its entity_display_view for pages, so I copied these changes to that split as well. @mark-bennett-uiowa, please feel free to give feedback about this.


richardbporter:
I'm looking through #1778 to find another issue to work on - is this addressing https://github.com/uiowa/uiowa/issues/955?

pyrello:
@richardbporter Yes.

pyrello:
Related to https://github.com/uiowa/uiowa/pull/1932#issuecomment-662518542.

pyrello:
Follow up to this issue: #1987.

bspeare:
@pyrello Here's what I am seeing for each section, maybe I installed the site wrong?

**One column template section**

Screen Shot 2020-07-24 at 12 52 19 PM

**One column**
Screen Shot 2020-07-24 at 12 52 33 PM

**Two Column**
Screen Shot 2020-07-24 at 12 56 22 PM


**Three Column** 
Screen Shot 2020-07-24 at 12 53 16 PM

**Four Column** 
Screen Shot 2020-07-24 at 12 53 29 PM

pyrello:
I believe that the issues about the wrong blocks showing up have been resolved.

joewhitsitt:
No longer blocked. #1932 was merged/deployed

joewhitsitt:
Pulling down for testing using a fresh default site install I am seeing similar problems that Ben was experiencing with the wrong block available in the wrong columns.

If page title hero and separator are supposed to be available at all except for the homepage, I think this could be better address by block_content_permissions similar to how vertical video was addressed.

I will see if I can get back to this later today.

pyrello:
I'm not sure how, but it looks like some of my updates (https://github.com/uiowa/uiowa/pull/1979/commits/7b1ac38a263e70ee6d9ed2b519d595530581a3df) were reverted somehow. The current version of this branch doesn't show those changes:

https://github.com/uiowa/uiowa/blob/9f7f301077fef455849dc561131036c4c188431e/config/default/core.entity_view_display.node.page.default.yml#L230-L260

Anyways, that is why it is "not working".

pyrello:
Also, please note that it is necessary to run `composer install` before testing. This includes an update to the `lb_direct_add` module which respects the layout builder restrictions.

pyrello:
One of the issues that are we are running into with this is that the layout builder restrictions aren't smart enough to differentiate between using layout builder for creating a content template and when its being used for per-node overrides. There also isn't currently a permission that allows someone with a specific role to bypass the restriction checks. This means that whatever restrictions we put in place will be in place for all users regardless of whether they are administrators or not. That may be okay, but it is something that needs to be considered.

pyrello:
The other situation that I have been trying to account for is when someone wants to build a page that is different than the default layout. Maybe they remove elements that were previously placed through the default template. At that point, if we have disabled all those options, then there is no way for them to restore certain elements that they might have removed.

joewhitsitt (approved):
I am still seeing (as a webmaster) the Page title hero and separator/spacer for the collegiate split. Works as expected for the default. I think the same permissions need to be adjusted for the collegiate split if desired: https://github.com/uiowa/uiowa/pull/1979/files#diff-39db5cfa12c0f3449fb551bdd9f70882

Having troubles testing the header section due to https://github.com/uiowa/uiowa/issues/1908

Approving for the default. Follow-up might be needed for collegiate/uiowa.edu as this would affect existing content.

pyrello:
@joewhitsitt My understanding is that the restrictions don't affect existing content because they only limit what gets showed as options for adding blocks via layout builder. If a block already exists, I don't think this prevents a user with appropriate permissions to edit it even if it on the restriction list.

joewhitsitt:
block_content_permissions if set for the collegiate split similar to how I set it for the default config will affect existing content because a webmaster/editor won't be able to edit/remove existing. if the goal of the pr is to hide page title hero and separate across all splits, then more work is needed because it does not currently achieve this for the collegiate split.

pyrello:
I think we will probably end up revisiting this issue. If there are more fields that we want to make initially unavailable for SiteNow V3 heading into beta testing, that is okay with me.

PR #1986, opened 2020-07-23
I1982 standardize layout builder consistency

### LINKED ISSUE #1982: Standardize Layout Builder Block Fields for Consistency
**Huge S/O to Sean and Joe for explaining how to handle this**

Make field implementation consistent across blocks and remove unused fields

- [ ] 1. Consistent Display Title/Title field across blocks
- [ ] 2. Consistent Link Fields 
- [ ] 3. Don't display the option to override styles if the block doesn't have any styles to apply

**1. Block Title Options**

- OPTION 1: Hide the block description field entirely. If we do this, then the blocks will automatically get block descriptions like “Card”, “Hero”, etc. 
- OPTION 2: Rename the block description field to ‘Administrative label’ to make it more consistent with SiteNow V2 and keep the current behavior. This would also require going through the blocks and making sure that they are consistent with the block description being available.

**2. Link Fields**
Make Link fields consistent across all blocks - some blocks require `'Link Text',` some give the _option_ for the user to provide `'Link Text'` and one block doesn't have the`'Link Text'` option available

**3. Block Overrides**
Some blocks have the 'Override Block Styles' option, but when selected, no additional styles are provided for the specific block. If the block doesn't have extra styles that can be applied, remove the field option.

### LINKED ISSUE #2049: Ajax error when adding non-block components via layout builder
```"
An AJAX HTTP error occurred.
HTTP Result Code: 200
Debugging information follows.
Path: /layout_builder/add/block/overrides/node.2/1/sidebar/field_block%3Anode%3Apage%3Afield_teaser
StatusText: OK
ResponseText: Error: Call to a member function bundle() on null in layout_builder_custom_form_alter() (line 37 of /var/www/uiowa/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module)."
```

This will be fixed by #1986 but it should be noted that the process callback runs twice: once when the off-canvas form is loaded and once on submit. We may want to figure out how to limit that to just the former.

dnkemper:
Resolves #1982, resolves #2049

**Title inconsistency**
The Image Gallery block and Text Area block didn't have a title field like the other existing blocks. They were displaying the input for the 'Display Title' field [similar to SiteNow v2 'Administrative Label' field for paragraphs] instead of displaying a separate 'Title' field in the templates. Added a title field to both blocks for consistency.

**Next tasks:** 
- [ ] ~~Change the 'Display Title' field text to 'Administrative Label' or something more accurate to represent what it does.~~ 
- [ ] ~~Also add more descriptive help text to 'Display Title' field~~
- [x] Find a way to not display block style override checkbox if the block doesn't have extra styles 
- [ ] ~~Make Link field consistent across all blocks~~
- [x] Fix Error message so user can easily spot the field that errored
- [x] Way to show field is required?


pyrello:
I was working on the custom code that changes the behavior for the block description field (which displays as "Title" text field and "Display title" checkbox) and I think that maybe we just want to remove this from the form entirely. It is not meant to be a title field and should never be displayed in the rendered output. Also, we specifically don't render it in most (if not all) of the block templates. I also don't see where it shows up anywhere as an actual administrative label, so I'm not sure what the value of letting the user set it would be. I can make those changes in #2000 as part of some other refactoring that is being done.

richardbporter:
>Fix Error message so user can easily spot the field that errored
 Way to show field is required?

We do seem to be missing some form validation styles like the red asterisks and the JS message that states "Please fill out this field." Does anyone know if we are unsetting a core library or something?

image

richardbporter:
> Fix Error message so user can easily spot the field that errored

Rendering them in red is hard to read. The asterisk I added affects both the frontend and backend. You can see it on a webform, for example. @bspeare Any thoughts on styles here?

The "error messages" are `strong` elements that appear above/below fields. I agree they don't stand out but with the dark `#off-canvas` background, red text doesn't really work.

image

config/default/field.field.block_content.uiowa_image_gallery.field_uiowa_gallery_title.yml
```diff
@@ -0,0 +1,32 @@
...
+    - field.storage.block_content.field_uiowa_gallery_title
+  module:
+    - heading
+id: block_content.uiowa_image_gallery.field_uiowa_gallery_title
+field_name: field_uiowa_gallery_title
+entity_type: block_content
+bundle: uiowa_image_gallery
+label: Title
+description: ''
+required: true
```
richardbporter: I can see a use case for an image gallery without a title.

richardbporter:
> > Fix Error message so user can easily spot the field that errored
> > Way to show field is required?
> 
> We do seem to be missing some form validation styles like the red asterisks and the JS message that states "Please fill out this field." Does anyone know if we are unsetting a core library or something?

I added some required field styles instead of trying to add in Claro's form styles.

config/default/field.field.block_content.uiowa_image_gallery.field_uiowa_gallery_title.yml
pyrello: Agree

richardbporter:
> Make Link field consistent across all blocks

Anyone know what the inconsistency is?

docroot/themes/custom/uids_base/config/optional/block.block.uids_base_page_title.yml
```diff
@@ -13,7 +13,7 @@ provider: null
 plugin: page_title_block
 settings:
   id: page_title_block
-  label: 'Page title'
+  label: 'Page title test'
```
richardbporter: This should probably be removed.

richardbporter:
> > Make Link field consistent across all blocks
> 
> Anyone know what the inconsistency is?

As far as I can tell, the differences are by design to accommodate the design system. For example, the optional link text in the card block allows for a button to render or not. The CTA requires link text as the entire component is not linked so a button is necessary. 

The only oddity I found is that the CTA does not require any fields which makes it possible to add a completely empty component. I feel like one or more fields should be required.

richardbporter:
I required the CTA link field as it looks weird without one. Its also titled "Call to Action" and if there's no link, there's no action and another component would be more appropriate.

joewhitsitt:
So in reviewing this PR, I realized that https://github.com/uiowa/uiowa/issues/2048 which we discussed this morning conflicts with the title field pattern established with the commits early in this PR for adding title fields to image gallery and text area. I think this can move forward as-is as long as we are okay with looping through and migrating data to the new parent title super field later.

Don't love the one-off form required styles. Is there a core issue out there for this or is another contrib module conflicting here?

Everything else, checks out IMO.

richardbporter:
> Don't love the one-off form required styles. Is there a core issue out there for this or is another contrib module conflicting here?

It's because uids_base is unsetting the classy/base library for some reason. It includes form styles like required, etc. I'm not sure of another solution.

https://github.com/uiowa/uiowa/blob/9e183d9dc4db37fb86f406f13d56320af4efa4e4/docroot/themes/custom/uids_base/uids_base.info.yml#L11

pyrello:
Per our discussion in Parking Lot (and after) yesterday, it didn't seem like there was any appetite to add a title field to the text area block at this point in time. However, that is being done as part of this PR: https://github.com/uiowa/uiowa/pull/1986/files#diff-bca9f01cd7d86e990301268c1da6c486R8. I think this should be removed, since we essentially decided not to take it on yesterday.

richardbporter:
I deleted the gallery and text area title fields for now. They'll get added back when we address #2048.

joewhitsitt (approved):

richardbporter:
Its probably best to merge this given the refactor to docroot/modules/custom/layout_builder_custom/layout_builder_custom.module and also the cherry-pick of 8d9a0b5 which was somehow lost from main. I think we can create follow up issues to change/improve anything in here.

PR #1988, opened 2020-07-23
Update Drush users commands from 2.1.2 => 2.1.3.

richardbporter:
This resolves an issue with multiple values used with the role option. Ex. `drush users:list --roles 'editor,webmaster'

### Testing
It works.
- https://github.com/richardbporter/drush-users-commands/blob/main/tests/ListTest.php#L40
- https://travis-ci.com/github/richardbporter/drush-users-commands/builds/176926185

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -222,7 +222,7 @@
             "version": "5.13.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Reviewed, not tested.

PR #1991, opened 2020-07-23
Feature: Layout Builder Slider block

pyrello:
This PR picks up from #1529 and splits out the Slider block from the rest of it.

# To test
* `blt ds --site=uiowa.edu` or another collegiate site of your choice.
* Edit layout on a page and add a Slider.
* Try to break it.
* Tell me what doesn't make sense or could be better.


cory-skeers:
Couple things:
- An image isn't required, but looks weird if there isn't one for a slide but images for others (might be a "with great flexibility comes great power to make ugly things" situation)
- When there is no image, looks like the headline is placed above the slider (in the screenshot, it is hidden behind the editor tabs)
- Does the slider require three slides? Whether it's controlled or helper text, easy to miss
- Headline level can be set individually on the slide, giving possibility to have strange hierarchies
- Think it's a LayoutBuilder thing in general, but the separation between "Create slide" and "Add block" is confusing--particularly it's easy to fill out everything for a slide, but click "Add block" without actually adding the slide, and losing that work
- Not sure if it's related to vertical videos/stats re-editing from the layout builder page, but I can't seem to edit the slider once the layout has been saved.

image

pyrello:
@cory-skeers Thanks for testing! Great feedback.

* For the slide, an image should probably be a required field. I'm a little tentative about making too many fields "required" just because that limits people's ability to start fleshing something out and come back to it later to fill in the unknown details.
* Slider does not require 3 slides.
* This is a good point about the headline. For the other collection types, I had created a top level 'Heading size' field and then programmatically set that for the individual items. It seems like maybe that would be appropriate for this too? @bspeare I wonder if there are any heading size options we could eliminate for this just because it is not meant to be used in multiple column layouts?
* How to handle nested items from a UX standpoint is definitely a thing that needs some more consideration. I'm wondering if there is a way that we can, upon click "Add block" on the slider block, that we can cause the IEF form to submit also. Also, we might want to add some styling changes to make the nested block more clearly separated or distinct from the parent form.
* @bspeare Is there an issue open yet for not being able to edit blocks after they have been saved in LB?

bspeare:
I think we could set all the headings to `<h2>` possibly.  Or just allow `<h2>` or `<h3>`.  We should restrict this to a one-col layout if it isn't set already.

bspeare:
@pyrello I can pull this down and see if I can apply the same fix that I set for the stat block for the editing button.

bspeare:
@pyrello I fixed the edit button and added a style fix to https://github.com/uiowa/uids/pull/291 for the mobile view of the slider.

pyrello:
These most recent updates addressed the following:
* Image is now a required field for the slide.
* The slide Headline field has been converted a text field.
* A "Heading size" field was added to the slider block type.
* Preprocess logic was added to the theme to pass the "Heading size" value from the slider to the individual slides.
* Added some basic style updates to the admin.scss file to help clarify the nested items compared to the rest of the form:
  Screen Shot 2020-07-27 at 2 18 43 PM

cory-skeers (changes requested):
The updates are looking really good! 👍  Able to make some pretty bad looking sliders, but impressed with how well it handles them--I think they fall into the "working correctly, but a poor choice" category.

I think it's an lb-wide thing, but can we add an indicator for required fields prior to trying to submit?

Only problem I ran into was the heading size seemed to be stuck at an h3. I could change it for the block and save, but on render it's still getting an h3.

image
image

cory-skeers (approved):
Added commit d28b796, and heading levels are being pulled down to the children.

Looks good to me!

pyrello:
Thanks, @cory-skeers!

PR #1993, opened 2020-07-23
Feature: Add 'Create URL alias' to publisher role

### LINKED ISSUE #1985: In base profile, allow content editors to set URL alias
# Problem
For the new Grad College site, I have a bunch of users that have the 'Content editor' role that are migrating content. As part of that process, I need for them to copy over the existing URL alias. Currently that role does not have access to that field.

# Proposed solution
Enable the 'Create and edit URL aliases' permission for the 'Content editor' role.

pyrello:
Resolves #1985 

# To test
This could probably get by with a code review. I updated the default config and then copied that one permission to the various splits that split that role.

Paging @mark-bennett-uiowa that this affects policy.clas.uiowa.edu

### DIFF
--- a/config/brand.uiowa.edu/user.role.publisher.yml
+++ b/config/brand.uiowa.edu/user.role.publisher.yml
@@ -59,6 +59,7 @@ permissions:
   - 'create remote_video media'
   - 'create terms in tags'
   - 'create twitter media'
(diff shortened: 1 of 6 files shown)


cory-skeers (approved):
Reviewed, not tested. 👍

Issue #883, opened 2020-01-02
Add media_entity_file_replace to provide file replace functionality


joewhitsitt:
## Problem
Unlike in D7 with file entity I believe, there is currently not a way to replace an existing file. There is a core issue out there https://www.drupal.org/project/drupal/issues/2648816.

## Proposed Solution
Until then this module solves the issue and is written and used by Princeton University ([bkosborne](https://www.drupal.org/u/bkosborne)).

https://www.drupal.org/project/media_entity_file_replace

joewhitsitt:
Browsing the issue queue for the module we would need to implement solutions for clearing file/media cache/purge when operations are triggered.

https://www.drupal.org/project/media_entity_file_replace/issues/3115477
https://www.drupal.org/project/media_entity_file_replace/issues/3109553

PR #1995, opened 2020-07-27
Feature: No upscaling for 'no crop' image styles and 'Original' image view mode

### LINKED ISSUE #1974: Add original view mode for small images
# Problem
When transitioning pages to the new system, sometimes there is nothing to be done but use an existing small image. Setting it to "Small - no crop" still causes the image to be processed and the resulting image can be larger than the original with all the attendant problems that has.

# Proposed solution
Add a view mode that displays the original image.

pyrello:
# Updates
Resolves #1974. 
* For default platform, I have removed upscaling from all the 'no crop' image styles. This should prevent them from appearing larger than the original image.
* Additionally, I have added the default image view mode to the collegiate text filters that allow images to be embedded. I also added some preprocessing code to rename 'Original' to 'Default,' similar to how it is done for SiteNow sites.

# To test
* `blt ds --site=cogscilang.grad.uiowa.edu`
* Edit the homepage: https://gradcogscilang.local.drupal.uiowa.edu/node/1/edit
* Insert a small image using the 'Small - no crop' option and observe whether it inserts the image as larger than it is.
* Click the 'edit media' button next to an existing image (you don't have to be able to see the image) and observe that the 'Original' option exists.


pyrello:
# Potential issues with this change
If there are any images using a no crop option (which I think is limited to collegiate sites right now) and they were counting on the upscaling to be there for the image to be the right size, this change will eventually mean those images will revert back to their original size. This only means the image will get smaller, not larger. Are there any images that you need to worry about this with @bspeare @mark-bennett-uiowa?

mark-bennett-uiowa:
@pyrello Not that I can think of the top of my head. If there is, I will adjust the images as needed in the case of an issue.

joewhitsitt:
Pulling down for testing.

joewhitsitt (approved):
"Original" works for v2 as well. I think getting rid of the upscaling is a good idea if even potentially a "regression" for some sites. It will encourage the use of appropriately sized images. However, I think something like #883 becomes more important if sites do have to swap out.

PR #1998, opened 2020-07-28
Add bigmenu module 

### LINKED ISSUE #1996: max_input_vars strikes again!
## Problem

Large drupal interfaces like menus and permission pages struggle with the server's max_input_vars limit. We solved this with Paragraphs by adding the experimental widget which collapses a nested item via ajax, reducing the amount of inputs on the form.

A site (though could be structured differently) recently ran into the max with their main menu system. As the platform grows and sites are populated, this issue will be experienced more frequently.

**Workaround**: Edit individual menu items versus trying to edit in the main manage interface.
**Workaround**: Permissions can be edited by role instead of the main manage interface.

Known Issue: https://sitenow.uiowa.edu/known-issues#uiowa-issue-1996

https://www.drupal.org/project/drupal/issues/2742371
https://www.drupal.org/project/drupal/issues/1565704

## Proposed Solution

- Up our max_input_vars number... again.
- Wait for core to fix this issue
- Investigate https://www.drupal.org/project/bigmenu for D8. This module collapses parent items via ajax which reduces the number of items being edited with each "save." We used the module on D7.

joewhitsitt:
relates to #1996

## To Test

- pull down a site with more than one-level deep main menu. cnm.uiowa.edu currently has a large menu.
- ensure the module is enabled and that you see the edit children column and links within the menu management interface
- notice how the site doesn't error if you try to modify a menu link via this interface.
- test to see if the usability is still adequate while providing error protection


richardbporter:
The menu on CNM is unusable without this module. On the flip side, I can see this being annoying for smaller menus with a few children.

Too bad there is no config to enable this when a certain amount of menu items exist.

joewhitsitt:
Right. The ux is not the same as D7. The immediate site specific issue has been resolved. I don't know of other sites with the issue so we could wait a bit on this and see what comes out of https://www.drupal.org/project/drupal/issues/1565704

joewhitsitt:
Going to close this for now:

- Basic PR that is contstantly out of date due to composer.
- We don't love the change in UX and the issue has not been reported by any other website on the platform.

Will leave #1996 open.

PR #2000, opened 2020-07-28
Feature: Add layout builder style groupings

pyrello:
# Update
Adds a patch to the `layout_builder_styles` module to allow grouping of styles together into separate fields. 
Screen Shot 2020-07-28 at 1 19 22 PM

# To test
* `blt ds --site=cogscilang.grad.uiowa.edu`
* `drush @gradcogscilang.local uli`
*  Visit the home page and click the "Layout" tab.
* Experiment with different options.
* Try to break it.
* Report feedback.


bspeare:
It looks like the "Override Block Styles" checkbox on blocks breaks with this new grouping.    The module code probably just needs to be updated with the grouping name.

pyrello:
I updated the code that handles this and moved it into the `layout_builder_custom` module to be consistent with how we are handling similar code for the the title. I also made some adjustments to labels and the ordering of elements on the inline block form.

Here is how the form displays now:
Screen Shot 2020-07-30 at 12 03 42 PM

bspeare:
@pyrello is there a way to we can force the "Background" option to be a single option and let the rest be multiple?

bspeare:
@pyrello Was that recent push meant to fix https://github.com/uiowa/uiowa/pull/2000#issuecomment-666454831?

pyrello:
@bspeare Yes, it should have fixed that issue.

pyrello:
> is there a way to we can force the "Background" option to be a single option and let the rest be multiple?

@bspeare I think so. I'll have to try something out.

pyrello:
I think I have figured out how to force the 'background' style group to not allow multiple selections.

pyrello:
Related to #1778

bspeare:
Both updates work great!

joewhitsitt (approved):
This is fantastic! I did notice the padding dropdown was open at first when configuring a section, but not a big deal.

Issue #1779, opened 2020-06-12
Moving to Layout Builder


pyrello:
- [x] #1660 and #1806: Finish SiteNow UIDS integration - This is needed to provide some relief with regard to interest in the new branding and design options that have been repeatedly demoed as part of the home page redesign project.
- [x] #1807: Move current SiteNow (V2) Paragraph config into a split.
- [x] Build SiteNow V3 using Layout Builder 
- [x] #1778: Layout Builder feature completeness
- [x] #1918 Documentation split v2/v3
- [x] Start creating beta V3 sites
- [x] Switch to all new SiteNow sites being created using V3

joewhitsitt:
A few thoughts. I am okay with transitioning to all new SiteNow sites using Layout Builder sooner than later with the "beta" label, letting customers know we are still working on it and the documentation/training that needs to be created. This will allow big sites to start putting content into the correct bucket. The feature parity might not yet be there, but we will communicate those changes as they come.

I am warming up to the idea of calling layouts V3. This will give us enough distinction when creating/organizing docs and support. We have console/meta info that would need to be updated based on the two being used.

I don't see automated migration working 100%. It is not a one-to-one migration and I think that any resources going towards automated migration should go towards getting people off D7.

Instead or at least as a primary step, and thanks to a single theme, I think we should investigate a way to have both on at once, either:

- All new basic pages start as layout pages, customers can transition over time.
- Find a way to have both on at once for existing paragraph pages, a guide for transitioning content to layout/preview and have a webmaster one-time switch to turn the page into a layout page so they can maintain menu placement, etc.

Ultimately, it isn't the most ideal to have two methods, but we had several methods in D7. As long as we prioritize and innovate for one method, with the work we have already done to create splits and separate logic it should be fairly straight forward to support and keep the other one running.

Automated migration D8/9 to D8/9 will always be the thing we never get around to.

mjoneill:
Please keep the conversation going. 

I added a meeting to our calendars to discuss on Wed with the team.  

I’d like to see if there’s a path forward we can all get behind and that allows us to: 

1.	Tell clients just starting redesign work what type of site they should build on (paragraphs/layout builder/Gutenberg)
2.	Tell past clients who already migrated to D8 how they will be brought forward
3.	Tell ourselves that we’re making a good decision with what we know
4.	Start decomposing the vision into actual work (including things we’ll need more clarity on)
5.	Whatever else we make visible…

joewhitsitt:
[...]

PR #2028, opened 2020-08-06
New collections block

### LINKED ISSUE #1555: As a site builder, I can add collections of items via Layout Builder

cory-skeers:
Collections block for adding accordions, dls, or tabbed content.

Resolves #1555, relates to #1779 

## Testing
Sync a site, config import, etc.
Add Collections to a page, and try out the three options (dl, accordion, tab). Check 'em out, break 'em, and such.

## Yet to do / known bugs
- [x] Functionality and theming for tabs, or disable tabs as an option 030dbf5
- [x] Default view mode selection error 030dbf5
- [x] h3-h6 accordion theming (uids)
- [x] heading size is available as an option, but unused for `<dl>`
- [x] refactor template structure to work off child view modes, and get away from the current theme preprocesses aside from the tab (resolved via cc6e8f8 and 9a48c39)

### Note re: heading size
Disabled heading size field for now. For accordions, it then defaults to h2. As styling for h3-h6 gets added later, this field can be re-enabled to offer more flexibility for setting hierarchy. Its use with `<dl>` can be addressed at that time. Tabs are currently unstyled and therefore disabled, so they are not affected by this. The collection heading size field is the same as used for the uiowa_slider, as well, so there are no field updates to worry about--it can just be re-enabled on the form when it's ready.


cory-skeers:
~Something appears to have broke from abee575 or 6ae23db, where the config form isn't loading.~

Never mind--was a local error.

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -314,6 +312,63 @@ function uids_base_preprocess_block(&$variables) {
...
+function uids_base_preprocess_block_content__uiowa_collection__accordion(&$variables) {
+  $items = $variables['elements']['field_uiowa_collection_items']['#items'];
+  foreach ($items as $delta => $item) {
+    $id = $item->get('target_id')->getString();
+    $paragraph = \Drupal::service('entity_type.manager')
+      ->getStorage('paragraph')
+      ->load($id);
+    if ($paragraph->get('field_collection_item_view_mode') !== 'accordion') {
+      $paragraph->set('field_collection_item_view_mode', 'accordion')
+        ->save();
```
pyrello: It looks like you are saving the paragraph in theme preprocess function, which seems like it would run quite a bit more than you would want it to.

pyrello:
@cory-skeers The one other thing that I wanted to do with this is to remove the `default` view mode from the options when the user is configuring the block. Nice work!

cory-skeers:
~Note: For this PR we can remove view_modes from the form to effectively disable them to allow for merging prior to all collections being ready (i.e. turn off tabs until they're styled).~

Addressed in 030dbf5

cory-skeers:
~View mode radio is currently set as required, but no default option is set (because the true 'default' is removed). Brings up an error when trying to add a paragraph prior to setting the view mode.~

image

This is addressed in 030dbf5

cory-skeers:
Heading size correctly applies to the accordion headings, but h3-h6 levels don't receive the full theming that an h2 receives.

image

docroot/themes/custom/uids_base/uids_base.theme
cory-skeers: Initial thought was to have the check before so it would only run/save once when the block view_mode is set (or updated). As long as the page is being edited through the UI, this save should run as part of the rendering process of the layout editor, keeping the main saving cost upfront on the editor.

I think you could get away with not saving at all, but setting the field each time--it would save us a save (ha) that's more costly, but need to run and set the variable in every preprocess. Not sure which is more efficient in the long run, especially getting into pages like https://medcom.uiowa.edu/theloop/covid-19-clinical-information (not ours, not Drupal, but good example of how many paragraphs we might end up with on a page).

Ideally, I'd rather this live with the block creation/update form submission, but haven't been able to nail that down--the block creation happens in a process that gets called after the submission itself that I haven't found a good way to hook into.

cory-skeers: Child view modes and a bit of refactoring of how the templates were laid out worked like a charm! 👍  Cleaned things up, and was able to then remove all these theme preprocess hooks but for the one handling tab labels.

joewhitsitt:
Just pulled this down for testing on a default site install. I see the Collection block but the only config is to add block which it doesn't seem to do. Is this correct? Both as webmaster and admin. In a one-column section.

cory-skeers:
@joewhitsitt Yeah, that's definitely not right.

Think it was the paragraphs-related config for the collection entities ending up in the sitenow_v2 config split. Without the split enabled, no paragraphs, so no paragraphs reference field...and so on. Got the uiowa_collection related configs out of the feature split and into the default.

joewhitsitt (approved):
Overall this looks nice! I added block_content_permissions to the default config so that webmasters and editors can work with it.

I appreciate the hold/disable on the dynamic headings and tabs features.

My only gripe is that by going with inline entity form, there doesn't seem to be a preview view mode that similar to what can be done with the paragraphs field widget. Without it, on a saved block, I can't see the content of any of the items unless I click edit. Imagining a list longer than a couple and a small edit to change order involves scanning the layout preview and then counting the field collection items to reorder. My understanding is that paragraphs is not block settings tray "ready" yet and is rough around the edges visually.

pyrello:
@joewhitsitt Did you test the paragraph style widget? When I tested different options, that type of widget didn't seem like it was workable. That is why I went with IEF.

pyrello:
@cory-skeers I'd like to test this before it gets merged, if you don't mind.

cory-skeers:
@pyrello Absolutely! Would appreciate anyone who's interested to get in there.

Re: the paragraphs widget, when I tested, it was workable, but required some additional styling. IEF works up until the layout gets saved, and it replaces the "orphaned block..." label with "collection items". Could hook in and update that to match the headline to make organization easier, but if you're going to do something like that already, I think it's debatable whether to do that or put the work into the paragraphs widget.

joewhitsitt:
>Did you test the paragraph style widget?

I did not. Just going off my experience with node/paragraphs. I approved with IEF, just seeing the difference in ux right now. IEF can be tweaked to offer a preview of sorts (see article authors on article).

https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_people/sitenow_people.module#L102

cory-skeers:
Thanks for the reference, @joewhitsitt!

I've updated it to use the collection item headline in the IEF, that should make it easy to keep organized. They should be unique, and enough to clarify what the content below is, anyway...

joewhitsitt:
Nice!

pyrello (approved):
This looks good as a starting point for these items. Nice work!

PR #2032, opened 2020-08-07
Create articles, people and events Layout Builder blocks.

### LINKED ISSUE #1789: Create Articles/People functionality in layout builder using custom block types

### LINKED ISSUE #2071: Create custom events block for use in Layout Builder
There are quite a few different options here. Considering there is already an event block built out, it could be confusing to have a separate events block for content hub integration if we plan on making the static event block generally available. Additionally, there has been talk of sites needing an event content type which further complicates things.

# Potential options
1. Create a separate event list block that pulls events from content hub.
  a. Keep the static event block in a split.
  b. Make the static event block generally available.
2. Integrate the event list block into the static block and allow the user to choose.
  a. This could also include a single event reference from content hub which would make for three choices.
3. An event content type could present a fourth option if integrated into this all-encompassing event block.

MVP, and probably the path of least resistance, is probably option 1 while keeping the static block in a spilt. However, I think the static event option is nice to have and offers more flexibility - we will eventually have requests for one-off events that are not on content hub for some reason. Thats why I prefer option 2 but have concerns of future integration of option 3, mainly over how intuitive and usable the interface would be. It could be done though with the states API and is at least accurately representing the myriad ways of creating/referencing events on campus.

joewhitsitt:
Resolves #1789, resolves #2071 

Introduces three new custom block types (articles, people, events) that are basically ports of the paragraph versions.

Created https://github.com/uiowa/uiowa/issues/2058 to address alternative layout/style options

## To Do

- [x] Fix caching issues with collection heading
- [x] Remove border-top, consistent top margin
- [x] Recreate and share fields between articles/people and events block
- [x] Add collection heading to events block
- [x] Add LB styles to events block
- [x] Add event location to teaser and style appropriately for hide-details LB style.
- [ ] ~~Update event styles to match existing in UIDS.~~
- [x] Update sitenow_events instance date formats.

## To Test

- Using a v3 site with articles, people and events (more than one of each), pull/sync. 
  - The default site can reinstalled locally as v3 by running `blt drupal:install`.
- Compile frontend SASS via ` blt frontend` or `yarn workspace uids_base gulp`.
- As a webmaster ensure there are three new blocks available (Articles, People, Events). Configure/place as you like.
- Within the layout interface and the rendered content (depending on content available) you should see a list of content and your configuration should match the output (hiding images, displaying a certain number, displaying a pager if more than number of items, filtered by tag, more link with custom path, etc.).
- Set heading field and make sure the teaser markup includes the correct heading tag (not the h5 class).
- Out of scope: there is intentionally no grid, masonry or other layout options included with this PR.
- For events, test out a single event page by changing the link path config at Admin -> Events. (drush cr afterwards).

I have not built this with collegiate split in mind as I am sure there are at least permission updates involved.


mark-bennett-uiowa:
Looks pretty good to me currently. Had to add a link to the /news view page in the menu structure, but the view was working. A few things I noticed:
**(0)** People is still very basic, not much to test here.
**(1)** Is there a reason behind Draft/Review/Published states for Authors? If the hope here is to attach articles to person entries existing on the site, what about people from outside the Dept/Unit/Org that wrote the article? _Are you going to add Richard Lewis from OSC to the Biology site so that his name appears as having written the article? If we leave this his entry as Draft, he wont show up as the article author as well as in the list of People on the Biology site. This has been a topic of discussion in the past, and I feel attaching an article author to a person entry on the site is not the path to go down. Having a text field (or unlimited entry field) for authors doesnt link to the person node on the site, but you dont have to add them to a site they arent directly part of._
Screen Shot 2020-08-13 at 10 11 07 AM
**(2)** In the News view page, the source link is attached to the Organization text; however, in the article itself, the two fields are kept separate. Should the news view keep these separate also?
Screen Shot 2020-08-13 at 10 14 12 AM
Screen Shot 2020-08-13 at 10 14 23 AM
**(3)** News feed contains twig theming:
```
<!-- THEME DEBUG -->
<!-- THEME HOOK: 'views_view_rss' -->
<!-- BEGIN OUTPUT from 'core/themes/classy/templates/views/views-view-rss.html.twig' -->
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0" xml:base="https://americanstudies.local.drupal.uiowa.edu/" xmlns:dc="http://purl.org/dc/elements/1.1/">
  <channel>
    <title>News</title>
    <link>https://americanstudies.local.drupal.uiowa.edu/</link>
    <description></description>
    <language>en</language>
    
    

<!-- THEME DEBUG -->
<!-- THEME HOOK: 'views_view_row_rss' -->
<!-- BEGIN OUTPUT from 'core/themes/classy/templates/views/views-view-row-rss.html.twig' -->
<item>
  <title>

<!-- THEME DEBUG -->
<!-- THEME HOOK: 'views_view_field' -->
<!-- BEGIN OUTPUT from 'core/themes/stable/templates/views/views-view-field.html.twig' -->
Test News Article Two
<!-- END OUTPUT from 'core/themes/stable/templates/views/views-view-field.html.twig' -->

```

mark-bennett-uiowa:
Here is the articles block in a section:
Screen Shot 2020-08-13 at 1 14 15 PM
The Test Article ONE links to the site article itself. Test Article TWO has the "Link directly to source" checkbox checked. The title of the Test Article TWO does go properly to the source link, but the display of the Written by: field(s) has changed and no longer has a link there.

joewhitsitt:
@mark-bennett-uiowa so, unfortunately none of the comments you made pertain to the scope of this PR. Articles and People already exist in the default config. This just adds the custom block types to display views of articles/people in layout builder. A LOT of content already exists with this structure. Follow-up issues could be created or extended through custom work to meet additional requirements or to fork and go solo. This is SiteNow trying to reach feature parity again with a new page building tool.

To answer your remaining questions:

The three default content types have moderation so published has to be set on the author inline editing form. We could change this to set as published by default and hide the field but that should be opened as a separate issue. I want to say there the use-case i had for the moderation field was to allow a draft article and a draft person to be created instead of forcing a person to be published right away before the article is published.

The hide on lists checkbox was created with the outside article contributor in mind. Someone that is helping write an article but should not be in the /people list.

/news and /people are view pages that exist by default and can be disabled through configuration. The help text on the blocks explains that a bit without adding more validation.

I will have to investigate the source, link directly to source, organization combo and how that might have changed with the uids_base adoption since they are all optional. This is technically out of scope of this PR though.

I think the feed contains twig debug because it is on locally for everything. It is not on in PROD: https://gis.sites.uiowa.edu/articles/feed

joewhitsitt:
Clarifiication on the article source stuff. The help text mentions the behavior a bit.

Screen Shot 2020-08-14 at 11 27 26 AM

The article teaser titles, if source link provided and link directly to source is checked changes the link to the source link instead of the node. The source link disappears because it is a redundant link in terms of a11y. I think this is working as designed. The full node has more space and so there is a different behavior there as it relates to the organization value.

cory-skeers (approved):
Reviewed and tested out on uiowa.edu using the default Keep Moving Forward! and Herky default content, and some new test articles/people.

Things look to be working well, and matches v2 implementation. Can get some strange hierarchies in sections with its `<h3>` tags, but that's a bigger question, and not a new one.

joewhitsitt:
Based on chats it sounds like this block may need the heading size treatment as well.

joewhitsitt:
Added heading treatment similar to collections/slider but for a freakin block > view > node view mode.

mark-bennett-uiowa:
@joewhitsitt This looks good for Articles now. I was seeing an Articles already enabled and then Sitenow Articles as an additional module that was available for enabling. It appears that has been fixed, and is working as needed.

I did notice that a top border exists on the first list item where one isnt needed. Can this be removed for the first list item?

Screen Shot 2020-08-20 at 3 22 29 PM

joewhitsitt:
@bspeare what are your thoughts on the top border from @mark-bennett-uiowa 's comment: https://github.com/uiowa/uiowa/pull/2032#issuecomment-677884167

bspeare:
@joewhitsitt I think we were using a style that removed that in v2.   I can pull this down and try to add that in if it helps.

joewhitsitt:
I am seeing it in v2 as well: https://gis.sites.uiowa.edu/articles

bspeare:
Good to see an example.  I think it's fine to leave as is for now.  I'm fine with removing it.  @mark-bennett-uiowa do you want to add a fix to this PR that removes that for the first child?

dnkemper:
Thoroughly tested and talked with Joe about some things yesterday (THANK YOU, JOE!).  Everything seems to be working great. The only issue I see is the lack of padding on L & R sides of content. It seems to be using the card_body class... if that's the case, does this happen to other blocks as well? 
Screen Shot 2020-08-21 at 1 14 56 PM

dnkemper:
UPDATE: Yes it is affecting other blocks.
Screen Shot 2020-08-21 at 1 20 47 PM

joewhitsitt:
@dnkemper does https://github.com/uiowa/uiowa/pull/2067 fix the card padding issue you experienced? since this just references the card component, I am not sure if display padding with bg colors is in-scope for this PR. Also, https://github.com/uiowa/uiowa/pull/2069 can be follow-up to this PR to address background color issues, if any.

dnkemper:
@joewhitsitt I'll scope it out and if it doesn't, I'll add a comment to #2069 if that works?

joewhitsitt:
#2067 for padding fixes
#2069 for background color contrast corrections

richardbporter:
>Check to see if I am putting the code/styles in the right place!

I think the modules are the right place for the code since it is running both paragraph and block implementations, at least in the short term. Putting styles in the is fine too.

mark-bennett-uiowa:
@joewhitsitt Just tested this out on American Studies (which may be causing some of the issues below - should I test this on a Sitenow v3 site instead?). Looks good, as the following things seem to work fine:

1. Hide descriptions, Hide details and Hide images all work properly for Articles and People blocks.
2. Display a certain number of entries works for Articles _but not for People_ (see below).
3. Display a pager works for Articles, _unsure if it works for People due to issue with #2 above_.
4. Did not test filter by tag, if someone else was able to test this.
5. More link with custom path works on both Articles and People.
6. No other errors that I can see, feel comfortable approving this even in its current state if this needed to get out the door.

Here are a few things I noticed though:
(1) The border above the first item looks even more out of place here (since its missing a heading).
(2) It doesnt seem to display any headings above the list(s). Am I missing a field here?
Screen Shot 2020-08-26 at 11 03 30 AM
(3) The pager system doesnt seem to be working on People when selected in the LB block settings (both of these blocks are set to show 2 and add pager):
Screen Shot 2020-08-26 at 11 18 44 AM
(4) The heading setting did not seem to work on either, here is the inspection:
Screen Shot 2020-08-26 at 11 43 24 AM

joewhitsitt:
@mark-bennett-uiowa i will look into these. thanks for the review.

@pyrello brought up some interesting points about maybe doing a block plugin instead or at least reusing the fields between the two.

joewhitsitt:
Just refactored to share fields and consolidate _some_ code.

Kind of needs testing again to make sure the field changes line up with the logic.

If you previously pulled, this might need a database drop or it will complain about missing field config

mark-bennett-uiowa:
> If you previously pulled, this might need a database drop or it will complain about missing field config

I definitely needed to do this @joewhitsitt . Retested, at the current state. Still works as my above testing, but it did fix the (3) The pager system doesnt seem to be working on People when selected in the LB block settings issue.

Im assuming that the two unchecked checkboxes mean they are still in progress?

joewhitsitt:
Yes, those checkboxes are needs to do. Your question about the parent field will be solved in a global way with https://github.com/uiowa/uiowa/issues/2048

joewhitsitt:
Here is a slack thread on the caching issue: https://iowaweb.slack.com/archives/C0164FNN1PV/p1598627548125900

richardbporter:
Heading size seems to work as expected without issue on events block. I'm guessing the cache problem is at the views level.

joewhitsitt:
i will try testing with views cache on those turned off.

Update: nope. with view cache turned off on those view displays, i am still seeing the change reflected in layout preview, but not on layout save.

richardbporter:
I can look at it.

Exactly what fields are we looking to share? Is there a preference on naming?

joewhitsitt:
It seems these are all collections and would match the existing shared fields.

- number of items
- more link
- more path

Does anything else match up?

joewhitsitt:
While testing you might encounter https://github.com/uiowa/uiowa/issues/2109 (multiple arguments) which I found out the first time I built these, but left as-is thinking it would get attention soon... :(

richardbporter:
I'm not having any luck. We solved something very similar in https://github.com/uiowa/uiowa/pull/1947.

richardbporter:
If we can't figure out the caching issue, one potential workaround would be to invalidate the render cache on node submit if the page contains an article/people view LB block. Ideally, we'd only clear the CID for the page in question.

richardbporter:
@joewhitsitt Can you review and test cfa7da8 with the cache issue for articles? It worked for me locally even with view caching re-enabled.

joewhitsitt:
Nice idea @richardbporter! And yes, this worked for me. Logged in and logged out. Do you want to extend this same solution to people while I work on the shared fields?

richardbporter:
Sure - I'm going to merge this into dev to and test it with Varnish as well.

joewhitsitt:
Just pushed another breaking change. Renamed certain fields to share with non-views "collection" type blocks (e.g. events).

joewhitsitt:
@richardbporter can you elaborate on the remaining events tasks?

Are the event styles supposed to match "event" like the homepage or remain similar to articles/people for MVP?

I'd argue that the date shouldn't be hidden with the details option. What separates article and event from a page is their relationship to dates, therefore I think they should always be displayed.

Looks like the events date format is the same as article:
Screen Shot 2020-08-31 at 2 30 23 PM

richardbporter:
>Are the event styles supposed to match "event" like the homepage or remain similar to articles/people for MVP?

I don't know. It seems odd that we would have two different styles for events. I think they should at least look similar in the short term. I was looking at this though and was unsure how to proceed.

> I'd argue that the date shouldn't be hidden with the details option. What separates article and event from a page is their relationship to dates, therefore I think they should always be displayed.

I'm ok with that. I think the location should show on the teaser though.

>Looks like the events date format is the same as article.

I noticed some wonkiness with all day events on Friday. I'll have to find some complicated dates to test this out. I also think `sitenow_events_format_instance` could be improved and possibly move to the preprocess for each theme function.

richardbporter:
I'm going to stash my UIDS styles for a follow up PR. I'm working on fixing a date instance bug. All that is left is:
>Add event location to teaser and style appropriately for hide-details LB style.

config/default/block_content.type.uiowa_articles.yml
```diff
@@ -0,0 +1,8 @@
+uuid: cbea77a6-0e50-488a-8d6c-c228b11dc1b9
+langcode: en
+status: true
+dependencies: {  }
```
richardbporter: I guessing this has a hard dependency on sitenow_articles, as does sitenow_people.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -172,17 +172,33 @@ function _layout_builder_custom_process_element(array $element, FormStateInterfa
...
-    }
+  switch ($block->getPluginId()) {
+    case 'inline_block:uiowa_text_area':
+      if (isset($element['field_uiowa_text_area'])) {
+        $element['field_uiowa_text_area']['widget'][0]['#format'] = 'filtered_html';
+      }
+      break;
+
+    case 'inline_block:uiowa_articles':
+    case 'inline_block:uiowa_people':
```
richardbporter, line 204: We should make this work for events too.

richardbporter (changes requested):

docroot/modules/custom/sitenow_events/css/single-event.css
```diff
@@ -6,6 +6,10 @@
   }
 }
 
+.sitenow-event .event-field {
```
richardbporter, line 9: Just noticed that `.sitenow-event` seems like the wrong selector for an event page because it doesn't exist. However, changing it to `.sitenow-event-single`, which exists, really borks the display.

richardbporter:
The styling of a single event page looks off to me. I made a comment about the selectors in single-event.css. Since HR is still using this library, I think it needs to stay in the module. Long term, it should probably move to the UIDS theme.

joewhitsitt:
Based on the slack conversation, the virtual link was cleaned up and an icon was added to show dates to align better with the location. I didn't want to get into figuring out a down arrow/up arrow toggle so I give you the ellipsis!

Screen Shot 2020-09-02 at 12 45 47 PM

Tested much of the events portion of this PR. Everything seems to be working as expected, including keywords which I just hotfixed for the version currently in main.

richardbporter:
Thanks. The only thing I was unsure of was the type on the single event pages - line height, font, etc. - it doesn't look on-brand to me but could be wrong.

joewhitsitt:
uids_base overrides can help with that, but I _just_ checked out another branch and would have to revisit later if nobody else is up to it.

To be on brand for now and revisit later, might I suggest mimicking the article full node styles:

Screen Shot 2020-09-02 at 1 15 11 PM

mark-bennett-uiowa:
> Based on the slack conversation, the virtual link was cleaned up and an icon was added to show dates to align better with the location. I didn't want to get into figuring out a down arrow/up arrow toggle so I give you the ellipsis!
> 
> Screen Shot 2020-09-02 at 12 45 47 PM
> 
> Tested much of the events portion of this PR. Everything seems to be working as expected, including keywords which I just hotfixed for the version currently in main.

This is HOT btw, LOVE the shaded past dates. Well done.

mark-bennett-uiowa:
> uids_base overrides can help with that, but I _just_ checked out another branch and would have to revisit later if nobody else is up to it.
> 
> To be on brand for now and revisit later, might I suggest mimicking the article full node styles:
> 
> Screen Shot 2020-09-02 at 1 15 11 PM

I vote to mimic what is there for articles.

pyrello:
Pulling this down to test again.

config/default/block_content.type.uiowa_articles.yml
pyrello: Added.

docroot/modules/custom/sitenow_events/css/single-event.css
pyrello: It seems like the event page looks pretty good now. We may want to have @bspeare take a look and refine it a bit when he gets back, but this looks like a great starting point!

pyrello (approved):
This looks good to me. There may be a little clean-up to do, but I think the data structure is sound.

richardbporter (approved):

joewhitsitt:
Ah crap. I was working on the style changes and also noticed something broke HR's events single... https://hr.local.drupal.uiowa.edu/event/54376/0

pyrello:
Dang! Sorry @joewhitsitt. What can I do to help?

PR #2034, opened 2020-08-07
Require cards be rendered horizontaly if placed in a onecol layout.

### LINKED ISSUE #1987: Allow card component to display in onecol layout as horizontal variant only
# Problem

The regular card component is not meant to render within a onecol layout.  However, the horizontal variant does render well within the onecol layout: https://uids.brand.uiowa.edu/components/detail/card--horizontal.html. 

# Proposed Solution

Force the `horizontal` style variant if card is placed within a onecol layout.

richardbporter:
Resolves #1987. Some PHPCS as well.

### Testing
- Try adding a card to a onecol layout.
  - See the form returns an error.
- Try adding a card to a non-onecol layout.
  - See the form styles can be changed.

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
+++ b/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
@@ -111,7 +111,7 @@ function _layout_builder_custom_block_validate(array &$form, FormStateInterface 
   $component = $form_state->getFormObject()->getCurrentComponent();
   $block = $component->getPlugin();
 
-  // Disable overriding layout styles if this is onecol card.
+  // Require horizontal style if this is card is in a one column layout.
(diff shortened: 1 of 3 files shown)


bspeare:
@richardbporter Both tests worked for me!  I can live without being able to add the border option to a one column card.

richardbporter:
I ran into an issue where the border style seems to be set even though it is not visible in the style field. I'm going to try the form validation method. 

Screenshot this for comparison later.
Screen Shot 2020-08-10 at 9 51 51 AM

richardbporter:
Got the custom validation working. If the horizontal style is not set, the form returns an error on submission. This makes it possible to use the border style or any future styles.

Screen Shot 2020-08-10 at 10 36 41 AM

bspeare (approved):
It works great! I tested it on https://uiowa.local.drupal.uiowa.edu/ and I was able to use both override style options.   Thanks!

PR #2041, opened 2020-08-14
Minimal HTML default text format

### LINKED ISSUE #2038: Minimal HTML as default for components over Filtered HTML
Part of #1778 

Change component text areas to use minimal_html by default or limit toolbar buttons to reduce kitchen sink.

~Without custom code, this is an all-or-nothing (ie setting it as default for all formatted text fields, including the Text Area block).~ Plan is to set Minimal HTML as the site-wide default, and hook_alter to set Text Area block's default to Filtered HTML.

cory-skeers:
Resolves #2038, relates to #1771 

Sets Minimal HTML over Filtered HTML as the default format for formatted text fields site-wide. This default is overridden for new Text Area blocks, which are set to Filtered HTML.

## Testing

- Sync a site/import config changes.
- Check that no existing content blocks are broken (specifically, blocks which had used Filtered HTML, such as the Hero block on uiowa.edu)--no existing block text format should change.
- Create new blocks. 
  - Text format for all but Text Area blocks should be Minimal HTML. 
  - Blocks such as Hero which allow Filtered should default to Minimal HTML, but allow for changing to Filtered HTML.
  - Change a Hero or other block to use Filtered HTML and save--check that it saves as Filtered HTML.
- Text Area blocks should default to Filtered HTML.


richardbporter (approved):
This seems to work as advertised.

PR #2044, opened 2020-08-14
Initialize hhpatuiowa.org.uiowa.edu multisite on uiowa02

joewhitsitt:


briand44 (approved):
Don't love the url but it looks like hostmaster has already approved it so I will approve this.

PR #2045, opened 2020-08-17
Enable layout-builder-operation-link contrib module.

cory-skeers:
Relates to #1778.

Enables https://www.drupal.org/project/layout_builder_operation_link, which adds an option to the Content admin page to go directly into editing a page's layout (similar to how you can currently go directly into editing the page).

## Test

- Pull, composer install, sync.
- Go to admin/content.
- Check that Basic pages have the Layout operation available, and that it works correctly.
- Check that Article and Person content do not have the Layout operation.
- Repeat masquerading with different roles (layout_builder_operation_link doesn't have permission separate from ability to access lb editor, so if the above works, this should too)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -105,6 +105,7 @@
         "drupal/imagecache_external": "^1.0",
         "drupal/imagemagick": "^3.0",
         "drupal/inline_entity_form": "^1.0.0-rc2",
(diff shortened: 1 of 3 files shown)


richardbporter (approved):
This is slick. FYI only webmasters seem to have layout access right now.

PR #2062, opened 2020-08-20
Update node body field to default to filtered_html.

### LINKED ISSUE #2052: Node body field should default to Filtered HTML in SNV3
Followup to #2041.

Minimal HTML was set as the default format for all text fields, with an exception made to set Text Area block textfields to Filtered HTML. Similar exception should be made for node body fields.

cory-skeers:
Resolves #2052 -- Sets node body format to filtered_html instead of the site-wide default minimal_html.

# Problem
As a result of #2041, site-wide text format default was changed to minimal_html, with an exception made for Text Area blocks to receive filtered_html as a default. As basic page body fields can be used to quickly add simple pages, this field should also default to filtered_html, and needs an exception made.

# Test
Sync.
Add a new basic page.
Check that body field format defaults to filtered_html.

## Note
General logic will work, but some of this will need to be updated in order to merge with updates in #1986.

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
+++ b/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
@@ -16,6 +16,7 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
     'layout_builder_update_block',
     'layout_builder_add_block',
     'layout_builder_configure_section',
+    'node_page_form',
   ];
 
   // Only continue if this is one of the above forms.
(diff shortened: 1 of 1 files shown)


joewhitsitt:
Did this introduce the override styles checkbox on the node form?

Screen Shot 2020-08-21 at 8 59 53 AM

cory-skeers:
@joewhitsitt It did--it's getting the 'else' form changes. Once #1986 gets merged, I'll update it.

Edit: Set the PR to draft until this is updated.

cory-skeers:
Got main merged in with the refactor, and moved it into its own hook to help avoid that type of issue in the future.

joewhitsitt (approved):

PR #2064, opened 2020-08-21
Temporarily get some v2 default content working.

### LINKED ISSUE #2023: New site provisions not receiving default paragraphs content
Following the SiteNow v2 split, new site provisions are not receiving their default paragraphs content. Pages, media, etc. are there--just the sections/paragraphs for the Home and About Us pages need to be reimplemented.

richardbporter:
Resolves #2023, kinda.

If we can live without v3 sites getting a home and about page in the short term, this gets some v2 default content back. Notably missing is the large image on the front page. I don't want to spend much more time on this but if someone more familiar with default content wants to jump in, go for it.

TLDR; It doesn't seem like the default node content can live both in the profile and a module.

### Testing
v3
- `blt drupal:install`
- Look at default.local.drupal.uiowa.edu.

v2
- `blt drupal:install --no-interaction && drush cset config_split.config_split.sitenow_v2 status true -y && drush cr && drush cim -y`
- Look at default.local.drupal.uiowa.edu.


joewhitsitt:
v3: basic site settings still has node/1 as the front page  and there is a menu link for node/2, both of which doesn't exist. this is my only concern with this change.

v2: also missing are the getting started cards which i think are the most important to getting started. is there are reason why these pieces of content were lost in the transfer?

i am almost in favor of keeping node/1 and node/2 established as basic pages for both versions and altering the site created email to note the issue than this middle situation. then we can focus on a new round of default content in the profile for v3.

richardbporter:
> i am almost in favor of keeping node/1 and node/2 established as basic pages for both versions and altering the site created email to note the issue than this middle situation. then we can focus on a new round of default content in the profile for v3.

Yea me too.

PR #2066, opened 2020-08-21
Adjust spacing between featured image and content

### LINKED ISSUE #2063: Increase margin for basic pages after feature image
Before the feature image for basic pages change was added a few days ago, the distance between the bottom of the section/page title hero block and the top of the main content was 3rem. Basic pages with a page title hero were created in a new section above the header, and the page title hero block was added to this section and styled in LB. The header had the title and breadcrumbs removed so that they would only appear in the new section with the page title hero block. It appears that this distance was due to the 3rem margin-bottom of the "empty" header section:
Screen Shot 2020-07-22 at 1 09 53 PM
When these pages were switched over to use the new feature image part of the header section, and the previously created section + page title hero blocks were removed, the title and breadcrumb were returned to the header, but the distance is now 1.25rem and due to the margin-top of the class="block block-layout-builder block-field-blocknodepagebody" div.
Screen Shot 2020-07-22 at 1 07 53 PM
Could this distance be increased if the feature image is used on a page?

cory-skeers:
Resolves #2063 

@bspeare This looked to me to need to be added to each of the one, two, three, four, and sidebar layouts, and to work as intended with it, but would love a check that I'm not blowing up spacing elsewhere.


bspeare:
@cory-skeers I adjusted the spacing and split the margins to display at a different size if there is no background image.   I still need to test this on more sites.   @mark-bennett-uiowa can you test this on your sites?

bspeare:
I'm not sure if this is worth adding to 2-col, 3-col and 4-col layouts since it was meant to provided default spacing for node body copy for general page use.   Maybe it should just be in use for 1-col and the sidebar menu layout?  Perhaps if users are using the multi-column layouts they should add padding through the section styles?

Screen Shot 2020-08-21 at 12 32 41 PM

cory-skeers:
@bspeare That makes sense to me--only apply it to node body copy, leave the others up to the user. I misunderstood the original issue.

bspeare:
@cory-skeers I think it was a good idea!  It was only after testing did I notice that there could be further spacing issues.

GaryRidgway (approved):
I have checked this and confirmed that on one column layouts there is a 2rem margin on top of the body field.

PR #2070, opened 2020-08-22
Initial Behat test setup.

richardbporter:
FYI - I did most of this off-hours outside of sprint scope.

Because of the amount of effort I put into our initial functional tests, I tried to resurrect them but was unsuccessful. It is difficult and time consuming to install our profile + config for _each_ functional test. Behat can run on the already-installed default site locally and in CI which is a better fit and more efficient. This is just a really simple PR to get Behat setup and running with a basic test.

The Behat Javascript driver does not work with self-signed certificates. There are two potential workarounds:
1. Don't redirect HTTP requests for local.drupal.uiowa.edu and override the protocol used in tests to http.
2. Install an SSL certificate generated from our CA.
  - http://docs.drupalvm.com/en/latest/configurations/webservers-apache/#enable-ssl-support-with-apache

I went with option 1 in this PR.

### Testing
- Outside the box:
  - `git pull`
  - `vagrant provision`
- Inside the box, i.e. `vagrant ssh`
  - Follow the directions in tests/behat/README.md


joewhitsitt (approved):
Looks cool. Thanks for the demo. Not tested

PR #2075, opened 2020-08-24
Feature: Banner block refactor

### LINKED ISSUE #1928: Refactor hero component to have hero and banner styles
# Problem

The Hero and Banner component share styles that can be refactored to offer one inline block component that will reduce the number of custom block options available. 

# Proposed Solution

Once https://github.com/uiowa/uids/pull/268 is completed, the Banner inline block component and the Hero component will need to be adjusted to offer more variation while being a little more prescriptive. 

- Text area becomes `minimal_html` instead of `filtered_html`. 
- Link field will need to be added

The Hero components featured on the home page will need to be set as custom styles and will not be featured as options within the Hero component.

bspeare:
Initial work to resolve https://github.com/uiowa/uiowa/issues/1928. 

# How to test banner block

```
blt ds --site=uiowa.edu
rm -rf ./node_modules
yarn cache clean
yarn install
drush @home.local cr
```

**Basic functionality**

1. Create a new banner block without using any of the "override options"
2. Create a banner block without a link
3. Crate a banner block without link text, is the banner still clickable?
4. Use any number of combinations in the "override options" area, do they work as expected?
5. Change the "Background" option to a pattern or background color, does it work different combinations in the "override options" area?

**Usage without layouts**

6. Move banner block around from one, two, three, and four column layouts.  Heading sizes and background opacity should change if used within two, three, and four column layouts. 

**Responsive behavior**

7. Collapse banners in different layout columns to different viewport sizes and note where it breaks.   Do we want to set a character count on the `Excerpt` to limit mobile overflow issues?

## Testing update hooks

1. Sync a site with existing, or add a hero and banner block to a site.
2. Run update hooks (8033, 8034).
3. Check that existing banner blocks still exist, now overriding styles with those marked `(default)`.
4. Check that, in place of hero blocks, banner blocks are present with the same content.
5. Check that any existing link in the original hero is present and functional as a banner button link.
6. Check that the newly created banner blocks are editable, and that the attached image can be replaced.


bspeare:
It looks like we still need to adjust the permissions to allow the banners to be placed in all column layouts in `config/features/collegiate/core.entity_view_display.node.page.default.yml` and `/config/default/core.entity_view_display.node.page.default.yml`

config/default/layout_builder_styles.settings.yml
```diff
@@ -1,5 +1,5 @@
 multiselect: multiple
 form_type: multiple-select
-groups: "padding|Padding\r\nmargin|Margin\r\nsection_type|Section\r\nheading|Heading\r\nbackground|Background\r\nalignment|Alignment"
+groups: "padding|Padding\r\nmargin|Margin\r\nsection_type|Section\r\nheading|Heading\r\nbackground|Background\r\nalignment|Alignment\r\nbanner_type|Banner type\r\nbanner_height|Banner height\r\nheadline_type|Headline type"
```
pyrello: I am thinking about the following changes:
* Banner type => Content orientation (or something like that). My concern is that the options in question aren't going to be clear to the user.
* Banner height => Height
* Headline type => Headline options

docroot/themes/custom/uids_base/scss/components/banner.scss
```diff
@@ -2,26 +2,54 @@
...
+    background: $secondary;
+    border: 1px solid $secondary;
 
-  a {
-    font-weight: $font-weight-medium;
+    svg {
+      color: $primary;
+    }
   }
+}
```
pyrello: Are these general rules that could be moved into a non-banner related file?

bspeare: Yes, we should move these to the button component.

config/default/layout_builder_styles.settings.yml
bspeare: Those descriptions sound better!

bspeare:
**Still to be done:**

- [x] Set default style classes to banner container
- [x] Allow banner to have a link without a bttn displaying
- [ ] If gold background is used, override text highlight treatment?
- [ ] Test on Slider

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -1088,3 +1090,118 @@ function sitenow_update_8032() {
...
+            preg_match('|<a.*?href="(.*?)".*?>(.*?)</a>|', $matches[2], $link_matches);
+            // Set URI based on link being absolute or relative.
+            $uri = (substr($link_matches[1], 0, 4) === 'http') ?
+              $link_matches[1] : \Drupal::request()->getSchemeAndHttpHost() . $link_matches[1];
+            $link = [
+              'uri' => $uri,
+              'title' => strip_tags($link_matches[2]),
+              'options' => [],
+            ];
+          }
```
pyrello: Just noticing this. I wanted to double check with @bspeare that this would be the desired outcome of this transition, to have a link pulled out of the hero excerpt and added explicitly as a link button on the new banner.

bspeare: The hero on https://coronavirus.uiowa.edu/ would be a good use case.   Would it remove that link markup and move it to the link field?

pyrello:
@cory-skeers I just wanted to add a note to make sure it is documented somewhere that the two hero blocks that we are not wanting to convert are the "Start your story here" and the "Welcome home" blocks on https://uiowa.edu/.

docroot/profiles/custom/sitenow/sitenow.install
cory-skeers: As it's laid out here, it would @bspeare. If we allow for filtered_html in the banner, we can bring the content over directly, but to keep it limited to minimal will require some sort of fenagling like this.

cory-skeers:
Thanks @pyrello! If it's only those two specifically, that'll be easy to add a uuid check or something similar for the "welcome home" block. The "Start your story here" block is currently being skipped for its having a video attached.

docroot/profiles/custom/sitenow/sitenow.install
bspeare: Thanks.   I definitely don't want to move to filtered_html :).

pyrello:
@mark-bennett-uiowa Are you using a video hero anywhere?

@cory-skeers I'm not sure if this would be an issue, but I'm wondering if it will be necessary when updating media references to do anything else to trigger media usage updates. So that a media object doesn't report that it is being used by an entity that no longer exists.

mark-bennett-uiowa:
@pyrello No, at this point, none of my sites are using a video, though they are using the Hero block. On the front page of all the sites, there is a Hero block that just contains an image, not a video, if you needed to test something.

docroot/profiles/custom/sitenow/sitenow.install
cory-skeers: As a reference, here's how the update_hooked coronavirus.uiowa.edu banner currently appears (no styling handling is being done yet):
image

pyrello: @cory-skeers So, it seems like we'll need to detect whether the hero was using the "Hero: Left aligned" LB style and apply the appropriate banner style. Otherwise, we should maybe default to a center aligned (horizontally and vertically) variant?

bspeare: We probably have more banners that we are replacing that use the style that @cory-skeers has hooked up.   Perhaps we just manually update the covid home page?

bspeare: But we also don't want sites like https://americanstudies.prod.drupal.uiowa.edu/ to get this style.  Is there way to only add that style above to banners?

cory-skeers: I added a bit of structure to handle applying styles. The initial setup defaults to `banner_centered`, and then if the original hero image had `block_style_hero_left_aligned`, replaces it with `banner_centered_left`. We can change the default, or flesh it out with more logic if different handling options need to be applied.

pyrello: I'm cool with the approach. @bspeare and @mark-bennett-uiowa have a better sense of what the default should be so I'll defer to them.

pyrello:
@cory-skeers Thanks for taking over the update hook! It was more complicated than I originally assumed and I don't think I would have gotten it done before I left for vacation. Great work so far!

cory-skeers:
Getting some visibility on this--
There is currently a bug with the newly created banners from the update hook where a new image cannot be added (the current image can be removed, or the background switched to a non-image style).

I believe it is an issue to do with how the `layout_builder_styles_style` or another part of the config is being added to the block. When loading and inspecting the block info everything appears to be correct, but when editing the block in layout builder, the default styles are not appearing despite being set.

cory-skeers:
I believe the update hook functions correctly now, properly allowing edits to be made to migrated banners. As styles are finished up, we can look at changing how the styles get applied, but the update hook may be ready to go!

bspeare:
Did we want to restrict the Slider from appearing in any other layouts besides the one column in this pr?  I don't think it is responsive enough to work in any other layouts. 

Screen Shot 2020-09-30 at 1 03 44 PM

joewhitsitt:
pulling down to start testing

joewhitsitt:
Haven't tested update hooks on a non-uiowa.edu website yet, but these are my observations so far:

During the setup: `warning Error running install script for optional dependency: "/var/www/uiowa/node_modules/gifsicle: Command failed.`

- Background: image, doesn't require an image, changes to black next block edit.
- Image is removed when switch to another background option and back. Assumed it would still be there just not used on render.
- Personal opinion - "reversed" is confusing. As it relates to backgrounds, inverted or just calling it what it is (gold, white) makes more sense as a label.
- Override style options is checked by default. Not consistent with other blocks.
- Potential overflow on all height options in desktop since there currently isn't a limit in place.
- Banners seem to go white on mobile which then has contrast issues depending on the section background color.

bspeare:
@joewhitsitt Do you recall what section combination created this?

- Banners seem to go white on mobile which then has contrast issues depending on the section background color.

joewhitsitt:
I believe it was one-column black bg section. I customized and placed a new banner below "Be a Hawkeye" on uiowa.local

bspeare:
>     * Potential overflow on all height options in desktop since there currently isn't a limit in place.

Regarding this issue, I wonder if we want to set a character count to truncate the banner text and add a read more link to reveal the remaining text?

Screen Shot 2020-10-01 at 11 50 43 AM

joewhitsitt:
Forgive me, I think truncating makes the most sense, but playing devil's advocate for a moment:
- with a html field you have to decide how you are going to truncate. strip html first or count the the long `<a>` tags with absolute URLs. 
- revealing the text in the same spot will cause the same overflow and just be user triggered instead.
- different banner heights may warrant differ limits?

bspeare:
> Forgive me, I think truncating makes the most sense, but playing devil's advocate for a moment:
> 
>     * with a html field you have to decide how you are going to truncate. strip html first or count the the long `<a>` tags with absolute URLs.
> 
>     * revealing the text in the same spot will cause the same overflow and just be user triggered instead.
> 
>     * different banner heights may warrant differ limits?

Those are good questions.   

- The `<a>` is definitely something we want to take into account. 
- I think a general trim count to use for all banner heights would be a good place to start.  
- The "read more" could trigger a class change on the banner container to overflow the contents in a more elegant way.

bspeare:
> I believe it was one-column black bg section. I customized and placed a new banner below "Be a Hawkeye" on uiowa.local

@joewhitsitt  I'm not able to reproduce this if you have time to try and test this again.

bspeare:
>     * Personal opinion - "reversed" is confusing. As it relates to backgrounds, inverted or just calling it what it is (gold, white) makes more sense as a label.

I changed it to "inverted" instead of using the color name since we aren't specifying that on any of the other style options within this block.

joewhitsitt:
I will try to recreate the mobile-white on Mon along with finish testing the update hooks. Got really into 2284/2290 and had to deal with a user support issue.

joewhitsitt:
Locally added a hero (video) and banner image block to a site using the main branch on a collegiate split site. Checked out this branch and went through the theme/yarn steps. Ran updb/cim/cr and I still have a hero video block in place. Also, as a webmaster, I can still hero in the direct add dropdown.

Still able to re-create the white bg on mobile issue at the 768px breakpoint (section black, banner with gold bg)
out

bspeare:
@joewhitsitt The default mobile display for the banners is to go to black text on white background.   Do you think it should just retain the background image/pattern for mobile?

joewhitsitt:
I didn't know that was the expected behavior. I assumed it would be the same for mobile similar to a text area gray bg. If the behavior stays the same, there are a few contrast issues to fix on mobile like I originally reported.

>Banners seem to go white on mobile which then has contrast issues depending on the section background color.

- Mobile and Desktop combinations are having issues with Blockquote link, text, border and a few combos have a desktop issue with the banner title (e.g. Section Black/Banner brain pattern)

cory-skeers:
> Ran updb/cim/cr and I still have a hero video block in place.

@bspeare Currently the update hook that creates banners in place of hero blocks skips over heroes with a video attached. Is there anything we should be doing to handle those, or leave those be for now like the homepage homepage?

bspeare:
> I didn't know that was the expected behavior. I assumed it would be the same for mobile similar to a text area gray bg. If the behavior stays the same, there are a few contrast issues to fix on mobile like I originally reported.

I think that's a good point and it can be easily fixed in this issue.  I think the white background default was more of a holdover from the behavior of the page title hero.

bspeare:
> @bspeare Currently the update hook that creates banners in place of hero blocks skips over heroes with a video attached. Is there anything we should be doing to handle those, or leave those be for now like the homepage homepage?

I don't think we want to update the Hero block with video.   Do we need to remove the ability for Webmasters to add new Hero blocks on Collegiate?  Or is the issue that the hook just didn't update a Hero with video?

cory-skeers:
> I don't think we want to update the Hero block with video. Do we need to remove the ability for Webmasters to add new Hero blocks on Collegiate? Or is the issue that the hook just didn't update a Hero with video?

For the update hook, we're good to go then. I think we just need to restrict adding new hero blocks.

bspeare:
With the latest updates I have:

- Adjusted responsive solution to allow for character overflow
- Adjusted color contrast issues for headings, blockquotes, etc. 
- Set mobile solution to match desktop

joewhitsitt:
Is this still being worked on (go back to a draft state)?

pyrello:
@joewhitsitt I believe this is ready for testing.

pyrello:
I am going to update the labels on the headline LB styles, but that should not affect any functionality for testing purposes.

GaryRidgway:
I have reviewed up to just before the update hook steps and it is all working as planned!, However I think that it might be good to limit the banners Excerpt area as stated above. perhaps something like 600 characters or so?

pyrello:
@GaryRidgway @bspeare re-wrote the CSS to allow for the case where there is a lot of text.

pyrello:
@richardbporter I'm not sure why this job is failing. It seems to be failing when starting a local test server. Which seems pretty random, since the other PRs are not failing for this reason. Wondering if you have any ideas or insight

richardbporter:
I've followed the instructions but can't get past a Twig error locally:
```
Twig\Error\LoaderError: Template "@uids_base/uids/heading.html.twig" is not defined in "@uids_base/uids/event.html.twig" at line 22. in Twig\Loader\ChainLoader->getCacheKey() (line 142 of /var/www/uiowa/vendor/twig/twig/src/Loader/ChainLoader.php).
```

richardbporter:
The above comment is on a fresh default install. I'm guessing thats why the test is failing - the response code is a 50x error (whitescreen).

pyrello:
I think I have it figured out. Thanks for your help, @richardbporter!

richardbporter:
Yes! I was worried this was going to be a rabbit hole.

pyrello:
I'm going to test the update hook part of this so we can get it merged.

GaryRidgway:
I have tested this for banners and it works, i am working on testing the hero hook right now

GaryRidgway:
I have walked through all steps and confirmed That everything has worked fine for the update hooks. The one thing i did notice is that hero Images which had bottom aligned text got their text center aligned after the hook ran. I don't think this is a breaking bug, and in the grant scheme of things is low impact, but I would like to hear the opinions of others before I approve.

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -1088,3 +1092,210 @@ function sitenow_update_8032() {
...
+            'banner_medium',
+            'headline_bold_serif',
+          ];
+          // Can add to this if more style transfers are needed.
+          foreach ($comp_styles as $style) {
+            switch ($style) {
+              // Replace default centered with left aligned.
+              case 'block_style_hero_left_aligned':
+                $new_styles[1] = 'banner_centered_left';
+                break;
```
cory-skeers: This code sets the style options for banners created out of legacy hero blocks.

I think `banner_centered` is accurate if no style is set (like the banner on coronavirus.uiowa.edu), but the hero's `block_style_hero_left_aligned` should maybe be updated to `banner_lower_left` to better fit existing styles (like americanstudies.prod.drupal.uiowa.edu).

pyrello:
@GaryRidgway Can you provide the example of the site and page where you seeing the example that you are unsure about?

docroot/profiles/custom/sitenow/sitenow.install
bspeare: @cory-skeers I see you just updated it lower left, but I would prefer the centered as the default update.

pyrello (approved):
Approving based on Alan's testing and subsequent changes.

GaryRidgway:
sorry, left for lunch. I used americanstudies.uiowa.edu

Issue #1739, opened 2020-06-04
Problems with background color for section or blocks.


mark-bennett-uiowa:
NOTE: This came from creating a temporary solution for a program block. Once program blocks are created, I will no longer be using this solution. Thought it would be helpful to catch these issues for the future though.
I hope this screenshot gives enough information as to the issues. The blue text represents a notice of behavior rather than an issue.
1: Section Padding: Top needs to be added to the section to add a colored spacing above first widget in section. I would assume this is intended and doesnt need to be fixed.
2: Uncolored background between two consecutive sections with background color.
3: Background color of card blocks is set to white, which does not show during layout building. Only appears upon save of layout.

To see this live: [https://africanamericanstudies.prod.drupal.uiowa.edu/undergraduate](https://africanamericanstudies.prod.drupal.uiowa.edu/undergraduate)

![Screen Shot 2020-06-04 at 9 32 41 AM](https://user-images.githubusercontent.com/3067017/83773577-4db94680-a64a-11ea-87bb-d80e334b138f.jpg)

bspeare:
A few notes on how this could be addressed:

- If a section has a background color applied, top and bottom padding should automatically be applied.  
- Each section has default bottom margins applied to it.   If there are two consecutive sections with background colors these margins should be stripped out. 
- Card currently has a white a default background color.   The card background should adjust based on the containers background color?

mark-bennett-uiowa:
@bspeare That all sounds great to me, but do we want/need to have a bunch of conditional logic on the styles? I dont want to cause problems with this if I'm just building something out incorrectly.

bspeare:
I think we want those in place so that it is easy for someone to design a section.   @pyrello do you have any thoughts on this?

pyrello:
I think ideally, we have sane default that make it easy to build this out without having to set a lot of overrides. If something outside of those defaults is desired, then that were the overrides come in.

Not sure if this helps clarify.

mark-bennett-uiowa:
This appears fixed (for the gray background) in #2076.

PR #2069, opened 2020-08-21
More bg-color combination work

joewhitsitt:
Short term solutions in uids_base for https://github.com/uiowa/uids/issues/242 and https://github.com/uiowa/uids/issues/262

Fixes #2074 
Fixes #2005
Fixes #2080

## To Test

Try out the blocks (cta, collections, card, text area, body, webform, accordion, event and the footer contact block) with different color backgrounds and the gray text area block option.

You will also notice that SiteImprove extension no longer complains about #2074.

PR #2076, opened 2020-08-24
Feature: Layout builder styles refactor in uids_base

### LINKED ISSUE #2053: Default padding for sections with background color is not being applied if "extended width container" is selected
# Problem 

Selecting "Extended width container" in the layout builder styles removes the default padding for sections using a background color:

Screen Shot 2020-08-20 at 12 58 21 PM
Screen Shot 2020-08-20 at 12 48 51 PM

# Proposed solution

Adjust css in `global.scss`.

### LINKED ISSUE #2054: Add default padding to text area block if it is used with a background color
# Problem

If you add a background color to the text area block, it will appear without default padding. 

Screen Shot 2020-08-20 at 1 19 38 PM

# Proposed solution

Add default padding to text area block if it is used with a background color. 

Screen Shot 2020-08-20 at 1 19 58 PM

### LINKED ISSUE #2059: Adjust use of default margins in layout builder sections
# Problem

Each section in layout builder is given default `margin-bottom: 3rem;` and to remove this default margin there is a layout builder style to remove it.  The default margin has been set to provide better spacing so that content editors don't create pages with components on top of each other. 

Screen Shot 2020-08-20 at 3 32 35 PM

# Proposed Solution

Create a more robust solution for adding default margins based on proximity of sections with a background color.  

Do we want to:

- Remove all margins and only add them by default using adjacent selectors (https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) if section with a background color is next to a section with a background color?
- Continue with current strategy?
- Other suggestions?

### LINKED ISSUE #2156: Can 'flex__wrapper' be removed from layout--onecol.html.twig
For at least 6 months, there has been a typo in the `layout--onecol.html.twig` template where `flex__wrapper` was accidentally changed to `flex__wra`. This is a commonly used template and it seems to have not caused any issues that I am aware of, so to me it stands to reason that the logic in the code that adds the `flex_wra` class (and containing div) is no longer needed.

@bspeare When you have time, can you speak to whether or not this can be removed?

bspeare:
Resolve #2059. 
Resolves: #2156 

Resolves #2054. Text Area blocks currently don't receive padding when a bg color is added, leading the text to run up against the bg boundary.

Resolves #2053 . Sections with 'extended-width' style losing top and bottom padding, leading contents to run up against edges. This adds in a $desktop-width-gutter padding in these cases.

Resolves padding issue with (vertical) card bodies.

# How to test updated layout builder styles

Test on uiowa.edu, https://hawkeyemarchingband.uiowa.edu/, and a SNV3 to make sure that https://github.com/uiowa/uiowa/issues/2059, #2054, and #2053 are fixed. 

- **Note:** these layout builder styles updates will require some manual updates to be made to the section containers on uiowa.edu.  

# How to test updated card background color variant

Card padding: https://github.com/uiowa/uiowa/issues/1739
- Change the background color of an individual card block to gray and test whether the card background color adjusts if horizontal and bordered variants are enabled.


bspeare:
@joewhitsitt I think I might try to resolve https://github.com/uiowa/uiowa/issues/1739 in this PR unless you were planning on addressing it in https://github.com/uiowa/uiowa/pull/2069.

joewhitsitt:
Fine with me. I was not trying to solve it beyond card defaulting to white bg

mark-bennett-uiowa:
Pulled down for testing.

mark-bennett-uiowa:
This appears to be working on hawkeyemarchingband.uiowa.edu (this site looks like it uses Paragraphs primarily rather than LB):
Screen Shot 2020-09-02 at 12 32 42 PM
Screen Shot 2020-09-02 at 12 39 22 PM

mark-bennett-uiowa:
This is working great on americanstudies.uiowa.edu local site.
Screen Shot 2020-09-02 at 1 17 33 PM

Here is the default padding on colored background at work. This works on save also:
Screen Shot 2020-09-02 at 1 18 25 PM
Screen Shot 2020-09-02 at 1 18 42 PM

Here is the 3rem padding on a colored background section (I ended up removing the Remove default column spacing style and this worked). The sections are now properly up against one another also for two concurrent background colored sections:
Screen Shot 2020-09-02 at 1 23 07 PM
Screen Shot 2020-09-02 at 1 29 20 PM

**NOTE:** Bordered and horizontal style cards are getting padding (the horizontal cards have some inherited padding from somewhere even though the styling says padding:0):
```
.layout.layout__container[class*=bg--] .card:not(.card--horizontal) .card__body, .layout.layout__container[class*=bg-pattern--] .card:not(.card--horizontal) .card__body {
    padding: 2rem;
}
.layout.layout__container[class*=bg--] .card.card--horizontal .card__body, .layout.layout__container[class*=bg-pattern--] .card.card--horizontal .card__body {
    padding: 0;
}
```

mark-bennett-uiowa:
**UPDATE:** Did notice these two small issues this PR:

**Issue 1:**
Screen Shot 2020-09-02 at 1 41 02 PM
Screen Shot 2020-09-02 at 1 41 13 PM

The section divs above are:
```
<div class="bg--gray layout layout--onecol layout__container">
<div class="layout layout--onecol layout__container">
```

And they get the following styling:
```
@media (min-width: 84.375em)
.layout__container[class*=bg--], .layout__container[class*=bg-pattern--] {
    padding-top: 3rem;
    padding-bottom: 3rem;
}
@media (min-width: 84.375em)
.layout__container {
    padding-bottom: 2rem;
    padding-top: 2rem;
}
```
So the gray background section gets 3rem and the default no background gets 2rem padding. Is this intended to have a smaller padding for no background? All colored backgrounds are tested and properly getting the 3rem top and bottom padding.

**Issue 2:** The last issue happens if we have a colored background on the section AND on the block contained within:
Screen Shot 2020-09-02 at 1 48 44 PM
Both section and block are getting a 3rem padding:
```
.layout__container .block-inline-blockuiowa-text-area[class*=bg--], .layout__container .block-inline-blockuiowa-text-area[class*=bg-pattern--] {
    padding: 3rem;
}
```

This may not be a concern moving forward now that #2069 has been merged. Only time this will happen is if we add a colored background to a block that already contained in a colored background section. It could happen, figured I would mention this.

pyrello:
@mark-bennett-uiowa This has been updated to allow for sections with backgrounds and those without to have the same padding.

The main place where this becomes noticeable is in back-to-back sections without a background:
Screen Shot 2020-09-08 at 12 01 03 PM

It seems like, if we want to keep `3rem` as the standard for all section containers without modification, then it seems like, anytime there is multiple no background sections in a row, then the user will likely want to remove padding on one or more of them.

Alternately, we could potentially automatically remove the leading padding from a no background section that follows another no background section.

I also added the narrow section, but I'm still working on that.

pyrello:
After playing around with this a little bit more, I am of the opinion that when we have multiple sections consecutively that have the same background treatment (including no/white background), we should strip the top padding from the second (and third, and fourth, etc.) sections that have the same background. To clarify, that would mean that each consecutive section with the same background would only have `3rem` of padding between them. This is as opposed to the total of `6rem` of padding between two sections with different background treatments. Here is an example of both:

Consecutive sections with no/white background (I used chrome's element highlighting to show the padding of the middle section):
Screen Shot 2020-09-08 at 3 11 24 PM

Consecutive sections with different backgrounds (No element highlighting):
Screen Shot 2020-09-08 at 3 01 50 PM

The downside to this approach is that it breaks our logic of every container has the same top and bottom padding, which makes the result predictable for users. However, I think this is warranted because I think it will be a fairly common use case for users to want to place multiple sections consecutively with the same background treatment. It seems unlikely that the user is going to want `6rem` of space between sections with no background color change. If we follow the consistent padding logic, then that means the user would be forced to apply the 'Remove top padding' style to all of those containers. If we follow my proposal, then the user would need to do nothing in most cases.

cory-skeers:
I like the adjusted padding approach, @pyrello. While it technically moves away from consistent padding amongst all containers, I think the the end effect is a closer match visually to the behavior between same-background and different-background sections.

mark-bennett-uiowa:
I chatted with Sean about this a little bit ago, and I am in agreement with this change. As long as the conditional logic works as intended and designers / editors are aware of this (and can add back in a space if desired), I say go for it.

pyrello:
I'm going to make a minor adjustment, but otherwise, I think this is ready for testing from anyone in @uiowa/developer. The change I'm going to make shouldn't change anything style-wise.

docroot/themes/custom/uids_base/scss/admin.scss
```diff
@@ -88,7 +93,7 @@ $imgpath: '../../uids/assets/images';
     &:hover,
     &:focus {
       .layout-builder__layout {
-        .lb__container {
+        .layout__spacing_container {
```
cory-skeers: Looks like this scoping places the editor outline on the interior container and not the selected padding. I think the outline can be moved up to the `.layout-builder__layout`, but not sure if that would cause regressions elsewhere.

image

pyrello: Nice catch, @cory-skeers!

cory-skeers:
I've tested this in a fresh SNV3 install and on uiowa.edu, and this is looking great.

For existing sites, I don't suppose we have a checklist of sorts on what might need to be updated manually? I'm thinking mostly for uiowa.edu as likely the highest profile, or if anyone already has a plan for what to update there as this goes out.

joewhitsitt:
pulling down for testing. sry for the delay

pyrello:
@joewhitsitt No problem. Much appreciated!

joewhitsitt:
Not sure if it is in scope of the card padding issue, but media images right are right up against the border on v3. Huge improvement to card padding in general though:

Screen Shot 2020-09-23 at 2 03 31 PM

joewhitsitt:
>Note: these layout builder styles updates will require some manual updates to be made to the section containers on uiowa.edu.

Do we know if this is the only site that needs manual changes?

bspeare:
I was planning on running through uiowa.edu and fixing any container issues when this is released.   I think a number of @mark-bennett-uiowa's sites will need a similar review.

pyrello:
coronavirus.uiowa.edu also has some updates that will be needed. We swapped a couple of the LB styles to try to limit the impact. I have a query figured out that gives us a list of the pages that we'll need to make updates on.

joewhitsitt (approved):
Going to approve with the assumption that everyone knows which sites they need to manually update upon deployment.

Not noticing any major "breaking" issues otherwise on v2/v3 sites tested

Created a follow-up to update docs: https://github.com/uiowa/uiowa/issues/2226

mark-bennett-uiowa:
> I was planning on running through uiowa.edu and fixing any container issues when this is released. I think a number of @mark-bennett-uiowa's sites will need a similar review.

Yep yep. I am planning on it, and since all but two sites are still in prod not live, I should be okay. Ill just have to see what exactly needs to get updated after release. I say go for it.

PR #2079, opened 2020-08-24
Update accordion ids

### LINKED ISSUE #2077: Collections Accordion ID Not Unique
## Problem

There appears to be an issue with setting a unique id for each accordion item within a collection.

Screen Shot 2020-08-24 at 3 07 33 PM
Screen Shot 2020-08-24 at 3 10 09 PM

cory-skeers:
Resolves #2077 

Accordion items in LB collections not receiving unique IDs. Updated where the accordion-item-id is set, and removed the single/multi-select identifier in the ID, as all accordions are currently multi-select.

## Testing
Sync, blt frontend.
Add at least 2 collections to a page, each with at least 2 items, and view mode set to 'accordion'.
Check that each accordion item within a collection has a unique ID, and that these are unique between the multiple collections.
Check that the "aria-controls" tag of the heading matches the id of the child `<div>`
Check that the "aria-labelledby" tag of the child `<div>` matches the id of the correct heading.
Check that the `<div>` "dataparent" tag matches the parent collection container id.

### DIFF
--- a/docroot/themes/custom/uids_base/templates/paragraphs/paragraph--uiowa-collection-item--accordion.html.twig
+++ b/docroot/themes/custom/uids_base/templates/paragraphs/paragraph--uiowa-collection-item--accordion.html.twig
@@ -41,7 +41,7 @@
 
 {% set accordion_item = {
   'accordion_h_size': h_size ?: 'h2',
-  'accordion_item_id': 'accordion-single-select-' ~ parent_id ~ '-' ~ delta,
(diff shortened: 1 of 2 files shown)


dnkemper (approved):
Followed testing instructions and everything worked great!

Side note: I did notice that there were background styling issues while testing this PR?  Can add a separate issue for that though if you like. 
Screen Shot 2020-08-24 at 4 39 08 PM

PR #2086, opened 2020-08-25
Create custom block type uiowa_events for use in layout builder.

### LINKED ISSUE #2071: Create custom events block for use in Layout Builder

richardbporter:
Resolves #2071 

### Testing
- Sync a site via `blt sync`.
- Add an Events block to a page. Try different combinations of filters.


richardbporter:
Do we want to axe the single event CSS/JS and theme it up in uids_base? There seems to be a problem with the current single event dates - some future dates are appearing as past.

richardbporter:
Also looking for help on the heading level field and the LB styles for hiding image, description, etc. - can I reuse the ones created in #2032?

docroot/modules/custom/sitenow_events/sitenow_events.module
```diff
@@ -6,169 +6,69 @@
...
- * @param int $parent
- *   Stores the current id.
- *
- * @return array
- *   Returns an associative array options tree.
- */
-function _build_options_tree(array $data, $parent = 0) {
-  $tree = [];
-  foreach ($data as $d) {
-    if ($d['parent_id'] == $parent) {
```
richardbporter, line 132: I think this loose comparison is a recursion trap but changing it breaks all the filters. It currently is equating null and 0.

joewhitsitt:
>Do we want to axe the single event CSS/JS and theme it up in uids_base? There seems to be a problem with the current single event dates - some future dates are appearing as past.

Can you elaborate? I am not sure I follow.

>Also looking for help on the heading level field and the LB styles for hiding image, description, etc. - can I reuse the ones created in #2032?

I think that is a good idea. I plan to revisit the PR to work out the kinks and maybe reuse fields between the two blocks. Can this PR wait for that?

docroot/modules/custom/sitenow_events/sitenow_events.module
richardbporter: Working on this.

richardbporter:
>Can this PR wait for that?

Yes.

> Can you elaborate? I am not sure I follow.

Should we move the single-event.css file to the uids_base theme and possibly refactor? I feel like we could use an on-brand accordion for the dates dropdown instead of rolling our own.

richardbporter:
Also. we could leverage some of the single event styles like location/date icons, etc.

richardbporter:
I'm going to merge this branch into #2032 and close this PR so we can address a couple interconnected issues.

Issue #1920, opened 2020-07-08
Further investigation of inline block/LB permissions


pyrello:
As a follow-up to #1890, we should see what options exist in the contrib space to allow for greater control over LB or inline blocks.

joewhitsitt:
- [x] Restrict vertical video to homepage.
- [ ] Restrict hero to collegiate. _being consolidated to banner_
- [x] Restrict separator to homepage only.
- [x] Remove Who's Online?
- [x] Remove Recently Updated?
- [x] Remove views until we are ready to let users place them.

richardbporter:
>Remove Who's Online?
 Remove Recently Updated?

I don't see these. Checked.

PR #2087, opened 2020-08-25
Hotfix LBC form alter fatal error.

### LINKED ISSUE #2085: Error: Call to a member function bundle() on null in layout_builder_custom_form_alter line 84
## Problem

As a webmaster I can't add any of the "More" blocks due to this console error.

richardbporter:
Resolves #2085 

I noticed its possible to add a collection with no items in it. While I'm in here, I can make the collection items field required? Or does that cause other problems?

### DIFF
--- a/config/default/block_content.type.uiowa_cta.yml
+++ b/config/default/block_content.type.uiowa_cta.yml
@@ -5,6 +5,6 @@ dependencies: {  }
 _core:
   default_config_hash: 9lZDUWJZh45bfBHWJzL9ac_3T_GCAFHUz0GKQqylx08
 id: uiowa_cta
-label: Call to action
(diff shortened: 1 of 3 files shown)


richardbporter:
Sorry for imposing my whitespace OCD. I tend to add whitespace subconsciously. :)

cory-skeers:
Ah, good catch--Yes, I think a collection item should be required.

richardbporter:
When I exported that change, it ended up in the sitenow_v2 split somehow. Probably some config dependency I'm not going to debug right now. I changed it manually.

joewhitsitt (approved):
Thanks! Now back to locking them down https://github.com/uiowa/uiowa/issues/1920

PR #2093, opened 2020-08-27
Add default table classes for view tables

pyrello:
This PR adds default table classes to views tables.


mark-bennett-uiowa (approved):
Inspected code, didnt test. Looks simple enough and good for me so far.

Issue #2078, opened 2020-08-24
Adjust placement / use of moderation control


bspeare:
# Problem

The "Main content" area is often empty if no Body content has been created.     Having this Main content area rendered within the page layout creates extra spacing due to the default spacing that we are adding to Sections. 

<img width="1752" alt="Screen Shot 2020-08-24 at 3 07 52 PM" src="https://user-images.githubusercontent.com/1036433/91091269-fb217200-e61b-11ea-95a7-3501613b049e.png">

# Proposed Solution

- Remove "Main content" area from rendering if there is no Body content.
- Remove "Links" field.
- Or if the section is empty, render the "Main content" section last so that the Moderation control options will appear at the bottom of the page instead of in the middle. 

<img width="1288" alt="Screen Shot 2020-08-24 at 3 13 40 PM" src="https://user-images.githubusercontent.com/1036433/91091593-6bc88e80-e61c-11ea-8b16-44af5e0feeca.png">

pyrello:
This is how this looks now:
![Screen Shot 2020-08-26 at 7 16 06 PM](https://user-images.githubusercontent.com/709009/91369155-a70bbe80-e7d0-11ea-9be8-1600c7614a04.png)

I'm cool with removing the "Links" because it doesn't seem like it ever gets used, at least so far.

I'm not sure about the rest of it. FWIW, this is the Coronavirus site, which splits `core.entity_view_display.node.page.default`, so it has its own template. This is also a different template than what is used in SiteNow V3.

cory-skeers:
I worked on this some, but had mostly focused on changing how it appears on the editor side, and didn't get to a good resolution.

The main part that needs addressing is on the page, rendering the field w/out content leads to extra spacing. I've added sections above to draw attention to it:
![image](https://user-images.githubusercontent.com/49074566/91449115-5764db80-e840-11ea-9c99-5c5b4fbd7fe4.png)

![image](https://user-images.githubusercontent.com/49074566/91449571-e2de6c80-e840-11ea-9b48-bc7c30b9ee78.png)

mark-bennett-uiowa:
I was curious about some of this also, as to (1) were fields like Links needed, etc., (2) is it possible to "hide visually" a section/block in LB, etc., and (3) to ensure that I could recreate the Main content section's default layout (Moderation control, Body, Links) if I moved to a different solution for certain pages and had to revert back.

[...]

PR #2111, opened 2020-08-28
Remove extra spacing if node body field isn't used

cory-skeers:
Relates to #2078.

Primary goal is to remove the extra spacing on an LB page if the body content field is empty.

The previous `{% if content.content %}` led to two nested `<div>` tags with no content but default margins, as content.content resolved to true even though it rendered to empty whitespaces.

I'm not a huge fan of the idea of rendering the content in order to check that there is content to display, but this is the lowest impact way I can think to do it (and at least the rendered content gets stored for use in display after, if it's not empty).

## Testing
Sync, blt frontend
Create a page, leave the body field (in the Edit tab) blank. 
The only `<div class="layout__container ...">` tags should be associated with content added through layout builder.
Edit the body field (in the Edit tab).
The body field content should appear on the page.

General testing that no regressions occur. This should only layout__onecol sections with no renderable content, though.


dnkemper:
Tested, and seemed to work great for me! Yay!

dnkemper (approved):

joewhitsitt:
Does something like the following not work? 
```
{% if content.content[0] is not empty %}
{{ content.content }}
{% endif %}

cory-skeers:
@joewhitsitt That didn't work for me, but it could have been a notation error on my part. When I tried, it always came up with something to resolve to true, even if ended up as empty space later.

joewhitsitt:
Ooh. What about?

```{% if content.content|render is not empty %}```

Only thinking about optimization/condensing, if what you have works, I am okay with moving forward.

cory-skeers:
The `{% if content.content|render is not empty %}` route ends up going through the render process once for the logic check, and once again for the page render.

Think I'm going to go down one more short rabbit hole to double check before merging.

PR #2106, opened 2020-08-28
Updated Drupal to 8.9.3

pyrello:
Resolves: #2103 

This update included:
* Updating `drupal/core` to `^8.9.3`
* Updating `drupal/date_recur_modular` to `^2.0`
* Removing `drupal.core` patch for 'Form blocks rendered inside layout builder break save'.
* Updating `drupal/core` patch for 'fix duplicate drag handles in claro' to a version that was compatible with `8.9.x`
* `.htaccess` was updated and I would love to have someone else look at it to make sure that changes seem okay.

# To test
* `composer install`
## CLAS Policy
* `blt ds --site=policy.clas.uiowa.edu`
* `drush @claspolicy.local uli`
* Navigate to home page and click the Layout tab.
* Ensure that you are able to save the form, despite the embedded form on the page.

## SN2
* `blt ds --site=default`
* `drush @default.local uli`
* Test against SN2 Paragraphs, Media, WYSIWYG, etc. functionality.

## SN3
* `blt drupal:install --site=default`
* `drush @default.local uli`
* Test against SN3 LB, Media, WYSIWYG functionality.

## Extra credit
* Test against coronavirus.uiowa.edu.
* Test against uiowa.edu.
* Test against any other split sites.

I don't think it is necessary to go too overboard with testing existing functionality, since all patches - except for the ones mentioned above - still applied.

PR #2115, opened 2020-08-30
Feature: Exclude .htaccess from being copied

pyrello:
This update adds the `.htaccess` file to an exclusion list so that it is not copied offer when file scaffolding happens during an update to drupal/core. Previously, it was necessary to compare the `.htaccess` file against its previous version to copy redirection rules back into it.

# To test
I'm not sure if there is an easy way to test this. One idea would be to pull down the branch, then make some change to the `composer.json` file, like removing a patch from `drupal/core` and then run `composer update --lock`. This will cause `drupal/core` to be removed and re-added so that patches can be re-applied. I think that would trigger the scaffolding command to run and you can then view that the `.htaccess` file remains unchanged. However, I'm not 100% sure that is how it works.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -262,7 +262,10 @@
             "initial": {
                 "sites/default/default.services.yml": "sites/default/services.yml",
                 "sites/default/default.settings.php": "sites/default/settings.php"
-            }
+            },
+            "excludes": [
+                ".htaccess"
+            ]
         }
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


richardbporter:
I don't think we want to do this because then we won't get htaccess changes provided by drupal/core like in https://github.com/uiowa/uiowa/pull/2106, which we want. We just need to diff the file and re-implement our redirects, etc.

We should use this feature to exclude files like robots.txt which we never want since we use the robotstxt file for multisite support.

PR #2130, opened 2020-09-02
Events Follow-Up

joewhitsitt:
- Override events-single styles in uids_base to match Article
- Fix for UHR's event single render being blank.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/content/sitenow-event-single.scss
+++ b/docroot/themes/custom/uids_base/scss/content/sitenow-event-single.scss
@@ -1,0 +1,39 @@
+@import "uids/assets/scss/_variables.scss";
+@import "uids/assets/scss/_utilities.scss";
+
+.sitenow-event-single {
+
+  .date-instance__next-upcoming {
+    display: block;
+    @include card-subtitle;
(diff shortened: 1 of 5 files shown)


pyrello:
I was getting tempted to try to implement the event template, but then I realized that the event template needs to be re-written and that's not going to happen until @bspeare gets back, so I'm leaving it for now.

pyrello (approved):
Looks good. Nice work!

richardbporter:
1960cea prevents overriding the `card_content` so the location shows on the teaser. Can we revert that or is there another way of handling it?

joewhitsitt:
I didn't realize @pyrello snuck that change in there. I am not familiar enough with Twig to work around it another way.

pyrello:
The `only` keyword just limits the variables to those that are explicitly passed. If you need a implicit variable (one that is available in the current scope but not defined in the current template), just explicitly add it to the `events_card` object that is passed to the template.

What is the variable that is missing?

pyrello:
FYI, I am working on a fix for this.

richardbporter:
Can you add it to https://github.com/uiowa/uiowa/pull/2133?

PR #2132, opened 2020-09-03
Fix single event contact fields.

richardbporter:
Contact fields were referencing pre-event system data, similarly to others.


pyrello (approved):
Seems to work correctly.

PR #2133, opened 2020-09-03
Hotfix: Pass through event-specific variables.

pyrello:
This is a follow up to #2130 and is meant to fix the issue where I broke the event location showing up.


richardbporter (approved):
This seems to work, although I don't totally understand why.

pyrello:
Once the `only` keyword is used, the only variables available within the `embed` tag are what are passed in with the object (e.g. `events_card`).

pyrello:
Part of the reason I started using `only` is because certain variables were getting implicitly passed through and causing unexpected behavior.

PR #2136, opened 2020-09-03
update fontawesome library

### LINKED ISSUE #2061: Update fontawesome icon library
## Problem

We need tiktok?!

joewhitsitt:
Resolves #2061 ?

Does just that. The library is updated, the contrib module we use on the backend is not updated. Good news? You can manually add `fab fa-tiktok` into a social media menu link icon field and have it work.

## To Test

Pull and create a social media menu link with the following value `fab fa-tiktok`. The icon picker does not work (shows blank).
Save the menu and view the rendered footer. See tiktok icon.

Screen Shot 2020-09-03 at 4 42 36 PM

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -219,16 +219,16 @@
         },
         {
             "name": "bower-asset/fontawesome",
-            "version": "5.13.0",
+            "version": "5.14.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Tested it and I was, in fact, able to add TikTok to a social media menu.

Issue #931, opened 2020-01-14
Multisite code is not scanned for standards and deprecations


richardbporter:
It would be nice to scan multisites for coding standards violations and deprecated code, especially with D9 on the horizon. Unfortunately, just adding the `docroot/sites/` directory results in very long build times. It is too large. 

We could maybe accomplish this with a custom fileset that excludes lots of BLT-related files that we don't need to scan. Just the minimum of what we care about.

https://docs.acquia.com/blt/extending-blt/#adding-or-overriding-filesets

richardbporter:
I'd be interested in working this in once the HR and UIowa Bootstrap workspaces are out of the build process.

PR #2140, opened 2020-09-08
Add dependency on jquery.cookie to LBC library.

richardbporter:
When editing a layout, there is a console error regarding jquery.cookie not being available: `Uncaught TypeError: $.cookie is not a function`. This PR adds a dependency to the module library to prevent it. I'm not sure how this didn't crop up earlier. Maybe the update to Drupal 8.9 removed the library from somehow.

### Testing
- Go to a layout enabled page and open the console.
- Edit the layout and trigger the off-canvas dialog (add a block for example). See error.
- Check out this branch. Rebuild the cache. Refresh the page.
- Repeat steps above and notice the error is gone.

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/layout_builder_custom.libraries.yml
+++ b/docroot/modules/custom/layout_builder_custom/layout_builder_custom.libraries.yml
@@ -4,6 +4,7 @@ layout_builder_custom.overrides:
   dependencies:
     - core/jquery
     - core/jquery.once
+    - core/jquery.cookie
   css:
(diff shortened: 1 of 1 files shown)


joewhitsitt (approved):
hotfix tested and approved, but we should probably update our custom code to use the new cookie stuff in D9 and not rely on the "shim". https://www.drupal.org/node/3104677

richardbporter:
Yep, I assume we'll need to audit all our code for deprecations. Somewhat related: #931

PR #2142, opened 2020-09-08
Initial covid data POC.

### LINKED ISSUE #2135: Create BI integration proof-of-concept
# Problem
We may need to integrate with BI data in different scenarios. Some scenarios may contain sensitive data that should not be publicly available. For example, the table on https://coronavirus.uiowa.edu/covid-19-numbers could be fed dynamically. 

# Proposed solution
Create a custom block for placement in LB that grabs data from API. This block would exist only for  coronavirus.uiowa.edu. The API endpiont and credentials should remain out of VCS: https://docs.acquia.com/resource/secrets/

richardbporter:
Resolves #2135.

### Testing
- Get the credentials from the [secrets file](https://docs.acquia.com/resource/secrets/#secrets-settings-php-file) on uiowa01 dev. Add them to the coronavirus site local settings file.
- Enable the uiowa_covid module.
- Place the block on page via layout builder.


richardbporter:
Proof this works... :)

PR #2152, opened 2020-09-10
Parent title field

### LINKED ISSUE #2048: Parent title field on certain block types
## Problem

To provide accessible DOM order to all parts of a page, we should provide a consistent option across certain block types to allow a parent title to be added.

## Proposed Solution

Look into extending the headline field or creating our own combination field.

- slider, collections, articles, people, image gallery
- possibly text area block in the future?

Instead of using a textfield, a long text field would allow future enhancements like word highlighting.

- optional
- long text field with a length limit (255?), 1 row display.
- heading size if not extending the existing headline field.
- visually hidden title (boolean)
- font selection (select)
- other style options?

If set, the parent title would set the children items appropriately making sure the DOM follows proper heading order. If not set, the children headings field would display and be configurable to set the children items.

cory-skeers:
Resolves #2048 

Compound title field to share across blocks. Includes the title itself, a way to control its heading size, and uses this to set child heading sizes (accordion headings, article titles, etc.) for proper DOM hierarchy. If no title is needed/wanted, the heading size can be set for the children directly.

## Testing

Pull
`composer install`
`blt frontend`
sync a site
Test on existing or new blocks:
* Articles
* Events
* Text Area
* Image Gallery
* Collection
* Slider
Check the following:
* Include a title and set different heading levels
  * For blocks with child elements, they should appear at the appropriate level below the title
* Try visually hiding the title, with various heading levels
  * The title should be hidden
  * The title should appear in the DOM on inspection
  * For blocks with child elements, they should appear at the appropriate level below the title
* Try setting the "child heading level" when the title is not filled
  * No title should appear on render, either visually nor in the DOM
  * The child elements' heading levels should render at the appropriate level

<details>
<summary>
Old notes
</summary>

Roadmap for parent title field.
* What is needed, assuming it requires a compound field?
  - [x] Title (text)
    * Plain, allow for formatting for future options (highlighting)?
      * For future options, do we need a minimal HTML field or similar, or can we use a plain, long field with tags like `<hl>` (or whatever is in place) that would be replaced with complete markup?
  - [x] Size (select list)
    * One field. Logic applies it to title if it is used, or children if not.
        * Any use case where you would _want_ to set both title and children sizes separately, that wouldn't break hierarchy?
  - [x] Display title (boolean)
    * How should this interact with children? Have a hidden title heading over visible children headings, or does this essentially mean don't use the title field?
  - [ ] Styles?
    * In addition to possible formatting in the text field that might apply to only part of the title, this would apply to all.
    * This can be added later more easily than others, as it would be adding a field or view modes, but the overall field won't be affected.
* What logic needs to be created
  - [x] Display/no display title
    * How does the hide title boolean (above) need to function?
  - [x] Control heading size of children
    * This is in place for collections/sliders currently (which both use paragraphs), and needs to be expanded for other block types.
  - [x] Update hook for blocks using existing Heading size field
    * Would need to set this new field for any block that uses this old field, and set heading size appropriately.
      * And probably need a two-step process. 1) update config to add new heading field alongside old heading size, and update hook to appropriately set the new field, and 2) Remove the old field
    * For any blocks which don't currently have this field, need to set this field (or make sure any NULLs are appropriately handled).
      * Collections (field was hidden--default to h2)
      * Text Area
      * Image Gallery
* What blocks need to use it?
  - [x] People/articles/events
    * Views-type collections, for a title above the view and setting the child heading sizes appropriately.
  - [x] Collections
    * Accordions (and future tab potential) set a title above and child headings. Descriptive lists the title and its heading size only.
  - [x] Slider
    * Title above, and slider heading sizes.
  - [x] Image gallery
  - [x] Text block(?)
    * Title above. Possibly unset heading sizes on that level or above?
        * e.g. Text block with an `<h2>` as a title has access only to 'Heading 3' and below from the style dropdown. Simple approach, would be possible to break by going into Source, but would provide some guidance to users (and potential confusion if they don't understand, and suddenly can't find a 'Heading 3.')

## PR current state
 * Layout Builder Custom defines a fieldtype and its formatter/widget.
 * Config sets this field on several blocktypes.
 * uids_base.theme paragraph preprocess sets heading sizes for collections and sliders.
 * sitenow_articles, sitenow_people, sitenow_events block preprocesses set heading sizes for their respective blocks.
 * Update hook sets fields appropriately for blocks which have a heading_size or existing title/heading_size field prior to those fields being removed from the block.
</details>


joewhitsitt:
There was talk about the text area block as well, but I will let @pyrello make that pitch. ;)

joewhitsitt:
**Comment Status: In Progress**

First off, IMO the main field here should be the title which is optional. If filled/not empty #states should show the other fields with reasonable defaults (e.g. h2). This keeps the form simple if not filled out.

> For future options, do we need a minimal HTML field or similar, or can we use a plain, long field with tags like <hl> (or whatever is in place) that would be replaced with complete markup?

I think the short term is to allow certain html elements and classes without a ckeditor format/toolbar. Help text can supplement if we want to expose the possibilities to more than power users. If we need to make it more user friendly, I think it is a separate format from minimal.

> Any use case where you would want to set both title and children sizes separately

I can't think of a reason to do this, especially since most of these children headings don't actually change the visual size.

> Display title (boolean)
> How should this interact with children? Have a hidden title heading over visible children headings, or does this essentially mean don't use the title field?

The optional title paired with the heading size is what should control the children. The visually hidden title should still have a heading size and set the children appropriately.

>Styles?

Not sure what options currently exist beyond selecting the font. Unless there is a list we are going to start with, I think this could be something added later but "studded out" in preparation for.

> How does the hide title boolean (above) need to function?

With v2, we controlled this with an if statement in each template.

pyrello:
FYI, I am reviewing this information. I'll follow up with another comment with any feedback.

docroot/sites/default/default.services.yml
```diff
@@ -0,0 +1,174 @@
+parameters:
```
joewhitsitt: we don't want docroot/sites/default/default.services.yml and docroot/sites/default/services.yml committed

joewhitsitt:
Started looking at this, I haven't tested all the blocks, but I am noticing that:

I am not presented an option to set children headings if no parent title value is given.

Parent Title value with Visually hide is not rendering to the DOM which is the ideal to keep the heading order and to provide screen reader navigation.

I can try to help with this later.

cory-skeers:
Looks like an update somewhere has broken rendering the block headline on collections (haven't checked sliders, but they are likely to have also broken).

# Current to do:
- [x] Fix bug that is preventing headline from rendering on collections (and possibly sliders)
- [x] Present an option to set child heading size if title field is not filled
  - [x] Update templates to utilize heading options according to this logic
- [x] Revert card changes

joewhitsitt:
pulling down for testing

joewhitsitt:
- Not loving the help text related to children items.
  - _"The heading size for the block title. Children headings will be set one heading level lower."_
  - _"The heading size for all children headings."_
  - Text area is not a "collection" so children help text doesn't apply and confuses me thinking that it will adjust headings within WYSIWYG (text area). form_alter change help text?
  - Image Gallery does not have children headings. form_alter change help text?
  - collection descriptive list didn't adjust child items. not sure it should, but again, the help text should reflect that.
    -  _"The heading size for the block title. Accordion item headings will be set one heading level lower. Does not apply to descriptive list."_
- headline has no targetable classes to style against. is there a uids default heading class that should apply here? `headline`
- not working on slider/collections.
- don't think we should allow h6 for parent as there is no h7 to apply to children.

cory-skeers:
Good points re: the help text areas. Think I'd started not really seeing them anymore, and they need updating.

For the sliders/collections not working, is it not setting the child heading levels, not displaying the headline, etc., or not working entirely?

joewhitsitt:
I think i set a h5 for slider parent title, and saw h3s for the single child i had. Collection accordions seemed to be similar. i can double check here in a bit. maybe it was limited to layout preview?

cory-skeers:
Made some updates to the help text, and its implementation for specific block use cases.

Removed the option for `h6` for headlines in cases where children are expected to be present, leaving it for the child heading option and blocks which don't have children (text area, image gallery).

cory-skeers:
Ah..yes. It looks like the original lb editor preview does not accurately set the child heading sizes, but it is set correctly after saving (for both viewing the page, and later viewings in the editor preview).

Not a breaking issue, but would be nice to get working in the original preview in case it may impact later styling options, and just for full accuracy in previews. Will look into it and/or create a follow-up issue for it.

## Edit:
The children headings are incorrect on the lb editor preview whenever the preview is in an unsaved state (Editing a block, adding a new block, etc., prior to saving the layout).

cory-skeers:
# To do
* Refactor theming templates to utilize uids `heading.html.twig`
* Include heading style options
* Track down editor preview heading difference bug

bspeare:
@cory-skeers are you able post screenshots and describe the style issues that you mentioned in standup?

cory-skeers:
@bspeare Adding the headline style options directly to the blocks that are using the headline field keeps it consistent with other block types, but without other adjustments gets confusing as to what it controls.

In the example below, the headline style options are added for an Articles block, but it's unclear if it should be applying to the overall block headline, the individual article headlines, or both. Depending on what kinds of options should be available, we could move the element on the form to make it clear it only effects the block headline, use it to effect all headlines, or add the styles directly to the uiowa_block_headline field where it could be handled separately from the children headlines and their possible styling.

image

bspeare:
@cory-skeers I would definitely want to see the "Headline options" within the block headline form section.   I'd also want to maybe limit the "Headline options" available.     Is that style option available in this PR?  I just tested this on uiowa.local and I don't see it.

bspeare:
@cory-skeers At this point, using the Articles list as an example, I don't think we want to change the headline style for each entry in the list since each component has a set grouping of styles.   I think the idea of changing the style of the block title is very appealing since most sites want to set a style for that heading or will do so by adding a separate block above if they really want to set a headline style.

cory-skeers:
@bspeare The style option isn't available yet in the PR (I'd been working with a half combo of this and the banner block refactor), but I'll start working on that next (including it with the field, so its more clear what its effecting) and let you know when it's available!

Do you have an idea of which styles we'd want to include, or which to limit?

bspeare:
@cory-skeers I'd probably want to start with just a few options like Zilla Slab or Zilla Slab with an underline:

Screen Shot 2020-10-07 at 3 02 47 PM

cory-skeers:
Refactored the field render to use `headline.twig.html`, and provide some classes (and some flexibility in adding/updating classes later). Also took a stab at getting style options in place.

image

Not a fan of how it's implemented (as a separate 'style' field, rather than using layout_builder_styles), but the main drawbacks of this way would be the slight difference in form style (screenshots below) and not automatically getting headline options pulled in. The former could be addressed through css, and the latter isn't too bad if these headlines will only be using a controlled subset of the full set.

image
image

bspeare:
@cory-skeers Thanks for posting the screenshots.   I think we might want to set fixed font sizes based on what layout column the block headline is within.  And s centered alignment option if we want editors to avoid creating a separate one column layout to create headings.  For instance, in one column layouts the heading would be sized larger. 

Screen Shot 2020-10-07 at 5 47 25 PM

cory-skeers:
@bspeare Are there fixed font sizes in place for any other components, and/or existing classes/variables to use, or will setting this need to be a new addition?

I'm also wondering about the center alignment option--would this be a single centered headline set above the content in the separate columns? My initial thought is then to provide a "section headline" option as well, where the section could have a headline, and each column/block could have a headline. But that also feels like blowing up the headline options...

bspeare:
@cory-skeers I think we might want to have the alignment informed by https://github.com/uiowa/uiowa/issues/2309.  

For the font size, perhaps we could start by setting it to `2.2rem` initially?

Screen Shot 2020-10-08 at 4 11 37 PM

config/default/field.field.block_content.uiowa_articles.field_uiowa_block_headline.yml
```diff
@@ -0,0 +1,25 @@
...
+  config:
+    - block_content.type.uiowa_articles
+    - field.storage.block_content.field_uiowa_block_headline
+  module:
+    - layout_builder_custom
+id: block_content.uiowa_articles.field_uiowa_block_headline
+field_name: field_uiowa_block_headline
+entity_type: block_content
+bundle: uiowa_articles
+label: 'Block Headline'
```
pyrello: I think this could just be called 'Headline'.

pyrello: And in any other place this label is used also.

config/default/layout_builder_styles.style.headline_all_caps.yml
```diff
@@ -0,0 +1,12 @@
+uuid: 62c70a56-e25d-4e66-9e2c-ed4b1cb5559d
+langcode: en
+status: true
+dependencies: {  }
+id: headline_all_caps
+label: 'Headline all caps'
```
pyrello: I'm going to adjust these labels in #2075. This will need to be adjusted after that is merged.

bspeare:
@cory-skeers are we able to target the heading field with a `block-title` class so that we can modify that size to `2.2rem`?  Would we add that here: https://github.com/uiowa/uiowa/pull/2152/files#diff-77fb9c8aa4d0580e6b5f78be4c85c98a? 

The other thing I noticed is that the heading option says "Bold Headline highlight" when it should be "Bold headline underline".

pyrello:
Does it need to be `block-title`? It might be good to name it something more generic in case this is used outside of blocks.

cory-skeers:
@pyrello It doesn't need to be named block specific. It might help to have a specific label or description in some places (collections, articles, etc.) to make its relationship to the children elements clear, but we can add that as needed--might be more clear on a per-block basis, anyway.

Any votes for replacement for `uiowa_block_headline` to namespace it outside of block? `uiowa_headline`, `uiowa_title`, or others.

pyrello:
I think the class should just be `headline`. I'm pretty sure that shouldn't conflict with anything internal and would be consistent with how we will eventually rename all the headline classes to `headline--bold--` style names.

cory-skeers:
Reworked machine names, field labels, etc., updated headline highlight to underline in label, and added a default `headline` class for scoping styling.

@bspeare Right now it sets headlines using the field to `2.2rem` unless it's in a one-col section, where it is set to `2.3rem`, to check that scoping works as intended.

richardbporter:
Is it possible to use this field programmatically? The aggregator block is provided by the uiowa_aggregator module in #2296 so I can't add the field through the UI.

pyrello:
I'm working on fixing the merge conflicts.

richardbporter:
If this is ready to go, I can work to add a custom render element after this merged as part of #2296.

richardbporter:
I guess I'm going to rebuild #2296 using the field API unless someone thinks otherwise. 😢

cory-skeers:
@richardbporter Sorry it took awhile to see this!

The field should be usable to be added programmatically with this--it gets defined as a complete, separate fieldtype (uiowa_headline).

From my standpoint, this is done other than review and testing (and possibly updating stylenames following the banner_block_refactor PR). Some things like a centered section option, or fine-tuning sizing based on column placement can continue to be added/adjusted after merging.

richardbporter:
My understanding is that its not possible (or easy) to use the field API in the forms API. I think we'd need to implement a custom form element using https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21FormElement.php/class/FormElement/8.9.x - like the address module is doing.

In the interest of time, I rebuilt the aggregator block using the field API. I was able the get this field working without any major issues in #2296. Nice job! 👍

richardbporter:
Did anyone test the update hook? Are there specific sites to test with the hook? Otherwise, I'd like to get this merged for #2296  if nobody has other changes.

joewhitsitt:
haven't tested since 9 days ago. not going to be able to do it before this PM.

joewhitsitt:
>Are there specific sites to test with the hook?

Same question.

cory-skeers:
The default content on fresh installs can be used to check that the Articles and Events headlines get set correctly, as well as slider (although possibly less noticeable, since it'd get set as an h2). People and Collections, I'm not sure of any examples in the wild.

richardbporter (dismissed):
Just noticed that https://github.com/uiowa/uiowa/commit/2f8ab63464eec1aab47248ec234346838370b638 was somehow lost in I'm assuming a merge conflict resolution into this PR. We'll want to double check that and any other regressions.

richardbporter:
I got sitenow_update_8035 to run on the default site without issue. Didn't know exactly what to look for but heading size was set to what it was before. 👍

cory-skeers:
Shoot. Hopefully there aren't many regressions in here outside of that one.

@richardbporter If that worked, then we're good! All the update hook should do is set the appropriate heading level if one had been set prior to this new field. Just had a little extra roundabout to make sure the config gets pulled in with the right order to get it updated and avoid config errors.

richardbporter:
I can try to find the offending commit tomorrow and reimplement the changes.

richardbporter:
Looks like this was introduced with the banner block refactor https://github.com/uiowa/uiowa/commit/1fd282c7e025f8184b14493024b6b1f3c7230b5b#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4c. I guess I'll create a fix off main.

richardbporter:
Cleared my review as I found the regression was unrelated to this PR.

richardbporter:
On fresh install, i.e. `blt drupal:install`, I'm getting the following error when editing the layout of the front page:
```
Twig\Error\LoaderError: Template "@uids_base/uids/heading.html.twig" is not defined in "themes/custom/uids_base/templates/field/field--field-uiowa-headline.html.twig" at line 47.
```

I compiled frontend assets... Anyone know if a template changed again? Otherwise I can dig into it.

cory-skeers:
@richardbporter Looks like a few template updates were needed. Think I got them all.

richardbporter (approved):

PR #2163, opened 2020-09-11
Replace medium image style with square 384 in sitenow_events.

richardbporter:
Otherwise the non-existent medium causes a whitescreen.


pyrello (approved):
Reviewed, not tested. Probably in the long run we should think about making this responsive.

richardbporter:
If its as simple as using that machine name, I can change it now. What is it?

pyrello:
It's not that simple. It is a different array structure, I think.

joewhitsitt:
it isn't that simple. you are hardcoding a render array and i think it expects a single image style here not a responsive collection

PR #2164, opened 2020-09-11
Add upgrade status module

pyrello:
Its only enabled on local split. This will allow us to start working on D9 compatibility issues, as we have time.


richardbporter (approved):
I'm ok with this if people want to use a UI instead of the CLI. We are currently running drupal-check as included in BLT - its running in CI and can be run locally with ` blt tests:deprecated:modules` or ` blt tests:deprecated:themes`. We just need to address #931 to scan all modules/themes.

pyrello:
@richardbporter It is my understanding that this provides both a UI and a CLI and that the CLI provides more information than is available using the existing methods. I can take a look at those existing commands to see if there is additional value being provided here.

pyrello:
I get this message when I run those commands:
```
⚠️  You're using a deprecated config option bootstrap. ⚠️️

This option has been replaced with bootstrapFiles which accepts a list of files
to execute before the analysis.
```

richardbporter:
Yea I'm assuming that is fixed in a newer version of BLT. You should get output after that though since it is just a warning.

richardbporter:
Whoops! Wrong branch. :)

PR #2166, opened 2020-09-12
Feature: views_tree via composer

### LINKED ISSUE #1479: Replace views_tree module with something that is currently support and available via composer
The views_tree module was embedded with the project because it is a dependency of the uiowa_book module and is not built in a way that supports inclusion by composer. It currently has deprecated code errors, so I think it is time to start looking for a new module to handle this functionality.

pyrello:
Resolves #1479. 
views_tree is able to be added via composer, so it can be removed from `contrib_dl`.


richardbporter (approved):
Nice! Would you say https://github.com/uiowa/uiowa/issues/1479 is resolved with this then?

pyrello:
@richardbporter Yes, thanks for pointing out the open issue!

PR #2172, opened 2020-09-15
Update sitenow_events image styles and single event page.

### LINKED ISSUE #2170: Pull large version of event image from content.uiowa.edu
## Problem

We are currently pulling the medium image size offered by the API which is 220px on the longest side. We are then scaling and cropping to a square 1:1 with the 385px image style. This is causing "pixelization" all of the time even if the source is a 1:1 aspect ratio.

## Proposed Solution

Regardless of what we plan to do with the event style in https://github.com/uiowa/uids/issues/309, we should be pulling a larger version so that event images aren't scaled up. This way, if an event admin uploads a large 1:1 image, it will display correctly on our v2/v3 events listings pages.

Might need to purge image caches as part of this.

richardbporter:
Resolves #2170.


joewhitsitt:
@richardbporter since you are increasing the scope quite a bit, you want to fix this dark/black contrast issue in v2 with virtual events?

Screen Shot 2020-09-15 at 11 02 31 AM

richardbporter:
Is this new or always existed?

joewhitsitt:
virtual is new, so it is new and i missed it last round of fixes

richardbporter:
How do I reproduce that?

image

joewhitsitt:
events list display v2 with bg-dark or bg-black

richardbporter:
Thats what I captured above - the event cards are white.

richardbporter:
Can't figure out the random test failures here. Moving to #2178

PR #2178, opened 2020-09-15
Update sitenow_events image styles and single event page.

### LINKED ISSUE #2170: Pull large version of event image from content.uiowa.edu

richardbporter:
Resolves #2170 

### Testing
- Add an events block to a v3 site.
- Inspect the image and see it is responsive using the large image format.
- Go to a single event page and see the image is responsive using the original image.
- Provide feedback on different single event styles/layout.
- Double check HR.


joewhitsitt:
why did you move away from the 1:1 aspect ratio? isn't there a square responsive image? after the Slack convo the other day, I told a customer to start using 800px 1:1 images because we would most likely be sticking with 1:1 images (circle mask compatible) and that it would be big enough for events.uiowa.edu largest mobile width.

Also viewing a [single event on law](https://law.local.drupal.uiowa.edu/event/56456/0) I got: `Notice: Undefined index: description in template_preprocess_sitenow_events_single_event() (line 20 of modules/custom/sitenow_events/includes/sitenow_events.single_event.inc).`

HR seemed okay.

Not all images look nice when moved to main column of the single event. (event/55036/0)

richardbporter:
>why did you move away from the 1:1 aspect ratio? isn't there a square responsive image?

I was trying to be consistent with articles/people responsive image style - thats what they're using. Also, the content hub docs recommend 3:2 aspect ratio: https://content.uiowa.edu/photo-guidelines I don't know why the styles it generates in the API are squares then...

>Not all images look nice when moved to main column of the single event.

I agree but the same can be said about images in the sidebar. I also don't see the point of implementing responsive images when its presented in a tiny sidebar.

Undefined index is fixed with c5b17c6.

joewhitsitt:
Wow. Never saw those image guidelines before.. Okay so I get consistency with Articles, but the person photo on full/teaser is `Rendered as Large - Square`

The nice thing about the images in the sidebar was that it represented the photo if it existed but assumed it would look bad and didn't make it the main feature of the page.

I guess I can drop the 1:1 for now, but as we finalize the event styles (circle is still a thing), we might have to add cropping functionality and a true 1:1 option for the API on content.uiowa.edu

joewhitsitt (approved):

richardbporter:
Still need to fix that location teaser font color issue with dark backgrounds. Working on it.

I don't care what we do with images. I was more motivated by seeing really tiny images and not being able to tell what it was. A lightbox would be nice... Also, events.uiowa.edu presents the image in the main column. There is a special image style for that. I'll investigate.

PR #2204, opened 2020-09-21
Remove hawk alert behat test.

### LINKED ISSUE #2203: Alert behat test failing
https://travis-ci.com/github/uiowa/uiowa/builds/185666940#L3196

richardbporter:
Resolves #2203 

I tried upgrading drupal/drupal-extension and increasing the AJAX timeout but this test is probably not a good idea. The module should be refactored to allow mocking a hawk alert instead of attempting a network connection. We may still run into AJAX timeout issues eventually though with other JS tests.

### DIFF
--- a/tests/behat/features/Alerts.feature
+++ b/tests/behat/features/Alerts.feature
@@ -15,12 +15,3 @@ Feature: Alerts
     And I am on the homepage
     Then I should see the heading "TEST"
     And I should see "This is a custom alert."
-
-  @drush
-  @api
-  @javascript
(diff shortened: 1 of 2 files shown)


richardbporter:
Actually I wonder if the issue is that this isn't using the Drupal AJAX system, just $.ajax. In any case lets just remove this test for now until we have more time to investigate.

pyrello (approved):
Reviewed. Build passed: https://travis-ci.com/github/uiowa/uiowa/builds/185675802

PR #2160, opened 2020-09-11
Preparations for Drupal 9

pyrello:
Working to accomplish the steps from "Before upgrading" in these instructions: https://support.acquia.com/hc/en-us/articles/360045796154-Upgrading-from-BLT-11-to-BLT-12
* Upgraded to `drupal/core:^8.9.7`.
* Ran `composer update`.
* Used BLT's recipe to migrate to using `drupal/core-composer-scaffold` instead of `drupal-composer/drupal-scaffold`. This also has the effect of moving to `drupal/core-recommended` instead of `drupal/core`.
* ~Fixed a bunch of new phpcs errors related to arrays needing to be broken up over more than one line when the total line length was greater than 80 characters.~ - Fixed in #2216 

# To test
Review the list of changes below (specifically for paragraphs and webform) and test against sites accordingly. **_If possible, please post comments below about what you have tested or plan to test and what site(s) you are testing/have tested against._**

# Composer updates
Both with the switch to `drupal/core-recommended` and running `composer update` there are a lot of updated dependencies besides what is shown in `composer.json`. `drupal/core-recommended` locks Drupal's dependencies to specific versions, so a lot of those packages are actually downgraded in this update. There are also the following contrib updates:
* drupal/block_content_permissions: 8.x-1.9 => 8.x-1.10
* drupal/config_filter: 8.x-1.5 => 8.x-1.6
* drupal/config_split: 8.x-1.4 => 8.x-1.5
* drupal/date_recur: 8.x-2.0 => 8.x-2.2
* drupal/embed: 8.x-1.3 => 8.x-1.4
* drupal/features: 8.x-3.8 => 8.x-3.11
* drupal/google_tag: 8.x-1.3 => 8.x-1.4
* drupal/honeypot: 2.0.0 => 2.0.1
* drupal/inline_entity_form: 8.x-1.0-rc6 => 8.x-1.0-rc7
* drupal/lazy: 8.x-3.4 => 8.x-3.5
* drupal/linkit: 8.x-5.0-beta10 => 8.x-5.0-beta11
* drupal/masquerade: 8.x-2.0-beta3 => 8.x-2.0-beta4
* drupal/memcache: 8.x-2.0 => 8.x-2.1
* drupal/menu_admin_per_menu: 8.x-1.1 => 8.x-1.3
* drupal/menu_breadcrumb: 8.x-1.13 => 8.x-1.14
* drupal/mimemail: 8.x-1.0-alpha2 => 8.x-1.0-alpha3
[...]

PR #2216, opened 2020-09-23
Update drupal/coder to most recent version and fix new PHPCS errors

pyrello:
As part of the process of getting ready for Drupal 9, `drupal/coder` will be updated, which introduces some new PHPCS sniff rules. These new rules will cause builds to break. I discovered this in #2160. This PR is an effort to preemptively fix those issues before we have to, to simplify the future process and testing.

# To test
This PR shouldn't change any functionality in our custom code. Please review the [code changes](https://github.com/uiowa/uiowa/pull/2216/files) and test as you see fit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,5 +17,5 @@ insert_final_newline = true
 [Makefile]
 indent_style = tab
 
-[composer.json]
+[composer.{json,lock}]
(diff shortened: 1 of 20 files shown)


richardbporter (approved):
Thanks!

PR #2221, opened 2020-09-23
Allow webmasters to configure uids_base theme settings

### LINKED ISSUE #1935: Investigate Theme Permission module
# Problem
We would like to allow webmasters to control the settings for the current theme, but we don't want them to be able to do anything else related to theme permissions, but the current theme permissions only cover all of the above.

# Proposed solution
Use the https://www.drupal.org/project/theme_permission module
* Create a new feature branch
* `composer require 'drupal/theme_permission:^2.0'`
* `blt ds --site=default`
* `drush @default.local uli`
* Set permissions for webmasters to only be able to administer `uids_base` and save
* Export default config: `drush @default.local cex`
* Search https://github.com/uiowa/uiowa/find/main for 'webmaster' to see which splits use a custom version
* For each instance listed, add the new permissions that were added to `config/default/role.webmaster.yml` to each instance of `role.webmaster.yml`.

### LINKED ISSUE #2222: Disable access to global theme settings
In looking into #1935, it seems that the theme_permission module alone will not be enough to lock down theme settings enough. We will also need some custom code to accomplish the following:

1. Disable access to the Global settings tab:
  Screen Shot 2020-09-23 at 10 23 17 AM
2. And we need to disable access to global settings on the UIDS base tab:
  Screen Shot 2020-09-23 at 10 24 15 AM

ethanstjohn:
Enabled the Theme Permission Module on the default site, went to permissions and ticked the box for Webmaster, administer themes uids_base. When masquerading as a webmaster I couldn't see any theme permissions enabled.

Resolves #1935
Resolves #2222 

## To do
* ~~Copy permission updates to various `user.role.webmaster.yml` configs in different splits as described in #1935.~~
* ~~Complete #2221~~

## To Test

As a webmaster for all sites except for HR, I am able to make changes to the uids_base theme settings but cannot configure other themes, enable/disable, change global theme settings.


joewhitsitt:
keep in mind that permissions need to be addressed in several splits and this only changes it for v2 defaults sites currently.

pyrello:
@joewhitsitt This doesn't cover those cases yet or even the default config yet. I just wanted this up so I could take a look. Then we'll account for the remaining cases.

pyrello:
I've checked this by syncing the default site. The permission appears to be set up correctly. But when I masquerade as a webmaster, I cannot access any theme settings at all. I opened (in separate tabs) the Appearance page, the settings for UIDS Base, and the settings for Claro. After masquerading, I reloaded each tab and all off them gave me the 'Access denied' page.

Not sure how to move forward from here.

joewhitsitt:
My bet is that the theme settings permission (core) needs to be on, but then we need to make sure uids_base is the only one with the contrib permission

pyrello:
I tried adding a patch (https://www.drupal.org/project/theme_permission/issues/3105637) to see if that caused it to work any differently. Same result, as far as I can tell. This module just doesn't seem to work with our platform.

pyrello:
@joewhitsitt I'll check that out.

pyrello:
That was it. However, this only gets us part of the way to what I think we need.

In addition to this module, we also need to disable access to the Global theme settings, which is still accessible:
Screen Shot 2020-09-23 at 10 23 17 AM

Also, we would want to remove access to the other settings on the UIDS Base settings page:
Screen Shot 2020-09-23 at 10 24 15 AM

pyrello:
For this PR to be viable, it needs to address #2222.

pyrello:
I updated the theme permissions for webmaster in the default config and the sitenow_v2 split. There is also an unrelated config update for the minimal filter in sitenow_v2, which is no longer different than default, so it is removed.

joewhitsitt:
Here are the split webmaster permissions that would probably want to have this permission:

config/default/user.role.webmaster.yml
 config/uiowa.edu/user.role.webmaster.yml
 ~~config/hr.uiowa.edu/user.role.webmaster.yml~~ _on uiowa_bootstrap and we don't want them changing stuff._
 config/brand.uiowa.edu/user.role.webmaster.yml
 config/features/collegiate/user.role.webmaster.yml
 config/features/sitenow_v2/user.role.webmaster.yml
 config/coronavirus.uiowa.edu/user.role.webmaster.yml
 config/policy.clas.uiowa.edu/user.role.webmaster.yml
 config/ighn.international.uiowa.edu/user.role.webmaster.yml

joewhitsitt:
https://github.com/uiowa/uiowa/pull/2221/commits/fd994c76d87566083ec6c614143fe48a847e9074 removes global settings
Screen Shot 2020-09-24 at 4 43 13 PM
Screen Shot 2020-09-24 at 4 43 04 PM

joewhitsitt:
Probably also want to disable favicon and logo settings?

pyrello:
I think so. I'm not actually sure what would happen if they changed it. The favicon might change, but the logo is statically defined in the theme.

joewhitsitt:
@mark-bennett-uiowa does clas.policy want webmasters to have the ability to adjust uids_base theme settings? If so, I can add.
`config/policy.clas.uiowa.edu/user.role.webmaster.yml`

mark-bennett-uiowa:
> @mark-bennett-uiowa does clas.policy want webmasters to have the ability to adjust uids_base theme settings? If so, I can add.
> `config/policy.clas.uiowa.edu/user.role.webmaster.yml`

@joewhitsitt No, not especially. I will go with the consensus here, and only allow admins to adjust the theme. Is that what the general consensus appears to be?

pyrello:
@mark-bennett-uiowa This PR is actually making the default behavior for the platform to allow webmasters the ability to control theme settings. But it is limited to just the theme settings for UIDS Base that are relevant to that theme.

mark-bennett-uiowa:
@pyrello Then sure, go ahead and add it @joewhitsitt , thanks! I misunderstood the purpose of this PR based on the title, my apologies.

joewhitsitt:
I think Ethan is out today, but IMO, this PR is ready for review.

pyrello:
I think we should disable access to the "Layout options" fieldset too and possibly remove that option altogether in a follow-up issue. We have not kept up with accounting for the different container options at the theme level and the two non-default options don't seem to do anything to the header and footer.

pyrello (changes requested):
Disable access to "Layout options".

pyrello:
Re-checking this.

pyrello (approved):
Reviewed file changes and tested on immuno.grad.uiowa.edu and coronavirus.uiowa.edu.

pyrello:
@joewhitsitt Nice work on this PR and glad to have your knowledge to be able to solve it so quickly. This is an important and solid update to the platform. And we're only about 4 months late to getting it done :)

PR #2223, opened 2020-09-23
Develop

GaryRidgway:
This will make the image visible on IE10+11. it doesnt make it work exactly as we have it in chrome, but the goal was useability.


GaryRidgway:
this feels like a lot of commits... did i do something wrong here?

PR #2229, opened 2020-09-23
Hotfix: Allow some self-closing tags in check for whether a region is empty

pyrello:
This is to fix issues from #2076 

The change is to a check that attempts to determine whether a region is empty. Previously we were stripping all tags, which meant that in cases where there was content that was only rendered as tags, it was getting skipped by the check.


cory-skeers (approved):
Didn't test, but looks good! 👍

PR #2230, opened 2020-09-24
I2149 - LB Tours Update

### LINKED ISSUE #2149: Update tours to handle v2/v3
## Problem

Our tours right now are v2 specific. There are targets in layout builder/v3 that we can target to provide guidance on.

## Proposed Solution

- Update content overview to highlight layout edit operation if it exists.
- Target the layout tab on node page view.
- Layout Builder general tour

dnkemper:
Resolves #2149 

## To Test

- [ ] do a fresh v3 site install of the default site
- [ ] Add a page - make sure tours button is visible
- [ ] Save the page - make sure tours button is still visible
- [ ] Edit the page - make sure tours button is still visible/functional 
- [ ] Click layout tab - make sure tours button is visible/ functional
- [ ] Add Person - make sure tours button is visible/functional
- [ ] Add Article - make sure tours button is visible/functional
- [ ] Edit Google Analytics - make sure tours button is visible/functional


joewhitsitt:
pulling down to look at:

>when I try to reference one block field in the tours, it will reference the other blocks that use the same field too… any idea on how I can prevent this since the blocks don’t contain any unique classes? It’s happening with events, people and the articles block

joewhitsitt:
So, I looked a this thinking it would just be simple to [string some classes together](https://github.com/uiowa/uiowa/blob/main/config/default/tour.tour.content-add.yml#L129) to scope this, but like you said, there are no unique classes or id's for the block edit form... Ridiculous...

My suggestion would be to try and create non-target items for these and weight the tour so that they fall in the right order regardless of which block is open for edit. If that doesn't work, it might be better to just keep block specifics out of the tour and point to the docs. We didn't do this for v2 paragraph blocks and additional maintenance will be needed to keep both docs and block tours up to date.

I realize this is still draft/wip but I noticed that layout tour item displays for a v2 paragraph site. layout builder is technically on for all sites, but editor/webmaster ability to interact with it doesn't exist in v2. Unless there is something to scope it to that only shows for v3 sites, I think these should be removed.

Screen Shot 2020-09-30 at 11 27 21 AM
Screen Shot 2020-09-30 at 11 31 06 AM

Thanks for fixing some of my poor writing with the existing tours.

dnkemper:
@joewhitsitt thank YOU for all of your help and for looking into this for me! I have two more appointments today but will start looking into it when I get back home. Thank you again!

dnkemper (commented):
- [x] Note to self: Line 63 on articles settings form- change people verbiage to articles

- [x] Content-add tour - line 80: put browse media and following verbiage in quotation marks

- [x] Question to ask: Social media wise, are we including Tik Tok?

- [x] Add tour for Google Analytics

dnkemper:
Question to ask during standup: Label tour items "For SiteNow v3" or should I just delete it all together?

joewhitsitt:
pulling down to test

joewhitsitt (approved):
In terms of test criteria, I approve. I might suggest a tweak to Google Analytics tour in the future. We don't want to push the campus-wide data option as much as we want site owners to handle their own analytics.

As a webmaster on a v2 site, I am still seeing the layout builder stuff on page view:
Screen Shot 2020-10-08 at 8 54 57 AM

Whatever you did for the page edit to hide the layout builder stuff on v2, can that be done for this tour as well?

PR #2231, opened 2020-09-24
Fix for recent config status check failures

pyrello:
Re-exported config for a few sites that have recently created `config:status` warnings during deployment.

Includes removal of minimal filter format from sitenow_v2, which is no longer different than default.


joewhitsitt:
pulling down

joewhitsitt (approved):
tested on avclub and a few other existing v2/collegiate sites.

Issue #2186, opened 2020-09-17
Revisit spam submissions again


joewhitsitt:
## Problem
Seeing an uptick in spam submissions coming into its-web.

## Proposed Solution
Revisit timeout or check in on newer/better solutions?

joewhitsitt:
https://github.com/uiowa/uiowa/issues/798#issuecomment-617894471, https://www.drupal.org/project/antibot, https://www.drupal.org/project/recaptcha

limit submissions per day

relates to https://github.com/uiowa/uiowa/issues/2200, https://github.com/uiowa/uiowa/issues/798

joewhitsitt:
Closing for now. New spam protection deployed. Now we wait...

PR #2235, opened 2020-09-24
Webform spam adjustments

joewhitsitt:
Relates to #2186 

- Adds antibot module which requires javascript for forms in addition to the current honeypot protection
- Prevents duplicate submissions by disabling submit button immediately after submit.

I would be curious what kind of/if any analytics show JS is enabled for our visitors.

## To Test 

Only way to really test this is to put it in front of the bots that are out there in the wild. Testing locally, make sure as a logged out user, you can still interact with forms and that you aren't restricted _too much_ from using them.


bspeare:
@joewhitsitt is there a site with a lot of webforms that you would recommend testing this on?

joewhitsitt:
i told Alan about the homepage. sitenow has a support form

GaryRidgway (approved):
I have pulled this down and tested a webform. I was able to fill it out and submit with no problems.

PR #2236, opened 2020-09-24
Some clean-up after LB styles PR

pyrello:
* Fixed some logic that detects the title block in a layout section and add a class so that it also works on articles.
* Removed the spacing on top of article dates in full mode.
* Commented out some rules that should no longer be active since the 'title-invisible' class was removed.


bspeare (approved):
I tested this on a few sites and it looks like it fixed the spacing issue for articles.  Thanks!

PR #2239, opened 2020-09-24
Remove access to spacer component.

pyrello:
This update adds the same restrictions to the spacer and page title hero blocks to the collegiate split and the coronavirus site.


joewhitsitt (approved):
reviewed, not tested. looks similar the work done for v3

PR #2254, opened 2020-09-28
Footer Block Edit Link

### LINKED ISSUE #2201: Webmaster role not able to edit footer text area?
## Problem
When the footer is empty the block does not display and doesn't show the contextual link. That is currently the only way we are allowing editors to navigate to the block to edit.

The block ID is not always the same to make it a bit more complicated.

## Proposed Solution

Need to figure out an identifier since the block id is not always the same.

Create a link somewhere that we can rely on regardless of whether or not the block is empty. Basic site settings follows the pattern we have established, but with https://github.com/uiowa/uiowa/pull/2221, we could put it in theme settings or add a link to the structure or configuration tabs.

joewhitsitt:
Resolves  #2201

## To Test

Pull and visit /admin/config/system/site-information on a couple uids_base sites. Early collegiate sites didn't have the footer contact block as id = 1 so you could pull down the homepage, etc.
Regardless of what site, as a webmaster you should be able to see the link above footer menus, click it and be taken to the footer contact block that serves content to the footer of the website.

This allows webmasters to get to the footer block even if the contextual filter link on the frontend is missing (empty block).

## Follow-up

Update footer documentation. https://sitenow.uiowa.edu/documentation/basic-site-settings-and-footer-configuration


cory-skeers:
Tested on americanstudies.local.drupal.uiowa.edu and uiowa.edu, and the functionality worked well for both.

On submitting edits, the user is taken to admin/structure/block/block-content which is an access denied for webmasters. The edits are saved without any issue, however. (Editing the block through the contextual link on the page returns the user to that page, so there is no access issue.)

Sitenote: The restriction on webmasters deleting the footer contact info block also works well, but cancelling from the page leads to the same admin/structure/block/block-content.

joewhitsitt:
First off, thanks for saving the day @cory-skeers, I could not figure out that build issue.

I might look into a route redirect for non-admins from admin/structure/block/block-content

joewhitsitt:
embarrassing coding standard commits included...

joewhitsitt:
Sorry, didn't do the convert to draft and re-trigger reviews. Added a destination parameter to kick edit/cancel to the front.

cory-skeers (approved):
Tested it on a few sites, and it worked, including the destination parameter when going from the basic site settings.

If you click the `Delete` tab, whether you choose `Cancel` or come back to `Edit` and then save, the destination parameter is lost and you're returned to /admin/structure/block/block-content. Seems like a separate followup issue in my mind, though, and the main purpose of this PR is still accomplished.

PR #2068, opened 2020-08-21
Feature: Layout Builder default content

pyrello:
Resolves: #1925 

See this comment for what needs to be done to finish the work on this: https://github.com/uiowa/uiowa/blob/1277ebbc13d305a1cfea42a7c7c937e14f249348/docroot/profiles/custom/sitenow/content/block_content/0c0c1f36-3804-48b0-b384-6284eed8c67e.json#L12

# To test
* blt drupal:install --site=cogscilang.grad.uiowa.edu
* Confirm that the default content installs and that there are no visible issues with it

PR #2270, opened 2020-09-30
Update drupal/layout_builder_lock to 1.0.0.0-RC5.

### LINKED ISSUE #2269: Non-admins can configure custom header blocks but receive access denied
https://www.drupal.org/project/layout_builder_lock/issues/3165130

richardbporter:
Resolves #2269 

### Testing
- On a stock v3 site, e.g. `blt dupal:install` the default site.
- As an admin, add a block to the header section (which is locked) of a page.
- Add a webmaster user and masquerade as that user.
- Edit the layout of the test page and see that the Configure button is visible.
- Open the console and click configure.

Repeat the steps above on this branch after running `composer install` and see the Configure button is no longer visible.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5636,17 +5636,17 @@
         },
         {
             "name": "drupal/layout_builder_lock",
-            "version": "1.0.0-rc4",
+            "version": "1.0.0-rc5",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/layout_builder_lock.git",
-                "reference": "8.x-1.0-rc4"
+                "reference": "8.x-1.0-rc5"
             },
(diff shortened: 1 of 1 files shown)


joewhitsitt:
testing in hopes it can go out today... or early tomorrow with everything else.

joewhitsitt (approved):
Nice test steps. Works!

richardbporter:
I'm a bit confused in relation to #2068 - this fixes those errors but is that desired behavior for webmasters?

joewhitsitt:
if it isn't header they should be able to edit right?

richardbporter:
Yea, lets go with. Its certainly better than the error.

PR #2279, opened 2020-10-01
Add right margin to card list media.

### LINKED ISSUE #2227: Fix padding around images for card lists in uids_base
# Problem 

#2076 may have created some padding regressions with card lists where media images are right up against the border on v3. 

94057836-cbde5c00-fda5-11ea-8c5b-2a30053bb228

# Proposed Solution

Adjust card list padding to create space for media images.

richardbporter:
Resolves #2227 

### Testing
- On a v3 site, create a page with events, people and article lists. 
- See that card--list media does not edge up against the right side of the container.
- Shrink viewport to small breakpoint to see media has no margin.


bspeare:
Pulling this down to test.

bspeare:
I added a rule to only add the margin if a background class is applied.  I tested this locally and it works on all list displays mentioned.

bspeare (approved):
I tested this locally and it works on all list displays mentioned.

richardbporter:
Thanks

Issue #2220, opened 2020-09-23
[Admissions Redesign - Core] - Blog and Video Feed from existing WordPress Site


mjoneill:
Blogger/Video new posts (pulls from feeds, shows student images and media)

mjoneill:
* http://blog.admissions.uiowa.edu/natalie/2020/09/masks-necessary-and-fun/ this is the source of the content
* Client does NOT need the content to live on admissions.uiowa.edu. It is fine for it to live on the blog site for now.  So first pass is that the content index view (three pieces of content that live elsewhere) are shown on this site.

bspeare:
**Additional notes:**

- Blog feed listed here is a view of "Blog teasers" created from Feeds Importer: https://admissions.uiowa.edu/#block-views-student-blogs-blogs
- Additional feeds listed are listed at https://admissions.uiowa.edu/admin/config/services/aggregator
- XML feed example: http://blog.admissions.uiowa.edu/adam/feed/
- D7 site uses core Aggregator module and D8 still has this module. 
- Investigate whether these feeds can be placed via Layout Builder.  I just tried adding an Aggregator block and had Ajax errors. 
- Is it worth adjusting these feeds to JSON or will XML work if we want to use decoupled Vue.js blocks to display this content?

bspeare:
List of current feeds:

- http://blog.admissions.uiowa.edu/andre/feed/
- http://blog.admissions.uiowa.edu/cori/feed/
- http://blog.admissions.uiowa.edu/ethan/feed/
- http://blog.admissions.uiowa.edu/faith/feed/
- http://blog.admissions.uiowa.edu/grace/feed/
- http://blog.admissions.uiowa.edu/kelsey/feed/
- http://blog.admissions.uiowa.edu/natalie/feed/
- http://blog.admissions.uiowa.edu/riya/feed/
- http://blog.admissions.uiowa.edu/counselors/feed/

briand44:
Fields that need displayed:

- Author Image - This currently is not in the feed. It is a content type on the admissions site. If we don't import these as entities we would need some way of grabbing an image from a specified file path.
- Title of blog
- Teaser - This is being truncated to a 100 characters.
- Author - only using first name, could be an issue in future.

bspeare:
Research spike:

- Investigate feed options
- Investigate ability to attach author via javascript
- Investigate front end / back end integration with Drupal
- Set config split for admissions.uiowa.edu

richardbporter:
[...]

PR #2284, opened 2020-10-01
Low hanging fruit

### LINKED ISSUE #1630: Improve unique id functionality to actually do something useful
## Problem

Customer pointed out that the unique id field is basically useless unless you are a power user. I had to write out the steps to successfully use the unique id field for an anchor link:

```
When you enter a unique ID, on render, the site converts the value to sanitized identifier (no spaces/special characters, lowercase). "This is my anchor!" becomes "this-is-my-anchor"

If you are linking to an anchor within the same page on SiteNow v2:

Create the unique id first and find it in the page by inspecting or copy it by using an already sanitized value.
Create the link that will send the page to the anchor.
Type the link text
Select the text
Click the link button
Type # followed by the unique id in the URL field within the link dialog window.
Save the link.
If you are linking to an existing anchor on a different page within SiteNow v2:

Create the link that will send the page to the anchor.
Type the link text
Select the text
Click the link button
Start typing the page title and autocomplete from the results.
This will turn the URL value into the permalink id path.
Append # followed by the unique id to the permalink path in URL field within the link dialog window.
Save the link.
```

The main problem is, you can type anything in to the field (special characters, etc.) and it saves and displays to the editor as such. On render however, it could be a completely different value. I think a first step would be to either dynamically change the help text to display the clean identifier that will be used and potentially saved to the db. On top of that, it would be nice to have these available for selection via https://github.com/uiowa/uiowa/issues/1412

Also, would it make sense to display the unique id in the paragraph preview if it exists?

Ultimately, this was something I put in to allow for better a11y for those that wanted to use it and to have it available day one when people started adding content. I forgot to revisit.

### LINKED ISSUE #2276: Typo in Google Tag Manager alert message
Screen Shot 2020-10-01 at 9 59 46 AM

### LINKED ISSUE #2278: Lower honeypot minimum time
## Problem

I believe the timer starts when the user starts interacting with the form. The minimum is currently 15 seconds which is a long time if the form is only collecting a few fields (newsletter signup name, email). This is causing poor ux for real users.

## Proposed Solution

With javascript required added-on now to prevent spam, maybe we can lower this limit a bit.

### LINKED ISSUE #2280: Unique ID logic needs to be changed
## Problem
We have been using `Html::getUniqueID` for this. https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AgetUniqueId/8.8.x

This is causing even more confusion with editors because anchors are getting a --2 appended to it.

https://github.com/uiowa/uiowa/search?q=Html%3A%3AgetUniqueId%28%24paragraph-%3Efield_uip_id-%3Evalue

## Proposed Solution
Create clean css type class instead. https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AgetClass/8.8.x

Consider working on https://github.com/uiowa/uiowa/issues/1630 to provide the user info about what the id will ultimately be.

joewhitsitt:
Resolves #2276
Resolves #2278 
Resolves #2280 
Resolves #1630

## To Test

- Confirm that warning message on Google Tag manager container settings page is free of typos: default.local.drupal.uiowa.edu/admin/config/system/google-tag/add
- Confirm as an anonymous user, you are able to submit a form after 5 seconds of interaction with said form.
- Look for unique ids to not have --2 or something appended to them but that they are still free of spaces, special characters. (e.g. https://sitenow.local.drupal.uiowa.edu/policies)
- clean unique ids are displayed in help text after field change
- clean unique ids are saved as value and displayed in paragraph previews.

### DIFF
--- a/config/default/honeypot.settings.yml
+++ b/config/default/honeypot.settings.yml
@@ -6,8 +6,8 @@ unprotected_forms:
   - honeypot_settings_form
 protect_all_forms: false
 log: false
-element_name: url
-time_limit: 15
+element_name: link
+time_limit: 5
 expire: 300
 form_settings:
   user_register_form: false
--- a/config/features/sitenow_v2/core.entity_view_display.paragraph.accordion.preview.yml
+++ b/config/features/sitenow_v2/core.entity_view_display.paragraph.accordion.preview.yml
@@ -18,6 +18,14 @@ targetEntityType: paragraph
 bundle: accordion
 mode: preview
 content:
+  field_uip_id:
+    type: string
+    weight: 1
+    region: content
+    label: inline
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
   field_uip_title:
     type: string
     weight: 0
@@ -31,5 +39,4 @@ hidden:
   field_uip_classes: true
   field_uip_colwidth: true
   field_uip_display_title: true
-  field_uip_id: true
   search_api_excerpt: true
--- a/config/features/sitenow_v2/core.entity_view_display.paragraph.accordion_item.preview.yml
+++ b/config/features/sitenow_v2/core.entity_view_display.paragraph.accordion_item.preview.yml
@@ -23,7 +23,14 @@ content:
     third_party_settings: {  }
     type: string
     region: content
+  field_uip_id:
+    type: string
+    weight: 1
+    region: content
+    label: inline
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
 hidden:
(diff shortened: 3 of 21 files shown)


richardbporter:
+1 on this. Have a ticket waiting for instructions on how to use anchors. Do we have docs on this? If not should we add some after this is merged? Currently working on #2220 but can maybe test later.

richardbporter (approved):
Works as described.

PR #2294, opened 2020-10-05
Export config after default install.

richardbporter:
Didn't create an issue for this but can someone run a default install locally and see if they get a config error on main? If so, this PR should fix it. 

### Testing
 On main, run:
`drush sql:drop`
`drush cr`
`blt drupal:install`

If you get an error, repeat above on this branch.

### DIFF
--- a/config/default/core.entity_view_display.node.article.default.yml
+++ b/config/default/core.entity_view_display.node.article.default.yml
@@ -34,7 +34,7 @@ third_party_settings:
           label: Header
           layout_builder_styles_style: {  }
         components:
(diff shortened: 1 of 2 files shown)


richardbporter:
Based on the CI fail, I'm guessing this my own local environment issue.

pyrello:
@richardbporter I was seeing this recently and it looked like maybe the format of the yaml config lists had changed. I'm not sure why this happens in some cases.

richardbporter:
I just ran it a third time and no errors. 🤷

PR #2296, opened 2020-10-05
Enable aggregator and allow feeds to be placed via Layout Builder.

### LINKED ISSUE #2220: [Admissions Redesign - Core] - Blog and Video Feed from existing WordPress Site

### LINKED ISSUE #2319: Update aggregator feed design to use the event template list design that #2255 is using
# Problem 

Using https://github.com/uiowa/uiowa/pull/2255, map the aggregator template created in https://github.com/uiowa/uiowa/pull/2296 to use the new event template list styles in that PR. 

Related to https://github.com/uiowa/uiowa/issues/2220.

richardbporter:
Resolves #2220, Resolves #2319 

### Testing
- Check out this branch and sync a v3 site.
  - Collegiate sites may have some confusion as the core aggregator block is available for placement.
- Go to Configuration -> Aggregator and add a few feeds. See #2220 for a list of feeds.
  - This feed has an image: http://blog.admissions.uiowa.edu/andre/feed/
- Run cron or import the feeds manually on the aggregator overview page.
- Place the custom combined feeds block via Layout Builder (now called RSS Feeds).
  - Verify items show in order of most recent.
  - Verify headline field works as expected with children.
  - Verify uploading a feed image displays next to items.
  - Verify that the "hide images" option will hide the feed images
  - Verify that the feeds are themed like the events component
  - Verify the no results text works as expected.
    - Requires deleting items from feeds.
- Test the aggregator pages and verify that they are lightly themed: https://default.local.drupal.uiowa.edu/aggregator, https://default.local.drupal.uiowa.edu/aggregator/sources, https://default.local.drupal.uiowa.edu/aggregator/sources/1

Do the above again as a webmaster to test permissions as that role. Verify you can't break anything.

Todo:
- [x] Investigate theming aggregator and aggregator items.
  - Can we make this look how we'd like?
- [x] Investigate a combined view of all aggregator feeds. Currently a route that lives at /aggregator
  - Can we alter the aggregator block or create a separate block to allow placement via LB?
- [x] Investigate aggregator feed block number of items starting at 2 instead of 1.

Assuming the above is possible, further todos include:
- [x] Lock down aggregator settings for webmaster to not change certain fields
  - Allowed html, import timings
- [x] Improve custom aggregator block to:
  - Allow single or all feed selection. 
    - Rename our block as part of this and remove the stock aggregator block from selection via LB.
  - [x] Add parent title field.
- [x] Theme aggregator and aggregator/sources routes. 
- [x] Update cron-related settings to match hourly default. Admins can override.


richardbporter:
Breaking change: swapped our image field for the standard RSS spec channel image. It works.

We need a way to only render the image if looking at "teaser" view mode. I'm confused about this because right now the view modes seem like default and summary but I'm not sure where summary is used. Also, our custom template doesn't seem to register for the single event feed block.

richardbporter:
The single feed block does not use the aggregator theme function. 🤦 We might be able to alter it or maybe just improve our custom block to allow single/multiple.

richardbporter:
Oh duh. We're selecting the default mode in our custom block. Probably not much reason to use summary then.

richardbporter:
Breaking change: removed core aggregator block from LB placement, renamed our custom block and configured for placement. Probably best to re-install a v3 site from scratch if you've already placed either block on a page.

I've created blank UIDS base libraries and attached them on to pages when the block is placed. There is also a separate library for aggregator feed pages which has different markup. We may not use those pages but just wiring it up in case.

richardbporter:
I'm going to look into improving our custom block configuration but was looking for assistance theming.

bspeare:
I wonder if we want to create a variant of the events component to utilize a design pattern like in the screenshot below for these blog posts. 

Screen Shot 2020-10-07 at 11 30 11 AM

richardbporter:
One thing to note that is feed images can't be required at this point. I think that was an issue with events.

bspeare:
I think we would maybe want some functionality to hide the image like with the events.

bspeare:
Also thinking that the work involved with https://github.com/uiowa/uiowa/issues/2306 can inform the styling of these feeds.

bspeare:
After implementing this on the Admissions dev site, it's time to set display styles within layouts.    We need a "Display Style" option on each block where the user can choose "List" or "Grid" list styles since there are use cases for list or grid. 

# Grid Style

- Grid style centered
- Grid style left-aligned

Screen Shot 2020-10-08 at 9 27 58 AM

# List Style

- List style centered
- List style left-aligned

Screen Shot 2020-10-08 at 9 31 34 AM

bspeare:
Or, maybe we just need a "Display Style" option on each block where the user can choose "List" or "Grid" list styles since there are use cases for list or grid within other layout options.

richardbporter:
Are we doing that somewhere already?

bspeare:
It was initially discussed in https://github.com/uiowa/uiowa/issues/1789.  I'll create a separate issue for visibility.

pyrello:
@bspeare It doesn't seem like we need two sets of styles for the orientation of the items, since the options are the same for grids and lists.

For all list type collections, we need just a left (default) and centered options.

bspeare:
@richardbporter are you able to set an image style on the local image field?

richardbporter:
Yes, used small__square which matches the external image style but this could be another one.

04051ca

richardbporter:
I'm thinking it doesn't make sense for anonymous users to access the aggregator pages. I removed permissions so only webmasters can access those pages. This is because the admin breadcrumbs take you to those pages. If we keep it like this, we should probably remove that listing item count config to avoid confusion with the block. Also, we'll probably get a pager request for the block which I had trouble with. I could be convinced otherwise if we think those pages are valuable. 

Example local URLs:
https://default.local.drupal.uiowa.edu/aggregator
https://default.local.drupal.uiowa.edu/aggregator/sources
https://default.local.drupal.uiowa.edu/aggregator/sources/1

They'll need some rough styling for webmasters and the same preprocess logic for the image as aggregator items.

bspeare:
@richardbporter Should we remove the option to place the "Aggregator Feed" block?  It seems like the "Aggregator Feeds" block will take care of single and multiple feeds.

richardbporter:
I think I did but if you're looking at admissions, I noticed that all blocks were available for placement. I'm guessing collegiate split sites are not restricting anything?

bspeare:
Ah, yes I was placing these on the admissions site.

bspeare:
@richardbporter does the feed source page output the uploaded image with the `feed-image` class for you?  On my admissions version it is missing that class:

Screen Shot 2020-10-12 at 2 55 21 PM

richardbporter:
It does not. I only added that to the image provided by the RSS feed because I wasn't sure what the markup was and how we'd target it via CSS. I don't think that class exists in the block either. Do you think we should add it to both? I think we could preprocess the field and add it in.

bspeare:
That's probably not necessary, I can just adjust the style scoping.   Thanks!

bspeare:
@richardbporter I'm done with https://github.com/uiowa/uiowa/issues/2319, whenever you want to put this out for review.

richardbporter:
Ok. Just wondering about the best way to get #2152 into this.

mark-bennett-uiowa:
@richardbporter pulling down for testing on interdisciplinaryprograms.prod.drupal.uiowa.edu.

richardbporter:
I think I jumped the gun on asking for a review. This will either need to be refactored as a block using the field API to make use of the heading field, or we'll need to provide a custom render element for the heading size field.

The former is probably faster but the latter might be more useful in the future.

mark-bennett-uiowa:
@richardbporter Looks like this is working, sort of. Keep in mind that this was run on https://interdisciplinaryprograms.prod.drupal.uiowa.edu locally. See screenshots below:
**(1) Here are the installed feeds, all of which have numerous entries, but not necessarily in the past few months. Manually "updated" the feeds, and notice that only 2 items were imported:**
Screen Shot 2020-10-14 at 2 20 47 PM
**(2) Here are the lightly styled aggregator pages:**
/aggregator
Screen Shot 2020-10-14 at 2 22 58 PM

/aggregator/sources
Screen Shot 2020-10-14 at 2 23 23 PM

/aggregator/sources/3
Screen Shot 2020-10-14 at 2 23 47 PM

**(3) Here is the frontpage of the DIP site above. The left column is one feed that does not have any imported items, the middle column is a feed with only one imported item (works), and the right column is a card item manually created and already in place on the DIP site as comparison. Notice the difference in styling, headers are not similar size, no result text needs reduced, and the entire "region" of feed item is clickable to remote source (can the title just be the clickable link??).**

aggregator_errors

richardbporter:
@bspeare I had to force push this up to resolve my diverged branch issue. Can you spot check styles are as expected?

bspeare:
@richardbporter sure, I'll test it now.

bspeare:
@richardbporter I've tested it locally and the styles look good.  Was there something you noticed that I should address?

richardbporter:
No, just wanted to make sure I didn't incorrectly resolve the conflicts I had locally or overwrite anyone else's changes.

richardbporter:
@mark-bennett-uiowa

If you post those feed URLs I could look into why there are not more items imported.

The heading level discrepancy should now be fixed with the use of the shared headline field from #2152. 

I don't think the UIDS card allows for just the header to be clickable.

mark-bennett-uiowa:
> @mark-bennett-uiowa
> 
> If you post those feed URLs I could look into why there are not more items imported.

@richardbporter Here ya go:
[https://globalhealthstudies.prod.drupal.uiowa.edu/news/feed](https://globalhealthstudies.prod.drupal.uiowa.edu/news/feed)
[https://internationalstudies.prod.drupal.uiowa.edu/news/feed](https://internationalstudies.prod.drupal.uiowa.edu/news/feed)
[https://latinamericanstudies.prod.drupal.uiowa.edu/news/feed](https://latinamericanstudies.prod.drupal.uiowa.edu/news/feed)

My guess is it's only pulling events in the past X days/weeks/months. The news is pretty spread out over time on these sites.

richardbporter:
Yea its controlled by the discard setting which is unfortunate. We'll have to create our own option set, which I thought about doing anyway. I'm thinking we replace "1 month, 3 weeks" and "3 months, 3 weeks" with 3 months and 6 months. That would get some of the posts in those feeds but not all. Would that work?

mark-bennett-uiowa:
No worries, and I think the 6 month option is perfect! Thanks @richardbporter

richardbporter:
Since the aggregator settings was essentially useless for webmasters, I removed their access to it, tracked the config and set the clear/import timing to 6 months. That should import a decent amount of news feed items in most cases. Sites can split/override if necessary.

I think I'm done here.

richardbporter:
The latest changes requires a cache rebuild.

joewhitsitt:
pulling down

joewhitsitt:
- As a webmaster v3 default, I don't see the aggregator settings.
- Number of items, it would be nice if it told me the max of 25 before I try to submit.
- OPML worked!
- Though "lightly themed," those  pages could use a bit more (e.g. container or padding, smaller feed image). These are access denied to non-admins

richardbporter:
> * As a webmaster v3 default, I don't see the aggregator settings.

Seems like you should. Just added access but administer seemed like it was already there: https://github.com/uiowa/uiowa/pull/2296/files#diff-637e5fac6952d1fe2855b01a564edaa212a4aa5ce3cc0ba5ee9f6d443089e95eR98

> * Number of items, it would be nice if it told me the max of 25 before I try to submit.

I added some field description help text. P.S. are there tokens for this?

> * OPML worked!

Cool! IDK what OPML is. Please note that is disabled for webmasters.

> * Though "lightly themed," those  pages could use a bit more (e.g. container or padding, smaller feed image). These are access denied to non-admins

I had not planned on letting anonymous users access these pages, at least not initially. I could be persuaded but wasn't sure if people wanted them indexed (the breadcrumbs are weird). We could expand permissions to authenticated users.

cory-skeers:
* The no results display has an extra ", minimal"--I'm guessing this is part of the text format showing up in the template.

image

* The no results text doesn't change according to the background, leading to black text on black background possibility.
* The headline doesn't adjust to background color (part of uiowa/uids#330
* Are the classes staying "events-card__*", or should they be separated? (Don't think there's any need to, just curious.)

The checklist of items to test went well (v3 default site), and I did everything masqueraded as a webmaster without any issues.

joewhitsitt (approved):
Tested webmaster access. Worked this time, not sure what happened last time.

My no results text is clean of formatter info.

OPML allows you to import multiple feeds at a time. (e.g. https://www.youtube.com/subscription_manager?action_takeout=1)

richardbporter:
Thats right. I disabled OPML for webmasters due to concerns about time outs or eating up too many resources. Ok to leave it admin only for now?

richardbporter:
> The no results text doesn't change according to the background, leading to black text on black background possibility.

Whats the best way to address this?

cory-skeers:
> > The no results text doesn't change according to the background, leading to black text on black background possibility.
> 
> Whats the best way to address this?

I think the _best_ way would be along the lines of SASS map like what was briefly discussed with regards to uiowa/uids#330; second best add styling to the "uiowa-aggregator-no-results" class with combinations for the various existing backgrounds; third best add the "events-card__description" class in addition to the "uiowa-aggregator-no-results" and piggyback off the styling that is in place there.

@bspeare or @pyrello, thoughts or preferences on this?

Edit: I'd be alright with merging this branch in and creating a separate a11y issue for addressing this--pretty low chance of a no-result RSS feed being added to a background where this is an issue prior to it being addressed, even if it's a separate issue.

bspeare:
@cory-skeers I'd prefer to create a follow-up issue to address this.  Perhaps while we are working on uiowa/uids#330.

richardbporter:
Thanks all. Merging this will close #2220 - is that ok? Seems like https://github.com/uiowa/uiowa/issues/2309 is still needed for admissions specifically. I would probably vote to close #2220 and create follow up issues as needed.

richardbporter:
Lets go with that. :)

PR #2303, opened 2020-10-07
Remove uiowa_blocks entry from #2009

### LINKED ISSUE #2009: Remove 'uiowa_blocks' module
After #2000 is merged, all the functionality that was provided by the `uiowa_blocks` modules will have been moved to other places. The block definitions themselves are maintained as part of the default config for the codebase and are no longer tied to the module. I think this would be as simple as just removing the entry from the `core.extension.yml` file.

ethanstjohn:
Removed uiowa_blocks entry from core.extension.yml, resolves #2009.


joewhitsitt:
@pyrello or @mark-bennett-uiowa do you know of sites that were using this that I can test against to make sure there aren't issues?

pyrello:
Since it was in the `core.extension.yml` file, it was technically enabled on all sites. The module contains no functionality, so I would think that you could test this against any site.

joewhitsitt (approved):
tested on policy.clas with no hiccups

richardbporter:
Updated PR description to auto-close linked issue.

PR #2307, opened 2020-10-07
Config updates

joewhitsitt:
Recent code/dependency updates have changed config in PROD and it hasn't be exported back.

Going through and updating where noticed.


joewhitsitt:
merged commits are not yet in PROD causing an issue. follow-up might be needed to update site and feature splits

PR #2314, opened 2020-10-08
Add character limit to uiowa_statistic prefix/suffix fields

### LINKED ISSUE #2240: Improve usability of prefix/suffix fields on stat component
# Problem

The purpose of the prefix and suffix fields are not exactly clear with how they are arranged on the edit form.    

# Proposed Solution

Rearrange the fields on the LB form so that the fields appear more connected. 

Screen Shot 2020-09-24 at 4 14 16 PM
Screen Shot 2020-09-24 at 4 10 11 PM

### LINKED ISSUE #2258: Add prefix/suffix maxlength validation to Statistic Block
## Problem
The stat component design can't handle the field 255 character limits available in the prefix/suffix fields.

## Proposed Solution

Look into custom validation function off form_alter to limit prefix and suffix to a TBD character limit. This will affect block saves going forward, but not existing blocks (grandfathered in until next save).

- Update help text to notify editors of limits.
- Update documentation/tours

cory-skeers:
Resolves #2258, resolves #2240 

# Test

- Sync a site, `blt frontend`
- Edit an existing stat block and create new stat blocks.
- Test that you cannot enter a prefix and/or suffix greater than 3 characters.
- For existing stat blocks with greater than 3 characters in either/both fields, they remain unchanged, but upon opening cannot edit/save until the new limit is enforced.
- Check the rearrangement, styling of the prefix/title/suffix field combo, and whether there are any hiccups or concerns


cory-skeers:
@bspeare Did you have a specific character count limit in mind for this? Also wondering if there's a strict character limit that's the same for both prefix and suffix, and if there's need to take both into account and further limit one depending on how the other gets used or if they can be handled completely separate from each other.

bspeare:
@cory-skeers I think 3 is a good character count.  @GaryRidgway @joewhitsitt can you think of any custom sites that you have built with really long prefix/suffix character counts?

cory-skeers:
Thanks @bspeare! Opening it up for review, since any existing stats that may have longer prefixes or suffixes won't break (just can't be updated without running into the validate), and the validate char count can be updated if needed.

If anyone has use cases for longer prefix/suffix values that aren't what we're trying to discourage, let me know and I can up the max!

bspeare:
The ux updates look great!   We can probably remove this `.` until we have more descriptions? https://github.com/uiowa/uiowa/pull/2314/files#diff-4ac0913b66c3cabf91f869fd2547561eR15

cory-skeers:
Oh weird! Glad you caught that extra period--leftover from an earlier edit.

dnkemper:
Looks great to me! I did notice that when you select "Alignment: Flex row center && Styles: Statistic: Horizontal" it doesn't make the statistic fields horizontal? Not sure if that even matters. BUT that's unrelated. GREAT JOB!
Screen Shot 2020-10-12 at 10 08 24 AM

dnkemper (approved):

Issue #2109, opened 2020-08-28
Views multiple contextual filters using ANDs does not work


joewhitsitt:
## Problem

The contextual filtering we use for articles blocks in both v2 and v3 doesn't work as expected when multiple arguments are given and we use the AND method of comma separating arguments.

You will notice the arguments don't work within the Views admin interface, so this isn't an issue with the custom code to build the view via paragraphs/blocks.

https://www.drupal.org/project/drupal/issues/1762770
https://www.drupal.org/project/drupal/issues/1248300

## Proposed Solution

- Wait for Drupal to figure it out and limit our views contextual filters to ORs only in the meantime.
- Provide an ANDs workaround like in Admissions' scholarships https://github.com/uiowa/uiowa/pull/2323/commits/6dd541be3b1038f5d9cf807727bdc53b0cf057c2

joewhitsitt:
This came up as I am working on Admissions scholarships. Multiple values for multiple arguments. Though the views help text says I can pass them as comma separated to achieve ANDs, I get the same sql query for both methods:
<img width="560" alt="Screen Shot 2020-10-21 at 1 52 16 PM" src="https://user-images.githubusercontent.com/4663676/96769464-ca07b880-13a4-11eb-8bea-309fa718c659.png">
<img width="1409" alt="Screen Shot 2020-10-21 at 1 52 00 PM" src="https://user-images.githubusercontent.com/4663676/96769462-c96f2200-13a4-11eb-961a-ee01bfce7848.png">

joewhitsitt:
I think this results in double results for people_list_block

joewhitsitt:
This is semi-related but adding more than one tag to the exposed filter tags on people_list_block causes duplicates: https://sandbox.prod.drupal.uiowa.edu/content-same-name

<img width="1071" alt="Screen Shot 2021-10-05 at 2 49 01 PM" src="https://user-images.githubusercontent.com/4663676/136092740-3a16c0e7-c688-41fb-a2a8-b30b43d49a37.png">

joewhitsitt:
Just tested again with the article block in v3. Multiple tags seem to be used as ORs in my test cases on sandbox.prod which is how they are strung together in code. This same code is being used for article paragraph and people paragraph (v2).

The help text is confusing:
>If selected, users can enter multiple values in the form of 1+2+3 (for OR) or 1,2,3 (for AND).

[...]

PR #2323, opened 2020-10-09
Scholarship Feature & Admissions Site Split/Customizations

### LINKED ISSUE #2214: [Admissions Redesign - Core] Scholarships
Scholarship content type and display (search?)

joewhitsitt:
Resolves #2214

- Establishes Scholarship content type feature split

## Custom scholarship additions for Admissions via site split:

- A single scholarships block view that takes multiple contextual filters on scholarship type/resident.
- Custom block to pull together the scholarships view block with easy UX for providing arguments. 
- Custom block "jump menu" leveraging the Forms API instead of the odd webform/javascript solution that currently exists. Allows configurable destination paths based on entity reference.
- Both blocks can be place where ever allowing content around it.
- Scholarship teaser displays all data per requirements in an accordion listing.

## To Test on v3 sites

- `blt dsa`
- `drush @alias.local config:import --source ../config/features/scholarship --partial`
- Enable the scholarships split admin/config/development/configuration/config-split
- As a webmaster, you are able to create scholarships. All the fields except type/resident are available. The permissions assume that this is a single extension of the v3 default. Adjustments might be needed for v2, collegiate or other combinations of splits.

## To Test On Admissions

- `drush @admissions.local config:import --source ../config/features/scholarship --partial`
- Enable the scholarships split admin/config/development/configuration/config-split
- `drush @admissions.local config:import --source ../config/admissions.uiowa.edu --partial`
- `drush @admissions.local cim`
- As a webmaster, you are able to create scholarships. All the fields from the issue are accounted for.
- As a webmaster, you are able to see Admissions Core blocks in the More.. section of layout builder and you are able to place and configure. The jump menu block takes you to the paths you have specified. The scholarships block displays scholarships in a working accordion that meets the filter criteria.


config/default/config_split.config_split.admissions.yml
```diff
@@ -0,0 +1,27 @@
+uuid: 1aa7f804-a75a-4201-bb26-1a50922ec5dd
+langcode: en
+status: true
```
richardbporter: Running `blt drupal:install` locally, I'm seeing the following error:
```
<li>The configuration synchronization failed validation.</li>
                  <li>Unable to install the <em class="placeholder">admissions_core</em> module since it does not exist.</li>
```

I think this is due to the admissions split being exported to the default config directory and enabled.

richardbporter (changes requested):

config/default/config_split.config_split.admissions.yml
joewhitsitt: good catch. i was treating it like a feature split...

pyrello:
@joewhitsitt I was hoping that the scholarship content type would be something that could be used outside of the Admissions site exclusively. For example, I think it could work for a page like this: https://www.grad.uiowa.edu/funding/fellowships/internal/lulu-merle-johnson

mark-bennett-uiowa:
@joewhitsitt Could possibly also use it here: [https://clas.uiowa.edu/students/scholarships](https://clas.uiowa.edu/students/scholarships) for individual scholarships like [https://clas.uiowa.edu/students/scholarships/ada](https://clas.uiowa.edu/students/scholarships/ada).

joewhitsitt:
I appreciate the interest in this work, however, I have been told to keep moving forward and that a follow-up issue should be made or a separate discussion scheduled. The scope/point value of the issue this PR seeks to resolve doesn't extend beyond the admissions website. This PR does not prevent a unified solution in the future.

joewhitsitt:
We have a card in parking lot to discuss further. Holding off on marking this ready for review.

joewhitsitt:
Per recent discussions, I am going to move the content type and fields to a separate feature split and remove any admissions assumptions for now.

cory-skeers:
Do we want to update the default formatting for any of the text fields to filtered_html instead of minimal? (Would have to handle it similar to how it gets updated for node body fields.)

Main thought had been formatting like bulleted lists might be desired, but looking at some example scholarships on the current admissions site, it doesn't look like it's necessary (or at least not by default). Just raising for visibility if there are other thoughts on default format.

joewhitsitt:
Though I think minimal for scholarship works here, I think it is going to become more of an issue as more functionality is added where should be filtered_html default. It would be nice to be able to scope on block content formatted fields instead of all formatted fields

cory-skeers:
@joewhitsitt I created an issue (#2377) to rescope how the default formatting gets set. The number of fields that would want a default minimal but need to allow filtered won't grow (at least not quickly), I wouldn't think, but content like this wouldn't be hurt by a default filtered, even if it's not strictly necessary. And having access to bulleted lists without changing text formatting would be nice...

cory-skeers:
I'm not seeing the jump menu block, either while masquerading or as admin. Checking to see if I have something borked with my local config.

joewhitsitt:
the jump menu block (under More..) is admissions split specific and requires the additional cim after partial import to turn on the module that supplies it.

cory-skeers:
The label and item for Contact (In the Scholarships block) don't get the styling for `<div "text-formatted"` that the fields do, and so they visually stand out from the rest. Not sure if it's something to be worked out now, or as part of a later scholarship content styling effort.
image

Otherwise:

- Tested pulling on a default install, and everything worked as advertised for what is available there
- Tested on admissions
  - Jump menu works great
  - Scholarship content works
  - Webmaster permissions are working for both the content type and the block types

**General thoughts, not necessarily needing addressing**
For the scholarship and resident type (particularly resident type), I'm guessing there's the possibility of a scholarship being available to more than one of each of those categories so they can't be single-select?

Filtering is a little confusing in the Scholarships block, but I think only when not using real scholarships. Selecting First-Year and Resident, for instance, could be interpreted as show me all First-Year scholarships as well as all Resident scholarships, not all First-year resident scholarships. BUT I don't think that would be a confusion for anyone working with the real scholarships data.

Similarly, it's not immediately clear that no checkboxes means all results, but in practice, it works as expected if the user picks only one type of filter (filter to resident, but no scholarship type filter gives scholarships of all types which are marked resident, vice versa). If all scholarships are needed, all filter boxes can be checked, as well.

**TL;DR**
Everything appears to be working as advertised. Just checking over things again for the sake of being thorough before giving an approval.

pyrello:
Pulling down to test.

pyrello:
I made a couple changes to bring the feature split portion of this in line with how we have other feature splits implemented.

joewhitsitt:
Pulling back to draft to fix a couple things

joewhitsitt:
This is coming up again: https://github.com/uiowa/uiowa/issues/2109 with this work. For now, it seems I am limited to ORs unless I can alter the query or perform secondary/tertiary passes to weed out ones that don't match. The help text should have the ORs mentioned. Training and properly tagging content will be the key.

joewhitsitt:
- Updated field__label in uids_base to match the style.

Screen Shot 2020-10-21 at 3 09 33 PM

joewhitsitt:
Made changes to the view query to treat the filter selections as ANDs instead of ORs. Selecting all the options, the content must match or no results are displayed. No selections, show all scholarships. Updated help text, provided comments and prepared for when ANDs might work in Core Views.

Feel free to review/approve from a technical perspective, but I think we want customer approval for this before it is merged to main.

joewhitsitt:
Based on recent feedback:

- [x] Add "Additional Text" field (place above contact)
- [x] Move type/resident on form below contact, make sure fields on form are in the same order as admissions' current frontend display.
- [x] Quick investigation in pulling in short description and value into accordion header.

joewhitsitt:
This is ready for review again. The admissions content/functionality specifics have been approved for initial deployment. Just need a technical review/approval.

pyrello:
I'm testing this.

pyrello (approved):
Appears to work as described. Nice work!

richardbporter (approved):

PR #2332, opened 2020-10-12
Remove layout builder override styles checkbox and related code.

### LINKED ISSUE #2311: Disabling layout builder style overrides does not remove styles
### Problem
If a block has style override classes applied, unchecking the "Override style options" and saving the the block does not remove said classes. 

cap

### Proposed solution
Update our layout_builder_custom code to remove styles if the box is not checked. Using states might be an option and/or custom validate/submit handlers.

richardbporter:
Resolves #2311 

### Testing
- Look at #2311 for a description (GIF) of the problem and how to perhaps reproduce locally.
- Check out this branch and configure a block via LB. 
- See that the original issue is resolved because this PR removes that checkbox entirely. Styles are added/removed just by editing the select lists.
- Test various blocks with different styles and ensure they are added/removed correctly.

This makes it easier for see what styles are applied and remove them in one step. The flip side is that the interface is slightly busier.

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
+++ b/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
@@ -7,7 +7,6 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\layout_builder_custom\LayoutBuilderPreRender;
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Reviewed the changes, and tested on uiowa.edu and a default install. Works as advertised. 👍

PR #2335, opened 2020-10-13
Hotfix: Default content fixes

### LINKED ISSUE #2330: Default content - Import homepage as nid 1
# Problem
When default content is created, we don't seem to have any control over which piece of content get nid 1. The is problematic because the homepage has logic attached to it that doesn't allow the user to delete it unless it is set to a different piece of content. Currently, the homepage is being imported as nid 2 during local development and nid 6 when installed in production. This means that the logic related to the homepage will either be broken locally or in production depending on what we set the default homepage nid to.

# Proposed solution
Figure out how to have the homepage be imported as nid 1, so it will be the same locally and in production and won't break related functionality.

pyrello:
Updates
* Two image filenames that were preventing them from being properly installed.
* Resolves #2330

# To test
* `composer install`
* `blt drupal:install --site=default`
* Check that home is nid = 1
* Build passed without image errors: https://travis-ci.com/github/uiowa/uiowa/builds/189778711


richardbporter (approved):
I changed my local MySQL auto_increment_increment to 5 and default content seems good after install. 👍

PR #2337, opened 2020-10-14
Removed slider block used in the v3 default content

### LINKED ISSUE #2331: Default content - Slider on homepage is un-editable after import
# Problem
After default content is created, the slider on the homepage cannot be edited via Layout Builder. It literally doesn't appear in the UI.

dnkemper:
Resolves #2331 

Believe this successfully removes the slider block from v3 default content. The last commit I made threw me off a bit (pretty sure that block content wasn't linked to other existing content? Then again I've been up all night so maybe I'm over analyzing things) but when I imported a fresh install of the default site everything worked as expected. 

# TO TEST
1. Do a fresh install of the default site
   1. blt drupal:install --site=default 
2. On the home page of the default site, make sure the slider and slides are removed and that there are no error messages (should go from ‘Welcome to your SiteNow site’ block to the ‘Getting started’ block)
3. Instead of the slider, there should now be three cards (‘Explore Features’, ‘Create Content’, and ‘Change Settings’) in a three column section under 'Getting Started'


pyrello:
@dnkemper A couple things:
* If we have decided to remove the slider content, then this update should add some new element to replace it that has the same content, since that is part of the getting started section.
* If you are removing default content file records, you should remove the corresponding files, which are stored in the same folder.

pyrello (changes requested):
@dnkemper A couple things:
* If we have decided to remove the slider content, then this update should add some new element to replace it that has the same content, since that is part of the getting started section.
* If you are removing default content file records, you should remove the corresponding files, which are stored in the same folder.

dnkemper:
@pyrello THANK YOU! I knew I was missing something that was right in front of my nose haha. Do you have any recommendations with what block I should add the information to? We haven't used a collection block but not sure if it would be an accurate usage of it? My brain needs caffeine stat

joewhitsitt:
I think the easiest would be to go back to individual cards for this like we had in v2. Collection will run into the same nested entity issue as slider.

dnkemper:
Thank you both for helping me!

pyrello:
I'm going to test this and see if the most recent dev version of `default_content` fixes the issues we have been seeing.

pyrello:
Is this PR rendered moot by the merging of #2345?

dnkemper:
@pyrello Yeah... not even sad. I'll close it. Thank you!

pyrello:
@dnkemper Nice work on this! Sorry it wasn't necessary in the end.

PR #2338, opened 2020-10-14
Allow saving while source edit mode is open.

### LINKED ISSUE #2274: Hitting update while in source editor doesn't save changes done in HTML.
# Problem 

Changes to markup are lost if you hit "update" to a text area block in source view.

Screen Shot 2020-10-01 at 9 38 49 AM

# Proposed Solution

?

cory-skeers:
Resolves #2274 

# Problem
When adding or updating a block with filtered or minimal html text field, if the source edit is open, changes will not be saved.

## Steps to reproduce

From issue: https://github.com/uiowa/uiowa/issues/2274
Add block to section
Create custom block
Edit text in body field
Save block
Edit block
Switch body field to ckeditor source mode
Edit text
Save
Changes are NOT saved!
Once I changed back to non-source mode before saving the changes are saved properly.

## To test
Try the above with the edited JS fix from this branch. Check that updates are now saved.


docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
```diff
@@ -32,4 +32,28 @@
           });
         }
     };
+
```
pyrello, line 35: Is this specific to layout builder? If not, I think we should maybe not add it through this module. Possibly via uiowa_core?

pyrello (changes requested):

docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
cory-skeers: I believe it is specific to layout builder, but I will check!

cory-skeers: @pyrello Tested editing a node body field and the footer contact info block, and both maintain changes when saving with the source edit form open, where text fields edited through layout builder do not. Looks to be limited to ckeditor within layout builder.

pyrello: Sounds like this is appropriately scoped. Thanks!

GaryRidgway (approved):
I have tested  this functionality and can confirm that when using a CKEditor WYSIWYG i can now save in source view

pyrello (approved):
Changing this, since Alan reviewed.

Issue #2339, opened 2020-10-14
Update metatag patch to be compatible with future versions


pyrello:
# Problem
In working on #2160, I was updating all the dependencies and discovered that the dev version of `metatag` that we were upgraded to was no longer compatible with the patch we were applying to that module. I found a newer version of the patch, which applies: https://www.drupal.org/files/issues/2020-01-17/metatag-n2782797-25.patch. However, in between the patch we were using and the new one, there was a change to the name of a sub-module which was created by the patch, from `metatag_fine_grained_perms` to `metatag_extended_perms`. This is a problem because, after the patch is updated, I can't sync the default site because it expects a module with the first name to be there and it isn't. Even if we lock in on the specific dev version that worked with the old patch for now, that means we won't be able to update that module ever.

# Proposed solution
This has to be done in multiple PR's, I believe.
## First PR
* Uninstall `metatag_fine_grained_perms` from default site.
* Re-export config to update permissions.
* Copy changes to permissions to additional permissions files.

## Second PR
* Update `metatag` module.
* Update patch to https://www.drupal.org/files/issues/2020-01-17/metatag-n2782797-25.patch.
* Add `metatag_extended_perms` to `config/default/core.extension.yml`.
* Update `modules/custom/sitenow_seo/sitenow_seo.info.yml` to refer to `metatag_extended_perms` instead of `metatag_fine_grained_perms`.

Initially pointing as 3.

PR #2340, opened 2020-10-14
Remove metatag_fine_grained_permissions module

pyrello:
The first part of a 2-step process for #2339 

# To test
I think this could be passed by review.


joewhitsitt (approved):
i understand

PR #2341, opened 2020-10-15
Update to metatag patch and metatag_extended_perms module

### LINKED ISSUE #2339: Update metatag patch to be compatible with future versions

pyrello:
Resolves #2339. 

Also addresses a few unrelated permissions updates for splits that didn't get them.

# To test
* `composer install`
* Review file changes: https://github.com/uiowa/uiowa/pull/2341/files. For each permission in the role config files, the update should be one-to-one with the new permission having an instance of the word `tag` removed from it. Additionally, there are some other updates to those role files to correct some omissions from previous updates.
* Test that patch applies: `blt ds --site=default`
* Log in to the site and masquerade as a webmaster and make sure that you can still access the metatag settings.


config/coronavirus.uiowa.edu/user.role.editor.yml
```diff
@@ -73,8 +73,6 @@ permissions:
   - 'create uiowa_hero block content'
   - 'create uiowa_image_gallery block content'
   - 'create uiowa_slider block content'
-  - 'create uiowa_page_title_hero block content'
-  - 'create uiowa_spacer_separator block content'
```
pyrello, line 77: @bspeare Are you aware of any reason why we need to leave these permissions intact on coronavirus.uiowa.edu? I ask because I was planning on removing them.

joewhitsitt:
no longer blocked?

pyrello:
No longer blocked

joewhitsitt:
pulling down

joewhitsitt (approved):
Tested on brand

PR #2344, opened 2020-10-15
Export collegiate config after failed admissions sync.

richardbporter:
Working through #aggregator, I ran into an error syncing admissions. I think this config needs to be exported to match a recent change to default config. I think this is what errored on a recent deployment.

### Testing
- Sync various collegiate sites.
- Test any related functionality.

Does this PR need to include updates to other splits with this config?


richardbporter:
I did not have any other syncing errors with these sites:

```
- default
- admissions.uiowa.edu
- brand.uiowa.edu
- coronavirus.uiowa.edu
- hr.uiowa.edu
- ighn.international.uiowa.edu
- policy.clas.uiowa.edu
- sitenow.uiowa.edu
- uiowa.edu
```

joewhitsitt (approved):
I was running into these too. Had to previous branches out there to fix it but they were failing. This worked locally on uiowa.edu and looks like a successful build this time. Step forward for sure.

PR #2370, opened 2020-10-19
Remove LBS checkbox, again.

### LINKED ISSUE #2357: Override layout builder style checkbox is back!
### Problem
The banner block refactor reintroduced this checkbox which was removed in https://github.com/uiowa/uiowa/pull/2332. 

https://github.com/uiowa/uiowa/commit/1fd282c7e025f8184b14493024b6b1f3c7230b5b#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4c

### Solution
Remove it again! There is a todo comment about adding a fieldset that always shows as well.

richardbporter:
Resolves #2357

In the absence of a fieldset, which I don't think is possible, I added some text to delineate the styles section of the off-canvas form. I called it "Appearance" but open to suggestions or removal. I also added some classes in case we want to add admin styles to differentiate these somehow.

### Testing
- Check out this branch locally.
- Add various blocks to a page via LB.
- See that the styles render below a div with text "Appearance".
  - [ ] Accept, edit or remove text.
  - [ ] Add our own admin form styles or remove applied classes.
- Ensure adding and removing styles is intuitive, works correctly and that https://github.com/uiowa/uiowa/issues/2357 is no longer an issue.


dnkemper:
I can start testing this!

pyrello:
I'm also checking this out.

pyrello:
I think that it looks good to have the label above the section. I'm not sure about "Appearance" as the label. Related to this, I think we should add something similar when the options are shown for a section and the "Appearance" label doesn't seem quite right for that.

Is "Options" too generic?

pyrello:
Screen Shot 2020-10-19 at 2 37 47 PM

Noticed this when testing on a vertical video on the homepage of uiowa.edu. Looks like there needs to be a check for whether any LB styles exist?

pyrello:
I re-added in the check for `$styles_exist` to only add the heading when there are style options to select.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -73,22 +70,7 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
...
-          // Set a state for the layout builder style to toggle
-          // based on the value of the override value specified below.
-          $form[$lbs]['#states'] = [
-            'visible' => [
-              ':input[name="layout_builder_block_override"]' => [
-                'checked' => TRUE,
-              ],
-            ],
-          ];
+          $form[$lbs]['#attributes']['class'][] = 'layout-builder-style';
```
pyrello: @richardbporter Just curious why this was added. I don't see a corresponding style rule or `#states` that would seem to necessitate this.

pyrello: I'm now reading your comment above regarding this.

richardbporter: Yea I thought we may want to add admin styles to make these look different. It could certainly be removed. Maybe a @GaryRidgway or @bspeare have ideas?

richardbporter:
What about Display Options?

richardbporter:
I went with "Display options" for now and added this to the section configure form, accounting for weights. It looks a little odd when there are no other form elements when configuring a section as a webmaster. There might be a way to address that in the helper function I created with element children. Didn't get that far though.

I also added a simple border bottom much like the heading level field styles. I removed the class mentioned previously as we're not using it at this point.

Configure block
image

Configure section
image

Configure section (no other fields besides LBS)
image

pyrello:
I'm not sure if this is ready for review yet, but I think it looks good and I'll approve when its ready.

cory-skeers:
Pulling it down to check it out--

cory-skeers:
Looking good!

I'd done a little exploring yesterday putting the fields into a fieldset and taking them back out on submit, but never got it working quite right. I think the heading/bottom border combo works just as well to offset them as the fieldset, without the need for a lot of troubleshooting.

I'll wait in case you had anything else to fit in here @richardbporter, but if you're finished, I'd also give it a 👍

richardbporter:
Yea thanks all. If we're ok with how it looks when configuring a one-column section, then I think I'm done. If not, we could add more logic to `_layout_builder_custom_add_lbs_heading` to address that.

pyrello:
@richardbporter I'll re-check that.

pyrello:
I tweaked the style a little bit more to give it some separation from the text above and an oh-so-slightly smaller text size.

Screen Shot 2020-10-20 at 9 41 12 AM

pyrello (approved):
Solid improvement and some good refactoring!

PR #2371, opened 2020-10-19
Remove Layout Options from theme settings

### LINKED ISSUE #2250: Remove "Layout options" from UIDS Base theme settings
Follow-up to #2221 

# Problem
A long time ago, we envisioned the possibility of having multiple options for what the default container of the site was (e.g. an entire site could be full width by default). However, we have not kept up with the options that were originally provided in terms of our theme development. At this point, those options are essentially non-functional and can only serve to break a site.

# Proposed solution
* Remove these options from theme settings form
* Remove them from the theme's config and schema
* Update hook to update theme settings on each site since this is locally managed (config ignored).

cory-skeers:
Resolves #2250 

'Layout Options' in the theme settings has become deprecated, and can (should) be removed.

This PR removes it from the theme settings form, the profile config and default config settings, and an update hook to remove it from the individual site database.

# Testing
Do a fresh install or a ds(a).
Go to /admin/appearance/settings/uids_base and check that Layout Options do not appear.
`blt frontend` and ensure no template errors occur on loading a page.

## Note
Can use `\Drupal::configFactory()->get('uids_base.settings')` to check the current theme settings on a site, and that the 'layout' config setting is not present, or `SELECT data FROM config WHERE name='uids_base.settings';` to check the database directly.


pyrello (approved):
Works as advertised. A subtle but solid improvement.

PR #2373, opened 2020-10-19
Site grad.uiowa.edu site split updates

pyrello:
* Adds a site split for grad.uiowa.edu
* Enables the book module for that site only
* Sets permissions book content type

# To test
Reviewing the files is likely sufficient.


richardbporter (approved):
We've used the book module on quite a few D7 sites. Its worked well for certain use cases. If you do anything more with it (styles, etc.), I'd be interested in seeing if we can create a feature potentially.

pyrello:
@richardbporter Good to know. I'm not totally sure yet if this is the approach I am going to take for building out https://www.grad.uiowa.edu/academics/rules-and-deadlines/manual, but I wanted to be able to start playing around with it. There seem to be some limitations with how it interacts with Layout Builder that are unresolved at this point.

mark-bennett-uiowa:
The [https://policy.clas.uiowa.edu](https://policy.clas.uiowa.edu) site has been using the book module in D8 since creation.

PR #2382, opened 2020-10-20
Move grad split config to grad split dir.

richardbporter:
#2373 exported the grad site split config into the default config and activated it. We should be able to move it to the grad split directory. I'm having trouble exporting things correctly locally.

### Testing
`blt dsa`

### DIFF
--- a/config/default/field.field.block_content.uiowa_articles.field_uiowa_headline.yml
+++ b/config/default/field.field.block_content.uiowa_articles.field_uiowa_headline.yml
@@ -11,7 +11,7 @@ id: block_content.uiowa_articles.field_uiowa_headline
 field_name: field_uiowa_headline
 entity_type: block_content
 bundle: uiowa_articles
-label: 'Headline'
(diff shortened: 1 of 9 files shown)


richardbporter:
Actually, this might be it. Looks all the config was in the split except the split config itself.

richardbporter:
Running `blt dsa` locally to see...

richardbporter:
Nope. We need to disable the split first, then import config. Working on that.

joewhitsitt:
pulling down real quick to test update hook on random site

joewhitsitt (approved):
removed book from the site i tested

joewhitsitt:
does uiowa need to finish before this one goes out?

richardbporter:
Probably

richardbporter:
The irony here is that we'll probably end up enabling book for everyone.

richardbporter:
But the split config needs to move anyway.

PR #2383, opened 2020-10-20
fix non layout pages

joewhitsitt:


cory-skeers (approved):
~Reviewed, not tested (yet)~

Tested on http://labfolland.local.drupal.uiowa.edu/thomas-folland, and the fix worked 👍

richardbporter:
#2382 has a cache rebuild if this needs it.

PR #2399, opened 2020-10-23
Bumped up uids version to 3.0.6

dnkemper:
Bumped up uids version to 3.0.6


dnkemper:
Joe and I were curious as to why the version number didn't get updated in the yarn.lock file on line 936 when running the `yarn upgrade @uiowa/uids` command but it's pulling the correct hash?

pyrello:
Closing in favor of #2400

Issue #2313, opened 2020-10-08
People Manual Sort doesn't work in v3


joewhitsitt:
## Issue

The recent addition of responsive tables is causing a conflict with the draggableviews table that is implemented via the SiteNow People manual sort option. As a result, the drag/drop doesn't work and there are table markup being displayed that shouldn't be.

V3: Non-functioning drag handles, non-hidden table markup.
<img width="612" alt="Screen Shot 2020-10-08 at 11 11 19 AM" src="https://user-images.githubusercontent.com/4663676/95485248-14c21300-0957-11eb-9a0e-0f3a4d5d5a75.png">

## Proposed Solution

- Provide another "turn-off" option to avoid providing any table styles or functionality to this particular table
OR
- Hardcoded scope to not apply functionality on top of the SiteNow People manual sort table.

joewhitsitt:
Removing the responsive tables library solves the v3 issue. Since header content (wysiwyg) can be place above this table view, I think we need some more disable functionality beyond the table--static class

joewhitsitt:
Added a known issue: https://sitenow.uiowa.edu/known-issues#uiowa-issues-2313

joewhitsitt:
This issue has also come up with the PHP Live Chat (text version) that we tried to implement for advisingcenter. The JS for that pulls into the website some table markup (layout purposes, not data). We can't control this markup to apply any opt-out classes.

I think we need to revisit the scoping of the responsive table functionality.

richardbporter:
Is this done?

PR #2401, opened 2020-10-26
People sort and other responsive table fixes.

GaryRidgway:
adding wrapper classes so this doesn't get applied on EVERY table, the list of classes should now be manageable in this file

This provides functionality that lays the groundwork for https://github.com/uiowa/uiowa/issues/2313

This also provides a small fix for collections not having a full flex basis on LB pages.

This also fixes Tables in any container, including the previously bugged collections.

This should be merged after https://github.com/uiowa/uids/pull/390 is merged, as the js in the `tables.js` file there is needed for this.


cory-skeers:
I pulled this and updated the uids hash to use uiowa/uids#390 for testing, and everything appears to be working great! 

One issue I came across, not sure if it's related to this or was present before: When manually sorting people, the `Show Row Weights/Hide Row Weights` button toggles the drag handles like it should, but the weights themselves aren't appearing for editing.

cory-skeers (approved):
Fixes worked with changing the uids hash to uiowa/uids#390 and using both updates together.

I don't think the `Show Row Weights` not actually showing the row weights is a result of this--can probably make a separate issue to address it, unless you think it can be tackled here too, @GaryRidgway?

joewhitsitt:
@cory-skeers yeah that is a known-issue that was supposed to be fixed soon after v2 MVP... Created a follow-up #2416, but I told Alan he could hide the button as a temp solution

GaryRidgway:
I have hidden the button for row weights for now

PR #2402, opened 2020-10-26
Update media_entity_facebook, lazy, schema_metatag

### LINKED ISSUE #2361: Facebook Post oEmbed (v2/v3) is going to break 10/24
## Issue

New and existing facebook post embeds will respond with a bad request.

## Proposed Solution
Wait for https://www.drupal.org/project/media_entity_facebook/issues/3176739

OR

Try implementing this https://www.drupal.org/project/drupal/issues/3176978#comment-13860046 using our existing core patch or via #2356

### LINKED ISSUE #2397: Update lazy to 3.6
## Issue

We lost the fade-in effect in https://github.com/uiowa/uiowa/pull/2160, https://github.com/uiowa/uiowa/pull/2160#issuecomment-710357346

## Proposed Solution

Evaluate the latest 3.6 version if this is fixed
https://www.drupal.org/project/lazy/issues/3171723

In testing, make sure images are locally loading and not supplied by stage_file_proxy.

joewhitsitt:
Resolves #2361
Fixes #2397
Relates to #2282 

I have added the API id and secret to each environment's Acquia secrets file.  For example, you can ssh into uiowa.dev and then view the secrets.settings.php file to use the credentials locally in a site's local.settings.php file.

This patch is a short term fix: https://www.drupal.org/project/media_entity_facebook/issues/3176739#comment-13882401

Also snuck in a fix a lazy load regression by updating to the latest stable release, and an update to schema_metatag which removes a patch.

**Reference**:
https://docs.acquia.com/resource/secrets/

## To Test
- `composer install && blt ds --site=default` (no database update)
- Assuming you are in an environment that is setup to use the credentials, you should be able to add a new Facebook post via media without issue.
- A non-working setup will either error out or silently fail but fail in bringing over any metadata related to the post.
- You can also check to see that Instagram is also working as this is now served by the same variable.
- Since we aren't writing the credentials to the db, there should be no config export.
- Fade-in effect should be back for image loading. #2397
- On uiowa.local.drupal.uiowa.edu, make sure that SearchAction is still in the `application/ld+json` schema.org markup.


pyrello:
I'm not sure if I'm doing something wrong, but when I attempt to add a video (to the homepage of the default site via a text area) nothing is rendering. I added the settings from the secrets file to the `local.settings.php` file for the default site.

I have tried starting over a couple times too.

joewhitsitt:
video from what source? example link?

pyrello:
Here is the link I was using: https://fb.watch/1nTMZ5CfeE/

joewhitsitt:
So, I don't think the https://fb.watch... is a valid oEmbed endpoint for Facebook. I was able to add the video post using the full URL: https://www.facebook.com/universityofiowa/videos/947406292673134

Screen Shot 2020-10-27 at 10 56 14 AM

pyrello:
Thanks for clarifying and sorry for the misunderstanding. I have done this correctly before, but for some reason I was thinking it was like a YouTube video.

pyrello (approved):
Reviewed, not tested.

PR #2408, opened 2020-10-26
Enable Chosen for config-split admin forms

### LINKED ISSUE #2385: Config split admin page not rendering select lists with Chosen
### Problem
The config split admin UI is no longer rendering multiple select lists with Chosen. This makes it difficult to find config entities and mark them for a complete or conditional split.

### Proposed solution
Investigate why this happening now and if it is affecting other admin UI select lists. Refactor our sitenow chosen library to fix it. This may be a good time to investigate a chosen.js replacement sine it is deprecated.

cory-skeers:
# Problem
Config split edit forms were using checkboxes, which then would not utilize the handy Chosen.js, making using the UI to edit config splits more cumbersome.

Resolves #2385 

# Test
`git pull`
Log into a local site.
Visit an edit form off of /admin/config/development/configuration/config-split
Check that form fields (such as "Modules" under the "Complete Split") appear as select fields with the select list appearing on click.
Save the form and check that any changes were saved correctly.

# Followup
This is a bug fix to allow easier use of the admin form, but doesn't address the fact that Chosen has been deprecated. A followup issue should be created to account for the deprecation, possibly to switch to Select2, but that change is a bigger haul, and not necessary to fix this bug.

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.profile
+++ b/docroot/profiles/custom/sitenow/sitenow.profile
@@ -334,9 +334,16 @@ function sitenow_form_views_form_administerusersbyrole_people_page_1_alter(&$for
 function sitenow_form_config_split_edit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   // Enable themes to be blacklisted.
   $form['blacklist_fieldset']['theme']['#access'] = TRUE;
-  if (!\Drupal::state()->get('config_split_use_select')) {
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_prepare_form().
+ */
+function sitenow_config_split_prepare_form(EntityInterface $entity, $operation, FormStateInterface $form_state) {
+  // Set a state variable to ensure config_split uses our Chosen
+  // select implementation instead of checkboxes.
+  if ($operation == 'edit' && !\Drupal::state()->get('config_split_use_select')) {
     \Drupal::state()->set('config_split_use_select', TRUE);
-    header('Refresh:0');
   }
 }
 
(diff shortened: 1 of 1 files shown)


joewhitsitt:
So, this "feature" was introduced and has a Chosen fallback, but depends on the chosen module. Could we some how call `useSelectList()` and return TRUE instead of a refresh?

https://www.drupal.org/project/config_split/issues/2965738

cory-skeers:
@joewhitsitt `useSelectList()` has a state variable override option to return TRUE without the Chosen/Select2_All modules being present. I'd set it in a form_alter if it wasn't already set, which required a forced reload on the first visit to a config form.

Moved the state variable check/set out of the `hook_form_FORM_ID_alter` and into `hook_ENTITY_TYPE_prepare_form`, which should make the forced reload no longer necessary for the same approach.

Still not a fan of the fact that the `prepare_form` checks the state variable first, only to have it checked again in `useSelectList()`...

joewhitsitt:
Thanks for the explainer, I missed the issue comment.

This worked. Sorry for digging, but why are we checking at all? Don't we want to force select (chosen) on the split add form as well generally when in doubt?

joewhitsitt (approved):

cory-skeers:
Removed the `$operation=='edit'` constraint--that would have prevented this from working if an admin opened the add (or any other config_split-specific) form prior to an edit form, and wasn't adding anything. 

The reason I went with the state variable check and set is I don't know a better way to make sure the state variable is always set across all sites and environments, as opposed to if the module used a config setting. Basically just want to set the state variable to true if it isn't currently set, and then let config_split build its form using select instead of checkboxes.

PR #2420, opened 2020-10-28
Use filtered_html as default text format (if available for field)

### LINKED ISSUE #2377: Rescope minimal/filtered html defaults
# Problem
Default text format needs to be set site-wide, and exceptions handled through hooks. Originally, this was set for text areas to default to minimal, with filtered the exception. The addition of the scholarships content type has raised this as potentially opposite what it should be, with filtered the default and minimal the exception.

# Proposed solution
Config update to set ckeditor default ordering with filtered above minimal to set filtered as the site-wide default.
Remove current exceptions which change uiowa_text_area and node body fields to filtered default.
Add new exceptions to capture block text fields where minimal default is desired and set them to minimal on creation.

cory-skeers:
Resolves #2377 

# Description

ckeditor text formats need to be set on a site-wide basis, and exceptions handled through custom code. Previously, minimal was set as the default, and exceptions added to certain fields (uiowa_text_area block field and node body fields). As new content gets added that may benefit from default filtered_html, want to get away from minimal as default, filtered_html as exception approach.

Further, there are not currently instances like the hero block which it's important to default to minimal while allowing filtered_html. This PR reverts to setting the default text format ordering as filtered_html > minimal > full_html > plain based on the field and permissions.

# Testing

- Sync a site.
- Edit content fields and check that the text formats are as expected
  - Add a new basic page and edit--is the body field defaulting to filtered_html?
  - Add a new uiowa_text_area block--is the main text field defaulting to filtered_html?
  - Add a banner (or other which should not allow filtered_html) block--is the excerpt or other text field defaulting to minimal, and not allowing to be switched to filtered_html?


joewhitsitt:
Wait, I thought we wanted minimal as default for block types but can't restrict it to minimal since the text area block is used for all kinds of things and filtered_html for content type fields. Maybe I didn't fully understand the original issues.

https://github.com/uiowa/uiowa/issues/2038
https://github.com/uiowa/uiowa/issues/1831

The other approach would be to update the allowed formats field settings for all content type fields to not have the minimal option. Testing would need to be done to see what happens to existing content with the minimal filter selected. Does it go back to filtered_html if minimal is not available or are we looking at an update hook. I don't think there are any allowed elements in minimal that aren't allowed in filtered_html.

cory-skeers:
So, we can restrict fields to minimal on a per-field, per-block basis, but if we allow both filtered_html and minimal on a field, we can't set which is the default on a per-field, per-block basis. When starting this issue, I'd thought we'd need to set filtered_html as the default, and then add new exceptions to set minimal for those blocks which allow for filtered_html and minimal.

Currently the only inline blocks that allow both are:
- Text area
- Collections
- Hero

The rest are restricted to plain and/or minimal, so this change won't affect them. The hero block was the block needing access to filtered_html but wanting to encourage minimal, but it's now deprecated/restricted, and doesn't need the exception.

I could see adding an exception for Collections to default to minimal, while allowing for filtered_html (there's already the use case of adding tables within accordions).

I think originally there were more blocks which allowed for but discouraged filtered_html, like the hero, or at least the thought that there would be, but over time it's become less of a concern.

mark-bennett-uiowa:
@cory-skeers Yes, yes, a thousand times, yes! Thanks Cory. I will try to get this tested later tonight or early tomorrow. Appreciate you taking care of this, especially as I have run into this with copying content between a dozen sites, lol.

joewhitsitt (approved):
I am fine with this. Just didn't seem that long ago that the opposite was desired. Maybe @bspeare should chime in, because I think this issue originated from him.

Long term, I think we will need something different to replace text areas like the footer contact info block where we want to restrict.

cory-skeers:
@joewhitsitt Thanks for taking a look at this. I'll bring it up in parking lot today.

My main thinking on it is that we're likely to see more cases where filtered is the desired default than minimal (at least, where we can't restrict the field to minimal only), and this way we only need to add exceptions for minimal, rather than every new filtered_html field.

I'd be shocked if we don't want/need to add the exception in to set the default to minimal for certain fields in the future, but I think we happened to do this at a time when we don't need one.

For the footer, we can either restrict to minimal only, or set the default to minimal for new sites, since it's a specific block we're putting content into.

Issue #2320, opened 2020-10-09
Person layout/design changes


joewhitsitt:
# Story

As a brand team design member or person content type stakeholder, I have seen and given approval for all the changes being made to the default person content type layout/design changes.

# Proposed Solution

As we develop how the _default_ person profile look going forward for SiteNow v3 and accounted for all the new fields we plan to add to the content type. We should seek guidance/approval from the Brand team and other stakeholders.

Relates to: 
https://github.com/uiowa/uiowa/issues/2205
https://github.com/uiowa/uiowa/issues/2293
https://github.com/uiowa/uiowa/issues/2297
https://github.com/uiowa/uiowa/issues/2291
https://github.com/uiowa/uiowa/issues/2234
https://github.com/uiowa/uiowa/issues/2498

joewhitsitt:
Thoughts from the design meeting. Let me know if I forgot some stuff, I am sure I did.

Good examples: [Tippie](https://tippie.uiowa.edu/people), [Harvard Library](https://library.harvard.edu/staff)

- Image placement, see Harvard, if no image, doesn't disrupt the flow of the rest of the content.
- Title below image.
- Consider if no sidebar content, the region collapses making the main full-width.
- Consider splitting sidebar into mobile top (positions/titles) and mobile bottom content (external websites), flowing above and below the main content. The idea comes from CSS Grid, but could be accomplished by a layout that has a sidebar primary and sidebar secondary together in desktop but separate in mobile.

pyrello:
It sounds to me like we may want to create a custom layout to handle this, which would allow us to accomplish the 3rd and 4th points above.

> * Title below image.

In this case, by title, are we talking about the person's name?

joewhitsitt:
Admissions now has a layout based person full node which can serve as inspiration or "merge to default"

joewhitsitt:
via Slack:
>https://github.com/orgs/uiowa/projects/1#card-46994730 has a list of related issues which are all closed now and so maybe it can be closed and moved to Done?

#2293 hasn’t been done for the default yet, just Admissions. I never really filled in this issue, but I think I originally thought this would be the design review/finalization task about look and placement of all our person fields for the default display. If that has already happened, I think this can be closed.

PR #2430, opened 2020-10-29
Person Contact Information

### LINKED ISSUE #2297: Person contact information
We need to be able to store and display contact information for a person. This includes the ability to list multiple office addresses.

Screen Shot 2020-10-05 at 4 40 46 PM

joewhitsitt:
Resolves #2297 

## Does

- Provides paragraphs "field collection" allowing multiple contact entries, split by optional label.
- Address field with all fields optional
  - Address has first name, last name, company as well as address line 2 to handle building/room/etc.
- Phone field (optional)
- Available on hr.uiowa.edu with a bare minimum implementation.
- Gracefully imports to person_extended websites. Since the form/displays config is split, these fields exist but are not available for use.
- Despite unusual markup structure, there are no a11y flags from SiteImprove related to this new functionality.

## Does Not

- BUILDING: VARCHAR2(4000) NULL
- ROOM: VARCHAR2(4000) NULL
- STAFF_CONTACT: VARCHAR2(128) NULL
- Separate fields for fax, cell, etc.

## To Test

- pull, `composer install && blt frontend && blt dsa`
- As a webmaster you should be able to add separate entries of contact information on v2, v3, HR and collegiate sites without a separate split for person content type form/display (e.g. person_extended)

## Notes

- Frontend blank space is registered if node is saved with a blank contact item as it technically exists as a field item of the parent paragraphs field. Logic could be added to account for this, but so far the space isn't really affecting anything.

default local drupal uiowa edu_people_herky-hawk (2)


dnkemper:
Going to start testing this. Definitely won't finish today but will try to get as far as possible

mark-bennett-uiowa:
Also pulling down for testing.

mark-bennett-uiowa:
Works pretty good. Since you can attach a first name and last name to the address, you could "simulate" the staff contact with that.
Screen Shot 2020-10-30 at 5 06 50 PM

**Thoughts**:
1. The styling for the link fields needs to be updated to match default font. Default font size is 1.2rem, the font-size of links is 16px.
2. Maybe shrink the line-height for the display of the address?? Its currently at 1.7. Changing this to 1.4 or 1.5 tightens up those lines a little more if desired.
3. We could set the default country to US if you wanted, so you wouldnt have to search through that list every time. It is set in the core.entity_form_display.node.<content_type_name>.default.yml file as below:
```
  person_address:
    weight: 12
    settings:
      default_country: US
    third_party_settings: {  }
    type: address_default
    region: content
```
See [https://github.com/uiowa/uiowa_person/blob/main/config/install/core.entity_form_display.node.person.default.yml](https://github.com/uiowa/uiowa_person/blob/main/config/install/core.entity_form_display.node.person.default.yml) for an example of how I did it on the uiowa_person module.

mark-bennett-uiowa:
@joewhitsitt Approved pending responses on thoughts from above.

mark-bennett-uiowa:
**UPDATE**: Decided to check the people list on /people. Image is still WAY too big:
Screen Shot 2020-10-30 at 5 20 35 PM

pyrello:
I updated the contact address field to set the US as the default country. I went about it slightly differently than @mark-bennett-uiowa's suggestion above, as this is how I could figure out how to do it in a way that I could export as config. I also added a form submit handler that checks the address to see if it has any other fields set than the country code. If there is no other fields on the address set, then it sets the address to nothing (which is how it comes through when submitted with the country set to '-None-'). I tried adding an empty address to test it. May need some other test cases.

mark-bennett-uiowa:
Checked code from @pyrello pushes. Looks good. Looks like the merge branch failed though.

joewhitsitt:
> **UPDATE**: Decided to check the people list on /people. Image is still WAY too big.

Out of scope, but I am sure we will address eventually in https://github.com/uiowa/uiowa/issues/2320

I plan to test @pyrello changes and I will see if I can font-size and line height tweaks @mark-bennett-uiowa suggested.

joewhitsitt:
The address has `<br>` between lines, so I am going to leave the spacing alone for now because the text technically already matches the paragraph default of 1.2rem size, 1.7 line height.

Extended the paragraph default to field `<a>`.

@pyrello the logic fails on edit of existing. I can't get anything address edits (after initially blank) to render/save.

pyrello:
@joewhitsitt I'll take a look.

cory-skeers:
@joewhitsitt I made some updates to the logic for addresses, if you want to test it again and see if it correctly render/save.

@pyrello Heads up that I made a few changes to the address unsetting logic, if you get a chance to look over it (basically just how and where deltas are handled).

joewhitsitt:
That worked @cory-skeers, thanks!

pyrello:
My local environment seems to be messed up at the moment, so I can't test this, but I reviewed the code changes and they look good to me.

pyrello (approved):
Tested prior to updates that fix functionality for existing addresses.

PR #2465, opened 2020-11-02
Create base area of study content type.

### LINKED ISSUE #2380: Investigate shared use case for Areas of Study
# Motivation
As part of #2365, there is going to be a content type created for the Admissions website for an Area of Study. This task would be to investigate whether we can use some of these fields for a shared use content type of the same name.

### LINKED ISSUE #2438: Build the base Area of Study content type
# Motivation

As part of #2365, we need individual pages to show information related to Areas of Study.

# Story

As a site visitor I want to be able to search for information about Areas of Study.

# Proposed solution

- [x] Build a base Area of Study content type that consists of the shared fields that are needed for both the Admissions and Grad Admissions sites.
- [x] Create an Area of Study feature split and export all related config for the content type this split.

## Notes about current content type's fields - this list is not comprehensive
Content type – Used for determining where the page should live in the URL alias structure (e.g. under `/academics`). Might be able to be replaced with a path alias pattern.
Mail item code – connects pdf output to mail item for that area of study – manual connection with MAU
Alternate titles – allows alternate names to find when looking by alpha
Alternate description – shown under name in that list
Academic Group – multiple – discussed above
Degree details – show/hide fields based on program type
Tracks - discussed above
Related links - ability to link to department, college and general catolog
Videos – youtube – open in modal now
Student profiles – shows thumbnail and quote on area of study
Intro – These need WYSIWYG
Why Study – These need WYSIWYG
Course work – These need WYSIWYG
Requirements
first year/transfer requirements
Transfer tips – pulls into other view – /transfer-plans
Application process – fy/transfer
Visit Information – can dump this
Program image – image for print but could be used on areas of study pages for visual
Search terms – probably not being used but wanted to provide that option

### LINKED ISSUE #2443: As a prospective student, I can start an Areas of Study search on the Admissions homepage
# Motivation

As part of #2365, Admissions would like a search block that can be placed on their homepage that will allow a prospective student to start a search for an Area of Study. The search block doesn't need to have any functionality other than being a form that will initiate a search on the Areas of Study page. Given the example given (see below), it seems like this feature is pretty straightforward.

# Proposed solution

Build a views exposed filter block for the Areas of Study page, which redirects results to the Areas of Study page itself. This block could, potentially, be used alongside any generalized A of S content type as potentially defined in #2380. 

## Examples of similar functionality
* https://admissions.wvu.edu/ - provided by Admissions as an example of this that they like.
* https://policy.clas.uiowa.edu/ - An example of a similarly built search box already used in our platform.

richardbporter:
# Update
 
* Resolves #2438 
* Resolves #2443 

# How to test

## Setup
- Go to Administration -> Configuration Split and enable the Areas of Study split.
- Import configuration for the site on which it was enabled on.
  - `drush @mysite.local cim`
- Add some Area of Study content and review fields.

## Test
- Add "Area of Study search" block to a page via Layout Builder and save.
- Test the search box and observe that it takes you to /areas-of-study

=======
Todo:
- [x] ~~Setup moderation~~ - Will move to subsequent issue.
- [x] ~~Output colleges as links to those colleges~~ - Colleges need more consideration and are removed from this version.
- [x] Refactor degree/type fields https://github.com/uiowa/uiowa/pull/2465#issuecomment-724871881
- [x] Remove image field
- [x] ~~Add help text to program type label fields to let the user know they are optional and only necessary when they want to alter the label.~~ - Leaving labels as required for now.


richardbporter:
College list needs to output a link to the college.

richardbporter:
~~Needs responsive tables!~~
Table seems fine without it.

pyrello:
Replace "degree type" and "degrees" fields with a paragraph reference field that is a collection of those two fields. The degrees list option should be optional with in the paragraph type.

pyrello:
I'm going to refactor the degree related fields as discussed.

richardbporter:
> Output colleges as links to those colleges

Do we want to hard-code those links or change this field as well? Almost seems like a multi-value link field would be more appropriate. I'm not sure how that affects sorting/filtering though.

pyrello:
I kind of wonder if there should be a mini entity for storing the college information? Its reused through different Areas of Study, I think.

pyrello:
I pushed my field changes. See what you think.

GaryRidgway:
I love the changes, to the backend, i think it's very clear. I think we could preprocess these on the front end in to a nice display, but i think that is of course out of scope of just refactoring the fields.

cory-skeers:
I think they [colleges as links] should be coded in somehow, whether it's hardcoded in something like a preprocess, saved in a site config, mini-entity, etc. I don't think there'll be nearly as much concern over links needing to be updated as if we needed to link to departments rather than colleges, but the chance anything would change and not get updated everywhere worries me.

@pyrello Looking at the new degree reference field now, and I think it's a good improvement. Only thing I'm noticing is Minor and Certificate being listed in abbreviation--do we need to include them or can we let the user choose -None- for that (or even hide it for those two degree types)?

And for this being the base area of study, are there strong opinions on building out the full list of degree options (PharmD, DNP, etc) or keep it small until the fields/forms are finalized, and add on later?

GaryRidgway:
I think I agree with Cory, would hiding minor and Certificate in abbreviations make sense? I feel like those are just types, or maybe I am misunderstanding?

mark-bennett-uiowa:
Minor and Certificate are not abbreviated in the General Catalog. You would see them as "Minor in xxxxxx" or "Certificate in xxxxxx". Even more than that, you can have "Graduate Certificate in xxxxx" and "Postbaccalaureate Certificate in xxxxxx" something too (see Classics).

pyrello:
Great feedback! I just copied what was already in the existing fields. I'm actually not that knowledgable about degrees (I have only ever received an AA), so I would defer to others wisdom in this area.

cory-skeers:
I removed the minor/certificate from abbreviation--I think hiding the abbrv field for those types would be a good next step.

I also added some abbreviations and both the additional certificate types. For the abbreviations, I just went off of what was available off the bat to me in MyUI for a "What if?" degree audit; might not be the best source of truth, but it included everything that was already in the field as well as a few extras.

cory-skeers:
Set it up to hide the abbreviation field unless Major is selected as type. Might need some adjustments based on what all gets included in this base type, or how the page display gets built.

Also doesn't unset any value; only hides the field. At this time I don't think we'll need to unset or change the value, as there'll likely be logic based on the type that takes care of this.

cory-skeers:
'Removed' the image field from the content edit form, moving it into the advanced node options as a featured image like with the other content types.

cory-skeers:
One option for hardcoding college links in place, in a field preprocess. (As a test, only College of Law is linked)

pyrello:
@cory-skeers I don't love this just because I am not a big fan of hard-coding this type of data, probably because I have to deal with so many problems from it in my current codebase. But... this may be a good way forward for now.

Alternatively, we could drop the College field from this for now until we can figure out how to implement it.

pyrello:
I'm going to work on getting a basic LB template in place.

pyrello:
I'm having some second thoughts about the way that the degree field collection is structured that come from trying to figure out how to process the inputs into the rendered style that exists on the current Admissions site. 

When the user enters the degrees, if they enter multiple degree options, for example (using Art as the AoS):
* Major - BA
* Major - BFA
* Minor

They have the option to set them in any order, so they could have set it up as Major, Minor, Major. However, since they are supposed to be grouped by degree type, this doesn't make any sense. I'm wondering about a structure that is more like:

* Majors
  0. (Paragraph)
    - Abbreviation
    - Label (optional)
  1...
* Minors
  0. Label (optional)
  1...
* Certificates
  0. Label (optional)
  1...
* Preprofessional Programs
  0. Label (optional)
  1...
* Online/Off-campus Programs
  0. Label (optional)
  1...

Which is ultimately very much like their current structure. This might still allow us to not include the Program Types checkboxes from the current site.

Also... I stumbled across this page: https://admissions.uiowa.edu/colleges/college-education, which seems to indicate that there is a college content type and a relationship with Areas of Study. So we may want to hold off on hard-coding college links until we can get clarification on that.

cory-skeers:
Interesting...I'd had mixed feelings about the hardcoding, but also wasn't keen on building anything beyond basic options if they weren't used elsewhere. I feel like linking to colleges may be better as an addition then, likely as some form of entity, and taken out of this base case.

I still like the idea of lumping all degrees together with type and label, with major a subset that also includes abbreviation, but I see your point. And the more I try to think of reasons to not split them up, or of alternatives to it, the less attached I am to the current setup over splitting it up. I think there could technically be some query optimization differences between the two and how they're stored in the database, but we're looking at a 2 - 3 join difference at fewer than 1k rows, and debatable between whether optimizing by type or not is even preferred, so...

Basically a long-winded way to say I'm good with splitting them apart like you've listed out here.

pyrello:
I'm still not sure about this data structure. For Minors, Certificates, Preprofessional Programs, and Online/Off-campus Programs, I think we will need either a checkbox to indicate that there is at least one program, or to require the label be entered for at least one item. I also thought about adding some pseudo-field checkboxes that would toggle the showing and hiding of the related label entry and cause the label field to be pre-filled with the name of the AoS, but at that point, I'm not sure we have gained much over the way this was previously put together. It seems like there should be some straightforward pattern for how to put this together, but I'm not finding it.

pyrello:
Just pushed an update to change the interface for entering program type information. This does not yet prefill the AoS name in the non-major program type labels for the first item.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -61,6 +61,40 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
       }
       break;
 
+    // @todo Move this to a more appropriate location.
+    case 'node_area_of_study_form':
```
cory-skeers: I think this will only trigger when creating new area of study content, but not when editing existing forms. But then for editing existing, if adding checkboxes, think will need to check for whether each program type has a value or not.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -61,6 +61,40 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
...
+        if (isset($form[$program_type])) {
+          $program_toggle = $program_type . '_toggle';
+          $form[$program_toggle] = [
+            '#type' => 'checkbox',
+            '#title' => $form[$program_type]['widget']['#title'],
+            '#weight' => $form[$program_type]['#weight'] - 1,
+            '#name' => $program_toggle,
+          ];
+
+          $form['#group_children'][$program_toggle] = 'group_area_of_study_program_type';
```
cory-skeers: I'm getting a WSOD error when trying to add new areas of study content from this line. Commenting it out, things seem to work as expected, but not sure if leaving it out has other effects I'm not noticing.

`The website encountered an unexpected error. Please try again later.
Error: Unsupported operand types in Drupal\field_group\FieldGroupFormatterPluginManager->prepareConfiguration() (line 114 of modules/contrib/field_group/src/FieldGroupFormatterPluginManager.php).`

pyrello: I think I may need to re-export config.

pyrello: I think this may be resolved by the last commit.

docroot/modules/custom/uiowa_core/uiowa_core.module
pyrello: Good point. I only started with the node add form, but this will need to be adapted for the node edit form as well.

pyrello:
I was trying to figure out how to pre-fill the first value in the non-major program types with the idea that I would have a custom submit handler that would examine the form and remove them if the corresponding checkboxes were unchecked, but then I realized that the value I would be trying to pre-fill them from (the AoS title) doesn't exist at the time the form is being created. So... to get that version of the solution done will require implementing javascript that does that when the title field is filled in.

I'm wondering if there is a way forward with the current version using a custom submit handler and automatically filling those field values if the corresponding checkbox is checked and there is not already a label present.

docroot/modules/custom/uiowa_core/uiowa_core.module
cory-skeers: That did it! 👍

pyrello:
I removed all the code for pre- and post-processing the form fields. I realized that I was going to a lot of trouble to have the user not have to sometimes fill out a field, but that the field was still going to be stored with that value regardless. There may be some value in revisiting this in the future but I don't think this needs to be in the first version of the base case.

Still need to finish updating the LB display and the view.

docroot/modules/custom/uiowa_core/uiowa_core.module
pyrello: I removed this, so it is moot now.

pyrello:
At this point, I have stripped the AOS content type down to just description, featured image, and the degree, certificate, or other similar fields. I could be convinced to even strip out those last fields for now. Or we could leave them and continue to refine them in a future update.

The exposed search block functionality works and redirects to the Area of Study page as expected. I went with a custom block + "noodle" code because there is no way in views (currently) to both a.) have exposed filters display in a block (which allows you to display them on a separate page) and b.) also have the filters display on the view itself. This functionality may also need to be refined further in a subsequent update as we will actually need different exposed filters on the view page itself. This might be accomplished by suppressing those in the code that displays the form in the custom block. Or a block with exposed view could be created and then manipulated through code to redirect to the page view.

Either way, I think this is at a good point for moving ahead. Open to feedback.

pyrello:
Oh, also, I left in the body field which gets added to the block type automatically when its created. It seems like there could be a need for adding text below the search field, although that could just be handled via a separate text area. In writing that, I talked myself out of it and I'm going to remove it.

cory-skeers:
I think this looks good to go. Gonna leave it for another approval for now, since I contributed to it.

richardbporter:
I don't think I can approve this since I created it.

pyrello (approved):
Tested/reviewed by @richardbporter, @cory-skeers, and myself.

PR #2488, opened 2020-11-05
Update to latest 1.x Composer version in CI.

richardbporter:
Composer 2 is out but there is other work needed to upgrade to it. This is a hotfix for failing CI builds to stick with Composer 1 short term.


pyrello (approved):
Reviewed, not tested

PR #2494, opened 2020-11-06
Move the headline field to outside of the slider container.

pyrello:
The slider component in UIDS has no explicit headline field defined, so the headline field attached to it in this project should be displayed outside the `<div class="slider">` container element. This should hopefully prevent unexpected behavior since we don't test for this type of condition in UIDS.

# How to test
* `drush @grad.local uli /`
* Inspect the HTML of the slider element on the page that is opened up and observe that the headline element is rendered outside the slider.


dnkemper (approved):
Nice!

Issue #2496, opened 2020-11-06
SSL for *.law.uiowa.edu on uiowa


dnkemper:

joewhitsitt:
law.uiowa.edu is on uiowa. this wildcard should probably be on the same application

dnkemper:
@joewhitsitt dang. The blt command told me there wasn't any ssl coverage on any app

joewhitsitt:
there technically isn't, it isn't smart enough to catch relationships like this.

dnkemper:
@joewhitsitt hahaha relatable 🥇

joewhitsitt:
This cert still needs to be created, but instead on uiowa.

joewhitsitt:
this is done.

PR #2495, opened 2020-11-06
Initialize lhpdc.law.uiowa.edu multisite on uiowa

dnkemper:

PR #2497, opened 2020-11-06
Add related SSL domains to UMC output.

richardbporter:
Relates to #2496, #2495 

The `blt umc` command does not indicate any related domains that should dictate where a new site should go. This is problematic because the assumption is to put the site on the app with the fewest DBs. This PR adds really quick and basic related domain searching to UMC.

### Testing
Run a few phony sites through the command and then abort.

- blt umc lhpdc.law.uiowa.edu
- blt umc foo.tracers.physics.uiowa.edu
- blt umc foo.pharmacy.uiowa.edu


dnkemper:
Awesome! I have a meeting that's about to start, but will test this ASAP!

dnkemper (approved):
YOU ROCK!

Issue #2489, opened 2020-11-06
Upgrade to Composer 2


richardbporter:
We need to upgrade to Composer 2 because Composer 1 is EOL. 

- [x] Revert #2488 
- [x] Remove hirak/prestissimo from CI, DrupalVM 
  - [x] #2454 
- [x] Upgrade acquia/blt to 11.6.0
  - https://github.com/acquia/blt/releases/tag/11.6.0
- [x] Upgrade the following packages:
  - dealerdirect/phpcodesniffer-composer-installer
  - grasmash/drupal-security-warning
  - oomphinc/composer-installers-extender
  - zaporylie/composer-drupal-optimizations
- [x] Upgrade DrupalVM configuration to use Composer 2
- [x] Advise team to update local Composer installs to match

richardbporter:
BLT 11.6.0 removes drupal-check, which I thought was valuable for us to run in CI. :( We'll have to dig as to why and how to add it back if desired.

richardbporter:
Looks like it is available as a separate plugin in BLT 12: https://github.com/acquia/blt-drupal-check I'd say we wait until then to add it back. In the meantime, we can leverage #2164 if necessary.

richardbporter:
I can't get the VM to use Composer 2 by default. I had to run `sudo composer self-update --2` on the box.

PR #2503, opened 2020-11-09
Feature scholarship weights

### LINKED ISSUE #2436: Add weight module
# Story

## Display
Alternate sorting options in view displays
As a site visitor, 
I want to view a list of scholarships which has been pre-sorted
so that I can quickly and easily find the scholarships that are most important to me.
## Backend UI
Weight module to set priority content
As a site owner
I need to be able to organize my scholarship content individually
so that I can prioritize the end user’s experience and emphasize (or deemphasize) certain scholarships, which are not sortable in this manner based on data within the system.

# Proposed Solution

Add the weight module to all content types and update view displays (the exception being people draggableviews) to consider weight as part of the sort criteria.

If a service-wide solution is not in-scope, this has been requested specifically for Admissions' Scholarship content type/view block.

https://www.drupal.org/project/weight

<details><summary>Original Issue</summary>
<p>

Admissions currently has their scholarships weighted so that certain ones always display at the bottom of view displays and others are always at the top. Though sticky could be setup for the scholarship content type, it only solves part of the desired behavior.

> We need to be able to weight the scholarships for the order they show on the individual pages/views. There are certain scholarships that always need to be at the bottom of the list (less important) and certain ones we want listed first. 

This would also improve person sorting globally across the site..

</p>
</details>

cory-skeers:
Resolves #2436; followup issues may need to be added for:
 - As an Admissions site owner, I can order scholarships content in relation to each other (e.g. see weights for other content, drag and drop, or other option) so that prioritizing scholarships is faster and has less room for human error.
 - As a site owner, I can weight content which appears in list views in ways other than automatic sorting, so that I can prioritize displays for criteria which is not based on data within the system.

# How to test

`git pull`
`composer install`
Sync admissions, run the update hook if working off a previously synced version
Add a Scholarships block to a page, note the default ordering (alphabetical by title)
Update weights for various scholarships (both positive and negative need testing)
Check Scholarships block, see that the ordering is correctly updating

Check that sites not on the admissions split are unaffected.


richardbporter:
@cory-skeers Did you run Composer commands on your host or the VM? What version do you have on each? `composer --version`

cory-skeers:
@richardbporter Typically on the host, which is now on version 2.0.4.

v1.10.5 in the VM. Gonna try again with that version.

richardbporter:
Ok yea - I think Composer commands should be run in the VM for discrepancies such as this or PHP versions, etc. We may want to get to #2489 sooner rather than later to avoid confusion. It could hit people's VMs if they destroy/create or reprovision. I'm not really sure.

richardbporter:
Of course I am unable to require drupal/weight due to memory issues which is something Composer 2 solves. :)

joewhitsitt:
`php -d memory_limit=-1 /usr/local/bin/composer...`

richardbporter:
I've got weight added locally without updating anything else. Let me know if you want me to push that up.

cory-skeers:
@richardbporter That would be great! I made some changes to the update hook as well to create the storage/pull in the config prior to attempting to update the existing field weights.

If you want to push that up, this should be good for testing.

richardbporter:
I'm seeing a broken/missing handler on the view. Probably an config import order-of-operations thing. 

image

richardbporter:
Or I'm doing it wrong. :)

cory-skeers:
Hrm. Looks like it's something in how the update hook is pulling things in that's getting things mixed up. Taking a look.

GaryRidgway:
I actually just tested this  as well, and I am getting the broken handler message as well

richardbporter:
A manual sync and then cim gets the field and view working together. I'm not sure what its choking on in the update hook yet.

richardbporter:
This will be a non-issue soon. Just need to upgrade stuff. https://www.drupal.org/project/drupal/issues/2901418

richardbporter:
Update hook aside, this works as expected. I'm not having any luck with it. I suppose this could go out first then with a follow-up update to set initial weights to 0. Thoughts?

cory-skeers:
@richardbporter I'm leaning that way too. The update hook that doesn't deal with config could be included later, with a quick check to set the weight to 0 only if it's still null, if it gets edited between this and the update hook going out.

richardbporter:
Yea that sounds like a plan.

richardbporter (approved):

PR #2508, opened 2020-11-09
fixing table width measuring

GaryRidgway:
Quick fix to address tables measuring their space incorrectly.

# ISSUE
When measuring their space, if tables were given a pixel measurement with a nonzero decimal point (ex. 966.5) then it would round up (ex. 966.5 --> 967). if this was too big for its layout, this would break flow unnaturally. Tables are now measuring to decimal point pixel values.

# How to test

pull down `https://ap-purchasing.fo.uiowa.edu` and test the `/purchasing/contracts/uniform-rental-program` page. if it is not breaking like on the live version of the site at `https://ap-purchasing.fo.uiowa.edu/purchasing/contracts/uniform-rental-program` then the issue is fixed.


joewhitsitt:
pulling to test quik

joewhitsitt (approved):
thanks for the quick fix!

PR #2509, opened 2020-11-09
Feature scholarship weights update

cory-skeers:
Followup to #2503

# Problem

field_weight is a new field, and when used for sorting views (notably in the Scholarships block), content with no value in this field appear before all content with a value, regardless of that value.

Update hook sets field_weight values to 0 if they have not been set.

# Test

- Sync admissions site, run the update hook
- Create a test page and add a Scholarships block
- Edit at least two of the scholarships included in the block
  - For one, set a positive weight value
  - For another, set a negative weight value
  - Try to leave at least one scholarship in the block untouched
- Check the test page, and see that:
  - Negative weighted scholarship is at the top
  - Positive weighted scholarship is at the bottom
  - Untouched scholarship(s) is in between those


pyrello (approved):
Reviewed, not tested.

richardbporter (approved):
Tested with admissions on 3.1.1-build. 👍

PR #2507, opened 2020-11-09
Feature: Adding fields for person type counselor

bspeare:
Resolves #2484 
Resolves https://github.com/uiowa/uiowa/issues/2486

Relates to:
https://github.com/uiowa/uiowa/issues/2320
https://github.com/uiowa/uiowa/issues/2293

# How to test

## Counselor Type

- On the admissions site, the counselor option should be available for selection under person types.
- Selecting Counselor should reveal the hometown field and the territory select list which includes US/Territories and an International option.
- Hometown and Territory are fields that should only exist on the admissions site and the full node should have the two fields represented.

## Counselor Page

- On the admissions site, the counselor page should be styled with a new layout through layout builder.  Check how the page looks with all fields filled out and some fields removed.  
- Check other `v2` (https://hawkeyemarchingband.uiowa.edu/) and `v3` profile pages and make sure they look the same. 

## Counselor Listing

- As webmaster you should see the Counselors view block and be able to place wherever. This should show an exposed territory filter and list only people marked as a counselor. The person teaser should include their territories.
- A url pattern /counselors/% with state abbr. or intl (international) should display just those counselors that have that territory selected.

PR #2521, opened 2020-11-12
Admissions Counselors (map, listing, person)

### LINKED ISSUE #1200: Add Colorbox to frontend build process

### LINKED ISSUE #2485: Counselor Map
# Epic
https://github.com/uiowa/uiowa/issues/2212

# Story

As site visitor, I can view counselors with a map page so that I can easily find the right counselor to contact based on my location.

# Proposed solution
- Assumes mapping solution investigation is complete. #2499
- [x] Build Map display based on the Counselor Person Territory field. https://admissions.uiowa.edu/counselors (blocked by https://github.com/uiowa/uiowa/issues/2484)

# Notes
- Person could end up being on the map multiple times.

joewhitsitt:
Resolves #2485
Resolves #1200 

admissions local drupal uiowa edu_counselors

This is a custom map block for Admissions but takes the opportunity to add the latest Leaflet library to the uids_base build process for use in other situations.

When done and https://github.com/uiowa/uiowa/pull/2507 is merged in, it should provide Counselors info in the map. A click on the state updates the Counselors view block filter and displays a pop up info box. A link in the info box will take the user to  /counselors/{state_abbr}.

# How to test

- Take a look at admissions.dev.drupal.uiowa.edu/counselors (but don't touch)

OR

- `composer install && blt frontend && blt ds --site=admissions.uiowa.edu`
- Create a few person nodes with counselor person type and different combinations of territory values.
- As a webmaster in layout builder, you should see the Counselors Map block under More...
- Placed with the Counselors view block, this map will update the exposed filter on that view. A query string also works `/counselors?territory=IA`

**Note:** There are a few select options that are not on the map.


joewhitsitt:
Merging into DEV for review

richardbporter:
Attaching the map JS in a behavior got the map to load in the LB preview. It also allows easier access to the settings as they are passed in. I would recommend going that route but thats going to require refactoring other parts of that JS.

joewhitsitt:
Ready for technical review but not merge

cory-skeers:
It looks like the map needs a cache clear before it will update if a counselor's addition or removal to a state (whether the whole person is added/removed, or territory added/removed) which otherwise would or would not be highlighted. Only an issue for the case where the changes being made are enough to affect whether a state should be highlighted or not.

In the case of a counselor being removed, the state remains highlighted and the "view counselors" option, which takes you to an empty view. In the opposite case, the state remains unhighlighted, and no option to reach the counselor list is available.

Not sure of the frequency of updates once everyone is in place in the original setup whether something like a forced cache clear is necessary, or if it can just be something to be aware of.

Edit: Haven't approved in case you want to look into the caching, as well as possible notes from the meeting, but otherwise this has my 👍

joewhitsitt:
Per Admissions meeting
- [x] Get caching right
- [ ] Try query string filtering again (e.g. `?territory=ia`), if so, get rid of info box and scroll down? Assumes, view and map are always placed together. https://github.com/uiowa/uiowa/blob/main/config/admissions.uiowa.edu/views.view.counselors.yml#L199
- [x] Zoom out/reset
- [x] Circle, smaller image counselor teasers
- [x] down play territory info

joewhitsitt:
@pyrello you mentioned you might look at the views query on this view. I think everything else has been addressed. I have a question into the Drupal views community about this. No bites as of writing this. https://drupal.slack.com/archives/C6SQM2J94/p1605795331061600

If you are able to figure it out, I will remove the view page display and update the map to not provide that popup link and instead maybe directly scroll down to the filtered view. For this to be ideal, it assumes map and view are placed on the same page.

pyrello:
Pulled down for testing

pyrello:
I tested the counselor's block and map, which works well!

joewhitsitt:
Thanks for figuring out the query issue! I think we got MVP approval from Admissions.

docroot/profiles/custom/sitenow/sitenow.profile
```diff
@@ -975,3 +976,25 @@ function sitenow_entity_insert(EntityInterface $entity) {
...
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * @param \Drupal\node\Entity\Node $node
+ *   Get straight to the node instead of through entity.
+ */
+function sitenow_node_presave(Node $node) {
+  $tags = ['node_type:' . $node->getType()];
+  Cache::invalidateTags($tags);
```
richardbporter: Why are these necessary? Aren't these running for all node types, not just person?

docroot/themes/custom/uids_base/gulpfile.js
```diff
@@ -73,7 +98,7 @@ function watchFiles() {
   // @todo Watch other changes?
 }
 
-const copy = parallel(copyUids, fontCopy);
+const copy = parallel(copyUids, copyColorbox, copyLeaflet, fontCopy);
```
richardbporter: Why do libraries need to be copied into uids_base at all?

docroot/profiles/custom/sitenow/sitenow.profile
joewhitsitt: Figured we might use the node_type in the future for other content. the default node_list wasn't specific enough for me.

pyrello: I think the point here is that there should be a check by node type, so that it doesn't run for every type of node.

docroot/themes/custom/uids_base/gulpfile.js
joewhitsitt: I was just following #1200. I thought they had to be local for use in libraries, etc.

docroot/profiles/custom/sitenow/sitenow.profile
richardbporter: Yea I don't know the reasoning behind this as it seems like default cache invalidation should work but I'm guessing this is going to result in A LOT more purges for all of our sites. I could test this theory if necessary but I think we might want to limit this to admissions person type specifically.

docroot/themes/custom/uids_base/gulpfile.js
richardbporter: IDK off the top of my head. It seems odd that a library would not be usable at the default location. Thats were FontAwesome is located and AFAIK that is working.

docroot/profiles/custom/sitenow/sitenow.profile
joewhitsitt: I had to implement this because the map was not aware of node add/update/delete for person because it is getting person data differently than through views. The block calls this cache tag. Nothing else so far should be using it so I don't see a lot of purges based on this if it isn't misused.

docroot/themes/custom/uids_base/gulpfile.js
joewhitsitt: i can try to use the default paths in the code base and remove if it works.

joewhitsitt: I hate when @richardbporter is right about things like this.

docroot/profiles/custom/sitenow/sitenow.profile
richardbporter: Oh gotcha. So this is a custom tag added in https://github.com/uiowa/uiowa/pull/2521/files#diff-7a4b444e952e83c3dd58f165a78332b15db945d70032fbaa8c2491bd7a912277R51?

I appreciate the abstraction but it still seems excessive to invalidate these tags for every node op on all sites when 90% of them don't use that tag. Considering how simple the hooks are to implement this for custom functionality, I would still vote limiting this to admissions.

docroot/themes/custom/uids_base/gulpfile.js
richardbporter: Sorry, just curious about things. :)

docroot/themes/custom/uids_base/uids_base.libraries.yml
```diff
@@ -163,7 +163,7 @@ colorbox:
     name: MIT
     gpl-compatible: false
   js:
-    colorbox/jquery.colorbox-min.js: { minified: true }
+    ../../../libraries/colorbox/jquery.colorbox-min.js: { minified: true }
```
richardbporter: I think you can use a docroot-relative path like we do for Chosen in the sitenow profile: https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.libraries.yml#L18

docroot/profiles/custom/sitenow/sitenow.profile
joewhitsitt: 395eceb - moved to admissions and limit scope to person

docroot/themes/custom/uids_base/uids_base.libraries.yml
joewhitsitt: done in cf1ee54

.gitignore
```diff
@@ -34,6 +34,8 @@ docroot/themes/custom/*/assets
 docroot/modules/custom/*/assets
 docroot/themes/custom/*/uids
 docroot/themes/custom/*/styleguide
+docroot/themes/custom/*/leaflet/dist
```
richardbporter: I think these can be removed now since all of /libraries is ignored.

docroot/themes/custom/uids_base/.gitignore
```diff
@@ -1,2 +1,3 @@
 !assets/js
 !assets/images
+!leaflet
```
richardbporter: Probably this as well?

richardbporter (changes requested):

richardbporter:
I'm getting an error trying to run the `watch` workspace script:
```
vagrant@local:/var/www/uiowa$ yarn workspace admissions_core watch
yarn workspace v1.22.10
yarn run v1.22.10
$ run-p es6-watch sass-watch
$ babel es6 --watch --out-dir assets/js --source-maps --verbose
$ node-sass -w --source-map true --output-style expanded sass -o assets/css --importer node_modules/node-sass-glob-importer/dist/cli.js
es6/counselors-map.es6 -> assets/js/counselors-map.js
Successfully compiled 1 file with Babel (1230ms).
The optional dependency chokidar failed to install and is required for --watch. Chokidar is likely not supported on your platform.
```

I think I have a fix for this error incoming.

richardbporter (approved):
Looks good. 👍 

Bonus points for getting colorbox out of our codebase. 💯

PR #2527, opened 2020-11-12
Graduate College migrations

### LINKED ISSUE #2334: [Grad College] Build article migration
The Grad College has about 450 pieces of article content, so ideally I would like to write a migration script to handle moving it.

### LINKED ISSUE #2540: [Grad College] Migrate Thesis Defense content
# Story

As a Grad College website maintainer I want to be able to batch migrate thesis defense content because there is a lot of it.

# Proposed solution

- [ ] Add a Thesis Defense migrate source plugin class.
- [ ] Add a Thesis Defense migration configuration.
- [ ] Testing

### LINKED ISSUE #2558: Create migration for Grad College SROP mentors
# Story

As a Grad College site owner I want to migrate SROP mentors via a migration process because the content type has a lot of fields and is more prone to human error.

# Proposed solution

- [ ] Develop a migration that can be used to migrate SROP mentors for the Grad College site.

### LINKED ISSUE #2559: Create a migration for Grad College SROP Scholars
# Story

As a Grad College site owner I want to migrate SROP scholars via a migration process because there are 420 of them and that would save a lot of time for the people migrating our content manually.

# Proposed solution

Build a migration to handle migrating SROP scholars for the Grad Colllege.

pyrello:
Resolves #2334 
Resolves #2540  
Resolves #2558
Resolves #2559 

* Adds a new feature split called `grad_migrate`, which duplicates and adapts some of the elements of the `sitenow_migrate` split.
* Adds a `grad_migrate` module to hold necessary classes for Grad College migrations.

# How to test

Code review should be sufficient to verify that none of the files being changed fall outside the Grad College site or updates to `sitenow_migrate`.


pyrello:
@cory-skeers I'm a bit stuck on this. I can get the article migration to show up. I can run it, but then nothing is imported. Not sure how to debug this either.

pyrello:
In an effort to make the existing elements more extendable for a variety of future migrations, I have re-factored the original source classes to extract out an abstract base node class as well as a trait for capturing functionality related to processing files into media. I am now working to understand the role of the `process` element in a migration config file.

cory-skeers:
Tested out the refactor with a d7->v3 migration, and all worked well, after pushing in some updates I'd had on another branch but hadn't made it into main yet (updates related to going to v3 rather than v2).

Should only effect grad-specific migrations if they were using the `updateInternalLinks()` and related functions.

pyrello:
Need to account for images/files that have already been moved over - possibly in prepareRow? If there is an existing D8 filename that matches the incoming filename, how can I reference that media item for downstream migrations?

cory-skeers:
> Need to account for images/files that have already been moved over - possibly in prepareRow? If there is an existing D8 filename that matches the incoming filename, how can I reference that media item for downstream migrations?

Do you need to reference them in inline text, or for image fields? In the SiteNow D7->D8 path, any body text updates are based on filename, so if you skip over existing files, those updates will still work without any additional effort (and the migration YML process should have an option for skipping, renaming, or replacing for files). If you need to reference based on original fid (like article featured image), then that'll be fun. Maybe a prepareRow to set the d7->d8 mapping manually, even if it skips over actually copying the file?

pyrello:
I went on a journey to try to figure out how to use the supplied D7 node migration only to decide that it wasn't worth the complication because it requires installing the `migrate_drupal` module, which is a hidden dependency because the D7 Node class extends a class from there. This introduces more than a hundred potential migrations, which makes debugging issues much harder.

So, now I have copied in some of the features of the D7 Node source class into the BaseNodeSource class.

Next steps:
- [ ] Get files working in the case where some files/images may already exist. I'm not sure, but I think I may need to sync files from prod to local to test this since I'm not sure if it will find the existing image via stage_file_proxy.
- [ ] Determine whether it is necessary to have separate source plugin classes for Article and Thesis Defense, which currently do nothing extra on their own. It may be possible that any variation can be handled through the migration config file.
- [ ] Troubleshoot file errors related to exif data
- [ ] Possible to only migrate file if it is actually used once migrated?

cory-skeers:
> * Get files working in the case where some files/images may already exist. I'm not sure, but I think I may need to sync files from prod to local to test this since I'm not sure if it will find the existing image via stage_file_proxy.

I think this will still work as expected with stage_file_proxy. The check for if the file exists should still return true based on constructing a uri based on the filename, even if the file doesn't actually exist in the directory. A very unofficial check: I ran the migrate on 400 files which listed as having completed successfully, but only 332 files have been copied to the directory.

> * Troubleshoot file errors related to exif data

What errors have you been seeing? I've gotten a few 404 errors, but haven't run into others yet. Have only run a few small batch imports though, so have only hit a small portion of the total files.

> * Possible to only migrate file if it is actually used once migrated?

As long as the `file_usage` is accurate, one way of doing this could be to add a condition to the `files::query()` to limit it to only returning those files which have ids in the usage table. Edit: I pushed a change that does this (basic constraint of having at least one entry in file_usage), and for grad.uiowa.edu it dropped files to migrate from 4705 to 4586. Should be able to narrow it down more if there are additional constraints on what entries in the file_usage table are useful (not revisions, not just the file entity itself), but that might take some more digging.

pyrello:
I'm running into a command line error when attempting to run `drush @grad.local ms`:
```
[error]  TypeError: Argument 1 passed to Drupal\migrate\Plugin\Migration::getProcessNormalized() must be of the type array, null given, called in /var/www/uiowa/docroot/core/modules/migrate/src/Plugin/Migration.php on line 644 in Drupal\migrate\Plugin\Migration->getProcessNormalized() (line 396 of /var/www/uiowa/docroot/core/modules/migrate/src/Plugin/Migration.php)
```

Steps to reproduce:
1. `blt ds --site=grad.uiowa.edu`
2. `drush @grad.local en grad_migrate`
3. `drush @grad.local cim --source ../config/features/grad_migrate --partial`
4. `drush @grad.local ms`

When I navigate to https://grad.local.drupal.uiowa.edu/admin/structure/migrate/manage/sitenow_migrate/migrations, I am seeing the migrations I would expect without seeing this error.

cory-skeers:
I'm not running into that issue (tested it again after the latest 'main' merge). Not sure where the difference between our experiences with it would come in.

pyrello:
I recently had to `vagrant destroy` my environment. I wonder if it is somehow related to that? I'll run through this again and see if I still get the same result.

pyrello:
Huh... not sure what the problem was. It's working now.

pyrello:
Here are a selection of error and warning messages I get when running `drush @grad.local mim d7_grad_file`:

This first one is the most common by far:
```
[warning] A non-numeric value encountered PelIfd.php:447
```

These next several seem like they might be related to focal point?
```
[error]  Window [0, 48061] does not fit in window [0, 8956]
```
```
[error]  Offset 6000 not within [0, 5999]
```
```
[error]  Invalid marker found at offset 0: 0x89
```
And this is the original error I was talking about before that seems to be related to exif data:
```
[error]  Size of Canon Camera Settings does not match the number of entries.
```

pyrello:
@cory-skeers To clarify this point:

> Get files working in the case where some files/images may already exist. I'm not sure, but I think I may need to sync files from prod to local to test this since I'm not sure if it will find the existing image via stage_file_proxy.

Taking `stage_file_proxy` out of the equation, what I was actually intending with this is that there may be files that have been manually migrated already. There is a manual content migration already underway on the Grad College site, so a good number of the files have already been moved to the D8 site as part of that process. I want to make sure that the process is not creating duplicates of files that have already been moved over.

cory-skeers:
> I want to make sure that the process is not creating duplicates of files that have already been moved over.

The `file_exists: use existing` in the `file_copy` plugin should get the behavior you're looking for as long as the filenames haven't been changed in the manual migration process. 

That said, it looks like the `migrate_map_d7_grad_file` table ended up with a row for every file from the existing site, none of the destination ids match the existing fids, and all the original fids still exist...so even if it correctly avoided downloading the files themselves, it created new file entities that aren't needed/desirable. Not sure exactly what's going on--the numbers for downloaded files, entities in the `file_managed` table, files errored, and rows in the `migrate_map_d7_grad_file` without an error status all aren't lining up as expected...

cory-skeers:
> [warning] A non-numeric value encountered PelIfd.php:447

Good news is, in trying to track this down, it looks like the warning gets thrown many times for a single instance (38 warnings for one image). Unsure of why it gets attempted so many times, but at least the overall number of files triggering the warning is fewer than it appeared. Also seems that the file gets imported correctly, so it can probably be safely ignored.

Bad news is I still don't know what's causing the issue. https://www.grad.uiowa.edu/sites/gc/files/u7/LUKE.jpg is an example image for which the warning gets thrown.

Another example: https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/IMG_0156_0.jpg

cory-skeers:
> ```
> [error]  Window [0, 48061] does not fit in window [0, 8956]
> ```

https://www.grad.uiowa.edu/sites/gc/files/portal_bgs/departments_and_programs.jpg was an example of one getting this error. It looks like the full file isn't getting added to the directory (9 kb vs. 62 kb), leading to an encoding issue. Manually downloading the file doesn't lead to the same issue

https://www.grad.uiowa.edu/sites/gc/files/image_attachments/2016_07_08-New Hancher-jatorner-0015.JPG

cory-skeers:
> [error]  Invalid marker found at offset 0: 0x89

https://www.grad.uiowa.edu/sites/gc/files/u418/Matt%20Drabek%20photo.jpg is an example of this error. Image gets added, and media created without any issue.

Looks like another encoding issue, where the data is starting with a PNG file signature, but the file has a JPG extension, but in this case despite the error, execution completes without issue.

Edit:
Similar, but slightly different issue here--
u20/Chad-Rhym.jpg
` [error]  Invalid marker found at offset 0: 0x50 
 [error]  ImageMagick error 1: identify-im6.q16: insufficient image data in file `sites/grad.uiowa.edu/files//Chad-Rhym.jpg' @ error/jpeg.c/ReadJPEGImage/1039.
 [command: identify -format 'format:%[magick]|width:%[width]|height:%[height]|colorspace:%[colorspace]|profiles:%[profiles]|exif_orientation:%[EXIF:Orientation]\n' 'sites/grad.uiowa.edu/files//Chad-Rhym.jpg'] `
Not sure what file signature it has that is different, and looks like this one did not get into the file directory/media created properly, but there is a second copy that did make it in without issue.

cory-skeers:
> [error]  Size of Canon Camera Settings does not match the number of entries.

Example file that gets this error: https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/Domann_0.jpg

Similar to some of the other errors, the file does get migrated, media entity created properly despite this. Would be good to catch and address this, but likely safe to ignore.

Other examples for this error:
https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/SROP06_raafi2.JPG
https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/SROP06_IMG_4624.JPG
https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/SROP06_silnettra4.JPG
https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/SROP06_Jennife2r.JPG
https://www.grad.uiowa.edu/sites/gc/files/srop_attached_images/SROP06_Alicia2.JPG

pyrello:
# Workflow notes
The following steps are needed to get the site ready to run or test migrations. While testing and debugging, these steps only need to be done when first switching to the feature branch or if newly added content is needed to test against. When getting ready to actually run the migration and move the content to production, the production site needs to be "frozen" so that no one is making any changes to it.
* `blt ds --site=grad.uiowa.edu` - Sync grad.uiowa.edu locally. 
* `drush @grad.local en grad_migrate` - Enable the `grad_migrate` module for that site. 
* ~~`drush @grad.local cim --source ../config/features/grad_migrate --partial` - Register and import the `grad_migrate` feature split.~~ This is no longer necessary as it happens automatically during the previous step.

At this point, the migrations should be setup and able to be run. The following commands are useful for testing a migration while you are still making changes to it.
* `drush @grad.local ms` - Show the status of all migrations.
* If changes are made to a source config file (e.g. `migrate_plus.migration.d7_grad_thesis_defense.yml`), then config needs to be re-imported for changes to take effect: `drush @grad.local cim`
* `drush @grad.local mim d7_grad_thesis_defense --limit=5` - Doing a limited test run can allow you to check the content being migrated, without having to wait for all of it to migrate.
* `drush @grad.local mr d7_grad_thesis_defense` - Roll back migrated content. Rolling back content allows you to test the same migration command multiple times, so is helpful if you are still actively developing the migration.

pyrello:
Todo items:
- [ ] Get Thesis Defense URL Redirect migration working.
- [x] Figure out how to migrate Defense Date smart date field.

cory-skeers:
@pyrello For the thesis defense URL redirects, do we need redirects for the old path pattern (thesis-defense/title) to the new pattern (thesis-defense/date/title), or are there potentially manually created redirects from the D7 site that also need to be considered?

pyrello:
@cory-skeers I was only thinking about creating new URL redirects to account for the change in path alias pattern between D7 and D8, but you raise a good question that I hadn't previously considered. There probably are some redirects that should be accounted for. I'll have to give this a bit more thought about how best to proceed.

pyrello:
FYI, I had been thinking about using a separate migration that would run after the Thesis Defense migration for the redirects, along the line of what is done in this example: https://github.com/evolvingweb/migrate_example_paths/blob/8.x/config/install/migrate_plus.migration.example_article_redirect.yml. I think it can be entirely defined in a migration yml file.

cory-skeers:
Is there an advantage in doing it through a separate migration over an event? Went with the event approach because it's more in line with how I think of the redirects as part of the migration process and I'm not familiar with the separate migration approach, but willing to pivot.

Right now the code works if it's in `sitenow_migrate`--think it's not firing either due to a conflict having `onPostRowSave` in both modules, or maybe a typo somewhere in its definitions (two benefits of the single YAML for sure). Wondering if it's better to leave it in `sitenow_migrate`, debug the issue with keeping it in `grad_migrate`, or return to your initial approach through the separate migration.

pyrello:
If you can just add a yml file, then it is easier to add it without changing anything. Also, the URL redirect can cleanly run after the associated migration runs without issue. This seems better to me for maintainability since it means you don't have to mess with the `onPostRowSave` event if you want to add more of these later.

cory-skeers:
Generally, if there's not a time you'd need to run the migration without the redirects being created, I'd prefer to use events and keep it as part of a single migration. There's the benefit in testing of being able to run partial migrations rather than needing to run the full node migration prior to the redirect, but beyond that, it's my personal preference/opinion.

I can take a closer look at the separate migration approach. If the `onPostRowSave` event doesn't actually have a conflict with `sitenow_migrate` (or if it does, but there's an easy workaround), I wouldn't be concerned about maintainability or the difficulty of adding additional redirects. But definitely don't want to start bloating `sitenow_migrate` more than it already is.

cory-skeers:
On a related note: When running the Thesis Defense migration, I'm getting 448 imported, but still a remaining 64 unprocessed. They aren't erroring, there are no messages in the `migrate_message_...` table, and the `migrate_map...` table only has the 448 rows (if a row were skipped w/out error, it should show up here with a `NULL` in the `destid` column). Not sure why they're not being processed, and it also prevents it from being marked completed for dependent migrations.

Edit: looks like the multi-value thesis chair field isn't being handled properly. It has 512 values, though there are only 448 thesis defense nodes, and the migration is getting confused.

cory-skeers:
Working on better handling for multi-field values now, which should resolve the 448 v 512 row discrepancy.

pyrello:
@cory-skeers Great! I was wondering about that, but I had no idea how to proceed.

That is a good point about having to run all the Thesis Defense migrations before running the TD URL Redirect migration. I'm wondering if we can get around that, for testing, by just removing the dependency and running them in order? It seems like the second one would only run against whatever had already been processed in the first. There's a little more to remember, process-wise, but it still seems cleaner to me when thinking ahead to doing a bunch of these.

cory-skeers:
Took a stab at the multi-value field handling with a new method on the `BaseNodeSource`. Could likely use some cleanup and maybe a rethink, but I think the general approach works: Fields that may or may not have multiple values get queried in `prepareRow` and added to the row.

I think you could accomplish a similar effect in the query using pivots, but it could get complicated fast, and this seemed the more easily expandable approach to me.

cory-skeers:
@pyrello True--removing the dependency during testing should be fine. And especially as all migrations will need to be handled by a dev using a local environment, I'm not too concerned about maintaining strict migration dependencies.

I got it working using this approach two ways, the first with some extra code added to the thesis defense source `prepareRow`. It's functional, but I don't like that we're re-running all of our queries when only one or the other type is necessary. It also loses out on the benefit of being self-contained to a YAML file this way.

Second attempt builds what _should_ be the slug based on first and last name fields (unsure if there are exceptions to this format), which were already available in the source plugin. Generated redirects are technically stored in titlecase which is slightly annoying, but functional nonetheless.

pyrello:
For the Thesis Defense, that appears to be working except for the Degree Major field. This is because the list that was used for this field is changing and there is some additional work that needs to be done to map the values from the first list to the second list. I'll be coordinating that effort next week, so we should be able to wrap that migration up pretty soon. I'm thinking it will just be a method on the Thesis Defense source plugin class that maps the old allowed values key to the new one via array. I'm adding some code stubs with comments as a placeholder for this.

At that point, I think it might be a good idea to split the rest of this PR into a new PR so that we can close this one and the related issue hanging out in our In Progress column.

pyrello:
The build is going to be failing for a bit while I finish working through the program list mapping.

cory-skeers:
There'll be more work needed I'm sure, but for the most part, the SROP scholar/mentor migration sources are set. The YAML mappings will need to be updated following #2834, but most of it can be reused with small updates (and that work can be split off into separate PRs.)

cory-skeers:
Getting this in w/the PR--looking to have a conditional file migration, where a separate, ante file migration wouldn't be required, and files migrated if/when they are needed for a given page.

- If there's a field_thumbnail_imageattached, check if the file exists on the new site and if so, get the correct mid, and if not, download the file and create a media entity, then map the mid
- Check for inline references in the body and, similar to above, map to an existing mid or download, create media, and map
- On migrate rollback, remove any files and media entities that were pulled in based on the article migration (in writing this, realized we'd have to also check that the file hadn't already existed, and wasn't being used elsewhere, so we don't accidentally remove files/media that overlapped)

Edit: Working on getting this going now, and will push up as progress is made.

pyrello:
Update about Thesis Defense blocker: I generated a list of the legacy program codes that are in use in the D7 Thesis Defense items and delivered this to Andy, who is going to use SQL to put together a map then manually update it where necessary for values that are not mapped correctly in the records he has access to. I attempted a programmatic solution for this via Python, but there is some information that is just not available via the data that needs to be manually adjusted.

cory-skeers:
Articles should work now to pull in files used both in the thumbnail image and in the body (the latter I haven't checked thoroughly--need to do a quick query for nodes that have inline entities to check against). The migration attempts to download the referenced file, create a media entity, and associate that media entity. Skips over files that it can't reach--404 errors, at least in part for trying to reach files in the public directory that are actually located in the private directory.

Also removes media entities upon article rollback, with 2 caveats: 1) it currently doesn't check that the media entity is used only by the articles migrated, and so could delete media entities that are used elsewhere; 2) it doesn't remove the files themselves, and relies on the system file cleanup...not 100% sure that is adequate to lead to its deletion.

pyrello:
I'm working on an update to the allowed values function for the degree majors for the Thesis Defense, but that doesn't really affect this work. The thesis defense migration runs without issue and spot checking the results seems to indicate that it all working, so I think we could consider this work done and make any adjustments that come up in follow-up issues.

pyrello:
@cory-skeers Are we doing any accounting for the Author field in the Article content type? We are not going to migrate that content type to the new system and instead will be using the Author field in the new platform, which points to the Person content type instead. Maybe we could check the title of the old Author item (there are only 5) and check for a matching Person record in D8 and provide that as the NID for the Author field?

cory-skeers:
As of now there's no work done with the Author field, but if there are already People to map to, that should be straightforward--even better if it's only a small number and it was a controlled field rather than free text.

I'll take a look.

cory-skeers:
Author migration seems to be working for articles--

Couple todos still listed in the row query:
```
      // @todo Join article people reference.
      // @todo Join tags reference.
      // @todo Join programs reference.
      // @todo Join editorial group reference.
      // @todo Check D8 migration status.
```

And there are a few fields (`lead`, `pull quote`, `photo credit`) that aren't handled but could be, probably inserted into the body content.

pyrello:
I checked the usage of the fields indicated in the last comment and updated the todo comments in the `Article` class to reflect the remaining ones that need to be dealt with. In all other cases, there are only one or two instances where it is used.

For 'Tags' I think we can just migrate it to the tags field on the new Article content type. In looking at the `sitenow_migrate` module, it looks like the Article migration there doesn't account for tags, so we would be having to figure this out. We could potentially handle it a couple different ways:
1. Create a separate Tag taxonomy migration and make it a dependency of the Article migration.
2. Handle it similar to how we handled files and embed the process of creating the term into the Article migration.

The second option sound easier for testing, but I'm maybe more inclined towards the first option, since this isn't a case where I think we need to selectively migrate the tag terms. I'd be interested in here your feedback @cory-skeers.

For the 'Program(s)' field, I need to look into how we want to proceed with that. Either we'll be merging it into the 'Tags' field on the new content type, or we'll be creating some code to compare it to the programs list that I'm adding in #2872. I want to check with the folks who actually write the articles on our site to see what they think.

cory-skeers:
There might be a process plugin that works for mapping taxonomy terms similar to the second option... I'll check on that.

If there isn't, I'd be inclined toward the first option as well.

Edit: Maybe there isn't... Then yes, I'd lean toward the separate migration, if there's no strong need to limit what gets pulled over. Especially with there not being actual files downloaded, and the trickiness of files and media being created, I think the added step/complexity to testing is worth the benefit of lower upfront cost in building it and more reusability options for other sites later.

pyrello:
We are going to move "Programs" into "Tags." Let me know what clarifications you need on this point.

cory-skeers:
@pyrello Would it be helpful to pull all taxonomies into Tags, or only Programs and Tags? I forgot there were other taxonomies to consider, like Editorial Groups. If we don't need to pull all taxonomies in, then we might want to go back to option 2, and set it up as a processTaxonomy trait that can be used for both Tags and Programs (and future taxonomies).

In either case, also wondering if terms should be prefixed by their source taxonomy if they're dropped into tags to offer some sort of separation--there's been some discussion about doing a similar thing for the case where custom fields are transitioned to using tags, including the field as a prefix to the value to make the tag.

pyrello:
We only need to move Programs and Tags. The rest should all be handled via different paradigms now.

If we prefixed the source, would that end up with Tags that read as "Program: Accounting" or something similar? If so, I think I'm not a fan of that idea, because it will be confusing for content editors going forwards and someone is definitely going to forget.

pyrello:
In testing the article migration, I noticed that inline image replacement wasn't working. I tried and failed to fix it with regex - not being a regex guru :| - and ultimately came across and example using the `Drupal\Component\Utility\Html` class to parse the body HTML to match images, generate their replacements, and perform the replacements.

I have not yet tested the Thesis Defense migration to ensure that it is still working.

# Remaining issues with Article migration

## Title is too long for 'name' column.
https://www.grad.uiowa.edu/news/2019-06-03/julia-lucass-walk-tennessee-to-arkansas-part-ii is an example where the images have `title` attributes that are too long to be inserted into the database for the D8.

## Lack of immediate file cleanup on rollback is causing some issues.
If I'm understanding the rollback code, it looks like we are leaving it to cron to clean up files, but to my knowledge, cron doesn't run on my local machine. I spent some time spinning my wheels trying to figure out why files were being reported as already there. More than anything, this is an awareness thing. When it is time to run the actual migrations, it is important to make sure that files are cleaned up. I think I'm going to just remove all the files in the directory beforehand, then rsync them from prod and then run the migration. I'll likely do a dry run or two first to assure myself that there are no issues.

## No handling yet (in refactored code) for case where image is not able to be retrieved for some reason.
This was previously handled by returning the original regex match, but I have not yet figured out the equivalent functionality with the way that I'm handling replacements now. I think that the `` node probably just should be removed.

## `@file_get_content` error suppression makes me nervous.
I may just need to understand what it is doing better, but it makes me think that we could be missing conditions that we should be figuring out how to handle.

## Nice to have: Attempt to match file against version with `_0` or similar parts of the stripped out.
That is usually a sign that the file is a duplicate. It would just be nice to identify and eliminate more of them. This is not important.

cory-skeers:
> In testing the article migration, I noticed that inline image replacement wasn't working [...] and ultimately came across and example using the Drupal\Component\Utility\Html class to parse the body HTML to match images, generate their replacements, and perform the replacements.

Interesting! I'm a fan of regex generally, but was not a fan of how fragile that approach had been (at least one major update had to be done following a media update, and something similar might have been the issue here). Will need to test with other migrations, but I don't see any reason this shouldn't work for all content.

> Title is too long for 'name' column.

Dang. Should be a quick update--I believe there's even a built-in Drupal method to handle this?

> Lack of immediate file cleanup on rollback is causing some issues.

Yep--the files are left there, and there isn't a check that the media files being removed aren't being used by another piece of content. It's possible that a media entity already used on existing (non-migrated) content would be removed on rollback if it were also used by new (migrated) content. 

Currently it's really only somewhat useful in testing, and without additional work, I think your plan for a clean migration after a test run or two is best.

> No handling yet (in refactored code) for case where image is not able to be retrieved for some reason.

Part of returning the original match was to make it easier for manual cleanup later, so removing it makes me nervous. Maybe could replace it with a `<!--...--!>` token so regular editing can be done, but there's a token for db queries and manual cleanup or replacement later?

> `@file_get_content` error suppression makes me nervous.

Should double check, but all of these errors should be 404 errors when attempting to download the file. Some of these were images that were true 404s, and others are files that are actually in the `private://` directory, and the attempted `public://...` path is 404ing. If it errors, no file content is returned, and the rest of the code gets skipped over.

For the private files, manually downloading and adding to the site prior to the migration is one way to handle it, assuming they should be added. I'll pull up my file notes from earlier.

> Nice to have: Attempt to match file against version with _0 or similar parts of the stripped out.

I think this is interesting, but could get complex fast. I've worked in a lot of CMSs where there are many files/images that get named the same thing but are actually unique (man_with_computer.jpg and man_with_computer_0.jpg gets added later, and they're just two separate images of different men with different computers).

Could check the file contents themselves, but there may need to be some handling of EXIF or metadata that could be updated without actually changing the image. Converting the raw data to a string and regexing it would be a small in the end, but tricky to figure out.An MVP would work without this, and might handle most if not all files without an issue, just not "robust" without it.

pyrello:
> Part of returning the original match was to make it easier for manual cleanup later, so removing it makes me nervous. Maybe could replace it with a <!--...--!> token so regular editing can be done, but there's a token for db queries and manual cleanup or replacement later?

I'm good with a plan like this.

Regarding the nice to have, let's just not worry about it, since it does introduce more complexity and need for additional testing.

pyrello:
I re-added the SROP migrations here and closed those individual PRs.

cory-skeers:
Added a quick truncate before setting the media name. I'd thought there was a built-in Drupal function that checked for length restrictions on database fields and truncated if necessary, but now am not finding anything like it. These few lines do essentially the same thing, just with explicitly defining the length limit.

Also added a token replacement as `<!--Missing image: $filepath--!>`, as long as there was an existing src attribute, so we don't have broken images but can query for them and clean up later.

Still need to check on what is resulting in the missing images, and if they're something we can account for here.

pyrello:
I had also been wondering about whether it might make sense to through a [MigrationSkipRowException](https://api.drupal.org/api/drupal/core%21modules%21migrate%21src%21MigrateSkipRowException.php/class/MigrateSkipRowException/8.2.x) so that more investigation could be done?

cory-skeers:
My thoughts would be if the image(s) was the only issue encountered, the row should still be migrated so that nothing else needs to be re-migrated or manually brought over, either by getting a node list from something like `select entity_id from node__body where body_value like "%<!--Missing%";` following migration, and/or maybe changing the moderation state to Needs Review so there's an in-system marker that further followup is needed.

pyrello:
I realized as I was looking at the mentor migration that we were going to need the same image file processing for the image field on that content type as on the article migration, so I did some work to abstract that functionality out into another trait that can be added to migrations that need it. This forced a bit of a reworking of how the other trait is implemented. It is now necessary to add to each migration that needs it, at least until we can come up with a better way to organize all this code.

Started going through mentor migration fields to update them to new names. Mostly, there was not a lot of updating. Added some todos to in comments.

cory-skeers:
Got the initial program mapping set for the Mentor. 24 profiles don't have a direct mapping--looking to see what the programs are now.

Edit: Here's the list.
```
+--------------------------------------------+
| field_grad_program_all_degree_value        |
+--------------------------------------------+
| UPDATE: Anatomy And Cell Biology           |
| UPDATE: Anatomy And Cell Biology           |
| UPDATE: Anatomy And Cell Biology           |
| UPDATE: Anatomy And Cell Biology           |
| UPDATE: Chemical and Biochemical Engr      |
| UPDATE: Chemical and Biochemical Engr      |
| UPDATE: Chemical and Biochemical Engr      |
| UPDATE: Chemical and Biochemical Engr      |
| UPDATE: Civil and Environmental Engr       |
| UPDATE: Civil and Environmental Engr       |
| UPDATE: Electrical and Computer Engr       |
| UPDATE: Health, Leisure and Sport Studies  |
| UPDATE: Med and Nat Prod Chemistry         |
| UPDATE: Med and Nat Prod Chemistry         |
| UPDATE: Occupational and Environ Health    |
| UPDATE: Occupational and Environ Health    |
| UPDATE: Other                              |
| UPDATE: Other                              |
| UPDATE: Other                              |
| UPDATE: Other                              |
| UPDATE: Physical Therapy and Rehab Science |
| UPDATE: Psych and Quant Foundations        |
| UPDATE: Teaching And Learning              |
| UPDATE: Teaching And Learning              |
+--------------------------------------------+
```

pyrello:
I can take a look at this in a little bit. Thanks @cory-skeers!

pyrello:
Thesis Defense migration appears to still be working great!

pyrello:
When I run the Thesis Defense redirect migration, I'm getting some errors for duplicate entries. @cory-skeers If you have time to look and see if you have any ideas, I would appreciate it.

cory-skeers:
> When I run the Thesis Defense redirect migration, I'm getting some errors for duplicate entries.

I'll take a look!

cory-skeers:
Few changes to the file download pipeline took care of a lot of the initial article missing inline images--down to 38 articles that are missing an image post-migrate.

cory-skeers:
Haven't checked that it's the case for all of them, but at least the first redirect duplicate is the result of duplicate thesis defense pages:
https://www.grad.uiowa.edu/thesis-defense/laila-mccloud
and
https://www.grad.uiowa.edu/thesis-defense/laila-mccloud-0

Both are being migrated, and the way the redirect migration is setup, it's creating the same redirect source for each.

pyrello:
Thanks for looking into this. In the worst case scenario where this runs and these failures occur, then it seems like the redirect just doesn't get created? I've checked in to see if I can just delete the duplicates. Any chance you can easily get a list?

cory-skeers:
Yes--here's the list of failed redirects. Not as easy to get the source ids, but it's a start.

It should be the case that it just doesn't create a redirect, so more of an issue that the redirect that gets created may or may not be redirecting to the "correct" thesis defense node, and/or there being duplicates.
```
thesis-defense/Laila-McCloud
thesis-defense/Timur-Dogan
thesis-defense/Douglas-Blake
thesis-defense/Kara-Hageman
thesis-defense/Natalie-Del Vecchio
thesis-defense/Stephanie-Lynch
thesis-defense/James-Oudenhoven
thesis-defense/Caleb-Lambert
thesis-defense/Carol-Seehusen
thesis-defense/Vu-Thuy-Nguyen
thesis-defense/Kara-Hageman
thesis-defense/Michael-Salino-Hugg
thesis-defense/Anton-Kananovich
thesis-defense/Michael-Salino-Hugg
thesis-defense/Jacob-Siebach
thesis-defense/Amos-Stailey-Young
thesis-defense/Alana-Tholen
thesis-defense/Shelby-Herig
```

pyrello:
This is great! Thanks!

pyrello:
It looks like you can add multiple process plugins to a field by just going to a format like:
```yml
field:
  - plugin: plugin_one
  - plugin: plugin_two
```
https://www.drupal.org/docs/contributed-modules/migration-tools/migrate-process-plugin-recipes#s-address-process-recipe

pyrello:
Articles and thesis defenses are migrated! I performed the migration locally, tested it on dev, then pushed it to production.

One thing I didn't think of before is redirect migrations for mentors and scholars. I think that can pretty much just be a copy of the one for thesis defense.

Mentor titles need to include the credentials, which I think I have working, but there is some weirdness with how it is displaying that I need to look into. Also, it didn't look like the image field was working for mentors.

I haven't tested scholars yet.

richardbporter (approved):
A very high-level approval. :)

PR #2542, opened 2020-11-16
Updated default menu path pattern to reflect parent path properly.

### LINKED ISSUE #1701: Pathauto not creating proper nested path
Assume we are placing all nodes under the Main Menu, and there exists the following structure:
Main Menu
 -- About (/about)
 -- People (/people)
 -- Undergraduate Programs (/undergraduate)
 -- Graduate Programs (/graduate)
 ---- Admissions (1)
 -- Resources

Note that the Undergraduate and Graduate Program pages have been given custom URL aliases . When a new node is added underneath one of them, and the "Generate automatic URL alias" checkbox is checked within the URL alias tab for this child node, the automatic path created does not carry over the custom URL from the parent.

The path for Admission (1 above) should be /undergraduate/admissions, however the automatic path created is /undergraduate-programs/admissions. The alias pattern is currently set to:
`[node:menu-link:parents:join-path]/[node:title]`
Is this the proper pattern? [https://www.drupal.org/project/token/issues/2769299](https://www.drupal.org/project/token/issues/2769299) lists an issue but it was from 4 years ago, and they say it was fixed.

mark-bennett-uiowa:
Updated default menu path pattern to reflect parent path properly. Parent paths were not using custom paths set in nodes when creating child pages underneath these pages.

By setting pathauto pattern to [node:menu-link:parent:url:path]/[node:title], this seems to have fixed this issue.

# How to test

- Set a custom path on any node.
- Create basic page child and grandchild nodes underneath this parent node, using any mixture of custom and automatic URL aliases.
- Make sure proper paths are getting set if automatic URL alias is used, reflecting any custom parent paths.

### DIFF
--- a/config/default/pathauto.pattern.menu_path.yml
+++ b/config/default/pathauto.pattern.menu_path.yml
@@ -9,7 +9,7 @@ _core:
 id: menu_path
 label: 'Menu path'
 type: 'canonical_entities:node'
-pattern: '[node:menu-link:parents:join-path]/[node:title]'
+pattern: '[node:menu-link:parent:url:path]/[node:title]'
 selection_criteria: {  }
 selection_logic: and
 weight: 0


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Thanks for bugging us about this. I think this is the ideal. In fact, I think this is needed and paths/redirects are updated on the next save of the child page.

I will say, it can get confusing depending on the order of existing pages that are saved, but all end up properly redirected in the end. It also is more intuitive to the user who is creating custom paths.

mark-bennett-uiowa:
Thanks so much @joewhitsitt , I appreciate the testing! Merging into main.

Issue #2346, opened 2020-10-15
Inconsistent Entity Usage with Media in block content


joewhitsitt:
## Issue

The entity usage UI table doesn't seem to be getting usage correct for block content. I am seeing usage (sometimes) for text area block media references. I thought I recently fixed this.

Example: This image is being used in more places than this one recorded usage.
https://sandbox.prod.drupal.uiowa.edu/media/36/edit/usage
https://sandbox.prod.drupal.uiowa.edu/cards

## Proposed Solution

- [x] Revisit entity usage configuration again as it relates to layout builder to get it working. Consult the documentation provided by the module/issue queue on how to set it up properly:
https://www.drupal.org/project/entity_usage
- [x] Confirm that there is no loss of information as it relates to v2 paragraphs.

PR #2545, opened 2020-11-16
Add video functionality to banner block

### LINKED ISSUE #2403: Add background video loop functionality to banner block
# Problem 

The current banner block does not allow for a video loop to be uploaded into the block. 

# Proposed Solution

Create a "Video" option within the "Background" options of the banner block where users can upload local video files. 

## Next steps

- [ ] Integrate UIDS work from uids#394

## Accessibility fixes

- [ ] Change markup in hero block template to disable autoplay by default and dynamically change based on block level override. 
- [ ] Add override checkbox/field to hero block in layout builder tray to enable autoplay.
- [ ] If autoplay is enabled, add sufficient warning text about vestibular triggers

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/2403. 

## How to test video and autoplay checkbox
1. `composer install`
2. `vagrant up --provision`
3.  `blt ds --site=admissions.uiowa.edu`
4. `blt frontend`
5. Add a video to an existing banner block, and verify that the video is loaded in paused state with a preview poster image both on the layout builder preview and saved page. 
6. Verify that the "autoplay" checkbox only appears if a video has been selected from the media library. 
7. Edit the banner block and check the "autoplay" checkbox and verify that video autoplays in layout builder preview and once the page is save. 
8. Edit the block again and uncheck the "autoplay" box and verify that autoplay stops in layout builder preview and saved page. 
9. Open the page in a new browser window and verify that the video loads based on your settings. 
10. Add another video banner to the same page and verify that you can set the autoplay option for the new banner block. 
11. Change your system accessibility settings to "Reduce motion" and verify that `prefers-reduced-motion` stops any video from autoplay. 

## How to test video cookie
(If at any point you would like to view the video uuids that are paused or how many videos the site is tracking for pausing, you can put `JSON.parse(jQuery.cookie('uiowa-video'))` in your console)
1. pause any video used for a banner in a published page.
2. Upon page reload, you should see that the video is still paused. You should also be able to move away from the page and come back and observe the same behavior. (If you were really die hard about testing this cookie, you could wait the 99 years that it will last to make sure it expires correctly)
3. create more video banners on the same page and test that they all have play and pause cookies that function independently.
4. Create a new page that also contains any number of video banners. 
5. Check that you can also set cookies for pausing on this page as well.
6. make sure that you can navigate away and between your video banner pages, and that the cookies persist through page closing, offsite navigation, etc.


pyrello:
It would be nice if we could just have a generic class that gets dropped on the video other than `.banner__video` so we don't need to pass in a component-specific class to the media field. Could we target something like `.banner__media .video` instead?

pyrello:
I just realized that if we set the autoplay field on the media entity itself, the autoplay setting will be persistent anywhere that it is used. This may not be the desired behavior in the event that a video asset is used in more than one place. 

Also, if we leave the autoplay field on the media entity, we'll need to figure out how to add an edit button to allow updating the setting from the banner form.

bspeare:
@pyrello it seems like this might not be working as it should: https://github.com/uiowa/uiowa/blob/feature_banner_video/docroot/themes/custom/uids_base/uids_base.theme#L207-L211.

If the `autoplay` checkbox is checked on the banner block, the `autoplay` attribute on the `video` doesn't change until the cache is cleared on the site.  Do you think that this is because we are not setting the `$view_mode` in the lines above?  You can see the `$view_mode` at https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_entity_view/8.2.x and https://drupal.stackexchange.com/questions/245610/preprocess-image-how-to-get-parent-entity.

bspeare:
@pyrello I'm getting an ajax error when I try to use the https://www.drupal.org/project/media_thumbnails_video within layout builder:

**Steps to reproduce:**

- Upload a new video to the banner block
- Click "save and insert"
- Check console log for ajax error

Do we want to try and add an image field to the video media type and use that as the poster image?

Screen Shot 2020-11-24 at 10 13 45 AM

bspeare:
Some further steps for enabling media thumbnail module:

1. Upload a video to the banner block. 
2. Enable media thumbnail modules
3. Go to `/admin/structure/media/manage/video/display` and change format to "Video extended"
4. This adds a poster image to the video markup
5. Does php-ffmpeg still neeed to be installed?  I see this error in the logs`FFMpeg\Exception\ExecutableNotFoundException: Unable to load FFProbe in FFMpeg\Driver\FFProbeDriver::create() (line 50 of /var/www/uiowa/vendor/php-ffmpeg/php-ffmpeg/src/FFMpeg/Driver/FFProbeDriver.php).`
6. regenerate thumbnails at `admin/config/media/thumbnails/refresh`
7. Check files folder to see if thumbnails were generated

Steps to reproduce ajax error:

1. Upload a new video to the banner block
2. Click "save and insert"
3. Check console log for ajax error

joewhitsitt:
Does it need https://packagist.org/packages/php-ffmpeg/php-ffmpeg?

bspeare:
@joewhitsitt That package is appearing in my vendor directory after running `composer install`, so I'm not sure where the problem is occurring.

joewhitsitt:
Maybe a @richardbporter question, but do we need to update our box config to use this similar to imagemagick? box/config.yml

bspeare:
The `autoplay` attributes are being set on the video through these preprocess functions: https://github.com/uiowa/uiowa/blob/feature_banner_video/docroot/themes/custom/uids_base/uids_base.theme#L204-L240.    However, there is a caching issue if you try to update the autoplay settings after creating a block:

Steps to reproduce:

1. `blt frontend` and `blt ds --site=admissions.uiowa.edu`
2.  Upload video file to banner image/video field and save page
3.  Video should be in default paused state
4.  Edit banner block and click on the "Enable autoplay" field
5.  After updating block the video should autoplay on layout edit screen
6.  Save page and the video should appear paused still
7.  Clear cache and refresh page and video should finally have the `autoplay` attribute

joewhitsitt:
going to see how i can help with the library and caching

joewhitsitt:
`{{ devel_breakpoint(attributes) }}` is breaking build

joewhitsitt:
I think i resolved https://github.com/uiowa/uiowa/pull/2545#issuecomment-733245035 but not sure if the paths i have will work for remote environments.

richardbporter:
It appears to be the same on AC and local. If it wasn't, config split could be used.

joewhitsitt:
i just realized i may have messed up video banner looks/functionality by switching to the video extended formatter where it sets the dimensions. hoping someone closer to the functionality can clean up my mess without backtracking the ffmpeg stuff. https://github.com/uiowa/uiowa/pull/2545/commits/81a4029c417c99ef35ddd76afd8b0b3e1b1131ee#diff-62e175340ff03d78ee228b43639e0631fdd355370a648571e323224e226e645d

bspeare:
@joewhitsitt Were you able to upload a video through layout builder and get the thumbnail generated?

joewhitsitt:
Yes. Forgot to mention. You need to vagrant up --provision i think to get ffmpeg working

bspeare:
@joewhitsitt Thanks for figuring out the thumbnail module!

joewhitsitt:
> Without this, sometimes `$media->_referringItem` is not set. Doing this ensures that `referringItem` is always available. This is part of a larger issue in D8/9 where there is not always a reliable way to get referencing items in the context that you need them.

@pyrello Sorry, just saw this after ripping that out. I think I fixed the autoplay cache issue. So is this referringItem still an issue?

pyrello:
@joewhitsitt Not sure, but your approach might still work for what we need? I never observed that `_referringItem` wasn't working, so not sure about how best to test it, but maybe @bspeare knows?

cory-skeers:
Running into a few instances where the autoplay checkbox doesn't appear. It appears to happen when:
- Editing a layout which does not currently have a video present in any banner (image/background banners being present don't seem to have an impact, but when there is an existing banner video this hasn't been an issue)
- The checkbox appears when first adding the banner video, but is not present when updating the banner
- Saving the layout appears to fix the issue (related to first bullet?)

I'm guessing it has something to do with the block not having been saved in the system yet when the update form is called, but the fact it seems to work correctly if there is another video block present (not necessarily the one being updated) on the page has me confused.

Edit: Looks like there's a js error that might be the culprit:
image

cory-skeers:
The video doesn't show up in entity usage; neither do images in banners, though, so not a PR-specific issue. Related to #2346 and some of the previous issues with getting the default content to work properly.

bspeare:
@pyrello for reference, the `media--video` class is being added here https://github.com/uiowa/uiowa/blob/feature_banner_video/docroot/themes/custom/uids_base/templates/field/file-video.html.twig#L18 and https://github.com/uiowa/uiowa/blob/feature_banner_video/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L579

pyrello:
I believe that I have resolved the issue with the autoplay checkbox failing to display when updating the block after the video has been added. While this wasn't directly tied to the javascript issues that were occurring, their occurrence seems to have been causing the javascript related to the checkbox to not run successfully.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
>If you were really die hard about testing this cookie, you could wait the 99 years that it will last to make sure it expires correctly

Not this diehard, but what I tested seems to work as expected. I did one double take when switching autoplay on for an existing banner video, but realized I had paused it previously and the cookie was controlling it. anonymous session updated when auto play was toggled and the cookies seem to handle multiple video preferences.

I am sure I missed the reason why the form has separate JS for show/hide logic but that too seems work. Maybe we can revisit this later and use core #states or ajax functionality.

pyrello:
We tried a few different options to try to make it work with `#states`, but these were all hidden form elements that don't seem to work with it. If someone can figure out how to make it work, then I'm happy to get rid of the javascript approach!

PR #2562, opened 2020-11-17
Config entity types

### LINKED ISSUE #2551: Create college config entities
# Story

As a site visitor, I want to see information on colleges which are related to the content I'm viewing, so that I can understand the full context of the content, as well as choose to learn more about the college from their webpages.

As a site owner I want a field list from which to select individual colleges so that I can easily connect content to relevant colleges.

As a SiteNow administrator, I want to manage a list of colleges in a shareable manner, so that I can maintain and update the list for all SiteNow users with minimal service disruption.

# Proposed solution

Create a custom config entity for colleges with a label, description, and homepage (link). Content fields can reference these entities, and they can be managed through config and shared across sites.

Work was originally started on this in #2465. Can be pulled out of there and into its own home, and then referenced by the area of study content type in that PR.

cory-skeers:
Resolves #2551 

# Motivation

Create shareable entities for a single source solution for referencing institutional content. As a starter, set includes the 12 colleges and several academic units from https://provost.uiowa.edu/organization.

# Testing

Sync a V3 site
Go to Admin > Structure, check that Academic Unit is available.
Check that there are 17 academic units prepopulated, including 12 colleges and 5 non-collegiate
Create a new or edit an existing block and/or content type
Add a new field, the Academic Unit type (listed under "Reference")
Edit the field widget settings and choose College, Non-Collegiate, or both.
Edit a piece of content with the new field, and check that the available options match the widget selection.
Add a new College and Non-Collegiate Academic Unit
Check that the options in the newly created field include the newly created entity

Note: The formatter is set to pass entity values forward, but as they are not currently built into any content, they won't be visible on any content to which they've been added.


joewhitsitt:
Just noticed the changes to Brand's lockup generator. Find and replace all?

cory-skeers:
> Just noticed the changes to Brand's lockup generator. Find and replace all?

Gah! Tried PHPStorm's refactor option to update a namespace, and didn't pay close enough attention to the preview.

joewhitsitt:
I think this is good, though originally I imagined them together with possibly a college/academic unit radio value. This allows use to use a single entity reference field for both. An entity reference view can provide a filter if only colleges or only academic units are to be used/available.

pyrello:
> Gah! Tried PHPStorm's refactor option to update a namespace, and didn't pay close enough attention to the preview.

PHPStorm causing all kinds of issues recently :)

I think that we probably shouldn't include descriptions because their inclusion would require us to have to update them more often. In cases like this: https://admissions.uiowa.edu/colleges/college-education, we'll probably still need local content entities.

cory-skeers:
@pyrello Removed the description field to help keep the maintenance load light.

@joewhitsitt Refactored a bit, so that the config entities for college and non-collegiate academic unit are actual subclasses of the generic 'unit' rather than only extensions. The generic unit can't be created by itself, but references to 'unit' pull in both non-collegiate academic units and colleges, while their individual references still only pull in their own entities.

pyrello:
Pulling down for testing.

pyrello:
I have tried this twice now on the default site and neither time have the config entities shown up when I logged in. I don't need to do anything other than sync to be able to see them, correct?

cory-skeers:
The configuration files are getting renamed as part of the sync process, it looks like... Can dig into it post-meetings.

cory-skeers:
@pyrello I believe this has been fixed. With that, though, I think there may be an order of operations issue here with how the config sync works alongside enabling the module.

pyrello:
@cory-skeers It now seems to be importing all the config entities correctly. I have added 3 fields to the person content type, referencing: 
* Academic unit
* College
* Non-collegiate academic unit

The second two work as expected, but the first one does not. It is showing no results. I don't see any additional options to set.

pyrello (dismissed):
Screen Shot 2020-11-30 at 9 54 03 AM

I added all 3 variations of this field and gave each one a different label, but they are all showing with a default label and help text. I think the desired behavior would be that they would show the assigned label and help text.

cory-skeers:
@pyrello Nice catch! Updated the form widget so that it should now keep the entered label/description.

joewhitsitt:
pulling down to test https://github.com/uiowa/uiowa/pull/2562#pullrequestreview-541052343

joewhitsitt (approved):
Fancy. Custom help text came through which I believe resolves https://github.com/uiowa/uiowa/pull/2562#pullrequestreview-541052343.

I noticed you can uncheck all and not have any options for the field formatter. Doesn't error on content save and provides no results text, so I don't think this is an issue but validation could be added in the future to require at least one.

cory-skeers:
Thanks for the testing! I'll create a followup issue for a validation/requirement so it doesn't hold this up, but makes it onto the docket for improvement.

PR #2577, opened 2020-11-19
Make block permissions consistent across LB (V3 based) sites

### LINKED ISSUE #2567: Update Admissions Split with Permissions
# Problem

Admissions is having troubles with aggregator as webmasters. Might be out of sync permissions.

# Steps to reproduce

- As a webmaster on Admissions, I am able to use the RSS Feed functionality all the way from configuration to display.

# Proposed solution

- [ ] Check the admissions' permissions and update. If aggregator stuff still can't be done. Dig further and FIX IT.

pyrello:
# Updates
* Webmasters and editors are not allowed to create new blocks of the following types: Hero, Hero Page Title, and Spacer/Separator.
* Since the above blocks may be placed already, we need to maintain the ability to edit or delete them.
* Added/updated/removed permissions related to metatags, which changed the format of their permission strings a while back.
* Resolves #2567. 

**Note:** This does not resolve the issue of Hero blocks being available. I'm not sure why, but I'm going to tackle that as a separate PR to make this (and that) easier to evaluate.

# How to test

I suspect that if this works on the default site, it should work on all affected (V3-based) sites. Si nce it just affects permissions I'm comfortable with a code review and testing against the default site with the idea that we can fix any potential issues that might be introduced on a subsequent update. I feel pretty confident that it won't be necessary, but my point is that this shouldn't affect functionality, just access.


joewhitsitt:
When this is ready for review, I can check to see if this also resolves https://github.com/uiowa/uiowa/issues/2567

pyrello:
@joewhitsitt You are welcome to test this whenever. I'm still working out some testing details so more details will be coming shortly.

joewhitsitt:
Confirmed that #2567 is fixed by this. Thanks

config/default/user.role.webmaster.yml
```diff
@@ -21,36 +21,6 @@ permissions:
   - 'access media overview'
   - 'access media_entity_browser entity browser pages'
   - 'access media_wysiwyg_browser entity browser pages'
-  - 'access metatag tag advanced__canonical_url'
```
joewhitsitt, line 24: Was this supposed to happen to the default webmaster?

joewhitsitt: duh. nvm

joewhitsitt (approved):

PR #2599, opened 2020-11-23
Change Scholarships Views Query

### LINKED ISSUE #2568: Update Scholarships Functionality to Handle Scholarship that appears on all?
# Problem

This scholarship, https://admissions.prod.drupal.uiowa.edu/scholarship/collegiate-and-departmental-scholarships, has all boxes checked with the expectation that it should show pretty much everywhere. I believe this won't work with how the view filtering works as it is trying to limit with ANDs and not treat them as ORs.

# Steps to reproduce

Customer expects https://admissions.prod.drupal.uiowa.edu/scholarship/collegiate-and-departmental-scholarships to show up on all of there existing view configurations.

# Proposed solution

- See if there is an easy fix to the view logic to account for this type of scholarship without starting over.
- Check in with Admissions based on findings and about how to proceed.
- Show on all checkbox?

joewhitsitt:
Resolves #2568

Client expects that checking all the type/resident checkboxes, in the case of https://admissions.prod.drupal.uiowa.edu/scholarship/collegiate-and-departmental-scholarships should display that scholarship on each of their configured views.

This basically means that the views query should be switched to ORs instead of ANDs and that there are no duplicates with these results.

I removed the heavy adjustment to get ANDs to work and confirmed that the ORs were working as expected similar to the current site:

- http://admissions.local.drupal.uiowa.edu/first-year-resident-scholarships === /finances/scholarships-nonres-first-year-students
- http://admissions.local.drupal.uiowa.edu/first-year-non-resident-scholarships === /finances/scholarships-first-year-students
- http://admissions.local.drupal.uiowa.edu/transfer-student-scholarships (has a duplicate view in the page) === /finances/scholarships-transfer-students
- http://admissions.local.drupal.uiowa.edu/international-student-scholarships === /finances/scholarships-international-first-year-students

# To Test

- Check the above current site -> new site views for a match in results. If it doesn't look quite right confirm that the scholarship has been created and tagged the same way on the new site.


richardbporter (approved):
> http://admissions.local.drupal.uiowa.edu/first-year-resident-scholarships === /finances/scholarships-nonres-first-year-students
http://admissions.local.drupal.uiowa.edu/first-year-non-resident-scholarships === /finances/scholarships-first-year-students

I think those two are mixed up. The others look ok but the ordering is different. I'm assuming thats just the weights?

joewhitsitt:
@richardbporter You are correct on both counts. Good testing.

PR #2606, opened 2020-11-24
Create Grad College Mentor

### LINKED ISSUE #2550: Create Grad College Mentor content type
# Story

As a [SROP Scholar](https://www.grad.uiowa.edu/ogi/srop/past-scholars) I want to be able to view a page that describes a mentor and the research opportunity that is available in working with them so that I can learn about my choices.

# Proposed solution

In the grad.uiowa.edu split:
- [x] Create a Mentor person type
- [x] Either: Add mentor fields to person content type and add them all to the mentor person type
- [x] Or: Add a field collection entity (either paragraph or fragment) and add mentor-specific fields to that item. Add an entity reference field on the person content type to the field collection entity and then add that field to the mentor person type.
- [x] Add new person type fields to the person LB template
- [x] Add new config items to grad.uiowa.edu split and re-export

dnkemper:
Closes #2550


pyrello:
I should have clarified this point: The fields that are in the Mentor field group on the D7 site, where they match up with existing person fields, don't need to be recreated. This is just a weirdness of how the GC site was built. Sorry for the confusion.

dnkemper:
@pyrello OH! So sorry for misinterpreting that! Thank you so much for catching it so fast!

Issue #748, opened 2019-11-14
Investigate workbench access type solutions in Drupal


joewhitsitt:
# Story

As a custom site owner I want the ability to limit editorial access to content based on association with a term or group.

joewhitsitt:
Here are some initial notes:

- A new taxonomy reference field per content type needs to be created (taxonomy access) and thus **splits the content type config**. The field is an actual field (not dynamic like moderation) and is not hidden if we strip workbench access configuration away from it.
- HR already has a "Units" taxonomy that could work for this if we want to add this functionality platform wide and reduce some "custom." It would probably require an update hook to limit the existing "references" field to not allow "units" and move existing unit selections to the tax ref field. This would reduce duplication/confusion.
- A single role can be eligible for multiple sections/units. Users are added manually through the workbench access configuration pages and only if their role is allowed within the section.
- It would make sense that webmasters would have the permission to bypass access restrictions but then that leaves publisher, editor and viewer roles.
- I don't see a way to auto map users based on role so managing users would be a manual webmaster job. There is a comma-separated list add method and a select-all remove method to speed up the process.
- Alternatively there is a menu access method but looks like it could be hard to manage if auto-mapping is not a thing and the menu is not org-chart driven.
- there were a lot of warnings/notices like "array flip expected but got," but they didn't seem to affect functionality.
- Default setup would take too long with a config split already in place.
- I think it could all benefit from more help text and more visual indicators that an editor doesn't see the edit actions because... "This content is managed by ___ which you are not a part of. Please contact a webmaster."
- I could see this being setup for the 80% and rolled out even with existing sites, but we could choose the naming convention wisely as ultimately, webmasters will be going to "Workbench Access" to configure their "Sections"

joewhitsitt:
After a discussion on Friday, we think:

- this field should be available on all sites
- to make adoption easier, units is the name of HR's vocabulary.
- HR should migrate units into it from field_reference
- HR needs to re-wire contact block
- allow webmasters to configure access
- lots of docs to supplement it.

i don't know what the dentistry requirements are but that may change this issue (which content types (HR it is only on page).

pyrello:
This is functionality that I am interested in for at least the Graduate College site. We are using this on our D7 site to control page access by editorial group.

I'm not sure there is a use case for me on the rest of my sites as the teams that manage them are pretty small.

joewhitsitt:
bumped at demo

briand44:
[...]

Issue #1905, opened 2020-07-07
Permissions for a role between webmaster and administrator?


pyrello:
This is to document permissions that may require an additional role for users who have a higher technical ability but are not considered developers.

- [ ] Global Metatags?
- [ ] Workbench Access
- [ ] Sitemap https://github.com/uiowa/uiowa/issues/1258

joewhitsitt:
I am starting to see the need for "Web Admin" as I work on #2549 / #748 . Someone to manage content ownership permissions for the whole site...

pyrello:
@joewhitsitt Is there a reason that this wouldn't just be the webmaster role?

joewhitsitt:
@pyrello For grad (2 webmasters) in #2607, it probably doesn't have to be, but as I was tinkering, either an individual or a role needs to be assigned to the top-level editorial groups to effectively manage the groups underneath. If webmaster was that role and there was several webmasters on the site, the whole access control functionality IMO wouldn't be very useful as they could bypass restrictions.

In the future I think we want to stay away from the site ownership designation stuff in v1 standard because nobody used it and it became out of date. Another role (used sparingly, though I am not sure it will be) which has a bit more permissions than a webmaster seemed to make sense for this content access stuff. Webmaster was supposed to be used sparingly so we might just be repeating history with another role.

richardbporter:
It might make sense to rename administrator to developer to denote that access to all features and functionality require knowledge of the application workflow and participation in our team. If that was done, administrator might fit between webmaster and developer.

joewhitsitt:
should webmaster become "administrator" and administrator become "developer" so we can make a move away from the term webmaster?

richardbporter:
Something else to consider, which may be better suited for a separate issue but wanted to jot down before forgetting, is whether or not the administrator role should be available for manual assignment. That is, do we want to only grant the administrator role to those in the appropriate AD group?

briand44:
Created a separate issue for ^^. https://github.com/uiowa/uiowa/issues/4845

pyrello:
Wondering if there is still any need for this? I don't think so from my end. I'm inclined to close it.

briand44:
I'm in favor of closing.

joewhitsitt:
heads up we still have https://github.com/uiowa/uiowa/issues/859 and i do like the idea of moving away from webmaster when it aligns with other work (https://github.com/uiowa/uiowa/issues/1905#issuecomment-1040446018)

[...]

PR #2607, opened 2020-11-24
graduate college - workbench access

### LINKED ISSUE #2549: Create workbench access feature split
# Story

As a custom site owner I want the ability to limit editorial access to content based on association with a term or group.

# Proposed solution

Building on the work started in #748, build a feature split to allow for sites to turn on workbench access.

joewhitsitt:
Resolves #2549
Relates to #748

This PR establishes workbench access on the grad.uiowa.edu website split. Originally, I thought it might make sense to establish some of this for the default to limit configuration splits, but with the complexity surrounding it, I decided to start with a full split for now.

- Editorial Groups taxonomy
- Reference field on page, article, book, person
- Editorial Access scheme that is enabled for page, article, book, person

# To Test

## Setup
`composer install`

## Testing
As a webmaster, this role is auto added to the Editorial Access scheme and allows management of users per Editorial Groups terms. Taxonomy hierarchy works with this system in that a parent term can manage all content classified with a child term. Webmaster is the only role that can add/edit terms for this.

Users are assigned by the webmaster role via a user's page. There should be a workbench access tab.

Content controlled by a term that a user doesn't have access to should get an access denied when they try to visit the edit form.

Additional grad website permissions might need to be adjusted based on this new functionality (e.g. book permissions)


joewhitsitt:
Continuing https://github.com/uiowa/uiowa/issues/1905#issuecomment-733745835, I was hoping to get @pyrello's opinion regarding the additional role before marking the PR as reviewable.

pyrello:
In the case of the Grad College, anyone with the webmaster role would be above the editorial group restrictions. Or at least, I can't come up with a reason why they would need to be subject to them.

I'd be curious if @briand44 has an opinion, since he is the other person I know who has articulated a need for this one some sites.

joewhitsitt:
Since Brian is out, to move this forward and if that is the case for Grad right now, I will remove the role and map it to webmaster. I think #1905 can be kept open as we discuss the future of permissions for default.

pyrello:
Testing this out.

pyrello:
My initial impressions of using this are that:
1. It would work for the described functionality.
2. It has some aspects of it that are undesirable, such as being able to select the whole "Editorial group" to add someone to, instead of just the terms/groups themselves. Also, although it was used with roles in the past, being able to add people directly to the groups makes that unappealing because it would just result in a bunch of extra config to have to maintain. And therefore, having the option to add roles is potentially not desirable.
3. Using this made me wonder if I even need this to be an access control feature. I don't think there is a problem with people modifying or deleting one another's content. Rather, I just need a way to present a certain subset of the content to the user when they log in so that they can focus on the content that they are responsible for.

I'll revisit this after the break.

Thanks for your work on this, @joewhitsitt!

pyrello:
I'm fixing the conflicts, FYI.

pyrello:
Getting #2623 merged will make resolving the conflicts in this a little more straightforward.

joewhitsitt:
Moved back to draft until we decide what to do with this. This is all grad specific, so if it isn't needed we can close this PR.

pyrello:
So, after some consideration, I think that we are not going to use this functionality in our D8 site. If we need this down the road, I think it will actually be good for us to consider it from a fresh perspective. Thanks for the work on this and sorry that it didn't end up getting used.

PR #2615, opened 2020-11-25
Do not generate/update the yarn lock file via blt frontend.

richardbporter:
PR #2521 introduced devDependencies to two Yarn workspaces. That was required for running `watch` commands locally. Because `blt frontend` will not install devDependencies, the yarn.lock file will be modified after running it. That is the default behavior of `yarn install`.

This PR adds the `--frozen-lockfile` option to the `blt frontend` command alias which will prevent `yarn install` from syncing the yarn.lock file and fail if it is required. https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile

The `blt frontend` command alias is essentially production mode, i.e. replicate exactly what is installed for prod. To get devDependencies installed locally, use `yarn install`.

### Testing
- On main, run `blt frontend`.
  - See yarn.lock modification. Discard.
- On this branch, run `blt frontend`.
  - See no yarn.lock modification.

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -46,7 +46,7 @@ validate:
 command-hooks:
   frontend-reqs:
     dir: ${docroot}
-    command: 'yarn install --production --ignore-optional --non-interactive'
+    command: 'yarn install --production --ignore-optional --frozen-lockfile --non-interactive'
   frontend-assets:
     dir: ${docroot}
     command: 'yarn workspaces run build'


richardbporter:
If this is confusing behavior, we could add an environment variable to Travis to control production mode. I think we could remove `--production` and leave the `--frozen-lockfile` option in place. Then `blt frontend` would install devDependencies locally. There would need to be a workaround locally if you wanted production mode - probably just a manual export statement beforehand.

pyrello:
This seems like better default behavior for most situations.

What happens if I am running a `watch` command and have run a development install and I don't realize it and commit that lock file. Will this prevent `blt frontend` from overwriting that lock file, even though that would be the desired behavior?

richardbporter:
> What happens if I am running a watch command and have run a development install and I don't realize it and commit that lock file

I think the lock file will always have devDependencies written to it but It seems to track them as optional somehow - see the failing main build log: https://travis-ci.com/github/uiowa/uiowa/builds/203669419#L4324 So its not problematic that they are written to it. They are just not installed when the `--production` option is used. 

> Will this prevent blt frontend from overwriting that lock file, even though that would be the desired behavior?

Yes but I don't think that _is_ desired behavior based on what I wrote above.

I'm thinking https://github.com/uiowa/uiowa/pull/2615#issuecomment-733887457 might be even better. Then developers can just run `blt frontend` and things just work locally.

pyrello:
I think we try it out for a little bit the way you are recommending and if that seems to cause trouble then we switch it.

pyrello:
Pulling for testing.

pyrello:
I think the argument against https://github.com/uiowa/uiowa/pull/2615#issuecomment-733887457 is that we don't always need dev dependencies locally. In fact, most of the time, we don't. Really only when we are developing for one of the yarn workspaces. If we are not, then I think we want conditions to be as close to production as we can get them. I don't know that it ultimately matters or not, but I think running the `yarn workspace ... --development` command when you need to do that work is reasonable.

pyrello (approved):
Works as advertised. Thanks!

richardbporter:
Thanks. This should get main to pass and deploy.

PR #2617, opened 2020-11-29
Added potential Google Analytics 4 fix

### LINKED ISSUE #2614: Google Analytics 4 Release
# Problem

Google Analytics released a new version which now uses a new property ID, beginning with a "G" instead of "UA-". We need to find a way to incorporate change since affecting D8 analytics currently

# Proposed solution

Apply patch that allows updated GA v4 prefix:
https://www.drupal.org/project/google_analytics/issues/3178179

dnkemper:
Closes #2614 

# How to test
Run composer install; composer update drupal/google_analytics; install site and see changes made there. Not sure how to test actual analytics without a life site to reference


joewhitsitt:
Patch applied locally to GA 3.1.0. So I created a test tag in our GA to use and implemented it for default.local.drupal.uiowa.edu with the Google Analytics Override (only PROD) disabled and was successful with pinging the GA mothership. No issues with entering the ID either into the form.

I suggest we get this out there. I will approve if you think this PR is ready.

Thanks @dnkemper 
Screen Shot 2020-11-30 at 10 09 58 AM

dnkemper:
Thank you so much, @joewhitsitt

joewhitsitt (approved):
tested. let's run with it

PR #2624, opened 2020-11-30
i2248 - Removed unnecessary padding with narrow container

### LINKED ISSUE #2248: Narrow container adding additional (unneeded) padding on sides for non-mobile breakpoint
So this was the observation I had about the narrow container, and why it looked "too" narrow. Digging through the styling, it looks like the narrow region is adding a padding-left:1.875rem and padding-right:1.875rem so that the text doesnt run against the side of the container (`<div class="layout__spacing_container">`), even when it doesnt need to (for non-mobile widths):
```
@media (min-width: 992px)
constructed stylesheet
.content__container .page__container--narrow .layout__spacing_container {
    max-width: 63.75em;
    margin: 0 auto;
    padding-right: 1.875rem;
    padding-left: 1.875rem;
}
```
The normal section and width above this region does not have this left and right padding (as an example).

Screen Shot 2020-09-25 at 9 36 18 AM

dnkemper:
Closes #2248 

# How to test
1. Install default site
2. blt frontend
3. Add page w/ section using narrow layout styling
4. Add block
5. Save page and inspect
6. Find .layout__spacing-container and make sure side padding is removed

### DIFF
--- a/docroot/themes/custom/uids_base/scss/global.scss
+++ b/docroot/themes/custom/uids_base/scss/global.scss
@@ -192,8 +192,6 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled 
       @include breakpoint(md) {
         max-width: 63.75em;
         margin: 0 auto;
-        padding-right: $lg;
-        padding-left: $lg;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


bspeare:
@dnkemper I added some padding back in for the `md` breakpoint since there is a range within that breakpoint where the container hits the edge.  That padding is removed when it hits `container` breakpoint size.

bspeare (approved):
Tested this locally, thanks!

dnkemper:
Thank you, @bspeare!

PR #2625, opened 2020-12-01
Remove twig_tweak module

pyrello:
Removes the `twig_tweak` module, which is no longer in use.

# How to test

I think it is enough that this passes tests. A search of the config directory results in 0 matches for `twig_tweak`, so it is not reference by the default config or any split.


richardbporter (approved):
Thanks

PR #2626, opened 2020-12-01
Bumped up uids version to 3.1.2

dnkemper:


joewhitsitt:
Isn't this the same issue we had earlier? https://github.com/uiowa/uiowa/pull/2626/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR938

Did the solution get documented?

dnkemper:
Neat. Yes, it is! The reasoning behind the fail in October was because a release was deleted?

dnkemper:
I'll fix it :)

pyrello:
To be clear, this is my bad. I created a new release for UIDS without following the full release process because I was testing out the fix to the build system. It should be resolved now.

joewhitsitt (approved):

dnkemper:
Thank you, @joewhitsitt!

Issue #2309, opened 2020-10-08
Create display options for lists, views and feeds in uids_base


bspeare:
# Problem

All of the v3 views and feed displays currently default to a list view within layout builder.    There is currently not a way to switch a display to a grid listing. 

# Proposed Solution

Create a "Display options" layout builder style that offers the ability to switch a view or feed to display as:

- Grid centered
- Grid left-aligned
- Grid 3-col
- Grid 2-col
- Grid 30-40-30?
- List style centered
- List style left-aligned

Initially added some thoughts on this here https://github.com/uiowa/uiowa/pull/2296#issuecomment-705612534.

pyrello:
Bringing my comment over from the thread this started on: https://github.com/uiowa/uiowa/pull/2296#issuecomment-705640786

I would consolidate this down to the following:

**Display list as**
* List (default)
* Grid

**Content orientation**
* Left-aligned (default)
* Centered

**Grid columns** (only shown if *Grid* is selected above)
* 3 columns 33-34-33 (default?)
* 3 columns 30-40-30?
* 2 columns

joewhitsitt:
Could Accordion be another display option? Might relate to https://github.com/uiowa/uiowa/issues/2486

joewhitsitt:
With the card styles I was doing in #2644, I am wondering if the content orientation is needed here? Could this just be adding the grid options? Also, do these grid styles already exist in UIDS or is that separate work that needs to be captured/figured out

bspeare:
@joewhitsitt Yes, I think that part is no longer needed.

PR #2644, opened 2020-12-02
Aggregator card style options

### LINKED ISSUE #2514: Allow card options to be applied to aggregator feed
# Story

As a site owner I want to be able to style my aggregator feed block with all the options that are available to cards so that I can have control over consistency in my design.

Related to #2351 

# Proposed solution

* For each card LB style option that is available, see if it can be reused for the aggregator feed block. If not, create a new LB style for lists.
* Add the aggregator block type to each LB style identified in the previous step.
* Export config updates.
* Add preprocessing logic for the aggregator block that takes in the LB styles and passes them through to the individual items being displayed in the feed.
* There is currently logic for preprocessing the card block that will need to be adapted and utilized to pass through styles and image styles to elements insides the feed item.
* There are examples of logic processing a parent item to pass variables through to a child item to make the headlines field function.

joewhitsitt:
Resolves #2514 

Piggy backs on some tricks that card and headline implemented. Most of this should be similar for the other listings except for the image handling.

# Questions

- Should BG Color be passed to children? - _Yes, and not applied to the parent_
- Can other aggregator styles be cleaned up? No idea what the _row stuff does. - _Yes_
- Is my copy/pasta from card for the parent headline padding okay? - _No, since we want section controlling bg, this is not needed_
- **Not sure the caching solution for the feed image is smart.**
- Should the defaults match what is currently out in PROD (this would differ from the card default)? - _Setting to what is currently out there similar to how card set white as default_

# How to test

- Pull a v3 site, cim, blt frontend, configure Aggregator to pull in a feed. Place the RSS Feed block and test out the layout builder style options now available.
- New instances should match the styles here: https://sandbox.prod.drupal.uiowa.edu/aggregator-0

Example feed: http://blog.admissions.uiowa.edu/andre/feed/

Screen Shot 2020-12-02 at 1 40 42 PM


bspeare:
@joewhitsitt  This looks great! 

Should we add the border option?

Screen Shot 2020-12-03 at 3 52 55 PM

We need to adjust the headline padding if there is no background on the aggregator and a section background is set:

Screen Shot 2020-12-03 at 3 59 21 PM

I think we might want to only have the background color on each aggregator item only and not on the container?  That way it functions closer to a card and when there is a `grid` setting for this list each grid item will be visible.  What do you think?

Screen Shot 2020-12-03 at 4 03 35 PM

bspeare:
- The styles listed on https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/content/aggregator-feed.scss are for the top part of http://admissions.local.drupal.uiowa.edu/aggregator/sources/1.  So we still need them. 
- We can get rid of https://github.com/uiowa/uiowa/blob/i2514-aggregator-card-options/docroot/themes/custom/uids_base/templates/content/aggregator-item.html.twig#L24. 
- Are we able to set default card classes for the list view on http://admissions.local.drupal.uiowa.edu/aggregator/sources to be something like the screenshot below?
Screen Shot 2020-12-03 at 4 20 07 PM


Do we want the defaults for the list styles to be the same as the card?  Stacked, widescreen, etc?  The answer might depend on what the default is when we build: https://github.com/uiowa/uiowa/issues/2309.

joewhitsitt:
Thanks for the feedback @bspeare. You avoided "going rogue" by 6 minutes!

> Should we add the border option?

Sure

>We need to adjust the headline padding if there is no background on the aggregator and a section background is set

I think I understand

>We can get rid of https://github.com/uiowa/uiowa/blob/i2514-aggregator-card-options/docroot/themes/custom/uids_base/templates/content/aggregator-item.html.twig#L24.

Uh, is generic card styles setup to handle the styles here? item-description is aggregator specific.
https://github.com/uiowa/uiowa/pull/2644/files#diff-67d117327c136a66e968a5b461a1d8478e0940c1f1315bfb594816ea1a64b2afR27

>Are we able to set default card classes for the list view on http://admissions.local.drupal.uiowa.edu/aggregator/sources to be something like the screenshot below?

Are you suggesting default config with site override capabilities here or do we just need to move aggregator to a separate case statement and set specific defaults: https://github.com/uiowa/uiowa/blob/5d8028d66d30c5fdbf2d0e49645ecff16b9eda24/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L149

> Do we want the defaults for the list styles to be the same as the card? Stacked, widescreen, etc? The answer might depend on what the default is when we build: #2309.

I was thinking they would match what is already out there for defaults so we don't pull a switcheroo on people.

bspeare:
> I was thinking they would match what is already out there for defaults so we don't pull a switcheroo on people.

Good point!

bspeare:
> Uh, is generic card styles setup to handle the styles here? item-description is aggregator specific.
> https://github.com/uiowa/uiowa/pull/2644/files#diff-67d117327c136a66e968a5b461a1d8478e0940c1f1315bfb594816ea1a64b2afR27

@joewhitsitt  Can you elaborate a little more on what you mean this one?

joewhitsitt:
Sure @bspeare To use the "hide descriptions" option on aggregator, the class name is different "item-description" so i scoped it using the aggregator-item class.

Screen Shot 2020-12-04 at 9 39 29 AM

bspeare:
@joewhitsitt Thanks, that makes sense.  We can either leave that wrapper class on or scope it to `.card` if that helps making the markup more maintainable.

joewhitsitt:
okay, i can move this to card

joewhitsitt:
The UX for this is a bit confusing if you have been working with the other blocks. (e.g. background controls children instead of the block itself unlike the rest of the blocks). Help text could help, but I bet most folks will figure it out via preview instead.

bspeare:
@joewhitsitt I agree that the UX could be a bit confusing, what do you think the help text should say?

bspeare:
@joewhitsitt The updates look good! I did some testing and found some card bugs that I put some fixes in for here: https://github.com/uiowa/uids/pull/428.

joewhitsitt:
> what do you think the help text should say?

"This is not the background color setting you are looking for."

not the background you are looking for

i say we document and see what the training sessions reveal

dnkemper (approved):

joewhitsitt:
We need a UIDS release to go out with this right @bspeare ?

pyrello:
I don't see an update to `package.json` or `yarn.lock`, so it seems not

joewhitsitt:
uiowa/uids#428 was merged in to fix some style issues that will be introduced with this

bspeare:
I think we can just merge this without that padding change and wait for those banner video prs to merge.

PR #2645, opened 2020-12-02
Attempt to alleviate crashing issues.

### LINKED ISSUE #2611: As a developer, I can use a modern version of Node.
# Story

As a developer I want to use a modern version of Node so that I can benefit from the active development of it.

# Proposed solution

We're currently using Node 10 which is only getting maintenance fixes until April 2021. We should update the version of Node used in CI and DrupalVM to the latest LTS. 

https://nodejs.org/en/about/releases/

richardbporter:
Resolves #2611 

This is an attempt to see if reducing large file operations on the VM can help alleviate computer crashing issues and general performance. It mostly moves frontend ops to your host. This has the added benefit of being faster and working with Babel. I upgraded to Yarn 2 for improved performance, at least according to https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern.

I was not able to get npm@7 to run workspace scripts how we are using them. That could be an option further down the road and then drop Yarn entirely. 

This PR also disables the `blt tests:yaml:lint:all` command which was crashing for me also. It seems unnecessary since we install from config in CI which I'm assuming will fail if any YAML is invalid.

Unfortunately, this probably requires removing `vm.enable` from your `local.blt.yml` file to avoid blt suggesting you run frontend on the VM. This also makes it more confusing where to run certain commands.

### Testing
- Remove `vm.enable` from `local.blt.yml` file if set.
- Reload VM config:
  - `vagrant reload --provision`
- Uninstall yarn on the VM:  
  - `vagrant ssh`
  - `npm uninstall -g yarn`
  - `exit`
- Follow README docs on frontend requirements installation: https://github.com/uiowa/uiowa/compare/no-yarn-plus-node-update?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R24-R26
- Build frontend
  - `blt frontend`
- Watch the uids_base workspace and make file changes.
  - `yarn workspace uids_base watch`
- Watch the uiowa_alerts workspace and make file changes.
  - `yarn workspace uiowa_alerts watch`
  - ES6 file changes should work as well. They did not previously on VM.
- Sync the following sites and check the themes:
  - default
  - hr.uiowa.edu
  - uiowa.edu
  - admissions.uiowa.edu
  - ighn.international.uiowa.edu


richardbporter:
I'm seeing crashes hitting sites in my browser now. This might alleviate CLI crashing but I wonder if we've reached a critical mass of site code on the filesystem.

richardbporter:
I'm not sure this really fixes anything but I would still appreciate feedback/thoughts.

richardbporter:
Closing this. We can resurrect later if needed.

PR #2668, opened 2020-12-07
Specify specific version of memcache to install in CI.

richardbporter:
PECL memcache v8 is out which requires PHP 8. Requiring a specific version should get CI running which is using PHP 7.3.


pyrello (approved):
Reviewed, not tested.

richardbporter:
We need this because drupal/memcache depends on it. Long term, we could modify our environment splits to not enable it in CI.

PR #2671, opened 2020-12-07
update rabbit hole

joewhitsitt:
Replacement PR for updating rabbit_hole, but with update hook config changes applied.
https://github.com/uiowa/uiowa/pull/2660


richardbporter (approved):
Only enabled in HR. Reviewed but not tested.

Issue #2058, opened 2020-08-20
Develop alternative layout/style options for articles/people


joewhitsitt:
## Problem
https://github.com/uiowa/uiowa/issues/1789 only went as far as getting the articles/people block functionality into layout builder as a list.

## Proposed solution

Decide how to provide other layout options within layout builder for this type of content.

- css class based styles?
- view mode switching?
- https://www.drupal.org/project/views_layout looked the most promising in a way to manipulate the view with the layout builder api.

bspeare:
@joewhitsitt can we close in favor of https://github.com/uiowa/uiowa/issues/4453?  Seems like we have a lot of issues about the article layout.

joewhitsitt:
I think this was related to providing different style/layout options for the v3 views blocks as it came from https://github.com/uiowa/uiowa/issues/1789 and then got later confused with the article/person node full view mode stuff.

this is done more or less.

Issue #2670, opened 2020-12-07
As a site owner, I can add/remove margins on components used in the same layout column


bspeare:
# Story

As a site owner, I want add/remove margins on components used in the same layout column so that I can clearly view components with and without spacing in a page design. 

All components do not have margins that will separate them if used in the same layout column.  See https://github.com/uiowa/uiowa/issues/2554 as an example. 

# Proposed solution

## First slice (present at mid-sprint demo)
- [ ] In `global.scss` add a global rule for components like `.block` within the same layout column like:
- [ ] Add the `.block` class to all custom block components that don't have it like the `card` and `stat`.
```
// Set spacing as the default for mobile. 
.layout__region .block {
  & + .block {
    margin-bottom: 1.5rem;
    @include breakpoint(sm) {
      margin-bottom: 0;
    }
    &:last-child {
      margin-bottom: 0;
    }
  }
// Add modifier class through layout builder styles to apply the spacing above the `sm` breakpoint. 
.layout--margin .layout__region .block {
  & + .block {
    @include breakpoint(sm) {
      margin-bottom: 1.5rem;
    }
    &:last-child {
      @include breakpoint(sm) {
        margin-bottom: 0;
      }
    }
  }

```

# Second slice (decide at mid-sprint demo whether to continue this work)
- [ ] Add a layout builder style checkbox like `Remove default component spacing` to unset the global rule. 
   - Update hook to check this box for existing pages
   - New pages have it unchecked
- [ ] Remove or adjust https://github.com/uiowa/uids/blob/main/src/components/card/card.scss#L355?
- [ ] We will also want to address this issue with components creating gaps between sections: https://github.com/uiowa/uiowa/issues/5631.  

This solution will also need to be implemented for setting/unsetting margins on lists in grids when https://github.com/uiowa/uiowa/issues/2309 is built.

bspeare:
Since all components don't have a `block` class on them that we can target, do we need to add a universal class on components so that we can target them with this adjacent sibling selector rule?

bspeare:
I wonder if the proposed solution should be to add the "Spacing" layout builder styles to all blocks and set the default value to `-None-` and allow site owners to "Add extra top spacing" and "Add extra bottom spacing" to each layout builder block.  

<img width="623" alt="Screen Shot 2021-06-03 at 3 36 13 PM" src="https://user-images.githubusercontent.com/1036433/120708661-7a808380-c481-11eb-84e6-82d04a0f0178.png">

[...]

Issue #2554, opened 2020-11-17
Stacked cards need a bottom margin


dnkemper:
# Problem

When you stack two cards in a column, there isn't a clear separation between the two separate cards
<img width="416" alt="Screen Shot 2020-11-17 at 10 52 03 AM" src="https://user-images.githubusercontent.com/16872949/99420588-01458880-28c3-11eb-8cff-46c3c482a001.png">

# Steps to reproduce

Add a card to a column and then add another card below it - boom

# Proposed solution

Going to add a bottom margin to create some space between them

bspeare:
This is a good idea.   I wonder if we should add a container setting to add/remove margins between blocks?  We could add a default margin for all stacked components within a layout and then create a setting on the container to `remove default component spacing` or something like that.

dnkemper:
@bspeare you are a styling wizard! Thank you!

bspeare:
Closing this in favor of https://github.com/uiowa/uiowa/issues/2554 which is already pointed.

PR #2676, opened 2020-12-07
Block List Grid Options

### LINKED ISSUE #2309: Create display options for lists, views and feeds in uids_base

joewhitsitt:
Resolves #2309
Relates to https://github.com/uiowa/uiowa/issues/2058

This introduces list/grid options for articles, events, people, rss feed. Editors can choose list or grid. If grid, they can choose 2 column or 3 column.

Depending on where the block is placed, the grid will work. Of course it will work in a single column desktop layout, but as the viewport changes and the columns become narrower the grid falls back to a list display similar to how individual cards stack currently.

# To Test

`blt frontend` and sync a v3 site (e.g. sandbox.uiowa.edu).
- New/existing blocks should have a list default, and if grid is selected, 3 column should be the default selection.
- If list is selected, the grid classes should not apply to the block render (e.g. grid--threecol--33-34-33).
- Try the different block options in different layouts with different percentage columns to make sure it is responsive enough that you don't cringe.


bspeare:
@joewhitsitt This looks good! 

- For the list and grid options I think we need the option to `add default spacing` or `remove default spacing` in the styles option.  

Screen Shot 2020-12-08 at 3 51 48 PM

- Do we want to create a `grid.scss` component in `uids_base` to put all this work?  It would be nice to set up the grid styles so that the classes could be reused elsewhere.   Are we able to use css grid or do we need the flexibility to expand/fill a container if there is only one blog post and the site owner has selected `Grid` and `3 columns`?

joewhitsitt:
@bspeare thanks for the early review. No changes to the config names or css classes? I know there is a pattern that is ideal, but not sure if I am following it correctly. definitely got confused at first with the card--list class i saw in the DOM. Also didn't know if there were already classes to use in uids or if this is all new. grid.scss in uids_base sounds great to me. i'll defer to you on grid/flex. i see there is a lot of flex already for section columns, etc.

with the add/remove default spacing styles, i assume similar validation will have to be done like with card border in certain layouts since i can't make that option show/hide based on grid being selected.

pyrello:
Any reason we can't use our existing layout classes? It is a little messy since we'll have a hierarchy like `.layout--onecol` > `.layout--threecol--50-25-25`, but I don't imagine that this will create a ton of conflicts and we can massage those out.

joewhitsitt:
> Any reason we can't use our existing layout classes

Or should it be a utility that both utilize?

pyrello:
I have wanted to add some version of our grid/columns to UIDS for a bit now because it would be useful in building out examples. Is that what you mean by both?

joewhitsitt:
still thinking fast in uids_base for now as we do this block stuff in this PR, but yes, build it like a utility with inspiration from what is currently there that other things like the layout columns can eventually use

Update - me looking at the current layout stuff after saying "fast": 😭

pyrello:
I think that is a good plan. I don't think it is necessarily going to be easy or fast, but I like the idea of not creating a bunch of redundant classes.

bspeare:
> with the add/remove default spacing styles, i assume similar validation will have to be done like with card border in certain layouts since i can't make that option show/hide based on grid being selected.

Good point on this one.  I think we would always want margins/column-gap set on the border option.

bspeare:
How do we plan on implementing this for views and aggregator?  Do we just add a class of `layout--twocol--50-50` to the view and set the css to something like:

```
.layout--twocol--50-50 {display: flex;}
.layout--twocol--50-50 > * {display: flex;flex-basis: 100%;flex-wrap: wrap;}
.layout--twocol--50-50  .views-row, 
.layout--twocol--50-50  .aggregator {flex: 0 1 49.5%;}
```

We could definitely use something like this for the student profiles pr view.

joewhitsitt:
Working on a mixin for this stuff so it can be used elsewhere (view pages, etc.), but I think this gets complicated because these list blocks can be placed in any section column (e.g. 4-col). I am beginning to think the same specs for the layout can't be used for this. Adjustments based on where the grid is placed needs to be accounted for.

bspeare:
@joewhitsitt I think it makes sense to keep separate for now and maybe make notes if there's anything that can be shared?

joewhitsitt:
So I have changed the approach a bit and pushed up something that can be reviewed. It would be nice if @pyrello and @bspeare had a chance to review particularly since we had the call yesterday but anyone else is welcome to give thoughts on this as well.

Layout builder currently gives us a lot of options in terms of column widths that these blocks are placed in. Not even the card styles are four-column friendly at certain breakpoints.

This approach bases the grid styles based on the parent. As a mixin it packs a power punch and has a lot of potential but is a small addition to an existing layout definition as well as usable in other situations (view page, etc.). For now, I have only added it to columns that have 50% or more of the page width.

There are so many combinations here (container (e.g. narrow), section columns (e.g. four-column) and all the different card style options and margin/padding controls. Assumptions on all of these things decides whether or not the grid can happen, otherwise it falls back to a listing.

Currently using css grid only because I started playing with it to avoid the full-width orphan common in flexbox. Have not tested browser support but I assume it gracefully falls back to a listing.

Open to other approaches, but I thought this accomplishes the task and doesn't restrict where these existing block types can be placed.

bspeare:
@joewhitsitt this works great!  A few notes:
- Perhaps remove the 30-40-40 option and just have the Grid Column options list as `2 Columns` and `3 Columns`?
- Increase the gap to `gap: 3rem;` to match the margins used in the layouts?  I know that you would have to recalculate the grid width with that change.  
- Add option to remove grid gap?

bspeare:
The card list style with a border gets in the way with the grid settings by adding a margin bottom in addition to the gap.   Should we override that margin (https://github.com/uiowa/uids/blob/main/src/components/card/card.scss#L355) in this pr until we figure out https://github.com/uiowa/uiowa/issues/2670 and https://github.com/uiowa/uiowa/issues/2554? 

Screen Shot 2020-12-16 at 1 15 06 PM

bspeare:
If we are forcing items set to "grid" to stack in different columns, should we reduce the `gap` when they are stacked?

**2 column layout example set to a `2 Columns` grid (`3 Columns` does remove the gap):**

Screen Shot 2020-12-16 at 1 43 06 PM

**3 column example:**

Screen Shot 2020-12-16 at 1 42 24 PM

joewhitsitt:
> If we are forcing items set to "grid" to stack in different columns, should we reduce the gap when they are stacked?

Good question @bspeare. You are testing my knowledge of flexbox/grid here.

The columns stack independently of the viewport to prevent too much squish in certain column widths. For the row gap, I hope breakpoint specific tweaks won't be too noticeable.

By the way, I went this grid route because I was having troubles with the sizing based on parent column width and the full-width orphan issue with flexbox.

pyrello:
Just to capture the state of play with this PR after discussion yesterday:
* We are putting this on hold for the duration of sprint 46 to allow for more conversation about the ramifications of moving forward with this approach.
* Part of the reason this is happening is because another potential approach has been articulated in #2707, which sets grid display options at the layout/region level rather than the block level.
* Finding a migration path between these two approaches is difficult and may not be possible if starting from this PR and eventually migrating to the approach in #2707.

docroot/themes/custom/uids_base/scss/components/grid.scss
```diff
@@ -0,0 +1,51 @@
...
+        .list-wrapper {
+          gap: 3rem;
+          grid-template-columns: 1fr 1fr;
+        }
+      }
+    }
+  }
+  .inner--grid.grid--threecol--33-34-33 {
+    @if $parent >= 75 {
+      @media screen and (min-width: 58.125em) {
```
bspeare: @joewhitsitt Do you think it would help to make some new breakpoint variables for reuse?  Or are these pretty specific to the mixin?

joewhitsitt: i grab this number from the three-col layout sass so that the grid would break down at the same time the three column layout does.

joewhitsitt:
Though a migration path out might be difficult in the future, if necessary, the migration from v2 into this method would be easier.

bspeare:
It looks like we also might need something like this for cards that are within a `views-row` or other wrapper to set/unset equal height: https://github.com/uiowa/uiowa/blob/i2457-student-profiles/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/person.scss#L133-L138.

bspeare:
@joewhitsitt Is it worth updating this pr with the grid mixins from UIDS?

bspeare:
@pyrello are we moving forward with this approach for grids or do you still plan to spend more time investigating https://github.com/uiowa/uiowa/pull/2707?

joewhitsitt:
Until we decide if we want to go this way, I wouldn't bother. Hopefully, we can take some time to decide which way to go because this is desired functionality that we have teased in demo

pyrello:
I'm not in a position to continue working on the trio of PRs that would stand as a possible replacement for this one. My hope was that letting this sit for a minute would allow some time to think through the possible ramifications and to possibly discover whether we are going down a one-way path with this. I have not gotten any additional clarity on it and I don't want to hold up releasing this feature, so I suggest that we go ahead with this. Hopefully, if there are downsides that we are not seeing now, we can identify them quickly once it is in the hands of users and make an adjustment then.

bspeare:
@joewhitsitt I did some slight updates on this before vacation a few weeks ago so I just pushed that work up.

joewhitsitt:
not affected by the pager issue as the classes are not passed to children

bspeare:
@joewhitsitt this is now working as expected.   I added a couple of fixes to get the grid styles working since we added https://github.com/uiowa/uids/blob/main/src/components/grid/grid.scss a while back.

bspeare:
Screen Shot 2021-02-17 at 4 09 21 PM

It is tempting to limit the grid to 50-50 on a 25/75 second region layout.   Or force a `stacked` option if placed in this section. 

Screen Shot 2021-02-17 at 4 13 52 PM

joewhitsitt:
I think I tried to mirror the same behavior that individual cards (aligned image) would do if placed in the various layouts. It was uncomfortable to watch while scaling down the viewport. It has been so long, so I defer to whatever you want to dial in.

bspeare:
@joewhitsitt I think what you originally built is fine.  It also seems tricky to override that based on how the mixin is structured: https://github.com/uiowa/uiowa/blob/4e9d1817785f69e9988499e32a257c80a4f8e46e/docroot/themes/custom/uids_base/scss/layouts/twocol.scss#L55.

GaryRidgway (approved):
I have looked at many variations for many of the different options here and it seems like the breakpoints and fallbacks at thinner widths are working for all that I tested. I am sure there are some combinations i did not get to, but i tested many. I think this looks good to me!

pyrello:
Pulling down to test.

pyrello:
Question: Why does the pager issue not seem to affect this PR when it is a problem in #2730?

pyrello:
Okay, I think I understand why this is. Because the LB styles are applied directly to the block rather than passed to the individual view rows.

pyrello (approved):
I have reviewed the code and tested the basics of the functionality. I'm good with moving ahead with this.

PR #2682, opened 2020-12-08
Update Node, Composer and document Yarn workspaces.

### LINKED ISSUE #2611: As a developer, I can use a modern version of Node.

richardbporter:
Resolves #2611 

Taking some learning from debugging my crashes and updating our Node environment to version 14. This also ensures Composer 2 is installed on the VM and documents how to run watch commands on your host. ES6 watch never worked on the VM and SASS was really slow because of polling.

I did not use `lts/*` as the NVM version like we do in uids because its possible CI, the VM and the host could get out of sync. 

### Testing
- `vagrant provision`
- `vagrant ssh`
- `composer --version`
- `node --version`
- `blt frontend`
- Check theme and workspaces built correctly. Visually test theme/workspace assets.
  - uiowa_bootstrap updated to Gulp 4
  - Check HR.
- `yarn workspace uids_base run watch`
  - Change SASS files or something.
- Optional: follow README instructions for running watch commands on your host.


joewhitsitt:
pulling down to test

joewhitsitt (approved):
based on test steps, this worked smoothly. hr looks okay. watch worked

joewhitsitt:
FYI for other testers: not compatible with quick PR switching

richardbporter:
Sorry for not specifying. Labeled with 'config update' as it modifies your VM config and some of that cannot be downgraded when switching branches.

pyrello:
Do you want other testers on this?

richardbporter:
I don't think so. Its probably best if everyone runs `vagrant provision` after this is merged.

PR #2666, opened 2020-12-07
Grad College content type updates

pyrello:
# Updates
## Grad College split
* Added Thesis Defense content type.
* Added the smart date module to handle date field for Thesis Defense.
* Added logic to hide end date and duration fields via `'#access' => FALSE` and logic to correctly set the end date in a custom form submit function.
* Added grad department field to Person content type and scoped it to show on Mentor person type.

# Default platform
* Resolves #2757: Added scoping for `.field_item` `font-size` adjustment to limit it to `div` and `p` tags.

# How to test
* `blt ds --site=grad.uiowa.edu`
* Check to see that Thesis Defense content type exists.
* Create a Thesis Defense item and view it.
* Observe that the Title field displays at the expected size for an `<h2>`.
* Visit the node/add/person page and select Mentor for the person type and observe that the Department field shows up.

PR #2697, opened 2020-12-11
Feature: Department entity config

### LINKED ISSUE #2622: Add a program config entity
# Motivation

There seems to be a use case for a reusable program list similar to what is being created in #2562. This is currently handled in the `person_extended` split via a vocabulary and term reference field. But this approach depends on users entering the departments manually, which introduces the potential for discrepancies between sites.

# Proposed solution

Create a custom config entity called Program and populate entries for it for all the established programs and subprograms.

pyrello:
Resolves #2622 

# How to test

Test instructions coming soon!


pyrello:
The list I need to use for department fields in the grad college site doesn't nicely line up to what I was able to pull from MAUI and I'm not sure if it can be used interchangeably for broader applications, so I'm closing this PR and I will make the updates I need in #2666.

PR #2699, opened 2020-12-11
Do not allow configuring page header section.

### LINKED ISSUE #2694: Regression? Webmasters are able to configure the Header section within a layout
# Problem

While we have locked out editors/webmasters from adding blocks to the "Header" section, they are able to edit the section and add background colors and container updates. 

Screen Shot 2020-12-10 at 12 10 03 PM

Screen Shot 2020-12-10 at 11 59 03 AM

# Steps to reproduce
Masquerade as a webmaster and edit a page with a featured image in layout builder and try to edit the "Header" section. 

# Proposed solution
Update section permissions

richardbporter:
Resolves #2694

For non-admins, this removes the ability to configure the header section if it exists. This goes for new nodes, default content going forward and existing default content out there that didn't have this lock in place.

**The Gotchas:**
- Existing sites that customized these regions will need an Admin to fix them.
- Any site thinking they should be able to edit these, will need to be split.

# To Test

As a webmaster of a v3 website, you should not be able to modify header section configuration for

- New pages
- about page of a fresh install
- node/1 if the defaults are reset and a header is provided
- existing node/6 (about) pages
- make sure v2 sites don't break as a result of the update hook
- same goes for admissions


config/default/core.entity_view_display.node.page.default.yml
```diff
@@ -59,8 +59,8 @@ third_party_settings:
         layout_settings:
           label: Header
           layout_builder_styles_style:
-            section_margin_edge_to_edge: section_margin_edge_to_edge
-            0: ''
+            - ''
+            - section_margin_edge_to_edge
```
richardbporter, line 63: I'm guessing this previously manually modified and `drush config:export` reformatted it.

richardbporter:
> node/1 if the defaults are reset and a header is provided

I don't understand this. Could you explain a bit? Thanks.

joewhitsitt:
> > node/1 if the defaults are reset and a header is provided
> 
> I don't understand this. Could you explain a bit? Thanks.

node/1 (Home) default content does not have the header section. in layout builder you can "reset to defaults" (deletes content btw) and restores the default template with the header section. If it is restored, there should be a header section and it should be locked down similar to new nodes.

richardbporter:
Thanks. I don't see any issue with this PR unless there are sites out there which should allow webmasters to edit the header. IDK if such sites exists and if they did how we would even identity them.

 I can't approve this PR since I created it.

joewhitsitt:
Any sites want to allow webmasters to do this should split this config.

Should this PR update the following sites?:

- ~~config/admissions.uiowa.edu/core.entity_view_display.node.page.default.yml~~ _covered_
- ~~config/coronavirus.uiowa.edu/core.entity_view_display.node.page.default.yml~~ - _doesn't have lock restrictions?_

richardbporter:
Good catch. I don't know.

joewhitsitt:
Updated coronavirus a bit. Webmasters can't change section configuration on new/existing.

richardbporter:
It'd be nice to get another @uiowa/developer to review this.

cory-skeers:
I tested that it worked on a fresh install earlier--pulling it back down to test in other contexts.

cory-skeers (approved):
Works as advertised. The gotchas are there, but can be addressed if needed following this.

Tested on fresh install, existing V3, plain V2 (protostudios), Admissions, and Coronavirus with no errors, and for the latter two, creating a new page or reverting an existing page to default (restoring the header) worked as well.

Issue #2351, opened 2020-10-15
Allow all card styles to be added to remaining list type blocks


pyrello:
# Story

As a webmaster, I want the ability to add recent card style options to my list blocks and single event blocks.

# Proposed Solution

This was recently done in https://github.com/uiowa/uiowa/pull/2644 and https://github.com/uiowa/uiowa/pull/2683 that set a foundation for this, but that must be extended to the other types as well as adapted based on how they are rendered (view, external events data, etc.).

- [x] Update the card styles configuration to extend to the other block types (articles, people, events, event)
- [x] Update layout_builder_custom to have the same show/hide options functionality.
- [x] Set the defaults for each block type by what is desired going forward.
  - [x] RSS and Article  = white, left align, images are right align wide small
  - [x] Event, Events, Person = white, left align, images are left align small circle
- [x] Write at least one update hook to match the defaults styles already out there in the wild so that the rollout is transparent for existing blocks.
- [x] Make sure the image view mode switching and caching is working based on the render of the current block.
- [x] Update block templates to remove hardcoded classes and instead take classes set by the block settings.
- [ ] Remove, update any uids_base specific styles.
- [ ] Check against v2 uids_base sites

A lot of this work can be referenced in those two previous PRs.

<details>
<summary>Original issue summary</summary>
# Story

As a site owner I want to be able to apply all the card style options to list blocks that utilize the card template. 

Related to  

# Proposed solution

After #2514 is completed, add these options to the remaining list blocks: articles, events, people. Additional consideration should be given to whether there is any possibility of this work affect V2 sites and whether testing will need to be done.
</details>

pyrello:
Brought over from #2306

bspeare:
We need this option for:

- Aggregator
- Events
- Event
- People
- Articles 

Are we able to use the existing layout builder styles on the custom block views listed above?  Or will it have to be recreated for each list type block?  Is there a list type not listed above that we are missing?

<img width="464" alt="Screen Shot 2020-11-06 at 9 27 10 AM" src="https://user-images.githubusercontent.com/1036433/98383656-6622f800-2012-11eb-8ddb-2c5549d585d6.png">

bspeare:
This functionality is needed on the admissions site so that they can adjust the display configuration of their blog feeds, counselors, and student profiles.

joewhitsitt:
@bspeare do we want to take this opportunity to change the default card styles for new blocks of each type or keep them the same as existing blocks out there?

- event
[...]

PR #2702, opened 2020-12-13
Views block display block options for building lists

### LINKED ISSUE #3115: Views block pager issue
# Problem
As identified in #2730, there is a problem when passing block styles to the view and clicking on a pager link in a views block. On the second page of results, any styles that were passed through to the view are lost. I'm putting this separate issue together to make it easier to see the relevant information to try to resolve this.

# Steps to reproduce
The easiest way to see this issue is to test #2730 locally.

# Proposed solution
The best ideas we have so far about why this is happening are:
1. We are failing to invalidate a cache or properly set a cache when adding LB styles to the view.
2. The method we are using to set the styles - using a custom `$view->display_hander->setOption()` only gets called when the block is rendered as part of the normal page rendering process. When called via AJAX the section in question doesn't get triggered.

Some research has uncovered this issue: https://www.drupal.org/project/drupal/issues/2605218, which has a couple patches that have been tried on #2730 to no avail.

pyrello:
Relates to https://github.com/uiowa/uiowa/issues/2351
Relates to https://github.com/uiowa/uiowa/issues/2058

# How to test

* `composer install`
* `blt ds --site=sandbox.uiowa.edu`
* `drush @sandbox.local uli`

## Test view block settings
* Click the "Layout" tab on a page and add one of the People view blocks by clicking "More...." and selecting it under 'Lists.'
* Test hiding and un-hiding fields via the "Hide fields" accordion.
* Test changing the sort options via the "Sort options" accordion.
* Test changing exposed filter settings via the "Exposed filters" accordion.

## Test building a view
_These instructions are in progress._

# Updates
Resolves #3115
 
This provides additional options that can be added view block settings when adding a block via Layout Builder. This provides a framework for allowing users to directly place view blocks instead of having to create custom block content types and adding code to render the view.

I have made some modifications to default config in this PR to make it easier to test:
* Added a People view block display with all allow settings turned on and the category set to 'Lists.'
* Updated the LB restriction settings for the page to allow all blocks of type 'Lists.'

## New Views block options
* [x] **Hide fields** - Toggle to allow showing options to the user to allow hiding fields when placing the block.
* [x] **Configure sorts** - Toggle to allow showing sortable list of sort options on the block settings.
* [x] **Items to display** - set as a number instead of a select list.
* [x] **Offset** - set the number of records to skip. This is useful when placing multiple Views blocks of the same kind.
* [x] **Show pager** - whether or not to show the pager.
* [x] **Display more link** - Choose whether to have a display more link and use a LinkIt field to enter the URL.
* [x] Display more link help text - Set help text that displays underneath the "Display more link" field.
* [x] Set above field to allow HTML.
* [x] **Add exposed filters to block settings** - Allows the developer to select exposed filters to display in the block settings.
* [x] URL field to set default path, which can accept tokens.
* [x] Add a token for people page view.

## Block settings and block display
* [x] Exposed filter fields - Set exposed filter values via block settings. Filter values are passed to block display.
* [x] Hide fields table - Select fields to hide from display.
* [x] Configure sorts - Display table of sort handlers and allow the user to change their order and set them to ascending or descending.
* [x] Items to display - Should work similar to existing people/articles blocks.
* [x] Pager offset - Should allow user to start at result X.
* [x] Show pager - Checkbox that displays pager or not on block display
* [x] Display More link field - Link field to select internal path or set arbitrary one.
* [x] Style "Display more link" autocomplete results to look like regular autocomplete results.
* [x] Display More link help text - Show the help text set in the view display settings.
* [x] ~~Add More link validation.~~ - It seems like the validation that is currently set works pretty well, but we may want to test this a bit more.

## Other
* [x] Add headline field.
* [x] Style "People list block" view.


pyrello:
An additional piece of information that is potentially relevant to how this would be implemented:

When creating a block display for a view, you can set a category for the block. The default is "Lists (Views)". In Layout Builder, the block category is what is used for creating the various groups of restrictions for what blocks are shown. If we created our own category - such as "Lists" - and applied it to views block displays that are specifically intended to be placed directly via Layout Builder, we could create a permissive restriction rule ("Allow all current and future blocks") for that category in Layout Builder Restrictions. This would allow us to maintain a default deny list of any views that currently exist or might be added in the default category while creating a curated list of allowed views without having to update the LBR settings every time.

joewhitsitt:
taking a peek

**update**: might be too early. i am getting an illegal choice error in trying to place the people block you just added. not sure if my paging hack is causing problems. I like the idea behind this if we don't take too many steps back away from what we have created so far (layout builder styles, headline, etc.)

pyrello:
@joewhitsitt Where are you seeing that error? In the console? If there are any more details you can provide, that would be great.

pyrello:
I hadn't thought about the headline field, but I think that we could provide the same option through the block interface and apply the logic behind the scenes.

joewhitsitt:
it was on the people block view render in layout builder. no results and an illegal choice on the exposed filters. again, it might be the way i hacked the people block together.

pyrello:
I'll check it out. Thanks for taking a look.

pyrello:
I'm stuck on 2 things right now:
* [x] I would like to render a LinkIt-enabled field for adding a "More link" but can't seem to figure out the syntax for this.
* [x] I would like to add a setting to views display that lets the help text for the more link field.

pyrello:
There is a view block setting to allow configuring the sorts. When it is enabled, it creates a tabledrag list of sorts that can be re-ordered and allows you to set the direction (e.g. ascending) of the sort.

Screen Shot 2021-02-17 at 3 00 12 PM

I am wondering if we want to allow allow the user to remove the sort? Or should all sorts have to be used?

mark-bennett-uiowa:
@pyrello Sean, this ability would be amazing! This would remove the need to manually sort long lists of people for me on most sites. Im sure there would be a few I couldnt avoid, but still very helpful to sort sticky differently than first and last name.

joewhitsitt:
>This would remove the need to manually sort long lists of people for me on most sites.

How would this remove the need to manually sort people?

pyrello:
I think there may be a bit of misunderstanding here. The option above would just allow you to prioritize a sort differently, but the sticky option would still depend on the sorting from the People page.

pyrello:
Ran into an issue while reading through the "display more link" help text on the existing People block: With the existing block, you can choose to display a more link and that will default to using the link defined by the user in the SiteNow People settings (or the provided default value). I'm having trouble coming up with a generalized way to connect a view block instance back to those settings. My best idea at this point is to create a token that represents the SiteNow People path setting, then add a field to the view block display settings that allows a default path to be set for the view, which can use a token. That would theoretically be achievable, but it feels a bit roundabout and potentially kind of a hack.

Some feedback on whether this seems like a good way forward or whether it seems like an indication that I'm trying to stretch the solution to meet the need would be helpful.

joewhitsitt:
Does this part have to tie 1:1 with the views more link? If I recall the current one doesn't so if you are thinking about a separate field (not sure how you store this data), why not just do the same thing? Maybe I don't quite understand.

richardbporter:
I couldn't get the pager to work on this branch but I'd be curious if that patch does anything on this PR.

pyrello:
@richardbporter I'm working on getting the pager functionality fixed exactly because I want to verify that we have the same problem. I'll let you know when its functional again.

pyrello:
As soon as I get the paging fixed on this PR, I'm going to test out whether the patch on this issue works for view block paging issues: https://www.drupal.org/project/drupal/issues/2605218

pyrello:
Adding the patch doesn't fix the problem in this PR either :|

richardbporter:
I'm not seeing the issue on this PR. I placed the the view block on a page and enabled the pager. The images don't seem to change but maybe that isn't illustrating it?

pyrello:
> I'm not seeing the issue on this PR. I placed the the view block on a page and enabled the pager. The images don't seem to change but maybe that isn't illustrating it?

Sorry, I did not re-export config. To see the issue, you'll need to pull and re-import config. Then you should be able to see the "Display options" in the block settings. Select "Left" for image placement. After that, the images should align left until you change the pager.

pyrello:
You can now see the potential resolution to #3115 via the PR. To see it, add the People list block a page and save different settings for the Image position display option (the default, despite what the option says, is "Right"). You'll also need to set the number of items to display to something less than the total items (e.g. 3) and select the "Show pager" option. After saving the block settings and the page, you should be able to page through results and see the image position you selected persist after using the pager. This should also be the case with heading size of the individual view row items.

I have added some todo comments to document some additional tasks that will be necessary to use this approach to move forward with the type of update we would like to make in #2730. However, it should be noted that we will also need to create new versions of the old blocks and write update hooks to migrate from the old blocks to the new ones.

bspeare:
I tested this locally and have a few notes:
- Tested toggling the view allow settings and they all work if unchecked or checked. 
- On the block, if "hide" the Title but leave the Credentials it will not display since we are rewriting the Title to contain the Credentials. 
- Also noted that the hide field settings, sort options, and exposed filters are ignored if you use the pager.

pyrello:
Ben's testing yesterday brought to light some lingering issues with finishing this, best summed up by this: https://www.drupal.org/project/drupal/issues/2605218#comment-14015595

richardbporter:
Just noticed that the "Show pager" option seems to be reset to TRUE every time you configure the block.

pyrello:
> Just noticed that the "Show pager" option seems to be reset to TRUE every time you configure the block.

That may be an issue with me renaming a variable and failing to rename it everywhere.

pyrello:
I [updated the patch](https://www.drupal.org/project/drupal/issues/2605218#comment-14020844) that is being used for this PR in a way that resolves most the the issues that were being encountered. The exception is exposed filters, which I can't even get to apply to the view in the block anymore regardless of AJAX. In fact, testing with and without the patch confirms that the issue seems to be unrelated to that. As best I can tell I considered the filtering working as of 18 days ago. I thought I had tested it more recently, but have lost track of when it happened. As far as I know that is the last thing that is broken.

Looking ahead to how to wrap this up once the above issue is resolved, there are a few items that need to be decided on:
* Should the block display class in this PR be implemented as an override of the core block display class (meaning that it replaces that class and is the class in use for all views blocks) or should it implement another views display plugin (meaning that it would be an additional and separate option from the existing block display)?
* How far should this PR go in "finishing" the new people block? Should we complete this option and transition existing people block instances over to it or do that in a follow-up PR?
* If we don't decide to "finish" the new people block in this PR, should we just hide it for now?

pyrello:
Here are the relevant pieces that are not working as expected right now:

In the following section of code, exposed filters are being determined from the view and added as fields to the block settings:
https://github.com/uiowa/uiowa/blob/055c18c06ed1518e84e4e9da8202845959ac866e/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L164-L203

Block settings for exposed filters are saved to block configuration:
https://github.com/uiowa/uiowa/blob/055c18c06ed1518e84e4e9da8202845959ac866e/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L476-L479

View exposed filters are updated based on block settings:
https://github.com/uiowa/uiowa/blob/055c18c06ed1518e84e4e9da8202845959ac866e/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L551-L553

The first two parts seem to be happening as expected. When I set breakpoints with `xdebug` and then save the block, I see the values coming through here:
https://github.com/uiowa/uiowa/blob/055c18c06ed1518e84e4e9da8202845959ac866e/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L551

This approach is adapted from this module which does basically the same thing (but only handles exposed filters): https://git.drupalcode.org/project/views_block_placement_exposed_form_defaults/-/blob/8.x-1.x/src/ExposedFormBlockDisplay.php

I'm pretty sure this worked at some point, but not sure how recently. I may have renamed a variable or array key that caused it to not work. Any help is appreciated.

richardbporter:
> Just noticed that the "Show pager" option seems to be reset to TRUE every time you configure the block.

This should be resolved with 5595aac.

richardbporter:
The only exposed filter to test this with people is tags right?

pyrello:
Yes. Although you could also modify the view to add more, if you wanted.

richardbporter:
I added a filter but how do I get it to show up in the allow settings? I only see tags still.

richardbporter:
I'm not having any luck with this. It looks like filters were maybe set differently previously: https://github.com/uiowa/uiowa/pull/2702/commits/a0417c318d05ded844cc84dfb166dd725701a9aa#diff-4d4408e67a8f25d1594cd344d3bdadde8d339e0a35bc8d48350beb35f2854c4cR223

Any idea at what point this was working?

pyrello:
My best guess is that I last tested it on 2/18, looking back through the updates to the OP to see when I added the checked checkbox. That was after I switched to the current method of handling the filters, which I feel fairly certain that I would have tested at least to the point where I was satisfied that they worked with Tags before marking them done. Also, since they are base on an existing module and I have compared the code between them multiple times, I don't understand why they work in the module and not in our custom code. This is one of the downsides of having a project drawn out over such a long time. I just can't remember when everything happened.

pyrello:
There is a patch that adds this functionality to `ctools_views`, which I have thought about copying, but it adds a lot of stuff that we are not already doing. It is at this issue: https://www.drupal.org/project/ctools/issues/2657060

In that patch, it allows for an exposed filter to still be displayed to the end user if the block editor doesn't select a value for block configuration. Because the approach is very different, it is hard to use as a basis of comparison. But I do note that in that patch, they also have the following line, which seems to be what is supposed to set up the filters that are set in block configuration to work on the view:
```
$this->view->setExposedInput($exposed);
```

joewhitsitt:
taking a look see

joewhitsitt:
so, i commented out the `public function usesExposed()` and it filtered as I expected. I noticed the view is setup to receive arguments and that the tags filter is “exposed” which might conflict with you setting usesExposed to false?

Screen Shot 2021-03-09 at 11 52 33 AM
Screen Shot 2021-03-09 at 11 52 06 AM
Screen Shot 2021-03-09 at 11 52 14 AM

We are adding a "hidden" class to the exposed filters for these blocks currently as it was the only way to get the items per page to work. `sitenow_people_form_views_exposed_form_alter()` and `sitenow_articles_form_views_exposed_form_alter()` - https://github.com/uiowa/uiowa/pull/513

richardbporter:
Yes! noticed that too but it seemed to cause other issues like no results when paging.

pyrello:
I just noticed this: https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/views/src/Plugin/views/display/DisplayPluginBase.php#L270-272

The `displaysExposed()` method seems to allow for determining whether the exposed form should show or not. Setting this to FALSE and removing the override of `usesExposed()` allows for the filter to be correctly set, but I'm still seeing the issues that @richardbporter described earlier with not showing the second page of results. Still digging into this aa bit.

pyrello:
I started another deep dive into this last night, trying to understand why the changes I made above caused everything to work correctly on the first load, but then for no results to be shown when the second pager link is clicked. I'm still tracking this through a bit, but it appears that the query that Views is constructing is not properly receiving the `items_per_page` and `offset` parameters when loaded via AJAX. On initial page load under the test I am doing, it receives the values `3` and `0` respectively for those values. When the second pager link is clicked, you would expect it to receive the values `3` and `3`. Instead, however, it is getting `9` and `9`, which is more results than test content I have created, which explains why I'm seeing no results.

joewhitsitt:
While working on https://github.com/uiowa/uiowa/pull/3243, I noticed a small downside in that using views block the "preview" of the view in layout builder is placeholder text compared to our current noodle blocks.

pyrello:
I noticed this too when testing that PR, but it was not my experience when testing this. It seems, with views and menus, that sometimes they display with a placeholder and sometimes without. I haven't been able to figure out the rhyme or reason of it yet.

pyrello:
The change to the view configuration does seem to cause this to work as expected. I'm still reacquainting myself with how to test it, but I'm encouraged that we may have a path forward.

I'm currently stuck on getting the person path token to work.

pyrello:
I'm almost ready for this to be tested again. The only thing I'm aware of that I need to figure out is how to strip the query string that gets added onto the URL of the more link.

pyrello:
This may require additional work, but I think it is ready for review.

pyrello:
I tried out adding the Person type filter to the person view that is included with this PR and it works! However, it is not very user-friendly yet. The same problem that we ran into with this filter elsewhere is present here, so we would need to adapt our solution to this before it could be utilized by end users.

cory-skeers:
Tested the view block settings, and they work really well!

One request would be if the accordions could appear more accordion-y. Not a blocker for this moving forward as more testing is done, but could be a helpful UX indicator when the accordions are closed.

image

pyrello:
Great feedback, @cory-skeers!

I think another improvement would be to have the accordion be open by default if the contents are different than the default.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -710,6 +818,65 @@ function layout_builder_custom_preprocess_media(&$variables) {
...
+  if (isset($row_styles) && !empty($variables['rows'])) {
+
+    // Convert single selection to an array for consistent processing.
+    if (!is_array($row_styles)) {
+      $row_styles = [$row_styles];
+    }
+
+    $default_classes = [
+      'card--list',
+      'card--person-teaser',
```
joewhitsitt, line 840: Should this be more general for non-person lists?

joewhitsitt:
So to merge, this person list view would be switched to disabled for now until we can use it to replace the existing block?

pulling down to test

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
bspeare: I think we need to generalize the `card--person-teaser` styles so that we don't have this issue.   Especially if we are using these styles for the new event content type.   I can make a follow-up issue detailing the work that needs to be done.

joewhitsitt (approved):
> So to merge, this person list view would be switched to disabled for now until we can use it to replace the existing block?

Assuming we can disable it for now, I approve to merge what is here.

i agree with https://github.com/uiowa/uiowa/pull/2702#issuecomment-832952177 but we have that issue on the menu block and others.

i have concerns using the core patch that nobody else from the community has weighed in on (https://www.drupal.org/project/drupal/issues/2605218#comment-14020844), but a lot of functionality is waiting on a solution for this so i think we should move forward.

i also don't love switching from view modes to individual fields for the view. i know this gives more flexibility, but will make it harder to account for when styling. if others agree maybe we can limit this to viewmodes for the default list blocks but keep the functionality open for other views.

richardbporter:
The core patch is also a concern of mine but we do have 13 core patches already, plus many more contrib. :)

pyrello:
Are there specific concerns about the core patch? I guess I don't see how this is any different than maintaining our own noodle code, except that the code is pulled in via patch. The patch is based on iterative work from the community with some adjustments based on my own work with this.

To be clear the patch (in some form) is absolutely necessary. If the patch is not okay, then this PR is a no go. The patch is what allows us to pull in block settings on AJAX requests. If we remove the patch, we are back to some elements of the view being reverted to the default view settings (e.g. not the block instance settings for the view).

pyrello:
@joewhitsitt When testing this, did you happen to test any existing view blocks to see if they were affected by it? I didn't include this in the instructions, so wouldn't expect that you had. But you are also a badass tester, so maybe you did?

richardbporter:
> Are there specific concerns about the core patch?

Just that it commits us to maintaining it, probably. A core update that changes this code may make the patch fail to apply, failing our builds. A security update would be worst-case scenario.

joewhitsitt:
Though I am a badass tester, I did not test other existing views. I assumed we "fixed" all those issues when the base functionality was added earlier to support admissions student profiles block. I can poke around this morning on some other views.

richardbporter:
Also, I'm not saying that is any worse than noodle code breaking after an update - could be the opposite. I have many concerns about many different things. Just expressing them. :)

pyrello:
I would assume that there is potential we broke all the views again with these updates, since this adds new functionality, including the core patch.

pyrello:
@richardbporter Me too! I wish that some of this functionality could get some priority to get into core. It is frustrating to have to depend on so many patches. But I think this is always going to be the case to some extent.

joewhitsitt:
Looks like it does break things. https://admissions.local.drupal.uiowa.edu/counselors no longer has the territory dropdown filter.

`displaysExposed()` seems to cause it not to show even though I resaved the view config to "allow" the territory filter.

Something like the below seems to work for the counselors situation. The exposed filter exists but is not "allowed" in the configuration of the list block so it assumes that it should be displayed. Thoughts?

```
  /**
   * {@inheritdoc}
   */
  public function displaysExposed() {
    if (empty($this->options["filter_in_block"])) {
      return TRUE;
    }
    return FALSE;
  }
```

pyrello:
Pulling down to take another look.

pyrello:
I think this is ready to go again.

joewhitsitt:
This fixed the counselors view. no issue with the student profiles block. poked around grad and didn't notice anything out of place. Thanks!

PR #2703, opened 2020-12-14
Concept: Outputting one block per view result

pyrello:
# Motivation

As the work moves ahead on #2676, I have been ruminating on the problem of being able to place views based lists of multiple items alongside individual component blocks. That is, if we a people (views) block styled to look like cards, the whole list of items will render as a single block, while an individual card will also render as a single block. This makes creating rules that target both of these things in an understandable way difficult. The idea here is that if we can create a way for individual views result rows to be placed as blocks, then they can be treated the same way that regular blocks can be. The idea for how to make this work comes from the [Paragraph Blocks](https://www.drupal.org/project/paragraph_blocks) module.

This PR is very much conceptual and still in progress. I published it so that the work is captured even if I'm not able to immediately finish it.

# Challenges
There are some gotchas when working through the logic of how this should work, namely:
* **How to deal with a large/unlimited result set?** I think it is clear that the usefulness of this functionality depends on limiting how many results can be placed. Even having 3 blocks per view display listed will quickly become unmanageable for finding and placing blocks. This is a limitation that also exists in Paragraph Blocks (presumably for similar reasons) that I was hoping to get around, but ultimately came to accept as an important limitation. Probably the best way to handle it is to create a setting on a view display that sets the maximum limit of blocks that can be created.
* **How to handle paging?** If blocks are placed individually, the concept of paging starts to fall apart. For one, where does the pager go? Does it become its own individual block needing placement? What happens when you click to a new set of results? Instead of having block that represents the Nth result, you'd need a block that represents the [items per page * (page - 1) + N]th result, which is quite a bit trickier to create and represent to the user.

# Assumptions
Based on the above, I'm making the following assumptions:
* Limited result set - this is currently hard-coded to 3, but could eventually be made configurable per view display.
* Paging options is hardcoded to "some" (which causes the pager to not be shown).

# How to test

So far, the only thing that is working is that the individual blocks per result are listed - up to 3 - to be placed. Actual rendering of the block is not working yet.


bspeare:
I tested this locally and have a few notes:
- Tested toggling the view allow settings and they all work if unchecked or checked. 
- On the block, if "hide" the Title but leave the Credentials it will not display since we are rewriting the Title to contain the Credentials. 
- Also noted that the hide field settings, sort options, and exposed filters are ignored if you use the pager.

PR #2707, opened 2020-12-14
Concept: Grid layout options per layout region

pyrello:
# How to test
* `blt frontend`
* `blt ds`
* Click 'Layout' tab
* Configure any section to see new options.
* After saving the section configuration and the page, inspect the section HTML in the page view.
* You should see classes added to the each region based on the grid setting you selected.


joewhitsitt:
Here are my initial thoughts about this method. Sean said this grid would apply to all blocks in the region. If so:

- How would this work with a card and a listing block (with title and more link) in the same region (column)? Do the individual list items flow as a grid as well? Grid styling would apply at the block level and again at the list level?
- Why would an editor use the 2 or 3 column layouts if they can get a grid of blocks inside a one-column layout?
- If a section region had multiple listing blocks, they couldn't have different grid settings.

That aside, I think the configuration is clean and potentially allows us better control of which regions should allow the setting based on the percentage widths.

Can you explain again what the benefit was again by doing it this way?

joewhitsitt:
Sean mentioned the viewbook's layout of majors: https://viewbook.admissions.uiowa.edu/ how it is a 2-column layout with text/card on one side and a grid of banners on the other side, with one non-list element at the end of the grid.

pyrello:
> How would this work with a card and a listing block (with title and more link) in the same region (column)? Do the individual list items flow as a grid as well? Grid styling would apply at the block level and again at the list level?

The approach in this PR alone would not work with card and list items together in a grid-formatted region. You'd end up with something like:

| column 1 ----- | column 2 ------- |
| card: --------- | list: ------------- |
| --------------- |  - 1 -  | - 2 - | - 3 -  |

There are two other PRs I created that seek to address aspects of this:
* #2702 would allow us to directly place view blocks instead of needing a block type wrapper to render them.
* #2703 would allow us to have individual view rows be placed as blocks, which would potentially allow us to drop the views wrapper also. This would mean that view row blocks could be placed alongside other blocks and have a common layout applied to both of them. There is a lot more nuance to this to make it work, but that is the basic idea.

> Why would an editor use the 2 or 3 column layouts if they can get a grid of blocks inside a one-column layout?

Good question.

> If a section region had multiple listing blocks, they couldn't have different grid settings.

Is this desirable behavior?

> Can you explain again what the benefit was again by doing it this way?

I think that having a single systematic way of applying this that doesn't rely on Layout Builder Styles (for which we are really pushing the boundaries of the use case) is preferable to maintaining a config-based approach that has to be maintained at the level of each block.

pyrello:
In discussing the needs of the Writing University site, it is apparent that there is a need/desire to figure out how to apply grid styles at the container/region level. That site is making extensive use of the "Grid 3x2 (stacked)" style - which was never supposed to be a thing outside of uiowa.edu... oops! However, it seems like there is a valid use case for making this work so that items can flow in a predictable order. There is an example of this in use at the bottom of this page: https://org-writinguniversity.prod.drupal.uiowa.edu/.

If a blocks are placed in a grid, they will flow like this:

1 2 3
4 5 6
7 8 9

And in mobile, they will follow that same order.

However, if you created the same order of blocks using columns via LB and viewed them in mobile the order would be:
1
4
7
2
5
8
3
6
9

This doesn't matter with some content, but with some the order is important.

So, I think that work on this PR should continue regardless of what happens with #2676. The combination of the two will create unfortunate combinations of options, but those things will already be possible with the existing tools.

pyrello:
Also, as part of this PR, I think we should migrate existing usage of the various grid options using LBS to the system used here.

pyrello:
Closing in favor of #4450

PR #2708, opened 2020-12-15
Make the section links show up on focus.

pyrello:
Small change to make the Layout builder section links show up on focus, which was not happening before.

# How to test
* `blt frontend`
* `blt ds`
* Click the "Layout" tab.
* To avoid having to tab endlessly, set the focus closer to the test by clicking in the empty space below the "Save Layout" button.
* Tab until you see the "X" and "Configure" links display.


bspeare (approved):
Great fix! Tested it locally on admissions.

PR #2709, opened 2020-12-15
Feature: Layout Builder double click block to edit

pyrello:
Adds a new module called `drupal/layout_builder_shortcuts` which provides the ability to open a blocks edit screen (while in Layout Builder) by double clicking the block.

# How to test
* `blt ds`
* Click the "Layout" tab.
* Test double clicking a block to see that it opens the block edit screen.
* Test editing a block through the contextual link.
* Test dragging and dropping a block.
* Test moving a block through the contextual link.
* Test removing a block.
* Masquerade as a webmaster.
* Visit the http://default.local.drupal.uiowa.edu/about and click the "Layout" tab.
* Attempt to double click a block in the Header section. This should not work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -123,6 +123,7 @@
         "drupal/layout_builder_lock": "^1.0@RC",
         "drupal/layout_builder_operation_link": "^1.0",
         "drupal/layout_builder_restrictions": "^2.6",
+        "drupal/layout_builder_shortcuts": "1.0.x-dev",
         "drupal/layout_builder_styles": "^1.0@beta",
         "drupal/lazy": "^3.6",
         "drupal/lb_direct_add": "1.x-dev",
--- a/composer.lock
+++ b/composer.lock
@@ -5877,7 +5877,7 @@
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/layout_builder_shortcuts.git",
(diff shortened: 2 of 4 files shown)


bspeare:
@pyrello Nice custom module! What does the addition to `layout_builder_custom.overrides.js` add that the module does not?

pyrello:
@bspeare Good catch! I originally added the JS and tested it in that file and forgot to remove it after I moved it to a separate module.

bspeare:
This still works well after removing the extra code.  One thing I noticed on https://uiowa.local.drupal.uiowa.edu/, is that if I double click on a component and save the block updates I am no longer able to double click on another block to edit.   I can still edit other blocks through the contextual links but the double click option is no longer working.

pyrello:
Hmm... interesting. I wonder if, after the LB element reloads, the script isn't getting attached again? Thanks for testing. I'll debug!

pyrello:
JS needed a bit of a rewrite to work correctly both after initial load and AJAX rebuilds. Should be working now.

cory-skeers (approved):
Tested on the conditions provided, and worked great!

Ran into one issue after moving a block by dragging, where the double-click and contextual links do not work until saving the layout. Contextual links don't work on `main` in this circumstance, though, so doesn't appear to be related to this PR.

pyrello:
@cory-skeers I'd be interested to hear more about this condition you encountered.

Issue #2530, opened 2020-11-12
Image issues with Articles news aggregator block and individual article content


mark-bennett-uiowa:
# Problem
The best way to describe this issue is to show it. These are examples from the [https://americanstudies.prod.drupal.uiowa.edu](https://americanstudies.prod.drupal.uiowa.edu) site. The first image is the Articles block news aggregator on the frontpage:
<img width="1361" alt="Screen Shot 2020-11-12 at 3 45 08 PM" src="https://user-images.githubusercontent.com/3067017/99000300-a85ea480-24fe-11eb-8ff2-f8a3d2b686d3.png">
(1) The images are too large (which I heard was being fixed in another PR) and work as if they are feature images.
(2) The entire block is linked, rather than just the title. If a user clicks on the description, they are redirected to the individual article. This is an issue with people, news and events at this point. Not sure if this is intended behavior or not.
(3) The only way to add an image to the article outside of the body field, is to use the feature image field/tab. This results in the following behavior on the individual article page:
<img width="1469" alt="Screen Shot 2020-11-12 at 3 45 58 PM" src="https://user-images.githubusercontent.com/3067017/99000584-202ccf00-24ff-11eb-9dc2-13235a4d0c9d.png">
(4) On the Articles view page, the image shows in the same manner as the front page issue above, rather than the events list does:
<img width="1217" alt="Screen Shot 2020-11-12 at 3 55 01 PM" src="https://user-images.githubusercontent.com/3067017/99000906-99c4bd00-24ff-11eb-86fc-fa03253e7e81.png">
<img width="826" alt="Screen Shot 2020-11-12 at 3 56 37 PM" src="https://user-images.githubusercontent.com/3067017/99000948-ab0dc980-24ff-11eb-9e42-ba3b6826b4ad.png">

# Steps to reproduce
Add a few news articles and add images to these entries under the Feature image field/tab. Events are pulled, so nothing to do here.
Add an Articles aggregator block on a page and display all options.

# Proposed solution
Can we switch this to a smaller, more-square, image with or without that circular border (like events)?
Can we hide the article image on the default display, but keep it in the edit page under an Article image tab rather than feature image tab?

joewhitsitt:
I think this needs to be broken up in to separate tasks.

In terms of "Aggregator" aka RSS Feed, card image options already exist and we are going to be working to extend this to other list blocks shortly: https://github.com/uiowa/uiowa/issues/2351

[...]

PR #2711, opened 2020-12-15
Featured image display settings

### LINKED ISSUE #2405: Change featured image field to allow users to set different heights
# Problem

Users have requested the ability to set different heights on their featured images in order to reduce page scrolling. 

Screen Shot 2020-10-26 at 2 41 32 PM

# Proposed Solution

Utilize the banner block refactor sizes to allow for users to set a small, medium, or large size on their featured image blocks. 

Screen Shot 2020-10-26 at 2 43 28 PM

## Next steps

- Determine a method to set the height for the featured image 
  - Through the node edit screen?
  - Or through a new custom block that can be edited within layout builder?

### LINKED ISSUE #2633: Global override option for featured image size
Followup to #2405 

# Story

As a webmaster I want set a global setting for featured image size so that it's easier to maintain consistency across the site (while still having the option for local settings if consistency is not my goal).

# Proposed solution

Add a site setting to allow setting a global default for image sizes.

Decision needed on how it interacts with individual page settings--disables the field, set a default but allow individual page overrides?

GaryRidgway:
* Resolves #2405. 
* Resolves #2633. 
* Removed some logic in `uids_base_preprocess_page` that sets a `has_node_image` variable that never gets used in a template.
* Unrelated update to `sitenow_articles` and `sitenow_people` settings forms that only runs a process to update all the nodes of that type if the path pattern has changed.

# TO TEST

## On default site
* `blt ds`
* You can set the new global settings for featured image display at https://default.local.drupal.uiowa.edu/admin/config/system/uiowa-articles and https://default.local.drupal.uiowa.edu/admin/config/system/uiowa-pages. The first is an added field at the bottom of an existing form and the second is a new form provided by a new custom module. The default setting for both should be set to 'Large.' Ensure that, despite the lack of an update hook, the defaults are in place.
* There is also a new 'Display setting' field in the 'Featured image' accordion in the right sidebar of the edit screen for pages and articles. You can view an example of this here: https://default.local.drupal.uiowa.edu/node/6/edit. The default setting per content item is to use the global setting for that content type.
* For both articles and pages, leave a piece of content at the default and test the global settings.
* Run `drush @default.local cst` to make sure that changing the global settings has not caused config to change.
* Then test all the per content item settings to make sure that they properly override the global settings.

## Check existing sites
This update changed some of the logic that determines the behavior of pages using the header layout.

* V2 sites such as hawkeyemarchingband.uiowa.edu
* admissions.uiowa.edu
* coronavirus.uiowa.edu


docroot/themes/custom/uids_base/templates/layouts/onecol--background/layout--onecol--background.html.twig
```diff
@@ -19,11 +19,10 @@
     'banner--gradient-bottom',
     'banner--gradient-dark',
     'banner--horizontal-left',
-    'banner--large',
+    "banner--#{field_featured_image_size}",
```
bspeare: @GaryRidgway Can you set it to default to `banner--large` and then have it adjust based on the size selection?

GaryRidgway: The select list is defaulted to large on the edit page. would you like it to default here as well or would the default on the edit page be enough for this?

bspeare: Probably something like this instead so existing featured image banners still look good: 
`field_featured_image_size ? 'banner--' ~ field_featured_image_size : 'banner--large',`

joewhitsitt:
pulling down to fix build and test

joewhitsitt:
Working to add this to

config/admissions.uiowa.edu/core.entity_view_display.node.page.default.yml
config/coronavirus.uiowa.edu/core.entity_view_display.node.page.default.yml

joewhitsitt:
So, I wonder if we can do a two for one with this PR as I don't like the none option here if it is the same as large. To remove the "none" option would me requiring the field and then running an update hook across all v3 sites to set the field. Otherwise editors working on pages without a featured image will get barked act to set a value even though they didn't provide an image.

Instead, there has been desire to have a featured image for teasers but not necessarily a featured image on the full node. Could "none" or not set default be large, remove the large option, add a "hide on full display" option which allows teaser/seo images.

bspeare:
@joewhitsitt That's a good question.  I put https://github.com/uiowa/uiowa/issues/2530 into parking lot for tomorrow because I know this is something that admissions would like to have.

joewhitsitt:
Also, we would have more flexibility to add/remove option in the future if we used handled these options through code similar to v2 styles, column widths, etc.

Otherwise, we may run into update hooks in the future to change this field. https://github.com/uiowa/uiowa/pull/2717

pyrello:
Here is a thought. What if we change this field label to "Display featured image as title banner?" and then re-label (programmatically) the 'None' option to 'No display.' Then the none option becomes the choice to not display the image as a banner. We would still need an update hook to set 'large' option for all the existing pages where a featured image is attached.

With regard to the idea of using an `allowed_values_function` callback to dynamically populate the values, I'm not necessarily opposed to that, but I think the reason it is being suggested is because it requires such a complex update hook to add/remove select options when using a static list. What I don't understand about using the 'allowed_values_function' is how does it protect us from having values stored in the database that are no longer valid? It seems like, if the values are dynamic, then if we can more easily add/remove them, then we can more easily create a situation we are storing values that are no longer meaningful/relevant/functional. I just don't understand how it works, so maybe this is accounted for in a way that I'm not aware of.

Additional notes for further development:
* Capture banner size from field in `uids_base_preprocess_layout` and set the banner class there rather than in the twig template.
* If we want the field to dynamically show/hide based on whether an image is present (we should), then we'll need some js logic like in `layout_builder_custom/js/banner-block-form.js` to check for the presence of the media item.
* Create a `sitenow_form_node_page_edit_form_alter` function in `sitenow.profile` to modify field settings, if necessary.

joewhitsitt:
>What I don't understand about using the 'allowed_values_function' is how does it protect us from having values stored in the database that are no longer valid?

My understanding, and I haven't worked with it in a while, is that it is similar to a taxonomy reference field. The value is stored until the next time the form/field is opened and saved. Like a taxonomy reference single list, it would return to none if the value was no longer present. Depending on what we are doing an update hook might be needed for quick updating, but otherwise we could let it clean itself up over time.

pyrello:
@joewhitsitt Thanks for the explanation. Based on that, I'm cool with this approach.

docroot/modules/custom/sitenow_articles/src/Form/SettingsForm.php
```diff
@@ -185,6 +185,24 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#size' => 60,
     ];
 
+    $form['global']['sitenow_articles_default_featured_image_size'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Articles Sitewide Default Featured Image Size'),
+      '#description' => $this->t('Sets the Default Featured Image Size for Articles'),
```
pyrello: Let's change this to 'Set the default behavior for how to display a featured image.'

docroot/modules/custom/sitenow_articles/src/Form/SettingsForm.php
```diff
@@ -185,6 +185,24 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#size' => 60,
     ];
 
+    $form['global']['sitenow_articles_default_featured_image_size'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Articles Sitewide Default Featured Image Size'),
```
pyrello: Can we change this to 'Display featured image'?

docroot/modules/custom/sitenow_articles/src/Form/SettingsForm.php
```diff
@@ -185,6 +185,24 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#description' => $this->t('Sets the Default Featured Image Size for Articles'),
+      '#options' => [
+        'do_not_display' => $this
+          ->t('Do not Display'),
```
pyrello: Can we change to 'Do not display'?

GaryRidgway:
For my new Laundry list of things:

- [x] Create allowed values function that can work for both pages and articles
- [x] Set 5 options for display in the allowed values, `Site default[DEFAULT]`, `Do not display`, `Small`, `Medium`, `Large`
- [x] change label to `Display setting` on node featured image size
- [x] Set `- none -` to Site default in hook
- [x] change template logic to work with new options.
- [x] Link to settings page in help text.
- [x] Create an articles field that also uses this same allowed values function
- [x] Move that field to the featured image dropdown section
- [x] Discern best way to give pages the same config as articles, and then execute on that, giving pages similar config to articles, defaulting to Large
- [x] link pages to the sitewide default
- [x] link articles to sitewide default
- [x] hide select list on articles and pages visually unless a picture is selected for featured image

joewhitsitt:
Build is failing:

```
FILE: ...ocroot/modules/custom/sitenow_articles/src/Form/SettingsForm.php
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 201 | ERROR | [x] Array closing indentation error, expected 6 spaces
     |       |     but found 8
 203 | ERROR | [x] Line indented incorrectly; expected at least 6
     |       |     spaces, found 0
 203 | ERROR | [x] 6 spaces found before inline comment; expected "//
     |       |     '#required' => TRUE," but found "//     
     |       |     '#required' => TRUE,"
----------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

docroot/themes/custom/uids_base/templates/layouts/onecol--background/layout--onecol--background.html.twig
pyrello: This logic is moved into the `uids_base_preprocess_layout` function to keep the logic in the twig template cleaner.

pyrello:
It occurs to me that HR is going to, at a minimum, be having the new global settings show up. But since that site is on a different theme, I think that implementing this feature would require entirely separate work in `uiowa_bootstrap`, which seems like it should probably be out-of-scope for this PR, since HR is hopefully moving to `uids_base` in the not-too-distant future. Probably the easiest way to resolve this is to communicate to them that these new settings have no effect for them until they move off their current theme. What do you think, @joewhitsitt?

pyrello:
Instead of going to the trouble to hide the "display setting" field on the node edit screen when no image is uploaded, I just added help text to the effect that it won't do anything unless an image is uploaded.

joewhitsitt:
I haven't pulled down yet, but would it be too much trouble to set a defaults and restrict access if the active theme is not uids_base with a @todo to remove when uids_base is our only theme? I say this, because we have been teased with uids_base adoption before and it still isn't done...

pyrello:
@joewhitsitt If I'm understanding your comment, I think we would need to set a check and restrict access in these places:
* Featured image display field on SiteNow Articles settings page via `'#access' => FALSE`
* SiteNow Pages settings page - this is new and only has the one field, so maybe we could add an access check to the route?
* Display setting field on Article add/edit form via `'#access' => FALSE`
* Display setting field on Page add/edit form via `'#access' => FALSE`

This would remove any indication that the feature existed for HR. As for setting a default, it is already set via the addition of `sitenow_articles.settings.yml` and `sitenow_pages.settings.yml` files, which establish the 'large' display option as the default (which matches up with the current behavior). I don't think we need to do anything else.

pyrello:
I updated the testing instructions to include checking that changing global default settings does not cause config to change.

joewhitsitt:
Pulling down to test

joewhitsitt:
- ~~Is the site-wide default supposed to be controlled by webmasters? If so, we need updated permissions splits.~~ 7515983 0fb8d13
- Remembered that uiowa.edu has custom templates without the featured header setup.
- ~~HR doesn't seem to need anything because permissions are split and so is the page display settings. Not sure why it isn't displaying on the article form.~~ 9fe9bd5 5e5a36bbca250adabed43dbce41e50ca9e60dfa1
- ~~We should break the SiteNow Article settings up. Update the first fieldset to be View Page Settings and then create a second field set for Article Settings.~~ 9f20e7f5d80f6bb8e5928d7d3f999be46891d3eb

pyrello:
@joewhitsitt These all look like solid improvements. And good catch on the split permissions, which I had totally spaced on.

joewhitsitt:
Does iowasuperfund want these page/article form and display settings?

pyrello:
Yes. I'm not sure how I missed that. I searched for instances of `entity_view_form` and `entity_view_display`.

joewhitsitt:
@pyrello could you check iowasuperfund? i exported the config for article which was split. the featured image i tested with is not showing and the form has an undefined index issue. then i realized that featured image isn't displaying on the prod site. just for teasers.

joewhitsitt (approved):
Tested on hr, hawkeyemarchingband, uiowa, gis.sites, admissions.

Aside from iowasuperfund, https://github.com/uiowa/uiowa/pull/2711#issuecomment-752712222, I think everything is working as expected

pyrello:
I'm checking iowasuperfund.

pyrello:
I think I decided to split the article type on iowasuperfund.uiowa.edu early on when I was getting the hang of doing development using this platform and the client told me he didn't think he'd ever need certain fields. However, it is not worth it to hide a few fields to have to keep that config in sync, so I removed it.

Issue #2649, opened 2020-12-02
Changes to Counselors Listing/Map


joewhitsitt:
# Issue

The admissions team came back with a few changes to the counselors map/listing functionality.

# Proposed Solution

- [x] Change views exposed filter - Any - to - All - in territory filter. This probably has to be done programmatically.
- [x] Remove territories except for states, dc, international recruitment from the select list field, export changes. This may result in data loss but the customer is aware.
  - [x] Move international recruitment to the top of the list.
- [x] Remove puerto rico from the map (e.g statesdata geojson)

GaryRidgway:
So I have done some investigation for removing the six necessary territories for checkbox two. After speaking with @joewhitsitt we have determined this issue to be a bit more complex than initially pointed, and so are going to merge checkboxes 1 and 3 with https://github.com/uiowa/uiowa/pull/2700 in and repoint this issue for the second checkbox.

joewhitsitt:
was there any work on this yesterday?

cory-skeers:
Finished up by #2717

PR #2717, opened 2020-12-16
Update admissions counselor territory options

cory-skeers:
Resolves final checkboxes of #2649.

# How to test

- Go to http://admissions.local.drupal.uiowa.edu/counselors
- Check that "International Recruitment" option appears at the top (below " - Any - ")
- Check that only "International Recruitment", states, and DC are included in the list (American Samoa, Guam, Northern Mariana Islands, Puerto Rico, United States Minor Outlying Islands, and Virgin Islands, U.S. have been removed)
- Check that a counselor person type has the same above options for their territory


joewhitsitt:
WOW! seriously?! Fast work @cory-skeers. Can we move this to admissions_core.install and remove the if at the beginning?

I will test here shortly.

cory-skeers:
Good call! Forgot there was a separate admissions module. Got it moved over and removed the initial check.

Had a nice base to work off of--guessing there are some spots that could be cleaned up that are leftovers from the previous update, but functionally it should be good to run and test!

GaryRidgway:
You are a madman Cory. A very skilled madman, but a madman none the less.

joewhitsitt (approved):
Worked for me! Thanks

PR #2719, opened 2020-12-17
Allow for minimal text in CTA

### LINKED ISSUE #2710: Allow Call to action block to use minimal html as an option
# Problem
A few site owners (CLAS, Admissions) have requested the ability to add links or minimal text formatting to the Call to action block. 

# Proposed solution
- [x] Create update hook to update field to (formatted, long) in place of (plain, long) and keep existing data.
  - [x] Update existing data to format as `plain_text` as part of this.
- [x] Allow `minimal_html` and `plain_text` formats. 
- [x] Ensure `plain_text` is the default format for newly created CTAs (with `minimal_html` as a secondary option).
- [x] Update component css (https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/cta.scss) and add `.cta__content p { margin: 0;}`

cory-skeers:
Resolves #2710 

# How to test

- Sync a site, make sure update hook / config updates fire without issue
- Check existing CTAs on the site, and see that their existing summaries are still present and set to `plain_text`
- Create a new CTA, check that it defaults to `plain_text`
- Try saving a CTA with `minimal` text format, check that it saves correctly.
   - Check that when editing the newly created block, prior to saving the layout, it retains `minimal` text format
   - Check that the CTA content doesn't receive extra margin spacing (the `<p>` when using `minimal` should be styled to `margin: 0`

Some preliminary test CTAs are available on the sandbox site at https://sandbox.prod.drupal.uiowa.edu/calls-action


mark-bennett-uiowa:
Well done Cory! I synced down American Studies and verified that by default, the CTA text area was set to plain text and summaries remained:
Screen Shot 2020-12-21 at 3 58 17 PM
I then changed it to Minimal HTML and received the popup warning about changing text formats. I saved the block changes and re-entered the edit to verify that it remained in Minimal HTML format:
Screen Shot 2020-12-21 at 3 59 39 PM
I also created a new section, added two new CTA blocks and verified that both were set to Plain text by default. I changed one of them to Minimal HTML and changes persisted:
Screen Shot 2020-12-21 at 4 02 08 PM

mark-bennett-uiowa (approved):
Approved. See testing above. Code looks good.

cory-skeers:
Thank you @mark-bennett-uiowa!

PR #2721, opened 2020-12-17
Initialize studentlife.uiowa.edu multisite on uiowa02

quamsta:
This is my first attempt at provisioning something for our studentlife.uiowa.edu site. Things I'm a bit unsure about/may have done wrong:

- I forgot to add a requester=dquam when running blt umc studentlife.uiowa.edu -- thought maybe this is related to SiteNow requests? Is it required if I'm the admin?
- Our SSL situation is messy. We have an existing cert for studentlife.uiowa.edu, but I have no idea if it's valid to use with Acquia. We basically want to start with a local dev environment only for now, so I'm hoping we can figure this out at another point?
- I didn't request a dev/staging CNAME, because of us just using local dev right now. Is this okay?

Thanks!

D


joewhitsitt (approved):
We can login as admin and give you access (requestor). You don't have to go live right away so the SSL step is not necessary yet.

dev/stage domains can be added/removed as needed. {sitename}.dev.drupal.uiowa.edu

PR #2725, opened 2020-12-18
Create uiowa_apr module and feature split.

### LINKED ISSUE #2463: Port Drupal 7 APR module to Drupal 8
Remaining issues:

- [x] Get directory introduction to show up
- [x] Add form validation for config page routes and paths
  - Otherwise the returned HTML does not output correctly if routes are bad, conflicting, etc.
- [x] Fix double slashes in sitemap controller output
- [x] Theme everything

richardbporter:
Resolves #2463 

### Testing
- `blt frontend`
- Sync the **default** site. Only certain URLs work with APR.
- Enable the APR split and import configuration.
- Copy the the API token from the D7 Tippie site APR config page.
  - This is the only API key that I know of. Maybe there is another?
- Go to the admin/config/content/uiowa-apr and paste the API token in and accept all other defaults.
- Test out the routes at their default locations:
  - /apr/people
  - /apr/publications
  - /apr/people/sitemap.txt
- Reconfigure the module with different options and test again.


docroot/modules/custom/uiowa_apr/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,230 @@
...
+    ];
+
+    foreach ($fields as $field) {
+      $path = $this->aliasCleaner->cleanAlias($form_state->getValue($field));
+
+      /** @var \Drupal\Core\Url $url */
+      $url = $this->pathValidator->getUrlIfValid($path);
+
+      // If $url is anything besides FALSE then the path is already in use.
+      if ($url && substr($url->getRouteName(), 0, 9) != 'uiowa_apr') {
```
richardbporter, line 213: @joewhitsitt Initially, I copied the path validation from sitenow_people et al. but it did not catch route conflicts with those modules. For example, changing the APR path to people did not catch the conflict with sitenow_people default path. Curious what you think of this method and also if you think we should add any specific help text related to sitenow_people.

joewhitsitt: if this works to catch view pages, etc. go with it. i remember you bringing this up previously with sitenow_people and sitenow_articles but we never revisited it. what kind of help text did you have in mind?

richardbporter: Not sure if we wanted to highlight any confusion with sitenow_people configuration or not. I guess we can probably add that to the docs.

richardbporter:
I cannot get responsive tables to work with the APR directory list view. @GaryRidgway and I talked briefly about this and we think its because the APR Vue app is replacing everything in the #apr-directory-service div. Debugging the scripts, it seems like the APR script is running before tables.js but for some reason there are no table rows when tables.js fires and the result is a table with headers only.

¯\\\_(ツ)\_/¯

bspeare:
@richardbporter Not sure about the card listing since all the images are different sizes.   Is there a D7 site example where the card listing looks good?

Screen Shot 2020-12-23 at 3 58 08 PM

joewhitsitt:
pulling down to test

joewhitsitt (approved):
I think the styles are wrangled enough. I used education's apr token with success. wasn't knowledgeable enough to customize Publications Departments but i can confirm that the webmaster role can configure.

richardbporter:
Thanks. I think we're just waiting for the `is-striped` class to be added to the table. Thats on the AIS side. Since this is not enabled for anyone, I don't think this needs to wait for it.

IDK if @bspeare has any other style changes planned but I think MVP is done. I'll merge this after CI passes.

bspeare:
I think this is good enough for MVP.   More work should/could be planned for making the profile page look better and similar to our other profile pages.

richardbporter:
> More work should/could be planned for making the profile page look better and similar to our other profile pages.

Agreed but that could be difficult if it depends on markup changes on the AIS side, especially since APR is going away.

Also just noting for reference that the grid display is disabled at this point. There are some comments in the code.

PR #2730, opened 2020-12-18
Card styles for all the things!

joewhitsitt:
# To Test

This extends card style options to events, articles, people blocks within v3. Styles should control child items. Generally, basic use of the website as well as the update hooks should not present errors.

uids_base style clean up will be saved for another day

## HR

- This PR should not affect HR's website. Article, Events, People listings should look the same as it is sub-themed off uiowa_bootstrap.

## v2 Websites (e.g. hawkeyemarchingband, diversity, law, brand)

- Check SiteNow Articles/People view pages including the People manual sort option. Should look the same.
- Check list/grid/masonry style options for events, articles, people, featured content paragraphs. Should look the same.

## v3 Websites (e.g. sandbox, uiowa, admissions, grad, international)

- Existing layout blocks should be set with PROD defaults and look similar:
  - Articles: left-aligned content, right-aligned small widescreen image
  - Events: left-aligned content, left-aligned small circle image with border
  - People: left-aligned content, right-aligned small square image
- New layout blocks should have the following defaults with the option labels showing (default) appended.
  - Articles: White background, left-aligned content, right-aligned small widescreen image
  - Events/People: White background, left-aligned content, left-aligned small circle image with border
- Changes between widescreen image and square/circle should change the image view mode and clear necessary caches.
- SiteNow People/Articles view pages should be unchanged for now and should look similar to a v2 website,  including the People manual sort option.


joewhitsitt:
If anyone has ideas about how to trigger a media viewmode change from several layers away, let me know. Block configuration > View > Node Teaser > Media Item. I think I hit a bump with the articles/people blocks.

Just tried to do this in the template with a hardcoded viewmode, but I don't think it generates the image style.

```
{% set image = content.field_image.0|merge({'#view_mode': 'large__square'}) %}
{% set image = image|render %}
```

joewhitsitt:
I didn't make it as far with @pyrello 's idea about using https://github.com/uiowa/uiowa/blob/1a71b9c20da97c8c435984eabc5fff7382b2c89c/docroot/themes/custom/uids_base/uids_base.theme#L200-L233 to solve the image viewmode switch. I didn't see the view (has the parent_lb_styles as a display handler option) as another level to traverse. I then started working with sitenow_articles_preprocess_views_view to see if I can pass the parent_lb_styles down to the article node but ran out of time.

cory-skeers:
Working on this, I tried going through `sitenow_articles_preprocess_views_view` and `uids_base_preprocess_field`. Going  from the field, I ran into the same dead end as you, where I could get up to the level of node, but not to the view that is pulling in the node.

Going from the view end, I could go down into individual result rows and the respective node attached, but got a little lost trying to set the media display to anything but the default set by the node display (teaser). This seems like the right way to go, but either I was trying to set the media display option incorrectly, or at this point (in the `...preprocess_views_view` hook) any attempted handling of the media gets undone by downstream processes handling the `teaser` display. The most I could accomplish was either it retaining the `widescreen` display option despite any changes, or to remove the image completely.

Had briefly entertained the idea of adding node display options and/or templates in addition to `teaser` to get around the above issue, but that seemed out in the weeds of clean, extendible components, and didn't dive into that rabbit hole.

pyrello:
@cory-skeers I wonder if you are running into an issue where some cache needs to be unset for the change you are trying to make to actually work.

joewhitsitt:
I think I got it, but had to change responsive_image_style_id which is different than we have done previously which concerns me on top of the concern over using this referring item thing.

joewhitsitt:
@dnkemper and I paired and knocked out the defaults for new and update hook for existing.

pyrello:
Pulling down to test.

joewhitsitt:
Planning to give this another look after disconnecting from it for a bit. Pulling down

pyrello:
@joewhitsitt If there is any clarifying you can do in the testing instructions, that would be helpful. Having a little trouble understanding what all I need to check.

joewhitsitt:
Premature with the review on this.

joewhitsitt:
- I think event styles (v3) will need to be adjusted based on this change. The section that contains these have specific style settings. Removing the styles affects the "Events" block's padding.
Screen Shot 2020-12-30 at 1 09 36 PM


Screen Shot 2020-12-30 at 3 11 46 PM

joewhitsitt:
Starting to think about splitting templates for the v2 split...
Screen Shot 2020-12-30 at 2 17 50 PM

joewhitsitt:
Except for https://github.com/uiowa/uiowa/pull/2730#issuecomment-752727709, I think this is ready for review again.

I won't merge until we have a plan for these advanced layout situations.

pyrello:
I have reviewed the code and I think the approach makes sense generally. It might be a good idea to add a comment somewhere in `uids_base_preprocess_media` that explains why the code is there instead of the other preprocess functions we tried first.

Also, I have tested some of the basic functionality of articles and events and have found no issues so far.

joewhitsitt:
@bspeare could you take a look at how this PR affects https://github.com/uiowa/uiowa/pull/2730#issuecomment-752727709 (event blocks on the homepage) and how we might work around it?

bspeare:
Pulling down to test

bspeare:
@joewhitsitt The first workaround that I can think of is to add the margin styles to individual event block component to set the spacing within each block:

Screen Shot 2021-01-04 at 8 54 31 AM

Or, another workaround could be to add a `Vertically centered` section style to apply `align-items: center; justify-content: center;` to the wrapper.

bspeare:
@joewhitsitt Do you still want help to "AUDIT any uids_base styles can be removed related to events, articles, people."?

joewhitsitt:
@bspeare if you can spare some time, you have a better eye for what is still needed

joewhitsitt:
>add the margin styles to individual event block component

Would these be hardcoded to the event template or as lb_style defaults? This would be the transparent deployment option right? The other would require manual adjustment after deployment?

bspeare:
> > add the margin styles to individual event block component
> 
> Would these be hardcoded to the event template or as lb_style defaults? This would be the transparent deployment option right? The other would require manual adjustment after deployment?

Good question.   Adding a `Vertically centered` section style might be the easiest to update pages that need a manual adjustment after deployment.   Introducing margin options on the individual event block might not be the best path forward.

bspeare:
@joewhitsitt I think the margin issue still needs to be solved via https://github.com/uiowa/uiowa/issues/2670.

bspeare:
@joewhitsitt actually, I don't think the vertical centering will help.  It will still create an uneven spacing if the event titles differ:

Screen Shot 2021-01-04 at 10 39 47 AM

 I think I'd like to introduce more granular padding styles to map to: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss#L230-L249.   Right now we just have an `All` option and for this case we could use `Top` and `Bottom` padding options for each individual block.

bspeare:
Should we remove the "Right" option?  It looks like it is no longer needed. 

Screen Shot 2021-01-04 at 10 50 33 AM

joewhitsitt:
I can remove the "Right" option.

Do I dare say we should remove event from the scope of this for now until we can address this specific setup?

I worry about adding more responsibility on a content editor to understand css properties, especially when our options can all be applied at once and could conflict on render.

bspeare:
@joewhitsitt That's probably a good idea.  It would be ideal if the event block would be replaced by the local event solution - do you think?

bspeare:
We are setting the `card__title` size based on the image size for the `left` or `right` options here with an adjacent selector if the media block exists: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/card.scss#L30-L63.   

Screen Shot 2021-01-04 at 10 52 40 AM

This doesn't translate well to the lists and we should probably adjust how this is implemented.

bspeare:
@joewhitsitt for the events, the `V2` version is using the `uiowa-events` class, is that correct?  And it looks like the `v3` version is using the `event-item` class.   Can we attach a custom `V2` events library to the events module?  If so, can you point me to where I can attach that library (https://github.com/uiowa/uiowa/tree/main/docroot/modules/custom/sitenow_events)?  It looks like the `V2` version needs a lot of custom styles while the `v3` version should be able to use most of the default card styles.

joewhitsitt:
I am removing the event block from this PR.

Correct: v2 has the uiowa-events parent class. v3 has the event-item child class. With https://github.com/uiowa/uiowa/pull/2676, I was going to introduce `uiowa-events-wrapper list-wrapper` parent class for v3.
To attach to v2 events, I would do it here: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_events/modules/sitenow_events_paragraph/sitenow_events_paragraph.module#L44

joewhitsitt:
> [`card__title`]... we should probably adjust how this is implemented.

The only way I can think to do it right now would be some css:nots on `block-inline-blockuiowa-events`, `block-inline-blockuiowa-articles`, etc. or "undo" those styles immediately after scoped to those same block classes. Do you have another idea @bspeare?

bspeare:
@joewhitsitt maybe we can scope it to a css:not `list-wrapper` - will that be the consistent class wrapper across all the lists?

joewhitsitt:
it can be. i will start adding that programmatically

bspeare:
> @bspeare if you can spare some time, you have a better eye for what is still needed

@joewhitsitt I think the css audit would probably be better handled for future work.  It will probably blow up the scope of this pr to add this work and testing.

bspeare:
> it can be. i will start adding that programmatically

I had the wrong class here.  `list-container` is the class that we want to use if we plan on implementing `.inner--grid.grid--twocol--50-50 .list-container` for the grid.   It looks like you remembered.

joewhitsitt:
I did list-container, but had to add list-item because the card block doesn't have anything to tie the `:not()` to.

bspeare:
@joewhitsitt testing http://sandbox.local.drupal.uiowa.edu/people-directory and I'm noticing that the right-aligned images have the `card--media-left` class on them when it should be `card--media-right`.  Also updating the "Image position" to `Left` or `Right` does not change the image position.

bspeare:
Should we update the teaser people view to look closer to the work done for the counselor teasers? Or create an issue for future work?

Current teaser:

Screen Shot 2021-01-04 at 3 22 28 PM

Counselor teaser:

Screen Shot 2021-01-04 at 3 22 43 PM

joewhitsitt:
oh gosh... what happened. i am seeing weird behavior now... i swear it wasn't always like this.

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -204,6 +204,7 @@ function uids_base_preprocess_media(array &$variables) {
   switch ($mediaType) {
     case 'image':
       $referring_item = $media->_referringItem;
+      // So far, only way to get media viewmode switching lb block > view > node > media deep.
```
pyrello, line 227: Thanks for adding this!

joewhitsitt:
> Should we update the teaser people view to look closer to the work done for the counselor teasers? Or create an issue for future work?
> 
> Current teaser:
> 
> Screen Shot 2021-01-04 at 3 22 28 PM
> 
> Counselor teaser:
> 
> Screen Shot 2021-01-04 at 3 22 43 PM

I was hoping to do SiteNow People and Articles style option functionality at a later time.

dnkemper:
When looking at v2 marching band site, if I add the articles paragraph and want to display three articles with a masonry setting, it only displays two across and wraps the third article underneath the 2nd article. Should three be displaying across a page or nah?
Screen Shot 2021-01-04 at 3 35 17 PM

joewhitsitt:
> When looking at v2 marching band site, if I add the articles paragraph and want to display three articles with a masonry setting, it only displays two across and wraps the third article underneath the 2nd article. Should three be displaying across a page or nah?
> Screen Shot 2021-01-04 at 3 35 17 PM

Masonry does that if there aren't enough items...

joewhitsitt:
`card--list` was affecting article/people blocks with switching media positions. implemented a workaround until we can set styles through config for SiteNow Articles and SiteNow People.

however, I am getting weird layout builder effects until I save the layout..
Screen Shot 2021-01-04 at 5 09 03 PM

bspeare:
@joewhitsitt Is that on the sandbox site?  I think that is because it it placed in the "Header" section.

joewhitsitt:
I think fcad9de, e7172c3 fixes the issue with the people block media position. With that, I think this is ready to test again. Just ignore the bizarre issues related to the header section as they are unrelated to this PR.

bspeare:
@joewhitsitt I just checked it out and it seemed to be working.  One thing we might want to change is to remove the margins to the `card__media` here https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/blocks/people.scss#L47-L51.  It is adding extra spacing that is probably no longer needed.

joewhitsitt:
Fixed for v2, people manual sort, v3 block lists.

Remove the card__media styles as requested. Switched the update hook to change existing to small. Large and medium were larger that what is currently out there. Small will be a visual difference, but one that most will appreciate.

bspeare:
Looks like we need to remove this extra margin on the image: https://github.com/uiowa/uiowa/blob/i2351-extend-card-options-part-deux/docroot/themes/custom/uids_base/scss/blocks/people.scss#L49

GaryRidgway:
Everything looks good, styles defaults etc, except for the following:

when i change a page in a pager, the styles get completely broken for people: https://i.gyazo.com/8be6586731928523e9dd89cd0e8f2e34.gif

as well, none of the  options that had the descriptions, details, or images hidden pulled over when i synced down and ran the hooks on specifically articles and people.

joewhitsitt:
@GaryRidgway 
Thanks for finding that paging issue. I will take a look in a bit.

>as well, none of the options that had the descriptions, details, or images hidden pulled over when i synced down and ran the hooks on specifically articles and people

Can you be more specific? Like v2, v3 the hidden styles were not applying?

pyrello:
> when i change a page in a pager, the styles get completely broken for people

Possibly related? https://www.drupal.org/project/drupal/issues/2605218

GaryRidgway:
Well, i was about to go to prod to take a picture of working style options but apparently something broke that and we didnt catch it at some point...

Screen Shot 2021-01-06 at 1 10 29 PM

It is there in the style options as well, so no hook has destroyed that in the db from my knowledge:

Screen Shot 2021-01-06 at 1 12 45 PM

I will create an issue in uiowa

joewhitsitt:
@GaryRidgway i just pushed a fix for the hidden styles issue

GaryRidgway:
ah okay, wonderful, thanks!

joewhitsitt:
> > when i change a page in a pager, the styles get completely broken for people
> 
> Possibly related? [#2605218: Views Block Display skips preBlockBuild() call on ajax rebuild](https://www.drupal.org/project/drupal/issues/2605218)

I tried patch https://www.drupal.org/files/issues/2020-04-22/2605218-49.patch but didn't see an immediate effect. The classes all get stripped off. My guess is the patch solves it for view blocks but not noodle blocks.

This is an issue with our headline field as well. If I have a h5 specified, it changes to an h2 on the next "page"

joewhitsitt:
The paging issue is a big problem for articles, people. Though headline is also affected, it is transparent to most users. Aggregator does not having paging. Might put together a separate PR that introduces this functionality for just events.

pyrello:
In testing this after I resolved the merge conflicts, I can't seem to get it to work. It is my understanding from reading through the comments that this PR was considered to be ready for testing, so it seems that subsequent updates to the platform have caused functional regressions that will need to be addressed.

Also, worth noting that it might be a good idea to attempt to use one of the newer patches from https://www.drupal.org/project/drupal/issues/2605218, because previously we had to use an older patch to be able to apply to a Drupal 8.x platform.

Finally, in reviewing this again, it seems like there is a lot of styles in the `people.scss` and `articles.scss` files that could just be replaced by applying established card style classes.

joewhitsitt:
>so it seems that subsequent updates to the platform have caused functional regressions that will need to be addressed.

 probably relates to the layout_builder_style thing.

The previous use of the patch didn't seem to help so not sure if a newer patch would make a difference.

From the PR summary:
>uids_base style clean up will be saved for another day

pyrello:
> The previous use of the patch didn't seem to help so not sure if a newer patch would make a difference.

There are different versions of the patch available for the version of Drupal we are now on.

pyrello:
I fixed the instances I could find where `#layout_builder_style` was being used. I also tried the more recent patch for 9.1 in the above issue, but it did not appear to fix our issue.

One other thing I noticed that I haven't jumped into yet is that setting the style on the first set of people on the test page (thank you Sandbox site!) seems to cause other instances after that to adopt that style.

Screen Shot 2021-02-15 at 5 08 35 PM

joewhitsitt:
Thanks for updating those @pyrello. Haven't had a chance to revisit this. Bummer about the pager still.

>One other thing I noticed that I haven't jumped into yet is that setting the style on the first set of people on the test page

Perhaps related to https://github.com/uiowa/uiowa/pull/2730#issuecomment-754667984

bspeare:
Just tested the articles page: http://sandbox.local.drupal.uiowa.edu/articles.   I'm not seeing the pager issue like there is with People, but all the article blocks are taking on the style of the first one and it is not in the Header section.   Images are rendering as widescreen ovals and it looks like there is some CSS work to be done to remove some old styles in `articles.scss`.

bspeare:
I guess the pager issue is still there on articles, it is just not as jarring because there are backup styles.

richardbporter:
This has some merge conflicts that need resolved first. Otherwise, testing this locally downgrades drupal-vm. I can start working through those now.

richardbporter:
I guess these update hooks should be renumbered as 9xxx right?

pyrello:
I can maybe get the merge conflicts worked out.

> I guess these update hooks should be renumbered as 9xxx right?
I believe so, yes.

richardbporter:
It was just sitenow.install which I pushed. Take a look and make sure it as expected compared with what is in main.

richardbporter:
I tried the latest patch. It didn't seem to work. I'm not sure that is what we are experiencing here.

richardbporter:
Sadly I did not get anywhere with this. It is odd that the other view operations seem to work fine. I would kinda expect the sorting and other things to be off as well but I wasn't able to replicate that.

joewhitsitt:
My workaround thought would be to keep the classes on the parent and handle as much with css similar to v2 paragraphs views. For the widescreen/square switch, could the view request a different node view mode where that media config is set? If we did more CSS then we would probably need different class names so that it doesn't affect the parent.

pyrello:
I think I understand now why the applied patches are not working. The patches in https://www.drupal.org/project/drupal/issues/2605218 are making changes to the `Drupal\views\Plugin\views\display\Block` class to update its `preBlockBuild()` method, which is ultimately called by `Drupal\views\Plugin\Block\ViewsBlock`. Because we are wrapping the view with a block content and then assigning our classes via `HOOK_preprocess_block()`, we are never triggering that specific `preBlockBuild` method. So the patch updates may fix the issue, but our implementation isn't affected by the fix.

pyrello:
Based on some additional investigation I did in #3115, I'm not sure that there is a good path forward for directly passing any data consistently between the block instance and the view. Moreover, the way that the LBS are being added to the block make it even less likely for us to find overlap. So, @joewhitsitt's idea about setting up the views items to get their styling from the parent block classes seems like our best (and possibly only) route forward.

richardbporter:
> Based on some additional investigation I did in #3115, I'm not sure that there is a good path forward for directly passing any data consistently between the block instance and the view. Moreover, the way that the LBS are being added to the block make it even less likely for us to find overlap. So, @joewhitsitt's idea about setting up the views items to get their styling from the parent block classes seems like our best (and possibly only) route forward.

If I understand correctly, that route would fix the styling issues but not the heading change, image style change, etc. Is that correct?

joewhitsitt:
We only have two image possibilities (square/circle, widescreen) which could be view modes on a separate but almost identical view displays. I am stumped on the headings unless we want to take a step back with that functionality and just make them divs instead of headings and style appropriately. An a11y loss but not a violation.

pyrello:
Good point. I hadn't thought about the fact that those things couldn't be inherited via cascading styles.

My next best idea about how to handle this is to create a drupal.org issue and introduce a patch that adds an event dispatcher that gets triggered via the `\Drupal\views\Plugin\views\display\Block::preBlockBuild()`. I think this would allow us to access the block and the view at the same time, even on a AJAX request and pass data between them.

richardbporter:
Whats weird is that it only seems to be those display_handler options that are not populated after changing the page. If you create a people block that filters by tag, say Mascot, and change the `node--person--teaser` template such that the parent_lb_styles equal the view arguments, they persist.

`{% set parent_lb_styles = view.args %}`

Notice the TID 16 set as a class on page 1 and 2.

Screen Shot 2021-02-22 at 4 39 34 PM

Screen Shot 2021-02-22 at 4 39 53 PM

pyrello:
I'll try making that change to see what gets populated. Maybe there is another way at this :)

richardbporter:
Its a bit hacky but might be a short-term workaround. It does look like we can pass additional, arbitrary arguments to the view. The values have to be strings though and the keys are numeric which isn't great. The template would need to be updated. Maybe there is another option for storing this data...

Long-term, something more like #2702 would probably be better but a few more noodles in the spaghetti for right now isn't so bad. :)

pyrello:
#2702 handles passing data in the same way, so whatever we figure out for this affects that PR too.

richardbporter:
I wonder if the `#context` render array property introduced in 9.2.0 would persist from block -> view -> node?

https://www.drupal.org/node/2971707

richardbporter:
> I wonder if the `#context` render array property introduced in 9.2.0 would persist from block -> view -> node?
> 
> https://www.drupal.org/node/2971707

Doesn't seem like it in my local testing with 9.2.x-dev.

b3d3bb3 is a POC of passing data as additional view args which doesn't seem to affect the view since they are not in use. Again, since the keys are numeric there is not a great way of accessing these in the template. Should the order change, this will break. 

Test out the people listing pages on sandbox again. Feel free to revert b3d3bb3 if there is another route being taken that I'm not aware of.

joewhitsitt:
Haven't tested this yet, but these do take a tag argument. Is that accounted for?

joewhitsitt:
pulling to test

richardbporter:
I briefly saw your comment about the tags - I think that is arg 0. This isn't as hacky as I thought because `views_embed_view` lets you pass any number of additional args. Arguments set up as contextual filters in the view need to come first, in order.

joewhitsitt:
It works (_except for the media viewmode switching_)!


https://github.com/uiowa/uiowa/pull/2730#issuecomment-783710168 (multiple teaser view modes to work around this potentially, but the people block already has a couple displays based on sorts)

pyrello:
I haven't tested this, but in looking at the code and reading the conversation, this approach sounds brittle. If we get it working, it seems like we are going to have to keep an eye on it to make sure that nothing changes unexpectedly. I'm not saying we shouldn't do it necessarily, just that it makes me a bit nervous.

> I wonder if the `#context` render array property introduced in 9.2.0 would persist from block -> view -> node?
>
> https://www.drupal.org/node/2971707

I failed to report back on this, but I reviewed the relevant views code in the 9.2.x branch to see if this was actually being implemented there yet and I didn't see any evidence of it. It remains a possibility that it might happen, but I don't think they are utilizing that property for views yet.

joewhitsitt:
Closing in favor of #2702. Will update the related issues

PR #2734, opened 2020-12-21
Feature stat no hover

### LINKED ISSUE #2583: As a site owner, I can adjust the stat component to display the content area without a hover effect
# Story

As a site owner, I can adjust the stat component to display the content area with a hover effect

# Proposed solution

Add a layout builder styles option like `Remove hover effect` to the stat component to show the content area and override the default hover effect. 

- [x] Refactor to remove `stat--transform` class if `Remove hover effect` option is selected: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-statistic.html.twig#L30-L32 
- [x] Apply a modifier class based on `Remove hover effect` of something like `stat--static` and target that class with something like:
```
 .stat--static .stat__content {
    position: static!important;
    clip: auto;
    height: auto;
    overflow: auto;
    opacity: 1;
}
```
- [ ] Edit the stat content field help text (it currently references the hover effect which may or may not be used: "Optional text to appear below the statistic on mouse hover, providing additional detail for the statistic."
    - Or, because the default behavior is to include the hover effect, should this remain as is?
- [ ] Determine style option naming
    - "Remove hover effect"
    - "Display content field by default"
    - "Persistent content field display"

cory-skeers:
Resolves #2583 

# How to test

`blt ds --site=default` or your preferred test site
`blt frontend`
Create (or edit an existing) stat block
Add the `Remove hover effect` style
Check that the stat no longer exhibits the hover behavior (the `content` text appears, correctly styled, statically below the statistic)
Check that a stat block without the `Remove hover effect` style still exhibits the proper hover effect (the `content` text appears hidden until hovering over the statistic block).

# Feedback
As part of this, we're trying to determine the naming for the style option. Few initial thoughts are:

- "Remove hover effect"
- "Display content field by default"
- "Persistent content field display"

Preferences from these options, or alternatives, highly welcomed!

Additionally, the help text for the content field references the hover effect, which is the default behavior. Depending on if the style label changes, we may want to update that help text as well.

### DIFF
--- a/config/default/field.field.block_content.uiowa_statistic.field_uiowa_statistic_content.yml
+++ b/config/default/field.field.block_content.uiowa_statistic.field_uiowa_statistic_content.yml
@@ -10,7 +10,7 @@ field_name: field_uiowa_statistic_content
 entity_type: block_content
 bundle: uiowa_statistic
 label: Content
-description: 'Optional text to appear below the statistic on mouse hover, providing additional detail for the statistic.'
+description: 'Optional text to appear below the statistic on mouse hover, providing additional detail for the statistic. To always show this text, remove the hover effect in the display option styles.'
 required: false
 translatable: false
 default_value: {  }
--- a/config/default/layout_builder_styles.style.cta_remove_hover_effect.yml
+++ b/config/default/layout_builder_styles.style.cta_remove_hover_effect.yml
@@ -1,0 +1,12 @@
+uuid: 70549ec1-6ee7-43f7-9c40-f00db44c27e2
+langcode: en
+status: true
+dependencies: {  }
+id: cta_remove_hover_effect
+label: 'Remove hover effect'
+classes: stat--static
+type: component
+group: default
+weight: 0
+block_restrictions:
+  - 'inline_block:uiowa_statistic'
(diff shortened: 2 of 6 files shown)


richardbporter:
> As part of this, we're trying to determine the naming for the style option. Few initial thoughts are:
"Remove hover effect"
"Display content field by default"
"Persistent content field display"
Preferences from these options, or alternatives, highly welcomed!

I read "Remove hover effect" as the most intuitive name before testing this out. After testing it, I can understand the other options now. I still think "Remove hover effect" is fine but maybe we want to mention it in the field description?

richardbporter:
Something like this, perhaps?

```Optional text to appear below the statistic on mouse hover, providing additional detail for the statistic. To always show this text, remove the hover effect in the style options below.```

cory-skeers:
I like it @richardbporter.

Added it with a slight modification: `Optional text to appear below the statistic on mouse hover, providing additional detail for the statistic. To always show this text, remove the hover effect in the display option styles.`

richardbporter (approved):
Nice! Works as advertised.

PR #2743, opened 2020-12-22
Change search query string to terms and update uiowa.edu schema to match.

richardbporter:
Resolves #2736 

Made some additional changes while I was in there:

- Code formatting
- Dynamically set search placeholder text based on CSE scope config
- Dynamically show "Search all..." link on results page based on CSE scope config

### Testing
- blt sync --site uiowa.edu
- Inspect schema source and check it matches.
- Test search form input and results page.
- Change CSE scope config and test out search again.

PR #2747, opened 2020-12-23
Fix some search regressions.

richardbporter:
Follow up to #2743 

Always show search-all link since CSE can be configured to search multiple domains. Set placeholder text back to "Search this site" since there is no way to detect otherwise. Change form element keys and document they control styling.

Testing
- Click search
- See placeholder
- Enter search terms and submit
- See search all link


pyrello (approved):
Tested, appears to work.

PR #2749, opened 2020-12-23
Fix undefined index on metatag config page.

richardbporter:
Just noticed a warning on the metatag config page. I think the serialization is messed up but imports/displays ok. Not sure how @joewhitsitt got this config exported like this.

# To Test

- No errors on the uiowa.edu metatag configuration page.
- Imports fine. Identical when export is attempted.
- Happy Google?
- Happy Michael?


joewhitsitt:
So, we were [relying on a patch to provide searchAction target](https://github.com/uiowa/uiowa/commit/bc7ac0f8f76419d598f9c9ed087d064c71e91a13#diff-ae135a64417fd3ee46c0c4e73dd99335f47b8bd3742912be112c5e42ffc82b6a) as Google wants it. That patch no longer applies but the workaround from the issue might work.

https://www.drupal.org/project/schema_metatag/issues/3055278

Need to try pushing this live and see if Google accepts it. Their new validator gives a green check mark. Whatever that means.
Screen Shot 2020-12-30 at 4 54 40 PM

richardbporter:
Thanks for explaining. I was confused how this got exported previously. We must have removed that patch at some point.

richardbporter:
I would say approve this @joewhitsitt for the next release then.

joewhitsitt:
I removed the patch thinking it had been merged in when we updated the module...

joewhitsitt (approved):

Issue #1626, opened 2020-05-18
Revisit split permissions configuration


joewhitsitt:
## Problem

We have entered into the growing-pains phase of split permissions. One little piece of config or functionality could trigger a split of permissions. As the number of roles increases and the number of sites splitting config we either need to figure out a better way of dealing with permission splits. Gone are the days where I was knowledgable about all of the custom sites on the service (hr, brand)

Permissions is the first major example of this, but other things will become more visible over time like basic page fields.

<img width="694" alt="Screen Shot 2020-05-18 at 11 31 25 AM" src="https://user-images.githubusercontent.com/4663676/82237620-914a4b80-98fb-11ea-8ba3-6fcb4955d652.png">

_this list is only going to grow_

A number of questions come up as I try to enable a new module for the service and set permissions accordingly.

- Should this split config get the new permission?
- Since it isn't required functionality, do they want it?
- What other things do I need to know about this site before making this change?
- If I make this change, how do I know the right person reviewing the PR will sign off on it?
- Should it just be a changelog type thing and notify folks to change things on their own time?
- Should I just make separate issues, tag the sites affected and leave it to the developer to decide?

I thought maybe once a site splits, their permissions should just be ignored and stay active config, but if we wanted to send out an updated permission to all, this would be an update hook and we would still run into the same issue of what sites should get it.

Should there be some kind of flag configuration in uiowa_core that states the "kind" of site it is like "Accepts all new base config unless it conflicts." But then if it conflicts, what then?

joewhitsitt:
Another thought I just had, was to double down on the number of roles and create webmaster_base, editor_base, etc. Make those pretty stock and then let most splits (unique content type permissions, etc.) be site specific roles with site specific permissions. Two initial problems I see is the number of roles (max_input_vars) and making sure the weights don't affect the base role configuration so those can stay synced.

pyrello:
Just dropping this here to review later: https://www.drupal.org/project/config_role_split

pyrello:
@joewhitsitt Do you think this is effectively handled by #4137?

joewhitsitt:
@pyrello from what I have seen, yes. Additionally, the work done to remove empty config directories also provides more visibility into what sites are not default.

Issue #2714, opened 2020-12-16
Create base MAUI integration with upcoming dates by session display


pyrello:
# Story

As a site owner I want to be able to show a version of the Academic Calendar, possibly filtered by event category, so that I can display upcoming events to my visitors.

Example: https://registrar.uiowa.edu/dates-by-session

# Proposed solution

Adapt whatever solution is being used to display the above page to create a similar feature for Drupal 8. Add the ability for the site to set an event category to limit the events displayed to just that category.

richardbporter:
I'll start looking at this.

richardbporter:
I'm wondering if we can treat this issue as a subtask/split from #2184 since I think only the upcoming dates by session view is needed. That might allow us to get it done faster and revisit the rest of the MAUI functionality later.

pyrello:
Sounds good to me.

richardbporter:
I had envisioned using custom blocks for this, maybe similarly to the aggregator block. Is that ok or were you thinking of a "page display" like SiteNow People or something?

pyrello:
I had just imagined a page like SiteNow People, but I'm open to having it be a block.

richardbporter:
The D7 panel panes allow for additional content to be placed before/after the dynamic displays. That was a common request and was why I initially was thinking of using custom block types in D8. If we use custom block types, we'll probably want to use the shared headline field which means we can't provide the block via the module (https://github.com/uiowa/uiowa/pull/2152#issuecomment-708500240). That means we probably have a lot of config to split out into a feature. It may be easier to just turn this on for everyone since there is no harm in doing so. We'd probably want to move the block placement in LB to the "More..." section off-canvas though as there will be quite a few blocks eventually.

The uiowa_maui.module and api/ functions _could_ be ported as-is with D8 replacements for deprecated code although I was thinking a custom service might be more appropriate. Then this issue would just need the dates_by_session pane from uiowa_maui_pane submodule ported to a block.

bspeare:
Would this be a good issue to implement the event list design suggestions from https://github.com/uiowa/uiowa/issues/2727?

richardbporter:
> Would this be a good issue to implement the event list design suggestions from #2727?

Maybe? Is that because they don't have an image and location like other events?

I didn't even think about using events for this but I guess that makes sense.

bspeare:
[...]

Issue #2359, opened 2020-10-16
Standardize multisite config split machine names


richardbporter:
### Problem
We cannot execute Drush/BLT commands targeting all multisite config splits because the machine names are all different. We (I) documented that multisite splits should be named the same as the URI. In hindsight, having them all be the same would be useful.

Ideally, we would be able to do something like this:
`blt uiowa:multisite:execute 'config-split:export site'`

### Proposed Solution
- [x] #4095
- [x] #4096
- [x] Update multisite split machine names to be "site". 
  - [x] Move site splits to subdirectory under config/sites/
  - [x] Update docs.

joewhitsitt:
"site" like `admissions` or "site" like `admissions_uiowa_edu`?

pyrello:
I think like `site_admissions_uiowa_edu`

richardbporter:
I was thinking literally `site` so that we can run the command above, i.e. export each site split en masse. Since site splits exist in site splits, there shouldn't be any conflicts.

mark-bennett-uiowa:
Ive been creating my site splits with labels like admissions.uiowa.edu. The machine name becomes admissions_uiowa_edu then. Shouldn't we avoid any naming conflicts either way (whether we prepend 'site_' to the front or not)??

richardbporter:
The directory would still be different per site obviously because they have to be. Just talking the machine name here.

richardbporter:
Here is en example scenario of why I thought this would be useful.

I need to update the same configuration in multiple site splits. ~~I write an update hook that modifies the configuration.~~ I run `blt uiowa:multisite:execute 'config:set foo bar'` on the sites and then run `blt uiowa:multisite:execute 'config-split:export site` to export each site's `site` config split. We can't do this right now.

richardbporter:
Sorry, update hook was a bad example. I meant just a Drush command to set some arbitrary config or add a permission, etc.

mark-bennett-uiowa:
Agreed. So directory would be `../config/sitename.example.edu`, label would be `sitename.example.edu`, and machine name is `sitename_example_edu`, correct?

richardbporter:
After all the subtasks, each site split would be named `site`  with folder location `../config/sites/sitename.example.edu`. The machine name is controlled by the config entity name so we'd be renaming all site splits to `config_split.config_split.site`. It'll require an update hook to import that.

richardbporter:
[...]

PR #2776, opened 2021-01-04
Feature: Config Role Split proof of concept

pyrello:
Relates to #1626

# Updates
Except for adding a module to the default config, I am going to keep this scoped to the Admissions website for now, just because the entire process of switching over to this system will take some time and it is annoying to have to keep things up-to-date as permissions are changing.

* Adds Config Role Split module to default `core.extension.yml`.
* Adds 3 `config_role_split.role_split.*.yml` files to their respective config splits: `admissions`, `areas_of_study`, and `scholarship`.
* Updated `config_filter` and `config_split` modules to their most recent versions.

# How to test

##  Test importing
* `composer install`
* `drush @admissions.local uli admin/people/permissions`
* The permissions should be a combination of the `user.role.*.yml` files in `config/default` and `config_role_split.role_split.*.yml` config files in `config/admissions.uiowa.edu`, `config/features/areas_of_study`, and `config/features/scholarship`. I don't have a better idea for checking this other than spot-checking items from all of these files to make sure that they are showing up on the permissions page.

## Test exporting
Export test instructions coming soon.


richardbporter:
This is interesting! I bet this would be a big DX improvement given the number of site/feature splits we have. This was one of the motivations behind #2359 but I think we'd still want that regardless.

pyrello:
This really only solves one aspect of a larger issue, where there is config that requires duplication. `core.entity_form_display`. and `core.entity_view_display` are two other examples where we end up with large chunks of duplication that need to be accounted for when we make changes to the default config. So, I think that #2359 is still very much needed.

But this would be a big DX improvement because there are so many files involved and so many opportunities to mess it up with the current method.

pyrello:
I'm stuck on this because updating Config Split to 2.x (or possibly Config Filter) seems to be causing some issues that I can't account for. To reproduce:

* `composer install`
* `drush @grad.local config-split:export site_grad_uiowa_edu`

`git status` shows:
  ```
    deleted:    config/grad.uiowa.edu/core.entity_form_display.node.person.default.yml
	deleted:    config/grad.uiowa.edu/core.entity_form_display.node.person.minimal.yml
	deleted:    config/grad.uiowa.edu/node.type.scholarship.yml
  ```

But those files are all validly different (because the Grad split adds extra fields to person node type and modifies the name and menu properties of the scholarship node type).

@richardbporter If you have the time/inclination to try to help me figure out why this is not working, I would appreciate it.

pyrello:
Another observation: The major downside of this approach is that we lose the ability to use the `user.role.*` wildcard conditional split option, which has the benefit of picking up newly added permissions without needing to know they are there. Config Role Split requires understanding exactly what permissions are needed and updating them as changes happen. Permissions won't get automatically "picked up".

I opened a new issue that captures my idea for how this could work better and in a more general fashion: https://www.drupal.org/project/config_split/issues/3191356

richardbporter:
I'll try to get #2714 to a point where there is data but needs styles and then take a look at this.

richardbporter:
I'm not sure whats going with the config export issue for me. I synced admissions, got a config diff error, exported and had A TON of config changes. I must be doing something wrong.

Overall, the UI for this is such that it doesn't seem that much better. Having to manually copy/paste those permissions around isn't great.

pyrello:
It seems like there is not much point in proceeding with this PR until there is a better UI for the Config Role Split for setting permissions.

I'm going to mark this as blocked for the time being. Hopefully they develop a better UI for this.

pyrello:
I still have a way to go with figuring out how to properly and fully test the changes of moving to new versions of config_filter and config_split, but I did some testing earlier today that was promising. Also added a patch which adds an Admin UI for setting the permissions in the role split.

Screen Shot 2021-06-29 at 2 44 55 PM

This PR is back in play!

pyrello:
I'm closing this in favor of #4091.

PR #2796, opened 2021-01-06
Feature: Fix to remove card heading and body size adjustments based on layout column

bspeare:
When the card "Display options" were introduced, we added an automatic heading and body size adjustment if the image size was set to `Medium` or `Large`.     Now that we are introducing https://github.com/uiowa/uiowa/pull/2730, this PR backs that change out until we can implement a better way to implement this sizing through the "Display options" so that it is clear to the site editor. 

**Before:**

Screen Shot 2021-01-06 at 9 08 57 AM

**After:**

Screen Shot 2021-01-06 at 9 09 14 AM

# How to test

Add a card to one column layout and set the image to `Medium` or `Large` with `Left` or `Right` positioning.   Verify that the heading size and card body font size remains the same.


pyrello (approved):
Reviewed, not tested.

PR #2800, opened 2021-01-06
update config_ignore config

joewhitsitt:
to fix deployments


pyrello (approved):
Reviewed, not tested.

PR #2803, opened 2021-01-07
Added command to fix xdebug log error to post provision script.

pyrello:
I'm not totally sure this works because I didn't want to run `vagrant destroy` ATM because of other work in progress. But it didn't cause a failure when I ran `vagrant provision`, so at a minimum it doesn't seem to cause any issues.

# How to test

If you have a local environment that you don't mind blowing up:
* `vagrant destroy`
* `vagrant up`
* `vagrant ssh`
* Run a PHP-related command... I think `drush` or `blt` commands normally cause the error.


richardbporter (approved):
Thanks. That was annoying.

PR #2804, opened 2021-01-07
Store APR publications departments as text and convert to JSON.

richardbporter:
Running through a quick APR demo myself, I had trouble getting this config to work. Asking folks to enter valid JSON is probably a bad idea. Tippie is not customizing the text of any of its department overrides. We can add that later through an alter hook if needed but seems like overkill.

### Testing
- See #2725 for setup.
- Enter department names, one per line. The placeholder values should work for tippie.
- Go to publications route and test out the department filter.


pyrello (approved):
Reviewed, not tested.

joewhitsitt (approved):
glanced. not tested.

PR #2822, opened 2021-01-08
Use string representation of show_switcher config in APR attribute value.

richardbporter:
Apparently I overlooked this the first time around. This needs to be written out as 'true' or 'false' otherwise the JS doesn't render. 

¯\\\_(ツ)\_/¯


pyrello (approved):
Reviewed, not tested.

PR #2827, opened 2021-01-08
Feature: Spotlight content type

### LINKED ISSUE #2823: Create Spotlight content type
# Story

As a site owner I want a way to feature one or more people on the site so that I can tell a story about them at is newsworthy or timely.

Example of this type of feature: https://www.grad.uiowa.edu/profiles/moala-keshei-bannavti

# Proposed solution

Build a spotlight content type.

pyrello:
# Updates

* Resolves #2823: Adds a Spotlight content type to the grad.uiowa.edu split.

# How to test

* `drush @grad.local uli node/add/spotlight`
* Create Spotlight with reference to existing person
* Create Spotlight with reference to new person


bspeare:
It looks like this layout is using the core two column layout and not the `uids-base` two column layout.  Should we try to change the icons or names to be able to distinguish between the two?

pyrello:
I wonder if we could just use a preprocess function to make them not show up regardless of whether they are allowed by LB restrictions or not.

pyrello:
Here is the current version of this:
Screen Shot 2021-01-17 at 12 38 34 PM

mark-bennett-uiowa:
@pyrello I think this looks great! Have you thought about adding a block implementation for a frontpage or another page to point to this spotlight? I think CLAS would use this for sure. I have been creating these using the existing widgets and not all that thrilled with how it looks.

pyrello:
@mark-bennett-uiowa I have not. For the launch of the Grad College site, I am planning on just using static cards on the home page: https://grad.prod.drupal.uiowa.edu/ (see the "Profiles of Success" section). I think eventually being able to reference them dynamically would be good also.

bspeare (approved):
I was able to create and reference and create a profile using the provided instructions for using the Spotlight content type.

pyrello:
Given that this was built as grad-specific and we didn't end up using it, I'm in favor of closing this and starting from scratch if this feature ever gets built.

PR #2833, opened 2021-01-08
Create MAUI module with configurable academic dates block.

### LINKED ISSUE #2714: Create base MAUI integration with upcoming dates by session display

richardbporter:
Resolves #2714.

### Testing
- `blt dsa` a site.
- Place the `Academic dates` block on a page somewhere.
- Test out the heading size field.
- Test out the exposed filters: session and category. The date list should update after changing either of them.
  - Do this with multiple blocks placed on the same page.
- Edit the block and select a session and category for pre-filtered data. This should hide the exposed filters.
- Test out the `@session` placeholder text you can use in the headline field.
- Test out the Hide Session Badges style option.
- Review the addition of the HeadlineHelper class and the changes in other spots:
  - uids_base_preprocess_paragraph()
  - sitenow_people
  - sitenow_events
  - sitenow_articles
  - The actual field itself in layout_builder_custom/
- Review code changes.


richardbporter:
I think this is ready for review but @pyrello might want to take the first crack at it in case anything was overlooked for GC site.

pyrello:
I reviewed the functionality and it appears to work almost exactly like the version on the Registrar site, with the exception of the shared date heading, which @bspeare is working on.

That being said, I realize I didn't do a good job of fully communicating what I needed this to do for the Grad College site, which is a little bit different. The main difference is that I need to be able select a session and set the category from the block settings and have them not show as form options to the end user. I can document this in another issue. Given that direction, it seems like we have an option between building this block out like we have with previous blocks, where we create a custom block type and add fields to it to capture those details. Or we could use a Block plugin class that would extend the base block class and allow for adding the fields to the block form via that class.

This is similar to how views can provide blocks to LB. I was exploring how to expose more view options to blocks to allow us to place them directly vs having to continue to create custom block with fields as wrappers around the view in the PR: https://github.com/uiowa/uiowa/pull/2702. After exploring that a bit, I think that is likely where we are headed long term.

I'm bringing this up now because it would require fundamentally reworking the approach that is taken in this PR if we were to go that direction and I think it would be better to do it now than to merge this PR as is and then rework it later.

pyrello:
I added an example of how what I'm talking about in my last comment could be done in e559782. It is meant to show a possible way forward, not trying to force the issue. If you decide not to go that route, I can revert it.

richardbporter:
Yea I went that way initially and actually prefer it but I don't how to use the shared heading field programmatically. Are you saying we don't that field for this block? I thought we did for a11y since it can be placed anywhere in the page.

pyrello:
I'm a little fuzzy on placing custom defined field API fields in a form. It seems like you should just be able to just create a field by referencing `'#type' => 'uiowa_headline'`. But I'm not sure if that would work or not. Outside of the basic form element behavior, it seems like we are manually applying the application of it to children elements in the various instances where it is implemented, so it seems like this would just be replicating that pattern.

richardbporter:
I tried that with the aggregator block but did not have any luck with it.

pyrello:
I also tried `'#type' => 'uiowa_headline_widget'`, thinking it might be the ID of the FieldWidget plugin class that mattered, but cannot get it to display that way either. I'm fairly positive you can reference a field type like this in form api, but getting good search results on the overlapping usage of "field API" and "form API" is tricky.

richardbporter:
Yea I never found any docs on it. My assumption was that it wasn't possible because they're two different APIs: form vs field. I briefly looked at creating a custom form API element that shared functionality with the field but gave up in the interest of time.

richardbporter:
We could pretty easily add checkboxes to the block that control the display of the exposed form elements.

pyrello:
@richardbporter Ugh. For some reason I thought you could implement items defined via the field api in custom forms. It's weird and confusing that Drupal would have these two different API's that would seem to overlap but don't.

pyrello:
> We could pretty easily add checkboxes to the block that control the display of the exposed form elements.

I guess that may be the way to go for now without blowing up the scope of this PR.

richardbporter:
> Ugh. For some reason I thought you could implement items defined via the field api in custom forms. It's weird and confusing that Drupal would have these two different API's that would seem to overlap but don't.

Yea I know. Another benefit of providing blocks programmatically is it allows them to be grouped under a custom category when adding them to a page. I think MAUI will need this eventually as there will be many blocks.

richardbporter:
Is there any way back from the custom block if we move forward with it? An update hook? I'd still prefer to provide custom module blocks as part of the module if we ever figure this out. I'd still prefer that for aggregator also.

I imagine the update hook would be difficult...

pyrello:
@richardbporter In this case when you say custom block, do you mean the block plugin implementation? Or using the custom block content type?

I think either way, there is an update hook path back to the other one if it comes to that. In the former, we'd be pulling block configuration and converting it into field values; in the latter, the reverse. Block configuration is just a set of key value pairs and field values are pretty easy to access from the block entity, so I don't think the update hook would be that hard.

richardbporter:
Would it be ok to rename the AcademicCalendar block? There will be a skeuomorphic calendar block at some point: https://registrar.uiowa.edu/academic-calendar I was hoping to use that name for it.
Like AcademicDateList or something?

pyrello:
@richardbporter I changed the class name. If you want to set it to something else, go ahead. This should all be cleaned up now.

pyrello:
# Todo
- [x] Finish building out `uiowa_core_headline_form_partial` to provide functional form partial for heading text and other settings.
- [x] Add headline form elements to block configuration form.
- [x] Add relative session selection to block configuration.
- [x] Add category selection to block configuration.
- [x] Update logic in `AcademicDateBlock::build()` to check block configuration to see if options were selected in the previous two items and use that to determine whether to show the exposed filter blocks.
- [ ] Add hidden fields with selected values from block configuration?

richardbporter:
Sorry for the big refactor. I ran with an idea.

> Add headline form elements to block configuration form.

The only thing not working with this is the visually hidden checkbox. I'm not sure how to pass that in as I didn't see other examples of it.

It may be easier to create a theme function and render the headline field in the block rather than passing so many variables down to the uiowa_maui_session_dates template.

docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Build the block.
+   */
+  public function build() {
+    $config = $this->getConfiguration();
+
+    // @todo Write a headline theme function to render it here and use DI here.
+    return \Drupal::formBuilder()->getForm(
```
richardbporter: We could render the headline here and avoid passing all the config around. I think we need a theme function defined to do that maybe?

Also a note to use DI here instead of \Drupal.

pyrello:
@richardbporter:

> It may be easier to create a theme function and render the headline field in the block rather than passing so many variables down to the uiowa_maui_session_dates template.

I agree with this approach. I think this would also be consistent with how we are handing this off to a separate template in other cases too.

Overall, this looks pretty good, as much as I could tell from a quick perusal.

I'm wondering if there is a reason to have the `HeadlineHelper` class in the the `layout_builder_custom` module instead of `uiowa_core`?

richardbporter:
> I'm wondering if there is a reason to have the HeadlineHelper class in the the layout_builder_custom module instead of uiowa_core?

Technically no. I put it there because thats where the field is defined. It could live anywhere though.

pyrello:
Probably the field should be in `uiowa_core` too. Only the processing that is necessary for it to work with LB should be in the `layout_builder_custom` module. I know it was already like that prior to this PR, just thinking about how we are structuring our dependencies.

richardbporter:
Yea that makes sense. Lets put it there along with the theme function.

richardbporter:
I'm looking at the relative session block config now.

richardbporter:
Ugh. The API call we're using to get dates right now only allows one session ID. I think thats why I called this DatesBySession in D7 and exposed the filter. There is another API call that accepts multiple session IDs but only shows upcoming dates, i.e. it filters out dates in the past. There is no category parameter to that API call either.

Do you need to show all dates for a session or would only upcoming dates suffice?

pyrello:
Per block, I am good with only being able to select one session at a time. Are you saying that we can't set two different sessions in two different blocks?

richardbporter:
> Per block, I am good with only being able to select one session at a time. Are you saying that we can't set two different sessions in two different blocks?

Multiple blocks on a page with different session selections should definitely work. I didn't think of that as a workaround. 

Do you know how the headline hide style is being applied? I'm still struggling with that.

pyrello:
@cory-skeers or @bspeare Any insight on how the hide headline style is being applied where it is currently being used?

richardbporter:
> @cory-skeers or @bspeare Any insight on how the hide headline style is being applied where it is currently being used?

I got it, I think.

richardbporter:
> Add hidden fields with selected values from block configuration?

I'm not sure what this means.

 I added some help text to the block session/category fields to hopefully make things clearer as it is a bit complicated. Not sure if _- Exposed -_ is the right word to use.

There is currently an issue with the ajax callback not working the second time. I'm investigating how to fix, otherwise we can just revert the unique ID change but that breaks multiple blocks on a page. I want to say this is essentially the issue: https://www.drupal.org/project/drupal/issues/2821793

richardbporter:
It may be possible to use the unique ID in combination with a custom ajax response that allows for a CSS selector instead of just an ID string: https://www.drupal.org/docs/drupal-apis/javascript-api/ajax-forms#s-ajax-commands-ajaxresponse

pyrello:
> > Add hidden fields with selected values from block configuration?

> I'm not sure what this means.

I don't remember what I meant by this. Maybe it will come back to me when I test it.

richardbporter:
I removed the unique ID for now as the session prefilter might negate the need for multiple blocks on a page for the grad college. Ideally, we'd figure this out.

pyrello:
I actually was into the idea of multiple blocks, just because then they could be placed side-by-side, or in different sections with different help text.

richardbporter:
I think we can get there with a bit more time/work. 

- [x] Another thing we might to do here, if we're happy with the HeadlineHelper, is standardize how other code is accessing the headline styles. I know there are few different variations going on in uids_base.theme and/or uids_base/templates.

pyrello:
I can take a look at this later to see if there is anything I can figure out. If we have something working that allows showing multiple sessions via one block, we could go ahead with that for now.

richardbporter:
Just saw this which might be ~~a better~~ the proper solution: https://drupal.stackexchange.com/questions/228884/submitting-a-form-with-ajax-when-multiple-instances-of-the-same-form-are-on-a-p

I can take a look later.

richardbporter:
I think I goofed something up with the date styles somewhere.

bspeare:
@richardbporter I wonder if changing https://github.com/uiowa/uiowa/pull/2833/files#diff-ac61a5ddf0f7f8843e25b7582823013c5918026b5b2713183acb5df76a321eddR20 to `card` would fix the issue?

richardbporter:
> @richardbporter I wonder if changing https://github.com/uiowa/uiowa/pull/2833/files#diff-ac61a5ddf0f7f8843e25b7582823013c5918026b5b2713183acb5df76a321eddR20 to `card` would fix the issue?

Yes, thanks!

richardbporter:
https://github.com/uiowa/uiowa/issues/2714 pointed to https://registrar.uiowa.edu/dates-by-session which has exposed filters but https://www.grad.uiowa.edu/deadlines does not have any. If grad doesn't need exposed filters, maybe we should just remove them as that is what causes the issue with multiple blocks - the ajax does not work properly.

I think I misunderstood the block session options as combining them but I think what is desired is:

- Current session (Winter 2020)
- Current session plus one (Spring 2021)
- Current session plus two (Summer 2021)
- etc.

That way, multiple blocks can be placed on a page in groupings by session. Is the monthly grouping necessary?

This change might make the headline field less useful. It would be nice if we could auto-populate that with the selected session through #states or a less-elegant string replacement if someone uses `@session` or something.

The registrar and other sites needs both the exposed filters and combined session listing so I could split these blocks in two. That might simplify things.

Also, there are fields on the grad deadline listing that are not in MAUI including the iCal, Location and the Deadline link - although that goes to a page with minimal info. Some MAUI dates have webDescriptions with links in them.

richardbporter:
I may have got the AJAX working correctly actually.

richardbporter:
> Another thing we might to do here, if we're happy with the HeadlineHelper, is standardize how other code is accessing the headline styles. I know there are few different variations going on in uids_base.theme and/or uids_base/templates.

I replaced one instance but the others are doing all different things and may be difficult to standardize. For example, some classes being added are not used anywhere else.

pyrello:
> Is the monthly grouping necessary?

No.

> This change might make the headline field less useful. It would be nice if we could auto-populate that with the selected session through #states or a less-elegant string replacement if someone uses @session or something.

Maybe we could have language along the lines of "Override title", which could allow the user to set the title of the block to something other than the session label. It might also be nice to be able to hide the title too. Being able to select the heading size is potentially still important for the same reasons it's important in other situations. If the heading is shown, either as the auto-populated text, or as overridden text, then you are selecting the top-level heading size - let's say, for example `h2`, and subsequent headings (e.g. date headings [`h3`], then event titles [`h4`]) should follow that same hierarchy. To facilitate this, it might be useful to add a method to the HeadlineHelper class that takes in an argument like (+1, -1, +2, -2) and returns an appropriate `h*` value up or down the hierarchy (obviously stopping at the top or bottom).

> Also, there are fields on the grad deadline listing that are not in MAUI including the iCal, Location and the Deadline link - although that goes to a page with minimal info. Some MAUI dates have webDescriptions with links in them.

No need to worry about the extra fields. I was hoping we might be able to find a way to include those iCal links, but when I discussed this with Josh, he brought up the point that they are not currently set up as calendar events that stay up-to-date, so if a date changes and someone added it to their calendar, it will be wrong from that point forward. Apparently, there is a way to create a link like that that stays up-to-date with changes. Either way, this is a feature we can punt on until after launch.

> I replaced one instance but the others are doing all different things and may be difficult to standardize. For example, some classes being added are not used anywhere else.

It may take some analysis by our resident CSS wizards to make sure that these classes are all needed and used.

richardbporter:
I'm able to build a page like the following with the current implementation:
image

An override style to hide session badges would be useful although we want to exclude subsession badges from that.

pyrello:
That all looks good to me.

richardbporter:
Now with less badges:

Screen Shot 2021-01-15 at 1 05 10 PM

pyrello:
Awesome. I think this will meet our needs nicely.

pyrello:
Testing

pyrello:
There are a couple issues that need some follow-up in future issues, although I don't think they are deal breakers for getting this PR wrapped up:
1. When adding multiple blocks to a page and one of the blocks has a category set, but the session is exposed, changing the session resets the block to showing all results. This doesn't happen when there is only one block on the page.
2. Changing the exposed filters (and therefore submitting the form) causes a refocus event that move the browser window up towards the top of the page, which is jarring for the user.

The first item is a bit of an edge case, because it is an unlikely scenario to come up very often. The second item is mitigated if you are just using this to display a single block with exposed filters near the top of the page.

pyrello (approved):
Documented a couple issues that I observed, but I still think this is good enough to go out.

pyrello:
Sorry, looks like I jumped the gun on approving this.

richardbporter:
No problem. I think that should fix the refocus. The ajax issue might take longer. I will leave it up to to merge this if you need to move forward. A follow-up PR is fine with me!

richardbporter:
> When adding multiple blocks to a page and one of the blocks has a category set, but the session is exposed, changing the session resets the block to showing all results. This doesn't happen when there is only one block on the page.

I saw this earlier once but I'm having trouble reproducing it consistently.

richardbporter:
> > When adding multiple blocks to a page and one of the blocks has a category set, but the session is exposed, changing the session resets the block to showing all results. This doesn't happen when there is only one block on the page.
> 
> I saw this earlier once but I'm having trouble reproducing it consistently.

@joewhitsitt Any interest in trying to reproduce this?

joewhitsitt:
I can take a look at this, but I am not sure when yet.

pyrello (changes requested):
Per conversation with Josh from Registrars office, this update should just be turned on for Grad College for now, so we can gather feedback from their calendar team before rolling it out to the whole platform.

pyrello (approved):

pyrello:
I created a new issue for the bug I encountered: #2889. I think we should go ahead and merge this, since the bug won't affect the Grad College and, by request, we are limiting this to just the Grad College for now.

joewhitsitt:
Was able to replicate the issue. Commented over here: https://github.com/uiowa/uiowa/issues/2889#issuecomment-763651202

PR #2834, opened 2021-01-08
Grad College SROP updates

pyrello:
# Updates

* Created Mentor content type in Grad split and added/moved fields from the person content type.
* Created Scholar content type in Grad split and added/moved fields from the person content type. In a few instances, new fields were created to replace the old ones.
* Updated the scholar view to point to the Scholar content type.
* Created layout templates for both of the above content types to look similar to the person content type.
* Created/updated templates to get the content types looking good.
* Added path aliases for both of the new content types.
* Added sitemap entries for new content types.
* Created a `grad` module for site-specific customizations (vs. customizations that might be added to multiple Grad College sites). Moved relevant code here.
* Turned the new module on in the grad site split.
* Removed fields from person content type on the Grad split that are no longer needed.
* Bug - Fixed an issue that was introduced in #2875 where the token manipulation wasn't scoped to nodes that contain that field.
* Some updates to `person.scss`

# How to test

* `drush @grad.local uli node/add/mentor` And add a Mentor.
* View the mentor page and check to see that it looks similar to a person page (e.g. https://demo.sitenow.uiowa.edu/people/herky-hawk)
* Visit https://grad.local.drupal.uiowa.edu/node/add/scholar and add a Scholar.
* View the Scholar page and check to see that it looks similar to a person page (see previous Herky link).
* You'll have already checked the bugfix above when you are successfully able to view the mentor and scholar pages.

**Visual Test for admissions:**
Updated person card templates to clean up content type specific css in `person.scss`
- Verify that  https://admissions.local.drupal.uiowa.edu/counselors looks like https://admissions.prod.drupal.uiowa.edu/counselors
- Verify that  https://admissions.local.drupal.uiowa.edu/student-profiles looks like https://admissions.prod.drupal.uiowa.edu/student-profiles
- Verify that https://admissions.local.drupal.uiowa.edu/people/amanda-cvitko-quinteiro looks like https://admissions.prod.drupal.uiowa.edu/people/amanda-cvitko-quinteiro


pyrello:
One additional change on this. The mentor field is currently implemented as a link, but I have yet to find any cases where the link is used in the current implementation. So, lets change the mentor field to a plain text field.

bspeare:
Pulling this down to test.

bspeare (approved):
Tested all conditions locally and made a few visual updates.

PR #2835, opened 2021-01-08
Config ignore fix

joewhitsitt:
Let's try it again. Builds error on a config_ignore diff. This force imports the config first so that there is no diff later.

Related convo: https://iowaweb.slack.com/archives/C8SM82ZFZ/p1610130577315200

# To Test

Following https://github.com/uiowa/uiowa#updating-dependencies

For any website plus HR try some of the following in an attempt to get a config export change for config_ignore.settings:

`blt ds --site=sitename`, login and save the config ignore settings without changing anything (/admin/config/development/configuration/ignore), `drush @sitename.local cex`

next try and do a direct sql-sync:

`drush @sitename.local sql-drop && drush @sitename.local cr && drush sql-sync @sitename.prod @sitename.local && drush @sitename.local updb` login and save the config ignore settings without changing anything (/admin/config/development/configuration/ignore), `drush @sitename.local cex`

Check HR to make sure uiowa_bar settings are still intact.

You may get diffs because there is new config that is not yet in PROD, but config_ignore.settings should not be one of them and `enable_export_filtering: true` should be what you get regardless of what you throw at it.

### DIFF
--- a/config/default/config_ignore.settings.yml
+++ b/config/default/config_ignore.settings.yml
@@ -28,6 +28,6 @@ ignored_config_entities:
   - views.view.articles
   - views.view.people
   - 'webform.webform.*'
-enable_export_filtering: '1'
+enable_export_filtering: true
 _core:
   default_config_hash: v7E8C8wJWeAW2BGohMNY1tZSBc4bexM6O62tGxecTfE
--- a/config/hr.uiowa.edu/config_ignore.settings.yml
+++ b/config/hr.uiowa.edu/config_ignore.settings.yml
@@ -3,6 +3,7 @@ ignored_config_entities:
   - 'block.block.main_navigation__superfish__horizontal:status'
   - 'block.block.main_navigation__superfish__toggle:status'
   - 'block.block.main_navigation__superfish__mega:status'
+  - 'config_split.config_split.apr:status'
   - 'config_split.config_split.areas_of_study:status'
   - 'config_split.config_split.collegiate:status'
   - 'config_split.config_split.person_extended:status'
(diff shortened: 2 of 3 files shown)


joewhitsitt:
Crap. It removed HR uiowa_bar settings. Working on a fix

pyrello (approved):
I tested HR via `blt ds`. I got lost in the second set of instructions, which seems to result in a lot of conflicts between the database (which reflects prod settings) and local config. That being said, I'm comfortable giving this the green light without additional testing. The updates hook runs as expected and I don't believe that this should break anything. The worst that will happen is that the build will fail again.

richardbporter:
Merging into dev and syncing a DB would be quicker than creating a release if we want to test in AC.

pyrello:
I merged into `develop` and `sql:sync`'ed the default DB from prod to local

joewhitsitt:
>Acquia CloudAPP Today at 9:32 AM
An updated deployment has been made to uiowa.dev using branch develop-build as develop-build.

Does this mean what I think it means? 🤞

pyrello:
I think it does. Maybe still try saving the config and see if there is a config diff?

joewhitsitt:
Saved config_ignore via UI on uiowa.dev. Then ran:
```
drush @home.dev cst
 [notice] No differences between DB and sync directory.
```

uiowa02.dev failed because of:
```
  Configuration <em class="placeholder">field.field.node.student_profile.fiel  
  d_student_profile_major</em> depends on the <em class="placeholder">node.ty  
  pe.area_of_study</em> configuration that will not exist after import.
```

PR #2839, opened 2021-01-11
SROP scholar migration

### LINKED ISSUE #2559: Create a migration for Grad College SROP Scholars

cory-skeers:
Resolves #2559 

# How to test

* `blt ds --site=grad.uiowa.edu` - Sync grad.uiowa.edu locally.
* `drush @grad.local en grad_migrate` - Enable the grad_migrate module for that site.
* `drush @grad.local mim d7_grad_srop_scholar` run the migration. Can append `--limit=#` to import only `#` of entries for testing, and use `drush @grad.local mr d7_grad_srop_scholar` to undo the migration and remove imported entries prior to running again.


pyrello:
Making a note of this so I don't forget: The mentor field on the Scholar content type is transitioning from a link field to a plain text field. If there is a URL, it can just be ignored. We'll just use the link text to fill that field.

pyrello:
@cory-skeers I attempted to update this branch with changes I made in #2527. However, I think, due to the fact that you had removed some files from this branch that were in the other branch, the merge resulted in the wrong files getting deleted. I attempted to go back through the commits to restore what was here, but I'm not sure that I did that correctly.

At this point, I might just be inclined to move whatever files from this were different from the other one back to #2527 to avoid anymore complications around some changes I made in that other PR.

cory-skeers:
> At this point, I might just be inclined to move whatever files from this were different from the other one back to #2527 to avoid anymore complications around some changes I made in that other PR.

I think that's a good plan. Originally planned on #2527 being merged in shortly, but on the timeframe now, I don't think there's any need for separate PRs.

It should only need the YAML and plugin files copied over (and the same for the mentor). I can work on that post-meetings if you haven't gotten to it before then.

pyrello:
Moving back over to #2527

Issue #2838, opened 2021-01-11
SSL Cert - se.iowastem.org on uiowa02


joewhitsitt:

richardbporter:
Not possible. #2522

joewhitsitt:
DCV validated. Supposedly. Moving forward to find out.

joewhitsitt:
Done

PR #2840, opened 2021-01-11
Initialize se.iowastem.org multisite on uiowa02

joewhitsitt:
SSL Issue: https://github.com/uiowa/uiowa/issues/2838


pyrello (approved):
Reviewed, not tested

PR #2841, opened 2021-01-11
SROP mentor migration

### LINKED ISSUE #2558: Create migration for Grad College SROP mentors

cory-skeers:
Resolves #2558 

# How to test

`blt ds --site=grad.uiowa.edu` - Sync grad.uiowa.edu locally.
`drush @grad.local en grad_migrate` - Enable the grad_migrate module for that site.
`drush @grad.local cim --source ../config/features/grad_migrate --partial` - Register and import the grad_migrate feature split.
`drush @grad.local mim d7_grad_srop_mentor` run the migration. Can append `--limit=#` to import only `#` of entries for testing, and use `drush @grad.local mr d7_grad_srop_mentor` to undo the migration and remove imported entries prior to running again.


pyrello:
Moving this back over to #2527

PR #2843, opened 2021-01-11
Delete accidentally exported APR settings from split.

richardbporter:
This must've been exported by mistake, maybe before the ignore rule was active.


joewhitsitt (dismissed):
reviewed, not tested

pyrello:
Um... wait... don't you want this file to set defaults?

richardbporter:
Hm. Good point. The config form sets all defaults so it works without it but yes, providing one seems like a good idea. I'm looking into it.

richardbporter:
I can't get the config to install when enabling the split and importing. I swear this worked before.

richardbporter:
I guess not. I reverted the last config_ignore update locally to see if something changed and it dit not make a difference. I guess ignored config in a split is not imported initially.

richardbporter:
There is no functional change here. Mostly just want to get that dummy data out of the config.

pyrello:
Thanks for looking into it. I guess the instances where I saw this happening previously were for default config, so that all makes sense. Although it is weird behavior in my opinion that it works one way for the default and different for a split. Might be worth opening an issue for Config Split 2.x.

joewhitsitt (approved):

PR #2844, opened 2021-01-11
Set destination param on login link

### LINKED ISSUE #2825: As an authenticated user, I want to be redirected to my destination after logging in for ease of use.
# Story

As an authenticated user, I want to be redirected to my destination after logging in for ease of use. 

Right now, if people try to access protected pages but they are not logged in, they are not redirected back to those pages after authenticating. This is annoying especially when asking viewers/editors/etc. to review unpublished pages in emails, Slack, etc.

# Proposed solution

Investigate contrib space to see if there is a plug-and-play solution although this could be as easy as setting the destination parameter on the admin login footer link href. The default login destination is the user page which isn't very useful at all. I don't see a problem with redirecting to the front page/news article/etc. or wherever someone clicked the admin link from.

richardbporter:
Resolves #2825 

Seeing if a quick fix work as expected. Deploying to dev.


joewhitsitt:
Did this work in DEV?

pyrello:
Yes. This works in dev. It works with unpublished, but also for admin links too, which is nice! Nice work @richardbporter

joewhitsitt (approved):
reviewed, not tested #FailFaster

Issue #2021, opened 2020-08-03
Improve search results page a11y


richardbporter:
# Problem 

The search results page has misleading colors that creates confusion over which link to click on. 

> I was using the search function on some of the sites that are using the new UIDS and noticed something odd. 
When you do any type of search, (take this https://uiowa.edu/google-search?search=search), search results are presented in a title, line break, URL, line break description format. However, the URL, despite being highlighted in a different color, isn't a hyperlink—only the page title is, which happens to be in black text with no underline.
This felt like an accessibility red flag and just a counterintuitive user experience. Just wanted to share this feedback.

<img width="870" alt="Screen Shot 2020-12-21 at 6 14 41 AM" src="https://user-images.githubusercontent.com/1036433/102776365-3c205d80-4354-11eb-9a5f-ab26a7f132dd.png">

# Proposed Solution

- [x] Adjust https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/search-results.scss and switch color treatment between Title and URL and add hover underline to the title as well. 

<img width="908" alt="Screen Shot 2020-12-21 at 6 15 21 AM" src="https://user-images.githubusercontent.com/1036433/102776452-5ce8b300-4354-11eb-87ef-baff0af5c3c4.png">

joewhitsitt:
This got merged.

PR #2847, opened 2021-01-12
Added hover underline and link color to title and black color to URL

ethanstjohn:
Solves https://github.com/uiowa/uiowa/issues/2021

The goal was to add functionality in the CSS of search results. Changing the title color to a link color and give it hover underline, while the URL which had link color should now be black to keep the emphasis off of the URL and onto the link.


bspeare:
Pulling down to test.

docroot/themes/custom/uids_base/scss/components/search-results.scss
```diff
@@ -79,7 +82,7 @@
   padding-right: 0;
 }
 #search-results .gsc-url-top .gs-visibleUrl, #search-results .gsc-url-bottom .gs-visibleUrl {
-  color: $link-color;
+  color: $blk;
```
bspeare: @ethanstjohn Can you change this to `$brand-cool-gray`?  I think having this as gray will help further distinguish that this url is not active.

bspeare (changes requested):

bspeare (approved):

bspeare:
Updated the url in the results to be `brand-cool-gray`:

Screen Shot 2021-01-13 at 9 17 51 AM

PR #2868, opened 2021-01-14
hotfix - events card styling in layout columns

joewhitsitt:
For layout builder columns other than one-column, circle image styling was applied. This is either hardcoded (event) or is now controlled by layout builder styles.

# Test

- event and events blocks in v3 layout builder.
- event should stay circle regardless of column
- change the image to square or widescreen in an events block and move between 1-4 column layouts. the images should stay as you configured.

of course at smaller widths, it switches styles entirely which is expected.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/components/event.scss
+++ b/docroot/themes/custom/uids_base/scss/components/event.scss
@@ -56,28 +56,6 @@
     }
   }
 
-  .layout--twocol .layout__spacing_container & {
-    @media screen and (min-width: 40em) {
-      .card__media img {
-        border-radius: 50%;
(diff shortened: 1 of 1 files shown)


bspeare:
Pulling down to test

bspeare (approved):
Tested on sandbox.local event list pages.   Do we still need to target the `uiowa-card-event ` class?  Is that still being used?

joewhitsitt:
it is still on the event block markup

https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-event.html.twig#L34

Issue #2474, opened 2020-11-05
Epic: Prepare platform for upgrade to Drupal 9


pyrello:
# Motivation

To stay current with the most recent version of Drupal

# Proposed solution

The following updates will allow us to update to Drupal 9.0.x. Some issues may require checking the warnings generated in the Upgrade Status report available on a site when viewing it locally.

- [x] uiowa/uiowa_bar#59: Update uiowa_bar module to be D9 compatible.
- [x] ~~#2592 Bump `drupal/big_pipe_sessionless:^2.0`~~ - This has to be done at the same time as moving to D9
- [x] #2593 Bump `drupal/components:^2.0@beta`
- [x] ~~#2594 Apply patch from https://www.drupal.org/project/config_ignore/issues/3150462 to make `drupal/config_ignore` compatible.~~ - This is no longer necessary because of #2772 
- [x] #2455 Bump `drupal/devel:^4.0`
- [x] ~~Resolve issues with drupal/field_group, possibly by applying the patch here: https://www.drupal.org/project/field_group/issues/3168213~~ - Not necessary, the library will get updated via D9 upgrade.
- [x] #2872 Bump `composer update drupal/fragments`
- [x] ~~Determine whether additional action is necessary to resolve `drupal/google_analytics` warnings in Upgrade Status~~ - Not necessary as all the warnings relate to an extension being uninstalled that is only required for tests.
- [x] Bump `drupal/honeypot:^2.0`
- [x] #2595 Bump `drupal/imagecache_external:^3.0.`
- [x] Determine whether additional action is necessary to resolve `drupal/media_library_theme_reset` warnings in Upgrade Status
- [x] Resolve `drupal/menu_block` compatibility issues - needs more research about whether patches are compatible with stable version or whether current dev commit hash we are pinned to works
- [x] Resolve `drupal/metatag` compatibility issues - needs more research about whether patch is compatible with stable version or we can roll with the dev version
- [x] ~~Determine whether additional action is necessary to resolve `drupal/paragraphs` warnings in Upgrade Status~~ - Not necessary as all warnings are related to tests or modules that are not installed.
- [x] Investigate purge warnings in Upgrade Status
- [x] ~~Determine whether additional action is necessary to resolve `drupal/search_api` warnings in Upgrade Status~~ - Not necessary as all warnings are related to tests.
- [x] Determine whether additional action is necessary to resolve `drupal/simple_sitemap` warning in Upgrade Status
[...]

PR #2872, opened 2021-01-14
Grad College and usability updates

pyrello:
# Updates
## Grad College
* Pulled in functionality to allow defining a main list of programs for the Grad college and to filter it for allowed values lists from #2527.
* Renamed field to better align with the allowed values.
* Update Grad programs and certificates view displays.
* Updated `drupal/fragments` to 1.0.0-beta3, which is needed for #2474. Seemed like a good time to do this, since this modules usage is limited to the GC split.
* Switched scholarship content type to use a Layout Builder template.

## Platform
* Added a menu route to 'Add content' that can be accessed from the admin toolbar without going to the Content overview page. Only accessible when the menu is showing on the side.
  Screen Shot 2021-01-14 at 10 25 52 PM
* Added `field--title--node--scholarship--default.html.twig` template.

# How to test

* `drush @grad.local uli`
* Note no problems with import.
* Check that the menu shows as pictured above.

## Check Admissions scholarship titles are displaying correctly
* `drush @grad.local uli scholarship/advantage-iowa-award` and check that heading looks like https://admissions.prod.drupal.uiowa.edu/scholarship/advantage-iowa-award


bspeare:
Pulling down to test

bspeare (approved):
New "Add content" link is in place and the Scholarship titles look correct.

pyrello:
Thanks!

PR #2677, opened 2020-12-08
Feature:  Default field styles

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/2648. 

# How to test

- Temporarily remove styles from https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/content/node--type--person.scss
- `yarn workspace uids_base gulp --development`
- Test on people pages for https://immuno.grad.uiowa.edu/people, https://github.com/uiowa/uiowa/pull/2621, and any other pages that have a lot of field and label variations.

PR #2875, opened 2021-01-15
Teaser token override, Grad body/bio teasers

### LINKED ISSUE #2819: Override field_teaser token output to allow automatic body field summary
# Story

As a site owner I want to have a meta description generated from body field (if it is used) so that I don't have to write custom meta descriptions for every single page.

# Proposed solution
* Create an implementation of `hook_tokens` in `uiowa_core` to override the value of `field_teaser`.
* The logic will check to see if there is a value for the `field_teaser` field and if so, display that. If not, it will check the body field for a value and create a summary for that. We'll basically be adapting the functionality in core for the body summary token: https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/node/node.tokens.inc#L129-164
* The Brand site has a code example of setting a token value through this hook, although in this case we would want to modify an existing token: https://github.com/uiowa/uiowa/blob/4d80082ca4907412b674799815a1c98a744ac086/docroot/sites/brand.uiowa.edu/modules/brand_core/brand_core.module#L328-L352
* Check for `$node->hasField('body')` before running the rest of the logic to prevent it from affecting V2 sites.

joewhitsitt:
Resolves #2819

For page, person, article if a field_teaser value is not provided and there is a body or person_bio field (sorry v2 page), then the stripped/trimmed 300 characters of the field are provided instead. This makes quick content populating easy to establish SEO for content until more refined teasers/meta descriptions can be created. This does not override field_teaser on teaser view modes.

For grad teaser view modes for page, person, article, if a field_teaser value is not provided and there is a body or person_bio field then a smart_trim 300 limit, stripped tags and ellipses suffix is provided. Though the field is not field_teaser, [thanks to recent general styling](https://github.com/uiowa/uiowa/pull/2677), this text shows similar to field_teaser text.

# Test

For sites that have the default content types and field_teaser, try entering teaser values to confirm the meta description is populated. Remove those values and try instead using body or bio content and see that a trimmed version shows as the meta description. This value will also populate other meta values based on the content type.

For grad, you will see body/bio text on /people for many individuals. If articles had the same setup, you would see that as well.


pyrello:
Pulled down for testing.

joewhitsitt:
Expanding scope to provide trimmed body/bio to teaser view modes on Grad with the idea that we will start the conversation of adding this to the default. We currently don't have a "hide descriptions" style option on SiteNow Articles/People view pages.

joewhitsitt:
@pyrello the Grad teaser portion is ready to test a your convenience.

pyrello:
I am not seeing this behavior when I run `blt ds --site=grad.uiowa.edu` on the people page, or on the article block list when I place it on a page. Am I doing something wrong? It sounds like the second case may not be covered.

pyrello:
Nevermind... I forgot to pull the changes.

pyrello (approved):
I tested this for meta descriptions on all default content types. Also for the display of the summary text on article and people teasers.

Nice work! Thanks!

joewhitsitt:
Updated documentation https://sitenow.uiowa.edu/node/81/latest and will publish when this rolls out.

PR #2903, opened 2021-01-21
Feature: Grad student profile

pyrello:
After working on #2827 (along with a conversation about Admissions today), I have realized that the content type I was creating was not ideal for some of what I am trying to build. This attempts to meet that need in a way that would be compatible with wider usage eventually.

Long term, I would like to see the differences between this and the version of student profiles in Admissions worked out and moved into some mutually compatible state. In the meantime, this was easier for me to get something done more quickly and I need to get this into the hands of people migrating content ASAP.

Adds a student profile content type, with the config currently living in the Grad split and a module to add templates and control other logic. Adds path alias and metatag defaults to be understood as a person schema object.

# How to test

* `drush @grad.local uli node/add/student_profile`
* Add student profile and observe it is similar in appearance to person page.
* Review code changes to check for instances where updates will affect the default platform.


bspeare:
Pulling this down to test.

bspeare (approved):
I've tested this locally and it works!

PR #2904, opened 2021-01-21
Create kang.lab.uiowa.edu and licensing.uiowa.edu, plus some extras.

richardbporter:
Since `blt uiowa:multisite:install` is most useful when more than one site needs installed, it can be quicker to run `blt drupal:install --site mysite.com` when only doing one site. I did this last week and noticed the post-install tasks did not run since they are included in the `uiowa:multisite:install` command. This PR moves them to a hook so it runs in both cases.

Also included AH application name in remote `drush status` output. This is a bit easier than parsing it out from the files path. 

# How to test

- Drop two site DB's locally.
- Run `blt uiowa:multisite:install` and double check the post-install tasks run - site name, requester, split if set.
- Drop one site DB locally.
- Run `blt drupal:install --site sitename.com` and double check again.
- Run `drush status` against some site in dev and check the output for Application.


richardbporter:
Just noting that licensing.uiowa.edu will need SSL coverage.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
nice. though I have to increase the size of my terminal window because i missed the application output the first time.

tested on webcommunity.sites which had a requester unlike international/law which were done way back when.

richardbporter:
Thanks. Yea a lot of info is cut off for me when I run terminals horizontally. We could move it to the bottom?

joewhitsitt:
>We could move it to the bottom?

Not necessary. Smug comment on my part.

PR #2907, opened 2021-01-22
Fix MAUI block AJAX bug with multiple forms.

### LINKED ISSUE #2889: Intermittent issue with multiple MAUI session date blocks under certain conditions.
# Problem

Following up on issue reported in #2833.
 
When adding multiple blocks MAUI Dates by Session blocks to a page, when one of the blocks has a category set, but the session is exposed, changing the session resets the block to showing all results. This doesn't happen when there is only one block on the page.

# Steps to reproduce
1. Add a  Dates by Session block to a page via LB. In the block configuration, leave both filter exposed.
2. Add a second Dates by Session block to the same page. In the block configuration, set the category to "Graduate Students" and leave the session filter exposed.
3. Save the layout.
4. Change the session in the second block and observe whether it is sticking to the correct category.

# Proposed solution

TBD

richardbporter:
Resolves #2889 

### How to test
Reproduce the bug on main according to #2889 using grad.uiowa.edu locally. Check out this branch, maybe rebuild the cache and check if the behavior is as expected.


richardbporter:
For reference: https://www.drupal.org/project/drupal/issues/2821852

pyrello:
Nice find!

joewhitsitt:
pulling down

joewhitsitt (approved):
Recreated from https://github.com/uiowa/uiowa/issues/2889#issuecomment-763651202 and worked.

For those tests, do those hardcoded sessions fall off over time? Will it have to be updated as time goes by?

richardbporter:
Nah its just some fake data that doesn't really need to change over time. Like this is just checking to see if the string is replaced: https://github.com/uiowa/uiowa/pull/2907/files#diff-b3f21792e994efe25da1d969ef5daa53fb7d9d09ed5088dfa4e9eba59a0c4dc1R111

Issue #2259, opened 2020-09-29
V3 version of V2's featured content


joewhitsitt:
## Problem

Several v2 websites are using Featured Content to reference existing content in a manually curated way and currently extends beyond article and person to include basic page as well. This type of functionality doesn't exist in v3.

This is commonly used to highlight a featured article, a manual sort of people, basic page content quick/related links.

Something to port over to will be necessary for v2 migration. https://github.com/uiowa/uiowa/issues/1779, https://github.com/uiowa/uiowa/issues/1803

## Proposed Solution

Recreate the featured content block in layout builder working around the current issue with the v2 version: https://github.com/uiowa/uiowa/issues/414

- [x] block type
- [x] headline/children
- [x] Permissions splits
- [x] entity view display splits, admissions entity print issue
- [x] styles

pyrello:
Are we okay with not considering this a blocker to V3 GA?

joewhitsitt:
Not a blocker.

pyrello:
As a first slice through this, we are going to create:
* A new custom block type
* with a multi-value entity reference field that can reference articles and people
* that relies on existing teaser displays for styling children
* with a Headline field
* that affects the h size of the children in the collection
* with permissions for Editors and Webmasters
* and allowed to be placed in Layout Builder

joewhitsitt:
Thought about providing a more link similar to the other "collection" blocks, but not necessarily as part of the current MMF

richardbporter:
Headline/children field is set up but we don't know of a way to get at the relationship between the block and node.

joewhitsitt:
Before this issue is closed, a follow-up task for page support should be created

pyrello:
I'm not sure we should close this issue without articulating all the things that we didn't do for the first pass at this feature.

richardbporter:
In honor of a11y day, I gave the headline issue a shot in #3552.

pyrello:
Follow up issue: #3567

PR #2919, opened 2021-01-23
Grad 3MT profile + list page updates

pyrello:
# Updates
* Added a "Honors/distinctions" field to the Grad Student Profile content type for displaying text like "Winner, People's Choice" or "Ballard and Seashore Fellowship Recipient."
* Added templates to display the above field using the UIDS Badge component.
* Add a student profile teaser template which uses the card component template + inner grid classes to accomplish a 67/33 or 33/67 layout.
* Updates to `grid.scss` that allow creating two-column grids of the above proportions.
* Added a Content list block type that allows for selecting arbitrary node content and ordering it.
* Added a Student profiles view block that can be added via LB.

# How to test
*These test instructions are a work in progress*
* `drush @grad.local uli node/add/student_profile` and create a profile, making sure to add an "Honor/Distinction."
*  https://grad.local.drupal.uiowa.edu/node/add/page and create a test page.
*  Click "Layout" tab and add a Content List block. Reference the profile you just created. Save the layout.


pyrello:
Here is what I am stuck on right now:
* The ordering or participants on the [3MT page](https://www.grad.uiowa.edu/2020-3mt-showcase) is arbitrary. Also, it will change between the finalists being selected and the winners and honorable mentions being chosen. The eventual order places the winners at the top and all the finalists below. I created the content list block type as a way to be able to control the order.
* It's not that important, but in the past we have added alternating background colors to be able to make each row stand out a bit more. The downside to the above approach is that the node teaser template where I am implementing the card knows nothing of its delta position within the entity reference field. This could probably be made available via preprocessing or by wrapping each field instance in a div with a background class.
* I'm wondering whether it makes sense to create a field specifically for the title of the presentation. I'm not sure what this would represent on the generalized student profile content type. Its kind of similar to a field that I used in the SROP content types called "Bio title" and it was just way to add a headline to the top of the bio. In the SROP instances this is used to give a headline to research projects. Not sure if this would be generally useful or something that would stay with the GC implementation.

joewhitsitt:
Is this the issue you referenced in standup that we should give thoughts on?

Are you essentially creating https://github.com/uiowa/uiowa/issues/2259 with this content list block type? If so, and I say this not having tested this PR, I'd use v2 as a guiding reference so that if this was pulled into default there isn't a large difference to account for with migration.

Not sure how to mix the arbitrary sort with the winners/finalists sort without causing UX confusion.

For alternating colors I could see that as a layout builder styles option that could be selected instead of a single background color.

I haven't given much thought yet to a "generalized student profile content type" so I'd just do what you need to do for Grad for now.

pyrello:
@joewhitsitt Thanks for your feedback. 

I'm leaning towards the following options, if I can get everything working together:
* Skip the content list block for now
* Use a view that is customized for GC purposes, since I'm not sure how broadly applicable this is.
* Create a view mode other than teaser for presenting the "large card" version of the content.
* Implement a weight field on the student profile to allow ordering the top items.
* Implement a "Honor/distinction year" field to allow for grouping people into cohorts.
* Adjust the view to allow for a contextual argument.

bspeare:
@pyrello are you still using https://github.com/uiowa/uiowa/blob/53f3eecba7cab876e550ff80a41045eca30b8295/docroot/modules/custom/grad_custom/grad_student_profile/templates/node--person--card.html.twig and https://github.com/uiowa/uiowa/blob/53f3eecba7cab876e550ff80a41045eca30b8295/docroot/modules/custom/grad_custom/grad_student_profile/templates/node--student-profile--teaser.html.twig?

bspeare:
Testing and reviewing now.

config/grad.uiowa.edu/views.view.srop_scholars.yml
```diff
@@ -1,6 +1,6 @@
 uuid: 7138678e-9e96-4ebc-b20b-578be83126d5
 langcode: en
-status: true
+status: false
```
bspeare: Did you mean to set this back to false after https://github.com/uiowa/uiowa/pull/2923?

pyrello: Nope

bspeare: ok, I'll fix it.

bspeare (approved):
I tested this locally and it works.

PR #2946, opened 2021-01-27
Brand - Stale lockup info issue

### LINKED ISSUE #2918: Brand - Inconsistent lockup download behavior
# Problem
>I've had happen a couple of times. I will make an edit to a lockup submission and then save/approve it, but then when the user accesses the lockup for download, after receiving the approval email, they are getting the version they submitted that doesn't reflect the changes I've made. It happened yesterday with this [Engineering-Hanson Center](https://brand.uiowa.edu/lockup/2991) lockup. It's only happened one other time that I can recall but I didn't think to document it in the rush to fix it.

>When I go to the lockup I get broken images where the preview would be and no download button option. So then I go in to edit and do a resave and that resolves it.

# Steps to reproduce

- ??

# Proposed solution

- Go through the lockup workflow for new and existing (with reroute emails on and routed to your email).
- See if node full images are broken and edits to existing are what is downloaded.

# Notes

- My initial suspicion is weird caching with browser and/or lazy loading for the node full images.
- Purge is supposed to happen on node update when the state is published. Make sure that is firing correctly (not local).

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/2918

Big shout out to @richardbporter  for figuring it out.

No testing necessary. Just a look over.


docroot/sites/brand.uiowa.edu/modules/brand_core/brand_core.module
```diff
@@ -394,7 +394,7 @@ function brand_core_node_update(Drupal\Core\Entity\EntityInterface $entity) {
         \Drupal::service('cache.render')->invalidateAll();
 
         // If available (not Local), try to clear the varnish cache for the files.
-        if (\Drupal::moduleHandler()->moduleExists('purge')) {
+        if (\Drupal::moduleHandler()->moduleExists('acquia_purge')) {
```
richardbporter, line 397: FYI - I think Purge is actually the right module to check for but unfortunately we don't have a local Varnish instance running to actually purge against. Ideally, we would because it'd be useful in situations like this. So the switch to acquia_purge is just a workaround I guess for now.

richardbporter (approved):

docroot/sites/brand.uiowa.edu/modules/brand_core/brand_core.module
joewhitsitt: unless we move from acquia though, we will be using both so i think this is good for a while.

PR #2951, opened 2021-01-27
Bug Fix: view person_card display content type person

### LINKED ISSUE #2845: Person_card view block does not display if person content type is initially created as in "draft" state
# Problem
The profile image, email, degree, or other contact information in the `person_card` view does not display if a Person is created and set to the "Draft" state. 

Screen Shot 2021-01-12 at 8 18 39 AM

# Steps to reproduce
Create a new Person with an image and set it to the "Draft" state. 

# Proposed solution
- I suspect this view just needs the published status filter criteria removed, saved, exported. admin/structure/views/view/person_card

OR

- Investigate permissions or caching that is blocking the `person_card` view from displaying.

ethanstjohn:
Potentially fixes: #2845 

Description: Removed person_card filter criteria for published content.

Test: Test by creating a new person in the "draft" state. It should display all expected and entered information.

### DIFF
--- a/config/default/views.view.person_card.yml
+++ b/config/default/views.view.person_card.yml
@@ -63,17 +63,13 @@ display:
           offset: 0
       style:
         type: default
-        options:
-          grouping: {  }
-          row_class: ''
-          default_row_class: true
       row:
         type: fields
         options:
-          default_field_elements: true
+          default_field_elements: false
           inline: {  }
           separator: ''
-          hide_empty: false
+          hide_empty: true
       fields:
         field_image:
           id: field_image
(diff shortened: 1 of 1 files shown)


config/default/views.view.person_card.yml
```diff
@@ -63,13 +63,17 @@ display:
           offset: 0
       style:
         type: default
```
joewhitsitt: These lines should have not changed. Only 518-530 should be removed:
```
-        status:
-          value: '1'
-          table: node_field_data
-          field: status
-          plugin_id: boolean
-          entity_type: node
-          entity_field: status
-          id: status
-          expose:
-            operator: ''
-            operator_limit_selection: false
-            operator_list: {  }
-          group: 1
```

joewhitsitt (approved):

PR #2952, opened 2021-01-27
Removed AoS requirements field and convert Areas of Study block to plugin class

### LINKED ISSUE #2877: Recreate AoS search block as Block plugin class instead of block content
# Motivation

Block content all belongs to the block category of the same name automatically and I'm not sure there is a way to change it. The way that I built the Areas of Study search block, it will always show at the top of the direct add button, even though it is only likely to ever be used a couple times on the Admissions site at most. Using [recent](https://github.com/uiowa/uiowa/blob/bc8771e025d97a91be98d63276900d1929ec2aac/docroot/profiles/custom/sitenow/src/Plugin/Block/SiteNowBlock.php) [examples](https://github.com/uiowa/uiowa/blob/865fa2cf39597d2ac3c4e48dadd4fc60f124d8dc/docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php) of blocks created via Block plugin class, we should convert this block to that methodology to allow us to classify it under the "Site custom" block category created in #2773.

# Proposed solution

 * [ ] Using example cited above, create a new AreasOfStudySearchBlock class to provide this block.
 * [ ] Delete the existing block from Admissions homepage.
 * [ ] Remove config and code related to Areas of Study search custom block.

### LINKED ISSUE #2942: Admissions - Add entity reference fields for admissions requirements and application process
# Proposed solution

* [x] Confirm with Scott how he wants to handle entering "Application Process" content.  Do they want to enter as pages similar to "Admission Requirements?"
* [ ] ~~If using a dedicated content type for "Application Process," add the content type. The fields/layout/etc. can be added in a follow-up task. If that is necessary, please create a the follow-up issue.~~
* [x] Remove the Admission Requirements content type.
* [x] Create paragraphs bundle to house the fields
  * Title/label
  * Intro
  * Requirement label
  * Requirement entity reference (page)
  * Process label
  * Process entity reference (page)
* [x] Create three separate paragraphs fields on AoS that allow one item from the bundle above.
* [x] Bring the local admissions requirements field into the same gray section to connect the "four" together.
* [x] Output in card template for follow up work in https://github.com/uiowa/uiowa/issues/2934

<details>
<summary>Original details</summary>
```
- [ ] Add 3 entity reference fields that link to pages, with the following names:
  - Requirements: First year
  - Requirements: Transfer
  - Requirements: International
- [ ] Add 3 entity reference fields that link to whatever content type is determined to be used for "Application Process" content based on previous conversation, with the following names:
  - Application process: First year
  - Application process: Transfer
  - Application process: International
```
</details>

pyrello:
# Updates
* Resolves #2877 

# How to test

* `drush @admissions.local uli /` and click 'Layout' tab.
* Add block, select "More...", and choose "Areas of Study Search"
* Save the page.
* Test that the search block takes you to the Areas of Study page.


pyrello:
@richardbporter I'm running into an error where it says the number of arguments being passed to the constructor function of the block class are not correct, but I compared this to the class you created for academic dates and I can't figure out what I am doing wrong.

docroot/modules/custom/uiowa_area_of_study/src/Plugin/Block/AreasOfStudySearchBlock.php
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Provides the Areas of Study Search block.
+ *
+ * @Block(
+ *   id = "uiowa_area_of_study_search",
+ *   admin_label = @Translation("Areas of Study Search"),
+ *   category = @Translation("Site custom")
+ * )
+ */
+class AreasOfStudySearchBlock extends BlockBase {
```
richardbporter: The class just needs to implement ContainerFactoryPluginInterface for the `create()` method to run, passing the fourth parameter to the constructor.

pyrello: Thank you!

pyrello:
I was originally going to complete #2942, but after discussing it with Scott, I think we need to regroup on that. I'll add more notes at that issue.

richardbporter (approved):

Issue #115, opened 2018-05-14
Add page number to default page title metatag.


briand44:
This is a good practice so Google doesn't count multiple pages as duplicate content. Currently this token has not been ported to D8. https://www.drupal.org/project/metatag/issues/2926852

joewhitsitt:
Removing blocked. We should at least chime in on the issue or investigate whether or not the latest comment meets our MVP needs.

richardbporter:
We can minimize the SEO issue by utilizing AJAX pagers whenever possible. In fact, the only use case we have for this right now is with articles and that view is already configured to use AJAX. It does not seem to be working, however.

PR #2954, opened 2021-01-27
Template consolidation and coding standards

joewhitsitt:
- Coding standard fixes
- Created generic unordered list field template to consolidate.
- Deleted person_education bulleted list template because it is not being used for person_card (full node) styles. Not aware of a grad or admissions content type or person variant using it.


docroot/modules/custom/sitenow_seo/sitenow_seo.module
```diff
@@ -22,7 +22,7 @@ function sitenow_seo_form_google_analytics_admin_settings_alter(&$form, FormStat
 /**
  * Implements hook_metatags_alter().
  */
-function sitenow_metatags_alter(array &$metatags, array &$context) {
+function sitenow_seo_metatags_alter(array &$metatags, array &$context) {
```
richardbporter, line 25: Wow were these hooks not working then?

joewhitsitt: i couldn't figure out a way to test this one.

joewhitsitt: the other was just a couple days live. i moved it from the profile and merged it but copied the function name from the one above.

richardbporter: I must've goofed this up a long time ago. I think I can test it.

joewhitsitt:
Hoping to get @pyrello to spot check this:
>Deleted person_education bulleted list template because it is not being used for person_card (full node) styles. Not aware of a grad or admissions content type or person variant using it.

docroot/modules/custom/sitenow_seo/sitenow_seo.module
richardbporter: Where are we using a non-AJAX pager anymore?

pyrello:
@joewhitsitt If we are doing something different with this field, it would be in a different content type than person and therefore use different templates. Otherwise, we want to stay in line with how this field is used, so this change should be good for Grad. Not sure about Admissions.

docroot/modules/custom/sitenow_seo/sitenow_seo.module
richardbporter: Original issue https://github.com/uiowa/uiowa/issues/115

richardbporter: Oh well. I'm sure its fine. Only needed for non-AJAX paged views but I can't find any.

richardbporter (approved):

PR #2961, opened 2021-01-28
Academic Units fix for saving data

### LINKED ISSUE #2801: Academic Units field not retaining value on save
# Problem

At least on mentor person type of grad college, and likely others, the Colleges form field (`field_uiowa_college`) isn't retaining fields entered in the editor. On save, no values entered are saved to the database.

# Steps to reproduce

`blt ds --site=grad.uiowa.edu`
Create a new person
Choose the `Mentor` person type
Add value(s) to the College field and save
Edit the Person and see that the College field is empty

# Proposed solution

Not sure if this is a local issue, or a regression in the form widget. An early iteration of the Academic Units from widget had a bug in how it set the default value which led to empty fields, and this may have resurfaced.

cory-skeers:
Resolves #2801

# How to test

## Adding a new field

- Sync a site
- Add a new field to a piece of content of type Reference > Academic Units
- Test with limited allowed values (1 or more) and unlimited
- In the Manage Form Display options, try adjusting the field widget settings to include college, non-collegiate, or both types

Check that the field functions as you would expect (e.g. it allows you to select only one option if 1 allowed value, multiple if unlimited values, that your choices are kept upon saving, etc.).

Adding a basic field should result in minimal on-page styling if no template adjustments are made--only label and plain-text display of the units chosen. Primarily looking that the field form is functional and that values are saving.

## Existing field

- `blt ds --site=grad.uiowa.edu`
- Edit an existing or create a new Mentor
- Select a value for the `College` field
- Check that the content saves
- Check that the display includes the (newly added) `College`


cory-skeers:
@pyrello I believe this gets the `College` field for Mentor content working correctly; it includes a small config update for the field along with the general fixes. A followup to this can be to get the old College values into place that weren't migrated (should be ~10 of them).

dnkemper (approved):
Nice job, Cory! Tested in grad and works with existing and when creating new. Also works when adding new academic units. 👍

Issue #2563, opened 2020-11-17
Style print/PDF version of Area of Study page.


pyrello:
# Story

As an Admissions staff member I want to be able to print (or download a PDF of) a styled version of Areas of Study pages so that they look beautiful when they are sent to a prospective student.

# Proposed solution

Based on the results of #2500, implement some way of outputting specific fields in a specific format for an Area of Study page when viewed as a printable page or PDF.

Todo:

- [x] Remove 'View PDF' link shown for authenticated users at the bottom of AOS pages.
- [x] Get TTF fonts working in PDF.
- ~~Create specific role for access AOS PDFs and listing view.~~
  - Authenticated role has permission. Webmasters can add users with viewer role for most-restrictive access.
- [x] Recreate view of AOS print links without print instructions: https://admissions.uiowa.edu/at-iowa-print
- [x] Set downloaded filename to mail item code field.
- [x] Refactor imported UIDS partials as PDF-local versions to avoid generation failure.
- [x] Create on-brand PDF with an existing template in mind: https://brand.uiowa.edu/template-library

cory-skeers:
Depending on the level of formatting needed, probably would be more involved than is necessary to get it working, but could be worth checking out https://www.latex-project.org/. Predominantly used for formatting academic papers, or print-ready materials that need special formatting like mathematical equations, but it is very useful when formatting PDFs in general.

https://www.drupal.org/project/latex, https://www.drupal.org/project/drutex are two possible modules at least as reference for implementation (very underused), and maybe get an idea of it's worth looking into or a non-starter.

richardbporter:
> Depending on the level of formatting needed, probably would be more involved than is necessary to get it working, but could be worth checking out https://www.latex-project.org/. Predominantly used for formatting academic papers, or print-ready materials that need special formatting like mathematical equations, but it is very useful when formatting PDFs in general.
> 
> https://www.drupal.org/project/latex, https://www.drupal.org/project/drutex are two possible modules at least as reference for implementation (very underused), and maybe get an idea of it's worth looking into or a non-starter.

Thanks but I think we're too deep into dompdf at this point. I also don't know of LateX would run on AC depending on the requirements.

PR #2970, opened 2021-01-28
Initial PDF functionality for Areas of Study.

### LINKED ISSUE #2500: Areas of Study PDFs
# Motivation
As part of #2365, "The area of study pages also have carefully designed PDFs that are generated and used internally by Admissions to provide print copies of this information. (https://admissions.uiowa.edu/at-iowa-print)."

# Story

As an authenticated user on the Admissions site,
I want access to a carefully designed PDF version of Areas of Study
so that I can print well-structured, on-brand copies of information from the website to give to prospective students.

# Example
[Accounting.pdf](https://github.com/uiowa/uiowa/files/5850453/Accounting.pdf)

# Proposed solution

- Explore options for generated PDFs in Drupal 8.
  - https://www.drupal.org/project/pdf_generator
  - https://mpdf.github.io/
  - https://www.drupal.org/project/printable
- Implement the best option as part of the Admissions split.
- Build an access-restricted page or similar option to generate PDFs and/or access pre-generated PDFs
   - If pre-generated prior to access:
     - Is it okay if the files are technically accessible publicly, but hidden behind an access page, or do they need to be locked down?
     - How/when are they re-generated? Content post-save hook, cron, etc.?

richardbporter:
Resolves #2500 

### ToDo
- [x] Hide View PDF link from anonymous users.
- [x] Get PDF images working.

### Testing
- `composer install`
- `blt sync --site admissions.uiowa.edu`
- `blt frontend`
- `drush @admissions.local uli /areas-of-study/accounting`
- See the `View PDF` link at the bottom.
- Click the `View PDF` and verify it renders without errors.
  - The featured image is hidden but images in text areas should be displaying.
- Log out and go to /areas-of-study/accounting again. See no `View PDF` link at the bottom.
- Go to http://admissions.local.drupal.uiowa.edu/print/pdf/node/761/ and verify access denied.

### Advanced
- Make a SCSS change in admissions_core/sass/pdf.scss
- Rebuild the workspace: `yarn workspace admissions_core run build`
  - Or watch for changes, see https://github.com/uiowa/uiowa/blob/main/README.md#workspaces
- Rebuild the cache: `drush @admission.local cr`
  - This is needed after any CSS change which is annoying. I'm not sure what cache bin to clear yet. 
- Append `/debug` to the print PDF URL of an AOS. This is what the library uses to generate and you can inspect styles/markup here.

I think most of the styling for this can happen in #2563 but I'm looking for initial thoughts on how this will work. Apparently, PHP PDF libraries have limited CSS support. The one we're using here, dompdf, has limited CSS 3 selector/property support. None seem to have flexbox or grid support. Because of this, we can't just import many of the UIDS component partials. Do we need to? Can we just write very basic styles for these PDFs?


richardbporter:
@bspeare Curious about your thoughts on this. I added some notes on testing. If you have a minute to check it out sometime, that would be appreciated.

bspeare:
@richardbporter I'll take a look this morning.

bspeare:
@richardbporter This looks like it will work.  The one question I have is whether we can group the fields in column wrappers to recreate their current pdfs.   We can probably set up the columns in the pdf template?

Screen Shot 2021-02-01 at 11 49 35 AM

richardbporter:
Yes that should be possible using the `node--area-of-study--pdf.html.twig` template. That was my thinking as well if they want to recreate those PDFs or do something basic. If more complicated layouts/features are requested, it might be a pain. Thanks for checking.

richardbporter:
> The one question I have is whether we can group the fields in column wrappers to recreate their current pdfs.

Or maybe that needs to happen in the `entity--print--node--area-of-study.html.twig` template? Either way, seems doable.

I threw a regular `` element in the node template and it renders ok. Seems like we can't use responsive/lazy images here which is fine, we don't need it for PDFs. The D7 PDFs had no images so I'll leave them out for now but we can preprocess an image style variable later if needed.

bspeare:
@richardbporter Some of them have images: https://admissions.uiowa.edu/printpdf/39

richardbporter:
Doh! Back to draft...

config/default/lazy.settings.yml
```diff
@@ -24,7 +24,7 @@ formatters:
   lazy_image: '0'
   lazy_responsive_image: '0'
 skipClass: no-lazy
-disabled_paths: "/rss.xml\r\n/entity-browser/*\r\n/entity-embed/*\r\n/media/*"
+disabled_paths: "/rss.xml\r\n/entity-browser/*\r\n/entity-embed/*\r\n/media/*\r\n/print/pdf/*"
```
joewhitsitt, line 27: look at you. clever

joewhitsitt:
pulling down to test

config/default/lazy.settings.yml
richardbporter: Yea I exported it to default in case anyone else uses PDFs. I don't think that will cause issue on other sites.

richardbporter: For the record, I don't think entity_print is something we should enable for all sites.

joewhitsitt:
This PR is guilty of not getting merged first:

```
diff --git a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.fragment.stat.default.yml b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.fragment.stat.default.yml
index ad7faa579..e9865ec00 100644
--- a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.fragment.stat.default.yml
+++ b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.fragment.stat.default.yml
@@ -70,7 +70,4 @@ content:
     settings: {  }
     third_party_settings: {  }
 hidden:
-  entity_print_view_epub: true
-  entity_print_view_pdf: true
-  entity_print_view_word_docx: true
   search_api_excerpt: true
diff --git a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.node.area_of_study.pdf.yml b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.node.area_of_study.pdf.yml
index ffea881a2..051e7b086 100644
--- a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.node.area_of_study.pdf.yml
+++ b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.node.area_of_study.pdf.yml
@@ -204,8 +204,5 @@ hidden:
   entity_print_view_epub: true
   entity_print_view_pdf: true
-  field_area_of_study_course_work: true
-  field_area_of_study_scholarship: true
-  field_area_of_study_stat: true
   field_image: true
   search_api_excerpt: true
diff --git a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.node.area_of_study.teaser.yml b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.node.area_of_study.teaser.yml
index e16779e8c..3ab8e966b 100644
--- a/tmp/drush_tmp_1612213473_60186ce137769/core.entity_view_display.node.area_of_study.teaser.yml
+++ b/tmp/drush_tmp_1612213473_60186ce182b4a/core.entity_view_display.node.area_of_study.teaser.yml
@@ -44,9 +44,6 @@ content:
     settings: {  }
     third_party_settings: {  }
 hidden:
-  entity_print_view_epub: true
-  entity_print_view_pdf: true
-  entity_print_view_word_docx: true
   field_area_of_study_career: true
   field_area_of_study_certificates: true
   field_area_of_study_course_work: true
```

joewhitsitt (approved):
Works as described.

>I'm looking for initial thoughts on how this will work.

Probably simple styles and/or a lite aggregate of styles. Can we not use any of the brand templates for this? We are technically going from web (uids) to print (brand).

richardbporter:
Sorry for goofing that merge up. 

I just hid the stories field on PDF as I don't think that will render. I also noticed fonts do not seem to be working in the PDF but are in debug mode. 

Maybe we should convert this back to draft and put more style work into it if all AOS fields are 100% done. Maybe #2563 could just be closed then and we wrap everything into here. Thoughts?

> Can we not use any of the brand templates for this?

I doubt it as this library converts HTML to PDF so we have to "rebuild" an on-brand PDF as HTML only for it to be converted into a PDF again. :)

richardbporter:
I'm assuming we need fonts to work so, back to draft...

bspeare:
I was able to get fonts to work with `!important` on the declarations.  I think we have all the fields to build this now.  Should we just try to organize it based on the existing examples: https://admissions.uiowa.edu/printpdf/39?

richardbporter:
Sure. I'm also wondering if we should not import any UIDS partials but just recreate only what is needed. My concern is that if any unsupported CSS gets added to a UIDS component in the future, the PDF generation will stop working.

richardbporter:
Where did you add that important? Can you push that up?

bspeare:
Sure, I'll try adding something.

richardbporter:
> My concern is that if any unsupported CSS gets added to a UIDS component in the future, the PDF generation will stop working.

Just to clarify: its not that the selector won't work in the PDF, its that the generation will choke and redirect back to the node. Here is an example:

image

bspeare:
After looking at this some more, I would just remove the brand styles in the `pdf` folder and just try to recreate their D7 pdf using `Roboto` or `Arial` since the pdf doesn't seem to be picking up the custom fonts.

bspeare:
It also looks like the inline styles it is adding to the generated PDF is either `serif` or `sans-serif` so we probably don't have the option to integrate brand typography.  Is that your impression?

richardbporter:
It sounds like it should work based on the docs but only certain formats are supported: https://github.com/dompdf/dompdf/wiki/About-Fonts-and-Character-Encoding Are there `ttf` formats available? I see some included in the zip files on the brand site. We could probably add those to the admissions_core module.

bspeare:
Do you want to try and download `ttf` from the brand site and see if the PDF will use it?

richardbporter:
Ok this works in dev: https://admissions.dev.drupal.uiowa.edu/areas-of-study/accounting

If we're ok with how this works, lets merge it and follow up with more styling in #2563.

Issue #2893, opened 2021-01-20
Migrate articles from engineering.uiowa.edu to engineering.prod.drupal.uiowa.edu


briand44:
https://github.com/uiowa/uiowa/issues/2652

# Proposed solution

- [x] Add engineering migration module/split
- [x] Write engineering article migration YAML
- [x] Write engineering article migration source plugin
- [x] Map author field--prepend to Organization
- [x] Map primary department, department feed, and audience fields--include as prefixed tags

cory-skeers:
Including some of the correspondence on the migration here--quotes are questions asked of Bill, and following texts are his responses.

> "Legacy" fields, such as Legacy NID, Temp Department, etc.: Do these need to be kept, or can they be ignored? (It sounded like these might not need to be moved, but I don't think there was a 100% decision yet.)

I'm pretty sure these can be ignored. I think some were just part of the last migration from the 2016 redesign.

> Author field: These are inline entities in V3, and could be created as part of the migration with some extra time. There also might be some issues with naming inconsistencies if initials are used in some places and not others, titles, etc. 

> The quicker option would be to put the author into the "Source Organization" field without the "Source Link" -- It would look essentially the same to the end user, but not take advantage of linking various articles to the same author. There could always be manual cleanup later, or possibly an update hook if it's decided that it's needed later.

Yes, whatever you recommend on these is fine. A generic author is totally fine.

> Extra fields, such as Department or Audience: We've talked about including those using the V3 tags. I was thinking they'd be prefixed by each category, so they could still be clearly distinguished between each other and other tags (e.g. audience-faculty, or similar).

Yes, the department and audience would probably work best as taxonomy tags now. I believe my original request to Brian/Michael had an excel file that listed all the tags used in the old site and what I had set up in the new site, including node IDs. If you need those files, I can send them to you. It might help. I think there are about 25-30 tags being used. If those can be kept, it would help a lot (but translated to how the new site manages that. The exact method does not have to be kept.)

> Files: Is there a need to migrate files as well? Previously migrations have been set to first migrate all files, and then node migrations use this migration mapping to attach the new media entities in field images and inline references. Working on an option to conditionally migrate files only as they're found in migrated nodes, rather than migrating ALL files beforehand. Once this conditional option has been done generally, either getting it in place or just migrating all files take about the same amount of time. It just comes down to: do you only want to migrate files that are used in articles, or do you want all files, regardless of whether they're used?

[...]

PR #3032, opened 2021-02-05
Initial framework for engineering migrate module.

cory-skeers:
Relates to #2893.

Initial work for an engineering migration module for use with engineering articles.

# How to test

Need a local copy of www.engineering.uiowa.edu in Dev Desktop (on uiowa703, listed with the www).
Leave the DD Apache/MySQL stack running.

`git checkout engineering-migrate` and `git pull`
`blt ds --site=engineering.uiowa.edu`
`drush @engineering.local en engineering_migrate`
`drush @engineering.local mim d7_engineering_article`

It will most likely terminate. 
 - If it terminates with a ` [warning] Memory usage is now 442.78 MB (86% of limit 512 MB), not enough reclaimed, starting new batch` you can simply run the `mim` command above again. 
 - If it terminates with a `Drush command terminated abnormally`, you will have to run: 
     - `drush @engineering.local mst d7_engineering_article` to stop the migration, then 
     - `drush @engineering.local mrs d7_engineering_article` to reset the migration to Idle, 
     - and then run the `mim` command again.
   
`drush @engineering.local mim d7_engineering_article_redirect`

Check that migration completes successfully. Currently there is a discrepancy in row counts--a 'complete' migration should import 1139 articles.

Spot check articles and see that nothing looks out of place, that articles from the existing site which had featured images have featured images post-migration.

For inline images, can check the database for any body fields containing a `<!--Missing image: @filepath-->` token.

Spot check for any other gotchas that you can think to check for that may not have migrated over correctly.

With the redirect migrate, can start at https://www.engineering.uiowa.edu/news, pick and choose articles, and use the same aliases on the prod site.


cory-skeers:
Should look into batching these. In testing, running into memory usage errors and needing to run multiple imports in order to complete the migration.

Edit: There's a `batchSize` variable to set in `SqlBase` (extended by the source class), as well as a `batch_size` option in the source plugin, but either they are not working properly, or there is a separate memory usage problem not solved by batching.

cory-skeers:
Mapping appears to be complete and functioning, including pulling in tags and mapping author field as plain text into the source organization field.

joewhitsitt:
I am not going to be much help in finding the memory leak from the migrate side of things, but one thing I will say is that the Engineering site as it exists has had database corruption issues in the past that we have had to work through. Not articles related (to my knowledge) but stuff related to panelizer/page manager. Could it be choking on a particular node or could there a base64 image in the markup that makes this time out/leak? Keep in mind, articles were originally migrated from an old site so this could be a remnant of that, etc. Let me know if I am way off, but been thinking about how I could help and this is what came to mind.

richardbporter:
IDK if this is much of a solution but I was able to run both migrations in one go by adding `ini_set('memory_limit', -1);` to the `local.settings.php` file for engineering.uiowa.edu. This disables the memory limit for this site in the local environment. It can and probably should be removed afterwards.

Is this migration going to be run once locally and then synced? If not, disabling the memory limit won't work obviously.

cory-skeers:
Maybe I've just looked at it too much, but I'm confused by the row counts:

`drush @engineering.local ms` gives 1318 rows, when the true count is 1139. Only 1139 nodes will process, which is accurate. Looks like it's joining the `field_data_field_image` that leads to the higher count, but that field doesn't offer multiple values (so no `delta` or otherwise leading to otherwise repeated node entries), and the `leftJoin` should take care of the fact that there are image entries for nodes other than article.

Otherwise, the migration still crashes but can now be restarted and executed to the end (or the crash can be avoided altogether--thanks @richardbporter!).

Few loose ends to tie up, then ready for testing and migration.

cory-skeers:
Marking as ready to review, in case any others want to test. The issue has some notes on the different mappings (old department taxonomy being mapped to tags, etc). I'm currently running a test migration to check that everything looks good, which may lead to updates, but no further work is expected at this time.

cory-skeers:
~Inline "fid" text in the following pages; checking to see if they're links to PDFs, files, etc., that may be broken from migration.~

~43 total--not too bad.~

False positives. Looks like we're clear.

richardbporter (approved):
I tested this the other day. Looks like enabling it results in a config diff which is fine since it will only run locally and then be turned off.

PR #3044, opened 2021-02-09
i2358 - coronavirus v2 to v3 

dnkemper:
Relates to #2358

- Updated config split to not include v2

# How to test

PR #3048, opened 2021-02-10
Feature: Update SiteNow V2 Media embeds + Coronavirus split removal

### LINKED ISSUE #865: Adopt Core's Media Library for v2
## Problem
With 8.8, Media/Media Library is stable and includes CKEditor placement of media entities, however it looks like there are some steps involved to adopt it over our Media/Entity Browser approach along with some limitations to think about.

- [x] Find/replace database across all sites/all filtered_html and full_html fields update to change <drupal-entity> markup to <drupal-media>
  - Look into how Reports > Field List is generated to get all fields and their type (long text formatted)
- [x] Configure media library form modes and view modes for each media entity type.
- [x] Swap out the EB button for the Media button and configure.
- [x] media_library_media_source_info_alter() to allow add-form for non-core media entities.
- [ ] Styles need to be adjusted for `<drupal-media>`
- [ ] Monitor https://www.drupal.org/node/2657592 as key to focal point/media library integration.
- [ ] Deal with the coronavirus.uiowa.edu mix of v2/layout
- [x] ~~Regex account for sm/md/lg view modes that have different machine names in v3.~~ - It looks like we don't have to account for view mode changes because we can keep the existing view modes.

### LINKED ISSUE #2358: Transition Coronavirus site to V3
# Problem
The Coronavirus site remains in a weird state where it is still a V2 site, but has Layout Builder enabled. This makes it a bit of a wildcard for some updates. Since the site is mostly (or completely?) transitioned to Layout Builder, it would be nice to finish that process and make that site a standard V3 site.

# Proposed solution
- [x] Confirm with client that it is okay to remove any revisions of content not currently in use. - _available window Feb 9th_
- [x] Migrate remaining live paragraph content to its LB equivalent.
- [x] Delete pages/revisions that are no longer needed. - https://coronavirus.uiowa.edu/admin/content?title=&type=page&status=All&langcode=All https://github.com/uiowa/uiowa/files/5767302/Coronavirus.pages.using.V2.paragraphs.pdf
- [x] Delete any orphaned paragraph content.
- [x] Disable sitenow_v2 split via update hook
- [x] Disable coronavirus_uiowa_edu split via update hook
- [x] Query for every textarea field and convert `<drupal-entity>` instances to `<drupal-media>` and their respective view mode attributes.
- [x] Schedule content freeze
- [x] Make sure no new pages have paragraph items created by default. Gut. Save. Clear revisions.
- [x] Make sure none of the layouts for pages have the v2 content block field placed. Remove. Save. Clear revisions.
- [x] Run Cron, delete orphans (twice).
- [ ] Merge/deploy https://github.com/uiowa/uiowa/pull/3048

pyrello:
This is a work in progress.

Resolves #865
Resolves #2358

# How to test

* `blt ds --site=hawkeyemarchingband.uiowa.edu` or another V2 site to test. The update hook skips non-V2 sites - presumably because there is nothing to update.
  * Example pages with different size media:
    * https://hawkeyemarchingband.local.drupal.uiowa.edu/about/history
    * https://hawkeyemarchingband.local.drupal.uiowa.edu/band-140
    * https://sitesgis.local.drupal.uiowa.edu/gis-day
    * http://brand.local.drupal.uiowa.edu/
    * http://brand.local.drupal.uiowa.edu/logo
    * https://hr.local.drupal.uiowa.edu/careers/professional-and-scientific-positions/career-development-planning
* coronvirus.local.drupal.uiowa.edu
  * no longer a hybrid v2/v3. You should see no v2 content block stuff on the page content type instead you should see a body field. You should be able to save the page without error. Any WYSIWYG placed media should be using the `<drupal-media>` markup and look "normal"
  * depending on current site activity, you may encounter a import error or a save error if there is paragraphs items or a placed field_page_content_block field on the layout. we will be manually clearing these and testing in DEV before deployment to PROD.


joewhitsitt:
I know it is early in draft and most of this code looks foreign to me, but can we set a revision message ("Programmatically saved" or something) with the entity save?

pyrello:
> can we set a revision message ("Programmatically saved" or something) with the entity save?

In the case of nodes, I think this makes sense. However, do we want to set this on blocks, media, and paragraphs too?

joewhitsitt:
>However, do we want to set this on blocks, media, and paragraphs too?

Wouldn't hurt, but probs not. Though I think they all have revision log fields, I don't think revision messages are available for reading on those entities... yet.

pyrello:
> ...most of this code looks foreign to me...

@joewhitsitt I think the comments do a pretty good job of laying it out. I had only minimally worked with PHP's DOMDocument prior to using it in migrations. It allows us to use an existing HTML parsing engine to find elements and replace them (roughly lines 2000-2043 in `sitenow.install`). Besides that part, the rest is the same logic we would need if we were finding and replacing stuff in the database to loop through all the entity types, each bundle, and each field instance to find the ones to which we need to apply additional processing.

pyrello:
I could use some help with figuring out what we need to do to account for possibility of a media embed's view mode changing. I'm not sure if we just need to map view modes that are not supported in the default profile to one that is?

Also, at this point, I'm wondering if it just makes sense to move ahead with this now instead of waiting until after #3044?

joewhitsitt:
poking around

- Prints HTML markup instead of render:
Screen Shot 2021-02-12 at 10 46 25 AM

pyrello:
Hmm... good catch. Not sure why that would be.

joewhitsitt:
field_text_body_format is NULL for everything that got saved. Need to set that as the replacement value is set. (e.g. filtered_html, minimal depending on the original)

pyrello:
I think that should fix it.

joewhitsitt:
so at first I thought small__widescreen would be the best replacement for small, medium and large just because it was the smallest option available, but I tried that on a few and the change was pretty dramatic. These view modes are still in the default (drupal core), but no longer an option for selection when creating new or editing existing. I feel we should take the grandfathered approach here and let folks revisit these as they update their images. The other view mode is "full" which is "original," where no data-view-mode attribute is added, so no change is necessary there.

pyrello:
If we can just leave the existing view modes in, I'm all for that.

pyrello:
I have tested this against both images and remote videos and I'm satisfied that is working for them. If anyone knows of examples for testing this against audio or social media embeds, please post them. Given how it is has worked for images and videos, I'm inclined to believe that we are good to go on the SiteNow V2 updates of this PR.

I'm going to test Coronavirus updates to make sure that is working as expected also. After that, I think this is good to go.

dnkemper:
Going to test this!

dnkemper:
Everything seems to be working fine so far BUT looks like there has been revisions to pages that we already combed through/removed revisions for. So I guess my testing was a bust 👎

pyrello:
@dnkemper Does that mean that the update to remove the coronavirus site split failed? It is not a problem per se to have new revisions, only if those revisions contain paragraphs.

joewhitsitt:
Blocked from merge release until we test coronavirus in DEV on 2/15 after a revisions sweep/sync. The rest of the v2 sites are testable.

pyrello:
I'm seeing some instances where media embeds don't seem to be properly set up after updating: http://coronavirus.local.drupal.uiowa.edu/returning-campus/classes

Screen Shot 2021-02-15 at 11 52 33 AM

I'm looking into why this may be.

pyrello:
Going to test this out on dev to see if this is just an issue locally because of Stage File Proxy.

pyrello:
It's still an issue on dev. From what I can tell, the difference of whether it happens or not seems to correspond to view mode. For example, if you edit the source of one of the text areas where this is happening and change `data-view-mode="image_medium"` (or another view mode that isn't working) to `data-view-mode="full"`, then it shows up as soon as you click the "Source" button on the WYSIWYG.

It is confusing why this would be the case. I'm not seeing the connection between that and the settings in Filtered HTML:
Screen Shot 2021-02-15 at 1 07 32 PM

joewhitsitt:
pulling down

think i figured it out. testing another sync or two

joewhitsitt:
ran into a config diff on admissions (entity_print). switching back to draft

joewhitsitt:
admissions diff:

  entity_print_view_epub: true
  entity_print_view_pdf: true

on core.entity_view_display.media.image.large, core.entity_view_display.media.image.medium, core.entity_view_display.media.image.small

we can move forward for now, this would fail on admissions, but this entity_print stuff will come up repeatedly similar to `search_api_excerpt` did for CLAS sites. @richardbporter thoughts?

richardbporter:
> we can move forward for now, this would fail on admissions, but this entity_print stuff will come up repeatedly similar to `search_api_excerpt` did for CLAS sites. @richardbporter thoughts?

What did we do for `search_api_excerpt`?

joewhitsitt:
We turned it on for all sites and added/hid it on every single display setting possible so it became a norm of our default config.

joewhitsitt:
I am going to split these for Admissions for now, but it will come up again as we update entity view display config.

pyrello:
If `entity_print` is a utility module like `search_api` is - that is, it doesn't really change anything for the end user on its own - then maybe we just employ the same strategy?

richardbporter:
It does other things that we may want to hide or get working for everyone. For example, a bulk operation on the content overview page to Download PDF when selecting one or more items.

pyrello:
Okay... let's hold off on it now.

joewhitsitt:
https://coronavirus.dev.drupal.uiowa.edu/ got the update. can folks spot check if they have time?

look for placed media, save page content, etc.

dnkemper (approved):

joewhitsitt:
waiting for stage rsyncs on a couple sites to finish

PR #3053, opened 2021-02-11
Old webform style updates

GaryRidgway:
# How to test


docroot/sites/blt.development.services.yml
```diff
@@ -3,6 +3,6 @@
 # Enable twig debugging.
 parameters:
   twig.config:
-    debug: true
+    debug: false
```
richardbporter, line 6: I think this should stay set to `true` because this file is only loaded in the local development environment.

richardbporter (changes requested):

richardbporter:
Just posting here what we mentioned earlier. @GaryRidgway Do you know if this PR is still needed?

GaryRidgway:
@richardbporter I think this PR has basically just become outdated. For some reason it was gutted at some point and things have changed so much that I think there is a need for us to take another serious pass using this PR as a reference when we do.

Issue #3052, opened 2021-02-11
Blog entry migration followup


cory-skeers:
Followup to #2792 -- pull in categories, allow `<pre>` tags and update content.

# Proposed solution

Categories didn't get pulled in, but we have a mapping of all old nids to new nids. Create a followup migration to create tags and add them to the proper nodes.
Allow `<pre>` in `filtered_html`, then update nodes that used it previously.

cory-skeers:
Looks like only 12 nodes used `<pre>` on the old site, so likely easiest to manually update once that tag is allowed.

Edit: And it looks like the tags are preserved, so once they're allowed those 12 nodes will be up-to-date.

pyrello:
@cory-skeers Is this issue closed by #3055 and #3058?

cory-skeers:
@pyrello I need to run the migration and sync it up to prod yet before it'll be fully ready to close.

cory-skeers:
Migration (categories followup) completed 🚀

PR #3055, opened 2021-02-11
Allow pre tags in filtered_html text fields

### LINKED ISSUE #2791: Add support for <pre> tag in text area
# Story

As a site owner I want to be able to use the `<pre>` tag to display pre-formatted content so that I can display content like poetry without having to recreate its structure.

# Proposed solution

Add `<pre>` tag to list of allowed tags for filtered_html and/or minimal_html formats.

cory-skeers:
Resolves #2791, relates to #3052 

# How to test

`blt ds --site=writinguniversity.uiowa.edu`
Check that config update completes.
Visit http://org-writinguniversity.local.drupal.uiowa.edu/news/2019/04/jason-r-iowa-three-boxes
See that the text is in a `<pre>` tag, allowing for poetic license in spacing (https://www.writinguniversity.org/blog/jason-r-iowa-three-boxes can be used as reference)


cory-skeers:
Current styling for `pre`:

```
font-family: monospace,monospace;
    font-size: 1rem;
```

image

Works well for poetry, particularly on a site that is looking to go serif by default. @bspeare curious on your thoughts whether this needs adjustment prior to allowing?

joewhitsitt:
>and/or minimal_html

Do we want to address minimal as well?

cory-skeers:
> > and/or minimal_html
> 
> Do we want to address minimal as well?

Good question! It's at least needed in filtered_html, since it has a use case in articles which have a single text body that likely can't use minimal.

Not sure if it's needed in minimal, but I also have no objections to it. Line breaks are accounted for already for blocks like CTAs.

joewhitsitt:
If the immediate use-case is resolved, I am in no rush to add more to minimal. Reviewed. Not tested. Approved but waiting to hear what Ben has to say.

bspeare:
I think we want to add some different styles for this.  I won't be able to look further into this until later today.

bspeare:
I think the brand team would like to see this poetry formatted in the `<pre>` tag using the `Lora` font: https://github.com/uiowa/uids/pull/483.

bspeare:
Pulling this down to test.

bspeare:
Looks good.  I updated the UIDS version so that the Lora fonts are applied:

Screen Shot 2021-02-15 at 1 13 39 PM

bspeare:
Definitely no rush to add this to minimal.

bspeare (approved):
Tested this locally and it works!

PR #3058, opened 2021-02-12
VWU taxonomy migration

cory-skeers:
Relates #3052 

Adds taxonomy migration and postMigrateImport method (has to go in sitenow_migrate, since that event already exists) to add the taxonomy terms to the appropriate nodes already migrated.

# How to test

Need a local copy of writinguniversity.org in Dev Desktop.
Pull the `vwu` multisite from the Graduate College Acquia Cloud docroot.
Leave the DD Apache/MySQL stack running.

`git checkout vmu-categories` and `git pull`
`blt ds --site=writinguniversity.org`
`drush @org-writinguniversity.local en writinguniversity_migrate`
`drush @org-writinguniversity.local mim d7_writinguniversity_blog`
`drush @org-writinguniversity.local mim d7_writinguniversity_blog_categories`

The migration will appear to hang for a moment following the term migration--this'll be the event listener firing and adding the taxonomy terms to the relevant nodes.

Check that migrated articles have tags (http://org-writinguniversity.local.drupal.uiowa.edu/news/2012/03/welcome-our-new-website is an example one). Alternatively, check out taxonomy terms, and see that there are lists of referenced nodes for each.

### DIFF
--- a/docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
+++ b/docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
@@ -117,6 +117,9 @@ class MigratePostImportEvent implements EventSubscriberInterface {
    */
   public function onMigratePostImport(MigrateImportEvent $event) {
     $migration = $event->getMigration();
+    $source = $migration->getSourcePlugin();
+    $source->postImportProcess();
+
     switch ($migration->id()) {
 
       // Right now, page migration is set to run last.
@@ -128,11 +131,6 @@ class MigratePostImportEvent implements EventSubscriberInterface {
         $this->logger->notice($this->t('Checking for possible broken links'));
         $candidates = $this->checkForPossibleLinkBreaks();
         $this->updateInternalLinks($candidates);
-        break;
-
-      case 'd7_writinguniversity_blog_categories':
-        $this->addWritingUniversityCategories($migration);
-        break;
 
       case 'd7_file':
       case 'd7_article':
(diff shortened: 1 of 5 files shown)


docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
```diff
@@ -128,6 +128,12 @@ public function onMigratePostImport(MigrateImportEvent $event) {
         $this->logger->notice($this->t('Checking for possible broken links'));
         $candidates = $this->checkForPossibleLinkBreaks();
         $this->updateInternalLinks($candidates);
+        break;
+
+      case 'd7_writinguniversity_blog_categories':
+        $this->addWritingUniversityCategories($migration);
+        break;
```
pyrello: FWIW, I think we are going to need to look at a different way of implementing these MigratePostImportEvent items as this doesn't seem sustainable over several hundred migrations, even if we start culling out the ones that are done.

pyrello: One idea for how to do this: We could create an interface that we use for all our migration classes that sets an empty `protected function postImport()` that would be called and run from this file. In migration classes where we needed some additional action, we could just add an overridden version of that method.

cory-skeers: Nice idea--there's some cleanup I'll get to following it, but as far as this post import goes, got it set up this way.

docroot/modules/custom/writinguniversity_migrate/src/Plugin/migrate/source/Category.php
```diff
@@ -0,0 +1,112 @@
...
+
+    $to_update = [];
+    $entityTypeManager = \Drupal::service('entity_type.manager')
+      ->getStorage('node');
+    foreach ($results as $result) {
+      // Grab the old nid and map to the new.
+      $nid = $nid_mapping[$result->entity_id];
+      // Check if we already loaded the node, else load it.
+      if (!isset($to_update[$nid])) {
+        $to_update[$nid] = $entityTypeManager->load($nid);
```
pyrello: Getting an error on this line:
```
[error]  AssertionError: Cannot load the "node" entity with NULL ID.
```

I appears that there is the possibility of having a `$result->entity_id` that is not in `$nid_mapping`.

pyrello: I think I have resolved this. I'm not sure if we need to have an alternative action if a result nid is not found in the map.

cory-skeers: I'm having trouble reproducing this--originally added a logger as an `else` to the above to catch what id wasn't appearing in the mapping, but it never fired. Then tried to revert the key check, and I'm not getting the NULL ID error.

My guess would be that it's a discrepancy between the databases we're using, which concerns me--wondering if I have some remnant from having run the migration originally?

pyrello: I can try testing again to see if I can get info about what was failing.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
I didn't encounter the error. My db was up to date:

```
drush @org-writinguniversity.local mim d7_writinguniversity_blog
 [notice] Processed 0 items (0 created, 0 updated, 0 failed, 0 ignored) - done with 'd7_writinguniversity_blog'
vagrant@local:/var/www/uiowa$ drush @org-writinguniversity.local mim d7_writinguniversity_blog_categories
 [notice] Processed 53 items (53 created, 0 updated, 0 failed, 0 ignored) - done with 'd7_writinguniversity_blog_categories'
```
Screen Shot 2021-02-15 at 1 17 12 PM

Spot checked a multi word category and see an old 2015 article: http://org-writinguniversity.local.drupal.uiowa.edu/tags/writing-exercises-and-prompts

pyrello:
Hey... sorry. I'm still checking this out. I'll report back.

pyrello:
I cannot recreate the issue I ran into before. Hopefully it was just noise.

PR #3069, opened 2021-02-15
Feature: Adjusted headline text for Admissions AoS page

bspeare:
# How to test

Go to http://admissions.local.drupal.uiowa.edu/academics/accounting-0 and the "New Headline" text on the admissions requirements boxes should be changed to "Admission Process".


richardbporter:
I always run user-facing text through translate by default even though its unlikely translation is enabled on this site at a later time. However, if any variables get added to the text, translate will ensure they are properly escaped as long as placeholders are used.

richardbporter (approved):

PR #3072, opened 2021-02-15
Add basic validation to article view arguments and hide feed icons for filtered displays.

### LINKED ISSUE #1310: InvalidArgumentException: The timestamp must be numeric.
https://diversity.uiowa.edu/news/feed/media-advisory-columbus-junction-students-visit-ui-explore-health-science-careers-april-10/all 

InvalidArgumentException: The timestamp must be numeric. in Drupal\Component\Datetime\DateTimePlus::createFromTimestamp() (line 198 of .../docroot/core/lib/Drupal/Component/Datetime/DateTimePlus.php)

### LINKED ISSUE #2031: Remove monthly RSS feed on SiteNow Articles
## Problem
Based on the month of articles the user views, the RSS link changes to only show the current month's articles. Not particularly valuable.

## Proposed Solution
Change it so that is always the main RSS feed (link and text) or remove it completely from the monthly contextual display while maintaing the main feed. Beware: The path is dynamic based on site owner configuration.

richardbporter:
Resolves #2031 
Resolves #1310 

### Testing
- Sync a number of sites to ensure database update sitenow_articles_9001 runs ok.
- Log in to a site locally and test that the show/hide feed icon config for articles works as expected.
  - Note the change that it only appears on the main news page.
- Put in an article views argument as text and see that a 404 is returned as it does not validate.
  - https://default.local.drupal.uiowa.edu/news/asdf/asdf
  - https://default.local.drupal.uiowa.edu/news/feed/asdf/asdf

### DIFF
--- a/docroot/modules/custom/sitenow_articles/sitenow_articles.install
+++ b/docroot/modules/custom/sitenow_articles/sitenow_articles.install
@@ -16,3 +16,21 @@ function sitenow_articles_update_8001() {
     ->set('display.page_articles.display_options.display_extenders.metatag_display_extender.metatags.description', 'The latest articles from [site:name].')
     ->save();
 }
+
+/**
+ * Add numeric validation to articles view arguments.
+ */
+function sitenow_articles_update_9001() {
+  $config_factory = \Drupal::configFactory();
+  $config = $config_factory->getEditable('views.view.articles');
+
+  $config
+    ->clear('core')
+    ->set('display.default.display_options.arguments.created_year.validate.type', 'numeric')
+    ->set('display.default.display_options.arguments.created_month.validate.type', 'numeric')
+    ->set('display.default.display_options.filters.status.expose.operator_limit_selection', false)
+    ->set('display.default.display_options.filters.status.expose.operator_list', NULL)
+    ->set('display.default.display_options.filters.type.expose.operator_limit_selection', false)
+    ->set('display.default.display_options.filters.type.expose.operator_list', NULL)
+    ->save();
+}
--- a/docroot/modules/custom/sitenow_articles/sitenow_articles.install
+++ b/docroot/modules/custom/sitenow_articles/sitenow_articles.install
@@ -29,8 +29,8 @@ function sitenow_articles_update_9001() {
     ->set('display.default.display_options.arguments.created_year.validate.type', 'numeric')
     ->set('display.default.display_options.arguments.created_month.validate.type', 'numeric')
     ->set('display.default.display_options.filters.status.expose.operator_limit_selection', false)
-    ->set('display.default.display_options.filters.status.expose.operator_list', NULL)
+    ->set('display.default.display_options.filters.status.expose.operator_list', [])
     ->set('display.default.display_options.filters.type.expose.operator_limit_selection', false)
-    ->set('display.default.display_options.filters.type.expose.operator_list', NULL)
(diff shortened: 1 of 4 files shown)


docroot/modules/custom/sitenow_articles/sitenow_articles.install
```diff
@@ -16,3 +16,21 @@ function sitenow_articles_update_8001() {
     ->set('display.page_articles.display_options.display_extenders.metatag_display_extender.metatags.description', 'The latest articles from [site:name].')
     ->save();
 }
+
+/**
+ * Add numeric validation to articles view arguments.
+ */
+function sitenow_articles_update_9001() {
```
richardbporter: Is this the right numbering?

joewhitsitt:
pulling down to test

docroot/modules/custom/sitenow_articles/sitenow_articles.install
joewhitsitt: We've been doing it wrong:
```
Naming and documenting your function
For each change in a module that requires one or more actions to be performed when updating a site, add a new implementation of hook_update_N() to your mymodule.install file (assuming mymodule is the machine name of your module). Implementations of hook_update_N() are named (module name)_update_(number). The numbers are normally composed of three parts:

1 or 2 digits for Drupal core compatibility (Drupal 8, 9, 10, etc.). This convention must be followed.
1 digit for your module's major release version; for example, for 8.x-1.* use 1, for 8.x-2.* use 2, for Core 8.0.x use 0, and for Core 8.1.x use 1. This convention is optional but suggested for clarity.
2 digits for sequential counting, starting with 01. Note that the x000 number can never be used: the lowest update number that will be recognized and run for major version x is x001.
Examples:

node_update_8001(): The first update for the Drupal 8.0.x version of the Drupal Core node module.
mymodule_update_8101(): The first update for your custom or contributed module's 8.x-1.x versions.
mymodule_update_8201(): The first update for the 8.x-2.x versions.
```
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/9.1.x

joewhitsitt: 9.1.3 so 9101?

joewhitsitt (approved):
tested on gis.sites and iti

fyi, the feeds are still available but not linked to, (e.g. https://sitesgis.local.drupal.uiowa.edu/articles/feed/2017/07) so no concerns about broken feeds in-use.

thanks for getting to the bottom of this.

richardbporter:
> fyi, the feeds are still available but not linked to, (e.g. https://sitesgis.local.drupal.uiowa.edu/articles/feed/2017/07) so no concerns about broken feeds in-use.

Yea I think those arguments could be removed but I would put money on someone, somewhere using these and I didn't want to deal with that. This should work it self out naturally over time. :)

docroot/modules/custom/sitenow_articles/sitenow_articles.install
richardbporter: > 1 digit for your module's major release version

We don't technically have a release for this module so aren't we on 9001 still? Although we do write out the release version on deployment so maybe 9301 for 3.4.*?

Naming things is hard.

richardbporter: Lets save this for another day. :)

PR #3094, opened 2021-02-16
cleanup v2 so it uses default's editors

joewhitsitt:
Remains of v2's editors were still in the config split which still showed the entity embed button instead of the media library button after the update in stage.

To Test
pull and sync coronavirus, brand. compare editors between stage and local. local should have the media library button and it should be usable.

unrelated issue: claro f'd the media library modal.
Screen Shot 2021-02-16 at 12 51 36 PM


pyrello:
Huh... I thought this had already happened. I wonder if it got re-exported somehow.

I'm going to do my update hook code review and post any changes here.

dnkemper:
I can test this after grooming!

joewhitsitt:
Not seeing a diff in the amount of updated content with d1f8aae on coronavirus.

pyrello:
Me either. Not totally sure about what are the conditions under which content shows up as updated under the content list.

pyrello (approved):
Tested on coronavirus.

PR #3102, opened 2021-02-17
Move database size query to separate command and limit scope of status alter.

### LINKED ISSUE #1338: Path alias %files does not work with Drush rsync
```
vagrant@local:/var/www/uiowa$ drush rsync @home.test:%files @home.local:%files

In BackendPathEvaluator.php line 85:
                                                               
  Cannot evaluate path alias %files for site alias @home.test 
```

richardbporter:
The Drush status command is used by other commands (Drush, BLT, etc.) to check various things. Running our database size query every time was coming up in the slow query logs on AC. This moves it to a separate command to avoid that. This also only alters status output if no specific fields were provided via input option which I think was breaking a couple things.

I think this actually resolves #1338 - I've deployed this feature branch to dev.

### Testing
- On main, run `blt doctor` and notice that it reports errors.
- Check out this branch, and run `blt doctor` again. Should have no errors.
- Run `drush @home.dev uiowa:database:size`.
- Run `drush @home.dev uiowa:table:size`.
- Run `drush @home.dev status` and notice application is in output.
- Run `drush @home.dev status --fields db-name` and notice application is not in output.
- Run `drush @home.test rsync @self:%files @uiowa.test:%files` and see that the command errors.
- Run `drush @home.dev rsync @self:%files @uiowa.dev:%files` and see that the command will run but please cancel!

Todo:
- [ ] Update slackbot response if this works.

### DIFF
--- a/drush/Commands/UiowaCommands.php
+++ b/drush/Commands/UiowaCommands.php
@@ -50,6 +50,7 @@ class UiowaCommands extends DrushCommands implements SiteAliasManagerAwareInterf
     $defaults = $annotationData->getList('default-fields');
 
     // If no specific fields were requested, add ours to the defaults.
+    // @todo Is there a more-defined context for when to do this?
     if ($fields == $defaults) {
       $annotationData->append('field-labels', "\n application: Application");
       array_unshift($defaults, 'application');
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Got rid of some errors but I still have a local one:
```WebUriCheck:checkHttps The SSL certificate for your local site appears to be invalid for https://default.local.drupal.uiowa.edu. ```

Everything else printed out correctly as the test steps indicate.

PR #3105, opened 2021-02-17
Feature: Admissions pdf fixes

bspeare:
This pull requests fixes a few missed items missed in the pdf-styles pr.  

# How to test

- `yarn workspace admissions_core build`
- `drush @admissions.local cr`
- Go to http://admissions.local.drupal.uiowa.edu/areas-of-study/print-links
- Verify that the pdf looks similar to the example below

Screen Shot 2021-02-22 at 10 18 27 AM


richardbporter:
Notes from meeting:

- [x] Put textarea fields in two columns. This will help keep PDFs to 2 pages or less. Use Art and/or engineering as test cases once populated.
- [x] Output the link URLs in parentheses. This helps with physical copies.
- [x] Put the featured image in the PDF, somewhere.

bspeare:
Columns are now working by splitting the fields into two separate pages.  Now we need to figure out how to limit certain fields from overflowing to the second page.

bspeare:
Looks like the D7 site was using something like https://developer.mozilla.org/en-US/docs/Web/CSS/columns to split up all the content like in the Art example.    I'm not seeing anything on https://github.com/dompdf/dompdf/issues/662 that suggests they support anything like that.  @richardbporter do you think we can throw some JavaScript at it to make the columns wrap? https://adamwulf.me/columnizer-jquery-plugin/

bspeare:
@richardbporter  are we able to try any of the other libraries supported by the Entity print module?  Like `wkhtmltopdf`?

richardbporter:
We could try tcpdf but IDK if its any better. wkhtmltopdf isn't supported on AC: https://support.acquia.com/hc/en-us/articles/360020517573-Acquia-Cloud-and-wkhtmltopdf We might be able to get it work ourselves but I would say its not a good idea.

bspeare:
It might be worth trying `tcpdf` to check if the `columns` property works.

richardbporter:
I can't find any docs on tcpdf that suggest that is supported. You can download it with `composer require tecnickcom/tcpdf` and then change the Entity Print configuration to use that parser. Clear the cache and it should be in use. I did it locally and the PDF looks jacked:

image

richardbporter:
I'll look into loading JS on the debug view and see if that gets parsed. I'm thinking no it won't but I'll test it out.

bspeare:
Thanks.  That is pretty jacked!

richardbporter:
The module does not even support adding JS to the debug page. I added some inline JS in the template, changed the font color but as I suspected the PDF parser did not run the JS. 😢

richardbporter:
I can try to get Wkhtmltopdf running locally for the heck of it.

richardbporter:
Having some issues even getting the module to recognize wkhtmltopdf...

bspeare:
That is weird considering that it gives you `composer install` instructions on the config page!

richardbporter:
I wonder if instead of rendering individual fields in the template, we render custom column variables that contain dynamic content that we preprocess into those variables. There has to be some approximate amount of lines that fit per page column. If we could calculate that and the amount of lines per field, split them and transfer the overflowing lines to the next column, would that possibly get this working?

richardbporter:
> I wonder if instead of rendering individual fields in the template, we render custom column variables that contain dynamic content that we preprocess into those variables. There has to be some approximate amount of lines that fit per page column. If we could calculate that and the amount of lines per field, split them and transfer the overflowing lines to the next column, would that possibly get this working?

Probably not as the number of lines/characters would not be accurate enough, I don't think.

bspeare:
Any luck with `wkhtmltopdf`?

richardbporter:
No. If its this hard to set up locally, its probably harder on AC and not worth it. I asked if PDF Reactor could make an exception to remote URLs just to see how it works.

bspeare:
Here's how admissions is using columns with `mpdf`: https://mpdf.github.io/what-else-can-i-do/columns.html.

bspeare:
Here's an issue with displaying the url next to the text: 

Screen Shot 2021-02-18 at 9 45 08 AM

richardbporter:
I'm playing around with trimming the text area output to a maximum number of chars/ words. Not sure if that will work long-term.

mpdf is not supported by the module. I guess I chose the wrong module...

richardbporter:
Smart trim isn't that smart and cuts off text mid-sentence. 0feb75f can probably be reverted. I think if reducing text length is viable it will have to be done via text area summaries.

bspeare:
@richardbporter Probably so. Can you revert that one?

richardbporter:
wkhtmltopdf may not support columns out of the box either: https://stackoverflow.com/questions/36968778/css3-multi-column-layout-converting-to-pdf-using-snappy-wkhtmltopdf-wrapper

richardbporter:
Is it fair to say that we probably have three options?

- Use the current solution without columns and use text area summaries to provide more succinct version of fields in the PDF.
- If columns are required, try to get MPDF working with another module or manually. Port existing styles and functionality
OR
- Possibly use some combination of custom routes and print style sheets to do the columns. This would lack the PDF viewer functionality.

joewhitsitt:
https://www.drupal.org/project/smart_trim
>Whether the trim length is measured in characters or words

Words should be possible.

Update: sorry, misread your comment about mid-sentence.

bspeare:
@richardbporter I updated the options with to reflect the version using columns in this pull request. 
- Use the current solution (https://admissions.prod.drupal.uiowa.edu/print/pdf/node/981) without columns and use text area summaries to provide more succinct version of fields in the PDF.
- Use the current updated solution in this pull request with smaller font size and columns that show specific fields. 
- If fluid columns are required, like https://admissions.uiowa.edu/printpdf/95, try to get MPDF working with another module or manually. Port existing styles and functionality
    OR
- Possibly use some combination of custom routes and print style sheets to do the columns. This would lack the PDF viewer functionality.

bspeare:
@richardbporter Marking this as ready for review so we can get these updates to the live site until further discussion.

richardbporter (approved):
These look good.

PR #3109, opened 2021-02-18
Hotfix HR media

joewhitsitt:
Related to the helpdesk ticket.

Fix embedded videos on HR. Might have to be expanded to address uids_base v2 sites.

## To Test

- pull and sync HR
- `blt frontend` (uiowa_bootstrap and hr both need to build)
- Go to https://hr.local.drupal.uiowa.edu/advanced-layout-guide-not-publication and look at the videos section. It should fix the broken videos. Compare to PROD (unpublished) where two videos (small ones (default/full)) aren't rendering.
Screen Shot 2021-02-18 at 11 59 02 AM

PR #3111, opened 2021-02-18
HR Video - fix default and full size, fix ckeditor versions

joewhitsitt:
Follow up hotfix to https://github.com/uiowa/uiowa/pull/3109 where i got the video to render again but it was the wrong size

The default/full should be full-width on render and actually clickable/configurable in the WYSIWYG.

Looks like a lot but it is just wrapping an exclusion for ckeditor (everything is loaded in hr, so this problem doesn't happen in v2/v3) and making the default/full view mode full width responsive.

`body:not(.cke_editable) {}` is causing the large diff


pyrello (approved):
Reviewed, not tested

PR #3116, opened 2021-02-19
Admissions - AoS Visibility Issues/Subprograms

### LINKED ISSUE #3079: Add field for Tracks/Subprograms to areas of study and style output
Areas of studies can have subprograms, which are lists of related areas. These related areas are referred to differently by individual departments, so there is a subprogram type field that allows choosing from a specified list of items to refer to these as. The subprogram type that is chosen is used as the label for the subprograms on the display. These should be output under the program types in the sidebar.

Screen Shot 2021-02-15 at 3 32 06 PM

Screen Shot 2021-02-15 at 3 32 48 PM

# Proposed solution

### LINKED ISSUE #3092: As a site owner, I can add content to an area of study page and expect only the sections containing content to display
# Story

As a site owner, I can add content to an area of study page on admissions and have sections render only if there is content entered in each field. 

Currently, the view block and the "Admission Process" section display if there is not content in each section.

Screen Shot 2021-02-16 at 12 47 05 PM

# Proposed solution

- Update view block settings to not render if empty
- Find a solution for not rendering the "Admission Process" section if there is not content in the fields

joewhitsitt:
Resolves #3092 
Resolves #3079
Also includes small tweak for HR oEmbed

# How to test

- pull and run `blt ds --site=admissions.uiowa.edu`
- interact with the following nodes:

//admissions.local.drupal.uiowa.edu/node/991
//admissions.local.drupal.uiowa.edu/node/786
//admissions.local.drupal.uiowa.edu/node/add/area_of_study

## Information Block

- This now checks all the fields in the view for NOT NULL and strips tags from the block to check against.
- If the fields are empty, the block "Information" should not display (aka new AoS node). If they are filled out, the block should display (aka an existing AoS node).

## Admissions Requirements

- You should be able to remove the existing blank admission requirements. If all are gone, the heading doesn't show.
- New AoS content should not have these paragraphs open by default.

## Subprograms/Type

- Fields are available on the AoS content type. By default the label is Subprograms, but if you select a Subprogram Type the label of that selection is used for the Subprograms field label.
- Information block shows/hides based on the field values covered by the block.
- Using allowed values function in case the values change (add/edit/delete).

## Make sure PDF stuff is still working

- Had non-smooth merging with https://github.com/uiowa/uiowa/pull/3105 so please check that.


joewhitsitt:
Going to incorporate https://github.com/uiowa/uiowa/issues/3079 into this as it renders in the same block and will need the same NOT NULL view configuration

bspeare:
Pulling down to test.

bspeare (approved):
Testing criteria all worked locally.  I added these fields to the pdf template.  I did notice that the "Information" block on the page will still display if you have content in one of the licensure text fields when it is unchecked, but that is an edge case.

PR #3123, opened 2021-02-22
Enable and configure ultimate_cron.

### LINKED ISSUE #3059: As a developer, I can run cron jobs without consuming too many resources.
# Story

As a developer, I can run cron jobs without consuming too many resources.

# Proposed solution

Split up cron tasks ran every time using a contrib module like ultimate_cron. See https://befused.com/drupal/ultimate-cron for reference.

richardbporter:
Resolves #3059 

### Reminder
Cron runs once an hour in prod, offset by 10 minutes for each application.

Cron runs once every 3 hours in dev/test, offset by 10 minutes for each application.

### Changes
Enables ultimate_cron on all sites and configures all jobs to run hourly except Acquia Connector, Node Revision Delete and Tmp File Cleanup. Acquia Connector runs nightly at midnight, NRD runs nightly every hour from midnight to 2am and Tmp File Cleanup runs every 3 hours.

Still investigating cron logging to normal channels. This seems to use the DB to log... 👎🏻 

### Testing
- Go to https://uiowa.dev.drupal.uiowa.edu/admin/config/system/cron/jobs
- Review cron schedules, last run time, status, etc.
  - Note that I did not merge the latest NRD schedule change into dev. Its still set at 1-5 which I think is overkill.
- Run cron.
- See that AC and NRD do not run and last ran overnight.

I was trying to review stack metrics in dev but its not working.


richardbporter:
Looking at stack metrics in dev is not proving very useful as its on the same hardware as stage. However, I'm guessing this might not save as much memory as I hoped for. I think this is because our custom cron command is probably the thing that eats up the memory, not the cron tasks themselves.

I think we should move forward with this though if we want to be able to run cron more frequently like every 30 minutes.

richardbporter:
I'm not loving the logging functionality of this. Because they're stored in the DB, cron logs will no longer get forwarded to Splunk or be visible in streaming or server files. 

I'm also doubtful this makes a big difference in prod since probably most of the memory consumption is our custom cron command. We could just disable sending Acquia Insight data on cron and make it a manual process or create a nightly command just for that. A few other modules like NRD and Simple XML Sitemap have that config as well.

But... we won't know for sure unless we push it up to prod. Backing out of this would be pretty straightforward. I would say we push it, monitor things and go from there.

joewhitsitt:
Should I be concerned about this change as it relates to the brand cron job that we configured in Acquia Cloud?

richardbporter:
I don't think so. I guess one advantage of this approach is you could drop that custom job and schedule it through ultimate cron.

joewhitsitt (approved):
Poked around DEV a bit. I agree that this is the type of functionality that needs to be tested in PROD to see how it goes. "Ultimately" cron still runs with this module and had wide usage, so beyond that, everything is a step up in functionality.

PR #3148, opened 2021-02-26
Set date input placeholder attribute to expected format.

### LINKED ISSUE #3126: D9 Webform datepicker not working in Safari and IE
# Problem 

Datepicker does not appear on IE and Safari and manual date entry does not work either on: https://designcenter.uiowa.edu/form/submit-job-request.     

# Proposed Solution
- https://www.drupal.org/project/webform/issues/3027747#comment-14002223
- Remove datepicker?

richardbporter:
Resolves #3126

I thought this was a better workaround than adding back a jQuery UI or other JS polyfill. Safari will hopefully support date input elements soon.

### Testing
- `blt sync --site designcenter.uiowa.edu`
- Open https://designcenter.local.drupal.uiowa.edu/form/submit-job-request in Safari or IE.
- Notice that the placeholder value for the Due Date field is set to the expected format.
- Try to submit the form.
- Open the same form in Chrome, FireFox, etc. Notice the placeholder does not display.
- Try to submit the form.

Another workaround for this is to use the Date List webform component instead which offers dropdowns for each element of a date - year, month, etc. We should document this on https://sitenow.uiowa.edu/known-issues.

### DIFF
--- a/docroot/themes/custom/uids_base/uids_base.theme
+++ b/docroot/themes/custom/uids_base/uids_base.theme
@@ -850,6 +850,15 @@ function uids_base_preprocess_input(&$variables) {
   if (isset($variables['element']['#name']) && $variables['element']['#name'] === 'search-terms') {
     $variables['attributes']['class'] = [];
   }
+
+  // Set the placeholder attribute for date inputs if we know the expected
+  // format and no placeholder has already been set. This helps people input
+  // the date as a string for browsers that do not support the HTML input.
+  if (isset($variables['element']['#type']) && $variables['element']['#type'] === 'date') {
+    if (!isset($variables['attributes']['placeholder']) && isset($variables['element']['#date_date_format'])) {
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
I think this is a step in the right direction since the same element displays the data format requirement and this just adds the placeholder with the same name. It doesn't get in the way of the full override of the field with datepicker.

Tested in Safari. My local browserstack testing is no longer working at the moment. :(

joewhitsitt:
>Another workaround for this is to use the Date List webform component instead which offers dropdowns for each element of a date - year, month, etc. We should document this on https://sitenow.uiowa.edu/known-issues.

I like this idea

richardbporter:
I've added this to https://sitenow.uiowa.edu/known-issues#uiowa-issue-3148

PR #3154, opened 2021-03-01
I3060 accordion updates

### LINKED ISSUE #3060: As an admissions user, I want to be able to configure accordions.
# Story

As an admissions user, I want to be able to configure which accordions open or close automatically, and configure accordions so that only one is open at a time, so that pages with many accordions are clean and intuitive.

# Proposed solution

Add configuration to collections for accordions that allows this functionality

GaryRidgway:
Initial PR for tracking progress on #3060 

# How to test:
1. Sync down the sandbox site, you should see some imported config for a few new fields
2. `uli` in to the site and create a page
3. Create a collection bundle, and fill out some collection items. I've found 3 to be a good number.
    a. Verify that there is an option on collection item called `Automatically open`.
4. Verify that the `View mode` radio field is just above the collection items
5. Verify that if you select `Accordion` in the `View mode` that another field, `Accordion Type` appears just under it.
6. Verify that you can select either `Multi select` or `Single select` in the `Accordion type` field, and that the help text for this field describes how they work, and provides extra instructions for properly using the `Single select` option.
7. Verify that if you select `Multi select`: 
    a. Setting the `Automatically open` on any amount of collection items makes them automatically open when loaded. 
    b. Verify that many accordion items can be open at the same time.
8. Verify that if you select `Single select`: 
    a.  Verify that only one accordion can be open at a time after the first opening of an accordion.
    b. For the completion of this issue, it was deemed that with `Single select` it was up to the users to only have one item set to `Automatically open`. If they set multiple collection items to be `Automatically open` with `Single select`, the set items will automatically open, and then once a closed item is opened, the `Single select` collapsing functionality will activate. See example below.
    
aded300063e8bad71c341e17f06fa28d


richardbporter:
Dumb question: where is `$form['settings']['view_mode']` coming from?

cory-skeers:
> where is `$form['settings']['view_mode']` coming from?

Each type of collection (accordion, descriptive list, and tab, though tab is currently hidden as an option) are different display view modes of the collections block. The `$form['settings']['view_mode']` comes directly from allowing the editor to choose which view mode to use, with a few programmatic changes in the form alter.

richardbporter:
So is that like a field on the collection block or paragraph?

cory-skeers:
It's for the collection block and comes from it, but sits outside the collection block form itself.

It's like:

Layout Builder Form

- View mode field
- Block form
  - Headline field
  - Accordion type field
  - Paragraph references
    - Collection item headline field
    - Collection item body field
   
Thinks get weird going between levels. The form itself (and then the view_mode field) are accessible with the standard form_alter, but the block form (and paragraph) fields are more accessible using process_element. I'm unclear about how to easily move between the two.

joewhitsitt:
Pulling down to test

dnkemper:
Thank you, @joewhitsitt

joewhitsitt (approved):
Unfortunately this fails on Admissions because of https://github.com/uiowa/uiowa/issues/3204, otherwise this seems to work as expected. From what I can tell there is no loss with regard to the aria attributes.

Until we address #3204 can you split these files for Admissions and add the entity print stuff?
```
core.entity_view_display.block_content.uiowa_collection.tab     
core.entity_view_display.block_content.uiowa_collection.descriptive_list 
core.entity_view_display.block_content.uiowa_collection.default 
core.entity_view_display.block_content.uiowa_collection.accordion        
core.entity_view_display.paragraph.uiowa_collection_item.tab    
core.entity_view_display.paragraph.uiowa_collection_item.descriptive_list
core.entity_view_display.paragraph.uiowa_collection_item.default
core.entity_view_display.paragraph.uiowa_collection_item.accordion 
```

The single with multiple auto open scenario works as described but I guess I imagined it activated right away as the option only talks about load, not thereafter. As a visitor actually reading content, the existing functionality works for me and doesn't cause the "wait why did that one close?" moment.

I think the auto option help text should state "If Accordion and this option are selected, this item will be open on page load." This is because if I have Descriptive List selected, "Set this accordion to automatically open on page load" doesn't make sense.

joewhitsitt:
Assisting Alan.

joewhitsitt:
Steps to split/update

- Manually added wildcard items to the admissions split file
- Copied the affected files to the admissions config folder.
- Added to each copied config file:
```
entity_print_view_epub: true
entity_print_view_pdf: true
```
- DS'd admissions. Logged in after the fail and saved each display setting affected.
- Ran `drush @admissions.local csex admissions`
- Re-ran ds to confirm changes worked.

PR #3174, opened 2021-03-03
Update Webform module to 6.0.2

pyrello:
# Updates
* Updates webform module to 6.0.2
* Updates default `webform.webform.contact.yml` to include change from https://git.drupalcode.org/project/webform/-/compare/6.0.1...6.0.2

# How to test

* `blt ds`
* Observe update hook runs without issue.
* `drush @default.local admin/structure/webform/manage/contact/handlers`
* Click "Email Confirmation" to view those settings.
* Confirm that "To email" under the "Send to" section has been updated to "Current user email address [Authenticated only]."
* If you want to further confirm the change you can look at it in https://default.local.drupal.uiowa.edu/admin/config/development/configuration/single/export, select "Webform" and then "Contact (contact)" and search for `to_mail` and verify that the first value has been changed to `[current-user:mail]`.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -9694,17 +9694,17 @@
         },
         {
             "name": "drupal/webform",
-            "version": "6.0.1",
+            "version": "6.0.2",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/webform.git",
-                "reference": "6.0.1"
+                "reference": "6.0.2"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


joewhitsitt:
though presidentialsearch split this config, it is not affected by this change

joewhitsitt (approved):
We should probably make note of this in our article. Unauthenticated won't get confirmation emails now.

Also, this form is supported by the module, do we need to scan active db for other forms using this same token?

PR #3183, opened 2021-03-04
Add progress indicator to uiowa:multisite:install.

richardbporter:
### Problem
While creating sites I noticed the `blt uiowa:multisite:install` command has no output and is a time consuming process. This can feel like the command is stuck when installing multiple sites on AC. This PR adds a basic progress bar to indicate something. I know its using a deprecated `->io()` method but we have other instances of that already. We can replace them all in a follow up.

### Testing
- Comment out any multisite overrides in local.blt.yml
- Run `blt umi`
- See progress bar.
- Cancel command. No changes to actual command here.


GaryRidgway (approved):
A progress bar appears! And looks good!

Issue #2873, opened 2021-01-15
Create button block


bspeare:
# Story

As a site owner, I can create a button block component so that I don't have to use the CTA block component.

# Proposed solution

Create a button block which implements the markup and options from the UIDS 4 button component with these options:

Link field with a title. 

- Button type: 
  - Primary 
  - Secondary
  - Tertiary
- Transparent: true
- Button width: full-width
- Borderless: true
- Font: thin - Antonio is the bold default, Roboto is the secondary thin option. 
- Font awesome icon selector (default to arrow for first slice)
- Limit button link text to a certain number of characters?
- Button size: s, m, l - unclear if we want to release this option but worth exploring for the CTA row.

joewhitsitt:
I'd like to see if we can do a lighter approach for the icon field. https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L833

richardbporter:
Button with 255 characters: 
<img width="989" alt="Screen Shot 2021-01-19 at 1 25 58 PM" src="https://user-images.githubusercontent.com/1068523/105083444-859ac000-5a5a-11eb-92c7-99d18c76c184.png">

joewhitsitt:
A possible extension of this functionality would be to allow multiple (button group) and/or "display as vertical menu," "display as horizontal menu"

https://github.com/uiowa/uiowa/issues/3200?#issuecomment-794086415

briand44:
Sounds like we want to do this in 4.x of UIDS so https://github.com/uiowa/uiowa/issues/5408 should happen first.

pyrello:
In grooming we discussed how to proceed with this this issue. The main issue is whether there is any reason to consider implementing this with the 3.x version of the UIDS button component. Implementing the 4.x version of the button requires us to figure out #5408 first, but doing the 3.x version of this now means that we are committing to changing it over later. The 4.x version of the button is mostly done at this point, so it seems like the right time to focus on that work.

briand44:
Icon integration could be a future slice. Default to arrow.

briand44:
Font options

briand44:
Yes, we should limit number of characters allowed for button. Number is up for debate.

bspeare:
I removed the blocked label now that https://github.com/uiowa/uiowa/pull/5379 is merged.

PR #3186, opened 2021-03-05
Feature: Add main navigation menu depth to config ignore.

mark-bennett-uiowa:
Adds the following items to the config ignore settings:
block.block.main_navigation__superfish__horizontal:depth
block.block.main_navigation__superfish__toggle:depth
block.block.main_navigation__superfish__mega:depth

This allows developers to override the default menu depth to hide deep menus if we want them hidden, and the change wont get overwritten on the next release.

# How to test

drush @default.local cr
blt ds
drush @default.local uli
Change visible depth of any version of the main navigation from Unlimited to any number.
Navigate to Administration > Configuration > Development > Synchronize and make sure the change is in the ignored configuration entities (???)

This shows the reason for adding this capability;
On a wide screen display, the menu carries way over to the right hand side:
Screen Shot 2021-03-08 at 2 55 20 PM

On a more traditional sized display, the menu overlaps its grandparent menu in this criss-cross effect:
Screen Shot 2021-03-08 at 2 55 43 PM


joewhitsitt:
I am personally not a fan of this approach. To me it makes more sense to control this as part of a split. Most of our ignored configuration is because the site owner can change it in the database. Most of our developer changes are tracked in config. I wasn't part of the conversation though where this approach was decided on.

pyrello:
Why does this need to be tracked in config? Are we planning on changing the setting at some point down the road?

joewhitsitt:
>Why does this need to be tracked in config?

Why is anything tracked in config? My main point is that this isn't site owner controlled. Predictability, consistency, sustainability, readability.

>Are we planning on changing the setting at some point down the road?

Maybe. I don't know. It was the decided upon default.

If you want to make a case why it shouldn't be tracked in config compared to everything else, I think that would be a more productive conversation.

pyrello:
> My main point is that this isn't site owner controlled. Predictability, consistency, sustainability, readability.

The default behavior is that there is an unlimited depth. With deeply nested menus, this can get a little bit unwieldy from a UI perspective. In some cases, it makes sense to place a menu on the page and let that menu function as the primary way to drill down into a section. Without the ability to set the block depth per site, there is no way to truncate the display in the top menu of a potentially messy section of the navigation while still keeping it available for on-page navigation. This seems like it should be fairly particular to the needs of particular sites.

Creating a new menu to handle this issue has multiple potential downsides:
* It forces us to create a split for the site and specifically to split the `entity_view_display` config for the page type, which is problematic because we are creating additional maintenance that is necessary if we make any changes to that config for the default platform.
* Creating a separate menu is also potentially problematic from a site hierarchy perspective and I'm not sure what the impact is on the breadcrumbs of the page.

I agree that we should have sensible defaults, but I disagree that the user shouldn't be able to change this basic aspect of their site. That it was previously decided that it should be the default doesn't have any bearing on whether that is the right choice now.

joewhitsitt:
I was not aware that the depth switched from the v2 default of one sub-level when I originally commented on this PR. I agree that the unlimited is not ideal in almost all situations. It looks like I might have been a reviewer that let this slip past so I will take partial blame for not addressing this earlier. In fact, I don't think we even styled for this. I can't tell where one fly-out ends and another begins.

I am also not an advocate for more menus. We did it for HR and it has been a complicating factor. For admissions, it sounded like a faux menu block where they could have a list of internal/external links look like a menu would solve their issue.

I guess I don't understand this this hack/workaround of having things in the menu system that shouldn't be in menu system, but look like the menu system. I understand the unlimited depth puts a real twist on the issue.

>but I disagree that the user shouldn't be able to change this basic aspect of their site. That it was previously decided that it should be the default doesn't have any bearing on whether that is the right choice now.

I didn't make this point. This PR only allows the developer to make the change. If this should be controlled by the user, where is the site configuration that they can set this? In that case, yes, it should be ignored as it is controlled by the user.

joewhitsitt:
If there is a timeline for this, I will approve if there is a follow-up task to set a non-unlimited default and to give webmasters control of this.

mjoneill:
> For admissions, it sounded like a faux menu block where they could have a list of internal/external links look like a menu would solve their issue.

Agree, that's what they indicated when we discussed.  I am fine with this approach if it seems like the best way to deliver on the needs (but don't want to force us this direction if there's a better way).

bspeare:
> > For admissions, it sounded like a faux menu block where they could have a list of internal/external links look like a menu would solve their issue.
> 
> Agree, that's what they indicated when we discussed. I am fine with this approach if it seems like the best way to deliver on the needs (but don't want to force us this direction if there's a better way).

There are a lot of other sites besides Admissions that will need the type of functionality described here: https://github.com/uiowa/uiowa/issues/3200.   Which is why a faux block doesn't really solve the issue.

briand44:
>For admissions, it sounded like a faux menu block where they could have a list of internal/external links look like a menu would solve their issue.

My understanding is #3200 doesn't solve for the above comment either. #3188 would be the alternative, correct?

bspeare:
@briand44 https://github.com/uiowa/uiowa/issues/3188 could work based on how the module solution works or https://github.com/uiowa/uiowa/issues/2873#issuecomment-794109265 could solve for admissions.

pyrello:
I believe that the decision we made regarding this was to go a different route instead of the config ignore. We are going to change the menu defaults as so:
* Change `horizontal` menu block to have a max depth of 3.
* Change `megamenu` menu block to have a max depth of 2.
* Leave `toggle` menu block with unlimited max depth.

mark-bennett-uiowa:
Created #3228 in order to solve for the latest decisions on this fix.

mark-bennett-uiowa:
I still feel this would be helpful, as we may run into an instance where a site needs a default other than the defaults set in #3226 . Do we need to make this editable by users other than admins.

bspeare:
I would vote to close this issue.   As illustrated in @mark-bennett-uiowa's criss-cross effect screenshots, I don't think Superfish can handle depths greater than 3 levels.

mark-bennett-uiowa:
@pyrello @bspeare @joewhitsitt @briand44 are we good to close this issue? I would like to allow some flexibility for admins to handle this, but recent releases of the new standard depths and hiding menus features have taken care of most foreseeable issues.

richardbporter:
Sounds like this can be closed. Reopen if not.

PR #3187, opened 2021-03-05
trim trailing slashes from SiteNow request form URL value

### LINKED ISSUE #3176: Sitenow request form URL trailing slash error message is misleading
# Problem
The error message when submitting a URL with a trailing slash is not great:
image

# Steps to reproduce
- Enter a valid URL with a trailing slash. 
- See error related to path.

# Proposed solution
Trim any trailing slash automatically before running validation.

https://www.php.net/manual/en/function.rtrim

joewhitsitt:
fixes #3176

## To test

- checkout **main** and sync sitenow.uiowa.edu
- Fill out the form here with some different URLs (some including a trailing slash e.g. https://foo.uiowa.edu/): https://sitenow.local.drupal.uiowa.edu/request-site
- See the errors, see the problem.
- checkout this branch, pull, `drush @sitenow.local cr`, refresh the form and try the trailing slash URL again. It should pass.

Reroute email should prevent these emails from reaching its-web.


GaryRidgway (commented):
I was able to submit a new site with the url `https://yellow.lab.uiowa.edu/` and `https://yellow.lab.uiowa.edu////////////`.

these were both accepted, but when i checked the requests page the output was this:

Screen Shot 2021-03-05 at 4 08 09 PM

is this the intended result? if so, then I will approve.

joewhitsitt:
Darn, you got me @GaryRidgway. Not ideal that multi slash issue is, but I think we just want to deal with the common copy/paste from the browser issue where there is only one /. @richardbporter thoughts?

richardbporter:
I guess thats a valid URL...

richardbporter:
I'm not able to reproduce that. It trims all slashes for me.

joewhitsitt:
for the validation it trims, but the submission records it.

richardbporter:
Ah, good catch. We could change that. We're already setting some form values at the end of the validation.

richardbporter:
If not, I think this catching the most common copy/paste issue. This code was written a long time ago and might need an overhaul anyway. Cheers!

richardbporter (approved):
Yea this is fine for now. The URL might still have an extra slash or two, or two hundred, but we currently have that problem. We use the host in our command which won't include those. Plus you can type in https://uiowa.edu/////// and still get there so... 👍🏻

Issue #2987, opened 2021-02-01
Pharmacy news migration


briand44:
# Story

As a College of Pharmacy website owner, I would like news articles migrated from our existing website (https://pharmacy.uiowa.edu") to our new website (https://pharmacy.prod.drupal.uiowa.edu), so that site visitors can see past Pharmacy related news without the College of Pharmacy having to manually create hundreds of old news articles manually.

# Additional Info

>We would like to migrate all historical news articles from our current website to our new website, while also preserving backlinks. We have articles dating back to 2012.  See: https://pharmacy.uiowa.edu/news/archive

We need to create redirects from old paths to new as part of this migration.
 
>Ideally, we would also like hyperlinks removed from articles from 2016 and earlier. That is, links that appear within the articles themselves.

They are fine with just removing the links as part of the migration process, no need to provide a list of node id's that we removed links from.

# Proposed solution

richardbporter:
I think this is blocked by #3205. I've almost got that to a point where it can be merged.

richardbporter:
Sorry - poor planning on my part. There is a decent refactor of sitenow_migrate going on there.

richardbporter:
Here are the fields on their D7 article content type:

- Title
- Body
- Image	
- Migrated	
- Custom Person Directory	
- Department	
- Audience	
- Category	
- Person	
- Title for Gallery	
- Gallery	

Only image, title and body have a direct destination. The term reference could maybe migrate to tags. I'm not sure about the others. Migrating images is a little tricky since we're not importing all images.

Any idea what fields they want?

> They are fine with just removing the links as part of the migration process, no need to provide a list of node id's that we removed links from.

Just delete the anchor element? Or unlink it but leave the text?

briand44:
I will get clarification. My assumptions are we can ignore the following fields:

- Migrated
- Custom Person Directory
- Title for Gallery
- Gallery

I will ask about the reference fields:

- Department
- Audience
- Category
- Person

My understanding was unlink it but leave the text.

richardbporter:
Some additional tasks to be done in #3190 

- [x] Debug ~~and fix~~ obscure errors on import.
- [x] Double check sticky setting is migrating correctly. First article listed on /news is old.
- [x] Set an image style on body images - too big.
- [x] Fix teaser markup characters still being printed on some articles.
- [x] Report internal links

cory-skeers:
[...]

PR #3205, opened 2021-03-08
Area of Study migration

### LINKED ISSUE #2501: Migrate Areas of Study content
# Motivation
As part of #2365, "Migrate areas of study content from existing site into new site."

# Story

As a site owner of the Admissions site,
I want an automated migration solution to move existing Areas of Study content into the new website
so that the new content is available with minimal chance of human error that may otherwise occur during a manual migration.

# Proposed solution

Investigate a custom migration option for Areas of Study content.
If applicable, implement said migration option.

Blocked until Areas of Study content type is finalized.

cory-skeers:
Resolves #2501 

Content being imported:
- [x] Academic Groups
- [x] Colleges (#3236)
- [x] Areas of Study
- [x] Transfer Tips

# How to test
TLDR; https://admissions.dev.drupal.uiowa.edu/

- `blt sync --site admissions.uiowa.edu`
- `drush @admissions.local entity:delete node --bundle area_of_study`
- `drush @admissions.local entity:delete node --bundle transfer_tips`
- `drush @admissions.local entity:delete taxonomy_term --bundle academic_groups`
- `drush @admissions.local entity:delete taxonomy_term --bundle colleges`
- `drush @admissions.local en admissions_migrate --yes`
- `drush uli /admin/config/system/sitenow-migrate`
- Fill out the connection information for DevDesktop (assuming you have D7 admissions on DD local).
- `drush @admissions.local migrate:status` 
- `drush @admissions.local migrate:import --all`
- Go to https://admissions.local.drupal.uiowa.edu/areas-of-study
- Test site functionality around the content mentioned above.
- `drush @admissions.local migrate:rollback --all`

### Todo
 - [x] Track down error that is thrown once on `migrate:*` commands after a `cache:rebuild`.
   - 206d9e0
 - [x] Revert 31fb3c1 @richardbporter 
 - [x] Remove migrate_plus.migration_group.sitenow_migrate from other splits. This will be ignored and could contain sensitive info. Should not be tracked in config except defaults in sitenow_migrate split. 
 - [x] Test out a "standard" migration. @cory-skeers 
 - [x] Map each AOS with a stock featured image of pentacrest or similar. @richardbporter 
 - [x] Swap the title logic - https://github.com/uiowa/uiowa/pull/3205#issuecomment-801219039 @richardbporter 
 - [x] Delete basic pages in production using /colleges/college-name URL alias. @richardbporter 
 - [x] Fix character encoding issue with college term descriptions. @richardbporter 
 - [x] Attempt to fix broken links and/or create a report much like d7_page is doing.
 - [x] Ensure "Generate automatic URL alias" is unchecked but the initial /academic/aos-name URL alias is set. @richardbporter


richardbporter:
~~91887d7 will need to be reverted once we have a destination field for that.~~ #3213

richardbporter:
For the record, the "A-Z List" referenced in the old site is just the view page. That threw me.

richardbporter:
@cory-skeers @pyrello I've gone down a rabbit hole with this trying to understand all the different migrate core/contrib modules. I'm still confused on a few things but I'll summarize the changes here. I'm happy to revert or back out anything if this is ground already covered.

I wanted to be able to enable `migrate_drupal` to autoload the classes to leverage for migrations. I found an example of node_revision extending the Node source plugin. This  I've enabled it, exported config where necessary and reduced as much code duplication as possible. I also removed core default migrations in 507fc61. Following the docs, I was trying to handle as much processing in the YAML file as possible. It seemed like a lot to write queries in the source plugin and then map fields in the YAML file and possibly prepareRow.  All these changes get us to a point where we can, for example, run a basic taxonomy term migration straight from a YAML file (https://github.com/uiowa/uiowa/pull/3205/files#diff-dc0653605f585fcb551d49dea168c2d13433d45bee6ab44b9f341b54a411d1fb) - although that might not work for AOS or be necessary.

richardbporter:
I'm not I understand how to use the processMediaTrait's processImageField method.

richardbporter:
I'm also a bit lost on the migration path from text areas to the requirements paragraph type...

richardbporter:
I'm working on the majors field right now which might inform some of my previous comments.

pyrello:
Here is an example of `processImageField` in action: 
https://github.com/uiowa/uiowa/blob/a6e644196c48a97442bf3aed91d2e4f69ac99ccb/docroot/modules/custom/grad_custom/grad_migrate/src/Plugin/migrate/source/Article.php#L154

cory-skeers:
> I'm also a bit lost on the migration path from text areas to the requirements paragraph type...

For the migration into V2, there was a post-row-save event method (https://github.com/uiowa/uiowa/blob/a6e644196c48a97442bf3aed91d2e4f69ac99ccb/docroot/modules/custom/sitenow_migrate/src/EventSubscriber/PostRowSaveEvent.php#L150-L174). In that case, we were letting the node be saved without the content, then grabbing the section->paragraph that were autocreated on node creation and editing them.

Haven't looked into how the field is set here, but might be a similar path forward.

richardbporter:
Thanks. I _finally_ got something working for the degree/majors field in a custom process plugin. I tried to do it initially with a separate migration but couldn't get it working. Maybe at some point we can create specialized SQL sources that accept bundle and/or field name configuration for these types of field -> entity migrations. I still think migration dependencies and lookups are ideal.

Just logging that some of the D7 site majors follow a different pattern than most. These could be adjusted to add parentheses around the abbreviation or we could improve our logic.
```
[warning] Cannot split degree on abbreviation for AOS: Open Major.
[warning] Cannot split degree on abbreviation for AOS: Speech & Hearing Science.
[warning] Cannot split degree on abbreviation for AOS: Secondary Education.
[warning] Cannot split degree on abbreviation for AOS: Secondary Education.
[warning] Cannot split degree on abbreviation for AOS: Secondary Education.
[warning] Cannot split degree on abbreviation for AOS: Secondary Education.
[warning] Cannot split degree on abbreviation for AOS: Secondary Education.
[warning] Cannot split degree on abbreviation for AOS: Sport and Recreation Management.
[warning] Cannot split degree on abbreviation for AOS: Enterprise Leadership.
[warning] Cannot split degree on abbreviation for AOS: Public Health.
[warning] Cannot split degree on abbreviation for AOS: Public Health.
[warning] Cannot split degree on abbreviation for AOS: English and Creative Writing.
[warning] Cannot split degree on abbreviation for AOS: Global Health Studies BA, BS.
[warning] Cannot split degree on abbreviation for AOS: Global Health Studies BA, BS.
[warning] Cannot split degree on abbreviation for AOS: Social Justice.
[warning] Cannot split degree on abbreviation for AOS: Social Justice.
[warning] Cannot split degree on abbreviation for AOS: Sustainability Science .
[warning] Cannot split degree on abbreviation for AOS: Screenwriting Arts.
[warning] Cannot split degree on abbreviation for AOS: Education Studies and Human Relations.
[warning] Cannot split degree on abbreviation for AOS: Science Studies.
```

richardbporter:
Note that 31fb3c11db20d67283212ae2ea46baba3e298451 should be reverted.

richardbporter:
The error on the initial `migrate:status` seems to be caused by not using the default database key of `migrate`. I tried updating the migrate_plus group config but that didn't do it. I had to actually set it in local.settings.php. I think it can be safely ignored.

https://support.acquia.com/hc/en-us/articles/360050885633-Troubleshooting-the-No-database-connection-configured-for-source-plugin-error-during-a-website-migration

docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
```diff
@@ -82,22 +82,6 @@ public function __construct(EntityTypeManager $entityTypeManager, LoggerInterfac
     $this->entityTypeManager = $entityTypeManager;
     $this->logger = $logger;
     $this->connection = $connection;
-
-    // Switch to the D7 database.
```
richardbporter: @cory-skeers I'm trying to config-ignore the sitenow_migrate group so that sites can set DB connection info other than to DD. That change seems to have broken this code for some reason. I moved it to the d7_page case statement below for the time being. I'm not sure why it can't find the database key anymore.

I think we'll need to test out a "standard" site migration with this PR. Would you be able to head that up?

cory-skeers: I'll take a look!

pyrello:
I noticed when viewing the Admissions site on dev that the Transfer Tips were being imported as belonging to no user and so getting assigned to user 0. I think the syntax of my update is correct, but I wasn't able to test it because I wasn't sure how to get the database settings working with the new setup.

richardbporter:
Thanks!

docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php
richardbporter: I think this was caused by the migrate_drupal `migration_plugins_alter` hook doing some wonky stuff. It also caused that DB error mentioned in the thread. I unset it in 206d9e0 and this code seems to be working now in the constructor.

richardbporter:
I think this is ready for an initial review, at least the content on dev.

richardbporter:
I just re-ran the migration in dev.

One thing I noticed is that if you run any Drush migrate commands before saving the sitenow_migrate config form, the sitenow_migrate group config entity is created but missing some keys. That causes our migrations to not be found. I had to delete it and resave the form. Not sure of a better way to handle that.

pyrello:
Observations (not necessarily meant to be actionable):
* Embedded links are broken. I was thinking that we could potentially check against the migration nid list to fix these, but that list gets generated as the migration happens, I think, so this would have to be something that happened after the migration ran.
* Some weird markup on [this page](https://admissions.dev.drupal.uiowa.edu/academics/engineering): `<a href="/node/194">Biomedical&nbsp;&#8203;</a>`. However, this appears to be in the original too, so not much that can be done about that.
* Transfer Tips look great!
* Even though the URL is custom, the "Generate automatic URL alias" box is checked, so the next time the page gets saved, it is going to change the URL: https://admissions.dev.drupal.uiowa.edu/node/1326/edit

richardbporter:
Thanks for taking a look!

> Observations (not necessarily meant to be actionable):
> 
> * Embedded links are broken. I was thinking that we could potentially check against the migration nid list to fix these, but that list gets generated as the migration happens, I think, so this would have to be something that happened after the migration ran.

Yes there is no processing of textarea fields happening right now. I can see what is happening for the d7_page migration in `sitenow_migrate\EventSubscriber\MigratePostImportEvent`. I _could_ copy/paste some of that to work with AOS potentially. It'd be nice if that functionality was refactored to be usable by other migrations, especially since each migration class' `postImportProcess` method is called by sitenow_migrate automatically. That is nice and might be nice for other events as well.

> * Some weird markup on [this page](https://admissions.dev.drupal.uiowa.edu/academics/engineering): `<a href="/node/194">Biomedical&nbsp;&#8203;</a>`. However, this appears to be in the original too, so not much that can be done about that.

Good catch. I guess we'll have to check to see if these types of things are worth automating vs. manually fixing.

> * Transfer Tips look great!

Good. 

> * Even though the URL is custom, the "Generate automatic URL alias" box is checked, so the next time the page gets saved, it is going to change the URL: https://admissions.dev.drupal.uiowa.edu/node/1326/edit

I may not understand this. Do we need to create a different URL alias other than `/academics/aos-name` on import? Or just make sure that checkbox is unchecked?

cory-skeers:
> * Embedded links are broken. I was thinking that we could potentially check against the migration nid list to fix these, but that list gets generated as the migration happens, I think, so this would have to be something that happened after the migration ran.

There's a start of this in `sitenow_migrate` at https://github.com/uiowa/uiowa/blob/6ec2fda1aab34b6b19fa34f30866c141d013fa90/docroot/modules/custom/sitenow_migrate/src/EventSubscriber/MigratePostImportEvent.php#L211

A post migrate event listener that fetches mappings, queries for embedded links whether they are in the `/node/#` format, relative links, or absolute paths to the same site. It hasn't been run on a migration in awhile, and could use testing/cleanup, but it's a start.

pyrello:
> > Even though the URL is custom, the "Generate automatic URL alias" box is checked, so the next time the page gets saved, it is going to change the URL: https://admissions.dev.drupal.uiowa.edu/node/1326/edit

> I may not understand this. Do we need to create a different URL alias other than /academics/aos-name on import? Or just make sure that checkbox is unchecked?

I'm pretty sure that we just want the checkbox unchecked. Admissions has emphasized that they want to keep the existing URL paths as much as possible.

I'm guessing that this is a property on the node that could potentially be set to a static FALSE value in the migration yaml file.

cory-skeers:
> Yes there is no processing of textarea fields happening right now. I can see what is happening for the d7_page migration in `sitenow_migrate\EventSubscriber\MigratePostImportEvent`. I _could_ copy/paste some of that to work with AOS potentially. It'd be nice if that functionality was refactored to be usable by other migrations, especially since each migration class' `postImportProcess` method is called by sitenow_migrate automatically. That is nice and might be nice for other events as well.

There was a start toward abstracting these sorts of events into:
https://github.com/uiowa/uiowa/blob/6ec2fda1aab34b6b19fa34f30866c141d013fa90/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php#L264-L269

The post import event subscriber calls this function (which doesn't do anything in the base), that can then be extended in individual Source classes as a way to tap into the event subscribers without blowing up the switch cases.

I think we'd want to move the link replacement function(s) into either the BaseNode or a trait, and then they can be called by the postImportProcess with any of the necessary mapping designations.

richardbporter:
> 'm guessing that this is a property on the node that could potentially be set to a static FALSE value in the migration yaml file.

You would think so. I'll look into that.

pyrello:
@richardbporter Here is the logic for how to handle the alternate title: https://github.com/uiowa/uiowa/issues/2501#issuecomment-796760748

richardbporter:
I've updated the todo list in the PR description. Feel free to jump in with any of those. I put my name next to one but need to switch gears after that.

richardbporter:
>I think we'd want to move the link replacement function(s) into either the BaseNode or a trait, and then they can be called by the postImportProcess with any of the necessary mapping designations.

Ah, I see. Moving them to BaseNode and then calling it in the d7_page migration `postImportProcess` method makes sense. We'll just need to make the base method work with other fields. Maybe pass in a field name or something.

cory-skeers:
> Ah, I see. Moving them to BaseNode and then calling it in the d7_page migration `postImportProcess` method makes sense. We'll just need to make the base method work with other fields. Maybe pass in a field name or something.

Mmhmm, probably just pass in the field would be easiest. Only other part (maybe?) would be making sure we look at the correct `migrate_map_*` table or tables. If it's in the source plugin, we can make the mapping out of its id, but then that limits its ability to make connections to other content types (like both articles/pages in standard migrations, other content types for custom migrations). Could simply pass an array of migrate map titles too, would probably be the most straightforward? Not very elegant, but should work...

cory-skeers:
Getting some negative values for Unprocessed that I haven't seen before when testing a standard `sitenow_migrate` migration, and not sure where they're coming from. Doesn't seem to impact things, and the Total and Imported columns are still accurate for checking on progress.

richardbporter:
> Getting some negative values for Unprocessed that I haven't seen before when testing a standard `sitenow_migrate` migration, and not sure where they're coming from. Doesn't seem to impact things, and the Total and Imported columns are still accurate for checking on progress.

I saw that a couple times but always when some part of a migration didn't work correctly. The migration didn't fail necessarily but it seemed to jack up the counts.

cory-skeers:
> I saw that a couple times but always when some part of a migration didn't work correctly. The migration didn't fail necessarily but it seemed to jack up the counts.

Okay! Not all the migrations registered correctly, so there were other issues going on there. I'll disregard for now, and see if it clears up.

richardbporter:
> One thing I noticed is that if you run any Drush migrate commands before saving the sitenow_migrate config form, the sitenow_migrate group config entity is created but missing some keys. That causes our migrations to not be found. I had to delete it and resave the form. Not sure of a better way to handle that.

I wonder if our custom config form should just overwrite or re-import this config every time it is saved?

I don't love using CM for the DB credentials. The migrate_drupal UI using the private tempstore service for DB credentials which might be more secure than config. I'm not sure how we might use that though. I added Drush `sql:sanitize` integration to delete the sitenow_migrate migration group config so it should get removed when DBs are synced around.

dnkemper:
@richardbporter WHAT ARE YOU DOING? SLEEP!

richardbporter (approved):
I think I've got this to a point where the sitenow_migrate migrations are at least running and reporting the correct counts. There will certainly be follow-up issues for those and possibly admissions but this PR is getting quite large. I'd like merge it as-is as an incremental improvement and start working on #2987 

I had a hard time following a lot of the regex and SQL in the base migrations. If possible, I think we should try use the Drupal APIs as much as possible. 

Thanks for everyone's help on this!

pyrello:
> I had a hard time following a lot of the regex and SQL in the base migrations.

I'm glad to know that I'm not the only one :)

PR #3225, opened 2021-03-10
Fix for existing accordions

cory-skeers:
Followup to #3154 

Existing accordions won't have values for multiselect/singleselect (and accordion items won't have automatic open values), leading to a warning.

image

Updates here involve switching to magic getters to help handle the null value case, and slight adjustment to multi-select logic to help it default to multi if no value is available, to match existing functionality.

# How to test

Sync a site that has previously had accordions created (`blt ds --site=sandbox.uiowa.edu` works)
Log into the site (`drush @sandbox.local uli`)
Visit the page(s) which had previously created accordions (http://sandbox.local.drupal.uiowa.edu/tables)
See that there is no warning/error
Check the accordion functionality, and see that they are multi-selectable, and that no accordions are automatically opened
Go into the page's layout builder, edit the accordion (switch to single-select and/or update an accordion item to automatically open)
Check that everything saves correctly, and that the accordions function as expected on page view.


joewhitsitt:
yikes, pulling down to test.

joewhitsitt (approved):
Though I was not able to replicate the warning in prod, this code looks fine and the functionality still works locally.

PR #3229, opened 2021-03-11
Feature: Add pretty title to Admissions Area of Study content type

### LINKED ISSUE #3213: Add alternate name field to Areas of Study
# Story

As a prospective student I want to be able to find areas of study by searching for more common terms so that I can find them without having to know what they are called.

# Proposed solution

Add an Alternate name text field to the Area of Study content type. This field will be used for building the Area of Study listing.

pyrello:
# Updates
* Resolves #3213 
* Cleans up the `admissions` split with regard to the Area of Study content type.

# How to test

* Observe creation of two config items related to `field_area_of_study_pretty_title` and updating of 4 config files like `core.entity_*_display.node.area_of_study.*` and the `admissions` config split.
* `drush @admissions.local uli node/1141/edit`, add a Pretty title to Anthropology, and Save.
* Observe that the pretty title does not show up on the page.
* Observe that the pretty title does not show up on the PDF: http://admissions.local.drupal.uiowa.edu/print/pdf/node/1141
* `drush @admissions.local csex admissions` and `git status` to observe there are no changes to config based on changes to `admissions` split.

### DIFF
--- a/config/admissions.uiowa.edu/config_split.config_split.admissions.yml
+++ b/config/admissions.uiowa.edu/config_split.config_split.admissions.yml
@@ -31,33 +31,6 @@ graylist:
   - core.entity_view_display.paragraph.admissions_card.default
   - core.entity_view_display.paragraph.admissions_requirement.default
   - core.entity_view_mode.media.grid_gallery
-  - field.field.node.area_of_study.body
-  - field.field.node.area_of_study.field_area_of_study_academic_gp
-  - field.field.node.area_of_study.field_area_of_study_career
-  - field.field.node.area_of_study.field_area_of_study_college
-  - field.field.node.area_of_study.field_area_of_study_competitive
-  - field.field.node.area_of_study.field_area_of_study_comp_txt
-  - field.field.node.area_of_study.field_area_of_study_course_work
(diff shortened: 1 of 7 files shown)


richardbporter:
Screen Shot 2021-03-11 at 10 22 24 AM

Was Alt Description ever discussed?

richardbporter:
> Screen Shot 2021-03-11 at 10 22 24 AM
> 
> Was Alt Description ever discussed?

Nevermind.

richardbporter:
This works but the issue says it should accept multiple values. Is that no longer the case? If so, should only the first value be migrated?

pyrello:
Per discussion yesterday, only one value is needed. I'll ask regarding how to handle that for migration.

cory-skeers:
> This works but the issue says it should accept multiple values. Is that no longer the case? If so, should only the first value be migrated?

Worth double checking or getting clarification, but if I remember correctly the D7 field allowed multiple values but never had more than 1 value to a node.

pyrello:
I updated the issue to remove the requirement for multiple values. It was only added based on inspecting the D7 solution. It is clear from the conversation with the customer yesterday that this should be a single value field.

richardbporter (approved):

PR #3227, opened 2021-03-11
I2748 wysiwyg tables styles

GaryRidgway:
Tracking for https://github.com/uiowa/uiowa/issues/2748, previously housed in the branch `i468-uids-tables-additional-styles`, ranamed for its new issue name. This PR integrates the changes from https://github.com/uiowa/uids/pull/490 in to uids_base.

# How to test:
1. pull down this PR
2. Sync `sandbox.uiowa.edu` locally.
3. Navigate to http://sandbox.local.drupal.uiowa.edu/tables
2. Test that the following are true:
    1. Adding a caption to a table will show a grey caption above the table itself.
    2. Adding the `table--gray-borders` class to a table will put gray borders on a table.
    3. Adding the `table--hover-highlight` class to a table allows rows to be highlighted on hover.
    4.  Adding the `table--width-default` class to a table allows a table to shrink to only as wide as necessitated.

PR #3231, opened 2021-03-11
Hotfix: Display table outline by default in WYSIWYG

pyrello:
* Adds table border update from #3227 

# How to test

* `blt ds`
* `drush @default.local uli node/add/page`
* Add a table to the WYSIWYG area and observe that you can see its outline.


bspeare (approved):
Reviewed, not tested.  Looks good!

PR #3243, opened 2021-03-12
Admissions: Student Profile block revisited

### LINKED ISSUE #3138: Fix Admissions Student Profiles block
# Problem

https://iowaweb.slack.com/archives/C0164FNN1PV/p1614281335067900
The student profile block still displays even if there is no content to fill it.

The student profile block also doesn't display profiles at random. 

# Steps to reproduce

https://admissions.prod.drupal.uiowa.edu/student-profiles/aasthika-das
# Proposed solution

Could potentially follow Joe's fix?!
 - if no data, don't display student profile view block
 - Don't show student profile image of the full profile node you're currently viewing
https://github.com/uiowa/uiowa/blob/main/docroot/sites/admissions.uiowa.edu/module[…]--views-block--areas-of-study-block-block-information.html.twig 
- Display student profiles at random. If viewing a full node student profile, that profile shouldn't be displayed in the block. It should be random and display other student profiles.

### LINKED ISSUE #3232: Admissions Student Profiles Block with Filters
# Story

As an Admissions content editor, I want the ability to place a filtered student profile block on any page.

# Additional Info

Need the ability to filter by the following fields:
- Gender
- Student Attributes
- State
- Country if Not US

They would also like the ability to determine how many items are displayed.

They would also like the ability to control the title. It can default to Student profiles with a link to the student profiles page but they should be able to override the text. The link should always remain the same.

The design of this should be similar to the random student profiles block.

# Proposed solution

pyrello:
# Updates

* Adapts parts of #2702 to solve for a more limited use case of providing filtering options in the student profile views block.
* Resolves https://github.com/uiowa/uiowa/issues/3138
* Resolves https://github.com/uiowa/uiowa/issues/3232

# How to test

- Pull and sync admissions
- Apply and Request Info Buttons should still be placed on AoS and working.
- Apply button should still be placed on student profiles and working.
- Provide featured and thumbnail images for a handful of student profiles.
- Block should still be placed on the student profile template with the same defaults as PROD. (9 items, "Student Profiles", no filtering)
- Block should be available to place by webmasters on page content type
- The filters (below), number of items and headline should be available in the block configuration.
  - Gender
  - Student Attributes
  - State
  - Country if Not US
- The block should not display if there are no valid profiles. Check this by filtering the content overview to just student profiles and unpublish all.
- Pick a student profile and publish it. Visit the full node. Block should still not display because the current profile should not be included.
- Pick another profile and publish it. The block should be visible on this and the previous node showing the other profile.
- Background image fallback for student profiles without an thumbnail.
- View should display random but is cached so you will need to run `cr` and reload the page to test or place different blocks with the same filtering
- Check other views allowed to be placed by webmaster because the views label is now hidden and replaced by the headline field.


bspeare:
Pulling down to test.

bspeare (approved):
All testing criteria worked!  I added a few adjustments to make the button margins and headline styles work for each layout.  I did notice the icon appears if you leave the Headline empty on the view block.

Screen Shot 2021-03-16 at 4 06 48 PM

pyrello:
I'm going to test this as well, since I'm curious how it turned out.

joewhitsitt:
>I did notice the icon appears if you leave the Headline empty on the view block

I can fix that

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -141,6 +143,10 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
       unset($form['settings']['label']);
       unset($form['settings']['label_display']);
       unset($form['settings']['admin_label']['#title']);
+      // Hide ones implemented by Views because ListBlock has headline.
+      unset($form["settings"]["views_label_fieldset"]);
+      unset($form["settings"]["views_label_checkbox"]);
+      unset($form["settings"]["views_label"]);
```
pyrello: I'm wondering if this should be moved to `ListBlock::blockForm()`? Also, could we change these to [single quotes](https://www.drupal.org/docs/develop/standards/coding-standards#quotes)?

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -702,3 +708,12 @@ function layout_builder_custom_preprocess_media(&$variables) {
     ] + $variables['attributes'];
   }
 }
+
+/**
+ * Implements hook_views_plugins_display_alter().
+ */
+function layout_builder_custom_views_plugins_display_alter(&$displays) {
+  if (!empty($displays['block']['class']) && $displays['block']['class'] == Block::class) {
+    $displays['block']['class'] = ListBlock::class;
```
pyrello, line 718: Just want to reiterate that the implication of this is that all the views block displays will be handled by this class and wanted to check that it was tested with this in mind?

pyrello:
Just checking, is it the intended behavior for this new block if it is placed on a page like so:

Screen Shot 2021-03-16 at 5 07 34 PM

The padding around the edges looks strange to me.

bspeare:
> Just checking, is it the intended behavior for this new block if it is placed on a page like so:
> 
> Screen Shot 2021-03-16 at 5 07 34 PM
> 
> The padding around the edges looks strange to me.

That is the intended padding.  Probably not recommended to be stacked under a card. :)

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: I can move it. I think I did that first but went this direction because of the similar overrides above it. Yeah I can switch to single quotes. Need to figure out how to change xDebug to copy path using single quotes.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: >Check other views allowed to be placed by webmaster because the views label is now hidden and replaced by the headline field.

It was a test instruction. The ones I tested could be abused (and addressed in follow-ups) but nothing was "broken" that I saw. I will do another check while making the other change.

pyrello: No worries. I just wanted to double check since I hadn't been entirely sure about this as the "right" approach when I was working on it initially.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: Changed to single quotes. Can't seem to target these "settings" in blockForm. Added a todo to investigate at a later date.

PR #3265, opened 2021-03-22
Hotfix - Admissions - ListBlock

joewhitsitt:
To make student profiles block revisited possible, we are hijacking views block with a custom one which was hiding any exposed filters. This cause the /counselors filters to not render. We can't make this assumption so for now, I have removed this recent change and updated the student profiles block to hide the filters similar to articles/people blocks until we can figure out a better way.

## To Test

- pull and sync admissions
- go to http://admissions.local.drupal.uiowa.edu/counselors and see the territory dropdown.
- go to http://admissions.local.drupal.uiowa.edu/student-profiles/bhavya-vats and see that there are no exposed filters.


dnkemper:
I can review!

dnkemper (approved):
Looks great! 👍

PR #3271, opened 2021-03-25
Feature: Added menu per role module

### LINKED ISSUE #3188: As a developer I can hide specific menu items without disabling them
# Story

As a developer I can hide specific menu items without disabling them. I need to hide one menu item from the main navigation, but still show its children in a side menu.

This all stems from a specific need I have for the Social Work site, specifically for the Field Education section. This Field Education section contains information for both undergraduate and graduate students. On their existing site, they have actually duplicated the content underneath both areas, which is leading to inconsistent content. On their new site, I have combined all of this information into one section (Field Education) and created a custom link to this section underneath both the UG and Grad menus:

Screen Shot 2021-03-05 at 12 30 39 PM

The problem with this is the Field Education section of the menu needs to be hidden so as to not show in the Main Navigation display. This could easily be solved by "disabling" the menu item from the main menu (if that was the only item), but this item has children of its own. And it gets worse: If I disable the parent Field Education item, then all of its children under up floating to the Alumni section, where it definitely doesn't belong.

Screen Shot 2021-03-05 at 1 09 29 PM

Furthermore, the Field Education section's children need to stay visible in their side menu:

Screen Shot 2021-03-05 at 1 06 18 PM

So, Im looking for a way to "hide" the parent menu item (Field Education) from being displayed in the Main Navigation (and in turn its children), while still allowing the side menu for these sections to remain intact.

# Proposed solution

Menu Item Role Access may be a solution: [https://www.drupal.org/project/menu_item_role_access](https://www.drupal.org/project/menu_item_role_access)

Menu Per Role may be a solution: [https://www.drupal.org/project/menu_per_role](https://www.drupal.org/project/menu_per_role)

Menu Item Visibility does not look to be a solution, as it looks to be a dead project: [https://www.drupal.org/project/menu_item_visibility](https://www.drupal.org/project/menu_item_visibility)

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3188. 

# Work remaining to be done in this pr:

- [x] Only show checkboxes for admins (Current permissions overriden by `Menu Admin per Menu` module)
**or** 
- [x] Only show one checkbox to hide for Anonymous users (For all roles except admin).
- [ ] Fix superfish hover/arrow issues

# How to test
- `composer install`
- `blt ds --site=sandbox.uiowa.edu`
- As an admin, you can edit any menu item and hide the item to anonymous users. As an admin you shouldn't see any other options. Save and open the site in an unauthenticated session.
- As a webmaster, editor, etc. You can't see the hide functionality on the menu item but you can still see the menu link within admin/structure/menu. On the hidden menu item you should see a message saying that the menu item you previously set is hidden.
- You can add the `administer menu link visibility` permission to editor, webmaster, publisher and they should then see the hide functionality. This permission is not set as a default for any role. This functionality is limited to admins and site/feature splits right now.
- Even as an admin, you shouldn't see the "show" functionality.


bspeare:
This is the interface that webmasters will see if they want to hide a link.    Did we want to alter this interface and if so what method for altering it are we thinking?

Screen Shot 2021-03-25 at 8 37 52 AM

bspeare:
@mark-bennett-uiowa What sites would you need to use this on?  Only Social Work?

joewhitsitt:
it is probably going to be easier to do it for all sites, but i'd really want the functionality ironed out (straight forward ui, no superfish flyout issue, etc.) so that it is there for advance users but one that doesn't require a lot of support

mark-bennett-uiowa:
@bspeare Yes, I currently only need this on Social Work. However, @joewhitsitt is right, and Im sure that this will end up getting used elsewhere in the future.

bspeare:
Since admissions doesn't need this module immediately, should we just add this module to the service and only make it available for admins?

bspeare:
Looks like the `Menu Admin per Menu` module overrides any specific access permissions that this module provides so we are unable to just add this to the service for Admins.

bspeare:
Adding the blocked label until we can make changes to make this module work with the service:
- Only show checkboxes for admins
or 
- Only show one checkbox to hide for Anonymous users
- Fix superfish hover/arrow issues

joewhitsitt:
> Adding the blocked label until we can make changes to make this module work with the service.

Are these covered by other issues or are they todo items for this PR?

bspeare:
@joewhitsitt I updated pr description to clarify that.

joewhitsitt:
Took a look at this. I don't love this module but to move it forward, I removed access to the `Roles able to see the menu link` field since all checked or all unchecked have the same effect and that by default, a menu link should be visible if the content is published. Since the original issue is hide certain menu items, I left `Roles not able see the menu link` but wrapped it with a new permission in the uiowa_core module.

Since we don't know all the roles on a given site, controlling the options seemed tough and webmaster can already assign all other roles (non-admin).

The one gotcha with this I see is that if an additional role is added to a website with existing visibility settings, that role will see all menu items initially even if it is lower in the role hierarchy.

This is easy to rip back out. Permissions config will have to be updated if we want to give webmasters and other roles permission to use this by default.

joewhitsitt:
Is this the superfish/arrows issue? Children under academics are hidden.
Screen Shot 2021-04-13 at 3 52 00 PM

Not sure I see a good way around that. I am also concerned the children `<ul>` exists but is empty.

bspeare:
@joewhitsitt Yes, and I think those arrows still move if you hover over them?

joewhitsitt:
yes, they still move as well

bspeare:
@joewhitsitt This looks good.   As long as it works only for admins I think we should move forward with this so @mark-bennett-uiowa can use this with his sites.     There are limitations to the usage with the example shown above, but I think it works to solve mark's issue.  

@mark-bennett-uiowa can you review this?

mark-bennett-uiowa:
This works for my needs on the Field Education section for the SSW site. Still visible while logged in for admins, even with Authenticated Users selected, but works properly when Anonymous Users is selected.
Screen Shot 2021-04-15 at 12 08 11 PM

Screen Shot 2021-04-15 at 12 08 22 PM

mark-bennett-uiowa:
With the Authenticated Users checked, masquerading as a Publisher shows that I cannot see the menu under Edit Menu:

**As Admin:**
Screen Shot 2021-04-15 at 12 19 46 PM

**As Publisher:**
Screen Shot 2021-04-15 at 12 20 46 PM

joewhitsitt:
> With the Authenticated Users checked, masquerading as a Publisher shows that I cannot see the menu under Edit Menu:

So it sounds like unsetting all the options except for Anonymous might be the MVP here? Seems like a lot of module that we aren't using though.

joewhitsitt:
Reduced to just anonymous and added a message that editors can see.

Screen Shot 2021-04-21 at 10 24 13 AM
Screen Shot 2021-04-21 at 10 24 34 AM

joewhitsitt:
If we punt on the superfish arrows and the a11y surrounding an empty UL then I think this is ready for review.

mark-bennett-uiowa (approved):
Works for the Social Work menu need I have. Verified that only admins have the ability to hide the menu item:
Screen Shot 2021-04-25 at 12 54 23 PM

and that other users with menu editing access receive the Warning message:
Screen Shot 2021-04-25 at 1 05 48 PM

Update successfully hid the menu item for anonymous users:
Screen Shot 2021-04-25 at 1 01 46 PM

joewhitsitt:
I created a follow up issue for the superfish arrow/empty ul issue. https://github.com/uiowa/uiowa/issues/3406

PR #3279, opened 2021-03-26
Feature: Fixing sass mixin import for admissions_core

bspeare:
We were importing the wrong file for headings.  This pr fixes that with the mixin file instead.


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Headlines still look like headlines on Admissions custom stuff

PR #3280, opened 2021-03-26
Admissions - AoS Search Improvements

### LINKED ISSUE #3258: Admissions - AoS Listing/Search Improvements
## Issue

https://github.com/uiowa/uiowa/pull/3249 made things functional, but improvements can be made.

## Proposed Improvements
- [x] I think we are going to need to work on the design a bit. When I added a college, the "Search" button was pushed to a new line, making the whole box very big.
- [x] I think we should consider adding a reset button. The form is somewhat hard to use without it.
- [x] Probably need something other than '-Any-' as a placeholder for those last two boxes. The colleges field takes up a lot of space.
- [x] Can we break the filters into two lines and have it auto-submit to remove the search?
- [x] Is it possible to have auto-submit functionality like D7 site: https://admissions.uiowa.edu/academics/undergraduate-areas-of-study?
- [x] Can we use something like https://www.drupal.org/project/better_exposed_filters to make the filter more dynamic?
- [x] Could we have one search bar across the top of the page with Program Type, Academic Group, and College filters as selects/checkboxes in a sidebar for further filtering? https://www.arizona.edu/degree-search/majors and https://admissions.umich.edu/academics-majors/majors-degrees are examples with sidebar filter.
- [x] It was also requested that we only have two columns for the results (Name and College) and having the other column of information display under the name similar to how it is on their existing site.

joewhitsitt:
Resolves #3258 

Visually approved by Admissions

## To Test

- pull
- `composer install && yarn workspace admissions_core build && blt ds --site=admissions.uiowa.edu`
- Test the search box on the front page. It should populated the exposed filter on the areas of study.
- The filters should all work if the content exists. You can change your admissions blt file to pull from dev if you want more content.

To those who saw the white screen in demo, manually trying to filter values via URL is not something you should do...
https://www.drupal.org/project/drupal/issues/2932604, https://www.drupal.org/project/drupal/issues/2901127
With that said, Drupal should really do something to prevent this from happening...


joewhitsitt:
>I added the better filtering module to the view and it seems to work. Only problem is that if you pause, it automatically filters but takes your curser off of the text box. Not sure if that’s ideal

I haven't tested this yet, but I recall this when working on the people filtering PR which might make a difference in the behavior you observed:

>With Drupal 9, jQuery UI is no longer included in core and will need to be added through contrib modules. @see https://www.drupal.org/node/3067969
The jQuery UI Touch Punch library will not automatically be installed with this module, and will either need to be downloaded manually and placed in the libraries folder, or included with the path in https://www.drupal.org/project/jquery_ui_touch_punch/issues/3159222

- https://www.drupal.org/project/better_exposed_filters

dnkemper:
Thank you so much, @joewhitsitt! That's incredibly helpful!

joewhitsitt:
>So I applied another that keeps the curser on the textbox. Not sure if appropriate or not? Only issue is that it wasn't working with Brave which made me go crazy for a bit. Works with Chrome and Firefox though..

I can't replicate the issue in Brave.

I did notice that `views-exposed-form--minimal` is being applied after the ajax which makes me think we need to change how https://github.com/uiowa/uiowa/blob/main/docroot/sites/admissions.uiowa.edu/modules/admissions_core/admissions_core.module#L154 is applied.

I also noticed that when typing in the search input, it autosubmits and sometimes chops off the letter I just typed. I can't reproduce it consistently though.

I think we need some no results text as well. Missed that the first time around.

dnkemper:
@joewhitsitt that is mildly concerning about Brave 😬 I spent quite a bit of time and just couldn't get it to work. You're magic. Also, dang. Didn't experience the text cutting off.. is setting the BEF time to 0 messing that up? I was experiencing it cutting off when set to 500 though. BUT that was before I added the second patch. I also thought it was supposed to auto submit 😰

dnkemper:
@joewhitsitt P.S. the no results text is a brilliant idea!

joewhitsitt:
Don't know if autosubmit will work because this affects the homepage and we don't want folks whisked away before the finish typing.

The homepage search box should probably be refactored to just be a custom form otherwise we are going to keep running into these gotchas.

joewhitsitt:
>The homepage search box should probably be refactored to just be a custom form otherwise we are going to keep running into these gotchas.

I am going to work on this tomorrow as part of this PR.

pyrello:
Going to test this.

pyrello (approved):
Looks good to me. Nice work on this!

PR #3288, opened 2021-03-29
Fix broken links in AOS migration.

### LINKED ISSUE #3275: Replace broken links in admissions AOS miration postImportProcess.
There are 1500+ broken links in the AOS migration. 

The report is happening in the `postImportProcess` method now. I think we can stick with that and expand it. We can replace some using the migration map and others using a manual map of source -> destination NIDs that admissions created. 

We should double check all fields are being scanned that need to be.

richardbporter:
Resolves #3275 

### Testing
- Go to https://admissions.dev.drupal.uiowa.edu/ and review AOS for broken links. Note that we are only fixing internal links. 
  -  https://admissions.dev.drupal.uiowa.edu/business is a good test case.
  - Other AOS can be found on https://admissions.dev.drupal.uiowa.edu/areas-of-study

Developer testing
```
blt sync --site admissions.uiowa.edu &&
drush @admissions.local entity:delete node --bundle area_of_study &&
drush @admissions.local entity:delete node --bundle transfer_tips &&
drush @admissions.local entity:delete taxonomy_term --bundle academic_groups &&
drush @admissions.local entity:delete taxonomy_term --bundle colleges &&
drush @admissions.local en admissions_migrate --yes &&
drush @admissions.local uli /admin/config/system/sitenow-migrate
```
- Configure for local DD admissions site.
- `drush @admissions.local ms`
- `drush @admissions.local mim --all`

Worth noting that running the migration again will attempt to fix broken links again which at best won't work or at worst replace them incorrectly. So after the step above:
- `drush @admissions.local pmu admissions_migrate`
- `drush @admissions.local cim`


richardbporter:
Something is hosed now after implementing this manual map logic. I'm not sure whats going on yet.

richardbporter:
Seems like `postImportProcess` is getting called twice and the second time it tries to map already mapped links. That may just be one problem.

richardbporter:
> Seems like `postImportProcess` is getting called twice and the second time it tries to map already mapped links. That may just be one problem.

@cory-skeers Any idea why this would be happening? I hacked my way around it for now.

richardbporter:
The second problem is that text area fields are not being saved correctly. The text format is written out to the field value. This does not seem to be working: https://github.com/uiowa/uiowa/pull/3288/files#diff-0b0817f5e868309d8ac3df7e46550613f61ffd3929bb9d333e537e7b3d441356R182

cory-skeers:
> The second problem is that text area fields are not being saved correctly. The text format is written out to the field value. This does not seem to be working: https://github.com/uiowa/uiowa/pull/3288/files#diff-0b0817f5e868309d8ac3df7e46550613f61ffd3929bb9d333e537e7b3d441356R182

Is the format being saved correctly, even if it's also being written to the value?

Haven't been able to test it, but I think `getString()` from a text field might return both the value and the format, and then it's getting included in the following processing/re-setting.

cory-skeers:
> > Seems like `postImportProcess` is getting called twice and the second time it tries to map already mapped links. That may just be one problem.
> 
> @cory-skeers Any idea why this would be happening? I hacked my way around it for now.

Not sure, but taking a look.

richardbporter:
> > The second problem is that text area fields are not being saved correctly. The text format is written out to the field value. This does not seem to be working: https://github.com/uiowa/uiowa/pull/3288/files#diff-0b0817f5e868309d8ac3df7e46550613f61ffd3929bb9d333e537e7b3d441356R182
> 
> Is the format being saved correctly, even if it's also being written to the value?
> 
> Haven't been able to test it, but I think `getString()` from a text field might return both the value and the format, and then it's getting included in the following processing/re-setting.

Doh! Thanks!

richardbporter:
I might just roll with this for now. It sounded like https://www.drupal.org/project/migrate_tools/issues/3130358 but we're on 5.0. Maybe https://gitlab.com/drupalspoons/migrate_tools/-/issues/89? I'm not really sure.

richardbporter:
I also thought it was because the method exists in the base class but that didn't seem to be it either.

richardbporter:
Todo:

- ~~[x] Move migration config to admissions split and remove admissions_migrate split.~~
- [x] Double check node/61 is replaced if it is found

config/admissions.uiowa.edu/migrate_plus.migration.d7_admissions_aos.yml
```diff
@@ -75,11 +75,11 @@ process:
     plugin: static_map
     source: field_program_types
     map:
-      0: major
-      1: minor
-      2: certificate
-      3: preprofessional_program
-      4: online_offcampus_program
+      - major
```
richardbporter: 🤦🏻

richardbporter: This works fine actually. Exporting the split config reformatted it.

richardbporter:
> Double check node/61 is replaced if it is found

I'm not finding it. I think it was logged erroneously initially because MigrateEvents::POST_IMPORT is invoked 3 times and node 61 is in our manual lookup table. So it was trying to replace the replacement.

richardbporter:
> Worth noting that running the migration again will attempt to fix broken links again which at best won't work or at worst replace them incorrectly. 

I switched to using state which might help prevent this from happening.

richardbporter:
Got a CM problem where this split cannot be imported after disabling the split. Investigating...

richardbporter:
> Got a CM problem where this split cannot be imported after disabling the split. Investigating...

Fixed it up in hook_uninstall() of admissions_migrate which clears a config import. Updated uninstall instructions.

joewhitsitt (approved):
Reviewed, not tested locally.

DEV is looking good. I couldn't find an internal link that was broken during my spot check.

PR #3294, opened 2021-03-30
bumping uids release number

GaryRidgway:
Is resolving https://github.com/uiowa/uids/issues/445. This is intended to give headings and intros a scaling size based upon screen wdth.

# How to test

1. Pull down any site with headings or intros
2. Test that the sizing of any headers or intros scales between 600px and 1310 px viewport width, and that there are clamp rules to change their size and backup rules for unsupported browsers (Which should only be ie7 and before)

### DIFF
--- a/docroot/themes/custom/uids_base/package.json
+++ b/docroot/themes/custom/uids_base/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "ISC",
   "dependencies": {
-    "@uiowa/uids": "uiowa/uids#3.4.2",
+    "@uiowa/uids": "uiowa/uids#3.4.3",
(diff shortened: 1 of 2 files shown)


bspeare:
Pulling down to test.  Do we need the updated `yarn.lock` file for this bump?

richardbporter:
I would think so but perhaps not: https://travis-ci.com/github/uiowa/uiowa/builds/221707366#L1771

bspeare (approved):
Clampage looks good!

richardbporter:
Conflict with latest. Resolved but lets double check Travis build output for correct version.

richardbporter:
3.4.3 https://travis-ci.com/github/uiowa/uiowa/builds/221812341#L1793

PR #3331, opened 2021-04-02
Enabling menu item "Show as expanded" option to be used

### LINKED ISSUE #3308: As a site owner, I can control whether my superfish menu items appear as expanded in the main navigation
# Story

As a site owner, I can control whether my superfish menu items appear as expanded. 

Superfish sets all menu items to be expanded by default so that the "show as expanded" option provided by Drupal is overridden.   

Screen Shot 2021-03-31 at 3 57 39 PM

# Proposed solution

- [ ] Change the menu block settings for the Toggle, Horizontal, and Mega menu blocks
  - Update config for these menus https://github.com/uiowa/uiowa/blob/main/config/default/block.block.main_navigation__superfish__toggle.yml, https://github.com/uiowa/uiowa/blob/main/config/default/block.block.main_navigation__superfish__mega.yml, https://github.com/uiowa/uiowa/blob/main/config/default/block.block.main_navigation__superfish__horizontal.yml
- [ ] Create update hook to set all existing menu items as expanded. 
  - Turn on show as expanded options
  - Manually import the appropriate menu type block config per site. 
- [ ] Check against all sites and HR

pyrello:
Resolves #3308 

# How to test

## HR
- `blt ds --site hr.uiowa.edu`
- Compare https://hr.local.drupal.uiowa.edu/benefits/health-insurance with https://hr.uiowa.edu/benefits/health-insurance
- Check that main navigation looks the same.
- Check that secondary navigation looks the same.

## SiteNow V2
- `blt ds --site law.uiowa.edu`
- Compare https://law.local.drupal.uiowa.edu/admissions to https://law.uiowa.edu/admissions
- Check that main navigation looks the same.
- Check that secondary navigation looks the same.

## SiteNow V3 and special cases
Run through the same steps as above on some V3 site and these special cases: uiowa.edu, admissions.uiowa.edu, grad.uiowa.edu.

## Node revisions
* Visit http://admissions.local.drupal.uiowa.edu/node/301/layout and configure block settings for the menu.
* Verify that the correct depth (1) is set.
* Revert the node to an older revision.
* Repeat above steps to verify correct depth is still set.


pyrello:
I've tested this on coronavirus.uiowa.edu, hr.uiowa.edu, and uiowa.edu and all seem to look okay. It would be good to get some additional sites tested, if there are any that people have concerns about. Otherwise, I think its ready :)

pyrello:
About to push an update to add some output to the update hook.

pyrello:
This is ready now.

cory-skeers (approved):
Tested w/coronavirus.uiowa.edu, admissions.uiowa.edu, writinguniversity.org (different menu style), and cnm.uiowa.edu (~500 menu items to test the batching), and each synced without issue and with no noticeable changes to their current menu functionality.

joewhitsitt:
pulling down in a bit

joewhitsitt:
This has a big effect on the sidebar menu (hr, v2, v3). All of sudden children are exposed. I know that is the point of the issue, but I am not sure we have thought about it at the sidebar level. HR especially is not designed for this change so I at least think they should be removed from the update. If this stays the same, communication about this should happen and more control needs to be provided to site owners. Some wild examples: https://hr.local.drupal.uiowa.edu/benefits/health-insurance, https://cnm.local.drupal.uiowa.edu/prior-seasons, https://law.local.drupal.uiowa.edu/admissions

bspeare:
Yes, this looks like it will open up the sidebar navigation on sites using menus.     Should we wait to include a brand approved sidebar design https://github.com/uiowa/uids/pull/515 before this merges?

HMB: https://hawkeyemarchingband.local.drupal.uiowa.edu/drumhawk vs. https://hawkeyemarchingband.uiowa.edu/drumhawk

Screen Shot 2021-04-05 at 2 29 17 PM

Live site:

Screen Shot 2021-04-05 at 2 32 10 PM

pyrello:
Although we obviously don't want to unleash this on a bunch of sites without warning, but I actually like this as an option!

pyrello:
Some ideas for how to move this forward:
* Limit setting the `expanded` to just menu items that are top-level.
* Additional update hook processing to set block instance settings to handle this appropriately at the menu block level.

Thoughts about best way to proceed?

mark-bennett-uiowa:
@bspeare @pyrello I like this too, but worry about how this will look without an indent or other distinguishing feature. Take for instance: [https://cs.uiowa.edu/undergraduate-programs](https://cs.uiowa.edu/undergraduate-programs). It gets hard to tell the difference if all we are changing is the font size and a little padding.

Can we have a slight indention? Will this cause issues with mobile display? Other thoughts?

bspeare:
@mark-bennett-uiowa The hope is to get something like this implemented:

Screen Shot 2021-04-05 at 2 52 59 PM

pyrello:
After discussing it with @bspeare a bit, I think the right way to handle this is to set block instance settings for menu blocks other than the main menu in the navbar region to show 1 level of navigation rather than "Unlimited," which is what I think a lot of them are set to.

mark-bennett-uiowa:
> @mark-bennett-uiowa The hope is to get something like this implemented:
> 
> Screen Shot 2021-04-05 at 2 52 59 PM

So the main children show up with greater than signs, and their children dont have them? Thats promising.

And yes @pyrello that sounds good. I have some custom horizontal menus that shouldnt be affected by that either way, but can see the benefit. Thanks guys.

joewhitsitt:
+1 vote towards the menu config update from unlimited to 1 at least as a short term default.

pyrello:
A couple notes for myself:
* HR has a specific block config that can be updated.
* V2 doesn't allow any per item use of LB, so updating the `core.entity_view_display.node.page.default` to the new setting is sufficient to update it for all V2 sites.
* Update hook needs to target V3 sites where layout has been altered from default. Looking for block instance and updating the setting in the block instance.

pyrello:
Okay. By leveraging the history of other update hooks, I have come up with something that addresses the third case in my previous comment. However, the issue with this (and incidentally the update hooks that it was based on) is that it will update the node so that the current version of the menu configuration, if present, will reflect the changes we want. However, for someone to undo this accidentally, all they need to do is revert to a previous revision on a page where the menu is present AND the layout has been overridden.

Does that seem like an acceptable risk? Or do I need to come up with a way to make this change in old revisions as well?

cory-skeers:
> Does that seem like an acceptable risk? Or do I need to come up with a way to make this change in old revisions as well?

Oof... Good question. My gut reaction was that it would be an acceptable risk, thinking that a) there are a lot of pages where this wouldn't come into play anyway, and b) it retains the option of immediately reverting changes for people who are currently editing sites (make an edit, change their mind/find a mistake, etc., and revert to right before they edited), which seems like the more common use case of revisions.

But that also seems like a lot of assumptions on my part.

Would it just (if there's anything "just..." about this) be going through old revisions of the block itself, or the block and the page? I'm thinking just the block itself, and the page revisions would be taken care of as long as the VIDs are still valid after updating?

pyrello:
> Would it just (if there's anything "just..." about this) be going through old revisions of the block itself, or the block and the page? I'm thinking just the block itself, and the page revisions would be taken care of as long as the VIDs are still valid after updating?

I'm not totally sure about this. I think you would have to manipulate the records for the layout field revisions where the block is present because it is stored as flattened structured data in the field. I'm actually not sure if the information is stored another way besides that. In that case, I think we would be preserving people's ability to revert revisions, but by making it look like that was the option that was always set that way, which seems a little icky from a data integrity standpoint.

cory-skeers:
> In that case, I think we would be preserving people's ability to revert revisions, but by making it look like that was the option that was always set that way, which seems a little icky from a data integrity standpoint.

I think we're in a kind of icky situation needing to update already, where preserving the ability to revert revisions follows the spirit of data integrity, if not the letter of the law. The flip side is that the data integrity is literally preserved, but the presentation and behavior of that data to the user is different than when it had previously been saved.

>  I think you would have to manipulate the records for the layout field revisions where the block is present because it is stored as flattened structured data in the field

That sounds right. Which is icky on the one hand, but might make it easier to directly update without worrying about other hooks or interactions getting in the way.

My vote would be to take some time to figure out if we could, and what it would take, to update the old revisions, but I could still be persuaded if anyone else feels strongly one way or the other.

pyrello:
The conversation up to this point has helped to solidify my feeling that we should manipulate the previous revisions so that the change cannot accidentally be reverted. Thanks @cory-skeers!

joewhitsitt:
Can we update the revisions through a sql update instead of through the intensive Drupal load/edit/save process?

pyrello:
> Can we update the revisions through a sql update instead of through the intensive Drupal load/edit/save process?

I think we have to do it that way because we don't want to trigger all the other stuff that happens when a node is saved/updated.

pyrello:
A few things:
* I have refactored the second update hook to replace the serialized menu block settings directly in the database. This is very efficient compared to loading nodes and executing node operations, but it also scares me a bit.
* For some reason the serialized menu block settings are not stored consistently, so there are two update operations to catch both an integer case and a string case.
* I left in some debugging code because I think this actually makes it easier to have confidence in what is happening. I can remove it prior to merging.

pyrello:
Updated testing instructions to include checking older node revisions.

cory-skeers:
Tested on cnm.uiowa.edu and the updates work well for the sidebar menus (and the primary menu still works). No revision updates needed (v2) but the second update hook ran without issue. Plan on continuing testing, particularly for v3/revisions.

cory-skeers:
Getting a config diff error for `core.entity_view_display.block_content.uiowa_events.default` on trying to sync admissions to test this. I think it's a local issue, but throwing it out there since it's slowing down my ability to test, and in case it's not just flubbed up on mine.

bspeare:
@cory-skeers I had the same issue.

pyrello:
I was getting that issue as well. I'm not sure where it is coming from, but I think it is unrelated to this update

joewhitsitt:
i think it is related to past events. i will fix in a separate pr

update: https://github.com/uiowa/uiowa/pull/3345

cory-skeers:
Update: Following #3345 I tested this on admissions, and the menu updates worked, as did reversion shenanigans.

pyrello:
Need to test on presidentialsearch.uiowa.edu.

richardbporter:
Presidential search site seems fine. It updated 9 menu items. I did not notice any visual differences.

joewhitsitt:
going to sync a couple sites

joewhitsitt:
Tested on hr, brand, law, webcommunity.sites, admissions.

Nice work! 👍

Issue #3204, opened 2021-03-08
Consider adding entity print stuff to default config (but off)


joewhitsitt:
## Issue

Anytime default entity view config is changed and Admissions hasn't been updated, configuration imports fail.

https://github.com/uiowa/uiowa/pull/3048#issuecomment-779490809
https://github.com/uiowa/uiowa/pull/3154
https://github.com/uiowa/uiowa/pull/3331#issuecomment-814432897
https://github.com/uiowa/uiowa/pull/3928#issuecomment-896994724
https://github.com/uiowa/uiowa/pull/4111

## Proposed Solution

- Update the affected entity view configuration to include but hide the entity print stuff similar to how we handled the search api thing. https://github.com/uiowa/uiowa/pull/2092
- Check permissions on routes

joewhitsitt:
Strikes again: https://github.com/uiowa/uiowa/pull/4451

briand44:
Probe to see if having it "on" but not enabled creates routes or adds any additional overhead. If not, seems like we should move forward with this.

e-marie-w:
This afternoon I enabled the module (not certain on how to have it on but not enabled). I re-saved displays and exported the changed config, following what `core.entity_view_display.*` files Admissions had split. 
Basically,
```
hidden:
  entity_print_view_epub: true
  entity_print_view_pdf: true
  entity_print_view_word_docx: true
```
got added to a lot of default files and deleted from a lot of Admissions files. 

For some reason exporting `core.entity_view_display.block_content.uiowa_quote.default` changes some seemingly unrelated things, so I did not commit those changes.

I completely skipped `core.entity_view_display.node.page.default`, this was adding Entity Print in a few places (I'm guessing disabling it would stop this?) so I left it for later. 

I'm guessing other site splits will have custom `core.entity_view_display.*` files and need to be saved and re-exported as well. But I am out next week, so I am unassigning myself so someone else can take over.

joewhitsitt:
In terms of permissions and routes, making sure as a visitor, webmaster for the default site you can not see any entity print configuration, or have any of the routes associated with the module resolve while on the admissions site, everything should still work as originally designed.

https://git.drupalcode.org/project/entity_print/-/blob/8.x-1.x/entity_print.routing.yml

https://github.com/uiowa/uiowa/pull/2970
https://github.com/uiowa/uiowa/pull/3234
https://github.com/uiowa/uiowa/pull/3490

[...]

PR #3345, opened 2021-04-06
Admissions - Events Block Config Diff Fix

joewhitsitt:
relates to #3204


cory-skeers (approved):
Reviewed and tested, and works as advertised--fixes the config diff error. Thanks @joewhitsitt!

PR #3353, opened 2021-04-08
Uninstall jsonapi and delete webmasters REST export view.

### LINKED ISSUE #3179: Uninstall jsonapi, remove webmasters view and delete corresponding dispatch communications
I noticed quite a few MySQL slow queries related to the router table. Over half the entries in that table are from the jsonapi module which is only used for the webmasters API, as far as I know. It creates routes per CRUD operation per field per entity, I think, and we have a lot of those :). Considering how infrequently we use this API, it seems like overkill to enable this on every site. 

Also, the data.its endpoint that aggregates the API data has some performance problems. It might be more performant to get this list using a CLI report similar to the SSL/DB report in UMC. We have other use-cases for these types of reports like a site/application manifest, site versions and #2852.

richardbporter:
Resolves #3179 

### Testing
- See https://data.its.uiowa.edu/sitenow/webmasters no longer exists.
- Optional: see D8 Webmaster population, template and campaign have been archived.
- Sync a site.
- Test path /api/admin/v1/webmasters does not exist.
- Test other jsonapi endpoints mentioned in https://github.com/uiowa/uiowa/issues/3179#issuecomment-816802133 do not exist.
  - ex. /jsonapi/user/user


joewhitsitt:
pulling down to review

joewhitsitt (approved):
Also, I did a quick poke in the repo too to see if other sites have jsonapi references. none that i can see.

Issue #3338, opened 2021-04-05
Migrate events fire multiple times


richardbporter:
# Problem
All of our event subscribers for MigrateEvents fire multiple times. This is not the desired behavior as code running in those subscribers could have negative effects. For example, a PostImport event to replace internal links could run again and replace the links again, resulting in broken links.

# Steps to reproduce
Set a breakpoint or log some output in any MigrateEvents event subscriber, for example admissions, pharmacy or sitenow_migrate. Run the migration in question and see the breakpoint or log output multiple times.

# Proposed solution
TBD.

richardbporter:
While digging into this, I noticed a couple interfaces provided by migrate that could be helpful for our BaseNodeSource to implement. 

`docroot/core/modules/migrate/src/Event/ImportAwareInterface.php` provides `preImport` and `postImport` methods that are called automatically. We are not currently implementing this.

`docroot/core/modules/migrate/src/Event/RollbackAwareInterface.php` provides `preRollback` and `postRollback` methods that are called automatically. We _are_ currently implementing this since we extend `SourcePluginBase` which implements it.

These could replace the event subscribers and reduce some code.

cory-skeers:
> `docroot/core/modules/migrate/src/Event/ImportAwareInterface.php` provides `preImport` and `postImport` methods that are called automatically. We are not currently implementing this.

That would be great. I'll look at adding that into #3366, and we should be able to get rid of the `postImport` listener completely then.

cory-skeers:
This appears to be related to: https://www.drupal.org/project/drupal/issues/3185389

There's lots of adding/removing event listeners throughout various parts of the migration process, and the removal of listener(s) can cause other listeners to fire a second time. This is likely the cause of both our custom event listeners firing multiple times, as well as the cause for those times when `prepareRow` fires twice.

PR #3366, opened 2021-04-13
Sitenow_migrate updates

cory-skeers:
Had some local updates I wanted to get in place, and start an area where migrate platform updates can be added.

To start, was looking at creating basic link replace/report/removal functionality in one spot that can be added and/or overridden as needed.

- [x] Option to remove links on pages created before a certain date (used in pharmacy, at least one other request has mentioned this)
- [x] Option to replace links from a known mapping, if that mapping is available prior to migration
- [x] Option to replace internal links which may have been updated as part of the migration process
- [x] Update link fix option to account for multiple migrations
- [x] ~Fix the process order for standard sitenow migrations, at least, while allowing for incomplete migrations (eg only articles, only pages, people and articles but no pages, etc)~ Set up to run pages as a migration, as well as potentially articles and/or people prior. If a sitenow standard migration needed to be run without pages, but with link-replace functionality, minor code alterations can be done locally to accommodate.

# Updates included in this PR

`linkReplaceTrait` has been updated for tools available to future custom migrations. By default, when running a `d7_page` migration, it will attempt to update links on any page, article, or person which may have updated node ids for their destination.

Lots of code cleanup, including the removal of two of the three event listeners.

# How to test

`git checkout sitenow-migrate-updates`
`git pull`
`composer install`
Sync a sitenow standard D7 site in DevDesktop
`blt ds --site=default`
`drush @default.local uli /admin/config/development/configuration/config-split`
Enable the Sitenow Migrate split
`drush @default.local cim -y`
Go to https://default.local.drupal.uiowa.edu/admin/config/system/sitenow-migrate
Complete the form and save (help text is available for configuring to your DevDesktop instance)
`drush @default.local ms` <- view available migrations and their content counts. `d7_files` will probably have a weird number, possibly negative--that's okay; we're not concerned with it at the moment.
You can run `d7_article` and/or `d7_person` with `drush @default.local mim d7_article` and `drush @default.local mim d7_person`, respectively.
After running the above migrations (or not) `drush @default.local mim d7_page`.
If everything runs without error, congratulations!
Spot check some migrated pages, see if there's any weirdness or not. Try clicking on some links, and see if they bring you to the correct page.


cory-skeers:
@richardbporter I forgot--did you figure out why the event subscriber was firing multiple times, or did that end up going through with the workaround?

Originally in testing this, the `postImportProcess` method on `Pages` fired twice, I've made some (seemingly unrelated) changes since, and now it doesn't appear to be firing at all.

richardbporter:
I did not: https://github.com/uiowa/uiowa/issues/3338

cory-skeers:
> I did not: #3338

Perhaps I should have looked into it a little bit... Thanks!

config/features/sitenow_migrate/migrate_plus.migration.d7_article.yml
```diff
@@ -40,6 +40,4 @@ process:
 destination:
   plugin: 'entity:node'
   type: article
-migration_dependencies:
```
richardbporter, line 43: These are not dependent on d7_file to do lookups?

richardbporter:
Migrate counts seemed accurate for 4cast. 

Ran the migrations. Got an ImageMagick error that I think I've seen before:
```
[error]  ImageMagick error 1: identify-im6.q16: width or height exceeds limit `/var/www/uiowa/docroot/sites/default/files/2021-04/wysiwyg_uploads/sita poster from 4cast_0.jpg' @ error/cache.c/OpenPixelCache/3839.
 [command: identify -format 'format:%[magick]|width:%[width]|height:%[height]|colorspace:%[colorspace]|profiles:%[profiles]|exif_orientation:%[EXIF:Orientation]\n' '/var/www/uiowa/docroot/sites/default/files/2021-04/wysiwyg_uploads/sita poster from 4cast_0.jpg'] 
```

That seems to hose the d7_file counts. Pretty sure I saw that before and we can figure out later. :)

config/features/sitenow_migrate/migrate_plus.migration.d7_article.yml
richardbporter: Not seeing any lookups. Not sure why I thought that....

richardbporter (approved):
Nice cleanup! 

@pyrello and I talked at one point about investigating the value of making some of this trait functionality available in process plugins. I did a bit of this in admissions: https://github.com/uiowa/uiowa/blob/main/docroot/sites/admissions.uiowa.edu/modules/admissions_migrate/src/Plugin/migrate/process/CreateDegrees.php It might be nice to maintain that code in a more concrete context of a plugin so that DI and code intel are available. Also, it opens up the possibility of applying the functionality by running the source through the plugin in the migration definition YAML. 

Doesn't change anything here. This is a great step. Just wanted to get this idea out.

cory-skeers:
Thanks for the test, and the plugin comments, @richardbporter!

PR #3383, opened 2021-04-19
WIP of switching to Drush cron.

richardbporter:
Went down a rabbit hole of setting environment variables to test the expansion in the Drush config files. Setting the env variables basically tells BLT that its an AC environment which breaks ALL KINDS of stuff. We could override the EnvironmentDetector but that breaks other things which requires more work. The cache dir is just not expanded locally which is odd but works. 

### Testing
TBD


richardbporter:
I'd like to make #3457 a part of this.

richardbporter:
If this ever becomes possible in the future, it would probably be easier to just cherry-pick a few commits from this PR.

PR #3385, opened 2021-04-19
Switch to automated_cron.

richardbporter:
This PR uninstalls ulitmate_cron, installs automated_cron to run hourly, removes our cron blt command, updates config_ignore settings and changes AC and NRD cron settings. ~~I will set up a separate nightly job for AC data.~~ NRD is set to run once per day. Is that enough?

We can't reliably use automated_cron to run ulitmate_cron timings. Also, I'm concerned with ulimate_cron database logging. https://www.drupal.org/project/ultimate_cron/issues/3043908 I think ideally we would use Drush to run cron and #3383 has been started to do that. 

Looking back on https://github.com/uiowa/uiowa/pull/1658, it seems export filtering is working as expected now so I'm switching config ignore settings back to what they were. Thats based on https://www.drupal.org/project/acquia_connector/issues/2968641#comment-12657050.

### Testing
- `blt dsa`
- Make sure no config diffs.
  - `blt ume cst`


richardbporter:
Noticed an issue with acquia_connector Drush command on dev. Looking into it.

richardbporter:
Looks like AC data is not sending correctly right now anyway so I'm not going to fix it in this PR.

joewhitsitt:
so the just of this is we take a step back from being able to provide scheduled content more granular than an hour? also won’t we have the same or similar issue with every site running the same jobs on the hour?

richardbporter:
>so the just of this is we take a step back from being able to provide scheduled content more granular than an hour?

Yea we can't commit to that using our current method. It seems to not scale.

>also won’t we have the same or similar issue with every site running the same jobs on the hour?

They won't all run on the hour. It is completely dependent on traffic to the site to trigger cron.

Maybe hourly is too often to start with? We could change back to 3 hours initially.

joewhitsitt:
pulling down to test

richardbporter:
Thanks. Just switched back to 3 hours to start with. I guess it'd be nice to get this out. Still waiting to hear on that site request.

joewhitsitt (approved):
I'll miss you ultimate_cron, though I can't say I got to know you.

PR #3411, opened 2021-04-26
Configuration Split Status

joewhitsitt:
Drush command that works with blt ume to check the status of a given split.

I am flexible on where this lives, just need something to tell me the usage of a given split.

## To Test

- pull and `drush cc drush`
- with a few multisites synced (v2 and v3) and present in your local.blt.yml file, run the following command:
`blt ume "sitenow:config-split:status sitenow_v2"`
- you should get output "true/false" depending on the site.

### DIFF
--- a/docroot/profiles/custom/sitenow/composer.json
+++ b/docroot/profiles/custom/sitenow/composer.json
@@ -7,5 +7,12 @@
             "name": "ITS",
             "homepage": "https://its.uiowa.edu"
         }
-    ]
+    ],
+    "extra": {
+        "drush": {
+            "services": {
+                "drush.services.yml": "^10"
+            }
+        }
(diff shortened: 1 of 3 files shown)


richardbporter:
Does this do anything that `drush @mysite.env config:get config_split.config_split.my_split status` doesn't? :)

joewhitsitt:
Ha! Where were you when I asked https://iowaweb.slack.com/archives/C8SM82ZFZ/p1619465743122400

joewhitsitt:
For my future self that comes back here looking for the way to do this without getting a "too many arguments" error:

`./vendor/bin/blt ume "config:get config_split.config_split.sitenow_v2 status"` from the root of an application environment

PR #3379, opened 2021-04-16
Mega PR: WYSIWYG Buttons Refactor, Admissions card link updates, Table updates

joewhitsitt:
Resolves #3078 
Relates to https://github.com/uiowa/uiowa/issues/260
Relates to https://github.com/uiowa/uiowa/issues/2532
Resolves #3262 
Resolves #3330 
Resolves #3395

- Removes our custom LinkIt solution for adding button classes to WYSIWYG links and replaces it with style dropdown options on the `<a>` tag.
- On all sites but HR, loops through text_long and text_with_summary type fields along with our ignored view config and replaces bootstrap button classes with uids classes.
- Does the same thing to remove table classes.
- Add bold and light intro classes to style dropdown on the `<p>` tag.
- Splits the filtered_html text format for HR which has different classes for the lead and buttons.

# To Test

To make sure that you get the UIDS update, follow these steps:
```
rm -rf ./node_modules
yarn cache clean
yarn install
yarn workspace uids_base gulp --development
yarn workspace admissions_core build
```

## Button updates

- Sync `sandbox` `hr` and other sites (e.g. a website with separate content type with text_long or text_with_summary fields).
- See https://sandbox.local.drupal.uiowa.edu/bootstrap-buttons. All of the markup has been changed from btn* classes to bttn* classes. Revert to a previous revision. You should still see the bttn* markup.
- Make changes to the content to confirm the markup stays and that there are no weird errors.
- See https://sandbox.local.drupal.uiowa.edu/sandbox-people. See the header content buttons are bttn* markup.
- See https://hr.local.drupal.uiowa.edu/advanced-layout-guide-not-publication and see that the buttons on the page are still the HR bootstrap based styles. Edit the page and see that Lead and Button style options still use bootstrap classes.

### Go the extra mile

- Query the entire database to see if `btn-` exists. It shouldn't except maybe for a key_value_expire table entry.

## Area of Study (Admissions) card link updates

- `blt frontend`
- Edit http://admissions.local.drupal.uiowa.edu/academics/accounting and add links to to the Admissions Requirement link fields.  Save the page and verify that the link button works but the entire card is not linked. 
- Edit the page again and add a "Story" to the page and verify that the story card is still clickable. 

## Print links (Admissions) permissions updates
- Masquerade as a webmaster and go to https://admissions.local.drupal.uiowa.edu/areas-of-study/print-links.   Try clicking on a pdf link in the "Tests" table at the bottom of the page.   You should **not** be seeing an "access denied" message. 

[...]

PR #3414, opened 2021-04-26
Admissions - AoS Permissions Update

### LINKED ISSUE #3248: Notice: Uninitialized string offset: 0 in sitenow_preprocess_node()
# Problem
sitenow_preprocess_node() assumes moderation. we now have custom content types without moderation which causes this notice `Notice: Uninitialized string offset: 0 in sitenow_preprocess_node() (line 791 of profiles/custom/sitenow/sitenow.profile).`

# Steps to reproduce
- Unpublish a student profile on admissions and view it.

# Proposed solution
- Update logic to check for moderation_state

### LINKED ISSUE #3380: Admissions - Allow un-publishing of Area of Study
# Story

As an Admissions site editor, I would like the ability to unpublish and republish areas of study content so that we can accommodate programs that pause or are removed and then start up again.

# Additional Info

It appears this is just a permissions issue.
Screen Shot 2021-04-20 at 1 12 28 PM

# Proposed solution

- [ ] Update permissions in Admissions split to allow "override Area of Study published option" for editors, publishers and webmasters.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/3380
Resolves #3248

This removes the AoS Block from the AoS feature split because it failed on import due to several fields not existing in the feature split. I don't think any other sites besides Admissions are using this feature.

## To Test

- pull, sync the default site, login, enable the areas of study feature split, `cim` - the import should run successfully
- as a webmaster with the area of study feature split enabled, you should be able to add, publish and then unpublish an area of study without error.
- sync admissions, as a webmaster you should be able to unpublish an area of study without error.

## Note

~~I created this PR separately from https://github.com/uiowa/uiowa/pull/3379 because I thought that was blocked. It also has changes to Admissions permissions, so which ever gets merged first will have to be updated in the other.~~


joewhitsitt:
No changes to permissions post-#3379 were needed

bspeare:
Pulling down to test.

bspeare (approved):
I was able to follow the testing instructions without any errors.

PR #3428, opened 2021-04-28
Re-enable composer security check and disable drupal security check.

richardbporter:
I actually forgot to rollback #2985 after upgrading to BLT 12 and Drupal 9. This re-enables Composer security checks and disables Drupal security checks (temporarily).


joewhitsitt (approved):
reviewed, not tested

PR #3430, opened 2021-04-29
Migration physics

### LINKED ISSUE #3369: Build Physics and Astronomy news migration
# Background

Physics and Astronomy news entries (News, machine name 'clas_news') have the following fields:

- Title (machine name: title)
- News Tags (machine name: field_clas_news_tags; a taxonomy-ish term reference field to tag entries as Space Physics, Plasma Physics, etc news)
- News Date (machine name: field_clas_news_date)
- News Description (machine name: field_clas_news_description)
- News Link (machine name: field_clas_news_link; URL and Optional Title)

I need to get feedback on how often the News Tags are used, which I will update here.

# Proposed solution

This may be a relatively straight-forward migration of the following (OLD --> NEW):

Title --> Title
News Tags --> ignored for now, Relationships (will add these tags in preparation of migration, can ignore x,x, and x)
News Date --> Authored on date
News Description --> Body (can we take summary and copy to Article Summary??)
News Link title --> ignored
News Link URL --> Source Link

mark-bennett-uiowa:
Here is the migration module for the Physics and Astronomy news items. Will solve #3369.

Still to do:

- [x] Properly handle lookup and storage of tags from old site into existing taxonomy (tags).
- [x] If physics_tags.yml file is not needed for creation of taxonomy terms, can delete.

# How to test

- In DevDesktop, clone site and pull database from uiclas application (may not be able to do)
- git checkout migration_physics (and the usual stuff)
- blt ds --site=physics.uiowa.edu
- drush @physics.local uli
- Enable "Physics Migrate" module in Extend
- Go to Configuration > System > SiteNow Migrate
- Enter default values from comments except for Database name (physics_uiowa_edu) and Base URL (https://physics.uiowa.edu).
- Go to Adminsitration > Structure > Migrations > Edit SiteNow migration group.
- Execute Articles migration (should be at least 334 items).


cory-skeers (approved):
Tested it out, and migration worked well!

Made a few small tweaks to the manualLookup override to get the terms working, and we should be able to do without a separate term migration. The one gotcha left there is the "Atmospheric and Environmental Physics" term wasn't in prod yet. If no other tags are created prior to its being added, then it can be created manually and the migration run without issue. Otherwise, will just need to update the `tid` for that one tag if it is different.

mark-bennett-uiowa:
@cory-skeers thanks for testing and updating this! I ended up adding a tag on the new site for "Atmospheric and Environmental Physics" so we shouldn't have to worry about that one. Gonna pull this down to test and demo to them in 30 mins.

mark-bennett-uiowa:
This works flawlessly @cory-skeers thanks for the help! I will push a small update to the tid you mentioned above, or do it through the UI here. I think this is ready to go. Can we "hold" this until right before launch, in case they add additional entries, or should we be manually adding those?

cory-skeers:
@mark-bennett-uiowa It shouldn't be a problem holding it until closer to launch.

The main thing to keep in mind is that there should be a small chunk of time while the migration gets setup and run for production that editors should be out of prod to avoid any possible data loss with syncs. But as long as there's no issue there in getting ready to launch, the migration should be able to be run any time.

Technically, we should be able to run the migration once now, and then again prior to launch and have the second run only import articles created since the first run, but it's never been tested, so you'd be heading out into uncharted territory going that route.

mark-bennett-uiowa:
@cory-skeers sounds good. Lets hold it for another few weeks then, and try to run it mid-May. We are shooting for an end of May launch if possible.

PR #3435, opened 2021-04-30
Parking and Transportation migration

### LINKED ISSUE #3264: Migrate articles for Parking and Transportation
# Story

As a Parking and Transportation site owner, I would like articles migrated from the Parking and Transportation Drupal 7 site (https://transportation.uiowa.edu) to https://transportation.prod.drupal.uiowa.edu so that users will be able to view all of the past news articles.

# Additional Info
They would like us to map tags according to the following screenshot.

Screen Shot 2021-03-19 at 3 49 46 PM

They would like us to programmatically create redirects from old urls.

We should migrate all existing news articles from old site.

They may manually create policy information as news articles.

# Proposed solution

cory-skeers:
Resolves #3264 

# How to test

Sync D7 site in DevDesktop (transportation.uiowa.edu on uiowa703).
`git checkout p-and-t-migration && git pull && composer install`
`blt ds --site=transportation.uiowa.edu`
`drush @transportation.local en transportation_migrate -y`
`drush @transportation.local uli /admin/config/system/sitenow-migrate`
Complete the configuration form using the help text to configure for DevDesktop
`drush @transportation.local ms` -- shows migration status; two migrations should appear, one for articles and one for article redirects
`drush @transportation.local mim --all` -- runs both migrations
There is an ImageMagick error that can be ignored, as can the link replacement notices (no attempt is currently being made at fixing them, so the notice is informational only)

Spot check imported articles and their content, that images are present where they existed in the D7 site, summaries, etc.
Check that article aliases from the old site work following the redirects migration.
Check that all imported articles have the news-ParkTransit tag.
Check that articles which had tags in D7 have the appropriate updated tag on import according to #3264 
Check that the title and published date are not duplicated in the body content (see example image below).

image


cory-skeers:
Title and date are visually repeated in a fairly regular pattern. Some cleanup can be done.
image

cory-skeers:
> Title and date are visually repeated in a fairly regular pattern. Some cleanup can be done.

Some correspondence with the client on this.

"The article title and headline are identical for most news articles. (We checked the recent news back to Jan 2020 and only saw  one exception.) So it should be fine to just use the headline.

Similarly, the published date and headline date are identical or within 1 or 2 days’ difference for most articles. So it will be fine to just use the published date."

Plan is to remove the dateline instance from the body, and grab the first headline, remove it, and replace the article title with it (many of the old article titles were repeats of the headline, but with an additional date added, which creates additional redundancy).

richardbporter:
I'll check this out.

richardbporter:
If you run the migration with verbose logging (`drush @transportation.local mim --all -v`), it seems like it is attempting to fix links. I see messages like `[info] Replaced term 6 in article CAMBUS driver Franks and student mechanic Madden together accomplish over 500 clean-up shifts.`

Is that expected?

Also, it appears quite a few articles already exist in the new site. Are those staying or should they be deleted beforehand?

richardbporter:
Seems like they repeated the article title/headline and date in the summary as well which makes that look a bit odd: 

image

I don't know if anything can/should be done about that.

richardbporter (approved):
Everything else checks out. Happy to help setup and run in prod when that time comes. Just let me know.

cory-skeers:
Thanks for the test, and offer of help @richardbporter!

Gonna take a quick look at the summaries before merging--would be a nice fix to get in.

Issue #3419, opened 2021-04-27
Admissions - Report of Long Text Full_HTML fields


joewhitsitt:
## Issue

We suspect that during migration, fields with full_html on the old site were migrated over to the new site the same way where editors don't have access. Additionally, fields have been marked full_html intentionally by developers to provide embeds.

## Proposed Solution

- [x] Query the admissions site for long text and long text summary fields with the full_html format. Provide the entity id associated with one.
  - There are a couple of examples of queries that could be run locally that output messages that could be turned into something to hand over to admissions.
  - https://github.com/uiowa/uiowa/blob/ea2052f6822e02b5c34ceaaca1c73e03c82ac29c/docroot/profiles/custom/sitenow/sitenow.install#L2449
- [ ] Ask Admissions which fields can be turned back to filtered_html and either run an update hook or manually correct.

https://iowaweb.slack.com/archives/C019B2C8K8U/p1619115149056800

joewhitsitt:
Forgot to bring this up for grooming/planning.

pyrello:
This is going to be slightly more difficult than I was originally thinking because of text areas in blocks in Layout Builder. The entity ID in the field table is for the block and so we can't provide it to the client like we can with a node ID.

joewhitsitt:
Changed the IS to show an earlier version of the MEGA update hook where it went through each db table instead where the entity id is available

pyrello:
The entity ID is the block content ID, which isn't helpful unless you know which node ID they correspond to.

joewhitsitt:
ah

cory-skeers:
> The entity ID is the block content ID, which isn't helpful unless you know which node ID they correspond to.

Get the list of full html block IDs, then check the `layout_builder__layout` field for instances of any of the block IDs in the blob text?

briand44:
Update: Scott mentioned in today's meeting that he expects all of the Areas of Study fields currently using full html, don't need to be (no embeds or anything currently in these). Likely all items listed in report will need fixed.

pyrello:
Does this mean he is requesting that we update the field format in addition to identifying it?

If so, we can maybe put some basic parameters around it (such as limiting it to Areas of Study content items), but beyond that, it is pretty much going to be all or nothing. That means if they did happen to have any content that is not compatible with the limited HTML format, it will get wiped out the next time the content item is saved.

briand44:
[...]

PR #3448, opened 2021-05-04
Update hook to fix `table` classes

### LINKED ISSUE #3447: `table` classes were broken in table update
# Problem
Classes starting with `table--` had the `table` part removed as part of the update hook in #3379. 

# Proposed solution
Update hook to match the remaining classes that were affected and to replace them with the correct version.

```
--gray-borders
--hover-highlight
--width-default
```

pyrello:
Resolves #3447 

# How to test

* `blt sync --site transportation.uiowa.edu`
* `drush @transportation.local uli admin/content`
* `yarn workspace uids_base gulp --development`
* https://transportation.local.drupal.uiowa.edu/cambus/routes-schedules-maps leads to a lot of pages with tables.
* Check this page to make sure that the corrected table class `table--hover-highlight` is still there.
* Add `table--width-default` to a smaller sized table like on the bottom of https://transportation.local.drupal.uiowa.edu/cambus/routes-schedules-maps/va-loop-shuttle-route and make sure the table does not stretch full-width across the page.


pyrello:
The downside to the approach I'm using is that if anyone has gone in and fixed this themselves, then this will break those instances.

One idea for how to refine this would be to add a condition to the update query that excludes fields where there is a match for the fixed terms.

joewhitsitt:
>if anyone has gone in and fixed this themselves, then this will break those instances.

could you loop back through looking for `tabletable--...` or whatever the breaking example will be and replace with `table--`?

pyrello:
Good idea!

pyrello:
That seems to work! To make it very clear for testing, I added some output messages.

Clever solution, @joewhitsitt!

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -2608,3 +2608,90 @@ function sitenow_update_9007() {
...
+                ->expression($value_column, "REPLACE($value_column, :old, :new)", [
+                  ':old' => $needle,
+                  ':new' => $replace,
+                ])
+                ->execute();
+            }
+
+            if ($update_count) {
+              \Drupal::messenger()
+                ->addMessage('Processed ' . $update_count . ' update(s) for ' . $table . ' table ' . $value_column . ' column.');
```
richardbporter: Passing the message through `t` with placeholders might be a more Drupal-y way of doing this rather than concatenation.

pyrello: Fair point.

richardbporter:
> Audit text areas with tables.

Is there a page on the site with a lot of tables?

pyrello:
Having the functionality in #3419 would just be useful generally it seems.

I'm not very familiar with that site. I'll take a look. @bspeare Any pages you would check?

bspeare:
I added a hotfix to make `table--width-default` work and updated the testing instructions.   I couldn't find a page where they are using the class, but https://transportation.local.drupal.uiowa.edu/cambus/routes-schedules-maps/va-loop-shuttle-route would be a good page to test out the class.  Try adding `table--width-default` to the bottom table.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
Test steps checkout. Let's get this out there.

PR #2765, opened 2020-12-31
Local events feature

joewhitsitt:
Relates to #2396
Resolves https://github.com/uiowa/uiowa/issues/3170

Initial event feature split with base functionality.

## To Test

- `blt ds --site=default`
- `drush @default.local uli`
- Go to https://default.local.drupal.uiowa.edu/admin/config/development/configuration/config-split/event/enable
- `drush @default.local cim && drush @default.local cr`
- As a webmaster, you should see the event content type. Add a few pieces of demo content. Title/date should be required. Schema Event should be available to editors to modify. Initial styled output of full and teaser view modes. Keep in mind that this is MVP to move forward with Admissions' requirements. Additional fields and functionality will be added later.

PR #3483, opened 2021-05-06
Schema.org Event Date iso-8601 format

joewhitsitt:
Follow-up oops fix to #2765 

https://developers.google.com/search/docs/data-types/event#date-time-best-guidelines


pyrello (approved):
Reviewed, not tested.

PR #3487, opened 2021-05-06
Followup updates to transportation migration

cory-skeers:
- [x] Replace hardcoded links
- [x] Check default image style settings
- [x] Decide on default image style

# How to test

Sync D7 site in DevDesktop (transportation.uiowa.edu on uiowa703).
`git checkout transportation-migrate-followup && git pull && composer install`
`blt ds --site=transportation.uiowa.edu`
`drush @transportation.local en transportation_migrate -y`
`drush @transportation.local uli /admin/config/system/sitenow-migrate`
Complete the configuration form using the help text to configure for DevDesktop
`drush @transportation.local ms` -- shows migration status; two migrations should appear, one for articles and one for article redirects
`drush @transportation.local mim --all` -- runs both migrations
There is an ImageMagick error that can be ignored, as can the link replacement notices (no attempt is currently being made at fixing them, so the notice is informational only)

## Main things to check

Find a page with images (such as https://transportation.uiowa.edu/news/recognizing-service-cambus-drivers-transit-operator-appreciation-day from the current site) and check that the images are added as `large__no_crop`.
Edit an article that was migrated and check that the source org/link fields are empty.


cory-skeers:
Received the following feedback. Images are largely added with defaults, so it may be a need to adjust the default used.

> "Thanks for the demo of the migration in the sprint meeting. I did notice that in the demo, photos in articles seemed to be very small knowing that I had previously set pixel widths to typically 500-1000 px (example Transit Operator Appreciation Week).
Is that something that we’ll have to manually check and update?"

joewhitsitt:
pulling down to test

joewhitsitt:
I am seeing small no crop on the example page (which was successfully redirected)

Screen Shot 2021-05-10 at 10 06 13 AM

The source link field also says `<nolink>`

FYI, I have not previously run the P&T migration

cory-skeers:
Well shoot. I've tried a database drop, file removal, and resync and run, and am getting the `large__no_crop` as well as the empty source link field. Not sure where the difference is coming from.

The change that relates to the source link field comes from `migrate_plus.migration.transportation_articles.yml`, and the removal of
```
37. field_article_source_link:
38.   plugin: field_link
39.   source: field_clas_news_link
```
There was another extra field being mapped (`field_article_source_link_direct`) that I just removed, but that should have been filling with false anyway, so that shouldn't make a difference.

Gonna try digging and see why this might not be working.

joewhitsitt:
Looks like the test instructions have the wrong branch name. Trying again with `transportation-migrate-followup`

cory-skeers:
> Looks like the test instructions have the wrong branch name. Trying again with `transportation-migrate-followup`

Oof, good catch. That was quite the copy-paste failure on my part.

That makes so much more sense that you weren't seeing the updates! Sorry about that. Thoroughly embarrassed.

joewhitsitt (approved):
Works as expected

PR #3495, opened 2021-05-07
Create transportation site split and cost calculator block.

### LINKED ISSUE #3434: Port Cost Calculator to D9 for Parking and Transportation
# Story

As a Parking and Transportation website owner, I would like to have a cost calculator so that commuters can compare costs based on different modes of transportation.

# Additional Info
  Current cost calculator -> https://transportation.uiowa.edu/cost-calculator
  Desired location on new site -> https://transportation.prod.drupal.uiowa.edu/cost-calculator

Currently is a custom D7 module.

There currently is a admin config page to set default values.
The existing site has a taxonomy for modes of transportation that is widely used across the site and also used with this functionality. It doesn't sound like this taxonomy will exist on the new site so those modes of transportation will have to be hardcoded or handled in some way specific to the cost calculator (See screenshot below).

Screen Shot 2021-05-04 at 8 12 41 AM

# Proposed solution

- [x] Port the calculator over as a custom block with hardcoded terms and other variables, updating outdated PHP along the way. Don't recreate the configuration form or related taxonomy

richardbporter:
Resolves #3434 

# How to test

- `blt sync --site transportation.uiowa.edu`
- `drush @transportation.local config:import --source ../config/transportation.uiowa.edu --partial`
- `drush @transportation.local config:import`
- `drush @transportation.local uli /cost-calculator`
- Place the Cost Calculator block on the page.
- Compare values with the old calculator by putting in the same numbers for both and verifying the math works out the same : https://transportation.uiowa.edu/cost-calculator
- Make sure you can place as many calculators on one page as you want and they all work separate of each other.
- ??
- Profit


cory-skeers:
Pulling down to test

cory-skeers:
Pulled down and followed the instructions (sidenote, need to enable the module as well, after not thinking and scratching my head for a bit when not seeing the block).

Can confirm that the prices I tested with matched results from the original site. Also added multiple calculators (so many calculators), and they work independently from each other. Some overlap occurs when placing calculators in a 4 column section, but surprisingly little.

The new 360 Express yearly savings calculations appear to work, but looks like we drop off the decimal. Not sure if that was intentional (eg Parking cost = $7,385.47, 360 Express should be $7,385.47 - $690 = $6,695.47, but displays as $6,695). Edit: Realized this is a difference in the `number_format`.

Not sure if styling is in-scope, but on a single column section, the calculator gets placed along the left edge (after appropriate padding) creating a lot of whitespace where the old site had it more centered. Not necessarily a problem if blocks can be placed more judiciously in the layout.

Also tested/confirmed that the blocks can be placed by a webmaster.

blt/src/Blt/Plugin/Commands/MultisiteCommands.php
```diff
@@ -14,8 +14,6 @@
 use AcquiaCloudApi\Endpoints\SslCertificates;
 use Consolidation\AnnotatedCommand\CommandData;
 use Consolidation\AnnotatedCommand\CommandError;
-use GuzzleHttp\Client as GuzzleClient;
-use GuzzleHttp\Exception\RequestException;
```
cory-skeers, line 18: Just checking that this change was intentional?

richardbporter:
I think importing configuration again would enable the module after enabling the split. I updated the test instructions. This is similar to the BLT update process - it runs cim twice.

docroot/sites/transportation.uiowa.edu/modules/transportation_calculator/src/Form/CostCalculatorForm.php
```diff
@@ -0,0 +1,220 @@
...
+            ],
+            [
+              'mode' => $this->t('Carpool'),
+              'cost' => '$' . number_format($yearly / 2, 2),
+              'savings' => '$' . number_format($yearly / 2, 2),
+            ],
+            [
+              'mode' => $this->t('380 Express'),
+              'cost' => '$690.00',
+              'savings' => '$' . number_format($yearly - 690),
```
cory-skeers: Do we want the decimals here? `number_format($yearly - 690, 2)`

richardbporter: Yes! Thanks.

blt/src/Blt/Plugin/Commands/MultisiteCommands.php
richardbporter: Yes - I commonly run `./vendor/bin/phpcbf` which tries to automatically fix PHPCS violations. Then I run `./vendor/bin/phpcs` after that to fix what it couldn't. I guess these classes were not in use there.

cory-skeers (approved):
Thanks for the explanation with the blt stuff @richardbporter, and the update to decimals!

Everything looks good on my end for this.

PR #3505, opened 2021-05-07
Create transportation calculator settings form and wire up to blocks.

### LINKED ISSUE #3449: Transportation - Cost Calculator Configuration Form
Follow-up to https://github.com/uiowa/uiowa/issues/3434

# Story

As a webmaster, I can update the transportation cost calculator with the necessary variables and have it feed the calculator similar to the D7 website.

# Proposed solution

- [x] Create configuration form to handle the calculator variables.
- [x] Wire up form config to calculator
- [x] Update transportation's permissions to allow webmasters to update the configuration form.

richardbporter:
Resolves #3449 

# How to test:
- `blt sync --site transportation.uiowa.edu`
- `drush @transportation.local uli /cost-calculator`
- Make sure the Cost Calculator block is on the page, and that pressing the submit button does not throw any errors.
- navigate to `https://transportation.local.drupal.uiowa.edu/admin/config/system/cost-calculator-settings` where all variables for the calculator are set.
- Change some of these around a bit and note that the calculator reflects accurate changes with these fields after calculation post save.
- navigate to `https://transportation.local.drupal.uiowa.edu/cost-calculator` and change some values that would be editable by anonymous users.
- Verify that these change the calculated results accordingly.
- masquerade as a webmaster.
- Verify that you can change the values in the fields on `https://transportation.local.drupal.uiowa.edu/admin/config/system/cost-calculator-settings` as a webmaster.


docroot/sites/transportation.uiowa.edu/modules/transportation_calculator/src/Form/CostCalculatorSettingsForm.php
```diff
@@ -0,0 +1,148 @@
...
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    parent::validateForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $this->config('transportation_calculator.settings')
+      ->set('distance', $form_state->getValue('distance'))
```
richardbporter, line 152: All of the other fields need to be saved here just like distance.

richardbporter: 👍🏻

richardbporter:
@GaryRidgway FYI those config sets can be chained like so: https://github.com/uiowa/uiowa/pull/3505/commits/88250ea4ad9c33616db7875fa6f1a6ccc042588c

cory-skeers:
Pulling down to test

cory-skeers:
Noticed that you can set "average working days per month" higher than possible days per month (e.g. 45 avg working days). Only noticed because negative values are restricted; don't think it's necessary to change, but wanted to point out as I came across it.

Did notice in testing, both the config and calculator are integer values for days of travel and distance. This matches the current calculator, but particularly for distance, I could see a use case for decimal values. Wondering if that's open for discussion or not.

If they stay integer, looks like a validator might help. Right now it lets you submit the form, comes up with accurate results, but displays an error, which is confusing.
image

GaryRidgway:
Thanks Cory! these are some useful insights. I'll look in to them

GaryRidgway:
@cory-skeers 

I added mins and maxes (0->31) for both days of travel and average working days. this should clear up any unreal number of days in a month.

I have also added decimal allowances up to two decimal points for both distance and days of travel so that those can be more finely configured. This should allow for those half travel days and 0.03mile distance calculations!

cory-skeers:
Nice, @GaryRidgway! I'll pull down the updates and check them out shortly.

cory-skeers:
Changes look great!

Last little details that have to do with the calculator and not the config: Distance and parking cost allow for negative values, where they probably don't make sense. Original calculator allowed these as well.

Entering invalid values are flagged as errors, but the results still display for the user.

cory-skeers (approved):
Updates have been tested and signed off.

I HEREBY APPROVE.

PR #3509, opened 2021-05-10
Feature: Investigation of media_iframe module

### LINKED ISSUE #3450: Investigate using media_iframe for UICapture/Panopto Embeds
# Story
As a site owner, I can embed a UICapture or Panopto video in my site so that I can share those videos with visitors.

# Motivation
UICapture/Panopto embeds do not use OEmbed, which is necessary for the remote video media type that we are currently using. We need another solution for this type of embed.

# Proposed solution
- Investigate whether we can use the [media_iframe](https://www.drupal.org/project/media_iframe) module for embedding UICapture/Panopto embeds. This will probably require setting up a new test media type with a field using the `media_iframe` type.
- Recommend whether we want to use that module directly or create our own custom module based on it.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3450. 

# How to test

- `composer install`
- `blt ds --site=default`
- `drush @default.local uli`
- Go to https://default.local.drupal.uiowa.edu/media/add/iframe and add a panopto video like https://uicapture.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=e4de089e-a011-4a5a-bf53-ad1701038fe1
- Add that to a page through the media library through the "Iframe" tab in the library
- Save page and view result

This patch could be useful:

- https://www.drupal.org/project/media_iframe/issues/3205908


joewhitsitt:
taking a look see

bspeare:
This investigation is complete.

PR #3511, opened 2021-05-10
Add style options to people list block

### LINKED ISSUE #3489: Add LB style options to people list block
# Story

As a site owner I use card styles with the people list block created in #2702 to take advantage of all the card options that have been developed.

# Proposed solution

- [x] Enable the Person block list view.
- [x] Provide layout builder style options related to card (background, image position, image format, content alignment, image size, card border)
- [x] Default styles for new instances should be `white, left align, images are left align small circle`
- [x] Update layout_builder_custom to have the same show/hide options functionality as aggregator.
- [x] Style adjustments and update templates to receive block style configuration.
- [x] Disable people list view to prevent premature deployment of the whole feature.

## Relates to

https://github.com/uiowa/uiowa/issues/3501
https://github.com/uiowa/uiowa/issues/2351

pyrello:
Resolves #3489 

# How to test

* Coming soon


joewhitsitt:
`$variables['#cache']['max-age'] = 0;` would fix the caching though it might not be the best solution

richardbporter:
> `$variables['#cache']['max-age'] = 0;` would fix the caching though it might not be the best solution

I wonder if that might be a better temporary option as the current implementation runs every time the layout builder edit screen is rendered since it bypasses the render cache. That could result in those media tags being invalidated quite a lot.

Scoping to field_media_image seemed to work for me locally. `$variables['content']['field_media_image'][0]['#cache']['max-age'] = 0;`

richardbporter:
I don't think the cache change affects anything other than media within this new list block view, at least not from what I can tell. So nothing to test there. 👍🏻

richardbporter (approved):

PR #3530, opened 2021-05-15
Update hook to convert people custom blocks to people list blocks

### LINKED ISSUE #3518: Update people blocks instances to people list view blocks.
# Story
As a site owner, all my existing people blocks will use the new list view created in #2702 so that I can use all the new functionality.

Split from #3489 

# Proposed solution
- [ ] Update hook to:
  - [ ] Query existing instances of the `uiowa_people` block in LB sections.
  - [ ] Unserialize the section into its class structure.
  - [ ] Loop through components and find the `uiowa_people` components.
  - [ ] Remove existing block content instance.
  - [ ] Create an instance of people list view block to set default block configuration.
  - [ ] Pull existing LB styles and translate `block_hide_*` style options into appropriately hidden fields.
  - [ ] Create a new component using the view block configuration and append it to the section.
  - [ ] Serialize the result and update the existing record with the new value.
- [ ] Update block restriction setting to remove access to the `uiowa_people` block.

pyrello:
Resolves #3518 

Removes all current and previous instances of `uiowa_people` custom block in LB and replaces them with instances of `views_block:people_list_block-list_card`. Because of modifications made in #3511, these changes should not affect SN V2 sites.

# How to test
- `blt ds --site iti.uiowa.edu`

- [x] Observe update hook runs without errors, reporting back the number of instances that were replaced. 
- [x] Check https://iti.local.drupal.uiowa.edu/our-research/centers/atmospheric-and-environmental-research-lab and inspect source to confirm that the people block on the page is using the `view-people-list-block` class.
- [x] Visit https://iti.local.drupal.uiowa.edu/node/251/layout to check that the "tempstore" version of the page is showing the right block type. You can check this by clicking "configure block" and checking that the settings for the block look like the new block settings.
- [x] Click the "Discard changes" button at the top of the page and confirm, then click the "Layout" tab check the current version of the node similarly to above.
- [x] Spot check previous revisions similarly to above to make sure that previous revisions are using the correct view block.

## User stories
- As a site owner all my pages with people blocks use the new people list view block.
- As a site owner all previous revisions of pages that contain people blocks use the new people list view block.
- As a site owner with unsaved changes in Layout Builder, if my unsaved changes contain a people block they are now using the new people list view block.


pyrello:
This is getting pretty close. The new block is now showing up in the place of the existing block. There are a few more todos to address and then there is a weird issue that was exposed during writing the hook where duplicate entries may be present that needs a bit more probing before I would be comfortable rolling this out.

pyrello:
@cory-skeers After reviewing your addition to the update hook, here are my observations (which informed the changes I made):
* The extra DB query and update was happening for both the `node__layout_builder__layout` and `node_revision__layout_builder__layout` and I think we only want it to happen for the first one.
* If we reuse the section that was generated to update the field table, there is the possibility we will overwrite some intentional changes by the user that haven't been saved yet. While our policy here is unclear, I think we should err on the side of not overwriting those intentional changes.
* I'm not sure how to go about reusing the code section where we convert the component section short of major refactoring or abstracting that section into a helper function. I'm leaning towards a helper function, but have never written one specifically for an update hook, so was uncertain about moving ahead with that.

pyrello:
This ready for review.

joewhitsitt:
pulling down to test

joewhitsitt:
Didn't seem to convert https://engineering.local.drupal.uiowa.edu/node/251 correctly. The "details" fields are hidden locally when email and phone are being displayed in PROD.

pyrello:
> Didn't seem to convert https://engineering.local.drupal.uiowa.edu/node/251 correctly. The "details" fields are hidden locally when email and phone are being displayed in PROD.

Any idea how the display of the fields were getting set? In the example page that I originally was testing this against, those fields are not shown: https://iti.uiowa.edu/our-research/centers/atmospheric-and-environmental-research-lab

joewhitsitt:
Looks like the ability to set lb styles for these were removed: https://github.com/uiowa/uiowa/commit/0664cb2ca852792f215c081d054895050551346c#diff-c373b8e74b61573fd63c0331a659c9ef582bb43de2c5f8487a5a966f206c7980

pyrello:
Ahh... good sleuthing!

pyrello:
I was making an assumption on this since I wasn't seeing LB style options that they didn't exist. So I need to revisit the logic that checks styles on this.

joewhitsitt:
It also didn't update https://sandbox.local.drupal.uiowa.edu/node/241 (temp) where i removed the slider block and created a people block

pyrello:
@joewhitsitt So, to clarify the above points:
* On the engineering page, it didn't convert the block at all? Or it just failed to account for the styles?
* Re: sandbox site page, I'm not clear on how to recreate what you are saying didn't work.

pyrello:
Screen Shot 2021-05-25 at 1 13 56 PM

Came across another issue here, where the grid style doesn't seem to be working correctly.

joewhitsitt:
>On the engineering page, it didn't convert the block at all? Or it just failed to account for the styles?

Looked like just the hide styles

>Re: sandbox site page, I'm not clear on how to recreate what you are saying didn't work.

https://sandbox.local.drupal.uiowa.edu/node/241 (new people block not saved)
When I clicked into configure the block after update and before save it didn't look like it was converted at all.

pyrello:
I think I have resolved both of the issues pointed out by @joewhitsitt above. Just need to work on setting a default value for list format.

pyrello:
I believe all issues are fixed now and this is ready to be tested again.

dnkemper (approved):
I believe I tested everything in issue description and comments... Nice work!

PR #3535, opened 2021-05-17
Alerts a11y improvements.

### LINKED ISSUE #1898: Hawk Alerts need a landmark
## Problem
When rendered (see uiowa.edu stage site) there are two rendering of hawk alerts. One in the header and one in the drupal announce, both of which do not have an appropriate landmark. We should wait for a core solution for the Drupal Announce area (https://www.drupal.org/project/drupal/issues/3098857) but try and address the block version.

## Proposed solution

- [x] Update the JS block version with role="region" and aria-label it.

### LINKED ISSUE #1899: Hawk Alert accessibility issues
# Problem

Our Hawk Alert implementation has a couple problems with how we are inserting Hawk Alerts into Drupal's message system. First, a Siteimprove rule is be triggered for the `#drupal-live-announce` region: https://alfa.siteimprove.com/rules/sia-r54. The rule and subsequent testing seems to indicate that this might be getting triggered because we are inserting Hawk Alerts into the region with full HTML markup. Second, Hawk Alerts are getting wrapped in with a `role="alert"` which automatically causes them to be announced separately implicitly using `aria-live="assertive"`. Testing with VoiceOver indicated that this was resulting a broken and unpredictable user experience where all the `assertive` messages interrupt each other. This can result in only one Hawk Alert being read aloud in its entirety (if there are multiple).

## Additional information
Siteimprove rule: https://alfa.siteimprove.com/rules/sia-r54

## Steps to reproduce
- Visit https://admissions.stage.drupal.uiowa.edu/. If you don't see a couple Hawk Alerts, you may need to run the following command:
```
ddev drush @admissions.test config:set uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active
```
- Turn on VoiceOver (Cmd-F5)
- Reload the Admissions staging website.
- There are 3 messages that should be read out loud: The concatenated `#drupal-live-announce` version of the Hawk Alerts and both of those same messages separately via the `role="alert"` attribute. Obviously, its not the desired outcome that all 3 of those messages would be getting populated, but if they are all going to exist, they should all be ready out by VoiceOver. But you should hear that only one of them gets read aloud.
- Repeat a few times to hear how it changes which message gets read aloud.

# Proposed Solution

There are multiple ways we could go about fixing it, but this is the one that I think results in the best outcome and user experience:

```js
messages.add(alertContent, {
  id: id,
  // type: 'error', <-- remove this
  dismissible: false,
  announce: Drupal.t('Hawk Alert: @alert', { '@alert': item.attributes.alert }) // <-- add this
});
```
By removing the `type: 'error'`, we would no longer cause the individual Hawk Alerts to get wrapped in `role="alert"` when they are inserted into the status message region of the page (where they are fully marked up to display as HTML). This prevents the problem where the alerts are doubled in the live announcement system. By setting the `announce` property to a non-HTML string, they will get fed into the `#drupal-live-announce` region consistently with how other messages are populated and this will hopefully be enough to stop triggering the Siteimprove rule.

<details>
<summary>Original issue summary</summary>

## Problem

hawk alert needs aria-atomic="true" on role=alert element

## Proposed Solution

From https://github.com/uiowa/uiowa/issues/1899#issuecomment-654428792, we might not be able to do this without an update to Core's messenger API.
</details>

richardbporter:
Resolves #1898 

Related to #1899

### Testing
- Check out main.
- `blt sync`
- `drush cset uiowa_alerts.settings hawk_alert.source 'https://emergency.stage.drupal.uiowa.edu/api/active.json'`
- Run SiteImprove locally on default site on main branch. 
- See ARIA landmark errors regarding perceivable content in Hawk Alerts.
- Check out this branch.
- `yarn workspace uiowa_alerts build`
- Refresh page and run SiteImprove report again.
- See errors are not present and `hawk-alert-message` divs have ARIA role and label.


pyrello (approved):
Visually tested.

PR #3552, opened 2021-05-20
Wire up featured content heading size field.

richardbporter:
Related to #2259 

This is just how my brain wanted to solve this. If there is a referring item going the other way around, that'd be fine too, just not seeing it.

### Testing
- `blt sync`
- Add a featured content block to page referencing >= 2 articles/people.
- Test out the usual headline field stuff:
  - With no headline value and different child heading sizes.
  - With a headline value and setting various sizes for it. Then inspect the children heading sizes.
  - Make sure the headline changes are as expected after saving the page.


joewhitsitt:
pulling down to test

joewhitsitt:
This works and I don't see any visual regression caused by this method. I think this would be a good enough solution for now that we can easily switch out if a better solution comes along.

Anyone **not** want this solution and would rather wait or keep looking?

pyrello:
I was curious to test it out, but don't wait on me if you think this works.

joewhitsitt (approved):

PR #3553, opened 2021-05-20
More Allowed Table Attributes

### LINKED ISSUE #2826: Allow more table attributes
# Story

As a table wizard, I can provide additional attributes to my tables, captions, summary to make them more accessible to all.

These suggestions were provide by Todd recently.

# Proposed solution

- [ ] ability to set ID for individual cells (r1c1?)
- [x] scope/ID/headers
- [x] role [presentation, grid]
- [x] aria-label/labelledby

joewhitsitt:
added more table attributes, resolves #2826

`<table>` now allows `id` `role`(presentation and grid only) `aria-label` `aria-labelledby`
`<th>` was allowing `scope` but now allows `id`
`<td>` now allows `headers`

I couldn't find an accessible use-case besides serving as an anchor for allowing more granular `id` capabilities so I left that off for this pass.

## To Test

- Using a table, an editor can manually add these attributes to table markup through the source display and not have the attributes stripped on render.
- I am not aware of any site splitting the filtered_html config.

### DIFF
--- a/config/default/filter.format.filtered_html.yml
+++ b/config/default/filter.format.filtered_html.yml
@@ -64,7 +64,7 @@ filters:
     status: true
     weight: -49
     settings:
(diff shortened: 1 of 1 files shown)


richardbporter (approved):
Works for me.

Issue #1955, opened 2020-07-17
Update basic page teaser template to use uids card


joewhitsitt:
## Problem

Of the three default content types, page is the only one not using the card component.

## Proposed Solution

- Change the twig markup to match the card component similar to article, person.
- Account for the styling changes based on the different markup for "featured content" list/grid/masonry and anywhere else a page teaser is called.

joewhitsitt:
I think this was resolved in https://github.com/uiowa/uiowa/pull/3555

PR #3555, opened 2021-05-20
Allow pages in featured content block.

richardbporter:
Relates to https://github.com/uiowa/uiowa/issues/1955

This allows basic pages to be referenced in the featured content block. I had to adjust the node--page--teaser template to account for v2/3 differences.

# How to test

**Make sure you have the latest version of UIDS:**

```
rm -rf ./node_modules
yarn cache clean
yarn install
```

**v3**
- `blt sync` a v3 site of your choosing.
- Create a page and add the Featured content block with some **page** and **article** references.
- Check the display. It should card-ified with media-right, widescreen small.

**v2**
- `blt sync --site hawkeyemarchingband.uiowa.edu`
- Create a page and add some Featured content through paragraphs with **page** and **article** references.
- Check the default display. Both **page** and **articles** should appear in a list.  Try adjusting grid and masonry settings and verify that they appear stacked. 
- Check https://hawkeyemarchingband.local.drupal.uiowa.edu/news and verify that list looks like prod but without line dividers. 
- Update Articles list on https://hawkeyemarchingband.local.drupal.uiowa.edu/ to use grid or masonry and verify that card is stacked.


joewhitsitt:
poking to see how this affects v2/hr.

it might be worth switching v2 to a node--page--teaser--v2 type template or something, otherwise we might have to deal with differences between the two
Screen Shot 2021-05-20 at 4 06 35 PM

richardbporter:
Much confuse. HR looks to be using the UIBS page teaser template so that site should be ok right?

I'm testing hawkeyemarchingband with the old template before these changes but something isn't right.

richardbporter:
This can wait until next week. I'd be curious to get @bspeare's thoughts on the template/preprocess changes.

bspeare:
I took a quick look at hawkeyemarchingband and noticed a couple of possible regressions.  I'd like to take a little more time after standup to verify.

richardbporter:
Thanks. We were also curious if any CSS could be possibly be deleted along with this change. Can't remember exactly why that was but maybe @pyrello remembers?

richardbporter:
Sounds like we may have caused the regression with v2 grid person reference. I guess we should set this back to draft and address it here.

richardbporter:
I'll work to get #3552 merged and then integrated in this PR for page titles.

bspeare:
The issue with v2 is that for Page and Article the card styles were being manually set in CSS based on whether grid, or masonry was set on the Paragraphs.   Are we able to preprocess in those card classes based on the grid or masonry choice or should we just fix those in the CSS?

richardbporter:
> The issue with v2 is that for Page and Article the card styles were being manually set in CSS based on whether grid, or masonry was set on the Paragraphs. Are we able to preprocess in those card classes based on the grid or masonry choice or should we just fix those in the CSS?

I'm sure we could but I don't know what exactly would be needed or how we broke this with the last PR. Maybe we could touch base today.

pyrello:
Since the grid or masonry settings are just a field value, it should be pretty straightforward to check it and add a condition to set classes based on the value.

bspeare:
@pyrello that sounds good.  If the setting is grid or masonry, the cards need to have the stacked values https://github.com/uiowa/uiowa/blob/7a2c1478d9b6f3fcb331490c5f0724ad3114a4eb/docroot/themes/custom/uids_base/uids_base.theme#L251-L252.

bspeare:
As we have discussed a little with the people list, cards are not built for having border-bottom separators like the list pages have had in the past.   Do we want to add the padding/margin back in to fix the style regression in the screenshot below or remove the line separators and create a line separator variant for the card component?

Screen Shot 2021-05-24 at 11 33 15 AM

pyrello:
I have not accounted for the "News" page yet, but I have this working for articles in terms of showing the correct classes based on the classes chosen by the user. However, it is requiring a cache clear to see the updates, so I guess we need to also trigger cache clearing for something, but I don't have that figured out yet.

pyrello:
News page looks okay. I think that was something I figured out last night and just forgot this morning. It doesn't trigger the `$is_stacked` condition, so doesn't get those classes.

joewhitsitt:
Did the strategy change since this morning?

bspeare:
@joewhitsitt I updated the pr to remove the card classes for v2 articles so that they can use the CSS that was previously working in `uiowa-paragraphs-lists.scss` as a way forward to bypass the caching issue and pager issue.

joewhitsitt:
pulling down to test

joewhitsitt:
Small issue combining content types in featured content v2. Not sure if it is a blocker though:
Screen Shot 2021-05-25 at 12 59 20 PM

After seeing this, I pulled main again and did all the same steps to test with the same content:
Screen Shot 2021-05-25 at 1 14 45 PM

bspeare:
@joewhitsitt I added spacing fix for those person teasers in featured content.

joewhitsitt (approved):
Let's move forward!

PR #3556, opened 2021-05-21
FontAwesome Role Presentation

### LINKED ISSUE #1320: Font Awesome <svg> arrow a11y
<svg> arrow should use role "presentation", not "img". Assigning "img" role triggers ALT requirement, but ALT text is unnecessary as the purpose and content of the link are satisfied by the child text node (e.g., "SOFA"). Note: this is probably a better solution than FontAwesome in this use case.

FontAwesome recently updated to to not strip out `role="presentation"` from the SVG icon that is rendered. https://github.com/FortAwesome/Font-Awesome/issues/14791

## Proposed Solution

- [x] Make sure our fontawesome dependency is up to date
- [ ] Audit our existing usage to make sure that classes don't break with the update
- [x] Updated _our_ markup to include `role="presentation"` and test to make sure it is not longer stripped out.

## Follow-up
- [x] Create a WYSIWYG Filter plugin that adds role presentation attribute to spans with the fa class.
- [x] Update filtered_html and minimal to allow role attribute on `span`

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/1320

Short of hearing back on https://github.com/uiowa/uiowa/issues/1320#issuecomment-846075644, I decided to move forward with the quick wins.

## ToDo

- [x] Event icon field
- [x] Happy config import for both install/sync

## To Test

- Anywhere you see a font awesome icon (except existing footer contact content) you should see the svg markup with `role="presentation"` instead of `role="img"`
  - Places to look:
    - Button arrows
    - Person Card Details
    - Event information
    - Social media menu items
    - ...
- In the WYSIWYG if you place an icon manually `<span class="fa-user fas"></span>` on render it should include `role="presentation"`. If you place an icon manually and try to add a role yourself that isn't presentation, it will override with the correct "presentation" value.
- Search the docroot for icons that I might have missed (our code base, not contrib).
- **Note:** FontAwesome itself has not been updated as part of this PR so icon classes haven't changed, etc.
- Not sure if any a11y checker (siteimprove, webaim, lighthouse) is still flagging this, but you shouldn't see alt text errors for icons.


bspeare:
Pulling down to test.

bspeare:
@joewhitsitt This works with Minimal and Filtered, should it work for Full_HTML too?

joewhitsitt:
good catch. omission on my part

joewhitsitt:
can't seem to get both install and sync to be happy with the minimal config file..

richardbporter:
I think cfd6563 and bf8ce26 resolved the config export differences with fresh installs and whats in prod. Sounds like a core issue: https://www.drupal.org/project/drupal/issues/2855675 The code in cfd6563 only affects the sort of the exported config, not the actual sorting the filters execute in. That is still controlled by the weight.

Now I'm running into an issue where disabled filters like lazy are being added to text formats when simply resaving them without any changes... 🤔

joewhitsitt:
taking a look

joewhitsitt:
Nice! I am not getting the lazy issue.

`drupal:install`, `uli`, resave config, `cex` = identical
`ds`, `uli`, resave config, `cex` = identical

Not sure how you found that issue btw.

richardbporter:
> Now I'm running into an issue where disabled filters like lazy are being added to text formats when simply resaving them without any changes...

Sorry, this was inaccurate. It happens when reordering filters on a text format in the UI. However, it seems to happen on main in my testing so I don't think it is related to this PR or the hook I just implemented.

So like, rearrange two filters for filtered HTML and then export config. Turns on lazy_filter for some reason.

richardbporter (approved):
I visually compared text format filter order for all text formats. Looks identical. 

Sounds like https://github.com/uiowa/uiowa/pull/3556#issuecomment-847304197 is a problem in main right now so it doesn't need to be fixed in this PR necessarily.

PR #3560, opened 2021-05-24
added git push command to output after site revision

### LINKED ISSUE #3100: Add git push command to output after site provision
# Idea

Add the following to the output that is shown at the end of the `uiowa:multisite:create` command:
` * git push --set-upstream origin branch_name`

dnkemper:
Closes #3100 

# How to test
1. Run the blt umc command with a dummy site - `blt uiowa:multisite:create test.law.uiowa.edu --requester=dnkemper`
2. Go through normal commands you would go through to provision a site - at end, should give you the command to push up the specific branch you're on ie Push this branch and merge via a pull request: git push --set-upstream origin i3100-git-push

3.  After testing, be sure to run the `blt umd` command to deprovision the dummy site


blt/src/Blt/Plugin/Commands/MultisiteCommands.php
```diff
@@ -704,14 +704,25 @@ public function create($host, array $options = [
...
+
       $steps += [
-        'Push this branch and merge via a pull request.',
         'Coordinate a new release and deploy to the test and prod environments.',
       ];
     }
 
     $this->say("Committed site <comment>{$host}</comment> code.");
     $this->say("Continue initializing additional multisites or follow the next steps below.");
+    $this->say("Push this branch and merge via a pull request: <comment>git push --set-upstream origin {$branch}</comment>");
```
pyrello: Instead of having this here, can we add it as the first item in `$steps` below?

dnkemper: I tried that and it either wouldn't call the branch variable or it would still place it above the steps regardless of where I put it :( If you can figure that out though, I'd be so open to learning how!

pyrello:
The `$branch` variable is only available if the `--no-commit` option isn't used, so I moved the message into that conditional section.

dnkemper:
@pyrello That makes sense! Thank you so much!

pyrello (approved):
Tested this and it works well!

dnkemper:
Thank you for your help, @pyrello

PR #3563, opened 2021-05-25
Started working on pdfs for grad

### LINKED ISSUE #3009: As site visitor to grad.uiowa.edu/academics/manual, I can download a PDF of the entire manual.
# Story

As site visitor to https://grad.uiowa.edu/academics/manual, I can download a PDF of the entire manual.

# Proposed solution

Leveraging entity_print, create a download link on the top level page that gets all the child pages and generates a PDF document with everything in it. The entity_print module does this on the content overview page when selecting multiple pieces of content. We should be able to dig into how that works and do it ourselves.

dnkemper:
Closes #3009 

# How to test

### DIFF
--- a/docroot/sites/grad.uiowa.edu/modules/grad_core/grad_core.module
+++ b/docroot/sites/grad.uiowa.edu/modules/grad_core/grad_core.module
@@ -7,6 +7,7 @@
 
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Template\Attribute;
 
 /**
  * Implements hook_entity_presave().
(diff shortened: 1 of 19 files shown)


docroot/sites/grad.uiowa.edu/modules/grad_core/grad_core.module
```diff
@@ -23,6 +24,22 @@ function grad_core_entity_presave(EntityInterface $entity) {
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function grad_core_preprocess_entity_print(&$variables) {
+  if ($variables['theme_hook_original'] == 'entity_print__node__page') {
+    // The debug path nests this differently for some reason.
```
richardbporter, line 32: This was a bad comment on my part. It would be better to explain why we do this like I just did in 49c87b4ce3be2900477499958f6e86bdcf53eba7.

PR #3566, opened 2021-05-25
Bulk moderation state actions

### LINKED ISSUE #696: Bulk update moderation states
Customer created all content in draft and wants to bulk update to published. This currently isn't available.

cory-skeers:
Resolves #696

Remove non-functional un/publish actions, add functional moderation state transition actions.

# How to test

`git pull && composer install`
Sync a site
Log in and visit /admin/content
Check that the generic "Publish" and "Unpublish" options in the Actions dropdown are no longer present
Select some content and try out the new options--Archive current revision, Unpublish latest revision, Publish latest revision
Masquerade as different roles, and check that you cannot perform operations you don't not have permissions for (eg Editor role cannot Archive content)

Possible followups:
Only show options that are available to the current role? Good UX cleanup, but not default behavior of the Actions menu.


dnkemper (approved):
Your timing is truly impeccable! Nice work!

PR #3580, opened 2021-05-28
Update hook 9009 fix for some erroring sites

cory-skeers:
# How to test

Run the update hook on a site that doesn't have an existing `key_value_expire` table and check that it completes without error


richardbporter (approved):
Tested with endingviolence.uiowa.edu. It re-ran with no errors and 0 records found.

PR #3587, opened 2021-06-01
Add person type and person type status filters to person list block

### LINKED ISSUE #2524: Allow person type/former filtering on People View Block
# Story

As a webmaster, I want to be able to filter people by person type/statuses when creating my pages with paragraphs or layout block content.

# Proposed solution

- [ ] Add the person types and person types status fields to the Person list view filters and expose them.
- [ ] Need to figure out a way to generalize the solution that was put in place to improve the UX of the person types field.
- [ ] Update preprocess_paragraph in SiteNow People to account new view contextual filter options.

### LINKED ISSUE #3519: Clean up people block config, noodle code, etc.
# Motivation

After completing #3518, the `uiowa_people` block type and any noodle code, configuration, or specific styles that were necessary to support it can be removed.

# Proposed solution

- [x] Replace preprocess_block logic in SiteNow People related to people block.
- [x] Remove People block content type.
- [x] Remove fields that are unique to People block content type.
- [x] Re-export config.

### LINKED ISSUE #3607: Re-run people block update and remove the block
# Motivation

With the people block update hook, we didn't remove access to the old block instance, so folks could have been placing them since we ran the update.

# Proposed solution

Create a new update hook with the same logic as before. And also remove access to the block this time.

pyrello:
Resolves #2524
Resolves #3519 
Resolves #3607 

# How to test
- `blt ds --site sandbox`
- Observe running of update hook without issues.

# Stories (All V3)
- As an editor, I cannot place or access the old version of the people block.
- As an editor, all of my people blocks use the new version of the block.
- As an editor placing a people list block, I can limit the list to just people of a certain type.
- As an editor placing a people list block, I can see the appropriate person type status based on the type I selected.
- As an editor placing a people list block, I can limit the list to just people of a certain type status.
- As an editor editing a people list block, if I have already selected filters, I see the Filters section as open.
- As an editor, I can use all of the above options with paging enabled on the block and it is not broken.


pyrello:
This seems to be working now. However, both @joewhitsitt and I had experienced some weirdness with the person types status not getting set correctly on the block update form, so any testing should probably focus on that a bit more heavily.

pyrello:
@joewhitsitt clarified that the way this is not working is that when you first add the block and set a value for person type status, it doesn't get saved. It does work on subsequent block update saves.

joewhitsitt:
i think i got it swapping #access for #states

downside: it shows for person types that don't have a former option but defaults to any

pyrello:
I believe all the outstanding issues have been resolved.

joewhitsitt:
I have tested this PR but because of my involvement on it would be great to get another reviewer on this.

mark-bennett-uiowa:
@joewhitsitt @pyrello I will pull this down to test right after lunch, unfortunately had meetings all morning.

cory-skeers:
Pulling down to test. If you're still looking to test it too, @mark-bennett-uiowa, don't let me stop you!

cory-skeers:
Can we add the People List block to the direct add list of blocks as "People," so an editor can add the equivalent block from the same spot post-update?

cory-skeers:
When editing an existing block with one person type selected, if you change the person type and try to set a person type status other than `- Any -`, it creates a validation error "An illegal choice has been detected. Please contact the site administrator."

It looks like the select list text correctly updates to the new person type, but the value it tries to submit is based on the former person type.

Steps to reproduce:

- Create a people list block, filter to Faculty
- Save the block
- Edit the block
- Change filter to Graduate Assistant and Person Type Status to Former Graduate Assistant
- Try to save the block

mark-bennett-uiowa:
@joewhitsitt @pyrello Ran my own testing on this also. Looks great from an approval PoV. I did have these follow ups along with @cory-skeers:

> Can we add the People List block to the direct add list of blocks as "People," so an editor can add the equivalent block from the same spot post-update?

Would like this also, especially with the previous People block option being removed with this PR.

> When editing an existing block with one person type selected, if you change the person type and try to set a person type status other than `- Any -`, it creates a validation error "An illegal choice has been detected. Please contact the site administrator."
> 
> It looks like the select list text correctly updates to the new person type, but the value it tries to submit is based on the former person type.
> 
> Steps to reproduce:
> 
> * Create a people list block, filter to Faculty
> * Save the block
> * Edit the block
> * Change filter to Graduate Assistant and Person Type Status to Former Graduate Assistant
> * Try to save the block

Also ran into this same issue, but had to follow Cory's steps exactly. If you change the Person Type prior to saving or leave out the Person Type Status option, this doesn't happen.

We should definitely figure out the ability to select multiple person types at some point. For instance, Graduate Assistants and PhD Candidates both fall under Students as "sub-categories". An editor may fail to check both Student and PhD Candidates, or Student and Graduate Assistant, when creating a person node. If they are only creating one Students page using the Student person type filter, then some results may get ignored unless they checked the Student person type.

Other than these above, I seemed to throw the kitchen sink at this and it worked really well. Has my approval for merge if we want to create follow up issues on some of these concerns above.

mark-bennett-uiowa (approved):
Approved, as long as concerns above are understood.

docroot/modules/custom/sitenow_people/sitenow_people.module
```diff
@@ -125,15 +130,117 @@ function sitenow_people_form_alter(&$form, FormStateInterface $form_state, $form
...
+
+        $type = $form_state->getValue([
+          'settings',
+          'override',
+          'exposed_filters',
+          'field_person_types_target_id',
+        ]);
+
+        if (!$type && !empty($element['field_person_types_target_id']['#default_value'])) {
+          $type = $element['field_person_types_target_id']['#default_value'];
```
cory-skeers, line 243: It looks like when the field_person_types_target_id changes in an open form, the `$form_state->getValue(...)` ends up returning `null`, and so it updates the type to the default value, which is whatever the form was previously saved as. Still unsure why the field appears to update correctly, but then submit with this reverted value.

cory-skeers (changes requested):

docroot/modules/custom/sitenow_people/sitenow_people.module
cory-skeers: I made an update to the `$type` setting and checking. It functions as expected now, but will throw a form value warning when a person type status is set and person type changes (say, "Former Graduate Assistant" is selected and person type is changed to "Faculty"--the person type status options will update to "Any" and "Former Faculty", and can be submitted correctly, but a warning that "An illegal choice has been detected. Please contact the site administrator." will be displayed on the field.

pyrello:
> Can we add the People List block to the direct add list of blocks as "People," so an editor can add the equivalent block from the same spot post-update?

@cory-skeers I'm not sure that I agree with just adding the people list block back in with the other direct add options. I would like to see either:
a) All the list type options are listed under "More... > Lists"
b) I get [this MR](https://www.drupal.org/project/lb_direct_add/issues/3098555) finished for the `lb_direct_add` module which allows for the option of having other options besides custom blocks listed in the direct add widget and that introduces the option to have groups in the DA widget like "Components" and "Lists"

pyrello:
> We should definitely figure out the ability to select multiple person types at some point. For instance, Graduate Assistants and PhD Candidates both fall under Students as "sub-categories". An editor may fail to check both Student and PhD Candidates, or Student and Graduate Assistant, when creating a person node. If they are only creating one Students page using the Student person type filter, then some results may get ignored unless they checked the Student person type.

@mark-bennett-uiowa  The reason I did not move ahead with multiple value support at this point is because it requires us to take even more ownership of manipulating the form and form state throughout the form submissions cycle. If we switch that field to accept multiple, then we have to hook into form validation to allow that to work correctly and then we may have to contend with the fact that the form submission will be expecting a single value.

If you look at the view settings, there is no option to change this filter to a multiple value field, as is the case with other filters. 

Rather than investing in that ourselves, I would suggest that we put our efforts behind enabling full config entity support via reference fields in this issue: https://www.drupal.org/project/drupal/issues/2429699. That support is indicated in the OP, but I can't find any evidence that it was actually implemented that way, as my comment indicates: https://www.drupal.org/project/drupal/issues/2429699#comment-14122220

Your comment about sub-categories also would seem to indicate that the way we are handling this in the first place is wrong. Maybe we should add a property to the person type to allow setting a parent and thus enabling the construction of a hierarchy?

docroot/modules/custom/sitenow_people/sitenow_people.module
pyrello: The fix here feels wrong to me, as the comments above the `$form_state->getUserInput()` method seem to indicate that there is almost no reason to use it instead of `$form_state->getValues()` instead:
```
   * These are raw and unvalidated, so should not be used without a thorough
   * understanding of security implications. In almost all cases, code should
   * use self::getValues() and self::getValue() exclusively.
```

@cory-skeers Since we are calling the `getValue` version of this directly above that and it is failing to produce the correct value sometimes, that leads me to believe that there is additional processing necessary in `person_types_callback` to ensure that there is alignment between the values expected and the values provided:
https://github.com/uiowa/uiowa/blob/02ed5a20c64decd337254fa894c1b4afd2842e9c/docroot/modules/custom/sitenow_people/sitenow_people.module#L249-L251

I am in agreement that we should figure this out before releasing it because it indicates a potentially larger problem in the way that we are handling the AJAX callback that could be more problematic than what we have discovered so far.

cory-skeers:
> > Can we add the People List block to the direct add list of blocks as "People," so an editor can add the equivalent block from the same spot post-update?
> 
> @cory-skeers I'm not sure that I agree with just adding the people list block back in with the other direct add options. I would like to see either:
> a) All the list type options are listed under "More... > Lists"
> b) I get [this MR](https://www.drupal.org/project/lb_direct_add/issues/3098555) finished for the `lb_direct_add` module which allows for the option of having other options besides custom blocks listed in the direct add widget and that introduces the option to have groups in the DA widget like "Components" and "Lists"

My personal preference in the long run would be b, but I'm not opposed to a, or moving forward with this as is for now as long as there's some communication about the change. As it is, it could look like an editor no longer has the option to add people blocks, and there aren't a lot of reasons to go into "More..." yet. I think I'll feel it's less important as more list blocks are added.

pyrello:
> My personal preference in the long run would be b, but I'm not opposed to a, or moving forward with this as is for now as long as there's some communication about the change. As it is, it could look like an editor no longer has the option to add people blocks, and there aren't a lot of reasons to go into "More..." yet. I think I'll feel it's less important as more list blocks are added.

This is something that needs to be reflected in the updated documentation, as I missed this detail when I was updating it.

pyrello:
I think this is ready to be tested again. Nice work, @joewhitsitt on figuring out the workaround!

@cory-skeers I think you'll have to explicitly approve it since you requested changes previously.

pyrello:
Also, grad.immuno.uiowa.edu is a site where we are already using the person types and person type statuses extensively:
- https://immuno.grad.uiowa.edu/people/faculty
- https://immuno.grad.uiowa.edu/people/students
- https://immuno.grad.uiowa.edu/people/students/former

joewhitsitt:
Testing it all together.

Capturing thought for a later issue: See the need for a "current" student, etc. status. Immuno.grad has students marked with former. Former students show in the student type filter (any status). What if I wanted a list of current students that didn't contain former students as well?

> Can we add the People List block to the direct add list of blocks as "People," so an editor can add the equivalent block from the same spot post-update?

I think moving away from block types building block types all the time, there is going to be more to find in the More... section so this will be less noticeable over time. Articles will eventually be there as well. These types of blocks will be used less on a page-by-page basis. Creating a distinction between content blocks and "dynamic" or referenced blocks I think makes sense in the long run.

joewhitsitt (approved):
Didn't know what site to test the update hook on but it is just rerunning what has previously run successfully in PROD and the old block is indeed getting removed this time.

shrug ftw

cory-skeers (approved):

PR #3606, opened 2021-06-01
Simple sticky table header experiment

### LINKED ISSUE #3576: Drupal sticky headers scroll sync experiment
Create an experimental implementation with using Drupal sticky headers to create responsive sticky headers that scroll vertically and horizontally for basic tables only. **This does not need to account for sticky first column or tables with `colspan`s.**

dnkemper:
Closes #3576 

# How to test

Visit https://default.dev.drupal.uiowa.edu/table-test and view the page with the browser shrunk to a mobile device size.


pyrello:
I think I got this experiment working. Need to add the `.sticky-enabled` class to a table to test it.

richardbporter:
Going back to tables.js, I've wondered if it would be possible to attach the library only if the text field contained a table. Otherwise it gets loaded on a lot of pages - any page with a textarea field, for example the homepage.

pyrello:
To be clear, the intention behind this is to see if we can get a solution in place that works for simple tables (just sticky header row, no `colspan`s or sticky first column). The experimental link above contains some flaws that would need to be corrected before we rolled this solution out. Specifically, the sticky header visually overlaps the responsive table container.

Before we make additional investment in the solution, we would like to get feedback from @inspector508 about whether the sticky header solution from Drupal core is considered accessible enough for us to move ahead with it.

bspeare:
@GaryRidgway Can this be moved out of blocked/waiting since you met with Todd?  What are the next steps for this?

bspeare:
Closing this in favor of https://github.com/uiowa/uids/pull/584.

PR #3612, opened 2021-06-03
Add additional default person types (adjunct, visiting, postdoc, alumni)

### LINKED ISSUE #3310: Add additional types as default options in Person Types
I anticipate that we will all have one-off examples of person types that we may run across around campus. However, Alumni seems to be an option that would be campus-wide, whether it is used or not for a particular site.

UPDATE: Adding the following as additional types:
Alumni
Adjunct Faculty
Visiting Faculty (may be combined with Adjunct Faculty)
Postdoc (or Postdoctoral Researcher)

Would this be an acceptable option to add to the default configuration as a standard option?

mark-bennett-uiowa:
This is based off of academic site needs for CLAS and Nursing (maybe others). Resolves #3310.

Faculty already covers full faculty, associate faculty, assistant faculty, lecturer, and other instructional track faculty.
Adjunct faculty is needed to cover secondary/tertiary appointments or people who are not technically part of the primary faculty.
Visiting faculty is needed to cover temporary appointments, external to the University, often without HawkIDs.
Postdoc is needed to cover post doctoral graduates who may be performing research on campus.
Alumni is self-explanatory.

Sampada put this best, when she mentioned "Prospective students can currently see Adjunct or Visiting Faculty listed as Faculty, and are interested in their research, but by the time they get here, the faculty member is gone. Its false advertising."

# How to test

- Pull PR.
- Check that these four new Person Types exist under Admin > Structure > Person Types.
- Check that these four new Person Types exist when adding a new person.
- Check that these types appear at the bottom of an individual profile when checked.


dnkemper:
I can test this after meetings today!

dnkemper:
Does it matter that you can still check former adjunct faculty (person type status) when you deselect the adjunct faculty person type? I mean, it doesn't show when you save the person then go back to edit it? So assume it's okay and will approve! Nice work! 
Screen Shot 2021-06-03 at 1 50 57 PM

dnkemper (approved):

joewhitsitt:
>Does it matter that you can still check former adjunct faculty (person type status) when you deselect the adjunct faculty person type? I mean, it doesn't show when you save the person then go back to edit it?

i would say that is a problem

if you compare it to the behavior of staff. when staff is unchecked, the former student option should no longer be visible.

mark-bennett-uiowa:
@dnkemper @joewhitsitt That is a problem that was already there, and exists with the other "former" options also. Though it seems to happen randomly for some reason, depending on the order of checking and unchecking options above. Ive been testing it for several minutes now, and only randomly get it to occur. It seems like an AJAX issue.

dnkemper:
If that was already there with other person types then, can we just create a new bug issue for it?

mark-bennett-uiowa:
We could, it looks like it is displaying the "former" checkbox options based on previously saved Person Types values. For instance, I have a test person in here that was saved with Adjunct Faculty, Faculty, and Visiting Faculty, and no "former" options checked. If I uncheck one of the Person Types, the "former" option goes away. If I uncheck all three Person Types, then all three previous "former" options re-appear but are unchecked, even though I dont have any Person Types checked at the time. I should have Alan teach me to record this so I can show off this behavior.

joewhitsitt:
Haven't had time to test this myself but bug aside, can these be split or status set to disabled by default until we have the larger conversation about whether these should be available to all sites?

Based on the conversation at demo, there is still a desire to align with the central people solution and provost is setting criteria for faculty.

mark-bennett-uiowa:
I guess that would be possible, but I would need this on nearly ever CLAS site, Sam needs these in Nursing sites, and others mentioned needing these, at least until the entire campus moves to the new Directory Profiles. Also, I dont see the difference between these and say Graduate Assistant or PhD candidate which are already part of the default profile.

mjoneill:
> I guess that would be possible, but I would need this on nearly ever CLAS site, Sam needs these in Nursing sites, and others mentioned needing these, at least until the entire campus moves to the new Directory Profiles. Also, I dont see the difference between these and say Graduate Assistant or PhD candidate which are already part of the default profile.

I agree. 

I'm not sure I understand the difference either.  This may be a signal that we should discuss something further and get a better understanding before making large commitments.

Let's find some time to invest.  @mark-bennett-uiowa would it make sense to include Sam if I can find some time for us to explore further?

mark-bennett-uiowa:
Absolutely @mjoneill invite Sam. Some of this was in response to a meeting she and I had yesterday in regards to how to create something for the Nursing site.

pyrello:
I set this to draft again so that it isn't merged prior to meeting about how we want to proceed with person types.

mark-bennett-uiowa:
Approved verbally by Sean after his assistance on this latest change.

PR #3621, opened 2021-06-04
Create Panopto media type and update permissions for SiteNow v3.

### LINKED ISSUE #3392: Embed UICapture/Panopto videos
# Story

As a SiteNow content editor, I would like the ability to embed UICapture/Panopto videos on my website similar to YouTube and Vimeo.

Panopto does not provide an oEmbed option.

# Proposed solution

- From what we learned with https://github.com/uiowa/uiowa/issues/3450, create a generic iframe or Panopto media type where an editor provides a share URL and the url is parsed into a rendered embed. Panopto provides an embed api instead of us maintaining an iframe object ourselves.

# Nice to have

- Embed Panopto Playlists
- Embed a Panopto Department

The second point is that I think we should explore is the ability to embed playlists. It uses the same "Embed.aspx" player, but the "id=" changes to "pid=" so maybe this could be a checkbox when users are entering their information. I didn't think of that initially, but I think it would be good to at least explore. Here is an example of the embed player loaded with a playlist: https://uicapture.hosted.panopto.com/Panopto/Pages/Embed.aspx?pid=f5642cc1-3863-4171-8fa0-acbd0102df02

The last thing that I had was something to support users who might want to share using the full player without needing to support embedding that player directly. If one embeds a folder, users that select content and click-through will be directed to the full player. That will also allow people to embed a Panopto "Department" within their existing sitenow sites. Below is an example of a department in an iFrame:
`<iframe src="https://uicapture.hosted.panopto.com/Panopto/Pages/EmbeddedList.aspx?folderID=cbca7443-6b29-4931-83be-acbd010219f2&embedded=1" height="405" width="720" style="border: 0;" ></iframe>`
Folders diverge from the format of videos and playlists, but I think this would be a better route than trying to embed the "Viewer.aspx" player. Much like the playlists, I don't think this is a "need" but I could see it being useful so it might be good to dig into some.

# Resources
- https://uicapture.hosted.panopto.com/Panopto/Pages/Sessions/List.aspx# (click a video and see/click share icon for URL to parse)
- https://www.codimth.com/blog/web/drupal/create-ckeditor-custom-plugin-drupal-8
- https://support.panopto.com/s/article/Learn-About-Panopto-Embed-API
- https://www.drupal.org/docs/8/core/modules/media/creating-and-configuring-media-types
- https://www.drupal.org/docs/8/core/modules/media/creating-a-custom-mediasource-plugin-for-external-assets
- https://git.drupalcode.org/project/media_entity_remote_image/-/tree/8.x-1.x
- Michael Espey (ITS)

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/3392 for v3.

### Setup
- `composer install`
- `blt dsa`
- `blt frontend`

### Testing
- As an editor/publisher/webmaster on any v3 site, I can embed reusable [UICapture](https://uicapture.hosted.panopto.com/) videos in WYSIWYG fields.


joewhitsitt:
realized my responsive sizing is flawed and causes large padding bottom issues

bspeare:
@joewhitsitt are you able to utilize https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/video/video-utilities.scss instead?

joewhitsitt:
Possibly though the media entity is different and so `media--type-remote-video` won't cover it. I was just trying to get something working because we realized that the width/height attributes weren't responsive and might not be worth capturing on media add.

bspeare:
Sounds good.  Maybe we can develop a more effective mixin that can be used in both places.

joewhitsitt:
Currently built on top of a link field for basic link validation similar to http://drupal.org/project/media_entity_remote_image but I am starting to wonder if that is the best field type to use.

richardbporter:
The thumbnails are visible in the media library but are only created once when the media element is uploaded the first time.

I'd be curious if we dropped the JS approach and constructed the iFrame ourselves if that would get the WYSIWYG preview working and simplify the LB interface preview inconsistency.

richardbporter:
Maybe we should scratch the HR checkbox, at least initially?

> responsive uiowa_bootstrap - doesn't have the same video utilities and has some odd hacks instead. could be improved. HR will be a heavy user of panopto initially.

bspeare:
That does seem like a good follow-up slice.

bspeare:
Does anyone know what Joe had planned for these items?

- change link field description
- "UICapture" co-branding while it is still limited to uicapture.hosted.panopto.com?

richardbporter:
> change link field description

I think the description of what to input when adding a Panopto video could be improved.

> "UICapture" co-branding while it is still limited to uicapture.hosted.panopto.com

I think we should label/brand this as UICapture initially b/c it is restricted to uicapture.hosted.panopto.com. So updating the media library label from Panopto to UICapture would one thing to do there. There might be other spots as well.

richardbporter:
Wow that preview is gigantic...

bspeare:
@richardbporter Try running `yarn workspace uids_base gulp --development`?

richardbporter:
> Currently built on top of a link field for basic link validation similar to http://drupal.org/project/media_entity_remote_image but I am starting to wonder if that is the best field type to use.

I can't remember the reasoning to not use a link field.

richardbporter:
Still need to clean up 46fa470 to use DI, handle Guzzle exceptions and put thumbnails in a subfolder. Also permissions for splits need to be updated.

richardbporter:
I'm trying to add permissions to other splits but not seeing them available on other sites like uiowa.edu.

cory-skeers:
Trying to pull this down to test, but having local env issues (unrelated to this PR--started before pulling the branch, and had thought they'd been resolved, but have persisted). Will plan on moving forward with testing if/when issues get resolved, but in the meantime other testers are welcome if there's capacity.

cory-skeers:
Couple notes (no changes, just notes)

#2346 is present with these, so there's some difficulty telling which videos are being used or not after they've been added. Not specific to this PR--just wanted to note.

If you put a Panopto video in an accordion item, when viewing the page if the video plays and you close the accordion, the video continues to play. Same with all videos, not just Panopto. Not a blocker, just not a 100% end user experience.

cory-skeers (approved):
Tested with various WYSIWYG fields on nodes and blocks, and with the various roles, and all seemed to work as expected. 💯

PR #3629, opened 2021-06-09
Hotfix for `sitenow_update_9011()`

### LINKED ISSUE #3628: Orphaned people blocks after updates to convert them
# Problem
Running the update hook in #3587 results in the following error message on the pharmacy site:
```
  Entities exist of type <em class="placeholder">Custom block</em> and <em class="placeholder">Custom block 
  type</em> <em class="placeholder">People</em>. These entities need to be deleted before importing.  
```

The update hook shows the following, indicating no updates are being made:
```
>  [notice] Message: People block records found: 0
> 
>  [notice] Message: Tempstore records found: 0
```

This seems to indicate that there are orphaned people blocks.

# Proposed solution

Modify `sitenow_update_9011()` as follows:
```php
/**
 * Redo 9009.
 */
function sitenow_update_9011() {
  sitenow_update_9009();
  // Delete remaining orphans?
  $result = \Drupal::entityQuery('block_content')
    ->condition('type', 'uiowa_people')
    ->execute();
  if ($result) {
    $storage_handler = \Drupal::entityTypeManager()->getStorage('block_content');
    $entities = $storage_handler->loadMultiple($result);
    $storage_handler->delete($entities);
  }
}
```

Thanks to @richardbporter for discovering the issue and @joewhitsitt for digging into it.

pyrello:
Resolves #3628 

# How to test

- `blt ds --site pharmacy.uiowa.edu`

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.install
+++ b/docroot/profiles/custom/sitenow/sitenow.install
@@ -3110,4 +3110,14 @@ function sitenow_update_9010() {
  */
 function sitenow_update_9011() {
   sitenow_update_9009();
+
+  // Delete remaining orphans?
+  $result = \Drupal::entityQuery('block_content')
+    ->condition('type', 'uiowa_people')
+    ->execute();
+  if ($result) {
+    $storage_handler = \Drupal::entityTypeManager()->getStorage('block_content');
+    $entities = $storage_handler->loadMultiple($result);
+    $storage_handler->delete($entities);
+  }
 }


richardbporter (approved):
Synced the site without issue. Not sure what to test beyond that. 👍🏻

pyrello:
I think the main question is, are we worried at all about the possibility that these blocks are still referenced in some way that we have not discovered?

joewhitsitt:
So, if this resulted in data loss related to unique body copy, files or something I would be more concerned but the worse case in this situation I believe is a broken reference to blocks in lb that no longer exist. These blocks should not be placed outside of that situation. The editor would find these eventually, remove and recreate.

It is of course puzzling how these orphans or whatever exist.

pyrello:
Possibly related: https://www.drupal.org/project/drupal/issues/3042418

Sites I'm seeing this error on in stage:
- careers.uiowa.edu
- iti.uiowa.edu

pyrello:
Also related: https://www.drupal.org/project/drupal/issues/3005948

After reading through this, I think it is possible that this is a cron timing thing, where the block instance was removed from a node, then cron hasn't run yet. Does that make sense, @richardbporter?

richardbporter:
Maybe. I re-synced on main and ran cron a few times after the config error but the row still exists in the table and I'm unable to import config still.

richardbporter:
Yea I can't get it to clean up with cron. Switching to panopto for now.

richardbporter:
I can't find that UUID anywhere else in the DB. I synced the other sites without errors.

pyrello:
So, it seems like this should be okay. Should we go ahead with it?

richardbporter:
I think so.

PR #3644, opened 2021-06-11
Create uiowa_profiles module and split for integration with profiles client.

### LINKED ISSUE #3286: Directory Profiles Integration
# Story

As a SiteNow site owner, I would like the ability to integrate with the campus Directory Profiles solution, so that I can have information about faculty/staff/etc visible on my website without having to manually create them.

# Additional Info

This should be similar to the existing APR integration that the College of Pharmacy is using on SiteNow v3.

Denny provided some initial information https://git.uiowa.edu/java/profiles/-/issues/174

# Proposed solution

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/3286

### How to test

1. Check out and pull down `i3286-directory-profiles`.
2. Sync down the `default` site.
3. Run `drush @default.local uli`.
4. Enable the Profiles split and import configuration.
5. Navigate to https://default.local.drupal.uiowa.edu/admin/config and, under the `System` header, click `UIowa Profiles Integration`.
6. create a new profiles instance using the `+` button
7. for each profiles instance:  
    7.1. use either of these API keys here for separate instances to check if multiple profiles pages work: https://iowaweb.slack.com/archives/C0164FNN1PV/p1623858986410400?thread_ts=1623858978.410300&cid=C0164FNN1PV.
    7.2. fill out the rest of the form and save configuration
    7.3. Go to the path you previously defined.
    7.4. You should see a Directory profiles listing for microbiology, as that is the API key we have used as a test.
8. verify that you can safely delete profiles instances
9. try and break things.

### Todo
- [x] Move to feature split
- [x] Decide whether or not to render the Drupal page title and breadcrumbs and adjust accordingly. (Passing in page title and breadcrumbs to Vue app may not be supported yet.)
- [x] Fix page size config (Passing this in to the Vue app is not supported yet.)
- [x] Render intro config on directory page (Passing intro config into Vue app is not supported yet.)
  - We can try to render this ourselves but then we'd need to hide it on person pages which could be difficult.
- [x] Fix 404 on individual profile page refresh
- [x] Fix sitemap
- [x] Misc. styling 
- [x] Decide what to do with the canonical link config.
- [x] Split webmaster permissions.


richardbporter:
This is failing because there is no SASS to compile in the workspace build script: https://travis-ci.com/github/uiowa/uiowa/builds/228947761#L1863

Do we envision needing our our styles for this?

bspeare:
Testing

bspeare:
@GaryRidgway I am getting an error that https://default.local.drupal.uiowa.edu/modules/custom/uiowa_directory_profiles/assets/js/basePath.js is missing.    Should I ignore it or is that needed for this module?

GaryRidgway:
is that not in your module files? I thought i committed it up...

GaryRidgway:
hmmm, maybe I didnt, IDK why it wouldnt get pushed up, but I'll find it and get it up there

GaryRidgway:
@bspeare Thanks for catching that, should be working now!

bspeare:
Another issue, when I add the API key and try to save the page I get this error:

Screen Shot 2021-06-16 at 1 53 58 PM

GaryRidgway:
ah yeah we ran in to that too. if you change one letter in there it will work, but i think @richardbporter said it might be a caching issue? I will look in to that

bspeare:
I was able to get all the steps to work.   I will be happy to approve once https://github.com/uiowa/uiowa/pull/3644#issuecomment-862633988 is figured out.  Is there a list of style adjustments that we have or will send to Denny?

GaryRidgway:
I had these three that I made for Denny:

https://git.uiowa.edu/java/profiles/-/issues/222
https://git.uiowa.edu/java/profiles/-/issues/221
and my comment son this https://git.uiowa.edu/java/profiles/-/issues/204/designs/ccom-breadcrumb.png

did you see any others @bspeare ?

bspeare:
There are others, but a lot of it is coming from our default styles.  So I think we will need a `yarn workspace` to compile SASS and make style adjustments.

richardbporter:
I'd probably recommend removing that workspace build if we aren't going to write our own styles. One less thing to run in the Travis artifact build script.

bspeare:
I think a lot or most of the style issues we should be addressing in the module.

GaryRidgway:
the only thing I heard from denny is that he wanted us to let him know style issues so that he could deliver a completely styled solution. I am okay with adding more overrides but that is the strategy I was going for

bspeare:
There are going to be things that we will need to fix within our theme.   At the very least we will need to attach some CSS to this module to fix some issues that are coming from `uids_base`.

bspeare:
Remaining questions on this pr:
- Add a css file for future style updates and remove the workspace?
- Does https://github.com/uiowa/uiowa/pull/3644#issuecomment-862629504 need to be fixed before this is approved/merged?

richardbporter:
> Remaining questions on this pr:
> 
> * Add a css file for future style updates and remove the workspace?
> * Does [#3644 (comment)](https://github.com/uiowa/uiowa/pull/3644#issuecomment-862629504) need to be fixed before this is approved/merged?

Yes. We'l also want to put this module in a split similar to APR since enabling it leaves config different.

richardbporter:
That should be fixed now.

bspeare (approved):
All of the testing steps worked fine.  Nice work!

richardbporter (changes requested):
Still need to put this in a split.

richardbporter:
Also, the sitemap is not working: https://default.local.drupal.uiowa.edu/directory-profiles/people/sitemap.txt

richardbporter:
The directory page size config is not working either. Default is 30 but its showing 10 per page.

richardbporter:
I think we should wire up that sass into a library included on the page, at least in the short term, to adjust styles until they are addressed in the profiles app.

bspeare:
There's one style we need to adjust and that is the `figure` margin that is setting the card image with extra margin.   That is coming from the `normalize.css` in `uids_base` and we need to override that so set `margin: 0` on it.

richardbporter:
I think we've taken this as far as we can. Seems like we need a decision on the breadcrumbs and page title discussions happening in GitLab in addition to the checkboxes I added above.

Another thing to consider is https://git.uiowa.edu/java/profiles/-/issues/116 mentions multiple directories which would require quite a bit of refactoring.

richardbporter:
If we decide to hide the Drupal page title and breadcrumbs, which I'm leaning towards, the controller title callback needs to be updated to remove that from the render array. The title callback might be able to removed altogether but I can't remember if that canonical meta tag needs to be set there or not.

Also, if the styles.scss are not going to leverage uids imports then I would vote to remove that workspace from the build process and just use vanilla CSS.

bspeare:
I added a few styles and leveraged a couple variables from the imports.  Do we want to hide the "People" page title on the individual profile pages or will that be handled by Denny?

Edit: maybe that's what @richardbporter is talking about above.

richardbporter:
> I added a few styles and leveraged a couple variables from the imports. Do we want to hide the "People" page title on the individual profile pages or will that be handled by Denny?
> 
> Edit: maybe that's what @richardbporter is talking about above.

Yes. I updated the checkboxes to highlight this and added some more details.

richardbporter:
I just noticed that the canonical link config is not working correctly. I removed it for now but we may need to do it via JS since clicking a profile does not trigger a page load.

richardbporter:
Wow. Took forever to track down what was causing https://travis-ci.com/github/uiowa/uiowa/builds/232664894#L2221 but it seems like a reference in a totally different file was: ccd650e It only errored when the `uiowa_profiles.module` file was present though... 🤷🏻‍♂️

bspeare:
@richardbporter crumbs should be looking good now.

richardbporter:
It looks like Drupal, maybe metatag, generates a canonical link already. 
image 

This is set on the directory listing and on individual profiles unless you do a full page reload.

We allow webmasters to set another URL as the canonical source for profiles but we're not rendering it right now because it was just adding another one to the HTML `head`. Since the directory doesn't do a page reload when navigating to a profile, Drupal/metatag's canonical link is wrong. I think we'd need to alter it in JS to handle it correctly.

However, I'm unsure about the use case for this. Is https://profiles.uiowa.edu the canonical source of all the profiles? Should we remove this config option?

joewhitsitt:
I don't have an answer to your question but as it relates to Drupal, you could alter what gets spits out based on the route via `hook_metatag_metatags_view_alter()`

richardbporter:
The breadcrumbs are incorrect on a full page load of a profile (slug). I'm working on adjusting that.

docroot/modules/custom/uiowa_profiles/src/Form/SettingsForm.php
```diff
@@ -226,6 +232,10 @@ public function buildForm(array $form, FormStateInterface $form_state) {
             ],
             '#attributes' => [
               'data-directory-index' => $key,
+              'id' => 'delete-profiles-instance-' . $key,
```
richardbporter: This is causing the form to submit. I think we need to let Drupal set the ID as it includes a hash, I think to prevent CSRF.

GaryRidgway: Good catch, fixed!

richardbporter:
>Update webmaster permissions everywhere.

I think I goofed this up. Webmaster perms should be split out in this feature split and not updated everywhere else since this won't be turned on by default.

richardbporter (approved):

richardbporter:
Oops. Thought I was approving just my changes I requested. I'd like to take another look at this and might be good for someone other than us to look at it as well.

cory-skeers:
Pulling down to take a look and test

cory-skeers:
Two related things with the configuration form:

- Can't delete a directory with invalid inputs. For instance, add a new directory, change your mind, and try to delete it--it won't because of the required fields. Workaround is to throw some text in, and then delete
- If there are errors on a directory other than the one in focus, it is not apparent why you cannot save the configuration. The form errors are set on the field of the "hidden" directory

GaryRidgway:
> * Can't delete a directory with invalid inputs. For instance, add a new directory, change your mind, and try to delete it--it won't because of the required fields. Workaround is to throw some text in, and then delete

This might be fixable by making the fields not required? IDK how I feel about that though. any thoughts?
> * If there are errors on a directory other than the one in focus, it is not apparent why you cannot save the configuration. The form errors are set on the field of the "hidden" directory

This one I might be able to make work. Let me see if I can work any magic

richardbporter:
>This might be fixable by making the fields not required? IDK how I feel about that though. any thoughts?

I think we should leave them required. I'll look into options for that one.

richardbporter:
Seeing this error in console log on settings form: 
```
uiowa-profiles-settings-form.js?qxbzua:18 Uncaught TypeError: Failed to execute 'observe' on 'MutationObserver': parameter 1 is not of type 'Node'.
    at uiowa-profiles-settings-form.js?qxbzua:18
```

Making those fields not required leads to other validation errors. Not sure what other options we have there.

GaryRidgway:
> * If there are errors on a directory other than the one in focus, it is not apparent why you cannot save the configuration. The form errors are set on the field of the "hidden" directory

@cory-skeers  this has been fixed. It did however lead me to other issues in the form, upon save. If validation is done with tooltips, as is done with the page size, it cannot be focused like the links can. Perhaps this, having the links in the very in your face error message, is good enough for now?

cory-skeers:
@GaryRidgway I think the tooltips are generally preferred, but if it's between having the tooltips or being able to bring focus to the proper tab, I think being able to focus on the tab is worth the tradeoff. I'd be open to others' thoughts on it though.

Is it only set that way on the Delete button and not the Save Configuration button, or is that difference something on my local setup?

GaryRidgway:
It should work on save as well, for example if you have the same path in multiple instances it will error on submission.

I am unsure on how to get it to focus on tooltip though, I haven't tried to hook in to that before to do anything with it before so BLECH

cory-skeers:
> Making those fields not required leads to other validation errors. Not sure what other options we have there.

I want to say there should be a way to limit validation errors if the button press is Delete and not Save, but I'm not sure that's still an option. I think in D7 it was a `#limit_validation` attribute or something similar, but I don't see it as an option for D9.

cory-skeers:
> It should work on save as well, for example if you have the same path in multiple instances it will error on submission

I think I have something wonky going on then, as I'm also now able to save multiple instances with the same path.

GaryRidgway:
> > It should work on save as well, for example if you have the same path in multiple instances it will error on submission
> 
> I think I have something wonky going on then, as I'm also now able to save multiple instances with the same path.

That is confusing. I just tested it and it did not allow me. what instance do you see when you hit the saved path?

cory-skeers:
> > > It should work on save as well, for example if you have the same path in multiple instances it will error on submission
> > 
> > 
> > I think I have something wonky going on then, as I'm also now able to save multiple instances with the same path.
> 
> That is confusing. I just tested it and it did not allow me. what instance do you see when you hit the saved path?

The first instance that was saved that way.

It looks like it's that I'm entering the path without the first `/`, and so it's seeing it as unique even though the path is already taken. Entering the path with the slash rightly leads to an error.

GaryRidgway:
> > > > It should work on save as well, for example if you have the same path in multiple instances it will error on submission
> > > 
> > > 
> > > I think I have something wonky going on then, as I'm also now able to save multiple instances with the same path.
> > 
> > 
> > That is confusing. I just tested it and it did not allow me. what instance do you see when you hit the saved path?
> 
> The first instance that was saved that way.
> 
> It looks like it's that I'm entering the path without the first `/`, and so it's seeing it as unique even though the path is already taken. Entering the path with the slash rightly leads to an error.

Ah okay, perhaps we should be assuming the slash and adding one to the beginning of any entry that doesn't have one?

richardbporter:
Setting `'#limit_validation_errors' => [],` on the delete buttons gets rid of the error but then deletes all the directories for some reason... 🤦🏻

GaryRidgway:
> Setting `'#limit_validation_errors' => [],` on the delete buttons gets rid of the error but then deletes all the directories for some reason... 🤦🏻

that is confusing and nonsensical!

GaryRidgway:
I would assert that if we can get the path validation to work correctly, then that should be enough with the newly added error focusing. In the future [we have a path forward to validate API keys in a similar way](https://git.uiowa.edu/java/profiles/-/issues/301#note_93727) so the tooltip's days are numbered for that field, and the `Page size` field already has a description describing a min and max.

cory-skeers:
> Ah okay, perhaps we should be assuming the slash and adding one to the beginning of any entry that doesn't have one?

I think that makes sense. The Sitenow People and Sitenow Articles configuration forms save and display the path without the slash, so some users may be used to entering paths that way.

GaryRidgway:
> > Ah okay, perhaps we should be assuming the slash and adding one to the beginning of any entry that doesn't have one?
> 
> I think that makes sense. The Sitenow People and Sitenow Articles configuration forms save and display the path without the slash, so some users may be used to entering paths that way.

hmmm, is it better to REMOVE the leading slash then instead of adding one?

cory-skeers:
> > > Ah okay, perhaps we should be assuming the slash and adding one to the beginning of any entry that doesn't have one?
> > 
> > 
> > I think that makes sense. The Sitenow People and Sitenow Articles configuration forms save and display the path without the slash, so some users may be used to entering paths that way.
> 
> hmmm, is it better to REMOVE the leading slash then instead of adding one?

Good point--probably the former 😜

GaryRidgway:
@cory-skeers i used the path cleaning library we used everywhere else, thanks to @richardbporter for the help on that one. Should be recognizing them as same paths now.

cory-skeers:
> @cory-skeers i used the path cleaning library we used everywhere else, thanks to @richardbporter for the help on that one. Should be recognizing them as same paths now.

Awesome--checked it and it's working great.

richardbporter:
I'm out of ideas on the Delete validation. 

Having those fields be required actually prevents adding a new directory when the previously-new directory does not have values yet.

GaryRidgway:
> I'm out of ideas on the Delete validation.
> 
> Having those fields be required actually prevents adding a new directory when the previously-new directory does not have values yet.

We could fill the fields with dummy values on submit so they are not empty? just slap a `1` in all fields and then boom, it validates.


Otherwise I think this may be a slice we come back to. I say we have another enhancements slice later because trying to make sure this works now may be too much to bite off.

richardbporter:
I'd be ok following up on that if it becomes really problematic for people, unless someone has another idea.

cory-skeers:
> I'd be ok following up on that if it becomes really problematic for people, unless someone has another idea.

That sounds good to me. In the meantime, there is also the workaround of manually doing what Alan suggested:
 > just slap a 1 in all fields and then boom, it validates.

cory-skeers:
Other than the delete validation that has a workaround and can be addressed later, everything appears to be working as expected, despite efforts to try and break it.

I think it looks good to go! 💯

PR #3645, opened 2021-06-11
Add operator option for person type status to people list block view

### LINKED ISSUE #3642: Add ability to select people who are NOT of a type status (e.g. NOT former faculty)
# Story

As a content editor I want to be able to filter people lists by people who are NOT of a particular type status (e.g. NOT former faculty) so that I can create lists that exclude these people.

As part of #3587 we introduced filtered based on person type and person type status. However, those updates only allow people to create lists of people that are formerly of a type, not to list people of a type excluding former members of that group.

To be able to create lists of current faculty or current students, we need this ability.

# Proposed solution

Adapt code from the patch in https://www.drupal.org/project/ctools/issues/2657060 to modify the display, validation, and submit handling of the filters in the block form. Probably also it would be a good idea to adjust the select option labels for the person type status operator field (via the process function) so that they make more sense in context.

### LINKED ISSUE #3696: Adjust/remove the frontend "Person Types:" label
I was migrating some staff/person pages from our older sites and I noticed that this label "Person Types:" seemed a bit odd:

image

Could we rethink the wording of "Person Types:" on the frontend or possibly prevent this label/field from being rendered? I know both possibilities were discussed at the web community meeting.

pyrello:
Resolves #3642 
Resolves #3696

# How to test

- `blt ds --site sandbox.uiowa.edu`
- `drush @sandbox.local uli node/1/layout`
- Using the stories below, add a person list block and test out different scenarios.

## Stories

* As an Admin user, I am able to create a person block listing people who are current "Staff".
* As an Admin user, I am able to create a person block listing people who are former "Staff".
* As a non-Admin user, I am not able to access the person type status filter and the AJAX update doesn't run.


pyrello:
This seems to work, but I think that the UI is confusing. I'm also a bit concerned about how much we are relying in code on the filter machine names that are configurable through the view. I don't immediately have great ideas for how to handle this.

I'm wondering if, instead of trying to address the above issues now, we just limit access to this filter to only admins. This would allow me to go ahead with being able to create the views I need for AMCS and Human Toxicology, but would allow us to continue to work on how best to present these options to editors and site owners.

cory-skeers:
> I'm wondering if, instead of trying to address the above issues now, we just limit access to this filter to only admins. This would allow me to go ahead with being able to create the views I need for AMCS and Human Toxicology, but would allow us to continue to work on how best to present these options to editors and site owners.

I could get behind that plan. We have some steps forward, and I believe the AMCS/Human Toxicology are the primary drives for needing this implemented sooner rather than later. Beyond getting it functional as is, you've described where I stopped working on it until you got back, hoping you'd already had plans in mind for moving it forward.

pyrello:
I think this is working.

dnkemper:
I can test this!

dnkemper:
When I add the current staff vs the former staff, it shows the same people... does that matter?
Screen Shot 2021-06-23 at 3 20 41 PM

cory-skeers:
@dnkemper what are the form settings for both of those?

dnkemper:
@cory-skeers The 'current' is a one of any staff (makes sense) but the second, is a one of former staff. So shouldn't we only see the 'former staff' person?

cory-skeers:
It seems like the person type status value isn't getting set correctly now, so whatever the form gets set to, it's either all of any staff or none of any staff. Not sure what the issue is.

dnkemper (approved):

pyrello:
> It seems like the person type status value isn't getting set correctly now, so whatever the form gets set to, it's either all of any staff or none of any staff. Not sure what the issue is.

@cory-skeers Is this comment based on testing or @dnkemper's screenshot above?

Issue #3657, opened 2021-06-15
Collection form UX regressions


cory-skeers:
# Problem

Parts of previous updates to the collections LB form have regressed.

- View mode appears above the headline field collection, separate from Accordion Type (they should appear together, below the headline collection)
- States on the Accordion Type field is not currently working (ie it is available even if a non-accordion view mode is selected)
- (Minor) Dynamic help text modifications are no longer being applied.

![image](https://user-images.githubusercontent.com/49074566/122123317-78101900-cdf3-11eb-80d1-9ff20133d8a0.png)

# Steps to reproduce

Add or edit a collection. Choose Descriptive List in View Mode. See that things are out of order, and Accordion Type is visible and doesn't make sense.

# Proposed solution

Investigate code around https://github.com/uiowa/uiowa/blob/11e2828b51d310fc31e6d2c225ef3ae63527c874/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L724-L751

PR #3658, opened 2021-06-15
Feature: banner backgrounds and collections hotfix

### LINKED ISSUE #3660: Banner background image/color options are broken
# Problem

Banner background image/color options are broken from this commit: https://github.com/uiowa/uiowa/commit/2b3f45603e823ef06e5ad132c8cd5dd3b58e791b. 

Screen Shot 2021-06-16 at 8 59 57 AM

# Steps to reproduce
Try adding a banner on prod. 

# Proposed solution
Fix the issue with the noted commit.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3660. 

# How to test
- Try adding a new banner and verify that the Background select appears:

Screen Shot 2021-06-16 at 9 39 25 AM

- Try adding a collection and verify that the view mode sits above the collection items:

Screen Shot 2021-06-16 at 9 40 07 AM

- Create a people block and use the person type and person type status filters to person list block and verify that they aren't broken.


cory-skeers:
Pulled this, and it looks like it resolves #3657 as well (haven't dived into what changes have been made or might still need to be made, but in its current state the collections form works like a charm with it).

bspeare:
@cory-skeers any thoughts on how we should fix this part?   This is the line that was causing the form problems: https://github.com/uiowa/uiowa/pull/3658/commits/7aadae53fc46c93e34707132bc1960f478909e4e#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4cL147.

bspeare:
Here's the commit where that check was added: https://github.com/uiowa/uiowa/pull/3587/commits/2b3f45603e823ef06e5ad132c8cd5dd3b58e791b.

cory-skeers:
> @cory-skeers any thoughts on how we should fix this part? This is the line that was causing the form problems: [7aadae5#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4cL147](https://github.com/uiowa/uiowa/commit/7aadae53fc46c93e34707132bc1960f478909e4e#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4cL147).

It looks like for collections, at least, it has the `$form['settings']` but is failing because it doesn't have a `$form['settings']['override']` set. Can we just change it to `if (isset($form['settings'])) {`, or do we need to check for `override` for other blocks?

cory-skeers:
@bspeare I'm thinking at least as a hotfix, take the `isset($form['settings']['override'])` check out for now--we may need to add it in later for specific blocks, or check for something else, but as a hotfix, it should get things functional again without explicitly breaking the people block if that gets handled in `sitenow_people`. I think that's the main block that has/needs an `override` key.

Edit: And after, maybe we can get @pyrello's input on the change, and help with accomplishing what that commit had originally intended.

GaryRidgway:
pulling to test!

GaryRidgway:
All of this is working well! I did notice that while I was checking if the filters on the `people list` block that if I checked `faculty` on a person as a `person type`, that `former faculty` did not show up. I had to save the node and re-edit the person to then find a new checkbox field titled `former faculty`. is this something we should deal with here? or make a separate issue?

bspeare:
@GaryRidgway I noticed the same thing and tested it on prod and had the same result.   I'm not sure if that is the intended functionality - @cory-skeers @richardbporter @mark-bennett-uiowa are you familiar with those checkboxes?

briand44:
>I noticed the same thing and tested it on prod and had the same result.

Then I would create a separate issue if it is not desired functionality and not deal with it here.

GaryRidgway (approved):
Works for everything listed!

cory-skeers:
> I did notice that while I was checking if the filters on the `people list` block that if I checked `faculty` on a person as a `person type`, that `former faculty` did not show up. I had to save the node and re-edit the person to then find a new checkbox field titled `former faculty`. is this something we should deal with here? or make a separate issue?

This is when making a new person, right? Not the people list block filters? That should be separate from this, but also not the intended functionality...

GaryRidgway:
it was during the editing of a person, not on the filters, just thought I would surface it!

PR #3659, opened 2021-06-16
Escape panopto user input before render.

richardbporter:
No functionality changes here.


cory-skeers (approved):
Reviewed, not tested, and should only have the desired impact 👍

PR #3665, opened 2021-06-16
Person Type Status fix

### LINKED ISSUE #3662: "Person type status" not showing on initial "person type" selection
# Story

As a site editor I want to be able to assign the `Former faculty` person type status to a `person` without having to assign them `Faculty`, save the page, click edit again, and then check the `Former faculty` box.

Currently, when you check `faculty` on a person as a `person type`, `former faculty` does not show up.You have to save the node and re-edit the person to then find a new checkbox field titled `former faculty`.

# Proposed solution

Possibly use states to show and hide this field?

cory-skeers:
Resolves #3662

# How to test

As a site editor creating a new person, I should see all available person type status options (eg Former Faculty) based on my currently selected person type(s), both when creating a new person and when editing an existing person.

### DIFF
--- a/docroot/modules/custom/sitenow_people/sitenow_people.module
+++ b/docroot/modules/custom/sitenow_people/sitenow_people.module
@@ -269,7 +269,7 @@ function _person_types_callback(array &$form, FormStateInterface $form_state) {
   //   the triggering element is the person types filter? Currently
   //   we're skipping if we're on a node form, but scoping could be
   //   more specific.
-  if ($form['#form_id'] === 'node_person_form') {
+  if (in_array($form['#form_id'], ['node_person_form', 'node_person_edit_form'])) {
     return $form;
   }
   $type = $form_state->getValue([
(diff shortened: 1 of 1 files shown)


bspeare:
pulling down to test

bspeare:
@cory-skeers What if you want to change an existing person to from faculty to former staff?  Should that functionality be working as well or only on initial node creation?

cory-skeers:
@bspeare Good catch! Made a small update so that it should work the same way on both new nodes and editing existing nodes.

richardbporter:
I wonder if we can work an unrelated fix in to this. The `@cs.prod` site has a config diff right now that will be blown away on the next release. I'm trying to figure out how to get it in a default state. The site has the additional_person_types split enabled. We added `config_split.config_split.additional_person_types:status` to config_ignore.settings in prod but that didn't fix it. That line needs to be added regardless.

richardbporter:
> I wonder if we can work an unrelated fix in to this. The `@cs.prod` site has a config diff right now that will be blown away on the next release. I'm trying to figure out how to get it in a default state. The site has the additional_person_types split enabled. We added `config_split.config_split.additional_person_types:status` to config_ignore.settings in prod but that didn't fix it. That line needs to be added regardless.

Nevermind. I'll create a separate PR for that.

bspeare (approved):
The multiple person types now appear when creating a new person or editing and existing person.  Thanks!

PR #3687, opened 2021-06-18
Update uiowa_auth and samlauth.

richardbporter:
Related to https://github.com/uiowa/uiowa_auth/pull/50

### Testing Setup
This has been merged into develop and deployed. You can use a site on dev like https://default.dev.drupal.uiowa.edu and a site with custom roles/auth config like https://brand.dev.drupal.uiowa.edu/. You can accomplish the below stories by logging in as user 1 (`drush @default.dev uli`) in one browser and then using another browser to log in with your HawkID. The stories will require you to delete your account on the site you are testing with and remove/adjust the role mapping config in Configuration -> HawkID Authentication.

### Stories
- As a user logging in to a site with a role mapping, my account is created and granted the appropriate role(s) upon log in.
  - As a user who has already logged in to a site via a role mapping, my role(s) is revoked if the role mapping changes to an attribute or value that I do not have.
- As a user logging in to a site without a role mapping or an existing account, I receive an error during log in and my account is not created.
- As a user logging in to a site where a manually created account exists with my HawkID, I can log in successfully and my email is synced correctly.
- As a user logging in to a site where a manually created account exists with my HawkID but has been blocked, I cannot log in.

### Todo
- [x] Merge https://github.com/uiowa/uiowa_auth/pull/50, create a new release of that repo and update this PR to use the new tag.
- [x] Update the [secrets.settings.php](https://docs.acquia.com/resource/secrets/#secrets-settings-php-file) config override files on each application environment to use the new config at the same time as code deployments. Change `$config['samlauth.authentication']['idp_x509_certificate']` to `$config['samlauth.authentication']['idp_certs']` and make the value an array. Already done on uiowa/01/02 dev envs. This might require an off-hours deployment and/or ITS Alert.


config/default/samlauth.authentication.yml
```diff
@@ -1,32 +1,64 @@
...
 idp_single_log_out_service: ''
 idp_change_password_service: 'https://apps.its.uiowa.edu/pwtool/change'
+idp_cert_type: single
 idp_x509_certificate: ''
+idp_x509_certificate_multi: ''
 unique_id_attribute: 'urn:oid:1.3.6.1.4.1.6159.1.41'
 map_users: true
+map_users_name: true
+map_users_mail: true
+map_users_roles:
```
richardbporter, line 44: This is not great for splits. Sites with role splits will either need to split this config out or we can dynamically override it in uiowa_auth.

richardbporter: Done

joewhitsitt:
testing

joewhitsitt (approved):
I have run through the stories successfully using default and brand DEV

richardbporter:
I was able to log in after redeploying to dev but I'm a bit confused as to why that was. I feel like it _shouldn't_ work until the secrets file is updated. I think a planned maintenance ITS alert would be a good idea.

joewhitsitt:
we'd like to wait on this

richardbporter:
I just set both config overrides in uiowa.test secrets file and was able to log in. I think we can do that in advance and then schedule the deployment through data.its very early and then remove the old config once complete.

richardbporter:
I've updated test/prod env overrides. This should be merged tomorrow afternoon and a new release created afterwards. The new release tag should be scheduled for deployment using dataits.

PR #3692, opened 2021-06-21
Fixed min height for text block

### LINKED ISSUE #3653: Text area blocks lose the contextual edit link if they are empty
# Problem
Text area blocks lose the contextual edit link if they are empty

# Steps to reproduce
Try adding a text area with a few line breaks to a layout and then try to edit the block after saving. 

# Proposed solution
- In `admin.scss` add a `min-height: 10px` or greater rule for text area blocks.

dnkemper:
Closes #3653 

# How to test

### DIFF
--- a/docroot/themes/custom/uids_base/scss/admin.scss
+++ b/docroot/themes/custom/uids_base/scss/admin.scss
@@ -1605,6 +1605,11 @@ only screen and (min-resolution: 1.5dppx) {
   border-color: transparent;
 }
 
+// Edit layout for textblock not displaying correctly
+.block-inline-blockuiowa-text-area {
+  min-height: 10px;
+}
+
 // Layouts not covered by lb_direct_add module
 .layout--page--left-sidebar .layout__region--main .layout-builder__add-block.open .layout-builder__direct-add__list > .links {
   @include breakpoint(sm) {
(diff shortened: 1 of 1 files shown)


docroot/themes/custom/uids_base/scss/admin.scss
```diff
@@ -1605,6 +1605,11 @@ only screen and (min-resolution: 1.5dppx) {
   border-color: transparent;
 }
 
+// Edit layout for textblock not displaying correctly
+.block-inline-blockuiowa-text-area {
+  min-height: 10px;
```
cory-skeers: Could we increase the min-height a bit? The configure button is reachable, but can sometimes overlap other parts where it gets less clear what it's attached to.

image

~40px gives enough space to fully contain the button. Not sure if that's so much space that might make the preview a bit misleading for what is actually in the text block, though.

image

dnkemper: Woof! Thank you! IMO the 40px works great!

cory-skeers (approved):
Looks good and works like a charm!

dnkemper:
@cory-skeers marvelous! Thank you so much!

Issue #3585, opened 2021-06-01
Some sites are randomly missing layout options and content.


bspeare:
# Problem
Some sites report that their layout options are missing a two-column layout option and content is missing.   
- Content is also missing on the page that was placed in a two-column layout 
- The option to create a two-column layout is missing from the add content page 

 # two-col layout history:
- https://tippie.prod.drupal.uiowa.edu/testing-page-parent/testing-page-3  - 5/27/21
- https://jobsexecutive.prod.drupal.uiowa.edu - 5/27/21
  - https://github.com/uiowa/uiowa/compare/3.6.18...3.7.0
- Transportation - 4/15/21
  - https://github.com/uiowa/uiowa/compare/3.6.4...3.6.5
- Pharmacy - 2/3/21
  -  https://github.com/uiowa/uiowa/compare/3.5.1...3.5.2
- Mark B CLAS 11:44am June 10th
  - https://socialwork.uiowa.edu/locations/iowa-city
  - https://github.com/uiowa/uiowa/compare/3.7.4...3.7.5
- Brandon E 12:06 Aug 16th - https://iowaweb.slack.com/archives/C0164FNN1PV/p1629133587288100
  - https://gpsg.prod.drupal.uiowa.edu
- https://archaeology.uiowa.edu 11/06/2023 at around 1:30pm
- https://medicinecommunity.prod.drupal.uiowa.edu/ 11/13/2023
- https://ocrc.prod.drupal.uiowa.edu/ 3/19/2025 - this was a new site provision that went out in a release the previous day and was working initially. There were no other commits in the release.
- https://lllc.prod.drupal.uiowa.edu/  - 06/12/2025 - https://github.com/uiowa/uiowa/compare/3.31.1...3.31.2
- https://spanish-portuguese.uiowa.edu/ - 06/16/2025 - https://github.com/uiowa/uiowa/compare/3.31.2...3.31.3

# Known Issue
https://sitenow.uiowa.edu/known-issues#uiowa-issue-3585

# Workaround

- Clear site and varnish cache

bspeare:
We are defining our layouts here: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.layouts.yml.

richardbporter:
It might be good to create an AC ticket explaining this issue to see if they can help us or provide information.

richardbporter:
[...]

PR #3714, opened 2021-06-25
Adjust uids_base layout definitions based on docs.

richardbporter:
This changes some uids_base layout path and template values based on the [docs](https://www.drupal.org/docs/drupal-apis/layout-api/how-to-register-layouts#full-annotation-reference). Might be related to #3585, might not. :)

# User stories
As a V3 site owner, I can add new sections to a page in LB using 1 column, 2 column, 3 column, 4 column, and page with sidebar layout.
As a V3 site owner, I can see existing content in the above layouts.
As a V2 site owner, my content continues to display as it was.


pyrello:
It seems like maybe the `path` option isn't actually necessary for our purposes.

bspeare:
Pulling down to test

bspeare (approved):
Great fix! Layouts continue to operate smoothly.

PR #3715, opened 2021-06-25
Headline size mismatch w/render cache

### LINKED ISSUE #3558: Node teaser heading size is set incorrectly on first cached render
# Problem
The node teaser heading size is set incorrectly on first cached render for bundles that implement the headline field.

# Steps to reproduce
Add the Articles block to a page and enter a title and set the heading size to h3 for example. Inspect the LB edit page markup and see that child elements are set to h4 which is correct. Save the page and inspect the same markup and see that the child elements are set to h2 which is incorrect.

If you edit the block and change the heading size from h3 to h4 for example, then child elements will be set to h5.

I think this problem is due to the fact that the block render cache tags do not yet exist the first time it is added to a page. That means we aren't really adding anything to the entity cache tags the first time:

https://github.com/uiowa/uiowa/blob/fa510cd3d76680f68bf51a09d04ec60b4c7d9702/docroot/modules/custom/sitenow_articles/sitenow_articles.module#L507-L512

We use this code in multiple places - person, articles, featured content, etc. After saving the page and editing in LB again, the cache tags are populated with `block_content:244` for example. That cache tag does not exist the first time.

# Proposed solution
TBD.

cory-skeers:
Resolves #3558 

## Story

As an editor I can place articles, people, featured content and any duplicative combination of referenced content with different heading sizes and get the expected result the first time and going forward for both layout builder preview as well as page render.

<details>
<summary>original pr summary</summary>
First attempt to get something going. Idea was to catch unsaved blocks (which don't yet have an id), save them, and then manually add the proper cache tag to the results. Runs into a duplicate uuid issue which interrupts save and causes WSOD.

There's another possible issue with this approach if a single node appears in more than one block list, with different heading sizes in each. The teaser render may get cached from any one of the block instances, and that headline size will then be used for all block instances if the cache is hit, regardless of their individual settings (the render cache tag will also be invalidated upon updating any one of the various block instances, as all `block_content:#` tags should be added to the nodes).
</details>


joewhitsitt:
i have been meaning to look into this. could we create a custom cache tag for these that we then target? it will be unique to the block and not cause a WSOD and would differ per block. trying and failing to find an example where we did this. we might have ripped it out before merge

joewhitsitt:
I wanna say that something like the following prevents the WSOD and solves the first issue. The multiple on a page with different settings issue still remains however.

```
        // If the block is brand new, we need to save it before it will have an id.
        if ($block->isNew() && $block->save()) {
          // Manually construct a custom tag with the newly generated id,
          // because the block content tag isn't yet in our variables.
          $block->addCacheTags(['sitenow_articles_custom:' . $block->id()]);
        }
```

This gets added later in:
```
// Associate the block's cache tags with each entity in the view result so
      // that the appropriate render cache ID is cleared when the block changes.
      foreach ($view->result as $result_row) {
        $entity = $result_row->_entity;
        $entity->addCacheTags($variables['content']['#cache']['tags']);
      }
```
so the if check/add in this pr wouldn't be needed.

joewhitsitt:
So I have been testing with articles block, people listblock and featured content. i think the latest works for everything but featured content people when there is an existing people listblock already placed. even multiple instances of a person across different featured content blocks seems to work. trying to track down the difference here.

joewhitsitt:
So `view` is present when a people listblock is placed first and a featured content block (with a person) is placed second. So then in the teaser template the view heading size wins out.

```
{% set headline_level = heading_size ?: 'h3' %}
{% if view %}
  {% set headline_level = view.display_handler.getOption('heading_size') %}
{% endif %}
```

cory-skeers:
This is working great! Pulled it down for testing, and haven't been able to reproduce any of either the first-render or multiple-use issues.

It has my approval in spirit, but can't give it because I technically opened the PR.

pyrello (approved):
Approving based on @cory-skeers approval.

joewhitsitt:
thanks both!

PR #3716, opened 2021-06-25
International - Article Migration

### LINKED ISSUE #3375: Migrate news articles for International Programs
# Story

As an International Programs site owner, I would like articles migrated from the International Programs Drupal 7 site (https://international.uiowa.edu) to https://international.prod.drupal.uiowa.edu so that users will be able to view all of the past news articles.

# Additional Info

Need  to get additional info in regards to:
- when they hope to launch - end of August(ish)
- migrating tags - yes
- migrate author labels (multiple) to organization field
- creating programmatic redirects
- if we are migrating all articles or articles going back to a specified date - migrate all articles beginning January 2015
- how handling broken links

# Proposed solution

joewhitsitt:
Relates to #3375 

## Story

As a viewer of the International v3 website I can see articles dating back to the beginning of 2015 along with the images and tags associated with them.

## Test

Sync D7 site in DevDesktop (international.uiowa.edu on uiowa703).
git pull && composer install
blt ds --site=international.uiowa.edu
drush @international.local en international_migrate -y
drush @international.local uli /admin/config/system/sitenow-migrate
Complete the configuration form using the help text to configure for DevDesktop (db = international_uiowa_edu)
drush @international.local ms -- shows migration status; two migrations should appear, one for articles and one for article redirects
~drush @international.local mim --all -- runs both migrations~
drush @international.local mim international_articles --limit 50 (will run a batch of 50. Need to repeat until all 900 articles are migrated in order to test the full migration, but spot checks can be done prior). Can run drush @international.local mim --all when ready for the final run, and it will run the final articles as well as the redirects.

At the end of each import, a list of possible broken links will be output to the command line (nodes with /node links in them). For sanity's sake, you can temporarily remove the `$this->reportPossibleLinkBreaks(['node__body' => ['body_value']]);` line from the `postImport` method until the final run.

## Todo
- [ ] Deal with memory issue
- [x] Broken links report?
- [x] Author field (multi term) to organization string


joewhitsitt:
```
[warning] Memory usage is now 427.59 MB (84% of limit 512 MB), not enough reclaimed, starting new batch
 [notice] Processed 328 items (328 created, 0 updated, 0 failed, 0 ignored) - done with 'international_articles'
 [error]  Migration international_article_redirects did not meet the requirements. Missing migrations international_articles. requirements: international_articles.
```

cory-skeers:
https://gitlab.com/drupalspoons/migrate_tools/-/issues/92 This was one of the issues I was keeping an eye on before; doesn't look like there's been much active progress on it lately, but might be worth checking out.

I've previously tried specifying batch sizes, and limiting imports (`--limit=#`), but whatever main memory leak is causing the issue still exists. I think the underlying issue has to do with how the migrate module runs queries and stores that information--on subsequent runs with a highwater mark, memory usage starts to build up before it even gets to the new records.

## Workaround 1

For some migrations, it's enough to simply run the migration multiple times, and each time it'll get a little farther. Each subsequent run is slower than the last, though, and even that has the possibility of eventually hitting memory issues prior to importing any new content if the node count is high enough.

## Workaround(ish) 2
A not-so-robust workaround if you're migrating locally is to up the allocated memory. Doesn't solve the issue at all, but combined with multiple runs, can be enough to get it to complete. Takes forever, though, because even if it takes longer to error out, the migration will start to slow down as the memory usage builds up.

cory-skeers:
## Untested Workaround 3

https://www.mediacurrent.com/blog/memory-management-migrations-drupal-8/

They wrote a script that repeatedly calls a migration with a (low) set `--limit` in order to makeshift batching. I haven't tried it myself, but from my experience subsequent calls with or without a `--limit` will still progressively slow, as there's a large memory usage buildup as the migration works through processed entities. But that may be a route forward.

joewhitsitt:
Thanks for these @cory-skeers. Sorry to call you out earlier. I am getting no where fast with #\1. Not having run a migration in production is this just an issue locally or will this most likely be a problem in PROD as well?

bspeare:
Do we know which workaround we want to use here?  Or is this blocked until we can decide?  This will be an issue on PROD as well.

joewhitsitt:
Not sure how my memory limit locally differs from prod and haven't tried the third option yet. I don't think it is blocked. If I recall we just needed to start testing the work we had done and maybe provide a broken links report? There was a question about the author field on the issue as well.

cory-skeers:
Started looking into implementing something like https://www.drupal.org/project/drupal/issues/2688297#comment-13523145 into the `BaseNodeSource`. Not quite plug-and-play ready, so not to the point of knowing if it's a way forward or not.

GaryRidgway:
hey all, thought it might be useful to try and track the amount of nodes migrated each iteration when cory ran the migration on his machine. They are listed below:

158
190
182
134
105
65
64
3 (last remaining nodes)

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php
```diff
@@ -287,8 +290,9 @@ public function downloadFile($filename, $source_base_path, $drupal_file_director
    * @throws \Drupal\Core\Entity\EntityStorageException
    */
   public function createMediaEntity($fid, array $meta, $owner_id = 1) {
+    $file_manager = $this->entityTypeManager->getStorage('file');
     /** @var \Drupal\file\FileInterface $file */
-    $file = $this->entityTypeManager->getStorage('file')->load($fid);
+    $file = $file_manager->load($fid);
```
pyrello, line 295: We can avoid having to set `$file_manager` (and then later unset it) by wrapping loading the `$file` in a try/catch statement. Ditto the `$media_manager` variable later on.

pyrello: I see now that this is how it was working before, but without the try/catch block.

cory-skeers: Interesting! Didn't know that about the try/catch approach. I'll take a look at that when I get back to this.

cory-skeers:
After watching the memory usage, I believe the `file_save_data` is the (main) culprit causing the issue. Particularly with file-heavy pages, this line tends to be the point at which memory spikes and never comes back down. Pages without files see few-to-no spikes (and not as large), and similarly the surrounding code involved in processing files doesn't appear to take up significant memory.

https://github.com/uiowa/uiowa/blob/d8612f1cef1a3b6aab7d6fd2b88f890a4c9996d3/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php#L255

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php
pyrello: The try/catch block just accounts for the potential of an exception. Ideally we would code around all pieces of code that can throw exceptions with try/catch blocks (if there isn't already one in the method being called). In this case it allows us to chain the `load` method off of the `getStorage` method without having to worry about the fact that we might have gotten an exception from `getStorage`.

cory-skeers:
I believe the author field migration is working correctly.

For the memory issue, it is still lingering, but it is now possible to a) complete the migration for all 900 relevant nodes (the other 1409 that appear in migrate:status are older than 2015) and b) utilize a `--limit 50` parameter on repeated imports to avoid any memory crashes. There is a small time addition on subsequent runs as the migration still checks if each node has been migrated previously or not, but memory usage does not compound on subsequent runs.

Right now, post-import will generate a link report, but does not attempt to fix any links (even if they are pointed at potentially migrated articles). This can be used to see the scale and scope of potential broken links to make any decisions regarding automated updates.

joewhitsitt:
looks good here. going to take a look locally

joewhitsitt:
From what I have seen, the content is being imported as expected, potential broken links reported. The limit allowed me to get further than the initial time. I'll be honest I did not see it to the end. Can't approve as author of this PR but here is a 👍

pyrello (approved):
Approving based on Joe's approval.

PR #3729, opened 2021-06-30
fix for transportation headers

GaryRidgway:
transportation headers fix


joewhitsitt (approved):
i think this is worth pushing. fixes the videos for Alan and I locally

PR #3754, opened 2021-07-01
Update README.md

makurj:
Correcting https://github.com/uiowa/uiowa/blob/main/config/README.md#site-split

# How to test


pyrello (approved):
Good catch! Thanks for making the update.

PR #3768, opened 2021-07-07
Feature: I3757 superfish ipad

bspeare:
# How to test


quamsta (approved):
Looks good on the devices I tested on: iPad Pro 12.9, iPad simulator, Chrome touch simulator. Good, clean fix. Thanks Alan! 

Edit: Also just tested with an iPad Sixth Generation from one of our students, and it's fine there too.

PR #3790, opened 2021-07-13
Fix: Changed breakpoint for container top padding

### LINKED ISSUE #3720: Adjust breakpoint on `layout__container` for padding consistency
# Problem
The layout container padding that is added or removed through the "Spacing" options on layout builder sections is not consistent for 

# Steps to reproduce
Add `Add extra top spacing` to a layout section and observe how the spacing gets smaller if you drop below the `container` breakpoint. 

# Proposed solution
- Change the spacing to be increased at a lower breakpoint so that there is not an unexpected spacing shift if the browser width drops below the `container` breakpoint. 
- Update https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss#L38 to `sm` intead of `container`.

pyrello:
Resolves #3720 

# How to test
* As a content editor, when I apply the "Add extra top spacing" style to a Layout Builder section, I don't see a change in padding size when I adjust the size of the viewport past the normal width container breakpoint.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/global.scss
+++ b/docroot/themes/custom/uids_base/scss/global.scss
@@ -35,7 +35,7 @@ $imgpath: '../../uids/assets/images';
   // This will be overridden by a subsequent style for 1+n containers.
   padding-top: $mobile-width-gutter;
 
-  @include breakpoint(container) {
+  @include breakpoint(sm) {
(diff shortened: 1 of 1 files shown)


cory-skeers:
Pulling down to test

cory-skeers (approved):
👍

Issue #3804, opened 2021-07-14
Add exposed filters to views block display


pyrello:
# Story

As a content editor, I want to be able to expose filters on my view blocks.

# Proposed solution

- [x] ~Help get MR at https://www.drupal.org/project/ctools/issues/2657060 to a fully working state.~ MR got merged and is part of the more recent version of `ctools`.
- [ ] Update to most recent version of `ctools` module.
- [x] Refactor code that transform the people type selector based on updates.
- [x] Add patch to account for issue with multiple exposed forms in blocks for the same views display, possibly the one at this issue: https://www.drupal.org/project/drupal/issues/3163299
- [x] Account for pager and offset options showing up in exposed filter form.
- [ ] Refactor code to work with updated version of `ctools` module.
- [ ] Update hook to convert existing configuration to the new format.

<details>
<summary>Original proposed solution</summary>
Modify the `Drupal\layout_builder_custom\Plugin\Display\ListBlock` class in the following ways:
- For each filter that can be set on the block settings form, add a checkbox to the form to allow the user to expose the filter. Name the checkboxes fields `$filter_id . '_expose`.
- Add `'#states'` to each filter field to disable the field if the corresponding checkbox is checked.
- Adjust the logic in the `blockSubmit()` method to check for an instances where the above fields are set to TRUE and set a new option value on the block called `expose_form` to TRUE if any of them are. Here is a quick example I came up with:
  ```
  $expose_form = FALSE;
  $exposed_inputs = $form_state->getValue([
    'override',
    'exposed_filters',
  ]);
  $exposed_filter_values = [];
  foreach ($exposed_inputs as $input) {
    // if input ends with _expose, $expose_form = TRUE;
    // else add to $input to $exposed_filter_values.
  }
  $block->setConfigurationValue('exposed_filter_values', $exposed_filter_values);
  $block->setConfigurationValue('expose_form', $expose_form);
  ```
- In the `preBlockBuild()` method, check the `expose_form` configuration value from the block and set an option on the display handler (`$this->view->display_handler->setOption('expose_form', TRUE)`) if this value is set to TRUE.
- Adjust the `displaysExposed()` method to check the display handler option and return `TRUE` if any filters are exposed.
[...]

PR #3815, opened 2021-07-15
Add research areas filter to people list block settings

### LINKED ISSUE #3812: Add research areas to people list block filters
# Story

As an editor, I can filter people list blocks by research area.

# Proposed solution

Update the people list block views settings to include research areas as a filter and include it with the filters that can be shown in block settings. Re-export config.

### LINKED ISSUE #3818: ListBlock class doesn't respect custom filter identifier
# Problem
When building a view, there is an option with exposed filters to choose a custom identifier for the filter (which gets used as the query key when the form is submitted). Our custom ListBlock display plugin does not respect this functionality and builds the form incorrectly when a custom identifier is used.

Screen Shot 2021-07-16 at 7 30 10 AM

# Steps to reproduce
- `blt ds --site sandbox.uiowa.edu`
- `drush @sandbox.local admin/structure/views/view/people_list_block`
- Customize the filter identifier for tags. It's default is `field_tags_target_id`
- Visit https://default.local.drupal.uiowa.edu/node/1/layout
- In PHPStorm, set a breakpoint here: https://github.com/uiowa/uiowa/blob/66c4ecf126569c4530cd01686d3919d90071fae7/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L225
- Turn on the "Start listening for connections..." option.
- In Layout Builder, add a People block
- In PHPStorms debug console, inspect `$form['settings']['overrides']['exposed_filters']`.
- You should see an array entry under `field_tags_target_id` AND one under your custom identifier.
- You'll also notice once you finish script execution and the block form finishes loading that the Tags field does not have a label (it is the first field in the Filters section).

# Proposed solution
Fix is in #3815

pyrello:
* Resolves #3812
* Resolves #3818

# How to test
* Review bug reproduction steps in #3818
* `blt ds --site sandbox.uiowa.edu`
* `drush @sandbox.local uli node/51/layout`
* To test that bug is fixed, follow similar steps, setting a breakpoint at line 229 in the ListBlock class. Observe that `$form['settings']['overrides']['exposed_filters']` doesn't contain an entry under the key `field_person_research_areas_target_id`. Also observe after script execution completes and the block form loads that the Research Areas filter field has a title showing.

## Story
As a content editor, I can set choose research areas as a filter to display people when adding or editing a people list block.


dnkemper:
I can test this since related to #3804

dnkemper (approved):

joewhitsitt:
Does this affect existing blocks or does it work as-is until next block edit?

Update: Tested to see if the new filter affects existing. Does not seem to.

PR #3821, opened 2021-07-16
Remove uiowa_people part 3

joewhitsitt:
Runs previous update hooks to remove uiowa_people from the sites.

careers.stage is an example of where this is still an issue. https://github.com/orgs/uiowa/projects/1#card-65146215

## To Test

- `git checkout main`
- `composer install && blt frontend`
- change the drush -> remote to `careers.test`
- `blt ds --site=careers.uiowa.edu`
This should fail!

- `git checkout uiowa-people-removal-part-3`
- `blt ds --site=careers.uiowa.edu`
This should run smoothly!


richardbporter:
> change the drush -> remote to careers.test
> blt ds --site=careers.uiowa.edu
This should fail!

This can be done in one command using the `--define` option.

 `blt sync --site careers.uiowa.edu --define drush.aliases.remote=careers.test`

richardbporter:
This works but I'm confused why it needs to be run again when it was run previously in `sitenow_update_9011`: https://github.com/uiowa/uiowa/blob/854d509802ef86a1052ff2ce8cb104125e6e306e/docroot/profiles/custom/sitenow/sitenow.install#L3107-L3122

So this would be the 3rd time running for all sites, right?

joewhitsitt:
Yeah, I didn't do any real fact finding to determine why this is needed again. Fingers are just crossed that this would be the last time.

richardbporter:
Should we limit to only careers or do you think it needs to run everywhere again?

joewhitsitt:
Could limit it to careers but maybe there is another one out there

richardbporter (approved):
I added a count message just so we could see if there were other instances of this. That work?

Issue #3661, opened 2021-06-16
People block pager count is broken when set to 1


bspeare:
# Problem
The people list block pager displays an inaccurate count when set to `1`. 

# Steps to reproduce
Check out the top two blocks on https://sandbox.prod.drupal.uiowa.edu/people-directory. 

# Proposed solution
- ?

richardbporter:
Investigate opening bug in ctools_views project to potentially address there.

joewhitsitt:
There might be pager id issue here. We got a separate ticket saying a pager wasn't even displaying. The difference is there is another block where the pager is not set.

Might be worth checking into what the pager id is set to and if it conflicts with another block on the page.

joewhitsitt:
Pager ID takes integer, and ideally we would supply something unique which our ListBlock submit doesn't seem to have access to.

Drupal 10 plans to make this dynamic https://www.drupal.org/project/drupal/issues/3225987

A ListBlock specific hack would be to set a block config value `rand(10, 1000)` update the pager id on preBuild. Still a collision possibility but less so. Views would have to be re-saved to get the unique id but this ensures the same id is used for submit/load/ajax. Unfortunately, `Html::getUniqueId` itself won't work for this because we need an integer.

Having a unique pager id seems to resolve this issue as well as the missing pager issue that was recently reported.

pyrello:
I may have accidentally stumbled on to a fix for this. Going to test today.

pyrello:
> I may have accidentally stumbled on to a fix for this. Going to test today.

This did not pan out as I had hoped. I had just not constructed the right case for testing the issue.

pyrello:
I worked on trying to adapt a patch using the code in this comment: https://www.drupal.org/project/drupal/issues/3225987#comment-14177968

This caused the pager to render correctly on first load, but broke the behavior when first clicking a page link. This would cause the first item to reload. Subsequent clicks after that worked correctly.

I like the approach of doing this dynamically and not getting into having to generate and store a value. It seems that some additional tinkering with this patch could probably get us there: https://git.drupalcode.org/project/drupal/-/merge_requests/1439.diff

dnkemper:
NOTE:

Load page
[...]

PR #3829, opened 2021-07-19
Add exposed filters to views block display

### LINKED ISSUE #3804: Add exposed filters to views block display

cory-skeers:
Resolves #3804 

# How to test

* Check that this doesn't break v2 person paragraph functionality. Test that it doesn't affect any existing view blocks with exposed filters.

## Stories
- As a non-administrator editing a people block, I am not able to filter by person type status.
- As an administrator, when I select a person type, the person type status options update to reflect on the corresponding options.


cory-skeers:
Update to current state for anyone who may jump in on this: Exposing filters functionally works. Could use some work on how the filter fields work (eg Person Type is a text field, not select list).

Main issue: The filters don't function. Hitting "Apply" appears to work, but the results do not change. May be related to https://www.drupal.org/project/drupal/issues/2605218 and any work that went into getting around that.

joewhitsitt:
taking a look

joewhitsitt:
oops. regression to getting the person type filter showing the options... duplicate filter on settings tray
Screen Shot 2021-07-21 at 2 12 26 PM

cory-skeers:
Update re: the AJAX issue--it appears the call gets made, but the display doesn't update. There are no console errors, but choosing an option that doesn't have any results (Research Areas is a good one for testing on sandbox) and refreshing the page will display a "no matches" Drupal message.

cory-skeers:
Also a note to remember, if anything needs to be updated in code: AJAX reloads do trigger `elementPreRender`, but not `preBlockBuild`.

cory-skeers:
Followed the ajax query through, and it is being formed correctly and returning correct results. So it does seem to be narrowed down to the display not updating with the returned results.

cory-skeers:
Apparently `preBlockBuild` is supposed to fire on AJAX reloads, and I'd had something going on where it wasn't.

Removed lines in that method that set exposed filters to match config values, which fixes the AJAX loads, but may have broken functionality elsewhere. Need to check on that.

cory-skeers:
To do: Update Person Type field on form to operate as a select list.

cory-skeers:
Person type and person type status fields are working as exposed forms; research areas and tags are not, and exposing them leads to "no results."

cory-skeers:
So, the Research Areas and/or Tags exposed filter fields seem to work properly if they are used along with either or both the Person Type and Person Type Status exposed filters. But they don't work on their own. "Don't work" here means returning no results regardless of the filter selections.

e.g. Exposing Research and Person Type will lead to a functioning filter form; Exposing only Research will not; Exposing only Person Type will.

There is currently extra stuff going on with Research/Tags to turn them into single select dropdowns when exposed, but removing this and leaving the multi autocomplete has the same issue.

joewhitsitt:
Taking another poke-see

joewhitsitt:
While trying to figure out:

>But they don't work on their own. "Don't work" here means returning no results regardless of the filter selections.

I set research on the block config to "history" and then made the filter exposed which on save, caused this illegal choice error. digging more. might be related?

Screen Shot 2021-07-27 at 11 15 32 AM

joewhitsitt:
Sorry, I am not getting anywhere with this. Just noticing things.

This is what research looks like when the block sends the filter to the view:
Screen Shot 2021-07-27 at 11 33 00 AM

This is what research looks like when the filter is exposed and submitted:
Screen Shot 2021-07-27 at 11 35 59 AM

pyrello:
I'm taking a look at this.

pyrello:
A couple things going on here:
- When exposing the operator for an exposed filter, the two items are built with a wrapper around them and our code mostly just assumes that the filters will all be direct children of `$form['override']['exposed_filters']`.
- Filters are stored early on in the process using their Views-generated ID (e.g. `field_tags_target_id`) and when the exposed filter ID is customized (e.g. `tags`). Our code doesn't fully account for that. I started working on this problem in #3815, but I think it will require a more significant refactor now that we are exposing filters.

I started working on some changes to account for these items, but more work needs to be done to fully account for the above conditions.

pyrello:
Changing the filter names was a little disruptive as we had some bunch of instances where the `type` filter was hard-coded in form override code in `sitenow_people`. I attempted to fix that, but it's not totally working yet. When a selection is made for person type, it is not populating the corresponding option in person type status.

cory-skeers:
> I set research on the block config to "history" and then made the filter exposed which on save, caused this illegal choice error. digging more. might be related?

Could be! I'd be interested to see if this still happens if you set research on the block config, and then make both the research and person filter exposed. If it doesn't happen there, it's likely related. If it does, it may be a separate issue (might need to remove values in `blockSubmit` in addition to disabling the field prior?). I don't have it pulled down currently to test, but can try later.

When the exposed filter is sent to the view, the data appears the same between the two scenarios of only the research filter exposed and both the research and person type filters exposed--in the former, no results are returned, but in the latter the correct results are returned using seemingly the same data structure.

cory-skeers:
Haven't dug into the proposed solutions in this issue (https://www.drupal.org/project/drupal/issues/2842525), but getting the Person Type Status exposed filter to populate with only the correct values given a chosen Person Type isn't as straightforward as adding an `#ajax` attribute to the exposed filter form, as I'd tried.

pyrello:
When building out the configure filter functionality, I used the patch at this issue as a little bit of a guide: https://www.drupal.org/project/ctools/issues/2657060

However, at the time we were not trying to expose filters, so I didn't go all in for their solution as it seemed overly complex. I'm going to revisit this to see what there is to learn from it.

pyrello:
> Haven't dug into the proposed solutions in this issue (https://www.drupal.org/project/drupal/issues/2842525), but getting the Person Type Status exposed filter to populate with only the correct values given a chosen Person Type isn't as straightforward as adding an `#ajax` attribute to the exposed filter form, as I'd tried.

I think that we may want to consider using custom JS here instead of continuing to go the AJAX form route. We could load JS that is triggered by changes to the person type field and updates the person type status field to only allow the appropriate options. If we did this, we could completely bypass this problem, I believe.

pyrello:
Note to attempt to address #3661 as part of this work.

pyrello:
As part of the refactor on this, some of the work is now going on in this MR for the `ctools` module: https://git.drupalcode.org/project/ctools/-/merge_requests/4/diffs

pyrello:
As I noted in our huddle yesterday, there seems to be an inconsistency in the way that views expects the filter data for entity references to be stored vs. how a form builder expects it to be stored. 

Views expects something like:
```php
$handler->value = [
  0 =>  16,
];
```

While FormBuilder expects something like:
```php
$handler->value = [
  0 => [
    'target_id' => 16,
  ],
];
```
If FormBuilder doesn't get the above format, then there is a fatal error when the values are passed into `EntityAutocomplete::valueCallback()` because it is passed to an array_map function that needs the above format to work.

In the current version of the patch we are using for Ctools (https://www.drupal.org/project/ctools/issues/2657060), I have ensured that the filter data for entity references gets stored in the block configuration in the second format. I was in the process of trying to understand the implications of the above information when I ran out of time to work on this.

Also, per https://www.drupal.org/project/ctools/issues/2657060#comment-10824250, I added patches from the following two issues:
* https://www.drupal.org/project/drupal/issues/2653938
* https://www.drupal.org/project/drupal/issues/2662978

I reviewed the patches and the second one at least makes sense to me logically. However, after adding those patches, the stored filter values from the block stopped working on the view block. So there may be something there to investigate.

pyrello:
I also add this issue for Drupal core, which would make all the checks we are doing in `ctools_views` for a filter's expose identifier easier: https://www.drupal.org/project/drupal/issues/3227124

pyrello:
A guide to working with the ctools MR that is part of this issue:

Note: I found it helpful to open a second terminal window to be able to run commands in the `ctools` directory while still being able to run `drush` and `blt` commands in the root directory.

To check out the MR and make changes to it:
```
cd docroot/modules/contrib
rm -rf ./ctools
git clone https://git.drupalcode.org/project/ctools.git
cd ctools
git remote add ctools-2657060 git@git.drupal.org:issue/ctools-2657060.git
git fetch ctools-2657060
git checkout -b '2657060-add-configure-filter' --track ctools-2657060/'2657060-add-configure-filter'
```
Make code changes, test, debug, and commit. Once you are satisfied that the changes should be added to the MR, `git push`.

To update the patch (for being able to share progress with other developers on the team):
- Make note of what the next comment number will be on https://www.drupal.org/project/ctools/issues/2657060. It will say the comment number after "Add new comment #" before the form to add a new comment.
- Make note of the last comment number where a patch was added.
- Download the last patch to the `ctools` module directory.
```
git diff 8.x-3.x > 2657060-[next-comment-number].patch
interdiff 2657060-[last-comment-number].patch 2657060-[next-comment-number].patch > interdiff_[last-comment-number]-[next-comment-number].txt
```
- Upload the patch and the interdiff and add a new comment.
- Update `composer.json` to point to the most recent patch and run `composer update --lock`
- Commit and push changes.

# References
- https://www.drupal.org/project/ctools/git-instructions
- https://www.drupal.org/docs/develop/git/using-git-to-contribute-to-drupal/creating-an-interdiff#s-create-interdiff-using-patchutils
- https://www.drupal.org/docs/develop/git/using-git-to-contribute-to-drupal/making-a-patch#s-make-the-patch

richardbporter:
I looked at this but I think I'll need help understanding in greater detail.

richardbporter:
Initial (potential) steps to get ctools tests running in uiowa app:

### Set up ctools or another project using source repo.
```
rm -rf docroot/modules/contrib/ctools
composer install --prefer-source
cd docroot/modules/contrib/ctools
```
Follow Git instructions on issue for setting up another remote within that project directory.

### Run test
Example
``` 
./vendor/bin/phpunit -c docroot/core docroot/modules/contrib/ctools/modules/ctools_views/tests/
```

If xDebug gets stuck on `Standard input code`, configure PHPStorm Debug servers to not break at first line.

The browser output URLs should work and are very useful.

#### Initial test debugging
```
POST request to: https://local.drupal.uiowa.edu/admin/structure/block/add/views_block%3Actools_views_entity_test-block_filter_text/classy
The website encountered an unexpected error. Please try again later.

Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for block.block.views_block__ctools_views_entity_test_block_filter_text with the following errors: block.block.views_block__ctools_views_entity_test_block_filter_text:settings.exposed.filter-field_ctools_views_text_value missing schema, block.block.views_block__ctools_views_entity_test_block_filter_text:settings.exposed_raw missing schema in Drupal\Core\Config\Development\ConfigSchemaChecker->onConfigSave() (line 94 of core/lib/Drupal/Core/Config/Development/ConfigSchemaChecker.php).
```

Setting `protected $strictConfigSchema = FALSE;` on `CToolsViewsEntityViewBlockTest` gets past this.

```
GET request to: https://local.drupal.uiowa.edu/admin/structure/block/list/classy?block-placement=views-block__ctools-views-entity-test-block-filter-text
The website encountered an unexpected error. Please try again later.

Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "ctools.views.handlers.helper". in Drupal\Component\DependencyInjection\Container->get() (line 151 of core/lib/Drupal/Component/DependencyInjection/Container.php).
```
There is no dependency on ctools itself which is where the `ctools.views.handlers.helper` service gets loaded. Adding it to the `$modules` property of the test or as a dependency of `ctools_views` gets past this.

Thats as far as I got. I can add this info the d.o issue if you'd like, or if it gets me credit. :)

richardbporter:
We might want to check BLT docs on testing in case we missed something on setting this up.

richardbporter:
I think 78cc8aa needs to be reverted. I took a deeper dive into this in #4063 and think we should probably use BLT as a wrapper for running core/contrib tests within this project. I can successfully run ctools_views tests in #4063 using the method outlined in the tests/README on that branch: https://github.com/uiowa/uiowa/pull/4063/files#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acR21

richardbporter:
Reverted. I updated the instructions in https://github.com/uiowa/uiowa/pull/3829#issuecomment-910490182 to use core's PHPUnit config. The top-level phpunit.xml.dist file should be used for our own application tests.

pyrello:
Running tests for `ctools_views` locally to see why they are failing and fixing the issues is the next step on this.

pyrello:
To trigger the fatal error that is forcing us to use both the `exposed` and `exposed_raw` filter config currently, follow these steps:
* `blt ds --site sandbox.uiowa.edu`
* `drush @sandbox.local uli node/1/layout`
* Add a people block in the section under the "Getting Started" heading. Set the exposed filter for Tags to "Mascot (16)". Save settings.
* In `Drupal\ctools_views\Plugin\Display\Block` (`docroot/modules/contrib/ctools/modules/ctools_views/src/Plugin/Display/Block.php`) on line 582 change
  ```php
  foreach ($config['exposed_raw'] as $key => $value) {
  ```
  to
  ```php
  foreach ($config['exposed'] as $key => $value) {
  ```
* Reload the page and you should see the error.

To the best of my knowledge, this is what we need to resolve to get rid of the `exposed_raw` config. The format of the values in `exposed` matches up with what Views is expecting (even though this is different than how these types of values are sometimes stored). This is primarily a problem with the `TaxonomyTermTid` filter (which is what is being used for the Tags filter), but we have to solve it in a generalized way by using filter plugin methods. It is not outside the realm of possibility to create a core patch to change something related to the `TaxonomyTermTid` filter if we can figure out how to make that work.

pyrello:
The error that is generated by the above steps specifically happens in `Drupal\Core\Entity\Element\EntityAutocomplete::valueCallback()` because that function is expecting the format of the stored term IDs as:
```php
$input = [
  ['target_id' => 1],
  ['target_id' => 3],
  // etc.
];
```
Instead, it is stored as:
```php
$input = [
  0 => 1,
  1 => 3,
  // etc.
];
```
The error happens here: https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/lib/Drupal/Core/Entity/Element/EntityAutocomplete.php#L126-128

However, when an exposed filter is shown on a Views page display this is not an issue so it seems like it should work here. Trying to figure out what we are doing different and correcting it is going to be part of the process here.

pyrello:
# Some points of code comparison:
## Form generation
`Drupal\views\Form\ViewsExposedForm:buildForm()`:
https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/modules/views/src/Form/ViewsExposedForm.php#L69-157
`Drupal\ctools_views\Plugin\Display\Block::blockForm()`:
https://git.drupalcode.org/project/ctools/-/blob/51d9d152d63195bcdc5f952d6d3606c4e7a17413/modules/ctools_views/src/Plugin/Display/Block.php#L260-348

## Form validation
`Drupal\views\Form\ViewsExposedForm:validateForm()`: https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/modules/views/src/Form/ViewsExposedForm.php#L162-175
`Drupal\ctools_views\Plugin\Display\Block::blockValidate()`: https://git.drupalcode.org/project/ctools/-/blob/51d9d152d63195bcdc5f952d6d3606c4e7a17413/modules/ctools_views/src/Plugin/Display/Block.php#L403-436

pyrello:
I think that all the relevant changes from #4479 have been moved over and therefore I am going to close this PR.

PR #3834, opened 2021-07-20
Hover color fix for black url button on yellow background banner

### LINKED ISSUE #3767: Banner URL button does not have hover effect
# Problem
If you have a yellow brain pattern banner and add a link to it, the button has no hover "underline" effect. It also looks like the ids coming in the CSS are different. See screenshot below:

image

# Steps to reproduce
https://tippie.prod.drupal.uiowa.edu/banners

makurj:
Fixes: #3767 

# How to test
1. Create banners with gold backgrounds and insert URL.
2. Hover to make sure a gold hover shows up on the black button.
3. Make sure other banners are okay.


bspeare:
Testing

bspeare (approved):
This looks good!  I updated the scope to only apply to the `.banner` so that it will not change if the background section is set to gold.

PR #3823, opened 2021-07-16
SiteNow Dispatch Module

GaryRidgway:
Resolves #3302 

## To test

Sync down a site
Add the block
enter a non-uiowa email and first & last name
check the ITS email that something got sent through
[Verify here](https://apps.its.uiowa.edu/dispatch/populations/718296392/members) that the email got added to the list.
Delete the example email that got added so it does not clutter the list
?????
Profit.

### \*\*hint\*\*
Ask Alan for an API key

PR #3853, opened 2021-07-22
Require dispatch signup form email field, change submit message and add headline field.

richardbporter:
Follow up to #3823.

### Setup 
- Configure SiteNow Dispatch settings with csi.drupal API key.
- Add a signup form to a page.

### Stories
- As someone subscribing to a list, I cannot submit the form without entering my email.
- As someone subscribing to a list, I see a confirmation message that makes sense when I submit the form.
- As a site user, I can add a signup form to a page with a headline appropriately sized and styled.

### Alan's added story
- As someone trying to add a sign up form block, I cannot add the block to the page until I have configured an API key.

### DIFF
--- a/docroot/modules/custom/sitenow_dispatch/sitenow_dispatch.module
+++ b/docroot/modules/custom/sitenow_dispatch/sitenow_dispatch.module
@@ -1,0 +1,18 @@
+<?php
+
+/**
+ * @file
+ * Contains sitenow_dispatch.module.
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function sitenow_dispatch_form_layout_builder_add_block_alter(&$form, FormStateInterface $form_state, $form_id) {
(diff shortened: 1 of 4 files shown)


richardbporter:
Hm. Noticed one other thing.

richardbporter:
First/last name are not required fields so we can't rely on outputting them in a confirmation message.

richardbporter:
Just noticed we don't have a title field on this block. That would be useful.

GaryRidgway:
I will approve this if someone else can approve my addition to it.

I've added a check to not display an 'add block' button when adding the block if there is no api key configured so that there are no incorrect form submissions.

cory-skeers:
Pulling down to test

cory-skeers (approved):
> ### Alan's added story
> * As someone trying to add a sign up form block, I cannot add the block to the page until I have configured an API key.

Checked out this story, and looks good!

Also went through the other stories for another pair of eyes.

richardbporter:
Thanks!

PR #3864, opened 2021-07-23
Feature: Empty hamburger bug fix

### LINKED ISSUE #3828: Hamburger Menu appearing in mobile view even when there are no child items
On bam.sites.uiowa.edu I noticed that the hamburger menu was appearing in mobile sizes even though there are no menu items to display. 

needless-hamberber

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3828. 

# How to test

- Test all menu options, toggle, horizontal, mega with and without enabled menu items.    If there are no menu items, then the hamburger menu should not show for mobile or for the desktop toggle version.


richardbporter:
This doesn't fix #3828 because of https://github.com/uiowa/uiowa/issues/3828#issuecomment-886909863. Looking into other options...

richardbporter:
Does this seem ok? https://github.com/uiowa/uiowa/pull/3864/files#diff-8b86d0481fbe21d2b3f16cde4ff1be44614d90bdd3154d9ea894b87fae016edbR987

richardbporter:
This seems to fix #3864 now. I tried to add some behat tests to cover these scenarios. Those appear to be working as well.

joewhitsitt:
pulling down

joewhitsitt (approved):
Looks good. Write a few more tests and I might get motivated to write one.

richardbporter:
One thing I'm confused about is why the off-canvas template works as-is but the superfish template change was required.

Issue #3865, opened 2021-07-26
Make sure file (image) migrations are getting metadata


cory-skeers:
# Story

If a file is imported without title/alt, the system defaults to "title" and "alt" which is not very usable for content editors or end site visitors.

Sitenow Migrate should be looking for these as part of the migration. If they are not found and/or there is an error, it should use a reasonable default, such as filename for usability.

# Proposed solution

Check the migration logic for files, both the Files migration and the BaseNode media trait extensions, and/or run some test migrations and check the end results.

cory-skeers:
Couple notes on this: Engineering migration specifically used `processImageField`, `replaceInlineImages`, and `replaceInlineFiles`, in that order. 

`processImageField` has been updated since that migration, so it's hard to say if there was an issue there or not without further digging.
Looking at the code, we should update this:
https://github.com/uiowa/uiowa/blob/907d836abe95b140916b5f64dd00847fddc72b1d/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php#L406-L409
As is, it's possible to have a `null` title.

`replaceInlineImages`: There are spots where the 'title' attribute doesn't exist, and in the case of engineering, the 'alt' attribute existed, but often was empty, so this could potentially be extended to account for those two cases.
https://github.com/uiowa/uiowa/blob/907d836abe95b140916b5f64dd00847fddc72b1d/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php#L506-L515

`replaceInlineFiles`, which is the method that uses the regex approach to look for both inline images and inline files:
https://github.com/uiowa/uiowa/blob/907d836abe95b140916b5f64dd00847fddc72b1d/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/ProcessMediaTrait.php#L105-L108
This was the fallback I'd thought was in place at the media creation level, but is not. We should add it there or make sure the other possible image paths above have it in place.

PR #3874, opened 2021-07-27
Create sitenow update to cc all and change engineering media title/alt fields.

### LINKED ISSUE #3870: Engineering media metadata
# Story

As a content editor for engineering.uiowa.edu, I would like my previously migrated media content to have its D7 metadata and/or other unique metadata so that I can easily find my content, and have accessible alt text readily available for site visitors.

# Proposed solution

Check that affected media is easily findable in the D8 database by either null values in title/alt or default values ("title" and "alt"). Find a way to update these entries using the D7 database (likely by matching filenames), or at least update the titles to something reasonable like filename minus extension.

richardbporter:
Resolves #3870

I'm not sure how to connect to the D7 database to make this better. If someone has an idea of how to do that, I think this could be closed in favor of that approach as it'd probably have better alt/title fields.

While syncing this site, I got a fatal error related to #3745 so I included a cc all in the update hook.

### Setup
```
blt sync --site engineering.uiowa.edu
```

### Stories
- As an engineering site content creator, I can browse and search media by name.
- As an engineering site content creator, I can embed images with decent alt text.

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.install
+++ b/docroot/profiles/custom/sitenow/sitenow.install
@@ -16,10 +16,12 @@ use Drupal\Core\Entity\Sql\SqlContentEntityStorageException;
 use Drupal\Core\Site\Settings;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\file\Entity\File;
 use Drupal\filter\Plugin\Filter\FilterHtmlCorrector;
 use Drupal\layout_builder\InlineBlockUsage;
 use Drupal\layout_builder\Section;
 use Drupal\layout_builder\SectionComponent;
+use Drupal\media\Entity\Media;
 use Drupal\menu_link_content\Entity\MenuLinkContent;
 use Drupal\node\Entity\Node;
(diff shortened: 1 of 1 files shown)


dnkemper (approved):
Can't think of an alternative way at the moment... works as should though and will approve it - understand if you want someone else to look at this though

richardbporter:
Thanks. I'd be looking at @cory-skeers as that someone else. ;)

cory-skeers:
> Thanks. I'd be looking at @cory-skeers as that someone else. ;)

I'll take a look!

cory-skeers:
At least at first pass trying to track down original file data, the small handful of example images I used didn't have any alt or title information on the D7 site. With that and that there are at least some that migrated with full title/alt, I'm inclined to think this was a lack of info in the original site combined with a failure of failsafe to give unique media titles. 

I'd still like to look a little further to check that that hunch is correct, but if it is, this is likely the best we can offer and would match what the D7 site had in place.

cory-skeers:
I went through some more examples, and still haven't found a situation where there was a title and/or alt that was missed, where there's anything more we could do than what's in this PR.

So we should still look at #3865, but for engineering I think this is the best approach.

PR #3876, opened 2021-07-28
SPPA requester role

pyrello:
I didn't realize that the requester role was specific to the SiteNow website. This PR adds that functionality to SPPA as well.

# How to test

I am comfortable with a basic review of file changes since these changes are limited to SPPA. Not sure how to test it until it is on PROD.


cory-skeers (approved):
Reviewed, not tested

PR #3883, opened 2021-07-29
Pharmacy map of Palliative Care Graduates

### LINKED ISSUE #3624: Pharmacy map of Palliative Care Graduates
# Story

As a Pharmacy website owner, I would like the ability for site users to be able to view a map of palliative care graduates in the US to showcase that program and help with recruitment.

# Additional Info

They like https://admissions.prod.drupal.uiowa.edu/counselors but would want some way to designate that there is a graduate in a state vs states with no graduate. Discussed a different color on the map or possibly dots for each location.

They are fine with results below the map but would be open to other options.

Could technically have multiple graduates in one location.

They data that would be available is:

- Name
- Job Title
- Area of Practice
- Location
- Photo - not required

The Palliative Care Certificate is a Registrar sanctioned certificate that has been around for 3ish years and averages 15-20 graduates per year so maybe around 50 records at most initially.

The College of Pharmacy is not currently using the person content type.

Also shared https://international.uiowa.edu/global-network and https://uihc.org/our-current-resident-class as examples but liked Admissions more.

# Proposed solution

- [x] Site split
- [x] ~~Add Palliative Care Graduate tag (in PROD) to target as MVP instead of a split person type.~~ _person type done instead_
- [x] Location field (state/territory dropdown)?
- [x] Copy and adjust guts of Admissions' counselors map
- [x] Copy and adjust guts of Admissions' counselors view
- [x] Make sure it doesn't split the basic page layout builder config
- [ ] Area of Practice field? https://github.com/uiowa/uiowa/issues/3624#issuecomment-892857891
- [x] Enable split/module in PROD
- [ ] Place blocks on layout builder page

dnkemper:
Relates to #3624

## Story

As a site editor, I can maintain a list of Palliative Care Graduates and the territories they relate to and display the results on a map within layout builder.

## How to test

- `blt frontend`
- `drush @pharmacy.local config:import --source ../config/pharmacy.uiowa.edu --partial`
- `drush @pharmacy.local en pharmacy_core`
- Create at least one person with the Palliative Care Graduate person type designation and a territory.
- Find the map and the list within the more section of layout builder
- See that the person and their related territory are display appropriately within the list and map.

## ToDo
For reference: https://github.com/uiowa/uiowa/pull/2521/files

- [x] Get person type filter on view to work (the string is not working) or switch to tag like stated in #3624
- [x] yarn workspace
- [x] libraries
- [x] map js
- [x] block calling map js and getting variables (allowed by lb page default) @bspeare


richardbporter:
After enabling the split, you can import config again to enable the module: `drush @pharmacy.local cim`. I had to run `blt frontend` to transpile JS before the map would render.

Is there any use to placing the map and/or list individually? If not, it is a bit odd that they are separate blocks in LB.

joewhitsitt:
> Is there any use to placing the map and/or list individually?

Technically no. One is a views block, the other is a custom block. We just built it the same as we had done for Admissions. As-is, this gives them the freedom to separate it within layout builder if the design dictates it.

richardbporter (approved):
Works for me. Fixed some PHPCS warnings.

joewhitsitt:
With Rick's approval, the basic functionality of the issue is done. There is still an outstanding question https://github.com/uiowa/uiowa/issues/3624#issuecomment-892857891 but it could be addressed in a follow-up PR if needed.

PR #3882, opened 2021-07-29
Revert "Added .htaccess redirect for www.(cs|math|stat).uiowa.edu/~ to homepage.divms.uiowa.edu/~."

richardbporter:
Reverts uiowa/uiowa#3879

PR #3884, opened 2021-07-29
Implement redirect from www.(cs|math|stat).uiowa.edu/~ to homepage.divms.uiowa.edu/~.

richardbporter:
Follow up from #3882

### Testing

```
rm docroot/.htaccess &&
composer install
```

Inspect `docroot/.htaccess` file.

Also, you run some URLs through https://htaccess.madewithlove.be/.


mark-bennett-uiowa (approved):
Pulled lastest PR, removed .htaccess file, ran composer install, and the file reappeared with the changes. Code is properly in place, so PR seems to be working.

Not going to test rewrite rules and conditions as I did that numerous times when creating it, and it has been working for several years.

Thanks @richardbporter

PR #3903, opened 2021-08-03
remove 'menuparent' class from cloned parent link to prevent FontAwes…

### LINKED ISSUE #3902: Missing FontAwesome icon on mobile Safari only in Superfish subnavigation
# Problem
Missing FontAwesome icon shown in subnavigation only when viewing from a mobile device (Safari tested).

superfish-icon-issue

# Steps to reproduce
On a mobile browser, expand the main navigation menu, and then expand a collapsible item in that menu and the first item (aka the clone of the original parent menu item) will have a missing FontAwesome icon. 

# Proposed solution
It looks like some JS may be erroneously applying the "menuparent" class to the list item tag underneath the anchor tag with the class "sf-clone-parent":

Screen Shot 2021-08-03 at 8 29 39 AM

When viewing from a non-mobile browser, the "menuparent" class isn't applied in this context and there isn't a missing icon.

Screen Shot 2021-08-03 at 8 31 49 AM

Investigate why on mobile browsers this "menuparent" class is applied.

quamsta:
Remove 'menuparent' class from cloned parent link to prevent FontAwesome expand icon from being shown. fixes #3902

# How to test

On a mobile device, check to make sure the cloned parent link doesn't have a missing FontAwesome icon showing (usually an [X]-like icon)


joewhitsitt:
going to deploy this branch to dev for mobile review

joewhitsitt (approved):
tested in dev. thanks!

PR #3919, opened 2021-08-05
Brand - set reply-to for email to keep drupal happy

joewhitsitt:
Setting up for demo, I was getting a reply-to not set error locally.

Relates to:
https://www.drupal.org/project/drupal/issues/111702

## To Test

`blt ds --site=brand.uiowa.edu`
set your email for reroute email (https://brand.local.drupal.uiowa.edu/admin/config/development/reroute_email)
`drush @brand.local cr`
`drush @brand.local brand-ld` should error (response 220 something) but instead of complaining that reply-to not being set, it should say that is set to your email address.

The 220 error is because of the mail server differences between Prod and Local. It can be fixed locally by re-configuring the mail system module. For another day.


richardbporter (approved):
👀

PR #3920, opened 2021-08-05
Trim profiles API key on save.

richardbporter:
### Setup
- Follow setup from original Profiles PR: #3644 

### Stories
- As a webmaster configuring SiteNow Profiles, I can copy/paste API keys with leading or trailing spaces for multiple directories and things just work.


cory-skeers (approved):
Didn't specifically test it, but on review, it looks good to go!

Issue #2305, opened 2020-10-07
Drupal secure content for authenticated users


joewhitsitt:
## Issue

It has been requested by a couple customers to have some content on their public website be viewable to authenticated users, or this idea of "password protected content."

## Proposed Solution

- Investigate what type of content is in scope. Reach out to customers about what they had in mind.
  - brand
  - law
  - studentlife
  - iowamommoodbooster.org
  - policy.clas.uiowa.edu
- Decide if this should be accomplished with a single site-wide authentication or have separate "passwords" per content
  - https://github.com/uiowa/uiowa_auth/issues/22
- Based on the scope above decide how to handle private files on a public website.

pyrello:
SPPA also needs something like this. In that case, viewing access for a custom content type and some view listing pages would need to be limited to only users of a certain role or roles.

It seems like, at a minimum, this would require custom code or a contrib module like https://www.drupal.org/project/node_view_published_override to allow setting the "View published content" permissions per content type. It also seems like it would require a custom role, since the Viewer role allows things that we would not want in this case.

quamsta:
We definitely need at least HawkID authentication for viewing some pages at Student Life. We have some clients that rely on this feature pretty heavily.

joewhitsitt:
>It also seems like it would require a custom role, since the Viewer role allows things that we would not want in this case.

Could the viewer role be repurposed and redefined if this ends up as a feature split instead of default?

mark-bennett-uiowa:
@joewhitsitt I think this would work for my needs in #3484 as long as we could accomplish this without giving these users editing rights. Closing that issue in favor of this solution.

mark-bennett-uiowa:
Moving this back to the top of Parking Lot because I now need this for both the iowamommoodbooster.org site and the policy.clas.uiowa.edu site.

Could we possibly investigate using one of these modules for protecting pages behind a password without giving them authenticated permissions??

[https://www.drupal.org/project/protected_pages](https://www.drupal.org/project/protected_pages)
**This seems like the best option** on paper as it allows sweeping protection of all protected pages vs individual nodes. Updated for 8/9 on April 29, 2021.

[https://www.drupal.org/project/protected_node](https://www.drupal.org/project/protected_node)
This is the node only protection module, no 8/9 version.

[https://www.drupal.org/project/shield](https://www.drupal.org/project/shield)
This is a walled-garden approach, and may seem too restrictive. Does have a Drupal 8/9 version, updated March 30, 2020.

[...]

PR #3925, opened 2021-08-05
SPPA - Private Page Content Type

### LINKED ISSUE #3922: SPPA - Create private page content type
# Motivation

SPPA has a career services manual that was previously only available to logged-in students. That page is not publicly visible. https://sppa.prod.drupal.uiowa.edu/student-resources/career-services-manual

# Story

As an SPPA site owner I want to be able to create page content that is only viewable by logged-in users.

# Proposed solution

- Add a "Private page" content type that is like the standard page, but has viewing privileges restricted to authenticated users. 
- Add this config to SPPA site split.

dnkemper:
Closes #3922 

# How to test 
1. `blt ds --site=sppa.uiowa.edu`
2. `yarn workspace sppa_core build`
3. `drush @sppa.local cr`
4. `drush @sppa.local uli`
5. Go to add a private page - should be same experience as adding a page
    1. Same options/blocks/styling/etc
6. Save the page with some content added from the ‘Edit’ section in addition to the ‘Layout’ section 
    1. be sure to add a statistic because had to fix styling to display similar to page
7. Save page
8. You should be able to view the page as a logged in user - if you view the node in incognito or logout - you should view an access denied page that says you’re not authorized to access the page


joewhitsitt:
> Featured content block: I added option for users to select published ‘private page’ content - incase they want a page that displays specific private pages?

- Will anon be able to see the content that is referenced in these? I don't think this is necessary.

> URL Alias: not sure what path to use?

- [node:title] like page should be good for MVP.

> Moderated content: Do I need the views file?

- I don't understand

dnkemper:
@joewhitsitt, thank you!
1. Removed featured content work
2. Updated alias
3. I will investigate this - pretty sure I exported something that didn't need to be exported 😬

richardbporter:
This seems to work as advertised but I noticed a couple things:

- There are quite a few files used within that section of the SPPA website. It would be good to set expectations with the customer that all files used within LB - docs, images, etc. - will _not_ be private.
- Assuming they are ok with that, I think we should output a disclaimer on the node edit and layout screens stating that. Otherwise, someone might upload a sensitive document expecting it to be private.

Overall, I would've preferred a simpler page as @joewhitsitt outlined in https://github.com/uiowa/uiowa/issues/2305#issuecomment-856869255. That would _actually_ have private files at the expense of LB which I think is a reasonable tradeoff. Does private information need to be presented in a unique, marketable way? I'm not sure. 

Considering it will take some time to convert all those existing pages to private pages, I wonder if this could wait for @pyrello next week.

docroot/sites/sppa.uiowa.edu/modules/sppa_core/sppa_core.module
```diff
@@ -7,6 +7,181 @@
 
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Custom node content type form defaults.
+ */
+function _sppa_node_form_defaults(&$form, $form_state) {
```
joewhitsitt: instead of duplicating _sitenow_node_form_defaults, you can call this and override after if necessary. like here https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L509

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -475,6 +475,13 @@ function layout_builder_custom_form_node_page_layout_builder_form_alter(&$form,
   $form['#attached']['library'][] = 'layout_builder_custom/layout_builder_custom.overrides';
 }
 
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function layout_builder_custom_form_node_private_page_layout_builder_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
joewhitsitt: can this move to sppa_core as `sppa_core_form_node_private_page_layout_builder_form_alter()` since this is a one-off for now?

docroot/sites/sppa.uiowa.edu/modules/sppa_core/sppa_core.module
dnkemper: @joewhitsitt thank you so much! So would I need to create a `sitenow_private_pages.settings_form` then for the featured_image_display settings? I get an error when calling the function.

joewhitsitt: ah i see... that stuff wasn't in there when i first created it. ideally it would be adjusted to be more flexible so it isn't following the `sitenow_` prefix. let me ponder this a bit

dnkemper: Okay! Thank you so much for your help!

joewhitsitt: I am going to fix the logic around the `field_featured_image_display` stuff in this function. plan on `sppa_core.settings_form` and you can comment out this snippet for now.

joewhitsitt:
pulling down to test overall functionality

- from what I can tell, sppa_core doesn't need a yarn workspace build because the admin.css is not needed for those stat form styles. those styles are already loaded by uids_base in the layout template.
- I agree with with what Rick said about outputting a message. _"I think we should output a disclaimer on the node edit and layout screens stating that."_ I think we broke node submission guidelines by not having the help module on/placed so doing a `\Drupal::messenger()->addWarning()` for the add/edit/layout for the private page content type would work.
- I wonder if a separate message should be output to viewers to somehow distinguish the difference between a page and a private page. "You are viewing private content."
- I think the override for Promoted to front page and Sticky at top of lists could be removed.
- field_publish_options isn't needed unless support for Visually hide title and/or Remove sidebar regions are added.
- i get access denied logged in as a student. permissions should be extended to this role
- field_teaser is used for teaser view modes and seo. since those aren't a thing with this content type, should field_teaser exist?

dnkemper:
- [ ] I agree with with what Rick said about outputting a message. _"I think we should output a disclaimer on the node edit and layout screens stating that."_ I think we broke node submission guidelines by not having the help module on/placed so doing a `\Drupal::messenger()->addWarning()` for the add/edit/layout for the private page content type would work.
- [ ]  I wonder if a separate message should be output to viewers to somehow distinguish the difference between a page and a private page. "You are viewing private content."
- [ ] I think the override for Promoted to front page and Sticky at top of lists could be removed.
- [ ] field_publish_options isn't needed unless support for Visually hide title and/or Remove sidebar regions are added.
- [ ] i get access denied logged in as a student. permissions should be extended to this role - @joewhitsitt I was following what existed for opportunity content type where students can `only` view the nodes. Or is that not working?
- [x] field_teaser is used for teaser view modes and seo. since those aren't a thing with this content type, should field_teaser exist?
- [x]  from what I can tell, sppa_core doesn't need a yarn workspace build because the admin.css is not needed for those stat form styles. those styles are already loaded by uids_base in the layout template.

joewhitsitt:
> I was following what existed for opportunity content type where students can only view the nodes. Or is that not working?

I didn't check why it wasn't working but i got an access denied when masquerading as `seif` (student) and trying to view a published private_page. Can you replicate? I am on a different branch.

dnkemper:
> > I was following what existed for opportunity content type where students can only view the nodes. Or is that not working?
> 
> 
> 
> I didn't check why it wasn't working but i got an access denied when masquerading as `seif` (student) and trying to view a published private_page. Can you replicate? I am on a different branch.

I tried to replicate, but it seems to be working fine for me? 

I added the warning message - wasn't sure how you were wanting it to display on the layout builder layout form? The way I did it didn't require me to mess with core, but understand if you're not a fan since it merges itself with the existing moderation message. I'd also greatly appreciate your content expertise with what to say haha. Fail. 

I am slightly confused with what you mean by distinguishing the private page vs normal page. Like to show authenticated users what node they are looking at when not editing the page or layout? If that's the case, wouldn't that get annoying for users? 

I got the sidebar/hide title page options working by adding the private page node to the existing functions, hopefully that's okay. I believe it's ready to be reviewed.

joewhitsitt:
taking a look

joewhitsitt:
>I added the warning message - wasn't sure how you were wanting it to display on the layout builder layout form?

#works4me. This is what I was thinking when I mentioned `node_private_page_layout_builder_form` in Slack.

>I tried to replicate, but it seems to be working fine for me?

My mistake. I saw an access denied after masquerading as a student because I was on the user admin page not the private page content. It didn't "back to site" like I expected.

I think https://github.com/uiowa/uiowa/pull/3925/commits/968c82e23a8b2795ebe8387dbda57e21b489481f was an overcorrection. The view role for other content types can see and "review" revisions. Think of an org admin that doesn't need edit access but needs to sign off on the copy before publish instead of reviewing a word document.

I don't recall what those `..layout overrides` permissions do, but if they have them for page, I would assume they would have them for this as well.



>I got the sidebar/hide title page options working

Nice. I moved it out of the profile code. Ideally we want less custom in the default.

>I am slightly confused with what you mean by distinguishing the private page vs normal page. Like to show authenticated users what node they are looking at when not editing the page or layout? If that's the case, wouldn't that get annoying for users?

I am not sure the message is the best solution for this, the work done to replicate the page content type is so good that as a logged in user I can't tell the difference unless I am familiar with the content or go to the edit screen where it says, "Edit Private Page..."

Screen Shot 2021-08-12 at 8 27 33 AM
Screen Shot 2021-08-12 at 8 27 43 AM

dnkemper:
> I think [968c82e](https://github.com/uiowa/uiowa/commit/968c82e23a8b2795ebe8387dbda57e21b489481f) was an overcorrection. The view role for other content types can see and "review" revisions. Think of an org admin that doesn't need edit access but needs to sign off on the copy before publish instead of reviewing a word document.

Even if it's supposed to be private to viewers? That was my uncertainty and I followed the opportunity content type 

>I am not sure the message is the best solution for this, the work done to replicate the page content type is so good that as a logged in user I can't tell the difference unless I am familiar with the content or go to the edit screen where it says, "Edit Private Page..."

What would you like me to do then? Still slightly confused. I'm sorry 😞

joewhitsitt:
viewer is an authenticated role which is different than anonymous (private) so they should be able to see this content per the issue: "page content that is only viewable by logged-in users."

>What would you like me to do then?
i haven't given it much thought besides a message. wordpress i think adds "private:.." to the page title. idk.

dnkemper:
> viewer is an authenticated role which is different than anonymous (private) so they should be able to see this content per the issue: "page content that is only viewable by logged-in users."

I KNOW NOTHING

> > What would you like me to do then?
> > i haven't given it much thought besides a message. wordpress i think adds "private:.." to the page title. idk.

I can definitely add a private prefix to the title field? I can also add a message - I misunderstood and thought you were wanting a message for both basic page and private. I'm sorry about that!

pyrello:
I am pursuing a different option which allows us to use the page content type and adds an option to make the page private: #3986. The benefit of this other approach is that it allows us to remove the `node_view_permissions` module, which had the unfortunate side-effect of creating a lot of duplicated config in config splits.

I think the points here about showing the user a message warning them about their files not being private were well-intended, but this got a little out-of-hand for an MVP feature for one site. Should we run into issues on SPPA or if we roll this out to everyone eventually, I think adding such messages would be good idea, but are not necessary now.

Leaving this open until we get some eyes on the other solution and decide that is how we all want to move forward.

PR #3912, opened 2021-08-04
Feature: Obermann articles migration

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3697. 

### Setup
Sync obermann.uiowa.edu D7 site in DevDesktop.

Sync obermann.uiowa.edu D9 site.
```
blt sync --site obermann.uiowa.edu &&
drush @obermann.local en obermann_migrate --yes &&
drush @obermann.local uli /admin/config/system/sitenow-migrate
```

Fill out form for DD.

`drush @obermann.local ms`
`drush @obermann.local mim --all`

### Stories
- As an obermann.uiowa.edu site owner, I can see articles created before 9/9/2020 in the new site and none created after that except the ones I created manually.
- As an obermann.uiowa.edu site owner, I can see those migrated articles have references to tags with the same name as a featured program if such a reference existed on the D7 site.
- As an obermann.uiowa.edu site owner, I can be provided a list of broken links in migrated articles so that I can fix them.
- As an obermann.uiowa.edu site visitor, I can get redirected to the new location of a migrated article in case I have it bookmarked.

PR #3947, opened 2021-08-11
Obermann migration followup.

richardbporter:
Follow up to #3912

### DIFF
--- a/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
+++ b/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
@@ -85,6 +85,9 @@ abstract class BaseNodeSource extends Node implements ImportAwareInterface {
     parent::prepareRow($row);
     $moderation_state = $row->getSourceProperty('status') == 1 ? 'published' : 'draft';
     $row->setSourceProperty('moderation_state', $moderation_state);
+    if (empty($row->getSourceProperty('body_summary'))) {
+      $row->setSourceProperty('body_summary', $this->extractSummaryFromText($row->getSourceProperty('body_value')));
+    }
   }
 
   /**
@@ -106,7 +109,7 @@ abstract class BaseNodeSource extends Node implements ImportAwareInterface {
     // that might cause weird spacing for our summaries.
     $new_summary = ltrim(strip_tags($new_summary));
 
-    $new_summary = substr($new_summary, 0, 200);
+    $new_summary = substr($new_summary, 0, 400);
 
     $looper = TRUE;
 
--- a/docroot/modules/custom/sitenow_migrate/tests/src/Kernel/BaseNodeSourceTest.php
+++ b/docroot/modules/custom/sitenow_migrate/tests/src/Kernel/BaseNodeSourceTest.php
@@ -1,0 +1,136 @@
+<?php
+
+namespace Drupal\Tests\sitenow_migrate\Kernel;
+
+use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Extension\ModuleHandler;
+use Drupal\Core\File\FileSystem;
+use Drupal\Core\State\State;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\migrate\Plugin\Migration;
+use Drupal\sitenow_migrate\Plugin\migrate\source\BaseNodeSource;
+
+/**
+ * Test description.
+ *
+ * @group sitenow_migrate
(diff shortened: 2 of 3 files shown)


docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
```diff
@@ -85,6 +85,9 @@ public function prepareRow(Row $row) {
     parent::prepareRow($row);
     $moderation_state = $row->getSourceProperty('status') == 1 ? 'published' : 'draft';
     $row->setSourceProperty('moderation_state', $moderation_state);
+    if (empty($row->getSourceProperty('body_summary'))) {
```
cory-skeers, line 88: By default, we weren't checking/replacing summaries, so not all migrations have this. There was at least one migration earlier that specifically didn't want anything like this (I think it was a grad site?), but I think that's the minority case. Might be worth making it the default behavior going forward.

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
```diff
@@ -106,7 +109,7 @@ protected function extractSummaryFromText(string $text) {
     // that might cause weird spacing for our summaries.
     $new_summary = ltrim(strip_tags($new_summary));
 
-    $new_summary = substr($new_summary, 0, 200);
+    $new_summary = substr($new_summary, 0, 400);
```
cory-skeers, line 112: Upping the length to 400 from the small 200 takes care of the test case, but guarantee a similar one won't pop up again, given a wordy-enough intro.

cory-skeers:
@richardbporter I made some slight adjustments here that take care of the Obermann test examples. 

I think it could still be a good prompting to look into using `smart_trim` for this, but if Obermann needs a speedier turnaround, this could do the trick.

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
richardbporter: Hmm. Confused. Wouldn't https://github.com/uiowa/uiowa/blob/028909b1bde3a4ea3a4978154d82551ab38d1135/docroot/sites/obermann.uiowa.edu/modules/obermann_migrate/src/Plugin/migrate/source/Articles.php#L57 have done this?

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
richardbporter: Oh I see. So if a sentence or breaking point or whatever is over the limit, it returns nothing?

richardbporter:
Thanks! I can try to clarify the issues once I understand this a bit better.

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
cory-skeers: Yup! I also forgot there was a handy `getSummaryFromTextField` that handled that. This should only be necessary for testing the base migration then, and unrelated to the obermann migration.

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/BaseNodeSource.php
cory-skeers: Yeah, this part led to both issues. For the date one, the abbreviating period was the first breakpoint it found, but starting farther out it finds a good spot past it. For the blank one, it kept shortening the string without finding a breakpoint, until it shortened it to a 0 length. 

In addition to lengthening the start point, maybe we should add a fallback if it gets too short, throw ellipses at the end of whatever we have instead of letting it get too far. Or just wait until implementing `smart_trim` 😜

richardbporter: Oh yea they want an ellipsis if its over the limit in #3949... Maybe we could try out smart_trim in the update hook for that issue to see how smart it really is and then come back to this. We can refactor these tests a bit to better cover what we're trying to ensure works.

richardbporter:
Closing in favor of #3952. No reason to duplicate smart_trim's TruncateHTML tests. We can reference this PR if there is other functionality we'd like to test in BaseNodeSource.

PR #3958, opened 2021-08-12
Feature: pdf color tweaks

### LINKED ISSUE #3950: Color tweaks to Admissions PDFs
>We have started actually printing some of the areas of study PDF’s and realized that we need to make some quick/simple color tweaks.

>The body text needs to be black (seems to be a darker grey now). Links in parenthesis can stay the same light gray color.
The breakout sidebar box needs to be a darker grey color. The printers we’re printing with are barely picking up that box. It doesn’t need much change, but slightly darker so that the box actually shows up in the prints.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/3950. 

Before:

Screen Shot 2021-08-12 at 12 11 43 PM

After:

Screen Shot 2021-08-12 at 12 11 21 PM

### DIFF
--- a/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/pdf.scss
+++ b/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/pdf.scss
@@ -45,11 +45,11 @@ p,
 ul,
 em {
   font-size: 0.7rem;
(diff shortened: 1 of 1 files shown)


richardbporter (approved):
Looks good. Did not print. :)

PR #3977, opened 2021-08-12
Education Migration

### LINKED ISSUE #3899: Migrate news articles for College of Education
Will get more details when we get closer to tackling this.

## Proposed solution

- Migrate all existing articles.
- Maintain published/unpublished status
- Tags, Affiliations are migrated to Tags. (Might have a list of consolidation for us to map)
- Source Title -> Organization
- Source Link -> Source Link
- Image -> Featured Image (site-wide default)
- Caption -> stripped and truncated to caption
- Body summary -> field_teaser
- Byline -> Prepend to body
- Short Description -> Prepend to body (uids-component--light-intro)
- body -> body (convert btn-primary to bttn bttn--caps bttn--primary, pull-quote to blockquote)
- Create redirects from old articles.
- Provide initial list of potential broken links.

### No action
- Image Gallery
- Header type
- Header image/video

cory-skeers:
Resolves #3899 

- [x] Migrate all existing articles. (total count: 609)
- [x] Maintain published/unpublished status
- [x] Tags, Affiliations are migrated to Tags. (Might have a list of consolidation for us to map)
- [x] Source Title -> Organization
- [x] Source Link -> Source Link
- [x] Image -> Featured Image (site-wide default)
- [x] Caption -> stripped and truncated to caption
- [x] Body summary -> field_teaser
- [x] Byline -> Prepend to body
- [x] Short Description -> Prepend to body (uids-component--light-intro)
- [x] body -> body (convert btn-primary to bttn bttn--caps bttn--primary, 
- [x] body -> body pull-quote to blockquote)
- [x] Create redirects from old articles.
- [x] Provide initial list of potential broken links.

# How to test

Sync the d7 instance locally, and start up the stack
`blt ds --site=education.uiowa.edu`
`drush @education.local en education_migrate`
`drush @education.local uli /admin/config/system/sitenow-migrate`
Complete the config form using "https://education.uiowa.edu" as the base URL
Run the migrations. -- Will likely need to limit the articles migration using something like `drush @education.local mim education_articles --limit=200`, otherwise you may run into a memory issue. (Although! I think all the errors have been worked out, where even if you error out you can re-run it without issues.)
At the end of an import, a lengthy terminal log will contain messaging regarding all articles with suspected broken links (ie relative `node/#` links which are no longer accurate). This report should cover all migrated pages--if multiple imports were run due to the memory issues, only the final report is needed.
Check things out and see that everything looks awesome.
Check that article summaries do not contain any `&#13;` characters (carriage returns that were not properly removed or converted).
Check that article redirects from the previous D7 alias work as expected.

Note for the pull quotes: The following nodes on the existing site contain pull quotes. The first two each contain an instance of a `div` pull quote which consist of an image and caption. These have been converted to images with captions in the migration, and not pull quotes. The rest are instances of `p` pull quotes, which have been converted directly to `blockquote` in the migration.
| entity_id |
|------------|
2041	
2536	
6521	
8586	
8636	
11651	

Some nids from the D7 site which contained a byline field value:

| entity_id	| field_article_byline_value	|
|--------------|----------------------------|
961	| By Andrea Zeek	
2041 | 	By Mikaela Parrick	
2146 | 	By Andy Goodell	
2241 | 	By Andrea Zeek	
2266 | 	By Mikaela Parrick	
2306 | 	By Sean Thompson	
2536 | 	By Andy Goodell	
2806 | 	By Andy Goodell	
2831 | 	By Andy Goodell	
3076 | 	By Andy Goodell	

Some nids which used the short description field:

| entity_id | 	field_header_short_description_value	
|----------|------------------------------------------------------------|
9776 | 	For UI REACH graduate Andrew Sukup, the sky is the...	
9786 | 	UI alumnus Jake Chung plants seeds of hope in comm...	
9851 | 	Iowa City New Horizons band members share love of ...	
9896 | 	Rusty Barceló serves as mentor for generations of ...	
10191 | 	Collaboration, creativity lead to new exhibit at T...	
11191 | 	1.2 million Iowans live in rural communities.	
11201 | 	Helping all students experience safe and supportiv...	
12296 | 	The UI trains educators and administrators who are...	
12656 | 	 This conceptual rendering from CMBA Architects sho...	

And nids from D7 where buttons were present:

| entity_id	 |
|-----------|
679	
3491	
15271	
15276	
15411


cory-skeers:
We put some code in to update instances of `btn-primary`, but in inspecting the database, it looks like there are no instances of `btn-primary` in articles (only in pages and events).

Tested by adding buttons to the local database, and the replacement process works for `btn-primary` and `btn-long`, both mapping to `bttn bttn--caps bttn--primary`.

richardbporter:
I can start testing this.

richardbporter:
Seeing lots of warnings after running with verbose logging: `drush @education.local mim education_articles -v`. Here are some, but probably not all:

- `[info] Trying to get property 'sourceid1' of non-object BaseNodeSource.php:252`
- `[info] Trying to access array offset on value of type null sitenow_media_wysiwyg.module:114`
- `[info] Trying to access array offset on value of type bool ProcessMediaTrait.php:112`
- `[info] Undefined index: target_id Articles.php:250`
- `[info] Undefined index: target_id Articles.php:254`

As I type this, I'm realizing I'm on PHP 7.4 locally so I'd be curious if these showed up for others.

richardbporter:
Just confirming that number of source records is the same between the article and redirect migrations 👍🏻. 

Still trying to nail down the phantom article issue I had on https://github.com/uiowa/uiowa/pull/3912#issuecomment-894464551. :)

richardbporter:
Switched back to PHP 7.3 locally but still seeing those notices.

I also seem to be stuck on the redirect migration:
```
 ------------------- --------------------- -------- ------- ---------- ------------- --------------- 
  Group               Migration ID          Status   Total   Imported   Unprocessed   Last Imported  
 ------------------- --------------------- -------- ------- ---------- ------------- --------------- 
  SiteNow             education_articles    Idle     610     610        0             2021-08-17     
  (sitenow_migrate)                                                                   15:21:15       
  SiteNow             education_articles_   Idle     610     1          609           2021-08-17     
  (sitenow_migrate)   redirects                                                       15:21:37       
 ------------------- --------------------- -------- ------- ---------- ------------- --------------- 
```

It will only import 1 item. I tried running in batches as suggested too but that didn't seem to make a difference, i.e. use the `--limit` option.

richardbporter:
I can't test the redirects obviously but everything else checks out. Too bad they didn't want to create People for the authors because I think that styling looks nice.

Nits:
I don't think these classes will cause problems but _may_ cause confusion if people think they'll actually do something, like change the alignment or font size:
Screen Shot 2021-08-17 at 3 29 01 PM

Screen Shot 2021-08-17 at 3 43 56 PM

cory-skeers:
> It will only import 1 item. I tried running in batches as suggested too but that didn't seem to make a difference, i.e. use the `--limit` option.

I'm wondering if it might be the `getLastMigrated` causing it to skip over them as if they'd been migrated? Working on another issue, but I can check that out after.

cory-skeers:
> I don't think these classes will cause problems but _may_ cause confusion if people think they'll actually do something, like change the alignment or font size:

I wonder if there's a good way to get a curated list of "supported" classes, and strip any that don't fit it? In some of the other migrations I've done little bits of stripping things out like that, but only for things I'd happened to notice appear the same way over and over.

richardbporter:
> > I don't think these classes will cause problems but _may_ cause confusion if people think they'll actually do something, like change the alignment or font size:
> 
> I wonder if there's a good way to get a curated list of "supported" classes, and strip any that don't fit it? In some of the other migrations I've done little bits of stripping things out like that, but only for things I'd happened to notice appear the same way over and over.

Good point. I guess I thought they were all stripped from Filtered HTML but that doesn't seem like the case. I'm not sure how to know what is valid and what is not but I would imagine there are few if any classes from v1 that exist in UIDS.

Maybe @bspeare has thoughts on that?

richardbporter:
> > It will only import 1 item. I tried running in batches as suggested too but that didn't seem to make a difference, i.e. use the `--limit` option.
> 
> I'm wondering if it might be the `getLastMigrated` causing it to skip over them as if they'd been migrated? Working on another issue, but I can check that out after.

Seems like that might be the case. I'm not following that logic otherwise I would fix it. :)

cory-skeers:
> > > It will only import 1 item. I tried running in batches as suggested too but that didn't seem to make a difference, i.e. use the `--limit` option.
> > 
> > 
> > I'm wondering if it might be the `getLastMigrated` causing it to skip over them as if they'd been migrated? Working on another issue, but I can check that out after.
> 
> Seems like that might be the case. I'm not following that logic otherwise I would fix it. :)

Redirects should import properly now. Made a small change to the `getLastMigrated` function, as well as skipping over the custom `prepareRow` stuff that's not needed for redirects.

I can try to look into those notices next.

cory-skeers:
Setting to draft. Going to take a look into the notices, and switch the redirects over to use https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/AliasRedirect.php

cory-skeers:
@richardbporter I think I've gotten all the kinks worked out--running with verbose output should only give the expected notices.

I think we would have missed either creating new affiliations or tags, and not noticed because it was in a hidden notice and not causing an error, so great catch!

richardbporter:
I'm able to run both migrations now. 👍🏻 

Spot checking source node 17256, it seems like the affiliations are different.

Source
Screen Shot 2021-08-18 at 10 15 45 AM

Destination
Screen Shot 2021-08-18 at 10 15 57 AM

cory-skeers:
> Spot checking source node 17256, it seems like the affiliations are different.

Hrm...taking a look.

cory-skeers:
Okay, so had looked at affiliations incorrectly. The issue has us mapping them to tags, but on the D7 site, the references were to nodes. I've split out the handling of tags and affiliations so that they are properly handled now, creating tags based on the titles of the affiliation references.

As long as I have the right understanding from the issue in how we're mapping these, then _now_ we should have them working correctly.

richardbporter (approved):
That seems to have fixed it. I spot checked a couple tags and affiliations. Also checked for any media items named "title" but found none.

Nothing else is sticking out to me.

PR #3992, opened 2021-08-17
Feature: deprovision wirrc.org and iowalakesidelab.org

bspeare:
Deprovisioning these because they require a non-branded theme.


pyrello (approved):
Reviewed, not tested.

PR #4009, opened 2021-08-19
Initial work on the Topic Landing Page

quamsta:
As we've gone through our migrations to SiteNow we've been a bit slowed up since we don't have a functional equivalent of a tag-based navigation that we use on many student-facing resource guides on campus. A few currently live examples include:

- https://basicneeds.uiowa.edu/resources/
- https://fsl.uiowa.edu/resources/
- https://offcampus.uiowa.edu/guide/
- https://sds.studentlife.uiowa.edu/campusresources/

This is a template we've been rolling out to our partners' websites and has proved to be fairly popular and we'd like to bring it into the Drupal/SiteNow world in some form.

This is our initial attempt at bringing it into SiteNow. It's currently not using Layout Builder (aside from a few CSS classes) and we'll need some assistance getting it fully brought into the stack.

topic-landing-preview

# How to test

You'll need to tag some pages or articles in your local environment with taxonomy terms in order for this template to function properly.

Make sure to run `blt frontend` in order to build the SASS files included in the commit.

Upon installation of the "studentlife_topics" module, a new content type "Topic Landing Page" becomes available. After creating a new "Topic Landing Page" node, you're able to select which tags (or terms) show up in the tag "bubbles" and in the navigation below the search while editing the page.

Screen Shot 2021-08-19 at 3 34 17 PM

A featured image can also be chosen as well, but really, that's it. 

# Stuff we haven't quite figured out yet

- Search is not scoped to the tags selected, it just searches the entire site
- How to fit it into the Layout Builder paradigm, currently this is an "as-is" twig template that can't be pulled apart too much. We'd love to figure out how to generalize it.
- How to make our projects more visible to the team. We're trying, haha.


bspeare:
Pulled this down and took a look at this and it seems like this could be quickly rebuilt to reuse existing styles and templates that would decrease the amount of twig functions, styles, and new markup used quite a bit.

1. You could create a student life topics view containing:
- Topics banner view block
  - Extend banner component that has styles that can handle most of this
  - List of tags
  - Search form
- Topics tag list view block
  - List of referenced tags
- Topics tag listing view block
  - List of referenced tags and associated content

2. Update the student life topics page to use layout builder
- Place the blocks listed above into a custom layout with the body field

bspeare:
Another option for this would be to move this into the SiteNow platform to use with the Page content type.

**Use the page content type**

- Change the field_tags taxonomy from comma separated to draggable
- Create one view with multiple blocks containing:
  - Exposed search filter for field_tags
  - List of field_tags on node
  - List of field_tags on node w/number of items
  - List of field_tags with referenced children

benlewzz:
Thanks @bspeare!! That sounds awesome, exciting to see it. Looking forward to talking more about it! 

Here are some things I'm hoping to bring over from the user-end: 

- Easy instructions and ability to create a topic landing page with pre-populated layout (duplicate a set of pre-made pages to help CMS user use it). Maybe duplicate page or set of pages? (I know this isn't available yet, but great use case).
- Easy for CMS users  to understand tagging / how to pull content into topic landing pages
- Search: not sure what we want to do here but this is obv. big priority

richardbporter:
One small concern I have is the term "Landing Page", at least in relation to Basic page which can be use for "landing pages" also. I don't know what else to call it though. I see some references to "Resources" so maybe something with that instead? If I'm the only one with this concern, not a huge deal. :)

joewhitsitt:
I am curious if a usability study or user tracking (e.g. crazyegg) was done on one of these pages in the past. Seems like duplicate content that attempts to catch a browsing users into the same content buckets but different flavors by the end of the page. What form do most users click to get the info they need?

search -> tags, comma-separated -> tags, buttons -> tags as linked headings with a sneak peak

If I was adding this for a site, I might wonder if all of these are worth being placed on the same page, so I like the direction that this would be captured more in config so that sites can override and choose à la carte what they want.

makurj:
I agree with Rick's comment on the term "Landing Page". "Resources" or "reference" or "table of contents" or "resource library", etc. could be possible options.

pyrello:
A few notes after our meeting:
* Recommended path forward is for a separate content type, something like "resource collection" or "topic page."
* We're going to have a follow-up working session to look at turning on Layout Builder as the template engine for the above content type. 
* This allows us to capture the layout details in config while shifting to using individual block templates for components on the page. 
* These block templates could potentially be used for other things in the platform so helps contribute as much back as possible.
* For non-admins, Layout Builder would not be accessible per content item for the above content type. Admins would be able to access the Layout tab and override the layout. I think this is what @joewhitsitt was indicating in his comment above.

joewhitsitt:
>I think this is what @joewhitsitt was indicating in his comment above.

Correct

quamsta:
Thank you for sitting down and discussing this with me, all! On board with not calling it a "Landing Page" for sure, we originally called it a "Topic Holder" for SilverStripe, but we were using the "Holder" term in a way that was common for SilverStripe sites at the time, so I was trying to adapt the name to a different platform. 

Digging "Topic Collection" quite a bit! 

@joewhitsitt, yeah, there are quite a few repeated chunks of the same info presented differently. We haven't done any extensive user testing on it, but it was modeled after a few support websites we saw in the wild that we really liked. 

The idea is that the smaller tags near the search bar that you see first (in white text) are a bit easier to see all at once on mobile devices, the larger tag boxes stand out on tablets and larger screens, and the full listings help people get through large amounts of content simply by scrolling if they haven't found a category that makes sense for what they're looking for yet. 

Having some of these components be optional or configurable by a dev seems okay, but as-is, it's supposed to have a purposeful flow: drive people to either search for something, get to a category that they're looking for, or all else failing, see what's available in a list without leaving the page.

bspeare:
Closed in favor of https://github.com/uiowa/uiowa/pull/4060.

PR #4020, opened 2021-08-23
add tag filter to content overview, moderated content

### LINKED ISSUE #781: Content Overview Filter By Tags
With #598 and #720 we will have adopted tags on all content types so why not allow people to filter content by it to get some backend use out of it.

## Proposed Solution
Add filter by tag autocomplete on the content overview page without compromising the view to future core updates.

Screen Shot 2020-04-30 at 10 14 12 AM

joewhitsitt:
Resolves #781

All this tag talk got me thinking about the early days of d8 and #781. simple change allowing folks to find content by tag.

Original proposed solution was to programmatically add the filter to the view. I think that is moot considering we have already "forked" the view by capturing it in default config and by creating it programmatically we have a greater risk of this breaking if the filter array options change.

Screen Shot 2021-08-23 at 3 24 12 PM

### DIFF
--- a/config/default/views.view.content.yml
+++ b/config/default/views.view.content.yml
@@ -2,8 +2,11 @@ uuid: 7f810d8c-2c52-4025-9b55-cd4d9e461105
 langcode: en
 status: true
 dependencies:
+  config:
+    - taxonomy.vocabulary.tags
   module:
     - node
+    - taxonomy
     - user
 _core:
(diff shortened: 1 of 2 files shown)


dnkemper:
YAS! I am a fan and can test this!

dnkemper (approved):
Works great! Also made sure that if content has multiple tags, all tags are counted for. NOIICE!

PR #4023, opened 2021-08-24
Retain periods in MyUI urls

### LINKED ISSUE #4003: Link field with MyUI urls containing `.` are converted to `_`
# Problem
MyUI urls used within link fields have `.` converted to `_`. 

# Steps to reproduce
View the difference between the text area link and the card link field on: https://sandbox.prod.drupal.uiowa.edu/link-test

# Proposed solution
?

cory-skeers:
Resolves #4003 

URL rendering replaces periods in url query strings with underscores (https://www.drupal.org/project/drupal/issues/2984272), which breaks certain MyUI links, and others, when used in link fields (https://sandbox.prod.drupal.uiowa.edu/link-test).

# How to test

As a content editor, I can place external URLs which contain query strings with periods into my content, and my site visitors will reach this intended destination.

Can use https://myui.uiowa.edu/my-ui/courses/by-department.page?q.academicUnitId=541&showResults=1 as a test URL.

Known areas where the issue crops up: Card link field, menu links, page redirects. Likely other areas that use link fields. Does not affect links placed in WYSIWYG text fields.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -245,7 +245,7 @@
                 "Themes have no context of the entity being rendered in preprocessing a layout when using Layout builder": "https://www.drupal.org/files/issues/2020-02-11/3111192-entity-14.patch",
                 "Add Views EntityReference filter to be available for all entity reference fields": "https://www.drupal.org/files/issues/2021-06-15/2429699-414.patch",
                 "Views Block Display skips preBlockBuild() call on ajax rebuild": "https://www.drupal.org/files/issues/2021-03-05/2605218-61.patch",
-                "Periods in query strings are replaced by underscores": "https://www.drupal.org/files/issues/2021-06-16/fix-mangled-query-parameter-names-2984272-59.patch"
+                "Periods in query strings are replaced by underscores": "https://www.drupal.org/files/issues/2020-06-25/fix-mangled-query-parameter-names-2984272-36.patch"
             },
             "drupal/default_content": {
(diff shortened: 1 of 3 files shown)


joewhitsitt:
I feel like the core patch might be the better path for more coverage unless I am misunderstanding the issue.

cory-skeers:
> I feel like the core patch might be the better path for more coverage unless I am misunderstanding the issue.

🤞 for the patch. I'm trying it out locally now. It looks like it should apply smoothly, but it touches such a common method it makes me nervous.

cory-skeers:
Patch fails to apply, looks like because it's trying to edit a test that doesn't exist. Not sure if it's a Core version diff issue, or if I'm missing or not understanding how the patch should be applied.

image

joewhitsitt:
patch 59 is for 9.3. we are still 9.1.12. i wonder if an earlier patch would apply.

cory-skeers:
> patch 59 is for 9.3. we are still 9.1.12. i wonder if an earlier patch would apply.

Shoot--I'd thought I'd tried that and got the same issue, but that turned out to just be me messing up the patch path.

Yes! The next-most-recent patch does apply.

dnkemper (approved):
Tested link fields - all could think of. Nice!

cory-skeers:
Thank you @dnkemper!

PR #4026, opened 2021-08-25
Add field_person_department to genetics.grad.uiowa.edu site split

### LINKED ISSUE #4021: Add department field to genetics.grad.uiowa.edu site split
# Motivation

The program field that is added to the person content type via `grad_person_extended` is not meeting the needs of the Genetics site. They would like a field that works similar to the Department field that I added to the person in the immuno.grad.uiowa.edu split.

# Proposed solution

- Copy the related config from the immuno.grad.uiowa.edu split to the genetics.grad.uiowa.edu split. **Note: This may require some modifications to `core.entity_*_display.*` config because we are not pulling over all the fields from Immuno.
- Change the form widget for the field to not use tags, which seems to be confusing to content editors on the Immuno site. Instead use a standard multiple-entry field widget.

### LINKED ISSUE #4037: Genetics - allow advisor field to contain multiple advisors
# Motivation:
Some students on genetics.grad.uiowa.edu have multiple PhD advisors. Currently, the person content type only allows users to save one advisor in the advisor field. 

## Story
As an editor/webmaster on a site with the `grad_person_extended`, I can set multiple advisors for a person of the student person type.

# Proposed solution
* `blt ds`
* `drush @default.local uli admin/config/development/configuration/config-split`
* Enable `grad_person_extended` split.
* `drush @default.local cim`
* Visit https://default.local.drupal.uiowa.edu/admin/structure/types/manage/person/fields and edit the Advisor field.
* Change the field settings for the Advisor field to allow unlimited entries.
* Change the form settings for this field to use "Autocomplete" option (not "Autocomplete (tags)").
* `drush @default.local csex grad_person_extended`

cory-skeers:
Resolves #4021 
Resolves #4037 

# How to test

`blt ds --site=genetics.grad.uiowa.edu`
Create or edit a person
Add one or more departments
Check that you can add one or more advisors
Visit https://gradgenetics.local.drupal.uiowa.edu/people/students and verify that education and hometown columns are showing
Visit https://gradgenetics.local.drupal.uiowa.edu/people/students/former and verify that started column is showing (there is no data yet)

`blt ds --site=immuno.grad.uiowa.edu`
Create or edit a person
Check that you can add more than one department.
Check that you can add more than one advisor


cory-skeers:
@pyrello or @e-marie-w: Couple questions.

- Does the department field needs to allow for multiple values? On immuno it allows only one
- Are the departments the same where we could/should populate the Departments taxonomy, or will those be added manually after
- Should the existing Program field remain, or does it need to be removed as part of adding the department field

e-marie-w:
> Does the department field needs to allow for multiple values? On immuno it allows only one

Yes it should allow multiple values, on Genetics some faculty have 2 or even 3 departments listed
> * Are the departments the same where we could/should populate the Departments taxonomy, or will those be added manually after

I believe we were going to let them manually add departments. Sean will know for sure.
> * Should the existing Program field remain, or does it need to be removed as part of adding the department field

We won't be using the Program field so it could be removed (I think), but if it's easier or less complicated to keep it that would be fine too (I think). Again, Sean will have the official answer.

pyrello:
>> Does the department field needs to allow for multiple values? On immuno it allows only one

> Yes it should allow multiple values, on Genetics some faculty have 2 or even 3 departments listed

If we can change this for Immunology too, that would be great!

>> Are the departments the same where we could/should populate the Departments taxonomy, or will those be added manually after

> I believe we were going to let them manually add departments. Sean will know for sure.

They will add these manually.

>> Should the existing Program field remain, or does it need to be removed as part of adding the department field

> We won't be using the Program field so it could be removed (I think), but if it's easier or less complicated to keep it that would be fine too (I think). Again, Sean will have the official answer.

I think that we might be using this field for students? The program field is added as part of `grad_person_extended`, so there is no way to remove the field. If it is not being used for students, we are overriding `core.entity_*_display.node.person.default`, so we should be able to hide it on the form and the page display.

cory-skeers:
Okay! I think this is all set, including updating the immuno.grad department field to allow multiple values.

It looks like there's data in the program field on genetics.grad (27 nodes have values), so that looks like you're right in that it's being used.

e-marie-w:
> It looks like there's data in the program field on genetics.grad (27 nodes have values)

That would be from when I was trying to make the program field work for the departments. Also, just to check, the department field is available to faculty and students, correct? Some of the students do use those HR department designations instead of a program name.

cory-skeers:
> > It looks like there's data in the program field on genetics.grad (27 nodes have values)
> 
> That would be from when I was trying to make the program field work for the departments. 

Ah...that makes sense! In that case we can still hide it if the student profiles won't need it.

> Also, just to check, the department field is available to faculty and students, correct? Some of the students do use those HR department designations instead of a program name.

Yep! It's on the person and not person type, so it'll be available for both faculty and students.

pyrello:
FYI, I am going to sneak a few more changes into this PR. Just waiting on confirmation from the customer.

cory-skeers:
> FYI, I am going to sneak a few more changes into this PR. Just waiting on confirmation from the customer.

Sounds good! Let me know if you need a review or another pair of eyes for anything.

dnkemper (changes requested):
All but one thing worked: in immuno.grad.uiowa.edu, I didn't see the advisor field while adding a person. https://gradimmuno.local.drupal.uiowa.edu/node/add/person Other than that, everything looks great!

dnkemper:
Okay.. started looking into this further. For some reason, it's getting a `display: none` class applied to the field in the form display

dnkemper:
Hopefully not overstepping, but made changes to fix issue - added fields to student person type for genetics and mimicked how it was done w/ immuno. Everything seems to be working the way it should -  advisor field not being visible was the only obstacle preventing this PR from being approved.. BUT want @cory-skeers to double check before I approve. 

P.S. I apologize. When I have an epiphany in the middle of the night,  I can't fall back asleep until idea is executed/issue is fixed. So annoying. Good night.

dnkemper (approved):

PR #4040, opened 2021-08-30
Always return an array if ProcessMediaTrait::getMid query fails.

richardbporter:
Seems like code using `getMid` expects to get an array back. If the query in `getMid` fails for some reason, it returns a boolean FALSE which causes warnings like:

```
[info] Trying to access array offset on value of type bool ProcessMediaTrait.php:112
[info] Trying to access array offset on value of type bool ProcessMediaTrait.php:144
```

This PR should ensure an array with those keys gets returned. Still looking at why the query is failing and if that is ok.


cory-skeers (approved):
Reviewed, not tested. Looks like a good, quick fix! 👍

richardbporter:
Noticed another warning where `$orientation` was not set in `hook_entity_presave`. I'm assuming file orientation is not available for every file so this just wraps it in an `if` statement to check first. 

d4e94607c50a0f181f5797e198cb5752a1cb1103

PR #4075, opened 2021-09-07
Recreate existing view page as a view block that can be placed on any layout builder page. #4069

### LINKED ISSUE #4069: Recreate existing view page as a view block that can be placed on any layout builder page.
Epic: https://github.com/uiowa/uiowa/issues/3837
# Story

As an editor, I want to be able to place a view block of area of studies within a layout builder page.

# Proposed Solution

- [x] Recreate the existing areas of study view as a block display and package up in AoS feature split
- [x] Disable or remove the view page from the feature split.
- [x] Make sure the styles are scoped correctly
- [x] Make sure the exposed filters still work

dnkemper:
Closes #4069 

# How to test
1. Checkout branch
2. `blt ds --site=default`
3. `drush @default.local uli`
4. activate areas_of_study split 
5. `drush @default.local cim`
6. `blt frontend`
6. `drush @default.local cr`
7. Go to a page -> layout builder -> add 'Areas of Study' block to page
8. Make sure displays purrrrtttyyy and functions properly

### DIFF
--- a/config/admissions.uiowa.edu/views.view.areas_of_study.yml
+++ b/config/admissions.uiowa.edu/views.view.areas_of_study.yml
@@ -88,7 +88,7 @@ display:
                   is_secondary: false
                 select_all_none: false
                 select_all_none_nested: false
-                display_inline: 0
+                display_inline: false
               field_area_of_study_college_target_id:
                 plugin_id: default
                 advanced:
@@ -564,6 +564,7 @@ display:
         empty: false
         header: false
         title: false
+        css_class: false
       filter_groups:
         operator: AND
         groups:
@@ -836,6 +837,7 @@ display:
             format: filtered_html
           plugin_id: text
       title: 'Majors, Minors, and Certificates'
+      css_class: table-responsive
     cache_metadata:
       max-age: -1
       contexts:
--- a/config/features/areas_of_study/views.view.areas_of_study.yml
+++ b/config/features/areas_of_study/views.view.areas_of_study.yml
@@ -1490,6 +1490,7 @@ display:
           inline: {  }
           separator: ''
           hide_empty: false
+      block_category: 'Site custom'
     cache_metadata:
       max-age: -1
       contexts:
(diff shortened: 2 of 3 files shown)


cory-skeers:
Pulling down to test

dnkemper:
You rock, @cory-skeers 🥳

cory-skeers:
It looks like the line height is different for the summary depending on if it is using a filled-in summary field (line-height 1.15 from `body` in `global.css` or a trimmed version of the body (line-height 1.7 from `p` in `paragraph.css`.

Here "Another area" has the summary field filled in, where "Test AoS" is left blank and pulling in a trimmed version of the body: 
image

bspeare:
@cory-skeers thanks for noticing that!  Can you check again?

cory-skeers:
> @cory-skeers thanks for noticing that! Can you check again?

Checked and looks good!

cory-skeers:
If there are multiple AoS blocks on a single page, interacting with any of the search blocks leads to a progress indicator that won't go away, and the results from any search are reflected in only one of the blocks. I believe this because each form element has the same id?

It's only an issue if there's a use case for including multiple AoS blocks on one page, and I'm not sure that you'd want to for this type of block.

joewhitsitt:
>If there are multiple AoS blocks on a single page, interacting with any of the search blocks leads to a progress indicator that won't go away, and the results from any search are reflected in only one of the blocks. I believe this because each form element has the same id?

>It's only an issue if there's a use case for including multiple AoS blocks on one page, and I'm not sure that you'd want to for this type of block.

I don't think that this should penalize #4075 as the search block was not in-scope. I don't think it is major concern, but I will try to reproduce and open a separate issue

joewhitsitt:
deleted comment thinking it pertained to something else. i don't think having multiple views blocks of this type is a concern.

cory-skeers (approved):
Tested the new block and everything is functioning well and looking purrrty!

Only issue arose when adding multiple instances of the block on the same page, where the exposed filters get applied to one block, and not necessarily the block to which they're attached. As it's only an issue with multiple instances, and this isn't likely to be the case (thanks @joewhitsitt!), I think we're still good to move forward with it.

dnkemper:
thank you so much @cory-skeers @bspeare and @joewhitsitt for helping move this across the finish line!

PR #4089, opened 2021-09-09
Do not render child heading size elements for ListBlock views that use the table plugin.

### LINKED ISSUE #4088: Utilize `$has_children` parameter in `ListBlock::blockForm` for table displays.
### Problem
As noted in https://github.com/uiowa/uiowa/pull/4081#issuecomment-915450999, the child heading size for the headline field does not make sense for all view blocks. For example, a table display that does not render headings in rows doesn't really need this.

There is now an optional parameter to `HeadlineHelper::getElement`, `$has_children`, that when set to FALSE will not return the child heading size form field.

### Proposed solution
TBD

Somehow allow for this optional parameter to be set in:
https://github.com/uiowa/uiowa/blob/a2a06288d220a145dd70a924ebb10826f62e984f/docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php#L138-L152

richardbporter:
Resolves #4088 

### Stories
As a site editor placing the AOS block, I don't see a child heading size form element that doesn't do anything anyway.
As a site editor placing the People list block, I still see a child heading size form element that still works.


cory-skeers (approved):
This looks good! Works as expected.

PR #4092, opened 2021-09-09
Feature: Quick topics page fix

bspeare:
Fixing bugs that I noticed during demo

- Switched views to sort alphabetically
- Removed tags field from layout builder template and replaced with a view so we can sort alphabetically

# How to test

```
blt ds --site=sandbox.uiowa.edu
blt frontend
Enable `topic_page` split at admin/config/development/configuration/config-split
drush @sandbox.local cim
```

- Add a topic collection page: https://sandbox.local.drupal.uiowa.edu/node/add/topic_collection
- Add body, tags, and featured image

- Verify that all of the tag views are listed alphabetical no matter the weight of their listing on the node edit screen. 
- Verify that the commas are white when there is a featured image and black when there is not a featured image. 
- Verify that the "Browse by all" section does not have weird content splits if you only have one or two items tagged.

### DIFF
--- a/config/features/topic_page/core.entity_view_display.node.topic_collection.default.yml
+++ b/config/features/topic_page/core.entity_view_display.node.topic_collection.default.yml
@@ -181,8 +181,9 @@ third_party_settings:
         layout_settings:
           label: ''
           layout_builder_styles_style:
-            - section_background_style_gray
-            - section_margin_fixed_width_container
+            0: section_background_style_gray
+            1: section_margin_fixed_width_container
+            section_alignment_start: section_alignment_start
         components:
           -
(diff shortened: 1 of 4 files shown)


quamsta:
Fixed a couple of CSS issues with the comma separated tags, by "fixed," I mean I removed them, I think they'll be a bit more of a pain than they're worth. Other than that, it's looking great. Thanks Ben!

quamsta (approved):
Fixed a couple of CSS issues with the comma separated tags, by "fixed," I mean I removed them, I think they'll be a bit more of a pain than they're worth. Other than that, it's looking great. Tags are sorting alphabetically everywhere. Thanks Ben!

Issue #4082, opened 2021-09-08
Replace AoS Search Block on Admissions with uiowa_core SearchBlock


joewhitsitt:
## Issue

We made a search block in uiowa_core that does what the aos search block does on admissions. admissions is the only site using the AoS search block. Let's reduce that to one.

## Proposed solution

- [x]  Update content type description to stop referencing view page and instead mention the "Areas of Study" block.
- [x]   Now that the uiowa_core search block is available. remove the current block from admissions homepage (prod), add the uiowa_core block and configure to point to the same /areas-of-study destination.
- [x]  Find usage on the site. I think there is one tucked away somewhere. We have past update hook code that can look for instances.
- [x]  Delete area_of_study_search block content type from split
- [x]  Update permissions to not include areas_of_study_category? Taxonomy vocabulary doesn't exist?
- [ ]  Sync changes (db) to stage/dev so we don't have missing block errors.
- [x]  Remove block from areas of study module

joewhitsitt:
This can actually cause an error if placed on a non-admissions site as the hardcoded view page used to send the search does not exist in the feature split.

<img width="283" alt="AoS-heads-up" src="https://user-images.githubusercontent.com/4663676/133153645-6956f4e7-a1ec-4ebe-bd98-ae269dd59b37.png">

briand44:
This block should remain on Admissions

bspeare:
@joewhitsitt for "Sync changes (db) to stage/dev so we don't have missing block errors." before I do this, is there anything on the stage/dev that needs to be saved before they are overridden?

bspeare:
We added an update hook in https://github.com/uiowa/uiowa/pull/4139 that just removes the block from the page revisions on admissions.uiowa.edu so the block would not be available if they wanted to revert the home page to a previous revision.  We need to check with the client to see if this would be OK.

bspeare:
Client is OK with losing the block from page revisions in the update hook, so I will move forward with getting the pr reviewed.

PR #4099, opened 2021-09-09
AoS Follow-up

joewhitsitt:
- changes combined filter parameter from `combine` to `title`. this provides consistency with the search block help text and the view filter label.
- remove area_of_study_search custom block type from feature split (not used and being listed in lb_direct_add)

## To Test

- using the search block and a placed areas of study view, set the parameter to `title` and try a search on the view. the search value should be populated in the view filters.
- you won't see Areas of Study Search as a lb_direct_add option above Articles.


bspeare:
Pulling down to test

bspeare (approved):
Testing criteria all checked out!  I did notice that you can still place the "Areas of Study Search" block on a page and errors out if you try to use it because the view is not enabled `Route "view.areas_of_study.areas_of_study" does not exist`.   I know https://github.com/uiowa/uiowa/issues/4082 is the next step to fixing that.

joewhitsitt:
#4082 might have to happen sooner than I thought. Thanks for the review.

Issue #3996, opened 2021-08-17
Linkcheck .prod.drupal links


mjoneill:
As a site builder, I have visibility into how many links I have to .prod.drupal, .stage.drupal, .dev.drupal, and .local.drupal because I need to fix these prior to launching (or as soon after as possible).

Previous probe - https://github.com/uiowa/uiowa/pull/373 - jw

mjoneill:
@cory-skeers et al, it seems like you had additional thinking on this.  Also, @joewhitsitt added a related probe...

richardbporter:
Another component of this could be adding validation to try and prevent internal links being used in the first place. Attaching a filter to the Filtered HTML text format would ensure that validation is run in many different places - blocks, content types, etc. That same validation could be added to link fields as well.

cory-skeers:
Previous probe has an interesting article as part of a go-live checklist: https://kevinquillen.com/mass-updating-drupal-fields-mysql-stored-procedure-pt-2

One thing that struck me about it is that the .prod/dev/stage links post-launch is the biggest and most visible issue, but absolute links are always a sticking point for content editors, and can lead to suboptimal situations where links either break in the future, or rely on redirects where `node/#` or `media/#` links would not.

My initial thoughts went in two directions:

- A tool where editors can run a "scan" of their site, which would check text fields, menu links, etc., for any absolute links to the site (or prod/dev/stage) and give them a report of which node, where the link is, etc.
- A validator of some sort to check on save and either give a notice, fix it (if possible) and give a message, or some other sort of intervention.

I'd lean toward the second. One issue with the first is the high number of areas a user can enter a link, and it can quickly become outdated (eg a new layout builder block type gets added, and it's not added to the query). It also requires an extra step on the user to get the report and then go to fix.

The second option is more proactive, in a similar way to how https://www.drupal.org/project/editoria11y (#3396) tries to make a11y easier and more in-your-face for the editor. It wouldn't help in-progress sites as much, but any new sites (and new edits) would benefit.

richardbporter:
It looks like it is possible to search SiteImprove inventory for specific URLs: https://my2.siteimprove.com/QualityAssurance/991404/Inventory/Links/1?Query=prod.drupal

However, I don't see the Inventory tab on other sites. Not sure why that is. Also, SiteImprove respects robots.txt so we'd have to open up internal domains for scanning before launch. Just throwing this out there.

richardbporter:
[...]

Issue #2795, opened 2021-01-06
As a webmaster, I can easily navigate the admin page to successfully configure my site.


richardbporter:
# Story

As a webmaster, I can easily navigate the admin page to successfully configure my site.

# Motivation
Our custom modules that have configuration pages are named differently in the admin section. Some are prefixed with SiteNow, UIowa, University of Iowa, etc. They are also all located in the System menu section which seems wrong for things like SiteNow Articles which should probably be under Content Authoring.

<img width="966" alt="Screen Shot 2021-01-06 at 8 52 06 AM" src="https://user-images.githubusercontent.com/1068523/103782252-d5aa6900-4ffc-11eb-91a9-d91d9e733c46.png">

IMO, the average person doesn't need to see a prefix at all and Articles Configuration would be better than SiteNow/UI/Uiowa Articles.

# Proposed solution

Decide on a naming convention and change existing admin menu links for custom modules. Also place custom module menu links in the appropriate menu parent.

This issue does **not** include changing module machine names.

joewhitsitt:
Glad you brought this up. Noticed this yesterday when testing the APR module.

richardbporter:
Example menu parent for SiteNow Articles: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_articles/sitenow_articles.links.menu.yml#L3

richardbporter:
> Example menu parent for SiteNow Articles: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_articles/sitenow_articles.links.menu.yml#L3

Further clarification on this example: I would change the title to just `Articles`, the `description` to `Configuration for article content.` and the `parent` to `system.admin_config_content` so that it appears under `Content authoring` as opposed to `System`.

[
<img width="1222" alt="Screen Shot 2021-09-09 at 1 53 09 PM" src="https://user-images.githubusercontent.com/1068523/132745893-ff9a4af7-5cb6-4885-85f4-a9c0e9b2a50f.png">
](url)

richardbporter:
Also, I think this is a good dev sprint candidate. :)

e-marie-w:
- Some of the custom modules don't have a .links.menu.yml file, I assume that means they aren't a part of the menu?
- Is there a specific site that'd be good to test this on locally, or would any do?
[...]

PR #4103, opened 2021-09-10
Sitenow Search module

### LINKED ISSUE #3373: Search and replace strings of text
# Story

As a site webmaster, I would like the ability to search for a string and replace all instances of it so that I don't have to make the change in every instance it is used.

# Additional Information

We have a custom module "Find Text" in use on SiteNow v1 sites.

https://www.drupal.org/project/scanner was also mentioned as a module that has been used for this type of functionality.

Probably some conversations needed here to understand what types of fields this could work on and what wouldn't be possible as well as if this would be safe for webmasters to use or if it should be limited to administrators.

# Proposed solution

cory-skeers:
Resolves #3373 

Ability to search all text fields for a given string or regular expression, and returns those field types/content and the node on which they appear. Currently doesn't search revisions, so only checks current 'published' text fields.

# How to test

- Sync a site
- `drush @site.local uli /admin/config`
- Go to the Sitenow Find Text page
- Try searching for some stuff, see if it works!
  - Node titles and fields directly attached to nodes should be searchable.
  - Fields attached to blocks should be searchable.
  - Fields attached to paragraphs (attached to blocks) should be searchable.
    - The focus was on v3, but paragraph content on v2 should be searchable as well.
  - Content (nodes and blocks) included as feature splits (such as AoS) should be searchable as well as standard platform content.
  - Menu items/links are not searchable.
  - Reusable blocks (such as in the footer) are not searchable.
  - Contents pulled in from other sources (directory profiles, aggregator content) are not searchable.
  - Note: Some default content may not show up such as Home page blocks--this is expected behavior.
- Masquerade as a webmaster and check that the page is available and you are able to perform content searches
- Sync some sites with various splits and check that they sync correctly, the module is available, and webmaster has permissions

Scanner offers some additional options, such as matching on word boundaries, limiting to published pages only, etc. (https://www.drupal.org/project/scanner) Some of these may be useful, some not, and some may be more useful if this was opened up to content editors, but not necessary for more experienced users who may be familiar with LIKE and regex options.

Also note for a possible followup: currently doesn't search menu items, which for things like #3996, is an extra consideration.


pyrello:
Sorry to make this suggestion this far into the work, but I think this module should be called something other than `sitenow_search`. I don't have a great alternate name off the top of my head, but hopefully it could be something that differentiates it from a user-facing search. Maybe something like `sitenow_admin_search` or `sitenow_text_finder`?

cory-skeers:
> Sorry to make this suggestion this far into the work, but I think this module should be called something other than `sitenow_search`. I don't have a great alternate name off the top of my head, but hopefully it could be something that differentiates it from a user-facing search. Maybe something like `sitenow_admin_search` or `sitenow_text_finder`?

Good point. I'd lean toward `sitenow_text_finder` between the two.

cory-skeers:
Leaving this set as 'draft', but would appreciate any feedback. At minimum, would like to still update module name and associative machine names prior to merger.

bspeare:
This works great!  I tested it out and have a few observations:

- Move to System category where all the SiteNow config pages are on admin/config?
- Add real world examples for the `Basic SQL LIKE operator modifiers`?  Is there a way to change the help text so a webmaster would understand the tips here?
- Help text for `REGEXP?` checkbox? :) 
- Change `Needle` label to something like "Search Text"?
- Add help text that this only searches text areas and text area blocks?
- Since this only searches the text area markup, searching for `University Housing &amp; Dining` offers results but `University Housing & Dining` does not.

richardbporter:
>Move to System category where all the SiteNow config pages are on admin/config?

Those are being moved around in https://github.com/uiowa/uiowa/pull/4118 for https://github.com/uiowa/uiowa/issues/2795 if you have any thoughts on where stuff should go.

cory-skeers:
> This works great! I tested it out and have a few observations:
> 
> * Move to System category where all the SiteNow config pages are on admin/config?
> * Add real world examples for the `Basic SQL LIKE operator modifiers`?  Is there a way to change the help text so a webmaster would understand the tips here?
> * Help text for `REGEXP?` checkbox? :)
> * Change `Needle` label to something like "Search Text"?
> * Add help text that this only searches text areas and text area blocks?
> * Since this only searches the text area markup, searching for `University Housing &amp; Dining` offers results but `University Housing & Dining` does not.

Thanks for the notes, Rick!

Moving some stuff around, and I'll take a stab at building out the form help text for some of these.

cory-skeers:
First slice of this is ready, to be available to turn on and use as an admin for tasks like #3996.

Couple of the possible followups:

- Turn on by default and open permissions to webmasters
- Explore other search options to include (toggle case sensitivity, whole word only, allow for escaping % and _ wildcards, options for word boundaries, etc)
- Option to search revisions
   - Possibly handled in a similar manner; would be nice to get field storage table names dynamically from the field definition. I believe there's a method for this, but that I couldn't get working yet.
- Include other site areas? Menu links, other blocks, etc.
- Replace text option (likely limited to admin?) with some sort of "undo" or revision reversion
- Refactor module methods into a controller or api to make it easier to use programmatically, like in update hooks.

GaryRidgway:
Checked this real quick to make sure it worked! and it does! so cool :D

cory-skeers:
Looks like some fields are no longer working--node body and collection item text area. Needs some looking into.

cory-skeers:
Note found out during demo: It's creating a 'layout' link for all nodes, including nodes that don't allow layout, such as people...

cory-skeers:
Think I got the main outstanding issues working here--body and collection fields are now searched and reported correctly, and layout links shouldn't appear for page results that don't allow editing the layout.

docroot/modules/custom/sitenow_find_text/sitenow_find_text.module
```diff
@@ -0,0 +1,198 @@
...
+            $tables['node__body'] = [
+              'values' => ['body_value'],
+              'parent' => ['entity_id'],
+              'type' => $entity_type_id,
+              'field' => $name,
+            ];
+          }
+        }
+        break;
+    }
```
pyrello: I think this could be simplified to:

```php
$parent_id = 'entity_id';

foreach ($field_map as $name => $info) {
  $table_name = $entity_type_id . '__' . $name;
  $values = value_columns($name, $info['type']);

  switch ($entity_type_id) {
    case 'node':
      $table_name = 'node__' . $name;
      if ($name === 'title') {
        $table_name = 'node_field_data';
        $values = ['title'];
        $parent_id = 'nid';
      }
      elseif ($name === 'body') {
        $value = 'body_value';
      }
      
      break;

  }

  $tables[$table_name] [
    'values' => $values,
    'parent' => [$parent_id],              
    'type' => $entity_type_id,
    'field' => $name,
  ];
}
```

cory-skeers:
Think this is still getting caught up on how it handles searching for parents in (some) cases where we have paragraphs under blocks under nodes. Reworking on that now.

cory-skeers:
Still going forward with the restructuring of parent fetcher because it's an improvement, but the specific issue I'd been running into during testing is different.

Looks like at least some of the default content provided on sites is not in the `entity_usage` table, and so the original methods of tracking down the parents weren't working. Is only an issue with some, but not all programmatically added content (some of our update hooks have explicitly added the usage connections for block/paragraph content, and so would still be in the `entity_usage` table). May be a future slice to explore this more.

joewhitsitt:
I believe the entity_usage table is updated on save. This was semi-recently "fixed" going forward for content and the default content predates that. Not sure if entity_usage table is captured anyway for default content export.

cory-skeers:
> I believe the entity_usage table is updated on save. This was semi-recently "fixed" going forward for content and the default content predates that. Not sure if entity_usage table is captured anyway for default content export.

Ah...that makes sense, and makes me feel even better about not worrying about it. I'll add it to the testing notes.

cory-skeers:
I believe this is ready for review/testing as a first slice.

e-marie-w:
I can test this. I don't think I should be the only one to do so but it's a start and good practice for me.

e-marie-w:
Working so far, returning matching content in titles and fields but not the footer or menu links. I found the descriptions/help text very useful. I haven't tested the wildcards or REGEXP or feature splits yet, plan to this afternoon.
> Currently doesn't search revisions, so only checks current 'published' text fields.

I assume the quotes are for how results can include the most recent version of a draft page that has never been published? Because I'm getting that but it seems useful/on purpose since no revisions are showing up.

e-marie-w:
Ignore previous question about drafts, finally noticed part about 'additional options include limiting to published pages.'

Briefly tested on a v2 site, worked as it should.

I enabled AoS on a site and it removed the find text module so I just re-enabled it. Then it worked fine to find AoS title and description content. I was confused why it wasn't finding the location or degree type fields until I realized they were Taxonomy (and so not counting as Content that could be found, I assume?) 

In theory I get the difference between the regular search and when REGEXP is checked, in practice not at all sure what to test that doesn't hand me the same results for both.

cory-skeers:
Thanks for looking into this!

> I enabled AoS on a site and it removed the find text module so I just re-enabled it. Then it worked fine to find AoS title and description content. I was confused why it wasn't finding the location or degree type fields until I realized they were Taxonomy (and so not counting as Content that could be found, I assume?)

That's exactly right. Anything that's part of a taxonomy or otherwise pulled in by a reference won't get searched at this point, with a few exceptions (if it references another node, like in a featured content block then it sort of works, because it'll search the referenced node separately, if that makes sense).

Do you think this is an area where the help text could be expanded to make it more clear?

> In theory I get the difference between the regular search and when REGEXP is checked, in practice not at all sure what to test that doesn't hand me the same results for both.

For the most part, anything in a regular search should still work when REGEXP is checked. If it is checked, though, you should be able to use a regular expression like `(Herky)|(About)` and get results that match either the `Herky` or `About`, where if it's not checked it should try to search for a string that matches "(Herky)|(About)" exactly.

e-marie-w:
> If it is checked, though, you should be able to use a regular expression like `(Herky)|(About)` and get results that match either the `Herky` or `About`

Ohhh okay, that helped a lot. Once I realized I shouldn't be adding spaces it worked perfectly! 

I don't think the help text needs to be updated to include the taxonomy. If it was research areas or another tag category I think I would know to just search the taxonomy instead. I just hadn't used Areas of Study before and it took me a minute to figure it out. 

The module works great for me and has my unofficial approval. I'd just feel better if a second set of eyes more used to testing also glanced at it.

cory-skeers:
Setting to draft while making updates to turn it on and update webmaster permissions.

bspeare:
Just tested on admissions.local and transportation.local and got access denied for webmaster masquerade for /admin/config/sitenow/sitenow-find-text.    Is that expected?

bspeare:
@cory-skeers thanks, that fixed it!

bspeare:
@cory-skeers should a search for "Welcome" on the admissions site bring back results for `field_uiowa_banner_title` for their home page "Welcome to Iowa" heading?  Same question for a search for "maui", should the results cover `field_uiowa_banner_link_uri` for "https://www.maui.uiowa.edu/maui/pub/admissions/webinquiry/undergraduate.page"?

cory-skeers:
@bspeare It should...I have a suspicion about that, but will need to look into it.

bspeare:
@cory-skeers sounds good.  I also had issues searching for text within cards.   I also struggled searching for text for paragraph items on a V2 site, but I know that isn't in scope.

cory-skeers:
It looks like the banner titles aren't being searched at all...not sure why yet.

cory-skeers:
Ahhhh....the "heading" type field wasn't getting added.

cory-skeers:
@bspeare I believe I got it working with the banner titles, as well as the maui link example. That may have helped with the issues searching for text within cards as well.

Edit: Maybe not...

cory-skeers:
@bspeare I think I got those issues worked out outside the v2 stuff.

>  I also had issues searching for text within cards.

This might be related to the banner title issue. If so, this should be fixed now. If not, do you have an example I could take a look at?

> I also struggled searching for text for paragraph items on a V2 site, but I know that isn't in scope.

Yep... It should work for chunks of v2, but it might get caught up on a few things. I think the 'section' paragraphs likely get in the way.

bspeare:
Thanks @cory-skeers, I will check again in a few minutes.

bspeare:
@cory-skeers the fixes work great!  One more question, is searching the headline field that is attached to the articles and events block in scope of this feature?

cory-skeers:
> One more question, is searching the headline field that is attached to the articles and events block in scope of this feature?

Thanks for thinking of / catching these!

It should be getting those now. The result is a little less helpful (all blocks that use the headline are using the same field, so they will all show up as "field_uiowa_headline_headline"), but they'll be searched now and still provide the general results.

bspeare:
@cory-skeers Headline search is now working, thanks!

bspeare (approved):
This works great! I verified the testing scenarios and found everything that I wanted to search for.   Even found some `prod.drupal` links buried on the testing sites.

PR #4104, opened 2021-09-10
Create some sites and remove all `.gitkeep` files from existing splits.

richardbporter:
Noticed these files that got created when the empty config directories were created still exist and they are not needed.


joewhitsitt (approved):
reviewed, not tested

PR #4106, opened 2021-09-11
Add 'no results' display to RSS feeds block dropdown field

dnkemper:
putting this out here in case have time to work on it this sprint

# How to test


docroot/modules/custom/uiowa_aggregator/uiowa_aggregator.module
```diff
@@ -174,8 +174,14 @@ function _uiowa_aggregator_get_feeds() {
...
 
-  foreach ($feeds as $feed) {
-    $options[$feed->id()] = $feed->label();
+  if ($feeds) {
+    foreach ($feeds as $feed) {
+      $options[$feed->id()] = $feed->label();
+    }
+  }
+  else {
+    $options = ['no-results' => 'No feeds available'];
```
richardbporter, line 183: This would allow people to actually select `No feeds available` from the list and not render anything. I think that'd be odd and IMO the message below would be enough.

dnkemper: Noticed that ☹️ COOL about the error message though!

richardbporter:
It doesn't look like the message shows up in the right place. I think we'd have to alter the form in layout_builder_custom to get this to kinda work as expected.

However, this all would be a lot easier if uiowa_aggregator was providing this block programmatically. We didn't do that initially because we didn't have a way to use the headline field in code - now we do. I wonder if we should groom an issue to convert existing blocks to a yet-to-be created uiowa_aggregator block. This has the added benefit of moving the RSS Feeds out of the LB 'Add Component' dropdown and into the More... category.

richardbporter:
Thanks for you work on this! We talked about closing this in favor of doing it as part of #4468. The code you wrote can be used for that issue with some slight modifications.

PR #4119, opened 2021-09-14
Reduce AC log output.

richardbporter:
All the `Skipping database...` messages in AC log output are unnecessary. Let's assume they're skipped and move that to a debug message which won't print with `-v` option.

### Testing
`blt artifact:ac-hooks:post-code-update uiowa test main main foo bar --environment=$target_env -v --no-interaction -D drush.ansi=false`


pyrello (approved):
Code reviewed. Looks good. Interested in seeing this during the next deployment!

joewhitsitt (approved):

joewhitsitt:
that's right. double approval

PR #4063, opened 2021-09-01
Patch entity_usage to fix WSOD with empty data- attributes.

richardbporter:
Resolves #3167 

### Testing
To see the error on the main branch, add the markup in https://github.com/uiowa/uiowa/issues/3167#issuecomment-789729782 to a text area block in LB on a page. Save and view the page. 

Check out this branch, `composer install` and follow the steps again. No more WSOD. 

Also included in this PR is some updates to dependencies and documentation to test Drupal core and contrib projects within uiowa/uiowa. If you are interested, follow the steps in the new tests/README file to try it out: https://github.com/uiowa/uiowa/pull/4063/files#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acR21

Issue #1977, opened 2020-07-22
Standardize JavaScript and CSS implementations


richardbporter:
Related to #1151, #1402, #4036

We should follow Drupal's lead and start using ES6 syntax for all but the simplest JS. Transpiling is kind of annoying though, and I'd be interested to see if we could just write ES6 syntax without transpiling based on our browser support policy.

Working with project asset directories is confusing as described in #1151. It would be nice if every `assets` directory was ignored and we had a better build process in place to move in all assets. 

Additionally, we seem to have a lot of code duplication between Yarn workspaces. A standardized, top-level `scripts` declaration in the root package.json file would be nice instead of having to copy/paste build scripts everywhere. Another option might a root Gulp file. This may need to wait until #2629 is done.

Also, a lot of our JavaScript is implemented differently. Some is vanilla, some is jQuery, etc. At the very least, all JS should be attached with Drupal behaviors, which is not jQuery, to avoid global variable leak, handle full/partial page reloads, etc.

Also, its important to note that jQuery is loaded on every page because of the dependency specified with BigPipe which renders all page requests. We might want to decide whether or not we want to use jQuery in our scripts. I think it at least needs to be passed to the IIFE and the docs still recommend using `once`: https://www.drupal.org/docs/drupal-apis/javascript-api/javascript-api-overview  Avoiding jQuery doesn't really gain anything other than making a future removal of jQuery a bit easier, whenever that comes.

https://www.drupal.org/node/2815083
https://www.drupal.org/project/drupal/issues/3052002
https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/big_pipe/big_pipe.libraries.yml#L8

GaryRidgway:
I think we are also at a point where things between UIDS and uids_base are clashing a little bit as well. Generally in UIDS we are using vanilla JS to keep it free of dependencies, and I think this is, by virtue of being the main coding style of UIDS, bleeding into uids_base.

Additionally i completely agree with you about being more careful about polluting the global namespace, and I myself have probably been guilt of not being as careful as i could be in the past.

Honestly though I don't think jquery is much more easy to use or valuable than vanilla ES6 at this point, so i think if vanilla JS has the small benefit of being more long term useful if/when we remove jquery, then we should use that. If I had my choice I would prefer vanilla ES6 with just the jquery "once" attachment function. This uses the recommended attachment function while keeping the core code consistent across both UIDS and uids_base.

richardbporter:
[...]

PR #4122, opened 2021-09-14
Initial visual regression testing setup using Percy.

### LINKED ISSUE #3713: As a developer, I can use visual regression testing as part of my code review workflow
# Story

As a developer, I can use visual regression testing as part of my code review workflow so that fewer visual regressions happen after code deployment. 

- https://css-tricks.com/testing-for-visual-regressions-with-percy/#unlocking-the-true-power-of-percy

# Proposed solution

Use Percy's free solution (https://percy.io/pricing) and see if it helps us find visual regressions that might be missed in code reviews.

richardbporter:
Resolves #3713 

I think this might be difficult to test in this branch alone. We may just want to merge it to evaluate as part of our normal workflow. If it doesn't work out, we can revert it later.

The way I understand this working is if Percy detects a visual difference with the baseline (main), it will flag a PR as needing review. We are not requiring that someone approve the Percy build in order to merge the PR. If no difference is detected, the PR shouldn't be flagged at all. I think.

### Pseudo-testing
- If you have a BrowserStack account on our team, you should be able to log in to https://percy.io/ and navigate to the uiowa builds.
- https://docs.percy.io/docs/visual-testing-basics

### Running snapshots locally
This isn't really meant to be ran locally but...
- Run `blt tests` and ensure that it doesn't fail. There should be a notice about Percy snapshots being skipped.
- If you do want to run this locally, export the Percy token, available in project settings on percy.io, in the terminal.
  - `export PERCY_TOKEN=xxx' 
  - https://docs.percy.io/docs/local-development
  - Edit snapshots.yml to point to https://default.local.drupal.uiowa.edu instead of http://localhost:8888 which is what BLT runs in CI via Drush runserver.
  - Run `npx percy snapshot snapshots.yml`
    - Note that Percy will install a version of chromium that does not work with `blt tests:drupal`. We could do more work in our replacement command if there is more of a use-case for running this locally, like setting Percy config dynamically so changing URLs isn't necessary.


richardbporter:
I think we’ll have to move the percy command to a BLT wrapper to at least make sure it doesn’t error locally.

richardbporter:
Not sure about these CI errors.

Using snapshot.yml:
https://app.travis-ci.com/github/uiowa/uiowa/builds/237649023#L1965

Using just one URL:
https://app.travis-ci.com/github/uiowa/uiowa/builds/237652699#L1960

richardbporter:
Updated local testing instructions.

richardbporter:
Watch out if you run this locally, it might update your Chromedriver, causing core/contrib functional tests to fail. 😢

richardbporter:
I think this might be working in CI now. The issue seemed to be that it was added as dev dependency. /shrug

I'm going to leave it at the top-level since its not strictly a uids_base dependency. I mentioned in #1977 that we may want to move away from Yarn and Yarn workspaces in the future and leverage a top-level JS/Gulp script to compile everything instead. This kinda reinforces that.

richardbporter:
The dynamic events block in default content is not good for visual regression testing. Different events get flagged against the baseline.

richardbporter:
Seems like there is no approval required if Percy does not detect any visual changes.

richardbporter:
We do not have to require Percy checks be approved to merge branches and probably shouldn't considering not everyone has an account.

richardbporter:
Thought of something.

richardbporter:
I'm going to use this PR for demo and leave it in draft state for now.

richardbporter:
What browsers do we want to use for snapshots? Each browser counts as a snapshot. Safari was just added. 

https://docs.percy.io/docs/cross-browser-visual-testing

bspeare:
> What browsers do we want to use for snapshots? Each browser counts as a snapshot. Safari was just added.

If we had to choose one, I think Safari would work since that browser usually has the most bugs.

bspeare:
@richardbporter I tried out `blt tests` for the local testing and got: 
```
FAILURES!
Tests: 51, Assertions: 163, Failures: 18.
[Acquia\Blt\Robo\Tasks\PhpUnitTask]  Exit code 1 
Failed to run installer database tasks: The database server version <em class="placeholder">3.22.0</em> is less than the minimum required version <em class="placeholder">3.26</em>
```

Do I need to update anything locally to get this to work?

blt/blt.yml
```diff
@@ -28,8 +28,6 @@ disable-targets:
     toggle:
       modules: true
   tests:
-    frontend:
```
bspeare, line 31: Just curious, what does removing `frontend` from the tests do?

richardbporter: We had previously disabled that command because we weren't using it. Removing it from `disable-targets` re-enables it which we want because we use it to execute the percy commands. https://docs.acquia.com/blt/extending-blt/#disabling-a-command

richardbporter:
> @richardbporter I tried out `blt tests` for the local testing and got:
> 
> ```
> FAILURES!
> Tests: 51, Assertions: 163, Failures: 18.
> [Acquia\Blt\Robo\Tasks\PhpUnitTask]  Exit code 1 
> Failed to run installer database tasks: The database server version <em class="placeholder">3.22.0</em> is less than the minimum required version <em class="placeholder">3.26</em>
> ```
> 
> Do I need to update anything locally to get this to work?

Never seen that before. I'll look into it.

richardbporter:
I think that error is Drupal core phpunit tests failing because SQLite is not up to date. We enabled Drupal core/contrib tests in https://github.com/uiowa/uiowa/pull/4063 to make it easier to run and contribute back to d.o projects. There is a brief overview in https://github.com/uiowa/uiowa/blob/main/tests/README.md but the TLDR; would be to add the below snippet to your `blt/local.blt.yml` file:

```
# Configure Drupal tests to run.
tests:
  drupal:
    # This is required for running functional-javascript tests.
    web-driver: chromedriver
    # Use MySQL. The default sqlite DB URL that BLT configures was not working.
    simpletest-db: 'mysql://${drupal.db.username}:${drupal.db.password}@${drupal.db.host}/${drupal.db.database}'
```

That is basically switching Drupal core tests from using SQLite to MySQL which we had to do anyway for unknown reasons. If you want to test a core/contrib project later, you can adjust it based on that README.

richardbporter:
> > What browsers do we want to use for snapshots? Each browser counts as a snapshot. Safari was just added.
> 
> If we had to choose one, I think Safari would work since that browser usually has the most bugs.

We can pick and choose on the project settings page: https://percy.io/0b75ef4c/uiowa/settings

bspeare (approved):
This works great!  I was able to test the snapshots locally and might have found something that we missed in the recent deployment?  I agree with @richardbporter, I think we should merge it in to evaluate as part of our normal workflow.    

Screen Shot 2021-09-29 at 1 43 00 PM

richardbporter:
> This works great! I was able to test the snapshots locally and might have found something that we missed in the recent deployment? I agree with @richardbporter, I think we should merge it in to evaluate as part of our normal workflow.
> 
> Screen Shot 2021-09-29 at 1 43 00 PM

Whoa nice! I mean not nice...

bspeare:
Actually, I don't think that is a regression.  But that is a good example of what Percy can catch!

PR #4129, opened 2021-09-15
Initialize star.org.uiowa.edu, upsa.org.uiowa.edu and disable checks.

richardbporter:


dnkemper (approved):
After looking at build fail output, the fix looks correct

PR #4013, opened 2021-08-20
Create COVID dashboard custom block.

richardbporter:
Resolves #4005 

### Dev setup
- Block is placed on https://coronavirus.dev.drupal.uiowa.edu/covid-19-numbers

### Local setup
- Get the credentials from the [secrets file](https://docs.acquia.com/resource/secrets/#secrets-settings-php-file) on uiowa01 dev. Add them to the coronavirus site local settings file located at `docroot/sites/coronavirus.uiowa.edu/settings/local.settings.php`.
```
blt sync --site coronavirus.uiowa.edu &&
drush @coronavirus.local cim --source ../config/coronavirus.uiowa.edu --partial --yes &&
drush @coronavirus.local cim --yes
```
- Place the block on page via layout builder.

### Stories
- As a coronavirus.uiowa.edu site webmaster, I can place this block and understand how it functions.
- As a coronavirus.uiowa.edu site visitor, I can see automatic updates around 10am on M/W/F and previous reporting data on other dates.
- As a coronavirus.uiowa.edu site visitor using IE11, I can have the same experience as the story above.

PR #4132, opened 2021-09-15
add graphing solution

GaryRidgway:
As a user, I want to be able to use graphs on my Sitenow v3 site.

# How to test

1. Pull this branch and do all the stuff to get it up and running.  
2. Enable the Graph split.

2. create an excel file and add some data to it. It muse be formatted like so, and save it as a .csv file.

| dates or other incremental header  |  dataset1 | dataset2  |
|---|---|---|
| Date1  | foo1  | bar1  |
| Date2  | foo2  | bar2  |
| ... | ...  |  ... |

You can have as many columns or rows as you would like, i just put in this amount for example purposes.

[Here is an example `.csv` for you to use](https://sitenow.uiowa.edu/sites/sitenow.uiowa.edu/files/2021-09/airtravel.csv)

3. open the excel file in a text editor and copy the plain text.
4.  add a "Graph" block to a page in layout builder
5. Give a summary of the data in the summary field.
5. paste the csv text in to the plaintext field
6. save
7. View that your first and second columns of data (that is not dates or whatever you used in column 0) are black and gold, and that any subsequent columns of data are randomly generated colors.
8. Inspect the canvas element that the Graph is drawn in and view that it has `role="img"` and `aria-labeledby="SOME-GRAPH-NAME"`
9. Verify that you can click on `Show tabulated data` and it will show you the table of the data that made the graph, and that the caption is what was entered in the summary field.
10. Verify that the summary field has the  same Id that was listed in the `aria-labeledby` on the canvas.
8. profit.


docroot/modules/custom/uiowa_core/src/Plugin/Block/GraphBlock.php
```diff
@@ -0,0 +1,115 @@
...
+      else {
+        $build['graph_container']['graph_table']['#rows']['row-' . $row_key] = [];
+        foreach (explode( ',', $row) as $column_key => $column) {
+          array_push($build['graph_container']['graph_table']['#rows']['row-' . $row_key], ['data' => $column]);
+        }
+      }
+    }
+
+
+//    $build['#header'] = [
```
joewhitsitt: Should all of commented out code be removed?

GaryRidgway: yeah i just cleaned it up

joewhitsitt:
captured from standup:
- trim csv input
- color control?
- aria-describedby link to table so screen reader knows table represents graph
- further a11y review before merge
- link the example csv to the "properly formatted table" help text

docroot/modules/custom/uiowa_core/uiowa_core.libraries.yml
```diff
@@ -2,6 +2,21 @@ gtag:
...
   js:
     js/tag-manager.js: {}
+graph:
+  version: 1.x
+  js:
+    js/csv_graph.js: {}
+  dependencies:
+    - core/jquery
+    - core/jquery.once
+    - core/drupalSettings
```
richardbporter: I'd be curious if adding a dependency here on `uiowa_core/chartjs` solved the loading issue.

GaryRidgway: like, depending on another library in the libraries file?

GaryRidgway: testing now

richardbporter: Yea IDK if you can depend on a library that you define in your own library file. But I assume Drupal would handle loading these accordingly if so.

GaryRidgway: it did not

docroot/modules/custom/uiowa_graph/js/csv_graph.js
```diff
@@ -0,0 +1,165 @@
+(($, Drupal, drupalSettings) => {
```
richardbporter: I missed our last IE11 conversation - are we still needing to support that? I don't think arrow functions work in IE11: https://caniuse.com/?search=arrow%20functions

Setting up a workspace to transpile this would be overkill though. I just ended up dumbing down my JS in https://github.com/uiowa/uiowa/pull/4013/files#diff-d42454e37e99d62e631edf86d0812c25abdbd610e6c382e13173a0822355deb5.

GaryRidgway: this is a totally valid concern. I think we should use the older version just in case because of IE11 concerns. I can make that change

cory-skeers: I think re: IE11 we're unofficially in the "as long as it doesn't break, it's fine regardless of how it looks" realm.

joewhitsitt: >as long as it doesn't break, it's fine regardless of how it looks

can we add that to our browser support policy? https://sitenow.uiowa.edu/policies

docroot/modules/custom/uiowa_graph/uiowa_graph.info.yml
```diff
@@ -0,0 +1,6 @@
+name: uiowa_graph
+type: module
+description: A simple graphing module for uiowa websites.
+package: Custom
+core: 8.x
```
richardbporter: I think `core: 8.x` can be removed.

docroot/modules/custom/uiowa_graph/js/csv_graph.js
GaryRidgway: It wasn't that bad to remove them, so no worries. Non-arrow functions are now in place.

richardbporter:
> captured from standup:
> 
> * trim csv input

Is trimming happening? I copied air travel and it had two blank lines which rendered.

richardbporter:
One concern with this is that most people will open CSV files in Excel automatically and not really understand how to copy/paste the source.

GaryRidgway:
> > captured from standup:
> > 
> > * trim csv input
> 
> Is trimming happening? I copied air travel and it had two blank lines which rendered.

Are you saying it completely rendered fine? I think I actually fixed the problem that was occurring that we "Thought" was being caused by not trimming the csv.

GaryRidgway:
> One concern with this is that most people will open CSV files in Excel automatically and not really understand how to copy/paste the source.

the other option is we could offer a file upload which would just parse something as a csv and then dump it in to the text field? I think with this slice the intent was to offer something simple to implement, which the text field solves.

richardbporter:
I gotta run but in gener

> > One concern with this is that most people will open CSV files in Excel automatically and not really understand how to copy/paste the source.
> 
> the other option is we could offer a file upload which would just parse something as a csv and then dump it in to the text field? I think with this slice the intent was to offer something simple to implement, which the text field solves.

yea thats fine for now

richardbporter:
Screen Shot 2021-09-22 at 4 13 21 PM

that CSV produces

image

GaryRidgway:
> Screen Shot 2021-09-22 at 4 13 21 PM
> 
> that CSV produces
> 
> image

I see, then yes we should trim those newlines.

richardbporter (approved):
I'm ok with this for a first slice to deliver for coronvavirus, gauge interest from other sites and learn. However, if we are getting lots of requests to enable this and for different graph types, we should reconsider a contrib project, unless those were already evaluated.

I might not understand the trimming thing. :)

GaryRidgway:
> I'm ok with this for a first slice to deliver for coronvavirus, gauge interest from other sites and learn. However, if we are getting lots of requests to enable this and for different graph types, we should reconsider a contrib project, unless those were already evaluated.
> 
> I might not understand the trimming thing. :)

trim has been added :D

richardbporter:
One sec here.

richardbporter (changes requested):
ChartsJS >3 does not support IE11. https://github.com/chartjs/Chart.js/issues/8042#issuecomment-856875306

The chart doesn't render at all in IE11:

Screen Shot 2021-09-23 at 8 18 49 AM

Screen Shot 2021-09-23 at 8 19 05 AM

Is that a requirement @mjoneill? Would displaying the table in IE11 instead be sufficient?

richardbporter (approved):
Data is there for IE11.

Issue #4149, opened 2021-09-20
Hover, Focus effects for Pagination on background colors


bewoldt:
# Problem
When the pagination is displayed on different backgrounds some of the hover effects are not visible or the color contrast is not enough.

# Examples
https://sandbox.prod.drupal.uiowa.edu/pagination

bspeare:
@bewoldt thanks for finding this!  You could make the updates for this at https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/pagination.scss.

joewhitsitt:
Closed with merge of https://github.com/uiowa/uiowa/pull/4166

PR #4166, opened 2021-09-21
updates to pagination with different backgrounds

bewoldt:
Resolves issue #4149


richardbporter (commented):
Tested with `blt frontend && blt sync --site sandbox.uiowa.edu`. Looking at https://sandbox.local.drupal.uiowa.edu/pagination. It is easier to read those numbers against the patterned background. Active page is more visible. 👍🏻 

The only one I was unsure of was the gray-on-gold. It's not as subtle as the others but IDK what if other color combos would be possible there. Thoughts on that?

image

GaryRidgway:
I think the Grey looks fine, I could see if they were white that they might blend in to the lines of the texture, so i think the grey is a nice offset. Looks good to me!

bspeare:
I think I'd rather see the pattern pager background remain similar to the solid background colors.   Can we just set the background for the pager to match the pattern background instead?  I think in the screenshots below the pager remains visible but doesn't draw all of your attention to the boxes. 

Screen Shot 2021-09-22 at 3 10 00 PM
  
Screen Shot 2021-09-22 at 3 10 37 PM

Screen Shot 2021-09-22 at 3 09 32 PM

GaryRidgway:
@bspeare If we are trying to draw less attention, then yes i think this is a good way to go.

bewoldt:
I made an update to reflect the comments made to make the link backgrounds the same as the section background color.

bspeare (approved):
Looks like a great improvement, thanks @bewoldt!

PR #4177, opened 2021-09-22
update for narrow container with hidden sidebar

### LINKED ISSUE #4080: Narrow container option doesn't work after removing sidenav
# Problem
If you have a page that has sub-pages and you check the option to Remove sidebar regions to hide the side-nav, when you change the layout of the content container to use the Narrow option the css styles get overridden.

Example page:
https://mentalhealth.uiowa.edu/recovery
This page has sub-pages, is set to use the Narrow content container width, and has the side-nav turned off.

bewoldt:
Resolves #4080


bspeare (approved):
Tested these will all variations of the layout w/sidebar and it is working.  Thanks for the bug fix @bewoldt!

Issue #4158, opened 2021-09-21
Decrease technical debt of ListBlock class


pyrello:
# Motivation

The ListBlock class was created originally as a fork of the `ctools_views` [Block](https://git.drupalcode.org/project/ctools/-/blob/8.x-3.x/modules/ctools_views/src/Plugin/Display/Block.php) class. At the time this approach seemed the best because we could make it work exactly the way we wanted. The downside was always that we were duplicating a lot of code that is already maintained in the contrib space. However, as we have learned, these classes are pretty hard to maintain and so we would like to reduce our footprint in that regard by shifting to a model where we extend the Block class instead of recreating it.

Getting this to a more stable state will allow us to continue with converting other views blocks to using this paradigm.

# Proposed solution

As was started in #3829, switch the ListBlock class to just extend `Drupal\ctools_views\Plugin\Display\Block`. This will require a lot of code clean-up and removal in the ListBlock class. All that we need to do in the ListBlock class is to apply changes to the existing logic that we want to apply to all of our views block displays.

pyrello:
Related to #3501

makurj:
I would like to pair on this.

joewhitsitt:
ditto

joewhitsitt:
ListBlock zoom huddle summary. We started combing line by line through ListBlock.php to remove anything that ctools_views is already providing. @makurj is going to push this initial work.

joewhitsitt:
i tried to move forward with https://github.com/uiowa/uiowa/pull/4198 but hope that @pyrello will have a chance periodically to check in on this.

noticing some UX setbacks but so far things are still rendering...

pyrello:
> i tried to move forward with #4198 but hope that @pyrello will have a chance periodically to check in on this.
>
> noticing some UX setbacks but so far things are still rendering...

So far this looks good. If you are running into specific issues, I'm happy to step through them with you.

joewhitsitt:
I feel like we haven't done that much gutting yet but switching from `Drupal\views\Plugin\views\display\Block` to `Drupal\ctools_views\Plugin\Display\Block` really changes things...

<img width="451" alt="Screen Shot 2021-09-24 at 4 41 05 PM" src="https://user-images.githubusercontent.com/4663676/134742747-65b28c96-d5aa-4463-8927-87a5df198ddf.png">

pyrello:
[...]

PR #4198, opened 2021-09-24
ListBlock Updates

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/4158

Created draft PR from Makur's branch

## To Test

- Interact with existing views blocks. Check for visual regressions from that of PROD (e.g. filters being revealed that weren't previously revealed).
- Create new people list blocks in v3 and make sure the settings provided are what are rendered.


pyrello:
Taking a look.

pyrello:
I did some refactoring with an eye towards making it easier to compare the `ListBlock` class and its new parent. Also made some changes to align with expected changes when we adopt a patch to allow exposing filters.

Also ended up creating a helper function for running update hooks to update a batch of blocks of a certain plugin id, but didn't end up needing it yet. I left it in because I think it will be helpful in the future.

@joewhitsitt Might be good for you to re-check this to make sure I didn't break anything. Everything looked good to me when I tested your code.

pyrello:
I'm giving preliminary approval of this based on the checking I did. Once @joewhitsitt has satisfied himself about how it works, then I'll give actual approval.

joewhitsitt:
So, it seems that https://github.com/uiowa/uiowa/pull/4198/files#diff-80bddf0efcb844ce671e57c638bfaa246f2fa0d0c37b1689cfcdc03657b3e565R975 will be a config update for all views blocks (article block, etc.), but one that doesn't need to happen before deployment? Is that correct @pyrello ?

pyrello:
It seems like views don't update their config definition until they are re-saved. At least, after adding that option, none of the other views were updated. I think that not having that item in there will cause them to continue to function they way they currently do. If we wanted to be as thorough as possible though, we would re-save every block and then re-export. Or if it seemed like something that could be handled via file comparison and copy/paste that would probably do the trick as well.

joewhitsitt:
from my tests, other view blocks were not affected beyond the new config but that the new config didn't prevent the new/existing blocks to work.

dnkemper (approved):
Nervous to approve this but tested it following the testing instructions and everything seems to work as expected

joewhitsitt:
Based on:

>I'm giving preliminary approval of this based on the checking I did. Once @joewhitsitt has satisfied himself about how it works, then I'll give actual approval.

>Nervous to approve this but tested it following the testing instructions and everything seems to work as expected

and my own testing, I am going to merge this.

pyrello:
Woo-hoo! Thanks for taking this on, @joewhitsitt. I think this work will make the future work in #3804 easier.

Issue #2848, opened 2021-01-12
Duplicate page or select template functionality


joewhitsitt:
# Story

As a webmaster, I want to quickly recreate certain layout pages on my site without configuring every section manually.

# Proposed solution

Some further discussion is most likely needed and a research spike into which solution will work best for our setup, but this has been repeatedly requested by customers.

## Potential Solutions
- [x] Investigate Entity Clone 
  - https://github.com/uiowa/uiowa/issues/129
  - https://www.drupal.org/project/quick_node_clone
  - https://www.drupal.org/project/entity_clone/issues/3060223
- [x] Investigate Replicate module (It works)
  - https://www.drupal.org/project/replicate
  - https://www.drupal.org/project/replicate_ui
- [ ] Investigate layout_library
  - https://github.com/uiowa/uiowa/issues/2378
  - https://www.drupal.org/project/layout_library
- [ ] Investigate Section Template library picker.
  - https://www.drupal.org/project/section_library
- [ ] Investigate Page template module
  - https://www.drupal.org/project/pate
    - https://www.drupal.org/project/pate/issues/3238285

bspeare:
The layout builder library is intriguing with this description "Site admins can prevent content-editors from creating their own bespoke layouts but give them the ability to customise layouts on a per-content-item basis."  I think this would be an ideal solution to have in place to limit content editors for creating layout and component combinations.

pyrello:
> The layout builder library is intriguing with this description "Site admins can prevent content-editors from creating their own bespoke layouts but give them the ability to customise layouts on a per-content-item basis." I think this would be an ideal solution to have in place to limit content editors for creating layout and component combinations.

@bspeare Are you talking about for content types where we don't allow for per item customization? Or pages? Or both?

I would have concerns about shutting down people's ability to use LB to create their own layouts with the page content type. With the patches to core and this module however, we could use multiple layouts as a starting point to providing better starting point suggestions for users.

One downside to the layout library module is that its current UI is a set of checkboxes on the node edit screen. For that to work, we would probably need to incorporate some icons (which I think would require some customizations on top of the module) and/or really good textual descriptions of the layouts themselves.

pyrello:
For the clone options, we would need to see how they work with our nested items, such as Accordions.

[...]

Issue #129, opened 2019-01-25
Add Entity Clone


joewhitsitt:
We have requests from current customers to allow the duplication of a page. I think [Entity Clone](https://www.drupal.org/project/issues/entity_clone) does this and supports entity reference revisions (paragraphs and nesting).

We could use this to create default unpublished templates for users!

joewhitsitt:
Particularly this issue, the ability to set permissions on a entity type per entity type basis: https://www.drupal.org/project/entity_clone/issues/2931143

joewhitsitt:
Good thing we were not early adopters. There are serveral issues with paragraphs currently.
> Bumping to critical, in our project we are extensively using (nested) Paragraphs, and this bug causes errors left and right.

richardbporter:
Will we still need this after #1779? I don't think the use case will be as strong since a default layout can be set for all pages.

pyrello:
We are still looking for a good solution to allow for different types of page templates, so if this might fill that role, I'd be interested in checking it out. I talked about this a little bit in this comment: https://github.com/uiowa/uiowa/issues/1550#issuecomment-647015619

joewhitsitt:
Another issue that might affect our solution (slider/collection paragraphs within a block within a page) -  https://www.drupal.org/project/entity_clone/issues/3050027

Might also be worth giving a look: https://www.drupal.org/project/quick_node_clone (mentioned in the above issue)

briand44:
closing this in favor of #4223

PR #4214, opened 2021-09-28
Try Replicate and Replicate UI modules

### LINKED ISSUE #4223: Ability to Duplicate Pages
# Story

As a content editor/webmaster, I would like to be able to duplicate an entire page so that I don't have to recreate it from scratch to have a consistent look.

# Additional Info

Replicate and Replicate UI modules have been discussed as a potential way forward.

Some additional information and conversations have been captured https://github.com/uiowa/uiowa/issues/2848.

Some more thoughts on possible issues with replicate https://github.com/uiowa/uiowa/pull/4214#issuecomment-929709156

# Proposed solution

Use replicate, replicate_ui, [this patch](https://www.drupal.org/project/replicate/issues/3100763), and possible a custom event listener to provide this functionality.

If entity_clone is used in place of replicate, there is a [patch here](https://www.drupal.org/project/entity_clone/issues/3050027) that is aimed at getting paragraphs nested under blocks working (may also be useful in setting up any necessary event subscribers for replicate).

pyrello:
Related to #2848 
Related to #4223 

# How to test

- Checkout branch and composer install
- Sync a site and uli
- From a node page or article, or `admin/content`, choose to "replicate" -- enter a new title
- Check that the page has been successfully replicated
  - All content should be present on the new page
  - New blocks should be created
  - New paragraphs nested under blocks should be created
  - Section components should have new UUIDs
  - New copies of referenced nodes (article authors, for instance), media, taxonomy terms, etc. should not be created, but the references should still be there
  - Check the revision overview tab--there should be a single revision, with a message that it was replicated from another node (with a link to that node).
- Only Page and Article content types should be replicable. For other content types, you should not have the option to replicate.
  - If you manually enter "node/#/replicate" for a non-replicable page, you should not be allowed to replicate, and should see a message stating this
- Publisher and webmaster should have replicate permissions. Related config splits that needed updating can be checked from https://github.com/uiowa/uiowa/search?p=1&q=filename%3Auser.role.webmaster or https://github.com/uiowa/uiowa/search?q=filename%3Auser.role.publisher
- Sitenow V2 sites should not have the option to replicate


pyrello:
I was curious about these modules after hearing about them in our meeting today, so I decided to put together a quick PR to try them out.

Out of the box, replicate doesn't appear to handle Layout Builder blocks appropriately. It seems to just copy the `layout_builder__layout` field, which makes it look like the page is duplicated, but when you dig into it, it seems that it is still pointing to the existing blocks. I think the ideal behavior here is that all new blocks would be created to avoid unforeseen complications with a single block splitting its revisions between two different referencing layout fields on two different nodes. I'm not sure what the problems would be, but it is not "correct" in how it should work.

I tried adding a patch from https://www.drupal.org/project/replicate/issues/3100763, which definitely seems like it should be working, but a preliminary test indicated that it was not. I wasn't able to investigate it anymore than that. 

Maybe if @cory-skeers has time to take a look he can validate what I am seeing or not.

cory-skeers:
> I tried adding a patch from https://www.drupal.org/project/replicate/issues/3100763, which definitely seems like it should be working, but a preliminary test indicated that it was not. I wasn't able to investigate it anymore than that.
> 
> Maybe if @cory-skeers has time to take a look he can validate what I am seeing or not.

Interesting...I'm planning on checking this out this morning, and see how it goes.

pyrello:
The other thing to check out that I haven't looked into yet is whether it appropriately handles nested paragraph in LB blocks. We would want the nested paragraphs to also be created as duplicates. I have a feeling that the patch isn't going to handle that.

However, I would say that if the patch works, then it also points towards a potential solution where we could create our own event listeners to trigger the duplication of those paragraph items.

cory-skeers:
> The other thing to check out that I haven't looked into yet is whether it appropriately handles nested paragraph in LB blocks. We would want the nested paragraphs to also be created as duplicates. I have a feeling that the patch isn't going to handle that.

Yeah, looks like the patch wouldn't handle that were it working. Can confirm that the patch does not seem to change the behavior, and the same block references are being used rather than new blocks being created. Looks like a good solution for it, though, if we could get it worked out and then extend it to work with nested paragraphs as well.

cory-skeers:
> Yeah, looks like the patch wouldn't handle that were it working. Can confirm that the patch does not seem to change the behavior, and the same block references are being used rather than new blocks being created. Looks like a good solution for it, though, if we could get it worked out and then extend it to work with nested paragraphs as well.

Tested on a different page, and the patch did work for basic blocks. In addition to updating it to replicate paragraphs, would probably want to update it to replicate our ListBlocks as well.

cory-skeers:
Hijacked this PR to test out entity_clone in place of replicate--can move to a separate one if needed.

Entity Clone with this [patch](https://www.drupal.org/project/entity_clone/issues/3050027) appears to be working right out of the box. Needs a more thorough testing (in general, and particularly with views-based blocks), but initial testing seems good.

joewhitsitt:
> Hijacked this PR to test out entity_clone in place of replicate--can move to a separate one if needed.
> 
> Entity Clone with this [patch](https://www.drupal.org/project/entity_clone/issues/3050027) appears to be working right out of the box. Needs a more thorough testing (in general, and particularly with views-based blocks), but initial testing seems good.

If so, I'd be interested in how this module has matured since I reviewed it way back when (permissions, etc.). https://github.com/uiowa/uiowa/issues/129

cory-skeers:
`entity_clone` is cloning too much...including media entities, article authors, etc. Some of this may be config options we need to update.

Edit: Updating config appears to have fixed it for some things such as media and taxonomy terms, but not others, such as article authors.

cory-skeers:
When replicate works, it does appear to work completely, creating new uuids for components and properly setting them.

Known issues so far:

- [ ] Does not work for view blocks (people currently, but articles/etc to come)
- [ ] Does not create new paragraphs (slides and collection items), though it does properly create a new block

Still need to test with more block types, and see if it is only the above cases, or if there are other issues. Initial patch testing  hadn't worked thoroughly, so there's still some uncertainty. Can confirm that it worked for Text Area, Collection (minus the paragraph Collection Items), and Articles blocks.

It does also properly update the Entity Usage table, with or without an additional save trigger.

TL;DR
Replicate needs more testing to find what other blocks may not work with the patch, and any others along with views blocks and paragraphs can be handled by an event subscriber.

Onward once again.

cory-skeers:
This is ready to be put through its paces in testing.

We'd want to update site splits (core.extension and permissions) prior to merging, but wanted to get it into testing and confirmed as a path forward prior to putting in that effort.

joewhitsitt:
>This is ready to be put through its paces in testing.

going to poke at this

joewhitsitt:
Sorry not getting very far..

```PHP Fatal error:  Uncaught Error: Class 'Drupal\replicate\Events\ReplicatorEvents' not found in /var/www/uiowa/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php:63```

**Comment by cory-skeers** [comment]
> Sorry not getting very far..
> 
> `PHP Fatal error: Uncaught Error: Class 'Drupal\replicate\Events\ReplicatorEvents' not found in /var/www/uiowa/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php:63`

Think this is an error because it's trying to load the subscriber before replicate gets installed. ReplicatorEvents is only used for a reference constant, so I hardcoded it as a temporary workaround so the functionality can be tested.

**Comment by joewhitsitt** [comment]
I am impressed so far. Some questions/comments:

- move local task before delete task and assign an icon
- yay! usage is up to date!
- Past revisions are "replicated" too. Is that necessary? I would imagine it would be one draft revision with updated author, datetime. It would be nice if the revision message included a /node/id reference/link to the original.
- This is turned on for all content entity types (default page, article, person) which can include custom types like Brand's lockup. On the surface, I don't think that would be a good thing to replicate. Could we layer in additional control to only allow the defaults unless a developer has included additional types?
- should we limit this to just publishers/webmasters to reduce usage and prevent abuse (db rows grow quickly with this)?
- i tried replicating https://sandbox.local.drupal.uiowa.edu/people-directory and the first two blocks were not created. not sure if this is a just me issue or not. (Pager issue when set to 1 and Former Staff)
- i would think this awesome enhancement would only be turned on for v3 correct? revise my brand lockup comment above to a v3 custom content type (admissions aos?)
- looks like the module needs co-maintainers
- resaved an accordion post-replicate and it seemed to break the markup id.
Screen Shot 2021-10-06 at 4 06 17 PM

**Comment by cory-skeers** [comment]
> Some questions/comments:...

Thanks Joe! Going to take a look into these

**Comment by GaryRidgway** [comment]
Perhaps this getting to dev to be vetted is not a bad idea?

**Comment by cory-skeers** [comment]
Set it back to draft as some further work/investigation gets completed. Still open for testing if anyone is curious and wanting to get more eyes on it.

**Comment by cory-skeers** [comment]
Moving into a checklist to investigate:

- [x] move local task before delete task and assign an icon
- [x] yay! usage is up to date!
- [ ] Past revisions are "replicated" too. Is that necessary? I would imagine it would be one draft revision with updated author, datetime.
- [ ] It would be nice if the revision message included a /node/id reference/link to the original.
- [ ] This is turned on for all content entity types (default page, article, person) which can include custom types like Brand's lockup. On the surface, I don't think that would be a good thing to replicate. Could we layer in additional control to only allow the defaults unless a developer has included additional types?
- [x] should we limit this to just publishers/webmasters to reduce usage and prevent abuse (db rows grow quickly with this)?
- [x] i tried replicating https://sandbox.local.drupal.uiowa.edu/people-directory and the first two blocks appear at the end of the section, rather than the start (Pager issue when set to 1 and Former Staff)
- [ ] i would think this awesome enhancement would only be turned on for v3 correct? revise my brand lockup comment above to a v3 custom content type (admissions aos?)
- [ ] looks like the module needs co-maintainers
- [ ] resaved an accordion post-replicate and it seemed to break the markup id.
- [ ] Config needs to be updated across affected splits (don't add to v2)

**Comment by cory-skeers** [comment]
> resaved an accordion post-replicate and it seemed to break the markup id.

Tested creating a new accordion on a page, replicating the page, editing the accordion, and saving the page; the markup id appeared correct. I might be missing something, it might be a more specific bug, or it could have been unintentionally fixed by another change.

**Comment by joewhitsitt** [comment]
so you got a number appended to `accordion-item--`? if so, fair enough. just curious

**Comment by cory-skeers** [comment]
> so you got a number appended to `accordion-item--`? if so, fair enough. just curious

Yup! (in my particular case, 271 in the original, 277 in the copy).

**Comment by cory-skeers** [comment]
> * Past revisions are "replicated" too. Is that necessary? I would imagine it would be one draft revision with updated author, datetime.
>  It would be nice if the revision message included a /node/id reference/link to the original.

Did some bit of cleanup with the revisions as part of the eventSubscriber (remove all revisions on the clone except for the revision).

Unless I'm mistaken, it looks like the eventSubscriber doesn't have any info on the original node's ID, so there might be a little more involved there in order to craft a full log message. Also not sure why, but the created revision is not showing up in the revision overview tab. It's in the `node_revision` table with log message, created time, etc., but not listed in the overview.

**Comment by cory-skeers** [comment]
> * should we limit this to just publishers/webmasters to reduce usage and prevent abuse (db rows grow quickly with this)?

That had been my thought. It seems to me like more of a "power user" type feature. I've updated the default permissions to reflect that (before being pushed out, these permissions and other config items need to be updated across splits).

**Comment by cory-skeers** [comment]
> * [ ]  i would think this awesome enhancement would only be turned on for v3 correct? revise my brand lockup comment above to a v3 custom content type (admissions aos?)

It seems to work surprisingly well for v2, but I'd worry about truly supporting it. I think it'd be between having it available but not officially supported/guaranteed on v2, or unavailable. Probably the latter, since I don't think we could even say that it won't WSOD under certain circumstances, only that I didn't run into that in my limited testing.

Re: the custom content types, do we want to restrict them from being replicated, if we limit permissions to publishers and webmasters? I think something like the brand lockup wouldn't want to be replicated, but content types like AoS or the Topics Landing pages I'm less sure about either way.

**Comment by cory-skeers** [comment]
> not sure why, but the created revision is not showing up in the revision overview tab. It's in the node_revision table with log message, created time, etc., but not listed in the overview.

I'm not sure if this is an issue with how I set the revision message and such, or if there's something else going on.

The primary chunk of code in question:

```
$entity->setNewRevision(TRUE);
$entity->setRevisionLogMessage('Replicated node ' . $this->getClonedNid());
$entity->setRevisionUserId($this->currentUser->id());
$entity->setRevisionCreationTime($_SERVER['REQUEST_TIME']);
$entity->save();
```

There is other code after this that removes old revisions, but it shouldn't effect this new revision (and testing without that code leads to the same results).

joewhitsitt:
As it relates to v2:
- Gives another sites another reason to switch to v3
- There is an entityreference bug that is more common in v2 where an existing reference no longer exists and causes save issues (temp unavailable). This is more so in v2 because of the nested collapsed paragraphs. Manually you have to inspect each paragraph item to remove the broken reference (recently happened to techforum) Not sure what happens in that situation with replicate, but reducing risk might be worth it.
- Less to maintain if it is as simple as split core extension and permissions.

I think there are use-cases to replicate other content types, but that it adds complexity to our service when the usage isn't necessarily there. I honestly don't think person content will be replicated. If we can implement some kind of opt-in functionality initially without too much trouble (UiowaCoreAccess control on the local menu task/routes) we can always take it down later.

cory-skeers:
Makes sense to me re: v2. Should be easy enough to keep it separate in the config split. `layout_builder_custom` holds the new subscriber (the event is also currently hardcoded to avoid issues with the order of installation), so there shouldn't be any issue there.

> If we can implement some kind of opt-in functionality initially without too much trouble (UiowaCoreAccess control on the local menu task/routes) we can always take it down later.

That sounds like a good plan. Local tasks and direct paths (/node/#/replicate) to the form, check the node content type, and hide/restrict if it's not a Basic Page?

joewhitsitt:
taking a look to see how i can help

joewhitsitt:
Found some issues related to revisions that might be causing the problem. Originally, I turned off NRD and Diff to get closer to core functions.
https://www.drupal.org/project/drupal/issues/2916172
https://www.drupal.org/project/drupal/issues/2769741
https://www.drupal.org/project/drupal/issues/2761409

I probably don't understand the long term consequences of this, but by setting, https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityBase.php/function/ContentEntityBase%3A%3AsetRevisionTranslationAffected/8.9.x, this seems to be resolved.

joewhitsitt:
I could not seem to get xdebug to break on alterRoutes to see what I could do there, but added ignored config to allow certain types to be replicated. I think I covered all the possible webmaster action points. modeled after the protected blocks (footer contact).

cory-skeers:
> I probably don't understand the long term consequences of this, but by setting, https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityBase.php/function/ContentEntityBase%3A%3AsetRevisionTranslationAffected/8.9.x, this seems to be resolved.

I don't either, but it works!

> I could not seem to get xdebug to break on alterRoutes to see what I could do there, but added ignored config to allow certain types to be replicated. I think I covered all the possible webmaster action points. modeled after the protected blocks (footer contact).

I like this approach. Seems easy enough to extend later if we need/want to, and gives a route for users to be nudged into contacting us if there are legitimate use cases.

joewhitsitt:
> gives a route for users who have the right permissions and can guess the path to be nudged into contacting us...

there i fixed it for ya.

cory-skeers:
Looks like users are still able to access "replicate" in the VBO operations (I think that's what controls it? The dropdown operations on `/admin/content`).

cory-skeers:
We're using https://git.drupalcode.org/project/layout_builder_operation_link to add links to layout builder in the VBO operations--might be a resource as a way to remove the option there.

joewhitsitt:
Looks like hook_entity_operation_alter should allow you to remove the option based on the same entity type check. https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_entity_operation_alter/9.2.x

cory-skeers:
Re: v2--

Rather than splitting core.extension (I'd thought it was already split), I made a small change to the update hook to, if the site is v2, simply not allow any content types for replication. It adds one more thing that we for sure will need to update as part of the transition from v2 to v3, but means one less split for testing/updating between now and then.

@joewhitsitt any issues with that approach?

joewhitsitt:
percy seems to fail on the default homepage youtube video, so i don't think it is related.

dnkemper:
I can test this!

dnkemper:
- [x] Check that the page has been successfully replicated
- [x] All content should be present on the new page
- [x] New blocks should be created
- [x] New paragraphs nested under blocks should be created
- [x] Section components should have new UUIDs
- [x] New copies of referenced nodes (article authors, for instance), media, taxonomy terms, etc. should not be created, but the references should still be there
- [x] Check the revision overview tab--there should be a single revision, with a message that it was replicated from another node (with a link to that node).
- [x] Only Page and Article content types should be replicable. For other content types, you should not have the option to replicate.
- [x] If you manually enter "node/#/replicate" for a non-replicable page, you should not be allowed to replicate, and should see a message stating this
- [x] Publisher and webmaster should have replicate permissions. Related config splits that needed updating can be checked from https://github.com/uiowa/uiowa/search?p=1&q=filename%3Auser.role.webmaster or https://github.com/uiowa/uiowa/search?q=filename%3Auser.role.publisher
- [x] Sitenow V2 sites should not have the option to replicate

dnkemper (changes requested):
Remove custom v2 site functionality then good to go!

dnkemper (approved):
JK! Only perms are there but no content type is replicable. Good to go. NICE WORK! Sorry about confusion!

PR #4215, opened 2021-09-28
Hotfix - Updated composer lock file and removed layout_library

dnkemper:
Make sure what's commented is removed https://github.com/uiowa/uiowa/blob/794f2dba43d198171a7414d7d15ec22621b5efc7/composer.lock#L7137-L7196

# How to test

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -7135,66 +7135,6 @@
             }
         },
         {
-            "name": "drupal/layout_library",
-            "version": "1.0.0-beta2",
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Code diff looks good, also pulled and ran a composer install and nothing unexpected: just removed layout_library as it should 👍

dnkemper:
Thank you so much, @cory-skeers 🥳

PR #4222, opened 2021-09-29
Skip percy snapshots on develop branch.

richardbporter:
Realized after merging #4122 that we don't want to auto-approve develop snapshots because that would mess with the baseline. I don't think we want to get notified of diffs on develop either because it could be unstable. This PR changes the snapshot to only happen when the TRAVIS_BRANCH is not develop.

### Testing
See that snapshots were skipped in develop: https://app.travis-ci.com/github/uiowa/uiowa/builds/238782507#L2051

See that snapshots were not skipped in this PR: https://app.travis-ci.com/github/uiowa/uiowa/builds/238781910#L2048


richardbporter:
I've updated the Percy project settings to remove `develop` from auto-approval.

pyrello (approved):
Looks good.

Issue #3951, opened 2021-08-12
Some sites are missing option to add the People block in LB


joewhitsitt:
We suspect that several sites (grad, pharmacy, transportation) did not get the allowed block permission when the people block was refactored due to split page default config.

## Proposed Solution

- Update sites that have split page.default to be up to date with people block and any other missing additions.

PR #4286, opened 2021-10-08
Transition v3 article blocks to views-based solution

### LINKED ISSUE #3272: As a site editor, I can hide the date on the articles teaser list
# Story

As a site editor, I can hide the date on an articles teaser list so that older dates at the top of the teaser list are not visible. 

# Proposed solution

- Replace articles list with exposed filters available in https://github.com/uiowa/uiowa/pull/2702?

### LINKED ISSUE #3501: Replace article custom block with views-based alternative
# Story

As a site owner I want to use a views block with new functionality from #2702 to build my article lists so that I can make use of the enhanced functionality.

# Reference

https://github.com/uiowa/uiowa/pull/2702 and related follow-up PRs (hotfixes, orphans)
https://github.com/uiowa/uiowa/issues/4267

# Proposed solution

- [x] Create the Article list view block similar to the People list view block and export config.
- [x] Provide layout builder style options related to card (background, image position, image format, content alignment, image size, card border, grid options)
- [x] Default styles for new instances should be `white, left align, images are right align wide small`
- [x] Add `sitenow_articles.path` token
- [x] Update layout_builder_custom to have the same show/hide options functionality as aggregator.
- [x] Style adjustments and update templates to receive block style configuration.
- [x] Update hook to replace existing article blocks with the views-based alternative.
- [x] Replace preprocess_block logic in SiteNow Articles related to article block.
- [x] Check and remove orphans?
- [x] Remove Article block content type.
- [x] Remove fields that are unique to Article block content type.

## Relates to
https://github.com/uiowa/uiowa/issues/2351
https://github.com/uiowa/uiowa/issues/3489
https://github.com/uiowa/uiowa/issues/3272

### LINKED ISSUE #3678: Default content on V3 home page install needs updating
# Problem

- The page is using outdated class names that we have deprecated 
- Events, Articles, and Banners need some updates as well.

# Proposed Solution

- Update all class names to use updated UIDS class names instead of legacy
- Update Events and Articles to use built-in headline field
- https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/content/block_content/582328d0-e1f0-4de6-843d-e2372c91e6e6.yml needs the layout_builder_style_headline_size = headline_medium

### LINKED ISSUE #4357: Unlabelled URLs alter size of person card
# Story #
User noticed unlabelled website links cause the person card to become bigger and not match person cards with labelled URLs. 

They were only labeling if there was more than one website link, so if this is expected behavior they suggested maybe changing the helper text.

Label:
labelled link
with title
No label/plain URL:
unlabeled link
no title

e-marie-w:
Resolves https://github.com/uiowa/uiowa/issues/3501
Resolves https://github.com/uiowa/uiowa/issues/3272
Resolves #3678
Resolves https://github.com/uiowa/uiowa/issues/4357 

# How to test

- [x] `blt drupal:install --site default` to see the new front page. no default content issues, articles list block is on the front page.
  - Approve new baseline for Percy
- [x] `blt dsa` a lot of sites to test the update hook against. no update/import issues.
- [x] compare existing articles block with the post-update replacement views blocks
  - hide images, descriptions, details should carry over. 
- [x] create a new articles views block, configure it how you will, get the expected result even when paging.
  - All the expected styles a available and working
  - More link goes to SiteNow Articles view page, custom path works if provided.
  - all styles should be "default" to start
  - changing image type, makes size options toggle
- [x] see some that fields located on the views config are not available to editors to toggle visibility
- [x] direct from source link functionality works even when source link is hidden.
- [x] organization is wrapped in source link if source link has value and source link is hidden.
- [x] source link is hidden by default for all new and existing
- [x] check to see if replicate is working for the new articles block

## ToDo
- [x] Investigate https://github.com/uiowa/uiowa/pull/4286#issuecomment-948901829
- [x] Update splits to make sure new block is available and old block is not.
- [x] Investigate empty headlines
- [x] WSOD https://github.com/uiowa/uiowa/pull/4286#issuecomment-949009898

# Additional information

- Due to the update hook, this should be run "off-hours," so merging/deployment should be coordinated.
- Website documentation needs to be updated.


joewhitsitt:
going to poke at this to get more familiar of where it is at

joewhitsitt:
looks like `sitenow_articles_preprocess_views_view_unformatted()` works but a closer eye on variable names required by docroot/themes/custom/uids_base/templates/uids/card.html.twig might need attention.

bspeare:
Here are a few items left to do to get the new template working:
  - [x] Get link to use path alias instead of node ID https://github.com/uiowa/uiowa/blob/5fdd59d5f6cecaccd5726c5c69ebff3dc25d606d/docroot/modules/custom/sitenow_articles/sitenow_articles.module#L323
  - [x] Render `field_article_source_org` as linked up to source link if "Link directly to source" is unchecked

e-marie-w:
On a page with only one article block it seems to be converted okay, but I get an error at the top of the page that starts with:
`Notice: Undefined index: field_article_source_link in Drupal\views\Plugin\views\style\StylePluginBase->getFieldValue() (line 814 of core/modules/views/src/Plugin/views/style/StylePluginBase.php).` 
On another page I maybe added a few too many article blocks, tells me website encountered an unexpected error and gives  me error that starts with:
`Error: Call to a member function getValue() on null in Drupal\views\Plugin\views\style\StylePluginBase->getFieldValue() (line 814 of core/modules/views/src/Plugin/views/style/StylePluginBase.php).`
Sooo at least it's the same line?

e-marie-w:
I've been testing the org and direct link stuff today. With and without an org the direct and non-direct links work. Still getting that undefined index error though. Sometimes it's on the page with the article block and sometimes on the actual articles. The only other thing I changed was which articles had authors so I'm going to test that some more.

joewhitsitt:
Can you parse the undefined index back to a particular line number?

joewhitsitt:
looking into this. seems to be caused by the `$source_link = $style->getFieldValue($id, 'field_article_source_link');` in `sitenow_articles_preprocess_views_view_unformatted`

e-marie-w:
> looking into this. seems to be caused by the `$source_link = $style->getFieldValue($id, 'field_article_source_link');` in `sitenow_articles_preprocess_views_view_unformatted`

Thanks for figuring out where it's coming from, @joewhitsitt! I was struggling.

joewhitsitt:
Bad news is that we can't rely on `$style->getFieldValue()` when a field is hidden from the view. Good news is that we are going to move a lot of those out to handle hidden fields.

e-marie-w:
> we can't rely on `$style->getFieldValue()` when a field is hidden from the view. 

I just discovered that setting hide styles on an article block before converting it to the view block seems to be the thing that completely breaks a page for me, is this why/related?

Update: `Hide descriptions` converts correctly. `Hide image` does not convert/take effect, but it also does not break the page. `Hide details` is the one that completely breaks the page (spits out `The website encountered an unexpected error. Please try again later. Error: Call to a member function getValue() on null...`

joewhitsitt:
I probably made @bspeare unhappy by borking the twig display and haven't done anything to limit what fields can be shown/hidden BUT I think I fixed the immediate errors and accounted for any field to be shown/hidden.

trying to avoid the preprocess_field with this because other views/displays rely on those.

e-marie-w:
I'll test more thoroughly after lunch but so far this fixed those pages that wouldn't show up at all! 

Update: I'm guessing the missing display options are related to the twig stuff, that's the only thing so far.

docroot/themes/custom/uids_base/templates/views/views-view-unformatted--list-article.html.twig
```diff
@@ -0,0 +1,53 @@
...
+      {% endif %}
+    {% endblock %}
+
+    {% block card_date %}
+      {% if card_date|striptags|trim is not empty %}
+        {{ card_date }}
+      {% endif %}
+    {% endblock %}
+    {% block card_tag %}
+      {% if card_tag.org|striptags|trim is not empty %}
```
bspeare: Will this solution bork an opportunity to use a single template for future list block templates?

joewhitsitt: At the moment yes, but I can try to get the render done in the preprocess again

joewhitsitt:
There is an issue with the tempstore portion of the update hook. Discarding changes post-update the updated block reveals itself.

https://github.com/uiowa/uiowa/blob/2facaf8cd8e5c189f04febf533e2d7cc0701aa1d/docroot/profiles/custom/sitenow/sitenow.install#L101

docroot/themes/custom/uids_base/templates/views/views-view-unformatted--list-article.html.twig
joewhitsitt: future single template back in play

joewhitsitt:
Thanks @cory-skeers, 6b148c6 seems to work for me.

joewhitsitt:
So, previously several fields were hidden using the details style option. I have changed the logic to link org if source_link is hidden but populated. But there will be instances where both org and source_link will show post-update. There is no way to know if one should be hidden over the other for a list.

It might be worth it to encourage hiding the source_link for new list blocks. More often than not source link is filled out to either be a direct link or to be part of the org.

joewhitsitt:
~~I don't think we can remove the block entirely quite yet. Scope would have to increase to also do https://github.com/uiowa/uiowa/issues/3678~~

Ah, crap, but we don't want a new instance out there to deal with... We **need** to do #3678 as part of this...

joewhitsitt:
I think we should default hide the source link for both the update hook and a new block. I think in most cases this will be desired. I will personally apologize to the few sites that display the source link and notice the change.

Source link is still used to wrap organization (if org exists and is not hidden) and is still used to link the card if direct is checked.

e-marie-w:
Time for fresh eyes I think, took me way too long to realize I needed to sync again to see the source link hidden change. Hiding the posted date as part of 'Hide details' was a planned change, correct?

Otherwise, only thing I found is an existing 'View More' custom path changed to the default path after the update hook.

joewhitsitt:
>Hiding the posted date as part of 'Hide details' was a planned change, correct.

No it wasn't. Good catch. Posted date can now be hidden, but existing blocks should not be hiding those.

>Otherwise, only thing I found is an existing 'View More' custom path changed to the default path after the update hook.

Ruh roh. Do you remember where?

e-marie-w:
> > Otherwise, only thing I found is an existing 'View More' custom path changed to the default path after the update hook.
> 
> Ruh roh. Do you remember where?

[https://sandbox.prod.drupal.uiowa.edu/article-lists](https://sandbox.prod.drupal.uiowa.edu/article-lists) I set it to go to the About page, just to test it. And it defaulted to the regular News page path.

joewhitsitt:
pushed fixes for both @e-marie-w

e-marie-w:
Now working as expected!!

richardbporter:
Dumb question: coronavirus.uiowa.edu homepage news block more link says "View more news" and after the update it says "View more" - is that expected?

Before:
Screen Shot 2021-10-20 at 2 44 53 PM

After:
Screen Shot 2021-10-20 at 2 45 00 PM

joewhitsitt:
>is that expected?

Not a dumb question. Good eye. This is now the same as the people block "view more" and will be customizable in the future: https://github.com/uiowa/uiowa/issues/4305

Since folks are renaming news to be other things (articles/updates/etc) they haven't been able to use the more link because it still says news.

We will be taking a bit of an a11y hit with this in the short-term but hopefully folks have a heading wrapping this or nearby to describe what the user might get.

richardbporter:
Thanks.

On https://genetics.grad.uiowa.edu/ homepage, the article body (or summary) doesn't seem to be showing anymore:
Screen Shot 2021-10-20 at 2 49 19 PM

The Hide fields: Article summary isn't checked though...

joewhitsitt:
D'oh! That is because of `grad_custom_preprocess_node()` where for a teaser view mode it is getting the trimmed/summary to display, but since this views block moved from view mode to fields, this won't work except on SiteNow Articles view page, etc.

@pyrello what do you think we should do for this custom stuff?

richardbporter:
> see some that fields located on the views config are not available to editors to toggle visibility

I'm not sure what this one means.

richardbporter:
>check to see if replicate is working for the new articles block

Not entirely sure how to do that. Might leave that to someone else.

I'm going to simulate a code deploy locally to better understand what to expect.

joewhitsitt:
> > see some that fields located on the views config are not available to editors to toggle visibility
> 
> I'm not sure what this one means.

title, content id are fields on people and articles views block that cannot be toggled by editors to hide. direct link to source is also included from articles block. https://github.com/uiowa/uiowa/pull/4286/commits/1028efa736a08853fcd16ec9e6fd36c8392600e2 https://github.com/uiowa/uiowa/pull/4286/commits/b06a426851571c470c388e10d48fee366bdcbc11

richardbporter:
>I'm going to simulate a code deploy locally to better understand what to expect. 

Looks like the block just doesn't render until the update hook runs. No WSOD though.

pyrello:
> D'oh! That is because of `grad_custom_preprocess_node()` where for a teaser view mode it is getting the trimmed/summary to display, but since this views block moved from view mode to fields, this won't work except on SiteNow Articles view page, etc.
>
> @pyrello what do you think we should do for this custom stuff?

It is not clear to me from looking at the code in that preprocess function why that would be resulting in what @richardbporter is seeing above. It is checking if the teaser is empty and then overriding the value of the `field_teaser` field with a smart trimmed snippet from the body field. If anything, the problem would be that the teaser would show regardless of setting it to hidden, not that it wouldn't show.

joewhitsitt:
It is preprocess_node (teaser view mode) which I don't think gets called anymore for the field-based views block.

This could move to preprocess_field and load the entity to get the other field values before adjusting field_teaser. Not sure how aware a field is of it's location (route) though.

richardbporter:
Basic replicate functionality seemed to work from a novice perspective. 

>compare existing articles block with the post-update replacement views blocks

Didn't check this box until grad_custom is figured out. Happy to help with that if needed.

pyrello:
@joewhitsitt is right that the preprocess hook isn't being called anymore. I'm inclined to just add article summaries on the articles that are showing now and instruct those site owners they need to do that going forward.

IMO this is a deficiency in the system that requires this field to be specifically filled out when the information can be generated and pulled from the body text, but that is a conversation for a different issue.

richardbporter:
Ok. Nothing else stood out to me.

>Due to the update hook, this should be run "off-hours," so merging/deployment should be coordinated.

+1 to this. There will be quite a delay where article blocks aren't showing.

@bspeare is the Percy snapshot as expected?

joewhitsitt:
> > Due to the update hook, this should be run "off-hours," so merging/deployment should be coordinated.
> 
> +1 to this. There will be quite a delay where article blocks aren't showing.

Switching to draft to prevent merge until we can clear the pipeline and schedule once approved.

pyrello:
I have added article summaries to the articles on the 3 sites I believe this would impact, so we are good to go from my end of things.

bspeare:
Tried pulling this down locally and syncing sandbox and got this error:
```
  sitenow   9023        hook_update_n   9023 - Convert uiowa_articles to  
                                        article_list_block.               
 --------- ----------- --------------- ---------------------------------- 

>  [notice] Update started: sitenow_update_9023
>  [error]  Call to a member function get() on string 
>  [error]  Update failed: sitenow_update_9023 
>  [notice] Message: Blocks of inline_block:uiowa_articles records found: 27
> 
>  [notice] Message: Blocks of inline_block:uiowa_articles records found: 152
> 
>  [notice] Message: Tempstore records for blocks of inline_block:uiowa_articles records found: 4
> 
 [error]  Update aborted by: sitenow_update_9023 
```

bspeare:
Screen Shot 2021-10-21 at 1 47 22 PM

Does this need to be added to admissions split? I'm not seeing the option here.

pyrello:
> Does this need to be added to admissions split? I'm not seeing the option here.

Yes, any instances of `core.entity_view_display.node.page.default` will need to be updated to add the block. Yet another example of why getting the CS2 stuff working would be great!

joewhitsitt:
i thought "lists" was an established category like "site custom"

pyrello:
> i thought "lists" was an established category like "site custom"

I'm not sure how this is even working for people blocks. I would expect "Lists" to show up in this section: 
https://github.com/uiowa/uiowa/blob/696cde6f206d95425b8da5aaab6a4cac8e31833e/config/default/core.entity_view_display.node.page.default.yml#L181-L198

joewhitsitt:
Added todo to investigate sandbox sync and lb page setting splits

bspeare:
Screen Shot 2021-10-21 at 3 39 11 PM

Ran this on pharmacy and it looks like it is outputting an empty headline field.

bspeare:
Also getting WSOD when trying to save the layout after adding an articles block to any page on Obermann and Pharmacy.  Log errors say:

Screen Shot 2021-10-21 at 4 16 32 PM

joewhitsitt:
:drupal_cry:

joewhitsitt:
Wow, nice testing Ben. Slowly working through all of these finds.

The headline has been out there since at least the ctools_views refactor. I think I fixed it for articles and people blocks.

>Tried pulling this down locally and syncing sandbox and got this error

Re-saved https://sandbox.prod.drupal.uiowa.edu/article-lists and it synced. This "solution" makes me very nervous.

joewhitsitt:
So, I suspect the WSOD is caused by too many items and not the warnings/notices from Ben's comment. I was able to break it on pharmacy using none, 0 (display all) and no pager.

Changed things to be default/migrated min 1, max of 50, default of 10. This matches the current block: admin/structure/block/block-content/manage/uiowa_articles/fields/block_content.uiowa_articles.field_collection_results

Looked into some of those other notices/warnings. Fixed a few. The remaining seem to be upstream in ctools_views and relate to sort plugin info, etc. Doesn't seem to affect sort order set by the user.

joewhitsitt:
Thank you @cory-skeers for that awesome find!

e1f4de0 solves the sync error from https://github.com/uiowa/uiowa/pull/4286#issuecomment-948901829

e-marie-w:
From what I saw checking the files & later read in #3951, it seems like Admissions, Grad, and Transportation are the only site splits that need the People List block specifically added. 

Will test more in-depth Monday. So far: Transportation works fine, Admissions adds new blocks fine but didn't convert the old article blocks correctly for me, and I haven't tried Grad yet.

joewhitsitt:
Though I think https://github.com/uiowa/uiowa/pull/4286/commits/0b09e68a74bc51b4956855443072e46f7e789b7b will work fine, I wonder why these are specific and not open like default. Is there a particular view in this category on these sites that they shouldn't have access to?

e-marie-w:
Grad also works fine, and I _think_ Admissions does. I thought the article block on the homepage wasn't converting correctly (it turned to just text areas and cards) but it looks like it's set up as text areas and cards on the production site so there is nothing to convert. I'm not sure if the existing article block is used anywhere on the site, is there a way to check? The changes worked for the other two so I assume we're good regardless.

cory-skeers:
> I'm not sure if the existing article block is used anywhere on the site, is there a way to check?

A quick way I use to check things like that is a database query. Can use `drush @site.env sqlc` to enter a command line interface for the local database, and then something like `SELECT count(*) FROM block_content WHERE type="uiowa_articles";` Gets a little more involved to find out _where_ any blocks are from here, or if they're actually on pages or only in old revisions, but it can at least give a yes or no if there are blocks, and generally how many may need to be checked.

bspeare:
Admissions is using the block on https://admissions.uiowa.edu/counselors-educators.

e-marie-w:
> Admissions is using the block on https://admissions.uiowa.edu/counselors-educators.

Oh awesome, thank you @bspeare! It works on this page so I am more confident in saying it does work on Admissions.

& I know how to check for any other instances now too thanks @cory-skeers!

joewhitsitt:
To prevent merge, I am going to switch back to draft, but this PR is reviewable again.

>Due to the update hook, this should be run "off-hours," so merging/deployment should be coordinated.

  >+1 to this. There will be quite a delay where article blocks aren't showing.

bspeare:
Screen Shot 2021-10-25 at 12 34 02 PM

I think we might want to set the default article image size to medium?  Small makes sense as a default for People lists because those images are square, but widescreen small is very tiny.

bspeare:
Also, is the source link supposed to display?  If you look at the two screenshots above the "Test" article should be displaying "uiowa.edu" as a link in the teaser.

e-marie-w:
> Also, is the source link supposed to display? If you look at the two screenshots above the "Test" article should be displaying "uiowa.edu" as a link in the teaser.

Source links not wrapped around an org name are hidden by default now, don't remember why just that it was a new change we made. Users can uncheck the box if they do want it displaying on the teaser still. 

I don't have a strong opinion on the image thing. Medium does look better when it's a two column layout but I'm not sure about one column. Then again, I usually see article blocks in the two column layout.

joewhitsitt:
> Also, is the source link supposed to display? If you look at the two screenshots above the "Test" article should be displaying "uiowa.edu" as a link in the teaser.

With the visibility of each field now possible, the functionality has changed slightly.

https://github.com/uiowa/uiowa/pull/4286#issuecomment-947671550

bspeare:
> I don't have a strong opinion on the image thing. Medium does look better when it's a two column layout but I'm not sure about one column. Then again, I usually see article blocks in the two column layout.

We had custom CSS in the previous articles block to make it look closer to `medium`, so I would vote for changing the default to `medium` instead of `small`.

bspeare (approved):
I've run through all the testing scenarios and blt'd a lot of sites and everything appears to be working.  Nice work!

joewhitsitt:
https://git.uiowa.edu/its-web/data.its.uiowa.edu/-/merge_requests/28

quamsta:
Don't want to hold this up, but can a future iteration/slice of the article lists have some kind of indicator that there's something clickable to see more of an article? Examples: underlined link in the headline, a visual only "Read more/ Continue reading" button, ellipsis (…) or some kind of visual way to show that there's more to an article or that it's just a preview?

joewhitsitt:
@quamsta is that captured by https://github.com/uiowa/uids/issues/460 ?

quamsta:
> @quamsta is that captured by [uiowa/uids#460](https://github.com/uiowa/uids/issues/460) ?

It's in the same neighborhood as that! Although I would assume a hover/focus state would only be applicable to desktops/mouses and not non-mobile/touch devices. 

What I'm going for is a way to universally say "there's more to this story/article" without the user having to focus on or hover over any single teaser. Could be as simple as an ellipsis. I can capture these thoughts in a separate issue if that's more appropriate

bspeare:
@quamsta Yes, please capture that in a separate issue :).

PR #4308, opened 2021-10-13
Replicate followup

### LINKED ISSUE #4223: Ability to Duplicate Pages

cory-skeers:
Resolves #4223 

Re-enables replicate functionality for pages, articles, and resolves WSOD issue with paragraph replicants.

# How to test

Specifically, there's a WSOD issue under certain circumstances.

# Steps to reproduce:

- Replicate a page which uses either sliders or accordions (blocks w/ nested paragraphs)
- Publish the replicant page
- Delete the replicant page
- Attempt to view the original page, and get a WSOD

If the updates here work as intended, the above should be able to be completed without resulting in a WSOD.

## Original testing instructions

Below is the original list of functionality to be tested, and what should be preserved along with solving the WSOD issue.

- Checkout branch and composer install
- Sync a site and uli
- From a node page or article, or admin/content, choose to "replicate" -- enter a new title
- Check that the page has been successfully replicated
- All content should be present on the new page
- New blocks should be created
- New paragraphs nested under blocks should be created
- Section components should have new UUIDs
- New copies of referenced nodes (article authors, for instance), media, taxonomy terms, etc. should not be created, but the references should still be there
- Check the revision overview tab--there should be a single revision, with a message that it was replicated from another node (with a link to that node).
- Only Page and Article content types should be replicable. For other content types, you should not have the option to replicate.
- If you manually enter "node/#/replicate" for a non-replicable page, you should not be allowed to replicate, and should see a message stating this
- Publisher and webmaster should have replicate permissions. Related config splits that needed updating can be checked from https://github.com/uiowa/uiowa/search?p=1&q=filename%3Auser.role.webmaster or https://github.com/uiowa/uiowa/search?q=filename%3Auser.role.publisher
- Sitenow V2 sites should not have the option to replicate

# Followup tasks to create/update issues when this is merged

 * Remove temporary workaround for replicator::constant error
 * Update v2->v3 transition steps to include updating uiowa_core.replicate_allowed in uiowa_core.settings (see update hook 9020)


docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -3539,3 +3539,30 @@ function sitenow_update_9021() {
     ->set('uiowa_core.replicate_allowed', [])
     ->save();
 }
+
+/**
+ * Re-activate replicate functionality for v3 sites.
```
joewhitsitt, line 3544: You could simplify this if it is the same.

```php
function sitenow_update_9022() {
  sitenow_update_9020();
}
```

cory-skeers:
My general understanding here is that the original replicate process prior to our EventSubscriber is creating at least a partial connection from the original paragraph to the replicant node's block, so that when the replicant node is deleted, so are its blocks, and then the blocks' paragraphs, including the original's paragraphs.

I think it's only partial, because we aren't seeing doubled-up paragraphs on the replicant, and are still seeing the paragraphs on the original. The only part that appears to not work as intended is how usage is being tracked upon deletion.

docroot/profiles/custom/sitenow/sitenow.install
cory-skeers: Good call! Assuming we won't have transitioned all v2 sites by the time this rolls through, it'll be exactly the same. And if we did transition before then...that's either great for us on the v2->v3 move or wow...this issue was a doozy.

cory-skeers:
> I think it's only partial, because we aren't seeing doubled-up paragraphs on the replicant, and are still seeing the paragraphs on the original. The only part that appears to not work as intended is how usage is being tracked upon deletion.

Paragraph's parent entity is being updated to the newly created block on the replicant, but all target_ids are correct.

So we either need to prevent the paragraph from being added/saved to the newly created replicant block, or re-set the original parent as part of our eventSubscriber.

cory-skeers:
Added a comment to the patch issue (https://www.drupal.org/project/replicate/issues/3100763#comment-14256939)

cory-skeers:
Included an updated patch that creates duplicates of entity_reference_revisions as well in the original listener patch.
https://www.drupal.org/project/replicate/issues/3100763#comment-14257205

cory-skeers:
Want to look into where other issues may be at with replicator, and if there might be current in-progress work on the entity references where something like this may be better suited, or if it'd be better to attach a listener after the block duplication.

cory-skeers:
Got a better understanding of the previous patch's "Create a duplicate of the inline block. For now we cannot use the Replicator service. The method "cloneByEntityRevisionId" has to be added." comment, and instead patched it to get around the issue and use the original Replicator service.

https://www.drupal.org/project/replicate/issues/3100763#comment-14257379

Still a bit more to look into, as well as want to get an MR going, but I think we're headed in the right direction.

cory-skeers:
Currently getting a "cannot load entity with NULL id" error when the replicant is attempted to be saved. The null id is to be expected and understood, but not why it's being thrown when attempting to save.

joewhitsitt:
plan on looking at this after standup

joewhitsitt (approved):
I tested the follow-up scenario as well as the general functionality again. Almost lost track of all the replicants i created/destroyed/altered/re-replicated. Also tried to cause chaos by removing/modifying new/replicated accordions.

Everything so far has held up.

PR #4319, opened 2021-10-15
Feature: Updated UIDS version hash

bspeare:
This pr updates UIDS with some list-item updates that allows better styles for nested lists:

Screen Shot 2021-10-15 at 11 42 00 AM

# How to test

Code review should be sufficient.


pyrello (approved):
Reviewed. Not totally clear on why there are new dependencies showing up in the lock file, but it may be related to upstream dependencies changing. Not too worried about it.

PR #3524, opened 2021-05-13
Added icons to local tasks tabs, added weight and UIDS red color to "Delete" tab

quamsta:
Resolves: https://github.com/uiowa/uiowa/issues/3508

This is my first attempt at doing anything Drupal admin styling/hook-related, so go easy on me! Feedback would be super-appreciated. Willing to work through this with anyone who wants to help

<details>
Preview:

Screen Shot 2021-05-13 at 12 26 42 PM

I wasn't sure how to safely add any icons to the tabs in the "Classy" admin theme due to its tabs already having ::before pseudo elements, so the admin menu only changes the "Delete" tab to the UIDS red.

Screen Shot 2021-05-13 at 12 26 10 PM
</details>

# How to test

`blt ds --site=sandbox.uiowa.edu`
`yarn workspace uids_base gulp --development`

After clearing your cache, you should see icons in the main editor that looks like https://github.com/uiowa/uiowa/issues/3508#issuecomment-854094772.  But like I said, once you click into "Edit" or other admin areas, the icons do go away. Would appreciate some guidance on adding those for consistency's sake.

Thanks!

PR #4321, opened 2021-10-15
Initial pre-footer setup.

### LINKED ISSUE #4296: Build solution for designable pre-footer region
# Story

As a site owner, I would like to be able to add content to a pre-footer region that is persistent from page to page.

# Proposed solution

* Build a custom content entity type for region areas. Possibly with a bundle for region? Investigate whether we might want to use this module: https://www.drupal.org/project/storage or Fragments. Whichever entity type gets used, it won't have any fields, but will have Layout Builder enabled.
* Set the Layout Builder Restrictions for the new entity to only allow one, two, and three column layouts. Only allow text area and CTA blocks to be placed.
* Create a way to display an entity in a region, either via a custom block type, or using a Views block display.

# Question
* Do we want to allow different options for how the pre-footer area looks? If so, how do we provide those options?

<details>
 <summary>Original issue</summary>
Original title: Add Give to Iowa block to pre-footer region

Screen Shot 2020-07-08 at 5 14 22 PM
Looking to add a Give to [Iowa] block in the pre-footer region so that it is visible on all pages and is directly adjacent to the footer region (no padding or margin).

The above screenshot shows an idea for this block, similar to a CTA block, with minor tweaks:

- Much smaller padding/margin on top and bottom compared to CTA block. (changed from 1.875rem to 0.875rem in this example)
- Button link is on the left side of the text.
- Text size of excerpt is much smaller than standard text. (used font-size:smaller in this example)
- No longer 50/50 flex but more like a 33/66 layout, so that there is much less padding/margin around button link.
</details>

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/4296

## Story
As a Webmaster, I can create `region_items` with layout builder and put them in a pre-footer section. I can create and manage the layout of these `region_items` from the "Region settings" page.

## How to test
1. Sync down any site.
2. find the "Region settings" in the "Sitenow" configuration section.
3. Make some `region_item`s 
4. edit their layouts
5. add one to the pre footer area.

The above steps should be enough, but they are also not detailed because UX is important on this, and it should feel relatively intuitive to make these. please list anything you feel becomes clunky or unintuitive, and of course any bugs you find, thanks :D


pyrello:
I removed the `uiowa_maui` module from the Thesis Defense split because it is not relevant there at all.

pyrello:
I changed the category for the `PreFooterRegionBlock` to "Restricted" and added that category as an empty allow list to the Region item fragment layout builder restrictions as well as the same settings for the page node.

joewhitsitt:
i want to comment before it gets to late. this work is currently spread across layout_builder_custom, uiowa_core, uids_base. is is possible/worth it to wrangle this in?

GaryRidgway:
> i want to comment before it gets to late. this work is currently spread across layout_builder_custom, uiowa_core, uids_base. is is possible/worth it to wrangle this in?

With a quick look over the code my instinct is that all the code feels like it is in the right place to me. What would be your suggestion on where to move it?

joewhitsitt:
unless this is supposed to be theme dependent, i think the uids_base logic and template could be called from uiowa_core. if it is supposed to be theme dependent, the stuff in uiowa_core should probably move and be theme settings.

uids_base and other things got big/bloated in the early days. we haven't revisited it but i think a lot could be moved elsewhere. just trying to make sure we don't create a bigger problem for down the road.

pyrello:
At this point it is pretty standard for our module features to have CSS/templates in `uids_base`. I agree that this is not necessarily the most ideal scenario, but I think sorting it out is more complicated than just moving the files back to the module of origin. I think the ideal best practice would be to have basic templates and CSS with the feature so that they are not unstyled outside the context of `uids_base`, but that the styling that is relevant to the theme should actually be applied by the theme.

GaryRidgway:
It has been suggested to have a per-page disable option for page pre-footers
Tara Lamb also asked if it is possible for pre-footers to have Banners:

>Call to Actions only allow for one action...correct?  We have three distinct actions that we call to on our site based on the audience.  Search/Request/Apply  https://distance.uiowa.edu/
-Tara Lamb

joewhitsitt:
>per-page disable option for page pre-footers

could be done similar to hide sidebar = https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L693 but would be limited to nodes. not sure how to extend to view pages, etc.

GaryRidgway:
- add inline entity form similar to article authors for the config page to select or create region items
- When added to the config page, any region should automatically be saved into config
- Edit Layout button should be added when an item is selected for a region
- Add redirect query to end of "Edit layout" link that sends the user back to the region settings config page

pyrello:
> > per-page disable option for page pre-footers

> could be done similar to hide sidebar = https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L693 but would be limited to nodes. not sure how to extend to view pages, etc.

FWIW, the block configuration settings are already set up to handle this type of setting, so if we are going to build this feature, we should probably just figure out how to expose that configuration form and no reinvent the wheel.

joewhitsitt:
to reduce view operations on the settings form, we can use `hook_entity_operation_alter` to check against the route and remove everything but the layout option

_most of_ these operation links are already set to return to the settings form

joewhitsitt:
Sounded like Alan was re-wiring the settings form after renaming the block to something more generic.

If we continue with using the block_id method for storing config (ignored), I don't think the update hook is necessary anymore. Sites will get the block via config and the settings form will be dynamic based on that block existing. The render will only happen if someone has gone to the form and set up a fragment for the reference. There is a chance of orphaned config if the block is deleted/replaced.

makurj:
Tested. Working as expected. I would like some more help text. For example: 
image

bspeare:
I tested this a little a noticed:

- Masquerading as webmaster I couldn't see the fragments after creating them on https://sandbox.local.drupal.uiowa.edu/admin/config/sitenow/region-settings.  @makurj did you notice that?

- We need to update the local menu tasks to look like the current layout builder editor interface.  This pr is useful for referencing how the icons were added: https://github.com/uiowa/uiowa/pull/3524/files. 


- The "Save layout" buttons need a little space and the fragment form ID could be added to this rule: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/admin/_layout-builder.scss#L9-L12. 

screen_shot_2021-11-01_at_11 16 14_am

makurj:
Nope, but we tested it with me as admin.

> I tested this a little a noticed:
> 
> * Masquerading as webmaster I couldn't see the fragments after creating them on https://sandbox.local.drupal.uiowa.edu/admin/config/sitenow/region-settings.  @makurj did you notice that?

GaryRidgway:
I've added the icons, the spacing before the layout buttons, and I have added help text to the settings page to help with clarity, though i did forgo the step-based descriptions to avoid having users need to jump around for individual steps. It now looks like this:
Screen Shot 2021-11-03 at 12 07 33 PM
 If we would like to go back to the step based descriptions I am totally okay with that, I would just love some input on how this is looking so far. Thank you all for the great feedback!

GaryRidgway:
Forgot to say, I didn't change the perms to get webmaster access working yet, because if we are waiting for config split 2 then we should probably allot time for that after it is ready.

cory-skeers:
Tested this out, and really liking it! Also tried breaking things, but everything worked well and without issues.

Couple notes:

- (Note, but no change) I personally want to move the "+ Add region item" button closer to or in the "Region items" fieldset. But this is consistent with other areas, like content and taxonomies, so don't think it should actually move.
- (Change, but may not be possible) If you edit a region item from the settings page, but then move to another region item edit option (eg go to Edit a region item, but then move to Layout), upon saving you will not be returned to the region settings page.
  - I'm assuming this is because the last page visited is no longer the settings page, and so you don't get redirected back. Not sure there'd be a way to change this without potentially messing with other functionality.
- (Future enhancement?) It would be great if on a page and/or when editing its layout, there was a contextual link (for logged-in users) to visit the regions setting page. Users with access to edit the Pre footer block get a contextual link for the block itself--could we adapt this, or add to it?
- (Future enhancement?) With the addition of the Fragments tab of admin/content, is there a way to connect this with the Region settings page? Additional actions added to the fragments listed, a message and link to the region settings page, or something else?

Edit: Also holding off on giving explicit approval until we can look at permissions.

joewhitsitt:
>With the addition of the Fragments tab of admin/content, is there a way to connect this with the Region settings page? Additional actions added to the fragments listed, a message and link to the region settings page, or something else?

Unfortunately, the fragments overview page is a sticky issue. Other sites have fragments for other purposes and I imagine this will grow. As a follow-up (not scope creep to this PR), I'd like to see fragment types (e.g. region item, program, location) listed as part of the display (filter/sortable) and maybe an operation link there for region item to go to region settings.

richardbporter:
This is my first time looking at this, very cool! For some reason though, the autocomplete on the region settings config page felt off to me. I thought a select list might be easier to use if for example, you can't remember what the fragment is called. Just my two cents.

Or maybe that view could list the region the fragment is in and/or allow placement of a fragment into a region somehow... :)

richardbporter:
The `administer fragment entities` permission did not seem to work for the region settings config page. We had to grant individual fragment permissions for it to work for other roles. We've updated the splits to remove the `administer fragment entities` and grant individual permissions to editor, publisher and webmaster roles. This is probably safer but more work unless CS2 merges soonish.

joewhitsitt:
pulling down to review

bspeare:
Is it possible to tack on https://github.com/uiowa/uiowa/issues/4389 to this?

joewhitsitt:
Didn't we groom it to be separate? Adding webform will at least be more config updates. Not sure if any style changes would be part of that.

bspeare:
> Didn't we groom it to be separate? Adding webform will at least be more config updates. Not sure if any style changes would be part of that.

It was marked as blocked because we didn't know when this pr would be come unblocked.  I was mostly wondering how easy it was to add.  Sounds like a follow-up pr will be best.

joewhitsitt:
Region items can be published/unpublished. This can lead to it not rendering to anonymous visitors yet there no indication beyond the fragment edit form that it could be unpublished. I don't think it is a blocker but we should probably create a follow-up task to make this more visible to editors

pyrello:
> Region items can be published/unpublished. This can lead to it not rendering to anonymous visitors yet there no indication beyond the fragment edit form that it could be unpublished. I don't think it is a blocker but we should probably create a follow-up task to make this more visible to editors

Does published/unpublished make sense for region items? Maybe we should just remove that option. Since a region item only shows up when it is placed, this seems a little redundant.

joewhitsitt:
fragments doesn't have anything like "override node options" so this most likely would be a form alter to set it as published and remove access

joewhitsitt:
Is this supposed to be on for v2 sites? As a v2 webmaster (hawkeyemarchingband) I can see region settings but not fragments/region item.

richardbporter:
> Is this supposed to be on for v2 sites? As a v2 webmaster (hawkeyemarchingband) I can see region settings but not fragments/region item.

We did not export sitenow_v2 permissions. I think we (mistakenly) assumed it would be v3 only as it uses LB but this is separate from basic pages so it probably could be. Right?

joewhitsitt:
Team decided to give this to v2 if it isn't too much cost. sitenow_v2, brand, hr, ighn.international, sitenow splits need to be updated.

richardbporter:
>brand, hr, ighn.international, sitenow 

I thought I did all the site splits yesterday that had user.role.* config split out. Commenting quick so nobody redoes that work without checking first.

dnkemper:
Rick you rock! I see brand split, hr split and ighn split

joewhitsitt:
Ah okay. just sitenow_v2 then

dnkemper:
Updated sitenow_v2 perms. Removing assignment.

joewhitsitt (approved):

PR #4322, opened 2021-10-18
#3327 - File replace caching

### LINKED ISSUE #3327: File replace caching
## Issue

Varnish for the file is not purged when replaced.

>Most likely, static files (.txt, .pdf, etc) served from your website will be cached by external caches for a long time, which means if you replace the contents of a file on your web server, users may not see the update content. External caches include things like a Varnish, a CDN, and a browser. You can control the behavior of external caches by emitting certain HTTP headers for static file responses. This cannot be configured in Drupal, because Drupal is not involved in serving static file assets. Your web server is.

-- https://www.drupal.org/project/media_entity_file_replace

## Proposed Solution

Any or some of these would make an improvement

- If you can, configure your web server to emit an ETag cache header for static file assets instead of a max-age header.
- Set a short max-age header for static file assets. Drupal's default .htaccess file (which provides configuration for Apache) sets a max-age header to 2 weeks for static files. You can lower this to something like 5-10 minutes. You may need to inform your content editors that all users may not see the updated file contents for some time.
- Button to update metadata - https://www.drupal.org/project/drupal/issues/2983456
- Image purging - https://www.drupal.org/project/purge/issues/2920332
- Varnish purge - https://www.drupal.org/project/varnish_purge

dnkemper:
Closes #3327 

## Steps to reproduce on https://default.stage.drupal.uiowa.edu
- Open two browsers or use an incognito window so you have an authenticated and anonymous session.
- In one window, log in to the default site and create a page that has as a link to a document (PDF, txt, etc.).
- Open the page in the anonymous session and click the link to the file. Inspect the file contents.
- In the authenticated session, go to media admin overview and **replace** the document with a file that has different contents. 
- In the anonymous session, refresh the file and inspect the content.

## How to test on https://default.dev.drupal.uiowa.edu
Repeat the steps above on the dev environment where this PR has been deployed.

*A note on audio and images: it seems like browsers cache these assets and the change is not immediate. We had to do a hard refresh to get the new assets to show when previously loaded on a page. An example would be the focal point thumbnail or the featured image on a page. 

**Additional note on purging: the paths to the files are _queued_ for purging and not necessarily immediate. There could be a delay between replacing the file and seeing the updated content.


richardbporter:
Tested on dev. Works great. Made a few small changes:

- consolidated the entity bundle and moduleExists check into one if statement
- Combine fid/file load and check to see if file load returns a truthy value
- Use `$file->createFileUri()` method instead of file_create_uri()

composer.json
```diff
@@ -355,7 +355,7 @@
         ],
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
         "nuke": [
-            "rm -rf vendor composer.lock docroot/core docroot/modules/contrib docroot/profiles/contrib docroot/themes/contrib",
+            "rm -rf vendor docroot/core docroot/modules/contrib docroot/profiles/contrib docroot/themes/contrib",
```
richardbporter, line 359: Oh I also included this because I was using `composer nuke` quite a bit. Turned out to be unrelated to this PR but I think removing `composer.lock` would make this more useful.

dnkemper: Baha you read my mind. I was figuring out where that came from. I've had to run that a bit as well so makes sense to me 🤷🏼‍♀️

dnkemper:
Update: last thing to fix is the focal point of an image not getting purged. SO CLOSE.

richardbporter:
Looking into https://www.drupal.org/project/media_entity_file_replace/issues/3192873

dnkemper:
Ooh ooh ooh .htaccess file magic gonna happen? 😍

richardbporter:
I think this can tested now but I'm looking for feedback on https://github.com/uiowa/uiowa/issues/3327#issuecomment-947797782. Maybe keep that in mind while testing and/or try not replacing an image file in stage to see how the default behavior "just works".

joewhitsitt (approved):
nice work. as-is, i think this is a great step forward. discussion on the issue brings up worries about media revisions (yes, that is a thing) and orphaned files. i will create a follow-up issue.

dnkemper:
@richardbporter and @joewhitsitt for the win!

Issue #4174, opened 2021-09-22
Rename Author field label in Card component


bewoldt:
# Problem
When making a Card component there is a field with the label "Author". I've personally used this field for many things other than an author name. Renaming it to a more generic label would remove confusion for site editors.

<img width="441" alt="Screen Shot 2021-09-22 at 10 25 00 AM" src="https://user-images.githubusercontent.com/731952/134373257-4a4746ed-28f9-42d3-88bd-935d195019e3.png">

# Proposed solution
Label options could be Subtitle, or Subheading

PR #4325, opened 2021-10-18
Update to block card field label

bewoldt:
Updates the Card block to change the field labeled "Author" to "Subtitle". 

Issue #4174


bspeare (approved):
Great label change, thanks @bewoldt !

PR #4333, opened 2021-10-19
Feature: Fix for media embed overflows in safari

bspeare:
Pushing this up since Joe showed me how to get the facebook secrets and it seems like an easy fix.


joewhitsitt (approved):
checked a few other media types (e.g. media--type-remote-video) as well.

PR #4334, opened 2021-10-20
Composer security update and Drupal update to 9.2.7

pyrello:
Updates Drupal to 9.2.7, which allows composer to be update to 2.1.9, which is the most recent secure version.

Also had to update to a new patch in https://www.drupal.org/project/drupal/issues/2605218 to get the update to work.

# How to test

* If builds pass, we probably think this is okay, except...
* Test views block pager functionality on sandbox locally to make sure that selected people list block properties stay in place between AJAX block loads.


pyrello:
As far as I can tell, the new patch that was used has not resulted in any regressions with the people list block stuff. I reviewed the code changes too and did not expect any regressions based on that.

dnkemper (approved):
Couldn't find any regressions as well. Nice work, @pyrello

joewhitsitt:
testing the pager real quick.

dnkemper:
Crap... sorry, @joewhitsitt

joewhitsitt:
apologies. didn't see you were testing it. if you think it is good. let's go with it

PR #4339, opened 2021-10-20
Use ReplicatorEvents::After_SAVE in place of hardcoded event service.

cory-skeers:
As a workaround for ordering of checks/installations, we'd hardcoded what should normally be an imported content.

Now that the Replicate module has been installed in a previous deployment, this simply replaces the hardcoding with the appropriate drupal/replicate constant.

# How to test

`blt ds` a site, and see that it completes without error

Extra credit:
Replicate a page, and check that our ReplicateSubscriber fired (quick check--see that the replicant revision has the original node id listed in its revision message)


richardbporter (approved):
👍🏻

PR #4091, opened 2021-09-09
Config split 2

pyrello:
Related to #4035
Resolves #4137 
Resolves #1626

# How to test
* Add your name after the split so others know you are testing it.
* `git checkout config_split_2 && git pull && composer install`
* Test feature splits:
  * Review file changes ([instructions](https://github.com/uiowa/uiowa/pull/4091#issuecomment-956973725))
  * `drush @default.local sql:drop && blt ds && drush @default.local uli admin/config/development/configuration/config-split`
  * Enable the feature split you are testing
  * `drush @default.local cim`
  * Test for regressions
  - [x] additional_person_types - @pyrello
  - [x] apr - @pyrello
  - [x] areas_of_study - @e-marie-w 
  - [x] event - @cory-skeers 
  - [x] grad_person_extended - @cory-skeers 
  - [x] profiles - @pyrello 
  - [x] scholarship - @cory-skeers 
  - [x] sitenow_v2 (@joewhitsitt)
  - [x] thesis_defense - @e-marie-w 
  - [x] topic_page - @quamsta 
  - [x] uiowa_graph - @pyrello 
* For each site split:
  * Review file changes ([instructions](https://github.com/uiowa/uiowa/pull/4091#issuecomment-956973725))
  * `drush @site.local sql:drop && blt ds --site site && drush @site.local uli`
  * Test for regressions
  - [x] admissions.uiowa.edu - _multi-split_
  - [x] amcs.uiowa.edu - _multi-split_ - @pyrello 
  - [x] brand.uiowa.edu (@joewhitsitt)
  - [x] coronavirus.uiowa.edu (@richardbporter)
  - [x] ~~education.uiowa.edu~~  - _empty, deleted_ - (@richardbporter)
  - [x] genetics.grad.uiowa.edu - @e-marie-w 
  - [x] grad.uiowa.edu - _multi-split_ - @pyrello 
  - [x] hr.uiowa.edu (@joewhitsitt)
  - [x] icsa.uiowa.edu (@joewhitsitt)
  - [x] ighn.international.uiowa.edu (@joewhitsitt)
  - [x] immuno.grad.uiowa.edu - _multi-split_ - @pyrello
  - [x] iowasuperfund.uiowa.edu (@richardbporter)
  - [x] pharmacy.uiowa.edu (@richardbporter)
  - [x] policy.clas.uiowa.edu (@joewhitsitt)
  - [x] presidentialsearch.uiowa.edu (@richardbporter)
[...]

PR #4341, opened 2021-10-20
Create ThankYouForm block and configuration in sitenow_dispatch.

### LINKED ISSUE #4127: Thank You Program
The College of Pharmacy would like to automate their thank you program similar to the existing OneIT thank you program.

Their existing manual process is:
User fills out webform to submit a thank you
Webform generates email to Pharmacy staff council
Pharmacy manually sends templated email (see below) to staff being thanked as well as to the staff members supervisor.

They would like the webform submission to continue to send email to Pharmacy staff council as well as trigger a lookup of the staff members supervisor based on staff members email, trigger an email to be sent from dispatch to the supervisor and staff member.

The email template in dispatch would include who it is being sent to, who it is from (person submitting webform), image, the message from the webform and some text saying that the staff members supervisor has also been notified. The email from would be College of Pharmacy Staff Council and the subject would be "[Name], You've been thanked!" 

We have existing functionality on the OneIT website (https://oneit.uiowa.edu) that does this same thing and we will likely need to account for this functionality on that site when it moves to V3.

My main concern here is if we have access to get Pharmacy Supervisor information from HR or if the current endpoint that is available to us is only for OneIT staff/supervisors.

cory-skeers:
Resolves #4127 

# How to test
**Note** that the Dispatch communication is set to test mode and all emails will be sent to ITS web.

Make sure https://git.uiowa.edu/its-web/data.its.uiowa.edu/-/merge_requests/30 is deployed to the test server.

Add below config overrides to local.settings.php for site. Values can be obtained from uiowa.dev [secrets.settings.php file](https://docs.acquia.com/resource/secrets/#secrets-settings-php-file).
```
$config['sitenow_dispatch.settings']['thanks']['api_key'] = 'xxx';
$config['sitenow_dispatch.settings']['thanks']['hr_token'] = 'xxx';
```

### Stories
- As a webmaster configuring sitenow_dispatch, I cannot enable/disable the Thank You Form nor configure it.
- As a webmaster/editor/etc., I cannot place the ThankYouForm Dispatch block if it has not been enabled.
- As an administrator configuring sitenow_dispatch, I can enable/disable the Thank You Form and set its configuration.
- As a webmaster configuring sitenow_dispatch, I can configure the Thank You form if it is enabled.
- As an anonymous user, I can submit a Thank You form submission for someone with a UI email address.
- As a supervisor of someone receiving a Thank You form submission, I receive a copy of the email if the form is configured to do so.
- As the maintainer of an email address configured to receive a copy of a Thank You form submission, I receive said copy.

### Test existing Dispatch config is unchanged after database update
- `blt sync --site sitenow.uiowa.edu`
- Make sure api_key is still set in `sitenow_dispatch.settings`


cory-skeers:
> I'm not sure how the original form works exactly but I'd be curious if the webform remote POST handler could work for this. [comment in 4127](https://github.com/uiowa/uiowa/issues/4127#issuecomment-943730273)

@richardbporter Reproducing the D7 OneIT implementation is essentially an HTTP POST request in part of the form submission code. Did you want to look into the webform handler approach?

Current D8-ified version of the original approach:

```
$response = $this->httpClient->get($endpoint, [
  'headers' => [
    'x-dispatch-api-key' => $apikey,
    'accept' => 'application/json',
  ],
  'method' => 'POST',
  'data' => $this->jsonController->encode($members),
]);
```

richardbporter:
When I suggested that, I wasn't too familiar with how the D7 module worked. It looks like it was just sort-of attached to one particular webform and there was all this other validation around it. Now, I think the webform POST is probably not worth using because I doubt that specific validation can be done through the webform UI. Doing it all in code is probably easier.

richardbporter:
I think I'e got this sending emails locally now using a new Dispatch population and communication called Thank You Program. The supervisor is always its web locally - just set with some dummy data for now. Still need to wire up additional placeholders. The idea with the placeholders is that form_alters can add to the `$form['placeholder']` array and they will be passed to the Dispatch template automatically.

docroot/modules/custom/uiowa_thankyou/src/Form/ThankYouForm.php
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $placeholders = $form_state->getValue('placeholder');
+    $hr_data = $form_state->getValue('hr_data');
+    $uiowa_thankyou_settings = $this->config('uiowa_thankyou.settings');
+    $apikey = trim($uiowa_thankyou_settings->get('uiowa_thankyou_dispatch_apikey'));
```
cory-skeers: Should we trim this on saving the SettingsForm instead, since that also makes dispatch calls to find the campaigns?

richardbporter: Yes, good call.

richardbporter: I was actually hoping we could use uiowa_dispatch for this but I don't think that will work. If we are going to use one campaign in our Dispatch client, we don't want people having access to all of that client's campaigns, comms, etc. so we'll need to keep this separate. Actually, we probably don't want people having access to pick another campaign/comm on this form either. Otherwise, they could select Inactive Site or SiteNow Announcements...

docroot/modules/custom/uiowa_thankyou/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * @param string $endpoint
+   *   Fully qualified url.
+   * @param string $api_key
+   *   Api key from Dispatch.
+   *
+   * @return object
+   *   The HTTP response from dispatch.
+   */
+  protected function dispatchGetData(string $endpoint, string $api_key) {
```
richardbporter: We should use sitenow_dispatch for this and other requests. There is a service that can make requests and handle logging. Might require some refactoring.

richardbporter: https://github.com/uiowa/uiowa/blob/a525a77eb6d73aaceae2c8efb6564b5215d87297/docroot/modules/custom/sitenow_dispatch/src/Dispatch.php#L59

docroot/modules/custom/sitenow_dispatch/src/Form/SettingsForm.php
```diff
@@ -55,42 +55,155 @@ public static function create(ContainerInterface $container) {
...
+    // Grab the current user to set access to the Thanks
+    // form settings only for administrators.
+    $current_user = $this->currentUser();
+    $form['thanks'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Thank You Email'),
+      '#description' => $this->t('Thank You email configuration. The <a href="@link">1row x 1col curated Dispatch template</a> is used.', [
+        '@link' => 'https://apps.its.uiowa.edu/dispatch/help/curatedtemplate/UI%201%20row%20x%201%20col%20-%20Version%202',
+      ]),
+      '#access' => in_array('administrator', $current_user->getRoles()),
```
richardbporter: The user 1 account may not have the administrator role. There is a service in uiowa_core that accounts for this: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_core/src/Access/UiowaCoreAccess.php

Here is a hook-based example of using it: https://github.com/uiowa/uiowa/blob/d8e08172bf66c74ada567ee5e2c9c887c3b34224/docroot/modules/custom/uiowa_aggregator/uiowa_aggregator.module#L29-L38

richardbporter: I think `$access->isAllowed()` would work here.

cory-skeers: Cool! Thanks for the example--got it updated.

richardbporter:
@cory-skeers The banner image is optional in the email template so I think its ok if we mark it as such which I did in 47f89bf. I also added alt text if the config is set because I guess that is required if the image URL is set. One thing I noticed is if you set an image and save the form and then want to remove it later, the form errors. Not sure how to get around that.

cory-skeers:
> One thing I noticed is if you set an image and save the form and then want to remove it later, the form errors. Not sure how to get around that.

Pretty sure this was because we were attempting to load a media entity when we didn't have an mid (which works, but then the field accesses were erroring). Updated that a bit.

If we remove the image, do we need to remove the `banner_image_url` config, or is it enough to set it to a null equivalent?

richardbporter:
> If we remove the image, do we need to remove the banner_image_url config, or is it enough to set it to a null equivalent?

Good question. I'm not sure. I'm deploying this to uiowa.dev for testing to try it out because local image URLs won't work. I'll set it up on the default site there.

richardbporter:
Seems like the image URL placeholder suffers from the same issue as unit, unitUrl, etc. that Dispatch Support is working on. :(

cory-skeers:
Some of the webform styling from `uids_base` isn't getting applied when section backgrounds are updated (gold button on gold background, black field description text on black background, etc).

I'm having trouble adding classes to the appropriate place. I can get it added directly to the form without any issues, but `block-webform` should be added to the parent div instead.

image

Another option could be to add the `block-thankyou-form` to `uids_base` alongside `block-webform`, but I'd been hoping to use the existing webform class directly.

Edit: It looks like the submit button for the thank you form is also an id "submit", but the standard form styling is scoped to "edit-actions-submit", so there might be more updating needed, or a different approach. Currently, it requires adding the "block-webform" class to the parent div, and adjusting the submit button's id.

richardbporter:
>but block-webform should be added to the parent div instead.

I would assume this could be added to `ThankYouBlock::build` but I'm having trouble getting that in the right spot too. Does it have to be preprocessed in?

cory-skeers:
> > but block-webform should be added to the parent div instead.
> 
> I would assume this could be added to `ThankYouBlock::buildForm` but I'm having trouble getting that in the right spot too. Does it have to be preprocessed in?

Possibly? I'd gotten sidetracked for a bit, but I think the parent `div` is a part of the template, and the classes are being added there based on plugin id and/or provider...So might need to/be able to pass additional classes in there, but I'm not sure.

Edit: Specifically, this template: `themes/custom/uids_base/templates/block/block.html.twig`

richardbporter:
I pushed up some code which adds that class to the block wrapper on all sitenow_dispatch blocks. I also prepended the thank you block ID with `sitenow_dispatch_` so you'll probably need to re-place it on a page.

I'm not sure what to test with this though. Setting a black section background doesn't seem to change a form field description to something readable. Should it? This is the kind of thing I usually need help with... :)

cory-skeers:
> I'm not sure what to test with this though. Setting a black section background doesn't seem to change a form field description to something readable.

Yep--that looks like a webform-wide issue. Opened #4385 and #4386 for it.

With that, it looks like your changes are working great for all but the submit button. I think that's from the submit button's id being different ("submit" instead of "edit-actions-submit"), but it at least isn't an accessibility issue (gold button background on a gold section background, but the button text is readable).

richardbporter:
>With that, it looks like your changes are working great for all but the submit button. I think that's from the submit button's id being different ("submit" instead of "edit-actions-submit"), but it at least isn't an accessibility issue (gold button background on a gold section background, but the button text is readable).

Changed to match for now but it seems like that is something we'd want to style regardless of ID. Otherwise people will have to remember to change the ID every time they build a form.

cory-skeers:
> Changed to match for now but it seems like that is something we'd want to style regardless of ID. Otherwise people will have to remember to change the ID every time they build a form.

Perfect. Checked it out, and it works great. Also opened an issue for changing the scoping of that style.

richardbporter:
Doh. Reverted b8aa694. That breaks Drupal's unique ID generation when multiple forms are place on a page. 🤦🏻

cory-skeers:
> Doh. Reverted [b8aa694](https://github.com/uiowa/uiowa/commit/b8aa694227bee1186731c8b1d4d196049e6cd446). That breaks Drupal's unique ID generation when multiple forms are place on a page. 🤦🏻

Ope.

We can probably wait for #4387 then. since the buttons are still usable, and we'll want to update it for other blocks/forms as well.

richardbporter:
I think I'm done with this until we can start working with the email placeholders.

richardbporter:
It's probably easiest to just set this up locally and use the admin form to set some placeholders and send an email. I would use the PR description and stories as a guide.

I just deployed the dataits MR to crow.

config/default/sitenow_dispatch.settings.yml
```diff
@@ -0,0 +1 @@
+api_key: null
```
richardbporter: I think we'll want to remove this file as it causes problems with config_ignore after #4091.

docroot/modules/custom/sitenow_dispatch/sitenow_dispatch.install
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @file
+ * SiteNow Dispatch install hooks.
+ */
+
+/**
+ * Ensure sitenow_dispatch.settings config entity exists.
+ */
+function sitenow_dispatch_update_9001(&$sandbox) {
```
richardbporter: Can't remember why this was new but the admin form probably was having issues if the config entity did not exist. This update might not be necessary if that admin form was tweaked.

cory-skeers:
It looks like the unit name placeholder is working, but if I'm setting these up correctly, the placeholder within the row1_heading we include in the default config isn't working, nor is the banner_image_alt (might just be that I provided a local image path, and it not loading an image broke the alt).

Manually adjusting the row1_heading placeholder and passing it a plain string works, but makes the interaction between the placeholder settings and the block form tricky. 

image

docroot/modules/custom/sitenow_dispatch/sitenow_dispatch.install
cory-skeers: I think several settings weren't sticking on first save if the config didn't already exist, but that's been tweaked now.

config/default/sitenow_dispatch.settings.yml
cory-skeers: Removed!

cory-skeers:
Updated with non-local banner image url and alt text, and they weren't being inserted into the template. Looks like we do still have some placeholder issues.

cory-skeers:
Notes for possible followups:

- Add some text to the email body if supervisor is checked ("A copy of this has been sent to your supervisor"), being added
- Possibility of sending extra copies to other email addresses? Additional config to list email address(es), and extra copies built similar to how the supervisor email is handled
- Possibility of custom logging? Either the log, or another entry possibly in a separate table. Possibility of exporting these results, CSV or something similar

richardbporter:
I'm planning on reducing placeholder usage and setting up a default image in the email for now.

richardbporter:
782376e94ef4d1c3054020f6799776cdf4dbbaf3 temporarily removes the banner image placeholder until that replacement becomes available on Dispatch. Can be reverted later.

richardbporter:
Got this looking better for tomorrow. 👍🏻

joewhitsitt:
pulling down to test

richardbporter:
Additional stories
- As an administrator disabling the thank you form on a site where the block is already placed, I can see the block does not render when disabled.
- As a webmaster configuring Dispatch settings, I am required to fill out the API Key when the Thank You form is disabled.

Additional test
- Confirm sign up form does not render if API key is empty. This can only happen when thank you form is enabled. This is goofy but it is possible to use the Thank You form with no Dispatch client since it uses a separate key.

joewhitsitt:
As a webmaster, I am able to still click "Add block" despite the warning and then I get a lot of warnings of missing values but the form renders.

richardbporter:
> As a webmaster, I am able to still click "Add block" despite the warning and then I get a lot of warnings of missing values but the form renders.

Good catch. Should be fixed.

richardbporter:
I'm kind-of regretting moving this into sitenow_dispatch. It made sense initially but the logic to enable/disable on the settings form is really awkward. We could still yet move it back out but if we plan on just enabling this for all users at some point, then it would make sense to stay put.

richardbporter:
3d85c1c was preventing admins from disabling the thank you form without also adding an API key... 🤦🏻

briand44:
>We could still yet move it back out but if we plan on just enabling this for all users at some point, then it would make sense to stay put.

I am not sure we want to enable this for everyone at some point. I wonder if there are cases where OneIT wouldn't want another related site having a Thank You form because they would rather they all use the OneIT form to make managing submissions easier.

joewhitsitt:
So the settings for this seem a bit wonky to me too. As an admin I could be easily confused what needs to happen if to stop a site from spamming, etc.

Rick and I talked about keeping it within the same module but moving the settings to a separate, dare I say it, "tab" (links.task see google tag manager for an example). Additionally, similar to our google tag manager implementation, it would be nice to have a kill switch for this as well (e.g. uiowa_core:toggle-gtag) that we can just hit with drush.

There seems to be some scoping issues with form submit buttons still. The thank you form has different behavior than a webform and causing some contrast issues (e.g. Gold). Preview/Render seem different sometimes too.

joewhitsitt:
Do you have a list of items you need help with or want to start a huddle?

richardbporter:
Right now, just the styling issues you mentioned:

> There seems to be some scoping issues with form submit buttons still. The thank you form has different behavior than a webform and causing some contrast issues (e.g. Gold). Preview/Render seem different sometimes too.

I'm working on the tabs + alternative enablement method.

richardbporter:
> I'm working on the tabs + alternative enablement method.

I tried hiding the thank you form via a dynamic local task/route access and enabling it with Drush but that was problematic. Went back to just making it visible for everyone. I guess that solves the discovery issue for folks. I think this is ready.

joewhitsitt (approved):
Based on the recent changes, I think this is good

PR #4347, opened 2021-10-22
Adjusted replicate node confirmation form text, fixes #4346

### LINKED ISSUE #4346: Replicate button confirmation dialog may need to be adjusted for our use case
image

We were exploring the "Replicate" functionality this afternoon. It's amazing. One minor thing that we noticed is that there's some inconsistency in the terms used in the default confirmation dialog before replicating a page. 

For instance, there's several references to "labels" when currently we're only replicating pages, so "title" may be appropriate here instead.

Also, the phrase "This action cannot be undone" is a little scary and might be a bit off-putting or potentially confusing if the user thinks it's referring to the label field above it. Also, it's technically not true, since you can "undo" a page replication by deleting the replicated page.

quamsta:
Trying to make the replicate page confirmation dialog a bit more friendly and more tailored to our current use case for the module: replicating pages. 

Changes:

- The title "Are you sure you want to replicate _node_ entity id [id]?" now reads "_Replicate Page_ [Page Name]," which matches the "_Edit Page_ [Page Name]" action title.
- The field label "New label (English)" now says "Replicated page's title"
- The description "This text will be used as the label of the new entity being created, in English." now says "This text will be used as the title of the newly replicated page."
- Removed the potentially confusing "This action cannot be undone." message 

Before:
Screen Shot 2021-10-22 at 8 54 48 AM

After:
Screen Shot 2021-10-22 at 8 51 28 AM

# How to test

Replicate a page to see these adjustments (after clearing the cache).

### DIFF
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -126,6 +126,21 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
       }
       break;
 
+    // Customize the replicate confirm dialog to our
+    // use case (pages with titles)
+    case 'node_page_replicate_form':
+      if (isset($form['new_label_en'])) {
+        $form['new_label_en']['#title'] = 'Replicated page\'s title';
+        $form['new_label_en']['#description'] = t('This text will be used as the title of the newly replicated page.');
+        // Removes the "This action cannot be undone" text.
+        $form['description'] = ['#markup' => ''];
+        // Changes title to match the "Edit" action title
+        // (e.g. "<em>Edit</em> Home").
+        $node_title = $form_state->getFormObject()->getEntity()->$entity->getTitle();
+        $form['#title'] = t('<em>Replicate Page</em> @title', ['@title' => $node_title]);
+      }
+      break;
+
     case 'menu_link_content_menu_link_content_form':
       // All checked or unchecked this seems to do the same. Hide until we
       // have a use-case.
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -136,7 +136,7 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
         $form['description'] = ['#markup' => ''];
         // Changes title to match the "Edit" action title
         // (e.g. "<em>Edit</em> Home").
-        $node_title = $form_state->getFormObject()->getEntity()->$entity->getTitle();
+        $node_title = $form_state->getFormObject()->getEntity()->getTitle();
         $form['#title'] = t('<em>Replicate Page</em> @title', ['@title' => $node_title]);
       }
       break;
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -98,14 +98,18 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
       ->load($entity->bundle())
       ->label();
 
-    $form['new_label_en']['#title'] = "Replicated {$bundle_label}'s title";
-    $form['new_label_en']['#description'] = t('This text will be used as the title of the newly-replicated @type.', ['@type' => $bundle_label]);
-    // Removes the "This action cannot be undone" text.
-    $form['description'] = ['#markup' => ''];
     // Changes title to match the "Edit" action title
     // (e.g. "<em>Edit</em> Home").
     $form['#title'] = t('<em>Replicate @type</em> @title',
       ['@type' => $bundle_label, '@title' => $node_title]);
+
+    // Lowercase for the rest of the form.
+    $bundle_label = strtolower($bundle_label);
+
+    $form['new_label_en']['#title'] = t("Replicated @type content's title", ['@type' => $bundle_label]);
+    $form['new_label_en']['#description'] = t('This text will be used as the title of the newly-replicated @type.', ['@type' => $bundle_label]);
+    // Removes the "This action cannot be undone" text.
+    $form['description'] = ['#markup' => ''];
 
   }
 
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -83,7 +83,9 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
   if ($route_name == 'entity.node.replicate') {
     // Get allowed replicate entity types.
     $allowed_types = \Drupal::config('uiowa_core.settings')->get('uiowa_core.replicate_allowed');
-    $node_type = $form_state->getFormObject()->getEntity()->bundle();
+    $entity = $form_state->getFormObject()->getEntity();
+    $node_type = $entity->bundle();
+    $node_title = $entity->getTitle();
     // Set message and prevent replicate.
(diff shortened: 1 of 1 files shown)


joewhitsitt:
So, replicate is also enabled for articles. It is also capable of being extended to other content types by developers. Can we look into ways to generalize when this fires and then gets the bundle name to use in the language?

Also, I think "This action cannot be undone" should be left in but can be modified to be more clear. This statement is common around Drupal for things (deleting content, etc.)

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -126,6 +126,21 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
       }
       break;
 
+    // Customize the replicate confirm dialog to our
+    // use case (pages with titles)
+    case 'node_page_replicate_form':
+      if (isset($form['new_label_en'])) {
+        $form['new_label_en']['#title'] = 'Replicated page\'s title';
```
pyrello: You can use double quotes to avoid needing to escape the single quote here.

pyrello: FWIW, this is the only time I will suggest using double quotes :)

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -126,6 +126,21 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
...
+        $form['new_label_en']['#description'] = t('This text will be used as the title of the newly replicated page.');
+        // Removes the "This action cannot be undone" text.
+        $form['description'] = ['#markup' => ''];
+        // Changes title to match the "Edit" action title
+        // (e.g. "<em>Edit</em> Home").
+        $node_title = $form_state->getFormObject()->getEntity()->getTitle();
+        $form['#title'] = t('<em>Replicate Page</em> @title', ['@title' => $node_title]);
```
pyrello: Per @joewhitsitt's comment, get the entity bundle to generalize "Page" to be whatever bundle type is being replicated. Since technically, this could apply to non-node entities, it might be good to add a check if the entity type is node and then set this title based on the bundle.

pyrello (changes requested):
See individual comments.

quamsta:
> So, replicate is also enabled for articles. It is also capable of being extended to other content types by developers. Can we look into ways to generalize when this fires and then gets the bundle name to use in the language?

Oh! Yes, I'll see if I can find a way to genericize it. Good call

> Also, I think "This action cannot be undone" should be left in but can be modified to be more clear. This statement is common around Drupal for things (deleting content, etc.)

Up for any ideas, but I feel "This action cannot be undone" is pretty unclear in this case. Makes sense for deleting items or other potentially destructive actions, but it might be a little harsh for something like copying an item.

docroot/modules/custom/uiowa_core/uiowa_core.module
joewhitsitt: would have to change how you get to the right form to alter though. currently `node_page_replicate_form` maybe there is another identifier that you can latch on to

joewhitsitt:
>Up for any ideas, but I feel "This action cannot be undone" is pretty unclear in this case. Makes sense for deleting items or other potentially destructive actions, but it might be a little harsh for something like copying an item.

Yeah, that is fair. It wasn't too long ago that this action was causing database corruption. Glad Cory found a fix and we are where it is today, but I am crossing my fingers each time that button is pushed. I guess the publisher doesn't need to know all that and "for your developer's sake, please use sparingly" doesn't quite roll off the tongue. I will back down on this change.

quamsta:
> > Up for any ideas, but I feel "This action cannot be undone" is pretty unclear in this case. Makes sense for deleting items or other potentially destructive actions, but it might be a little harsh for something like copying an item.
> 
> Yeah, that is fair. It wasn't too long ago that this action was causing database corruption. Glad Cory found a fix and we are where it is today, but I am crossing my fingers each time that button is pushed. I guess the publisher doesn't need to know all that and "for your developer's sake, please use sparingly" doesn't quite roll off the tongue. I will back down on this change.

Yeah, totally understand the reasoning behind wanting to keep a message like that with the difficulties the module presented, but from a user perspective the "cannot be undone language" may present the question of "should I not be doing this?" and I also had a user ask "Is it referring to the title field not being able to be changed after the fact?" since the message was showing up directly under the title field.

quamsta:
Also @joewhitsitt, I just committed a change that should allow for any allowed replicated content type to be referred to in this dialog (instead of just pages). I did a hacky "ucfirst()" to get the content type label to be uppercase in the title attribute, but if you know a better way to do that, please help haha.

joewhitsitt:
More processing but I think something like this would work and handle a world of multi-worded names in the future (e.g. Areas of Study):

```php
$bundle_label = \Drupal::entityTypeManager()
  ->getStorage('node_type')
  ->load($node->bundle())
  ->label();
```

>Since technically, this could apply to non-node entities, it might be good to add a check if the entity type is node and then set this title based on the bundle.

A lot of our code so far is only focused for "content" based on the replicate settings, so let's hold off of on future-proofing for all entities.

docroot/modules/custom/uiowa_core/uiowa_core.module
quamsta: Ended up using the `if ($route_name == 'entity.node.replicate') {` statement that's already being used instead of `node_page_replicate_form`

quamsta:
> More processing but I think something like this would work and handle a world of multi-worded names in the future (e.g. Areas of Study):
> 
> ```
> $bundle_label = \Drupal::entityTypeManager()
>   ->getStorage('node_type')
>   ->load($node->bundle())
>   ->label();
> ```
> 
>  > Since technically, this could apply to non-node entities, it might be good to add a check if the entity type is node and then set this title based on the bundle.
> 
> A lot of our code so far is only focused for "content" based on the replicate settings, so let's hold off of on future-proofing for all entities.

Okay, this is now using the bundle's label() instead of node_type. This does create a minor grammatical issue where the label is now capitalized when it maybe shouldn't be (e.g.,the  "Replicated Article's title" field label should probably be "Replicated article's title," since I've seen "article" and "page" used in lowercase before, but it's probably not a huge deal?

joewhitsitt:
pulling down to test

joewhitsitt:
I guess I misunderstood what you went with ucfirst() originally when providing you a way to get the human readable label.

Hope you don't mind, but I tweaked this a bit to get the best of both worlds. I still think the human readable label should be used but that it should be lower case. I think you protect yourself from apostrophe errors in the future "content" can be squeezed in. (e.g. senior thesis content's title).

Screen Shot 2021-10-22 at 4 03 30 PM

Screen Shot 2021-10-22 at 4 31 24 PM

I am going to approve, but if you disagree with my change, we can revert and go a different way.

joewhitsitt (approved):

quamsta:
> I guess I misunderstood what you went with ucfirst() originally when providing you a way to get the human readable label.
> 
> Hope you don't mind, but I tweaked this a bit to get the best of both worlds. I still think the human readable label should be used but that it should be lower case. I think you protect yourself from apostrophe errors in the future "content" can be squeezed in. (e.g. senior thesis content's title).
> 
> Screen Shot 2021-10-22 at 4 03 30 PM
> 
> Screen Shot 2021-10-22 at 4 31 24 PM
> 
> I am going to approve, but if you disagree with my change, we can revert and go a different way.

This is great! Thanks a ton Joe! Have a good weekend.

pyrello (approved):
Approving based on @joewhitsitt's approval.

PR #4369, opened 2021-10-28
Feature: Editor preview fix for vertical alignment

bspeare:
I noticed this while demoing the vertical alignment fix.    It was not showing up because we had some older CSS in `admin.scss` that didn't get removed when we moved to grid layout. 

Before: section with "vertically aligned to top" selected:

Screen Shot 2021-10-28 at 1 36 16 PM

After:  section with "vertically aligned to top" selected"

Screen Shot 2021-10-28 at 1 39 00 PM

# How to test

- It is most apparent if you try on the four columns of stats on https://socialwork.local.drupal.uiowa.edu/node/1/layout.


cory-skeers (approved):
Works as advertised. Nice cleanup!

PR #4375, opened 2021-10-29
Enable remote video on card block

### LINKED ISSUE #2724: As a site owner, I can place an embedded remote video into the card component
# Story

As a site owner, I can place an embedded remote video into the card component so that the card format can showcase YouTube videos.

Screen Shot 2020-12-17 at 3 58 12 PM

Screen Shot 2020-12-17 at 3 56 51 PM

# Proposed solution

Like the banner block has a "Background" select option, we would create a "Media" option on the card block that would allow users to select an image or video.  

Screen Shot 2021-10-08 at 4 17 37 PM

Once the video has been selected, we would hide some of the card formatting options like:

- Remove card image size options and set to large by default. 
- Set image as the default media option (will this need an update hook?)
- Use the banner image issue as reference for what needs to be done: https://github.com/uiowa/uiowa/pull/2545. 

Screen Shot 2020-12-17 at 4 06 11 PM

makurj:
Resolves #2724 

# How to test

As a content editor, I can add a remote video to a card. Do the "image format" and "image size" fields disappear on video addition?

### Video

If "Image position" is set to Stacked, the image size options should not be available.

Screen Shot 2021-11-03 at 1 42 39 PM

If "Image position" is set to Right or Left, the Large and Medium image size options should appear:

Screen Shot 2021-11-03 at 1 42 48 PM

Screen Shot 2021-11-03 at 1 42 56 PM

### Image

If "image format" is set to Circle and "Image position" is Stacked, then image size options should appear.  

Screen Shot 2021-11-03 at 1 43 39 PM

If "Image position" is set to Right or Left, the "Image size" options should be available no matter what the "Image format" is set to. 

Screen Shot 2021-11-03 at 1 43 49 PM
Screen Shot 2021-11-03 at 1 43 54 PM


bspeare:
I think we also might want to hide the `small` and `medium` image size options and set the default to `large`.   Here's what a video would look like if set to `small`:

Screen Shot 2021-11-02 at 9 56 46 AM

Edit: or just remove `small` if that is an easier option to implement.

pyrello:
@bspeare If it is showing that option while video is selected, then that is not the intended behavior. I would request changes to that effect.

bspeare:
@pyrello Thanks, I will look closely at this and make the request.

bspeare (changes requested):
The "Image size" option is still appearing when I have set a remote video and as soon as I select "Right" or "Left" alignment the "Image size" select option appears. 

Screen Shot 2021-11-02 at 10 21 41 AM

bspeare:
Looks like `card-block-form.js` is not loading on the page.

makurj:
Intended behavior for image: if image position is stacked, and circle is image format, image size shows up. Image size doesn’t show up for square and widescreen. If left is selected for position. With left for image position, no matter the image format, image size is always visible. Same with right image position.

Intended behavior for remote video: Image size and image format should not be visible irrespective of image position options.

docroot/modules/custom/layout_builder_custom/js/card-block-form.js
```diff
@@ -14,22 +14,22 @@
         const $image_format_field = $form_wrapper.find('.form-item-layout-builder-style-media-format').find('select');
         const $image_size_field = $form_wrapper.find('.form-item-layout-builder-style-card-image-size').find('select, label');
```
pyrello: I'm not sure about this approach. If I'm not mistaken, how this works is that you are hiding the `<label>` and `<select>` element, while leaving the `<div>` wrapper that contains them visible. While this technically works, it feels a bit like a kludge.

bspeare: Good point!  We are still refining this a bit.

bspeare: @pyrello Line 15 has been changed.   Do you have an issue with line 14 too?

bspeare:
I've updated the testing instructions with screenshots explaining the Display options to test.

dnkemper:
I can review this as an outside perspective since I haven't been involved in any discussion or work if that's okay? 🤷🏼‍♀️

makurj:
Sure! @dnkemper
Thanks, @bspeare for updating the issue with amazing screenshots!

dnkemper:
- [x] If "Image position" is set to Right or Left, the "Image size" options should be available no matter what the "Image format" is set to.
- [x] If "image format" is set to Circle and "Image position" is Stacked, then image size options should appear.
- [x] If "Image position" is set to Right or Left, the Large and Medium image size options should appear:
- [x] If "Image position" is set to Stacked, the image size options should not be available.
^^all pass in the testing - LOVE the screenshots btw to make sure I'm on the right track. Question though.. is there a way to set a name then or not? If not, why does it allow you to search the videos by name if the name can't be set? Can only set tags. Not sure if this matters or not? Looks great though other than that

makurj:
Thanks @dnkemper. That is odd. We don't have the option to add a name when uploading a remote video but we can add tags, and searching by tags works great. The name works also, but only if you know the name of the video from the source.

dnkemper:
@makurj OH! Okay! If that's the desired functionality, then I'm good on approving unless @bspeare wants more changes?

bspeare (dismissed):

bspeare (changes requested):
I am encountering an issue where the "Small" size remains if I update an existing card that was using the "Small" image size.   Try editing one of the left aligned small cards on https://sandbox.local.drupal.uiowa.edu/node/21/layout. 

Screen Shot 2021-11-04 at 10 54 55 AM

bspeare (approved):
That fixed the issue! Nice work @makurj !

makurj:
Thanks for all your help @bspeare!

PR #4376, opened 2021-11-01
View updates for Grad sites 

e-marie-w:
# Story
AMCS would like all graduates listed, instead of only most recent. 

Neuroscience would like the searchable people views like Genetics and Immunology, so like those two it needs a site split.

# How to test


pyrello:
@e-marie-w I think this is unblocked now.

pyrello (approved):
Reviewed, not tested. I got to view it via screen share.

PR #4404, opened 2021-11-05
Initialize byrne.lab.uiowa.edu multisite on uiowa

e-marie-w:
Provisioned bryne.lab.uiowa.edu on uiowa because 01 had more SANs and 02 more DBs so I went with middle option.


joewhitsitt:
future reference, *.lab and our other pre-approved patterns don't count against SANs

dnkemper (approved):

e-marie-w:
Ahh good to know, thanks @joewhitsitt! And thanks @dnkemper for approving!

dnkemper:
@e-marie-w @joewhitsitt I learned something new today as well! Should we add that to provision message? Looks like everyone has been provisioning sites in uiowa/uiowa01 vs uiowa02 or am I misunderstanding?

dnkemper:
OH DUH. Was thinking about databases. JK

joewhitsitt:
@dnkemper When the results come up, it will say that *.lab is already covered on all three in the terminal table because of the wildcard. it won't add to the SAN count. that is why dbs/sans count don't match.

PR #4406, opened 2021-11-05
Duplicate views result with multiple arguments on people_list_block and article_list_block

### LINKED ISSUE #4267: Duplicate views result with multiple arguments on people_list_block and article_list_block
# Problem
If content has more than one taxonomy term reference and those same terms are used in the people list block and article list block as multiple arguments, view results display duplicates.

# Steps to reproduce

- Create a person node, add two tags 
- Use those same two tags in a people_list_block
NOTE: do same with research areas - bug affects this filter as well

- Create an article node, add two tags.
- Use those same two tags in a article_list_block

https://sandbox.prod.drupal.uiowa.edu/node/551/edit

Screen Shot 2021-10-05 at 3 46 38 PM

# Proposed solution

- [ ] See if views distinct or aggregation can fix the issue without causing additional issues.
- [ ] See if setting "reduce duplicates" on the individual filters can fix the issue without causing additional issues.
- [ ] Make sure the way the arguments are passed in aren't related to https://github.com/uiowa/uiowa/issues/2109
- [ ] If the above don't work, last option could be to add a preprocess prior to render to manually skim the entity_ids and reduce duplicates. Increases noodle-code, but gives (almost) full control over timing and methodology of duplicate removal.

dnkemper:
Closes #4267 

Reduced duplicates in exposed filters - went with lightest fix

NOTE: this affects article list block - multiple people have multiple tags & people list block - when people have multiple research areas in addition to multiple tags

# How to test

1. checkout branch
2. blt ds --site=sandbox.uiowa.edu
3. go to https://sandbox.local.drupal.uiowa.edu/content-same-name and make sure no duplicate entries. Be sure to add new people with tags and add a pager (added: duplicate tag, duplicate bug test)
4.  Also test person research areas - looks like this was affected by same bug (added 'duplicate research area and duplicate bug test in prod - can filter on those)
NOTE: added research areas to prod - so can filter by them (duplicate research area and duplicate bug test)
4. Do same for articles list block too (added: bug test, same name)


pyrello:
@dnkemper I'm curious why the change happens twice in the People view, but not the Articles view?

dnkemper:
@pyrello clearly attempting to multitask this with Makur's environment and updating wiki doc was a terrible idea. People list block has a research area filter and the article list block does not. Will remove because this solely accounts for tags. Thank you for having a brain today because I clearly do not

dnkemper:
@pyrello helped me realize that the same bug affected the research areas filter on the people list block (ie if add same research areas to ppl and filtered list block by them it created duplicate nodes). Added back the 'reduce duplicates' functionality.

richardbporter (approved):
I'm not noticing any other issues that this might cause.

dnkemper:
@richardbporter YAY! Thank you so much for testing 🤗

PR #4407, opened 2021-11-05
media video/remote video fix appending to classes array

joewhitsitt:
Video card revealed a bad classes array. Fixed how media--video is added so that it doesn't erase other classes (e.g. align-right)

## To Test

Check banner local video to make sure the potential for other classes being passed in doesn't affect render.

https://sandbox.local.drupal.uiowa.edu/

Check remote video wysiwyg alignment and general card video renderings.

https://sandbox.prod.drupal.uiowa.edu/media


bspeare (approved):
Thanks for fixing this!

PR #4412, opened 2021-11-08
Added a new custom text field for button for listblock

### LINKED ISSUE #4305: Custom more link label via ListBlock
# Story

As a site editor I would like to add a more link to my people/articles/events (local) views block and choose to use a default label or provide a custom label.

# Background

https://github.com/uiowa/uiowa/issues/3501

https://github.com/uiowa/uiowa/issues/3501#issuecomment-874905664 was decided it should be a separate follow-up task.

# Proposed solution

- [x] Similar to the more link path provided via ListBlock, provide a more link label field.
- [x] For existing blocks out there, default to "View more" or whatever we have currently.
- [x] Update logic to use custom label if populated.

makurj:
Resolves #4305 

# How to test

1. `blt ds --site sandbox.uiowa.edu`
2. Check articles list block to ensure view more link works with and without custom text.
3. Check people list block to ensure view more link works with and without custom text.
4. If you leave the custom link text blank, it should default to "View more".

We have already tested the update hook via querying mysql before and after.


docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php
```diff
@@ -333,6 +356,11 @@ public function blockForm(ViewsBlock $block, array &$form, FormStateInterface $f
         $more_link_help_text = $this->t('Start typing to see a list of results. Click to select.');
       }
 
+      $more_link_text = $this->getOption('more_link_text');
+      if (empty($more_link_text)) {
+        $more_link_text = $this->t('Enter your custom text for the button');
```
pyrello, line 361: This seems more like the text you would want in `#description` than to be set to this value.

joewhitsitt:
pulling down for review

joewhitsitt:
Got an error when trying to visit https://sandbox.local.drupal.uiowa.edu/articles

`Error: Object of class Drupal\Core\Url could not be converted to string in Drupal\Component\Utility\UrlHelper::parse()`

On https://sandbox.local.drupal.uiowa.edu/people-directory

`Warning: strpos() expects parameter 1 to be string, object given in Drupal\Component\Utility\UrlHelper::parse()`

Taking a look

joewhitsitt:
Looks like this change is the culprit. @makurj or @pyrello recall why this was changed?

https://github.com/uiowa/uiowa/commit/13448b36f544459c80702c7e5eb2c8fa58b322d7#diff-4d4408e67a8f25d1594cd344d3bdadde8d339e0a35bc8d48350beb35f2854c4cL586

Update: Sean said it looks like a copy/paste error. Adding it back in.

joewhitsitt (approved):

PR #4422, opened 2021-11-09
Start of diff updates

### LINKED ISSUE #2302: Visual Inline Diff is not correctly configured for layout builder/content blocks
## Problem

Revision and diff information is available for v2 paragraph content blocks but doesn't seem to be configured correctly for v3 layout builder. Most revisions show "no visible changes"

Screen Shot 2021-11-02 at 12 57 00 PM

## Proposed Solution

Revisit diff configuration (admin/config/content/diff/general) and field diff configuration and get it a step further to provide layout builder revision information that is helpful when comparing revisions.

- [ ] Enable "Visual Inline" option for v3 sites, change order so it is not the default diff display.
- [ ] Since it uses uids_base to display the diff, adjust styling to handle the diff functionality (links/buttons, timestamps, revision info). It doesn't have to be perfect, but it should be representative, clean and functional.
- [ ] Audit field settings for diff to make sure we are comparing against fields that should be and that have diff functionality available (core or field specific)
- [ ] [Add local tasks patch if still needed](https://www.drupal.org/project/diff/issues/3169729)

### LINKED ISSUE #4465: Split diff plugin for revisions
Start of PR in #4422

There's a diff plugin in `layout_builder_custom` for use in split/unified diffs. The primary functionality works (but could probably use some cleanup). Otherwise, the primary goal of this slice would be to focus on the formatting of the output, and getting it into a more user-readable state.

Example of the initial output:
image

See if we can improve formatting some to make it more usable for users.

Visual Inline out of scope for this issue but would be solid improvement in future.

cory-skeers:
Resolves #4465 

# How to test

checkout branch and pull
blt frontend
sync a site
make some changes, compare some revisions
have a great time doing so
Try combinations of split/unified and raw/strip tags

As a content editor, are you able to understand what changes occurred, and navigate to layout builder to view the changes in place?

We have direct control over the array of strings that is sent to diff, so what from layout builder gets included, and how all labels and values are represented as strings. We do not have direct control over how the diff is calculated once these strings are formed, or how the results table is put together, but improvements to these could be tackled as followups.


cory-skeers:
Would be useful to attach a css override like below to help with formatting/readability issues. Revision diffs are in an admin page, so needs to be attached and override the claro theme.

```
table.diff.diff-responsive-table td {
  white-space: pre-wrap;
}
```

cory-skeers:
Looks like text area field changes are being split into separate result rows, which can sometimes lead to extra spacing, and sometimes lead to weird results like a "change" that is separate from any labeling. Need to look into...

image

~This happens after our plugin, and is a "feature" of diff.~

Edit: This has been worked around.

cory-skeers:
This is open for testing and feedback.

GaryRidgway (approved):
Working to the extent expected!

PR #4427, opened 2021-11-11
Hotfix - Region Items Permissions

joewhitsitt:
Removing v2 support
Giving editors add/edit, removing delete permissions.
Give view permissions across the board.
Update editor/webmaster to work with region item layout builder

Removed presidentialsearch editor as it was identical

## To Test

- pick a site, any site. if it is v2, you should not see the word fragment or region anywhere. for v3, editors/webmasters can add/edit/layout region items and configure region settings. editors cannot delete region items for their own good.


bspeare:
Pulling to test

bspeare (approved):
Tested on brand and diversity and the testing conditions worked!

PR #4442, opened 2021-11-16
Added fix for contextual links in lb

### LINKED ISSUE #2716: LB contextual links don't work after dragging block
# Problem
After dragging a block within the layout builder editor, contextual links no longer work until the layout changes have been saved.

# Steps to reproduce
Edit a page layout.
Drag a block into a new section or a new region within the section.
Attempt to edit/move/remove the block using the contextual link.

# Proposed solution
Unsure, but my guess is the contextual link is attempting to load a block based on the section/region in which the block is appearing in the layout preview, but as the layout has not yet been saved, the block does not exist by that reference. Some way that reference chain needs to be retained and/or updated upon dragging. Can maybe look into how the contextual move option works, as that appears to work correctly.

dnkemper:
Closes #2716 

P.S. Wanting Ben to review it because we know it doesn't work for him. 
# How to test

1. composer install
2. blt ds --site=sandbox.uiowa.edu
3. Try editing https://sandbox.local.drupal.uiowa.edu/node/586/layout and move the block above and then move the block way below to the bottom region


richardbporter:
Sounds like we are going to wait for core to resolve this.

PR #4451, opened 2021-11-17
Post CS2 Fixes

joewhitsitt:
- Admissions patch split for blockquote display settings due to entity print.
- Fix typo in AMCS split
- Update grad site permissions

## To Test

- no sync errors on admissions, amcs, grad


pyrello (approved):
Reviewed, and tested changes to grad.uiowa.edu but not the rest.

PR #4460, opened 2021-11-18
Only change image size when previously set to small

### LINKED ISSUE #4441: Image size resets to large when editing a remote video on a card set to medium image size
# Problem
Image size resets to large when editing a remote video in a card set to medium image size

# Steps to reproduce

Try editing the block on https://sandbox.prod.drupal.uiowa.edu/node/616/layout and note that the current image size is "Medium" but the block will reset to "Large" when you edit the block. 

# Proposed solution
Revisit https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/js/card-block-form.js

cory-skeers:
Resolves #4441 

# How to test

`blt ds --site=sandbox.uiowa.edu`
Try editing
https://sandbox.prod.drupal.uiowa.edu/node/616/layout
When configuring the video, the image size should remain set to "Medium"
Try different combinations of saving the block as medium or large, and see that it saves correctly and opens to the same size when re-editing
Try removing the video or adding a new block, setting the image size to small, and see that it still updates to Large when then adding a remote video


bspeare (approved):
Nice fix, thank you @cory-skeers!

PR #4474, opened 2021-11-19
Dependencies: Config Split and Layout Builder Shortcuts

pyrello:
* Updates Config Split to 2.0.0-beta5 and removes the patch that got merged.
* Updates Layout Builder Shortcuts to 1.0.1.

# How to test

This *shouldn't* change anything about how `config_split` works since it is just an update that includes that patch that we were using. I synced a couple sites (admissions and grad) and re-exported them to ensure there were no changes to config. I also tested the update to `layout_builder_shortcuts` when I was making it. I'm open to ideas on more/better testing we should do, especially related to HR (ahem, @joewhitsitt).


pyrello:
Closing in favor of #4448, which has this branch merged into it.

PR #4479, opened 2021-11-22
Views block exposed filters

### LINKED ISSUE #3804: Add exposed filters to views block display

pyrello:
A replacement PR for #3829 that resolves #3804

# How to test
```
composer install
blt ds --site sandbox.uiowa.edu
```
## Article and people list blocks
* View some existing article and people lists block and compare the settings with the production sandbox site to make sure that the update hook worked as expected.
* Place blocks on a page and try changing the following settings, saving the block, saving the layout, and viewing the page.
  * Test setting headline or not
  * Test changing items to display
  * Test changing offset
  * Test setting pager
  * Confirm that the 3 settings above work predictably with one another
  * Test changing sort options
  * Test hiding fields.
  * Test adding a more link, setting a custom path, and setting custom text
  * Test setting filter values
  * Test exposing filters without setting a value
  * Test exposing filters with a value set
  * Test setting multiple filters at the same time
  * Test setting multiple filters to different variations of exposing the filter and not
  * Test that exposed filter and pager for one block do not affect any other views blocks on the same page of either the same type or a different types
  * Confirm that "Items per page" and "Offset" do not show up in exposed filters

### People list block
* When setting the "Person type" filter, it should limit the options display in "Person type status" drop-down to the same option.

## Event list block
```
drush @sandbox.local config:set config_split.config_split.event status true
drush @sandbox.local cim
```
* Create some local events.
* Run through the above tests.


pyrello:
There has been quite a bit of work done on the patch code for the related issue adding this functionality to `ctools`: https://www.drupal.org/project/ctools/issues/2657060. Might be time to give it another spin :)

pyrello:
I got excited and a little bit of ahead of myself yesterday when I marked this ready for review.

cory-skeers:
Pulling this down to do some testing! Will update with how much I go through, if it's not all of it.

dnkemper:
@cory-skeers okay are you psychic? Because my environment is... not working at the moment

cory-skeers:
> @cory-skeers okay are you psychic? Because my environment is... not working at the moment

Ha! Not psychic, just had the PR open and hadn't refreshed to see your assignment 😬

cory-skeers:
Testing exposed filter on articles blocks, everything working well with a single articles block on the page. With two (or more) articles blocks on the same page, running into a few issues with the exposed filters (sort, hide, pager, etc working fine regardless of one or many blocks on a page).

If multiple blocks are exposing filters, using any of the filters is applying that filter to all of them on the page (eg adding 'bug test' tag filter to one applies it to all--'bug test' will appear in all the exposed filters and be applied to all blocks, with or without an exposed filter)

With two articles blocks, top with exposed filter and bottom without, using the exposed filter on the top block applies the filter to the bottom block but not the top from which it is exposed.

pyrello:
Converted to draft to investigate the reported issues.

pyrello:
My local testing did not result in the same outcomes that @cory-skeers was seeing, so we did a huddle to confirm that I was testing it right. It seems like there may be an issue with his local development environment that is affecting the testing.

pyrello:
This needs an update hook because we are ditching our homegrown views block filtering solution in favor of a contrib one.

pyrello:
The way that we built the local event content type and its related list block view is not ideal for exposing the virtual event field as a filter. We want it to effectively work as a true/false value, but it is actually a URL, so we are checking to see if the URL is there as our true value and false otherwise. However, this way of doing things means needing to add code to handle this specialized logic for the virtual event field.

We could alternately add a checkbox that serves as our true/false value and then leave the existing field to actually output the virtual event URL if it exists. The downside is that this extra field will require some of its own logic to handle the show/hide of the virtual event URL field.

However we decide to handle this, we should figure it out and include the change as part of this update, since we are already writing an update hook to handle changing the block configuration for these types of values.

pyrello:
After discussing this at our mid-sprint demo, most immediately we would like to sidestep the issue with the virtual filter by removing it. Longer term, we could consider adding a hidden boolean field to track whether the event has a virtual URL and set that on node save based on whether the user set a virtual URL field.

briand44:
Do we know if Admissions is currently using the virtual filter?

pyrello:
> Do we know if Admissions is currently using the virtual filter?

It does not appear that we are currently allowing that as an option.

joewhitsitt:
Going to pull and test

joewhitsitt:
Getting some notices on https://sandbox.uiowa.ddev.site/people-directory

`Notice: Undefined offset: 1 in Drupal\ctools_views\Plugin\Display\Block->preBlockBuild() (line 640 of modules/contrib/ctools/modules/ctools_views/src/Plugin/Display/Block.php).`

`Notice: Trying to access array offset on value of type null in Drupal\views\Plugin\views\display\DisplayPluginBase->getHandlers() (line 852 of core/modules/views/src/Plugin/views/display/DisplayPluginBase.php).`

`Notice: Undefined property: Drupal\views\ViewExecutable::$0 in Drupal\views\Plugin\views\display\DisplayPluginBase->usesExposed() (line 248 of core/modules/views/src/Plugin/views/display/DisplayPluginBase.php).`

There are contrast issues with the form labels on black backgrounds
Screen Shot 2022-04-07 at 10 55 41 AM

I am going to keep testing but wanted to surface these

joewhitsitt:
Looks like it unset the home state: california filter from this views block https://admissions.uiowa.ddev.site/node/4896

joewhitsitt:
Screen Shot 2022-04-07 at 11 25 29 AM

Filters need to respond to layout width

joewhitsitt:
This is unrelated but concerning (https://sandbox.prod.drupal.uiowa.edu/list-block):

Screen Shot 2022-04-07 at 11 31 48 AM

Anyone recall what this block was?

Was the event feature turned on temporarily and a block placed? I am seeing revisions from @bspeare along with Alex and Michael (webmaster masquerading?). The diff shows filters for both people and events

Discussion:
https://iowaweb.slack.com/archives/C0164FNN1PV/p1649350023246039

pyrello:
> Screen Shot 2022-04-07 at 11 25 29 AM
>
> Filters need to respond to layout width

@bspeare created a follow-up issue: #5109 

> Anyone recall what this block was?
>
> Was the event feature turned on temporarily and a block placed? I am seeing revisions from @bspeare along with Alex and Michael (webmaster masquerading?). The diff shows filters for both people and events

It seems likely that I did this when I was testing event list block updates. I removed the broken block from the page.

> Getting some notices on https://sandbox.uiowa.ddev.site/people-directory

It seems like something isn't working quite right. I suspect this is related to another patch we are using being out-of-date with 9.3. I'm going to set this back to draft until I have time to look into it.

pyrello:
This is why I think the patch might not be working anymore: https://www.drupal.org/project/drupal/issues/2605218#comment-14294915

pyrello:
Closing this in favor of #5951.

PR #4483, opened 2021-11-22
Event views block

### LINKED ISSUE #4126: Events list block
Blocked until https://github.com/uiowa/uiowa/issues/4305

# Story

As an admissions editor, I would like the ability to dynamically display local event content within a page.

# Proposed Solution

_Build off the default site with event feature enabled first!_

- [x] Establish new view block in events split.
- [x] Add to card-related LB styles
- [x] Use generalized card list template and preprocess unique fields into the correct place in the card
- [x] Extend to admissions split

# Follow-up

- Part of our strategy to consolidate event functionality https://github.com/orgs/uiowa/projects/1#card-63850273

pyrello:
Resolves #4126.

Adds an event list block for sites with the event feature turned on. In this first iteration, the event block will only display future events and does not allow filtering related to the date.

# How to test

**Update UIDS:**

```
rm -rf ./node_modules
yarn cache clean
yarn install
```

## Event list block
* `ddev blt ds --site sandbox.uiowa.edu`
* `ddev drush @sandbox.local uli`
* As a site editor, I can add an event list block to my page.
* As a site editor, I can hide any of the fields that are available to display.
* As a site editor, I can sort my events list block by date ascending or descending.
* As a site editor, I can filter events by event category, and tags.
* As a site editor, I can set block styles and the displayed event match my selections.

### Check for regressions with existing list blocks
* People and articles list blocks show the same default style options as they currently do.
* People and articles list blocks can still select the order of sorts via table drag utility.
* Generally spot check functionality to ensure that it still looks and functions as expected.

### Admissions
* As an admissions editor, I can use the event block as described in the previous stories.

## Card template refactoring   

As a part of this pr, we have updated the card template to create a `card_meta` block within the template to handle the meta information under each card subtitle.   This updates existing card templates to remove `card__date` and `card__label` classes and variables to use the broader  `card__meta` class.    This changes the card classes from:

```
[card__media]
[card__title]
[card__subtitle]
[card__date]
[card__label]
[card__body]
```
to:
```
[card__media]
[card__title]
[card__subtitle]
[card__meta]
[card__body]
```


bspeare:
I noticed a few things that weren't working that might be helpful to compare against the tests:

The hide fields option for location and virtual event are not working:

Screen Shot 2021-12-06 at 10 50 13 AM

The only image option available is the circle:

Screen Shot 2021-12-06 at 10 53 41 AM

And the default card options for the Card Block look off:

Screen Shot 2021-12-06 at 11 23 01 AM

pyrello:
Options for event list and default card options are fixed now. Still need to look at why the hide option for virtual and location are not working.

pyrello:
Virtual and location fields are now working as expected.

I am noticing some weird behavior with the date filter in the block form:
Screen Shot 2021-12-16 at 3 41 46 PM

It appears that whatever mechanism causes the show/hide when selecting the different operator options is not functional. I'm not sure if this is it, but I notice that the `states.js` files is not being loaded. I tried adding it directly to the form, but it is still not loading. Not sure what to do here.

Another thing that I think we might want to do here is add a datepicker for any date fields. It is weird to have to type in the date and to need to know what format it should be in.

joewhitsitt:
>Another thing that I think we might want to do here is add a datepicker for any date fields. It is weird to have to type in the date and to need to know what format it should be in.

if the input type can be set to `'#type' => 'date',` this is supported by most browsers now. https://github.com/uiowa/uiowa/pull/4549#issuecomment-989294236

[better exposed filters](https://www.drupal.org/project/better_exposed_filters) i think does this, but there are issues with their jquery based solution.

pyrello:
When syncing the admissions site, I am not noticing any sync conflicts. It seems like we might not need to do anything to accommodate that site split at this point. @joewhitsitt Can you think of anything I might not be thinking of in that regard?

joewhitsitt:
I think the IS was written with CS1 in mind. If the event split is turned on for Admissions and the block works, I can't think of anything off the top of my head.

pyrello:
I was able to come up with a temporary workaround to hardcode a fix for states specifically for this. Hopefully we can remove this when we move the a generalized solution for filters using `ctools_views`.

pyrello:
Did a similar fix for showing date fields instead of textfields. It makes me uncomfortable putting such brittle code out there.

pyrello:
Just realized there is one other change I need to make. Putting back in draft.

pyrello:
So... I realized after I starting working with the date filter that this adds a lot of complexity and that the way I was trying to handle it didn't really make any sense. If we are going to allow filtering by date, the most likely scenario is using a relative date (e.g. any events between now and 2 months from now). I decided to not get into that right now and just set a non-exposed filter that filters out events that are in the past. The user can still limit the number of events displayed by setting the number of items, so this seems like a reasonable starting point. We can get into more complex filtering options, but I think that should be done in future work and with input from the rest of the team and our partners about what they want.

GaryRidgway:
I am going to try and pull this... wish me luck

GaryRidgway:
I will use this comment as a notepad for things I find confusing:
### Things i find confusing
- The "Article summary" field toggle is actually toggling the event short description. I recommend changing this from "article summary" to "Event short description"
- The field toggles have a very different naming design than the articles and people ones, should these be standardized?
Event
Screen Shot 2021-12-21 at 3 36 57 PM
People
Screen Shot 2021-12-21 at 3 36 44 PM
Articles
Screen Shot 2021-12-21 at 3 43 54 PM

GaryRidgway (approved):
I have tested everything and found it all to work on both sandbox and admissions. I have found no regressions.

Additionally there were some QOL improvements that i thought could be made, and Sean made them as well. Thank you @pyrello!

This looks great, APPROVED!

PR #4485, opened 2021-11-23
Remove blt:init:settings from sync.commands.

richardbporter:
https://github.com/acquia/blt/issues/4428

# Testing
- On main, run `blt sync --site sandbox.uiowa.edu`
- See that the output from Drush is syncing the default site, not sandbox.
- Check out this branch and run it again.
- See the correct site is synced.

We could leave this out permanently as it doesn't really need to run per-site for every sync. People would just need to remember to run `blt bis` to regenerate local.settings.php files when necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ multisites:
 ```
 
 ### Common Tasks
-Multisites will not be able to bootstrap without a `local.settings.php` file. The `blt:init:settings` (or `bis` for short) command will generate local settings files _only_ for the multisite defined in BLT configuration. By default, this is all multisites but be aware that the `local.blt.yml` change documented above will override that.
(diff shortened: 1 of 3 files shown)


joewhitsitt:
pulling down to test

joewhitsitt (approved):
not going to pretend to know how/why the no-op addition but this works as described. do we need to update docs as part of this so folks get muscle memory with new sites, new setups?

richardbporter:
> not going to pretend to know how/why the no-op addition but this works as described. do we need to update docs as part of this so folks get muscle memory with new sites, new setups?

Yea probably. I can do that.

quamsta:
Ran into the blt problem this morning, tested this branch and it seems to work as expected. Guess this means bacon, lettuce, and tomato is back on the menu?

richardbporter:
@joewhitsitt What do you think of those changes?

joewhitsitt:
Looks fine to me.

joewhitsitt:
Crud. I tried ds sandbox from main. `Could not look up database name for @self`

joewhitsitt:
Duh Joe! read the docs...

PR #4486, opened 2021-11-23
Added some help text to the content field on the block quote component

quamsta:
Discussed this with @bspeare and @ABoston88, thought it might be a good idea to help alleviate confusion and maintain consistency across campus about whether or not to use quotation marks in the new block quote component. 

APA style guide says no quotation marks in block quotes, so I added some instructions (with the help of Ben and Alex for phrasing) to help content editors.


bspeare (approved):
I think this is useful supporting text.  Thanks for making this change!

PR #4495, opened 2021-11-29
Find Text updates--menu searching and new placement

### LINKED ISSUE #4348: SiteNow Find Text UX improvements
# Motivation

The Find Text feature is a tool rather than a configuration setting, so placing it in the context of the SiteNow configuration menu doesn't make a lot of sense, since you are not changing configuration when you use it. Also, I think we are a little over the top in wanting to name things branded with "SiteNow" and that this tool could just be called "Find Text" instead of "SiteNow Find Text." I think this would potentially reduce confusion for people using the tool without other context because the "SiteNow" in the name is not meaningful in relation to how it works.

# Proposed solution

- Change the menu and page title of the tool to just "Find Text"
- Move the menu item to be a top-level Admin menu item (as a sibling to "Content" and "Structure")

### LINKED ISSUE #4445: SiteNow Find Text - Main Menu links support
Epic: https://github.com/uiowa/uiowa/issues/4268

This is the next _first_ slice of extending SiteNow Find Text functionality

## Proposed Solution

- [ ] Working off the existing code, extend to provide search support to menu link titles and path to catch the common *.prod.drupal.uiowa.edu menu items.

cory-skeers:
Resolves #4348
Resolves #4445 

# How to test

- `git checkout find-text-menus && git pull`
- `blt ds --site=sandbox.uiowa.edu` or sync any other site
- (Permissions haven't changed, but for full testing experience, the following should be completable by a webmaster).
- Use the "Find Text" toolbar link
  - It should appear in the admin toolbar, labeled "Find Text" with an appropriate icon
- Search for content in the site
- Check that you can still find content within a page, either on node fields or in layout builder blocks/paragraphs, and that the results table displays the correct node
- Check that the `edit` and `layout` links work for nodes
- Search for content that should match a menu link title or uri
- Check that the menu appears in the results table, labelled as a menu item
- Check that the `edit` link takes you to editing the menu item that matches the search string
  - After saving from the menu link, you should be redirected back to the Find Text page (though the previous results do not persist through this)
- Check that the Find Text page's help text, link to documentation work correctly, as some changes have been made as part of this as well

## Edit
 - Try using both the "render" option and not
 - Check that highlighting works for both rendered/nonrendered outputs
 - Search for strings that might match text within HTML tags
   - Rendered output should still function. Good example of this might be searching for "table" and checking that table results still render as tables.
     - There might be results without highlighting -- this would indicate that it matched only content within tags
   - When not using rendered output, tag should appear complete, but with added highlighting 
 - For all of the above, check that labels appear correctly (bolded, text makes sense for the field it refers to)

### DIFF
--- a/docroot/modules/custom/sitenow_find_text/css/sitenow_find_text.icon.css
+++ b/docroot/modules/custom/sitenow_find_text/css/sitenow_find_text.icon.css
@@ -1,5 +1,7 @@
-.toolbar .toolbar-icon.toolbar-icon-sitenow-find-text-search-form::before {
-  background-image: url("/core/themes/claro/images/icons/868686/magnifier.svg");
-  width: 2.25rem;
-  left: 0.1667em;
+.toolbar-icon-sitenow-find-text-search-form::before {
+  font-family: 'Font Awesome 5 Free';
+  font-weight: 900;
+  font-size: 16px;
+  top: 0.67em !important;
+  content: '\f002' !important;
 }
--- a/docroot/modules/custom/sitenow_find_text/css/sitenow_find_text.icon.css
+++ b/docroot/modules/custom/sitenow_find_text/css/sitenow_find_text.icon.css
@@ -1,7 +1,5 @@
-.toolbar-icon-sitenow-find-text-search-form::before {
-  font-family: 'Font Awesome 5 Free';
-  font-weight: 900;
-  font-size: 16px;
-  top: 0.67em !important;
-  content: '\f002' !important;
+.toolbar .toolbar-icon.toolbar-icon-sitenow-find-text-search-form::before {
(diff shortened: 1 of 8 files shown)


GaryRidgway:
All current tests work great! @cory-skeers and I have talked about adding some potential UX improvements to the results portion of the search, so before I approve we will move forward on that.

GaryRidgway (approved):
This now feels incredible, thanks for working with me on all the UX updates that make this feel really great @cory-skeers ! well done!

dnkemper:
big fan of "this now feels incredible" haha

cory-skeers:
Thank you for all the testing and feedback, @GaryRidgway! Really happy with the current state of this!

PR #4496, opened 2021-11-29
Feature: `<a>` tags in event description

### LINKED ISSUE #4085: Events content overflowing in Safari
# Problem
Events info is overflowing into the next event in the next event doesn't have an image in Safari (Version 14.1.2). It's fine in Chrome.

https://tippie.prod.drupal.uiowa.edu/events

Safari:
image
image

Chrome:
image

# Known Issue
https://sitenow.uiowa.edu/known-issues#uiowa-issue-4085

### LINKED ISSUE #4477: As a site owner, I can display an events list with teaser text that allows `<a>` tags
# Story

As a site owner, I can display an events list from events.uiowa.edu with teaser text that allows `<a>` tags so that site visitors can click on register and zoom links without clicking through to the next page. 

# Proposed solution

- Update docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc to allow `<a>` tags by updating `Xss::filter` to include an allowed list or switch over to use `Xss::filterAdmin` which is what the full display uses for better or worse.  
- For reference https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Xss.php/class/Xss/9.2.x

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/4477. 
Resolves https://github.com/uiowa/uiowa/issues/4085. 

# How to test

This pr changes the events module to use the `description` instead of the `description_text` from content hub.

- Test on any site using an events block.   
- You should see the text formatted and links active if they haven't been truncated.   
- Libraries has a lot of links if you set the event count to 20 for testing.

### DIFF
--- a/docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc
+++ b/docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc
@@ -81,8 +81,8 @@ function template_preprocess_sitenow_events_teaser(array &$variables) {
     'all_day' => $variables['event']['instance_all_day'],
   ]);
 
-  if (isset($variables['event']['description_text'])) {
-    $variables['event']['description_text'] = Xss::filter($variables['event']['description_text']);
-    $variables['event']['description_text'] = Unicode::truncate($variables['event']['description_text'], 500, TRUE, TRUE);
+  if (isset($variables['event']['description'])) {
+    $variables['event']['description'] = Xss::filter($variables['event']['description']);
+    $variables['event']['description'] = Unicode::truncate($variables['event']['description'], 500, TRUE, TRUE);
   }
 }
--- a/docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc
+++ b/docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc
@@ -5,9 +5,9 @@
  * Teaser theme functions.
  */
 
-use Drupal\Component\Utility\Unicode;
 use Drupal\Component\Utility\Xss;
 use Drupal\Core\Url;
+use Drupal\smart_trim\Truncate\TruncateHTML;
 
 /**
  * Theme function for an event teaser.
@@ -83,7 +83,9 @@ function template_preprocess_sitenow_events_teaser(array &$variables) {
 
   if (isset($variables['event']['description'])) {
     $allowed_tags = ['a', 'strong', 'em'];
+    $trim_suffix = '...';
+    $truncate = new TruncateHTML();
     $variables['event']['description'] = Xss::filter($variables['event']['description'], $allowed_tags);
-    $variables['event']['description'] = Unicode::truncate($variables['event']['description'], 500, TRUE, TRUE);
+    $variables['event']['description'] = $truncate->truncateChars($variables['event']['description'], 400, $trim_suffix);
   }
 }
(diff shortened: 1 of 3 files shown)


cory-skeers:
Pulling down to test

cory-skeers (approved):
Tested and everything is working well! (Thanks for the extra tip of using libraries as a test case--very easy to find some links there.)

Both the enhancement and bug listed have been resolved. 💯

bspeare:
@cory-skeers thanks for testing!  I made one last minute update to only allow certain tags.   Do you mind taking another glance and let me know if you see any issues with those code changes?

joewhitsitt:
Another thing to consider is that the HTML tags count towards the 500 character truncate and may potentially cause broken HTML...

cory-skeers:
@bspeare Looking at the extra change, I think everything looks good! I can pull it down again to make sure, but I'll be a little bit.

cory-skeers:
> Another thing to consider is that the HTML tags count towards the 500 character truncate and may potentially cause broken HTML...

Ooh...also a good point to think about.

bspeare:
Moving this back to draft to see if we can figure out a solution that uses `smart_trim`.

joewhitsitt:
Yay! I think this smart_trim use will work nicely. I added the following description to test with after the xss filter.

```php
$variables["event"]["description"] = 'Nulla neque dolor, sagittis eget, iaculis quis, molestie non, velit. Donec vitae sapien ut libero venenatis faucibus. Fusce ac felis sit amet ligula pharetra condimentum. Curabitur vestibulum aliquam leo. Vestibulum suscipit nulla quis orci. Cras risus ipsum, faucibus ut, ullamcorper id, varius ac, leo. Nunc nulla. Nullam sagittis. Praesent vestibulum dapibus nibh. In <strong><a href="https://google.com">consectetuer turpis ut velit. This is my longer link that starts before the truncate cutoff</a></strong>';
```

And this is what I got. Nicely corrected HTML and it doesn't seem to count markup against the character count.

Screen Shot 2021-12-01 at 3 24 25 PM

docroot/modules/custom/sitenow_events/includes/sitenow_events.teaser.inc
```diff
@@ -81,8 +81,14 @@ function template_preprocess_sitenow_events_teaser(array &$variables) {
...
-    $variables['event']['description_text'] = Unicode::truncate($variables['event']['description_text'], 500, TRUE, TRUE);
+  if (isset($variables['event']['description'])) {
+    // Limit what html is allowed.
+    $allowed_tags = ['a', 'strong', 'em'];
+    $variables['event']['description'] = Xss::filter($variables['event']['description'], $allowed_tags);
+
+    // Use smart_trim's helper to safely truncate without breaking HTML.
+    $truncate = new TruncateHTML();
+    $trim_suffix = '...';
+    $variables['event']['description'] = $truncate->truncateChars($variables['event']['description'], 400, $trim_suffix);
```
cory-skeers: Was the change from 500 characters to 400 intentional?

bspeare: Good point! I think I forgot to commit that change back.

bspeare: I changed it back to 500, are we good to merge this in?

cory-skeers: I think so!

PR #4497, opened 2021-11-29
Fix newrelic appname.

richardbporter:
The `$site_dir` variable was removed in #4467 so individual sites are not reporting in New Relic. This PR updates the newrelic snippet to use BLT's new `$site_name` and also check if it is set in the first place.

### Testing
- Review code.


pyrello (approved):
Reviewed, not tested.

PR #4523, opened 2021-12-02
Grad admissions site split

### LINKED ISSUE #4475: Grad Admissions Cost Estimate
- [x] Estimated costs for domestic and international students for Graduate Programs. Tied to an area of study. Similar fields to existing Grad Admissions site (year and cost estimate for US as well as year and cost estimate for international).
- [x] Multiple grad areas of study could reference a single cost estimate.
- [x] The cost estimates would be rendered as links on the grad area of study (not outputting all info on the grad area of study).
- [x] Admissions likes the tabbed interface to switch between domestic and international estimated costs but are open to other ideas that make sense.
- [x] There won't be a list of all cost estimates, users would need to navigate to the grad area of study to find the cost estimate specific to that degree option.
- [x] Add entity reference field and cost estimate info field to AOS
- [x] URL path: /costs/title

### LINKED ISSUE #4476: Grad Admissions Colleges taxonomy
Each Area of Study would have the ability to reference one or more Colleges (similar to main Admissions site), and those Colleges would be output as a link on the Area of Study. The Colleges should also be able to contain information such as an image, description, link and display list of referenced areas of study.

### LINKED ISSUE #4521: Establish grad admissions site split for future aos work
Create site split for grad.admissions.uiowa.edu for epic work. 
Make sure users will have aos feature split perms and make sure it's turned on for following epic issues

### LINKED ISSUE #4568: Grad Admissions Areas of Study taxonomy
Areas of Study taxonomy (holds `see also` info and allows multiple pieces of content to show up as one row on the view display)

dnkemper:
Closes #4521 
Closes #4475 
Closes #4476 
Closes #4568

# How to test

1. checkout branch
2. blt ds --site=grad.admissions.uiowa.edu
3. composer install
3. drush @admissionsgrad.local en tvi
3. drush @admissionsgrad.local en grad_admissions_core
4. drush -l grad.admissions.uiowa.edu config:import --source ../config/sites/grad.admissions.uiowa.edu --partial
5. drush @admissionsgrad.local cim
7. blt frontend
8. . drush @admissionsgrad.local cr
6. drush @admissionsgrad.local uli
7. Add an estimated cost (http://admissionsgrad.uiowa.ddev.site/node/add/costs_estimate) and click "US Citizens & Permanent Residents". Select ‘provide a menu link’ -> Select ‘Cost and Funding’ -> Select ‘Accounting’ disabled link I added to prod. Save.
8. Add another estimated cost, follow same steps except select ‘International’. Save
9. Add a college. Go to http://admissionsgrad.uiowa.ddev.site/admin/structure/taxonomy/manage/colleges/add
8. Add an area of study (http://admissionsgrad.uiowa.ddev.site/node/add/area_of_study) and add an area of study (ex accounting) a program (Accounting (MAc), reference college you created, an estimated cost intro and reference estimated cost you created (SHOULD ONLY DISPLAY DOMESTIC - Show up as Accounting, Estimated Cost or whatever title you added)
9. Make sure the college displays as a link on aos. Click link and make sure college page displays the aos that referenced it and looks purty .
10. If college looks decent, go back to the aos page you created and make sure the estimated cost you references shows up as a purty button -> click said button and make sure it takes you to domestic cost page.
11. P.S. Also make sure there's a tabbed menu that shows the international estimated cost page you added and can switch over to it
10. Add another area of study and make sure can add same estimated cost


docroot/sites/grad.admissions.uiowa.edu/modules/grad_admissions_core/grad_admissions_core.module
```diff
@@ -0,0 +1,6 @@
+<?php
+
+/**
+ * @file
+ * Contains grad_srop.module.
```
joewhitsitt: can this be updated?

dnkemper: Whoops. Yes. Sorry about that... was using it while talking through config split scenarios.

joewhitsitt:
pulling down to test, but I think we want to hold off on merging this until we know we can build what we need with the AoS split or if we need something entirely custom.

dnkemper:
> pulling down to test, but I think we want to hold off on merging this until we know we can build what we need with the AoS split or if we need something entirely custom.

100%. Thank you, Joe! Also, sorry for going about it this way... wanted to get a feel for how config splits are going to work now and saw this as an opportunity to learn 😬

joewhitsitt:
I don't think this is working correctly and I am a little fuzzy now that we are on CS2.

Following the instructions above, the site split isn't registered to enabled through the UI. Historically it has been by letting the site know about the new config directory: https://github.com/uiowa/uiowa/tree/main/config#site-split

Enabling the module instead, the module install file has an update hook instead of an install hook to enable/import the AoS feature. The update hook will only run if the module is already enabled and sees this as an update.

I think the steps would be like this (creation/testing) until it is established in prod:

- Turn the module on (assumes we will need the module for something else later on)
- The module uses an install hook instead of an update hook to enable/import the AoS feature.
- Make at least one customization to the AoS feature (remove location taxonomy/field?) or remove permissions so site owners can work with this until we are ready to turn it on?
- The site split adds the module as a complete split item as well as the modification(s) above.

I can help square this away next week.

joewhitsitt:
Removed the patch permissions until grad.admissions goes custom (currently handled by AoS split when enabled). Changed the update hook to install and updated the instructions to get everything firing in the correct order.

>I think we want to hold off on merging this

Still stand by this until we are comfortable with site editors entering data or tell them that we will be wiping data before migration.

joewhitsitt (dismissed):

dnkemper:
Menu patch didn't work 😭

pyrello:
> Menu patch didn't work 😭

How are we determining that? It worked for me.

dnkemper:
PLEASE TELL ME IT IS JUST MY ENVIRONMENT
From: Sean Adams-Hiett ***@***.***>
Sent: Tuesday, December 21, 2021 5:45:01 PM
To: uiowa/uiowa ***@***.***>
Cc: Wisnousky, Danielle N ***@***.***>; Author ***@***.***>
Subject: [External] Re: [uiowa/uiowa] Grad admissions site split (PR #4523)


Menu patch didn't work 😭

How are we determining that? It worked for me.

dnkemper:
After a fresh sync and clearing my browser cache, when enabling the menu link, it did nothing (for cost estimate content)

pyrello:
Hmm... it is no longer working for me either. Which is a bummer :|

dnkemper:
DANG

docroot/sites/grad.admissions.uiowa.edu/modules/grad_admissions_core/sass/area-of-study.scss
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+.block-field-blocknodearea-of-studyfield-area-of-study-requirements,
+.block-field-blocknodearea-of-studyfield-area-of-study-procedures,
+.field--name-field-area-of-study-deadlines,
+.field--name-field-area-of-study-requirements,
+.field--name-field-area-of-study-procedures {
+  // @include breakpoint(sm) {
+  //   display: flex;
+  // }
```
pyrello: Can this be removed?

bspeare: Yes, I can clean that up.

joewhitsitt:
Pulled this down yesterday and noticed a couple things.

- I think AoS subtitle should be "minimal" like it was done for Admissions. Might be worth looking to see if other long text fields can be limited as well.
- Need to customize/split what SEO options should be available for these new content types. (e.g. schema person, etc. should not)
- The taxonomies are all [public](https://grad.admissions.uiowa.edu/content/accounting) but should they be? Should we rabbit hole AoS and degree type terms?

I don't know the alternative solution quite yet, but the "content overview append estimated cost type" doesn't do a whole lot to improve. The menu title will still be the title, appending on view render won't allow "international" to be filterable. I would think a presave that combines the faux title and the type together would make more sense (allows filtering in content overview). The menu title would still have to be set manually I would think since it isn't required and kicks off a separate process to create the menu link content.

dnkemper:
> Pulled this down yesterday and noticed a couple things.
> 
> * I think AoS subtitle should be "minimal" like it was done for Admissions. Might be worth looking to see if other long text fields can be limited as well.
> * Need to customize/split what SEO options should be available for these new content types. (e.g. schema person, etc. should not)
> 
> I don't know the alternative solution quite yet, but the "content overview append estimated cost type" doesn't do a whole lot to improve. The menu title will still be the title, appending on view render won't allow "international" to be filterable. I would think a presave that combines the faux title and the type together would make more sense (allows filtering in content overview). The menu title would still have to be set manually I would think since it isn't required and kicks off a separate process to create the menu link content.

You can change the menu titles and we will be doing the content so shouldn’t be a problem since selecting a specific area of study to relate to as I have demoed to Scott. International should not be filterable. Only domestic. I set it up that way for a reason. Totally agree on minimal since the edit form is ridiculous. Slightly confused… the titles should be accounting estimated cost and then if you select domestic or international it should make the content overview easier to distinguish as ‘accounting estimated cost - domestic’ and vice versa. As Scott agreed when demoing it. Everything was agreed upon Wednesday but I guess not being here doesn’t give me much sway on things. BUT if you have any ways to make it easier I’m totally okay with that. Just know we are doing the content and will be providing instructions. Last thing… that’s why I had the menu force module - to make the menu title required but if there is an alternative solution without enabling a module that would def make life easier

joewhitsitt:
@dnkemper, I didn't mean to disrupt your time off. Enjoy gosh darn-it. My comment were patched together "notes to self" thinking nobody else was around to hear.

>You can change the menu titles and we will be doing the content so shouldn’t be a problem since selecting a specific area of study to relate to as I have demoed to Scott.

Yes, if that workflow was approved, that is fine. As I said, "The menu title would still have to be set manually I would think since it isn't required"

>International should not be filterable. Only domestic. I set it up that way for a reason.

I think you misunderstood. I was talking about the content overview page (all content that you can filter on). If you are just appending the value to the already processed view, filtering on title will not bring back results for these.
Screen Shot 2021-12-29 at 2 45 34 PM

Example story: As an editor, I want to find and update all international cost estimates as I was just sent new information for next year:
Screen Shot 2021-12-29 at 2 45 21 PM

The content overview logic doesn't work for "[Graduate Programs in Liberal Arts & Sciences](https://grad.admissions.uiowa.edu/graduate-programs-liberal-arts-sciences-costs)" because the title is stored with `&amp;` instead of `&`

Screen Shot 2021-12-29 at 2 54 07 PM

I won't touch this part of the PR until your return when we can discuss it.

joewhitsitt:
I am going to be pushing a fix for the Areas of Study view page. Thanks to @briand44 for helping me with the relationship/aggregation.

Currently, results are being aggregated and not all nodes (degree types) are being represented for an AoS term.
Screen Shot 2021-12-29 at 3 10 01 PM

Here is the content causing the issue. Accounting (term) should have multiple degree options (MAc, PhD) which each linking to a separate node.

Screen Shot 2021-12-29 at 3 10 21 PM

The fix was to change the base view from node to taxonomy ([terms are not duplicated like currently](https://grad.admissions.uiowa.edu/graduate-areas-of-study)) then do a look up of all AoS nodes and their degree types and piece them together. This also simplifies bringing a single "see also" snippet per term which is added using the term description. With relationships and aggregation, the filters are available without duplicates.

Screen Shot 2021-12-29 at 3 12 16 PM

joewhitsitt:
It was discussed in parking lot that the minimal slice for the taxonomies is to keep them public like they currently are until more feedback is obtained. The rabbit hole alternative has additional SEO pieces to consider.

Screen Shot 2021-12-30 at 11 53 54 AM

joewhitsitt:
Screen Shot 2022-01-03 at 8 49 50 AM

Depending on the path forward for the estimate type indicator, I think this should be changed to `$node->get('field_grad_costs_select)->value;`

pyrello:
> Depending on the path forward for the estimate type indicator, I think this should be changed to `$node->get('field_grad_costs_select)->value;`

This is equivalent to `$node->field_grad_costs_select->value`, I believe. If we are going to worry about not using magic methods because PHPStorm can't interpret them correctly, we should maybe discuss and adopt this as a best practice so that we are doing it consistently throughout the code. Currently we use both forms in a lot of different places.

pyrello:
FYI, I am working to resolve some of the issues above. @joewhitsitt has convinced me that what we were doing originally with handling the storage of the title similarly to the person content type is actually the correct way to handle it. Apologies to @dnkemper for trying to take us down a different path with regard to that.

pyrello:
I ran into the issue with the `menu_link_weight` module causing a javascript error. I switched to commit hash for that module because there is a fix for the issue in the 1.x-dev branch, but no release has been created yet.

pyrello:
> The taxonomies are all public but should they be? Should we rabbit hole AoS and degree type terms?

It sounds like we are going to sort this out, but I don't think it should block use finishing this particular PR and could be added as a follow-up task to some of the upcoming work.

pyrello (approved):
I have tested the functionality and it all appears to work as it is supposed to. I reviewed all the file changes and everything appears to be in order.

PR #4547, opened 2021-12-07
DDEV Fixes

### LINKED ISSUE #4539: DDEV - Default site syncing to wrong db
# Problem

Even with a recreated local.settings.php file the default site seems to try and sync to uiowa_edu instead of uiowa causing a config diff error.

`You will destroy data in uiowa_edu and replace with data from ....`

If you delete the uiowa_edu db, the sync errors with `Could not look up database name for @self`

# Proposed solution

?

joewhitsitt:
Fixes https://github.com/uiowa/uiowa/issues/4539

This can be solved manually by deleting the local.drush.yml file in your default site directory

## To Test

- Comment out multisites array in local.blt.yml
- run `blt bis`
- run `blt ds` the database name should be `uiowa` and sync successfully

This deletes and replaces local.drush.yml files. Though the uri in the replacement file is the same for each site, it just works.

### DIFF
--- a/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
+++ b/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
@@ -190,6 +190,7 @@ class ReplaceCommands extends BltTasks {
     $this->taskExecStack()
       ->dir($this->getConfigValue('docroot'))
       ->exec('rm sites/*/settings/local.settings.php')
+      ->exec('rm sites/*/local.drush.yml')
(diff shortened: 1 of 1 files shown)


pyrello:
Testing.

pyrello (approved):
Tested. This regenerates the `local.drush.yml` file per site, but it has the new URL format. Also `blt ds` ran without issue.

makurj:
Seems to work okay. 
image

PR #4546, opened 2021-12-07
DDev followup: use `drupal:tests` in CI for functional JS tests and reconfigure/document local testing.

richardbporter:
Related to #4538 

This PR removes the BLT configuration required for the ddev chromedriver container to work from our `blt.yml` file. The reason for this is so we don't have to then override those values to the defaults provided by BLT in `ci.blt.yml`. The other benefit is should those values change in BLT, we'll get those updated values.

This change means it is necessary to specify more values in your `blt/local.blt.yml` files if you'd like to run Drupal tests locally. I've updated `tests/README.md` to reflect this.

Additionally, this PR changes how our tests our run in CI. We now use the `tests:drupal` command to run all phpunit tests and configure phpunit to use our repo config. This ensures the necessary chrome and chromedriver requirements for functional JS test.

### Random
Added `ddev phpcs`, `ddev phpcbf` and `ddev phpunit` commands because I got sick of typing `./vendor/bin/...`.

### Testing
- Reconfigure your `blt/local.blt.yml` file according to the `tests/README.md` instructions to run Drupal core/contrib test.
- Run `blt tests:drupal` and make sure it works.
- Run `blt tests:phpunit` and make sure the uiowa_search functional JS test located at https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_search/tests/src/FunctionalJavascript/SearchTest.php is not marked as skipped.
- Make sure the same test is not marked as skipped in CI log output. 
  - Previous example of skipped test: https://app.travis-ci.com/github/uiowa/uiowa/builds/243121493#L2168
  - Latest build (could change): https://app.travis-ci.com/github/uiowa/uiowa/builds/243128919#L2171

PR #4554, opened 2021-12-08
Find Text - basic permissions test

### LINKED ISSUE #4589: Taxonomy term content search in SiteNow Find Text
# Problem
Taxonomy content is not currently visible to the Find Text module (ie a search for "Space" won't return results for taxonomy terms labeled "space" or that include "space" in their description).

# Proposed solution

Increase the scope of `sitenow_find_text` to include text-based fields which are used by taxonomy terms.

cory-skeers:
Adds additional functionality to the Find Text module (taxonomy term searching) as well as several tests to aid in future development.

Resolves #4589 

# How to test

- `git checkout find-text-tests && git pull`
- Manually run the `sitenow_find_text` tests that have been added
   - https://github.com/uiowa/uiowa/tree/main/tests
   - All tests should pass. For extra thoroughness, you can "break" functionality and check that the tests no longer work. For instance, rename the page permission in the permissions yml, break a case statement in `get_text_field_tables` and see that the related entity tests break, etc.
- Check that the latest Travis CI build passed successfully
- Sync a site that has taxonomy terms, or create some new taxonomy terms
- Try searching for terms that will match taxonomy term names and/or descriptions, see that they appear in the results
- Try searching for a menu and node content, and see that they still appear in the results


richardbporter:
Nice!

Right now, every test of ours needs to be defined in `phpunit.xml.dist` as a `testsuite` so it runs whenever `phpunit` is run without a path argument: https://github.com/uiowa/uiowa/blob/20e013b8729c4b0e65330d55718e6a3ebc4710f4/phpunit.xml.dist#L25-L41

Also, I think this test will be skipped in our CI environment until #4546 is merged.

cory-skeers:
> Right now, every test of ours needs to be defined in `phpunit.xml.dist` as a `testsuite` so it runs whenever `phpunit` is run without a path argument:

Awesome--thanks! Will likely have some questions about that and reach out. Still need to tweak the second test I added to get it working.

GaryRidgway:
I had all of the tests fail for me sadly.

cory-skeers:
Not sure why, but they're failing on `$this->drupalLogin($this->user);` now, which hadn't been a problem before. But since they're all dependent on a user being able to log in and see the page, they all use this.

cory-skeers:
When `$this->drupalLogin($this->user)` runs, it's failing with the exception `"Button with id|name|label|value \"Log in\" not found."`

joewhitsitt:
How might this test be different than PeopleBlockViewTest and ConfigSplitTestTrait in terms of login?

cory-skeers:
> How might this test be different than PeopleBlockViewTest and ConfigSplitTestTrait in terms of login?

Just tested PeopleBlockViewTest locally and it failed on the `drupalLogin` as well.

Changing `simpletest-base-url:` to 'https://default.uiowa.ddev.site' gets the user login working correctly, but leads to failures accessing other pages later in tests...

cory-skeers:
Reverting ae883997e7235407a4ff2505e7a2d23c0ad16cfb, the tests pass, so it looks to be related to the change to `apache`.

cory-skeers:
Updated the tests README with different "simpletest-base-url" value that should allow for these and other tests to be run locally.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
`OK (5 tests, 56 assertions)`

Tested on Admissions and it all checks out

Possible future enhancements:

- limit to specific type
- result stats like "3 content instances, 2 taxonomy instances, 1 menu instance"

both would help encourage the user to keep scrolling to the end

cory-skeers:
Thanks @joewhitsitt!

PR #4555, opened 2021-12-09
card link indicator

### LINKED ISSUE #4372: Add a visual indicator to article teasers that there is more content upon clicking/tapping the article
Similar to https://github.com/uiowa/uids/issues/460, but more specific to articles and article teasers.
# Story

As a user I want to be able to see that an article is clickable and has more content than what the teaser is showing so that I don't miss important details in a longer article.
<details>

image_from_ios

</details>

# Proposed solution

# First Slice 

## Card block 

- Add circle button option if link text is not provided: https://github.com/uiowa/uiowa/issues/4372#issuecomment-962091338
  - Rework logic in https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/uids/card.html.twig#L108-L122 to display new button
 - Add a modifier class style adjustments to current button style for circle button in https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/card.scss
 - Search for all uses of `card` and test: https://github.com/uiowa/uiowa/search?q=card.html.twig
   - As you can see we have 25 templates using the card template such as views templates and custom modules. 
   - V2 Also uses this card template.  

# Second Slice

## List blocks (People, Articles)

- Add checkbox to display a button below each item
  - If box is checked, allow user to add custom link text like "Continue Reading"
  - If no link text is provided, then circle button will be default

Original solution:
<details>
I think adding any of the following to an individual article teaser may help alleviate this:

- Underlined link in the headline indicating that it's clickable (may be less visually appealing, but more functional)
- A "Continue Reading" button (may have accessibility issues, but could be done in a CSS psuedo-class, e.g. `:after { content: "Continue reading" }` to prevent screen readers from repeating "Continue reading" over and over).
- An ellipsis character after the body of the teaser (may be the most elegant, probably also could be done in a pseudo-class
</details>

GaryRidgway:
# How to test

## Circle button
- `blt ds --site=sandbox.uiowa.edu`
- Go to http://sandbox.uiowa.ddev.site/cards and run through the circle button usage in the white, gold, gray, and black backgrounds
  - Verify that the circle button appears if the card has a link but not link text
  - Verify that the circle button hover/focus styles are the same as the regular button, for example the hover border is white on a gold background for both the regular button and circle button
  - Review the code changes and verify that the circle button will only appear if  `'visible_button': true,` is set from the template that is extending the `card.html.twig` template like on https://github.com/uiowa/uiowa/blob/140a6a7b2f8bd26146ad85aee816a4c2abf1cd91/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-card.html.twig#L56
- Check on a v2 site that uses cards

## Random color contrast button issue
- Go to http://sandbox.uiowa.ddev.site/banners-and-buttons in a private browsing window and verify that the banner using a black background on the section and the banner has correct color contrast. 

Screen Shot 2021-12-14 at 11 40 14 AM

Screen Shot 2021-12-14 at 11 39 59 AM

## Check that card template refactor did not break usage on Admissions AOS:
- `ddev blt ds --site admissions.uiowa.edu`
- View the "Admission Process" section on http://admissions.uiowa.ddev.site/academics/marketing to see that the cards in that section do not have buttons.


bspeare:
I'm working on getting the button hover/focus working with background color combos this morning and after that I will put this up for review.

cory-skeers:
Pulling down to test!

cory-skeers:
I'm wondering if the left two cards here (third row of cards, if you look on http://sandbox.uiowa.ddev.site/cards) are intended to be different colored fills:

image

Which...now I realize is the same behavior as the text buttons. It's the difference between specifying a white background and choosing None. So I think it's intentional, and at least, it's nothing that's introduced in this work.

Carry on!

bspeare:
Thanks @cory-skeers!  Yes, the border changes a bit based on the bg color settings.  

The button border color without the background color is not gold so that it adapts to any background color that is added to the section.

cory-skeers:
Not sure which is the intended behavior here, but one difference between text or no text. Appears to only be black card backgrounds on a gold section background. The text-based button gets a white border on hover, where the circle button gets a gold border on hover (they both have white borders prior to hover).

image
image

bspeare:
@cory-skeers I updated it so both have gold when hovered.   The border around each should remain white.

cory-skeers:
> I updated it so both have gold when hovered. The border around each should remain white.

Looks good!

I've gone through the styling and options in v3, and all looks good. Checked two v2 sites with cards and appeared unchanged. Have also checked the banner contrast issue, and saw that it's been fixed.

Still to review:

- [x] Review the code changes and verify that the circle button will only appear if 'visible_button': true, is set from the template that is extending the card.html.twig template
- [x] Check that card template refactor did not break usage on Admissions AOS

cory-skeers (approved):
Reviewed and tested! 💯

bspeare:
@cory-skeers Thanks!

PR #4557, opened 2021-12-13
Trying out object handling patch for Config Split

pyrello:
# How to test


pyrello:
We are moving to a newer version of Config Split in #5625. At some point we will want to revisit this, but I'm going to close this and we can open a new PR when that work bubbles up again.

Issue #4468, opened 2021-11-19
Convert aggregator blocks to ListBlock and improve no results behavior.


richardbporter:
In reviewing https://github.com/uiowa/uiowa/pull/4106, it seems like that work would be much easier if the block was provided programmatically. We didn't do that initially because we didn't have a way to use the headline field in code - now we do. This has the added benefit of moving the RSS Feeds out of the LB 'Add Component' dropdown and into the More... category. 

### Proposed solution
Recreate the block in uiowa_aggregator and then write an update hook to convert all existing blocks to it. Resolve https://github.com/uiowa/uiowa/issues/2417 by adjusting the block form as needed.

richardbporter:
https://github.com/uiowa/uiowa/pull/4561 made me realize we're duplicating a lot of functionality in ListBlock implementations. The aggregator block was built before ListBlock and could probably be converted now.

PR #4561, opened 2021-12-13
Increase aggregator expire time and implement pager.

### LINKED ISSUE #2428: Allow feed items beyond the 6 month/25 max limit
## Issue

Some feeds contain an archive of content that extends beyond the 6 month limit. This would help avoid importing this type of content.

Example Newsletter Archive
https://apps.its.uiowa.edu/dispatch/archives/past/--iowa-law---september-25-lxj144j

https://careers.prod.drupal.uiowa.edu/gaining-experience

## Proposed Solution

Special consideration should be given to database sizes and query performance.

- [x] Remove max number of items limit
- [x] Add a mini pager for feeds with more than X.
- [x] ~~Option to disable default 6 month limit on a per feed basis~~

richardbporter:
Resolves #2428 

### Notes
- This will change the expire time for all feeds on all sites from 6 months to 1 year.

### Stories
- As a webmaster adding an aggregator feed, I can see articles dated past 6 months in the future imported into my site.
- As a webmaster placing an RSS Feed on a page, I can display all items beyond the _Number of items_ limit by using a pager.
- As a webmaster placing multiple RSS Feeds on a page, I can see that pagers work correctly for each block.


richardbporter:
Just noticed that the pager looks wonky in a grid layout like on https://careers.prod.drupal.uiowa.edu/gaining-experience. Could use some help with that.

joewhitsitt:
Can the page element just be moved outside the `aggregator-wrapper`?

richardbporter:
> Can the page element just be moved outside the `aggregator-wrapper`?

Yes thanks!

joewhitsitt:
pulling down to test

joewhitsitt (approved):
Noticed that multiple feeds with different pager config (e.g. display 1, display 5 at a time) seem to work okay because they aren't AJAX. Since they share the same pager id though, clicking through a different one can "erase" the previous pager's progress as it is replaced in the URL. Not the same issue that we encountered with views and less noticeable.

Thanks for figuring out pagers for custom code!

richardbporter:
I'm not sure what you mean. I'm able to page independently:
Screen Shot 2021-12-15 at 3 37 36 PM

richardbporter:
@joewhitsitt demoed the issue to me on Slack. I think this PR removes a blocker for careers and should merge for the short term. Long term solution is #4468 and the pager issue would be fixed by doing that.

PR #4458, opened 2021-11-18
Swap DrupalVM for ddev.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/3641

# DDEV UIowa Setup

_Note: If you don't have a current setup (M1), you will have to follow some onboarding docs to get git and homebrew setup._

- `vagrant halt` on your **current** setup
- [install docker 4.2.0 for mac](https://docs.docker.com/desktop/mac/release-notes/#docker-desktop-420) - _there is a bug with 4.3.0 and [here is the workaround if you find yourself up that creek](https://github.com/drud/ddev/issues/3404)_
- open the docker app and make sure it is "running"
- brew install drud/ddev/ddev
- mkcert -install
- git checkout `ddev` or `main` if this has been merged (hello future)
- if you were testing ddev elsewhere, headover there and `ddev stop --unlist uiowa`
- ddev auth ssh
- ddev composer install
- ddev start
- ddev auth ssh
- ddev ssh
- blt frontend
- blt bis
- blt ds

## Additional Notes

- the new local domain format is *.uiowa.ddev.site, so hr.uiowa.edu is hr.uiowa.ddev.site and gis.sites.uiowa.edu is sitesgis.uiowa.ddev.site
- If you have an existing docker install, make sure it is up to date, but not too far. 4.2.0
- ~~This does not clean up the drupalvm stuff completely.~~ 039c7d4
- You might have to run `ddev mutagen sync` if your containers seem out of sync.
- All Git operations need to happen on the host. BLT commands involving git were updated to require being run on host.
- Checkout the docs for more tidbits: https://ddev.readthedocs.io/en/stable/ (e.g. toggle xdebug, etc.)

## To Test

Try to perform normal local environment tasks:

- `blt frontend`
- `blt ds --site=sandbox.uiowa.edu`
- `drush @sandbox.local cr`

Additional resources:
https://iowaweb.slack.com/archives/C8SM82ZFZ/p1638808178159000

PR #4576, opened 2021-12-16
Hotfix: change ddev webserver to apache.

richardbporter:
Followup from #4458.

# Problem
We're not using apache locally and should be b/c thats what Acquia uses.

# How to test
- Go to http://default.uiowa.ddev.site and see that you are not redirected to https because nginx doesn't use .htaccess.
- Check out this branch and `ddev restart`.
- Go to that URL again.
- ??
- Profit
- Test other stuff.


bspeare (approved):
Redirect to `https` worked after `ddev restart`!

PR #4584, opened 2021-12-17
Hotfix: require UGD be run on host.

richardbporter:
Another oversight of #4458. We cannot run `blt ugd` on the ddev web container because there is no Git repo there.

### Testing
You could run `blt ugd` on your host and then cancel it, or just code review.


bspeare (approved):
This looks handy for the next release!

Issue #4571, opened 2021-12-16
As a webmaster configuring the Dispatch Thank You form, I can use optional placeholders.


richardbporter:
# Story

As a webmaster configuring the Dispatch Thank You form, I can use optional placeholders.

Optional placeholders don't work right with ad-hoc API data. They either render as empty string which is undesirable or the variable substitution breaks the communication. We've reached to Dispatch Support about this.

# Proposed solution

Once fixed, add back the optional placeholders removed in 782376e and 1dca667 as well as other ones specified in the [template](https://apps.its.uiowa.edu/dispatch/help/curatedtemplate/UI%201%20row%20x%201%20col%20-%20Version%202) that might be useful.

PR #4585, opened 2021-12-20
Fix the Thank You footer statement.

richardbporter:
Noticed the Thank You email footer statement is broken while setting up #4127 for pharmacy. This PR fixes it. It also ensure the request URL is always set in the footer statement, otherwise we don't know what site is generating the email. 

### Testing
Fill out the form at https://default.dev.drupal.uiowa.edu/about and check the shared web inbox for the messages. Look at the footer statement.


briand44 (commented):
Tested it and the url of the site showed in the footer of all the messages. Statement about supervisor appeared on the email to the person being thanked.

I did notice that there doesn't seem to be a way for the supervisor or additional recipient to know who was being thanked.

richardbporter:
> Tested it and the url of the site showed in the footer of all the messages. Statement about supervisor appeared on the email to the person being thanked.
> 
> I did notice that there doesn't seem to be a way for the supervisor or additional recipient to know who was being thanked.

Thanks. I'll add that to the footer statement. Thats the easiest place to do it until #4571 is done.

richardbporter:
Dev is updated if you want to take another look @briand44

briand44 (approved):

PR #4588, opened 2021-12-20
Views Pager ID Bug - use cache instead of tempstore

### LINKED ISSUE #3661: People block pager count is broken when set to 1

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/3661

This solution does the same thing as before but captures it as cache for both anonymous and authenticated.

Cache tables have a max so there is less concern of the bin filling up. https://www.drupal.org/node/2526150

I tried to manually purge old dom_ids from the table, but couldn't come up with a "fingerprint" to tie to the cache too without at least temporarily messing up the ajax pager for an anonymous page load.

I set to permanent because some were expiring and then causing issues for cached ajax requests.

## To Test

https://sandbox.uiowa.ddev.site/people-directory should match their view config and work correctly on-click. Pagers should also work for block/page re-save and after `cr`.


richardbporter:
Would it be worth it to push this out to sandbox dev and unleash SIteImprove on it? :)

joewhitsitt:
Sure. Can we trigger SiteImprove manually?

joewhitsitt:
Deploying to DEV and syncing sandbox down to DEV.

https://sandbox.dev.drupal.uiowa.edu/people-directory

joewhitsitt:
This solution might cause complications because our remotes use memcached instead of store cache tables in the db. Not sure how to get visibility into these for the pressure test. All I know is that we have a separate storage limit for this.

joewhitsitt:
Since all dom_ids refresh together, I had a thought about storing all dom_id/pager_id pairs in one item (e.g. current_path) to reduce entries and possibly be enough identifying information to re-write an entry based on the max pager id versus appending to the same entry. I don't know how that affects storage (row entry vs column bulk) or how complex querying this serialized data would be

joewhitsitt:
capturing the slack discussion related to memcache

https://iowaweb.slack.com/archives/C8SM82ZFZ/p1640123286263200

pyrello (approved):
Works as advertised. Not sure if we are still waiting to see what it does on dev.

joewhitsitt:
Testing in dev didn't work out because this stuff isn't stored in the site db. Rick mentioned he wasn't concerned with the solution. Only semi-related, bots might be causing performance issues by hammering the pager links and getting stuck in loops.

Going to merge for now and we can monitor and rollback if needed.

PR #4590, opened 2021-12-21
Hotfix - Null check on region items before block render

joewhitsitt:
## To Test

- Create a region item and assign it to the region pre-footer.
- Delete region item entity.
- Page should not error out.


richardbporter (approved):
Nice! Had to finish my commit to change from last week and wrote a quick unit test. Not the best test, but I was able to replicate the Exception and your change fixed it. Manually tested as well.

PR #4594, opened 2021-12-22
SMTP Test

### LINKED ISSUE #3498: As a site owner, I want generated emails sent reliably without risk of being filtered.
## Issue

There is very little we can do as developers to deal with emails that come from our websites. Our websites are increasingly using webform to create workflows for their organizations and expect email delivery to just work.

>The Cloud Platform email servers are a shared resource, so you can’t send mass emails (such as marketing messages or newsletters). This service is provided as a courtesy and has no uptime or service level agreements associated with it.

There is also a disconnect between emails coming from Acquia and how Office365 processes them causing developers to converse both with Acquia and ITS Email when issues arise.

## Proposed Solution

- [x] Reach out to the ITS Email Team about being our "third-party" mail service provider and discuss steps on making it happen. https://docs.acquia.com/cloud-platform/manage/email/3rdparty/
- [x] Deploy to standard/custom of 701-3 on 01/24
- [x] Dentistry - either use reply_to and change their site mail b/c its a subdomain right now or just implement a much lighter hook_mail_alter to set all from addresses to the noreply address
- [x] icgov
- [x] grad
- [x] clas

joewhitsitt:
Relates to #3498

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -184,6 +184,7 @@
         "drupal/siteimprove": "^1.5",
         "drupal/smart_date": "^3.1",
         "drupal/smart_trim": "^1.1",
+        "drupal/smtp": "^1.0",
         "drupal/stage_file_proxy": "^1.0",
         "drupal/superfish": "^1.4",
         "drupal/swiftmailer": "^2.0@beta",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "60f93d09ac6cc08010eb7668974623e7",
+    "content-hash": "e1645589aa1bafa66db8d7290b7bbf12",
     "packages": [
         {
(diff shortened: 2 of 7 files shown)


richardbporter:
Platform Email is able to send emails from any address of a verified domain. This is better that O365 SMTP which requires Send-As rights for every different from address. Therefore, I think we should close this PR in favor of #4619.

This was a good POC and if a site has a specific need to send emails through O365 from a small number of `@uiowa.edu` email addresses, they could implement it through split configuration and some work with the email team. In fact, it seems like SwiftMailer, which is already in our codebase for brand, can do SMTP.

richardbporter:
>In fact, it seems like SwiftMailer, which is already in our codebase for brand, can do SMTP.

Nevermind. Just noticed that [SwiftMailer](https://www.drupal.org/project/swiftmailer) has been marked obsolete and they recommend [SymfonyMailer](https://www.drupal.org/project/symfony_mailer).

joewhitsitt:
Thanks for reminding me. Created an issue: https://github.com/uiowa/uiowa/issues/4623

PR #4596, opened 2021-12-27
Remove gradient if page title is hidden.

### LINKED ISSUE #4531: Remove banner gradient overlay if page title is hidden
# Problem
If the page title is set to "Visually hide title" on the featured image, the overlay gradient meant to provide accessible color contrast ratios still appears. 

Screen Shot 2021-12-03 at 1 20 15 PM

# Steps to reproduce
Add a featured image to a Page and hide the page title. 

# Proposed solution
- Add a modifier class in `banner.scss` to remove the gradient: `.banner--gradient-bottom.layout--title--hidden::after`

cory-skeers:
Resolves #4531 

# How to test

~`yarn workspace uids_base build`~ (not necessary if you've built off of main recently)
Sync a site
Check a page with a featured image
If the title is not hidden, a gradient overlay should be present on the banner image (`linear-gradient(180deg,transparent 0,rgba(0,0,0,.65) 70%)`)
If the title is hidden, no gradient overlay should be present

### DIFF
--- a/docroot/themes/custom/uids_base/scss/components/banner.scss
+++ b/docroot/themes/custom/uids_base/scss/components/banner.scss
@@ -158,11 +158,6 @@
   font-size: containerClamp(2.5rem, 4.7rem);
 }
 
-// Remove the gradient overlay if the page title is hidden.
-.banner--gradient-bottom.layout--title--hidden::after {
-  background: none;
-}
-
 .banner__image {
   .contextual-region {
     // Remove contextual position for logged in view of banner image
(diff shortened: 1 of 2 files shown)


joewhitsitt:
Don't get me wrong, I like this simple fix, but I also don't like tracking down in the cascade why a class in the markup is not doing as it should.

There is a lot of logic already around this functionality in uids_base_preprocess_layout where this is taking place and `$title_hidden` is already available. I would vote to handle this there.

cory-skeers:
> There is a lot of logic already around this functionality in uids_base_preprocess_layout where this is taking place and `$title_hidden` is already available. I would vote to handle this there.

I'll take another look at this, and check things out. Thanks!

cory-skeers:
@joewhitsitt Okay! This is ready for another look. After going back through it, think this can be covered by a small additional check in the preprocess.

joewhitsitt (approved):
Tested the change. Thanks!

PR #4598, opened 2021-12-28
Fix ReplicatorEvents error

### LINKED ISSUE #4527: Class ReplicatorEvents not found in layout_builder_custom EventSubscriber
# Problem

Acquia cloud tasks spit out an error related to Replicate:

```
Error: Class &#039;Drupal\replicate\Events\ReplicatorEvents&#039; not found in Drupal\layout_builder_custom\EventSubscriber\ReplicateSubscriber::getSubscribedEvents() (line 73 of /mnt/www/html/uiowa02dev/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php).
```

# Steps to reproduce

- Inspect an acquia cloud task (e.g. deployment, db sync, etc.) right after `drush @self sql-sanitize`

# Proposed solution

?

cory-skeers:
Resolves #4527 

Acquia Cloud tasks are throwing an error, not liking the use of the class constant in event name subscriber. In looking through d.o issues and modules, it seems it's an issue with the order classes get loaded, and the workarounds are either to use string constants or check for class existence first (a core example here:  https://api.drupal.org/api/drupal/core%21modules%21content_translation%21src%21ContentTranslationUpdatesManager.php/8.2.x). 

Checking for the class existence first shouldn't effect any functionality, but should circumvent the error.

# How to test

Sync a site, replicate a page.
Check that the layout has successfully been replicated, and that new blocks have been created
(A quick way for the above is to use `drush @site.local sqlc` and run `SELECT count(*) FROM block_content` before and after replicating).

# Followup test

After merging, check tasks in acquia cloud, and see that the error is no longer appearing.

```
Error: Class &#039;Drupal\replicate\Events\ReplicatorEvents&#039; not found in Drupal\layout_builder_custom\EventSubscriber\ReplicateSubscriber::getSubscribedEvents() (line 73 of /mnt/www/html/uiowa02dev/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php).
```

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php
+++ b/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php
@@ -69,9 +69,11 @@ class ReplicateSubscriber implements EventSubscriberInterface {
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
-    return [
-      ReplicatorEvents::AFTER_SAVE => 'onReplicateAfterSave',
-    ];
+    $events = [];
+    if (class_exists('\Drupal\replicate\Events\ReplicatorEvents')) {
+      $events[ReplicatorEvents::AFTER_SAVE] = 'onReplicateAfterSave';
+    }
+    return $events;
   }
 
   /**


joewhitsitt (approved):
tested functionality locally. still works as expected.

PR #4599, opened 2021-12-28
Indication to scroll for timeline

### LINKED ISSUE #4565: Timeline indicator to scroll for content
As a visitor to a timeline on a webpage (https://175.uiowa.edu/milestones-university-iowa-history), I can see some sort of indication that I should scroll down so that I know there's content there.  Perhaps draw the first element by default?  something else?

cory-skeers:
Resolves #4565 

# How to test

`git checkout timeline-indication && git pull`
`yarn workspace uids_base build`
`blt ds --site=175.uiowa.edu`
Visit https://175.uiowa.ddev.site/milestones-university-iowa-history with various browser screen sizes, and check if it is apparent the user should scroll down for additional content.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/components/timeline.scss
+++ b/docroot/themes/custom/uids_base/scss/components/timeline.scss
@@ -207,7 +207,8 @@
   transition: opacity 500ms;
 }
 
-.timeline--card.js-scroll.scrolled {
+.timeline--card.js-scroll.scrolled,
+li:first-child .timeline--card.js-scroll {
   opacity: 1;
 }
 


cory-skeers:
In looking at this, I started with making the first element visible. I feel like it gives enough of an indication to scroll for content. There's a small window of viewport size that still allows for no timeline content to be seen on initial load, but that feels like a more common occurrence in web design, and less related to the intention of the original issue.

Another thought was to remove the line prior to the first item, but after seeing the content visible, I don't know that further treatment is needed. Thoughts @briand44?

Before treatment:
image

After treatment:
image

After treatment, and removing additional margin above first element:
image

The portion of screen seen where there is no content visible, a relatively short and wide viewport (visible first element, but no margin change:
image

joewhitsitt (approved):
Reviewed the change and saw the standup demo of this. The team thought this was a good step forward.

Issue #4502, opened 2021-12-01
Spacing with accordions


bewoldt:
# Problem

Accordions have a margin top to space them out but no margin bottom so when you add a block directly after an accordion there isn't any space between them.

<img width="1033" alt="Screen Shot 2021-12-01 at 11 25 23 AM" src="https://user-images.githubusercontent.com/731952/144285144-2f9c9d14-8cc2-4664-bad6-d51656e5906b.png">

Example page
https://recserv.prod.drupal.uiowa.edu/climbing-wall

Temporary solution is to add a single line break <br> to manually space out elements but it would be nice if that wasn't necessary.

bspeare:
I've noticed this too.  Could we just add the "Margin" styles to this block to create extra spacing as needed?

<img width="666" alt="Screen Shot 2021-12-01 at 12 02 00 PM" src="https://user-images.githubusercontent.com/1036433/144308927-b14d1595-05b5-4ebb-8ad7-ba72648f4759.png">

quamsta:
I ran into this too during a site migration that needed some headings and accordions:
<img width="1483" alt="Screen Shot 2022-01-03 at 3 54 56 PM" src="https://user-images.githubusercontent.com/472923/147984596-c60cf2aa-2b06-478f-996b-b9bb2314c687.png">

PR: https://github.com/uiowa/uiowa/pull/4612

bspeare:
Closed via https://github.com/uiowa/uiowa/pull/4612.

PR #4612, opened 2022-01-03
Added margin top and margin bottom layout builder styles to Collection block

quamsta:
See: https://github.com/uiowa/uiowa/issues/4502

# How to test

Pull down this branch and import the config changes, create a new Collection block and add the margin top and margin bottom styles and confirm they're applied.


bspeare:
Pulling down to test

bspeare (approved):
This adds the margin top and bottom style options to the collections block and they are applied after saving the block.

PR #4635, opened 2022-01-07
fix card aria

### LINKED ISSUE #4611: Check the aria markup for card variations
# Problem
aria-describedby is present for the card arrow button. woot! good job. It needs to be switched around though. The title has the text that is describing the link but are telling screen readers to look to the blank button for the link description.

Screen Shot 2022-01-03 at 1 13 03 PM

# Steps to reproduce
Card with title, arrow button link

# Proposed solution
- [ ] Revisit the logic around how the elements are sprinkled with aria.

GaryRidgway:
fixes #4611

## To testerino

1. sync down sandbox
2. go to `https://sandbox.uiowa.ddev.site/cards`
3. test that the card indicator cards do not have `aria-describedby` id's anymore, and that the headline isn't pointing to them.
4. ?????
5. This joke is old.

### DIFF
--- a/docroot/themes/custom/uids_base/templates/uids/card.html.twig
+++ b/docroot/themes/custom/uids_base/templates/uids/card.html.twig
@@ -34,7 +34,7 @@
 {% endif %}
 
 {# Do not show the aria-describedby attribute if title AND link text are not both present. #}
-{% if link_element != 'title' %}
+{% if link_element != 'title' or card_link_title == ''%}
   {% set card_aria_describedby = '' %}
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Reviewed and tested, and all looks good!

(Writing this out 'cause I needed it to avoid confusing myself) If there is button text provided, area-describedby is present on the h2.a link and points to the button text. If no button text is provided, it does not receive an id, and the h2.a link does not receive an aria-describedby attribute, and then only uses the h2 text.

PR #4639, opened 2022-01-10
fix featured item target null error

GaryRidgway:
This is supposed to fix an error with targeting deleted referenced items. this should check for those null items before trying to render them.

# How to test

don't test this, it works, I promise.


richardbporter (approved):
Yes.

GaryRidgway:
Agreed.

cory-skeers:
> Agreed.

Ditto.

PR #4648, opened 2022-01-12
Button position alignment style option

### LINKED ISSUE #2406: Add ability to position card button at the bottom of the card
# Problem

Card buttons are positioned immediately below the card title and body. 

# Proposed Solution

Add a layout builder style option to the card to position a button at the bottom of a card. 

Screen Shot 2020-10-21 at 12 13 50 PM

# Next steps

Screen Shot 2021-12-21 at 1 13 51 PM

- Add an option in the "Styles" menu to position button at the bottom of the card
  - The layout builder style would add a class to the wrapper that sets these values:
    - `.card__body { display: flex; flex-direction: column;}`
    -  card `.bttn { margin-top: auto; }`
- Test with all card variants
  - Test on all browsers, including IE
  - Test on a v2 site

GaryRidgway:
this should fix https://github.com/uiowa/uiowa/issues/2406

## How to test

sync down sandbox
add the `Card: Button position bottom` style to cards
Test that the button gets positioned to the bottom of the card, unless the section has  the `vertically align to top` style.

Test with all card variants
Test on all browsers, including IE
Test on a v2 site


joewhitsitt:
First question: why is the style called `Card: Button position`? as a user i want to force a button to sit at the bottom of my card, this style doesn’t tell me what it is going to do with my button.

Alan said that @bspeare might have the background on this.

joewhitsitt:
pulling down to test

bspeare:
> First question: why is the style called `Card: Button position`? as a user i want to force a button to sit at the bottom of my card, this style doesn’t tell me what it is going to do with my button.
> 
> Alan said that @bspeare might have the background on this.

I think we can change it to `Card: Button position bottom` or any other suggestion!

GaryRidgway:
Okidokey! I will change it in a moment

GaryRidgway:
@joewhitsitt go ahead and grab it again, it should be changed now

joewhitsitt:
looks like main was merged but there was no change to the file diff

GaryRidgway:
sorry, there you go. my apologies, i had to merge locally

joewhitsitt (approved):
Tried multiple combinations of v3 cards and it works as described. I can see folks getting confused by the section alignment, but one that we can address with training/support.

v2 card buttons are no longer, instead the link wraps the card only. buttons in v2 are wysiwyg buttons and I see no change there.

ie11 doesn't break

joewhitsitt:
Not sure if other button positions are in the future, but the class name could become a problem if just the label is changed here.

Issue #4242, opened 2021-10-01
When all content is private, store all files in private files system


briand44:
# Story

As an owner of a site where all content is private, I would like all my sites files to be private as well.

# Additional Info
Parent issue #4241 

# Proposed solution
- Most of the work for this has already been done.
- [ ] Write test to check for private scheme patch version of config for every default public scheme: https://github.com/uiowa/uiowa/issues/4242#issuecomment-1013324907

richardbporter:
We discussed a potential solution of setting the `$config['system.file']['default_scheme'] = 'private';` in a site's `settings.php` file right after it is created. This would prevent anyone from switching back and forth, at least without modifying code and understanding the implications.

However, we also discussed that SiteNow was created for public-facing websites from day 1 and that retrofitting an "intranet" solution might be overlooking lots of baked-in functionality.

pyrello:
A few notes from our grooming conversation about this:
* If a site were somehow able to switch back and forth between public and private, we would need some sort of (drush?) script to move all the files from one filesystem to another.
* Because of the above, it would be a lot easier to make it a policy that a site that is going to be private needs to be created that way from the start and will then always be private.
* Is it a possibility to set the filesystem type per media bundle? If so, we could potentially create a "private file" option for things that need to be protected in a mixed system.

richardbporter:
Hm. Changing the `system.file.default_scheme` from public to private locally doesn't actually do anything - thats just the default scheme used when creating new file fields. All of our existing media type file fields would still use the public file system. That config would have to be changed, as well as anything else configured to use the public file system (custom code, etc.).

pyrello:
> Hm. Changing the `system.file.default_scheme` from public to private locally doesn't actually do anything - thats just the default scheme used when creating new file fields. All of our existing media type file fields would still use the public file system. That config would have to be changed, as well as anything else configured to use the public file system (custom code, etc.).

That's good to know. So, we would probably need to set this up as a split and turn it on for sites prior to granting them access.

pyrello:
This is not comprehensive, but provides a few points to think about: https://industrialagency.ca/blog/private-files-drupal-8

richardbporter:
References to `uri_scheme`: https://github.com/uiowa/uiowa/search?q=uri_scheme

References to `public://`:
```
config/default/media.settings.yml:icon_base_uri: 'public://media-icons/generic'
[...]

PR #4666, opened 2022-01-14
Update all multisite default.local.settings.php files.

richardbporter:
While looking into https://github.com/uiowa/uiowa/issues/4242#issuecomment-1011432303, I noticed that all of our multisite local settings are out of date compared to what is in BLT upstream. This PR changes the `blt bis` command, again, to remove all `default.local.settings.php` files so the BLT default.local.settings.php file is copied in. We can use this command to manage local settings going forward, for example when updating blt.

Here are the recent updates to local.settings.php we are getting: https://github.com/acquia/blt/commits/main/settings/default.local.settings.php

Additional changes:
- require removal of multisites override in local.blt.yml so all files are regenerated
- Prompt for confirmation since developers may have config overrides or other settings set in local.settings.php files.

Devs probably don't need to run `blt bis` that frequently. Only when updating all settings files as part of a BLT update or when needing to work with a new site locally.

# How to test

`blt sync`
Poke around the site.
`blt sync --site mysite.com`
Poke around the site.

### DIFF
--- a/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
+++ b/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
@@ -181,6 +181,10 @@ class ReplaceCommands extends BltTasks {
    * @hook pre-command source:build:settings
    */
   public function preSourceBuildSettings() {
+    if (!$this->confirm('This will delete all local.settings.php files for defined multisites. Are you sure?')) {
+      throw new BltException('Aborted.');
+    }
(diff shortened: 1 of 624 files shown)


richardbporter:
Related: https://github.com/acquia/blt/pull/4460

joewhitsitt:
pulling down to test

joewhitsitt (approved):
Works for me.

Issue #3200, opened 2021-03-08
Epic: Menu block improvements


bspeare:
# Motivation

We have gotten several requests for different types of improvements to the menu system. This is a spot to collect and track them all.

# Tasks
* [x] #4703 
* [x] #4701 
* [x] #4823
* [ ] #4711
* [x] #4544 
* [ ] #4712
* [ ] https://github.com/uiowa/uiowa/issues/4764
* [ ] https://github.com/uiowa/uiowa/issues/7092
* [ ] https://github.com/uiowa/uiowa/issues/7764
* [ ] https://github.com/uiowa/uiowa/issues/9539

# Questions
- Do we want to limit this to only administrators?  If we are switching layouts based on the menu style in the future slice, does access to the per page overrides need to be restricted?
- Allow dropdown menus? https://github.com/uiowa/uiowa/issues/3200#issuecomment-1007513375

bspeare:
Here are a few creative examples of sites trying to create their own "menus" that could benefit from the ability to set the initial visibility level to "Active menu item" to display a sidebar navigation or a horizontal menu:

<img width="1474" alt="Screen Shot 2021-03-08 at 1 47 14 PM" src="https://user-images.githubusercontent.com/1036433/110373364-07e44100-8015-11eb-85fe-42a7352eb81c.png">

<img width="1762" alt="Screen Shot 2021-03-08 at 1 46 43 PM" src="https://user-images.githubusercontent.com/1036433/110373371-0a469b00-8015-11eb-871a-96dc7469916c.png">

<img width="1474" alt="Screen Shot 2021-03-08 at 1 45 39 PM" src="https://user-images.githubusercontent.com/1036433/110373384-0ca8f500-8015-11eb-9e16-aec66140fd13.png">

bspeare:
After discussion:

- Allow changing active or children with explanations that make more sense?
- Button block allow multiple, display as menu?

makurj:
I'd like to pair on this.

bspeare:
I've updated the proposed solution to reflect our grooming discussion today.

bspeare:
Added a pharmacy tag because they were also interested in this functionality.

bspeare:
I've updated the proposed solution to reflect our conversation from this morning.  I will work on adding some screenshots/gifs of how the dropdowns might look on our current pages.  

[...]

PR #4667, opened 2022-01-14
Add access to menu settings for content editors

### LINKED ISSUE #4544: Secondary Navigation: Visual Adjustments
# Story

As a designer, I want the ability to make adjustments to the secondary navigation style to be more visually appealing.

### LINKED ISSUE #4563: Update LB styles grouping solution
# Motivation

Since we implemented the [patch to allow for LB style groups](https://www.drupal.org/project/layout_builder_styles/issues/3075502#comment-14032178), there has been quite a bit more work on that issue and the solution has evolved to the point where, I believe, it is incompatible with the solution we adopted. Because the original solution was never adopted by the module, there will be no automatic upgrade path via update hooks and we will need to write this ourself to be able to adopt the new changes and eventually future versions of the module.

# Proposed solution

- [x] Analyze the changes to the way that group information is stored and document.
- [x] Update to version of module with patch committed OR a newer version of the patch that has been accepted for being committed.
- [x] ~Write an update hook to transition between the interim solution for storing group information and whatever the final solution is.~ The update hook in https://github.com/uiowa/uiowa/issues/4563#issuecomment-1026913079 was used to generate the config updates, but wasn't actually necessary to include in this PR.

### LINKED ISSUE #4703: Allow more access to menu block level and depth settings
Creating a new first slice for #3200. These features should satisfy the request made by Student Life

# Stories

* As an editor I want to choose horizontal or vertical options for my menu block so that I can take advantage of those different design options.
* As a platform manager I want the above choice to appropriately restrict the level and depth options so that content editors cannot select options that we have not accounted for in our design/theme.

# Proposed solution

- [x] Add vertical menu option, group it with the horizontal menu option, require a selection and only one. Set vertical menu style as the default.
- [x] Update hook to update existing menu blocks to be vertical if horizontal is not selected.on the Main navigation block to allow the menus to display sibling pages.   
- [x] Either a) move the existing menu orientation LB style field to before the menu settings details pane; or b) add a faux field before the menu settings details pane that pulls in those options. If b, then we'll need to handle passing through these field values to the LB style setting. Also the LB style setting will need to be hidden.
- [x] Either via `#states` or custom JS, dynamically change the selectable options in the level and depth fields based on the the selected menu orientation. Allowed values will be documented below.
- [x] Remove the access check that disables the menu settings fields for non-admins.
- [x] As a fallback if the menu is placed in a small column within layout builder and set to `Horizontal`, implement Container Queries polyfill (https://css-tricks.com/a-new-container-query-polyfill-that-just-works/) to style menu in horizontal/vertical style based on the layout builder region column.

### LINKED ISSUE #4794: Re-organize menu_block settings to make them easier to understand
# Proposed solution

Continue the work in https://github.com/uiowa/uiowa/pull/4667, and make these changes:

-  [x] Find a way to unset the "Fixed parent item" if "Make the initial visibility level follow the active menu item" is checked when the block is saved. 
-  [x] Find a way to reset the "Initial Visibility" level to `2` if "Make the initial visibility level follow the active menu item" is checked when the block is saved. 
- [x] Update and reorganize form display so the show/hide makes sense
- [x] Update the "main navigation" block under the "more.." option in layout builder to have the same settings that are provided through the block placed on the page. 
- [x] ~Test that everything is still working with our custom logic to hide "depth" setting and automatically set that value to `1` when "Horizontal" is selected.~ We decided to allow the depth setting to be used for both menu types in anticipation of #4823. 
- [x] Write test to assert that fields we have hidden in custom code are not showing up on the menu block form.

**Older partially completed and outdated solution:**
<details>
- [x]  Check the [`menu_block` issue queue](https://www.drupal.org/project/issues/menu_block?version=any_8.x-) to see if there is any upstream work going on to re-organize the form in a way that makes sense. If there is, these suggestions may need to be adjusted to reflect that work.
- [x] Validate that "Make the initial visibility level follow the active menu item." option essentially overrides anything that is set for "Initial visibility level" and "Fixed parent item" (See comments below).
- [ ] Move this section of the form into or before the Menu levels section:
  Screen Shot 2022-02-09 at 11 01 36 AM
- [x] "Make the initial visibility level follow the active menu item." should be checked by default.
- [x] Decide which of the radio buttons should be the default and make it so: "Active menu item" or "Children of active menu item"
- [x] If "Make the initial visibility level..." is checked, hide "Initial visibility level" and "Fixed parent item" options.
- [X] ~Probably just get rid of the "Advanced options" details section because it doesn't really help the form to break the options up that way.~ Renamed this to "Visibility options" instead.
</details>

GaryRidgway:
Resolves #4703.
Resolves #4563. 
Resolves #4544. 
Resolves #4794. 

# To test
```
ddev restart # Only necessary if you have not done so since #4854 was merged.
composer install
blt frontend
blt ds --site sandbox.uiowa.edu
```
- Spot check LB style settings to make sure that they are rendering correctly in block forms and that selected values are reflected in block display.
- Add or edit a menu block on a new or existing page.
- When adding a menu block, "Vertical" should be the default option for menu orientation.
- Otherwise, the "Number of levels to display should only allow 1-3 as options.
- You should be required to select either "Vertical" or "Horizontal."
- ~After you selected the "Horizontal" option and saved the block, if you configure the block again and choose "Vertical" this time you see 1 as the selected value under "Number of levels to display."~
- If you set the style to "Horizontal" in any layout outside of the one column layout, the design should appear as a vertical menu.
- In Layout Builder the preview for the menu should always show the placeholder text.
- If the menu block is placed in a one column layout, the placeholder style should not change.

## To test the test
- `ddev phpunit docroot/modules/custom/layout_builder_custom --filter testMenuBlockOverrides`
- Confirm the test runs without errors or failures.
- Click the last link to see the HTML generated during the test.
- Visually confirm that the form matches the expected fields and order that you observed in previous testing.


pyrello:
To close out this work, everything except for the updates to change the display of the layout builder style settings for menu orientation should be removed for now. We can refer back to this when working on future enhancements for #3200.

pyrello:
The remaining item to be done is to create the fallback CSS so that if a horizontal menu is placed in a column that is too small in width it reverts back to looking like a vertical menu.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -248,7 +212,6 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
               $defaults['layout_builder_style_card_image_size'] = 'card_image_medium';
               break;
 
-            case 'views_block:events_list_block-card_list':
```
joewhitsitt, line 251: Was removing `case 'views_block:events_list_block-card_list':` intentional?

pyrello: It was a duplicate line. I left in the one that was grouped with the block listing events from content hub.

joewhitsitt:
pulling down to test

joewhitsitt:
Though I am looking at this, I don't want to keep others from looking at it. I think it would be good if we could get more eyes on it.

joewhitsitt:
Capturing from a DM with Ben:

i am seeing the none option as a webmaster
the selection of vertical/horizontal is not saved (removed my bit about publishing because someone demoted Michael on the sandbox site)

Additionally want to report my other testing:

- admissions, v2 don't have any sync issues
- card options seem to be moved around but still functioning when selecting image stacked (sizes disappears)

pyrello:
>i am seeing the none option as a webmaster

Trying to get the `- None -` value option solved upstream: https://git.drupalcode.org/project/layout_builder_styles/-/merge_requests/5/diffs#note_69041

> i can’t seem to publish

Can you expand on this? Do you mean save the layout? Or move a draft to published?

> card options seem to be moved around

I will look into this.

joewhitsitt:
correction on the published - someone demoted Michael to editor on the sandbox site

joewhitsitt:
Sorry the duplicates, just wanted to clarify my testing:

> Spot check LB style settings to make sure that they are rendering correctly in block forms and that selected values are reflected in block display.
>Add or edit a menu block on a new or existing page.

Spot checked a few sites, overall the normal blocks seem to be working as expected.

>When adding a menu block, "Vertical" should be the default option for menu orientation.

Yes.

> When "Horizontal" is selected for menu orientation, the "Number of levels to display" field is hidden.
> Otherwise, the "Number of levels to display should only allow 1-4 as options.

Yes.

> You should have no option to not select either "Vertical" or "Horizontal."

None is an option.

>After you selected the "Horizontal" option and saved the block, if you configure the block again and choose "Vertical" this time you see 1 as the selected value under "Number of levels to display."

The value regardless of selection is not saving or changing how it is rendered.

>As a non-admin, I should be able to see all of the following settings

Confirmed.

pyrello:
I got the issue with the `- None -` option sorted out. I also figured out that removing the original LB style element from the form was causing the value to get skipped because of the way that module prepares styles for saving. Changed our way of hiding it to reflect this.

It seems that maybe the order of the styles or groups has changed and I can look into that more to see if there is way to keep it as it is.

pyrello:
I added weights to the LBS groups, so they should at least render in a predictable order now. Not sure why they are ordering differently outside of that, but I'm not sure it is a big enough deal to worry about.

bspeare:
It looks like we lost the "Styles" dropdown from each layout section:

Screen Shot 2022-02-02 at 4 30 04 PM

bspeare:
It looks like anywhere we were using the "Styles" option on a block or section it is no longer displaying.   This is how the card should look:

Screen Shot 2022-02-03 at 8 34 19 AM

pyrello:
Looking into the issue related to "Styles" drop-down not showing.

pyrello:
We were missing a default style group. Added that and the styles field started showing up. Only checked it on cards so far.

joewhitsitt:
pulling down to review again

bspeare:
As @joewhitsitt pointed out, we are having some weird layout builder previews of the menu and a different result when editing the page.   This seems to happen when the starting level is set to `1`. 

https://user-images.githubusercontent.com/1036433/152381850-c88b4d47-cd43-40aa-bd00-c56e88fb567e.mov

joewhitsitt:
my selections don't seem to match render. Alan originally had some code to "fix" the keys as they weren't 1:1

update on previous testing, none option was removed and selections are saving

joewhitsitt:
I think some description help text on the orientation field would be good to say that horizontal will only work in certain sections.

pyrello:
Switching back to draft until we can get more clarity on the source of the bug that has been identified above.

bspeare:
Drupal core with `menu_block` and `layout_builder`: The preview is still inconsistent if the level is set to `1`. 

https://user-images.githubusercontent.com/1036433/152398071-1a3f3b02-3589-4846-a244-851eb5e8e1ff.mov

SiteNow: Unchecking "Make the initial visibility level follow the active menu item" fixes the display issue but you can see after I save the page and edit again the form values have reverted to the `2` visibility level and "Make the initial visibility level follow the active menu item" is checked again. 



https://user-images.githubusercontent.com/1036433/152398231-a0ebd5a8-70d3-4314-8409-122f631ebfb3.mov

pyrello:
The SiteNow behavior on prod makes sense because we were setting this default value without regard to previous selections:
https://github.com/uiowa/uiowa/blob/38f35182f399aa3e61acce8c91ae5c13a4dfcd51/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L332-L333

Do we want to uncheck and hide the "Make the initial visibility level follow the active menu item" setting? Or do we need that setting to be available?

bspeare:
I think the "Make the initial visibility level follow the active menu item" is handy for adjusting how the menu acts on the page.  

Here's an example of how the menu changes if it is checked or unchecked. 


If it is checked, it will show the child pages under "How to Apply":

Screen Shot 2022-02-03 at 1 38 25 PM

If it is unchecked, it will show the sibling pages in the "Apply" section:

Screen Shot 2022-02-03 at 1 37 46 PM

What I have noticed in the Drupal install without SiteNow is that the menu preview in layout builder is inconsistent.   Is there a way to hide the preview for that menu block?

bspeare:
Still to do:
- [x] Hide `1` visibility depth
- [x] Figure out why "Make the initial visibility level follow the active menu item" is always checked
- [x] Add description help text on the orientation field to say that horizontal will only work in one column sections

richardbporter:
I get an error running `yarn install` for an optional dependency gifsicle. Running `blt frontend:setup` will run `yarn install` with the same options specified in our BLT config which will skip optional dependencies, among other things.

https://github.com/uiowa/uiowa/blob/28c4d9deda9959cf280dc33e2d3cdb3b0c2a1ce1/blt/blt.yml#L51

https://iowaweb.slack.com/archives/C0164FNN1PV/p1644250571621389?thread_ts=1644248133.624399&cid=C0164FNN1PV

pyrello:
I'm not sure we even want to run `yarn install` locally. Instead, I think we're supposed to run `yarn ci`? But I haven't looked into it in a while.

dnkemper:
> I'm not sure we even want to run `yarn install` locally. Instead, I think we're supposed to run `yarn ci`? But I haven't looked into it in a while.

WUT?? I would run `rm -rf ./node_modules; yarn cache clean; yarn install; yarn workspace uids_base gulp --development` when working in local environment w/ things that. need to be compiled. Also, never used `yarn ci`. Intrigued

richardbporter:
I see the label "requires cache rebuild" but no update hook to do so. Is that needed still?

pyrello:
Whoops. Not `yarn ci`. That's for npm: `npm ci`

We are wanting to install without updating dependencies: `yarn install --frozen-lockfile`. `yarn install` on its own can update dependencies, I believe.

pyrello:
> I see the label "requires cache rebuild" but no update hook to do so. Is that needed still?

@richardbporter Thanks for the reminder! I added an update hook. Since we are changing patches and versions of Layout Builder Styles, I think we need that to prevent WSOD for pages that call outdated code.

richardbporter:
>Otherwise, the "Number of levels to display should only allow 1-4 as options.

>After you selected the "Horizontal" option and saved the block, if you configure the block again and choose "Vertical" this time you see 1 as the selected value under "Number of levels to display."

I see 2-4 and 2 respectively.  The test text is just outdated, right?

richardbporter:
From the thread, it sounds like the layout builder preview is currently an issue and we're not aiming to solve it right now. However, it does make it really difficult to understand what all these setting actually do when changing the block config and just seeing placeholder text. 

One thing I don't understand is the "Use as title" config. That title doesn't seem to be visible. Is that going to cause confusion for folks?

bspeare:
Maybe we should hide the "Use as title" option until we can get to https://github.com/uiowa/uiowa/issues/4764?

pyrello:
> I see 2-4 and 2 respectively. The test text is just outdated, right?

Yes, although I think you just pointed out a problem with the code logic that we need to figure out. Currently, if "Horizontal" is selected for the menu orientation, then we set the depth to `1` in our form submit handler. However, we are removing `1` from the options for depth in the form. I don't think this technically would cause any issues because to change it to "Vertical," you'd have to save the form again, which means that the value of depth would be set to whatever was selected. But it probably would be a good idea to not try to set the default value of the depth field to a value that is unset. @bspeare Do you think there would be any issue with changing the automatically set value of depth to `2` when the user selects "Horizontal"?

richardbporter:
> > I see the label "requires cache rebuild" but no update hook to do so. Is that needed still?
> 
> @richardbporter Thanks for the reminder! I added an update hook. Since we are changing patches and versions of Layout Builder Styles, I think we need that to prevent WSOD for pages that call outdated code.

I wasn't able to reproduce a WSOD locally but I did get an AJAX error when trying to edit a block, making it impossible to do so. The update hook fixed it. 👍🏻 

Error details: `The website encountered an unexpected error. Please try again later.Drupal\Component\Plugin\Exception\PluginNotFoundException: The &quot;layout_builder_style_group&quot; entity type does not exist. in Drupal\Core\Entity\EntityTypeManager-&gt;getDefinition() (line 143 of core/lib/Drupal/Core/Entity/EntityTypeManager.php). Drupal\Core\Entity\EntityTypeManager-&gt;getHandler(&#039;layout_builder_style_group&#039;, &#039;storage&#039;) (Line: 196)`

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -282,7 +282,7 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
           // Limit the depth options to 2-4. We're removing the 1 option because
           // it causes issues with a discrepancy between the preview and what is
           // shown on the page after saving. Also, we consider it bad practice
-          // to shown the top-level menu again on a page.
+          // to show the top-level menu again on a page.
```
richardbporter: Blah. This commit is supposed to say 'change shown to show'...

richardbporter:
> Maybe we should hide the "Use as title" option until we can get to #4764?

I did this, for all users. Not sure if admins need access but if they do, we'll have to implement an access check. [98e246a](https://github.com/uiowa/uiowa/pull/4667/commits/98e246a43c02ac17168bd4296e9ca475e534b6fb)

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
richardbporter: And then I changed it again in [3babe1b](https://github.com/uiowa/uiowa/pull/4667/commits/3babe1b33f33c3ecb325ad3c68341c1747eeb148). TIL about gerund. /shrug

pyrello:
> I did this, for all users. Not sure if admins need access but if they do, we'll have to implement an access check. [98e246a](https://github.com/uiowa/uiowa/pull/4667/commits/98e246a43c02ac17168bd4296e9ca475e534b6fb)

For now, I decided to go without the access check and I think that makes sense for now.

richardbporter (commented):
Seems like all the test steps check out. The LB preview is really frustrating and these new settings might cause more people to ask about that.

I can approve but it seems like we'll want to schedule an off-hours deployment because people won't be able to edit blocks until the cache is cleared.

pyrello:
> I get an error running yarn install for an optional dependency gifsicle. Running blt frontend:setup will run yarn install with the same options specified in our BLT config which will skip optional dependencies, among other things.

We still need to get `blt frontend` sorted out in #4771, but `imagemin-gifsicle` isn't really doing much for us and I think we can safely remove it since it seems to cause a lot of hassles: uiowa/uids#630

bspeare:
> Do you think there would be any issue with changing the automatically set value of depth to 2 when the user selects "Horizontal"?

Screen Shot 2022-02-08 at 9 08 55 AM

Should we set this to `2` and get rid of `1` so we don't have the incorrect live preview issue?

pyrello:
> Should we set this to 2 and get rid of 1 so we don't have the incorrect live preview issue?

Done

richardbporter (dismissed):

pyrello:
Here are my suggestions to make this block make more sense:
* Check the [`menu_block` issue queue](https://www.drupal.org/project/issues/menu_block?version=any_8.x-) to see if there is any upstream work going on to re-organize the form in a way that makes sense. If there is, these suggestions may need to be adjusted to reflect that work.
* Validate that "Make the initial visibility level follow the active menu item." option essentially overrides anything that is set for "Initial visibility level" and "Fixed parent item".
* Move this section of the form into or before the Menu levels section:
  Screen Shot 2022-02-09 at 11 01 36 AM
* "Make the initial visibility level follow the active menu item." should be checked by default.
* Decide which of the radio buttons should be the default and make it so: "Active menu item" or "Children of active menu item"
* If "Make the initial visibility level..." is checked, hide "Initial visibility level" and "Fixed parent item" options.
* Probably just get rid of the "Advanced options" details section because it doesn't really help the form to break the options up that way.
* Test that everything is still working with our custom logic to hide "depth" setting and automatically set that value to `1` when "Horizontal" is selected.

bspeare:
I'll work on fixing the merge conflicts.

richardbporter:
> I'll work on fixing the merge conflicts.

This could be tricky. What I did for the uipress conflict was `git checkout --theirs -- config/sites/uipress.uiowa.edu`, manually sql-sync the uipress site, cr, cim, cim, updb and then export the site split.

For this, you could `git checkout --ours -- config/default/layout_builder_styles.* config/default/core.entity_view_display.node.page.default.yml`, manually sql:sync the default site, cr, cim, cim, updb and then cex.

richardbporter:
> > I'll work on fixing the merge conflicts.
> 
> This could be tricky. What I did for the uipress conflict was `git checkout --theirs -- config/sites/uipress.uiowa.edu`, manually sql-sync the uipress site, cr, cim, cim, updb and then export the site split.
> 
> For this, you could `git checkout --ours -- config/default/layout_builder_styles.* config/default/core.entity_view_display.node.page.default.yml`, manually sql:sync the default site, cr, cim, cim, updb and then cex.

Forgot the two cims - updated.

bspeare:
> 

Thanks @richardbporter, I'm not exactly sure what those commands are doing.   Would you mind running this on your end or huddling to explain this further?

richardbporter:
Looks like the update hook is not working with the changes to `unserialize` calls. :/ 

I tried adding more allowed_classes but there is a lot of potential objects stored in there. I hate to do this but we might just have to phpcs:ignore those lines. IDK if using the Drupal API to update these blocks instead of querying the database directly is possible, but that might be something worth looking into in the future.

I will work on reverting the `unserialize` changes.

richardbporter:
I set previously-run update hook usage of unserialize to disallow any classes and `_update_all_blocks_by_plugin_id` usage to allow all. Added comments as well. Update hook seemed to run fine after that.

richardbporter:
> > I'll work on fixing the merge conflicts.
> 
> This could be tricky. What I did for the uipress conflict was `git checkout --theirs -- config/sites/uipress.uiowa.edu`, manually sql-sync the uipress site, cr, cim, cim, updb and then export the site split.
> 
> For this, you could `git checkout --ours -- config/default/layout_builder_styles.* config/default/core.entity_view_display.node.page.default.yml`, manually sql:sync the default site, cr, cim, cim, updb and then cex.

I did this on the config in question. Might be worth double-checking things now.

richardbporter:
I think the stripped-down form looks much better for webmasters. 👍🏻

bspeare:
Noticing that the Background styles are now out of order.  Maybe this happened with the 9.3 update?

**Before:**

Screen Shot 2022-02-22 at 11 51 11 AM


**After:**
Screen Shot 2022-02-22 at 11 45 37 AM

richardbporter:
> Noticing that the Background styles are now out of order. Maybe this happened with the 9.3 update?

Could be, that re-sorted a lot of config.

richardbporter:
> > Noticing that the Background styles are now out of order. Maybe this happened with the 9.3 update?
> 
> Could be, that re-sorted a lot of config.

They seem to be in correct order in main/prod. Maybe I goofed up the merge conflict. I will look again.

richardbporter:
Nothing weight/sort related is sticking out to me. Must be related to the layout_builder_styles update? I found this: https://www.drupal.org/project/layout_builder_styles/issues/3159865

richardbporter:
I added a one-liner in our layout_builder_configure_section form alter for the time being to alpha-sort these. It seems like that issue MR is not the desired direction of the maintainers.

richardbporter:
> I added a one-liner in our layout_builder_configure_section form alter for the time being to alpha-sort these. It seems like that issue MR is not the desired direction of the maintainers.

Nope, that breaks saving the selected option I think. Reverted.

pyrello:
> Noticing that the Background styles are now out of order. Maybe this happened with the 9.3 update?

This could potentially be related to the update to Layout Builder Styles. Do we need to re-save the order of the styles? Or maybe something in that update caused the ordering to stop working? Or maybe some of our custom code is getting in the way?

**Edit:**

Whoops. Just noticed that @richardbporter already mentioned this.

bspeare:
I think we just need to re-save the order.  I can do that and update this shortly.

bspeare:
Should rename this block to be "Secondary navigation" or something else since this is not really the "main" navigation?

Screen Shot 2022-02-24 at 1 42 58 PM

bspeare:
Another thing to consider is that the defaults available when you choose the "main navigation" block through the "more..." option in layout builder are not the same defaults available when you create a new page and edit the "main navigation" block provided within the layout.

pyrello:
> Another thing to consider is that the defaults available when you choose the "main navigation" block through the "more..." option in layout builder are not the same defaults available when you create a new page and edit the "main navigation" block provided within the layout.

We will need to update our LB template for the page to match the default settings we are changing to in this PR.

pyrello:
@richardbporter I dismissed your old approval since the PR has changed since then.

bspeare:
Just tried to sync sandbox and I got this error: ` [error]  Update aborted by: sitenow_update_9035 `

pyrello:
Setting back to draft to check this out.

cory-skeers:
Testing the test worked well--going to take a look at the rest of it.

cory-skeers:
Could we adjust some spacing on the menu block in the layout preview? I've had a couple moments where the configure contextual button overlaps another component, and it's not always clear what I will be clicking on.

image

(I'm also running into what I think are local issues to me, so it's possible this is a me-only issue.)

pyrello:
@cory-skeers That should be cleared up by running `blt frontend`.

cory-skeers:
> @cory-skeers That should be cleared up by running `blt frontend`.

Alright, think there's something else going on here on my environment. `blt frontend` isn't helping at all.

pyrello:
> Alright, think there's something else going on here on my environment. `blt frontend` isn't helping at all.

@cory-skeers I am not having any luck replicating what you should this morning after standup. The `layout-builder-block--placeholder` class is being added to the menu block and the appropriate CSS is being applied. Either this is a problem with just your local environment for some reason, or I guess I need more information to figure out how to resolve it. Maybe we call it good if everything else looks okay?

bspeare:
I'm not seeing the class either.  I'm syncing a site again to double check.

pyrello:
I think the issue is fixed.

richardbporter:
I see one more call to unserialize with some allowed_classes specified. Should that also be changed to `'allowed_classes' => TRUE`?

https://github.com/uiowa/uiowa/blob/e2af9275f8118f4eac34b57c518518959aff9608/docroot/profiles/custom/sitenow/sitenow.install#L65-L74

pyrello:
> I see one more call to unserialize with some allowed_classes specified. Should that also be changed to `'allowed_classes' => TRUE`?
> 
> https://github.com/uiowa/uiowa/blob/e2af9275f8118f4eac34b57c518518959aff9608/docroot/profiles/custom/sitenow/sitenow.install#L65-L74

Yes.

pyrello:
There have been updates to the patch in LB styles, so I'm going to update that.

pyrello:
Ready for review again!

cory-skeers:
> * After you selected the "Horizontal" option and saved the block, if you configure the block again and choose "Vertical" this time you see 1 as the selected value under "Number of levels to display."

Checking I understand the desired behavior correctly--say I have a menu set to horizontal with number of levels to display set to 3, and save the block. If I then edit the block and change it to vertical, the number of levels to display should update from 3 to 1? If so, I'm not seeing it updating.

pyrello:
@cory-skeers This instruction is out of date and can be ignored

cory-skeers (approved):
From my testing, this looks good to go!

Verified the testing instructions and tested the test. Spot-checked through layout builder styles, and working as expected.

pyrello:
Thanks @cory-skeers!

Setting back to draft until we can plan a deployment.

joewhitsitt:
The pipeline is clear for this if it is ready

PR #4673, opened 2022-01-18
Create multiple sites and fix site creation issue.

richardbporter:
I broke site creation in https://github.com/uiowa/uiowa/pull/4666. This PR changes `blt bis` from throwing an error if multisite overrides exist to temporarily removing them. It does that  copying the file to the tmp directory and then putting it back. 

Another option would be to go back to the warning message and then only remove existing local settings files from multisites defined in config instead of _all_ multisites.

# How to test

- Check out this branch.
- Make sure you have some multisites overridden in `blt/local.blt.yml`.
- Run `blt uiowa:multisite:create foo.com --no-db --no-commit`
- While the command runs, open `blt/local.blt.yml` and notice any multisites config is gone. Also open the tmp directory in the repo root and see the original local.blt.yml file is there. When the command finishes, the original local.blt.yml file should be copied back to the blt directory.
- Discard changes.


pyrello:
Any chance we can just add the stage file proxy to the local settings files so it just works without having to add it?

richardbporter:
> Any chance we can just add the stage file proxy to the local settings files so it just works without having to add it?

That would be nice. stage_file_proxy is already in our local split. I wonder if there is a token we can use to set that origin URL in that split. Or a way to set it post blt sync. That won't work in every scenario because we don't know whether to use the actual prod URL or an internal prod URL. The actual prod URL would work more of the time?

richardbporter:
Added some post-blt command code in this PR to set the stage_file_proxy origin URL to the site production URL by default or custom BLT config, if set. This would handle most cases, I think but the default behavior is now stage_file_proxy on by default.

Custom config example for sandbox: https://github.com/uiowa/uiowa/pull/4673/commits/3dc8855d44c5ca999411896a427fbe68a0148e22

Another option would be to prompt for input but that requires interaction.

joewhitsitt:
pulling to test

joewhitsitt (approved):
👍

richardbporter:
Thanks. Apologies if this causes more regressions. 🤞🏻 😬

PR #4684, opened 2022-01-19
update from address to a noreply version

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/3498

Instead of sending to our inbox, we want direct replies to bounce so I replaced it.

To test the tests:
`ddev phpunit docroot/modules/custom/uiowa_core`


richardbporter:
I think we can do without the base64_decode and use plaintext - that was just to obscure our service account address. This one isn't real so no spam can get to it.

joewhitsitt:
but i was so proud of my encoding and figured, "if it ain't broke"

richardbporter (approved):

PR #4690, opened 2022-01-20
Move stage_file_proxy origin setting to postBis instead of active config.

richardbporter:
Follow up from #4673 

Using active config leaves sites in a config diff. Config ignoring just the stage_file_proxy.settings:origin did not work. This PR moves the setting to local settings file override after bis is run.

# How to test

- `blt bis`
- See docroot/sites/175.uiowa.edu/settings/local.settings.php has origin set to prod URL.
- See docroot/sites/sandbox.uiowa.edu/settings/local.settings.php has origin set to internal prod URL. That is set from the doroot/sites/sandbox.uiowa.edu/blt.yml file.
- Sync a few sites via `blt dsa`.
- `blt ume cst`
- There should be no config diffs.

### DIFF
--- a/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
+++ b/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
@@ -228,7 +228,6 @@ class ReplaceCommands extends BltTasks {
 \$config['stage_file_proxy.settings']['origin'] = '$origin';
 EOD;
 
-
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
Works as described. Tested on a mix of sites.

Issue #4695, opened 2022-01-24
Hide or adjust feed links in taxonomy term views (but don't remove the feeds entirely)


quamsta:
# Story

Related to: https://github.com/uiowa/uiowa/issues/4435

**Background: This request originated from from Student Disability Services due to user confusion with the "Subscribe to" links on the taxonomy terms on pages like:  https://sds.studentlife.uiowa.edu/tags/ui-scholarships-students-disabilities**

We'd like to hide the "Subscribe to" feed links in taxonomy terms (at least) because of the lack of RSS feed support currently in web browsers. When clicking on a "Subscribe to" link on a taxonomy page like the one below:

<img width="568" alt="Screen Shot 2022-01-11 at 10 22 23 AM" src="https://user-images.githubusercontent.com/472923/148981143-5370fdcc-2eb3-4b16-85b8-188d032130a1.png">

Users are presented with the following dialogs or raw XML code depending on the browser:

Chrome:
<img width="942" alt="Screen Shot 2022-01-11 at 10 28 15 AM" src="https://user-images.githubusercontent.com/472923/148982463-eed98e1f-5818-4845-9af1-d99d4fd259da.png">

Safari does a weird variety of things including:

![](https://user-images.githubusercontent.com/4663676/151036401-d3d0c6e2-e551-4d51-a7b9-fa556551bd50.png)

<img width="1227" alt="Screen Shot 2022-01-25 at 3 44 01 PM" src="https://user-images.githubusercontent.com/472923/151072009-611d6edf-d8af-4974-a7ca-727244e32b3d.png">

<img width="1120" alt="Screen Shot 2022-01-11 at 10 27 50 AM" src="https://user-images.githubusercontent.com/472923/148982299-18310769-d00f-4bfd-832d-cd3228341cf8.png">

Firefox:
<img width="486" alt="Screen Shot 2022-01-11 at 10 27 36 AM" src="https://user-images.githubusercontent.com/472923/148982264-520d6f1b-a9ae-4aea-9b4c-9ce6af772934.png">

# Proposed solution

As @mark-bennett-uiowa stated in https://github.com/uiowa/uiowa/issues/4435, the feeds themselves are often crucial to some sites' functionality. And as @joewhitsitt said in the same issue, RSS feeds may see a comeback in the browser. 

However, the anchor/link in the body of the taxonomy term's page really only results in user confusion unless they have an RSS feed reader extension installed in their browser.

[...]

Issue #4435, opened 2021-11-12
Boring "Subscribe to News" link on SiteNow Articles page.


mark-bennett-uiowa:
# Problem
The Subscribe to News link at the bottom of the SiteNow Articles page is an just a standard link. Compared to the View More Events button link or the View More button created on an Articles block, it could use some styling:
![Screen Shot 2021-11-12 at 12 08 39 PM](https://user-images.githubusercontent.com/3067017/141515227-e73e7e49-87bf-406c-8db4-696229d9cd0c.png)

# Steps to reproduce
Create a SiteNow Articles page and check Show RSS Feed icon in the SiteNow Articles settings page: /admin/config/sitenow/uiowa-articles. I dont even see the feed icon for my sites.

# Proposed solution
Style this link like the View More buttons from other blocks?

joewhitsitt:
This could also extend to all taxonomy term subscribe links

quamsta:
@mark-bennett-uiowa We're actually having the opposite problem here, where people are clicking on the "Subscribe to" link on a taxonomy term for subscribing to scholarships, for example:

https://sds.studentlife.uiowa.edu/tags/ui-scholarships-students-disabilities

<img width="568" alt="Screen Shot 2022-01-11 at 10 22 23 AM" src="https://user-images.githubusercontent.com/472923/148981143-5370fdcc-2eb3-4b16-85b8-188d032130a1.png">


...But they're just hitting an RSS/XML feed that they don't really understand or know how to use. Most browsers don't provide a path to subscribing to feeds anymore by default unfortunately.

I'm wondering if in some cases these feed links could go away entirely and those who want to subscribe to a feed can just use Feedly or whichever client and add the subscription manually by entering the url (like the one above) and the feed reader can find the RSS feed via a meta tag.

quamsta:
Also, we could just remove the link on the taxonomy term but leave it in the articles view, since it might be more useful in articles than in something specific like a taxonomy term.

quamsta:
Here are what all major browsers do when you click a feed link without any browser extensions or third party tools (1/11/22):

Chrome:
<img width="942" alt="Screen Shot 2022-01-11 at 10 28 15 AM" src="https://user-images.githubusercontent.com/472923/148982463-eed98e1f-5818-4845-9af1-d99d4fd259da.png">

Safari:
[...]

PR #4696, opened 2022-01-24
hid taxonomy term feed link

quamsta:
Addresses #4695 

This hides the feed link from the taxonomy term view.


joewhitsitt:
i know you are going for the easy/quick solution on this, but have you tried to hide it higher up while maintaining it in the head info? (e.g. template, view config). even a `@include element-invisible;` IMO is better than a `display:none;`

quamsta:
> i know you are going for the easy/quick solution on this, but have you tried to hide it higher up while maintaining it in the head info? (e.g. template, view config).

@joewhitsitt Everywhere I've looked online says that the feed link in the body of the document is directly tied to making the feed available in the `<head>`. If there's some other way to disable it via config, I'm all in for that. 

> even a `@include element-invisible;` IMO is better than a `display:none;`

The CSS mixin `element-invisble` we're using in the Drupal theme looks to be similar to a Bootstrap CSS class `.sr-only`, which visually hides the link, but still allows it to be spoken aloud by screen readers. 

This is decidedly not the desired behavior, since visitors that are using screen readers like NVDA, JAWS, or iOS or macOS's voiceover in combination with a common web browser will hear the link "Subscribe to [term]" read aloud and will still experience the confusing dialogs or XML presented to them. 

I would rather have the link not be visible or read aloud, hence `display: none;`.

joewhitsitt:
So I am still in the boat that thinks it should still be "available" so would element invisible plus a "RSS Feed link to [term]" be a compromise?

quamsta:
@joewhitsitt I want to say yes to this in the spirit of compromise and camaraderie, but keeping a specific link on a page visible for screen readers and other assistive devices while hiding it visually for everyone else doesn't seem to jive with the spirit of accessibility and equal access imho, even if we change the text.

I feel that if we're going to hide this link, we should hide it for everyone, and keep the feed in the `<head>` for consumption by any feed reading extensions or apps that people are using.

The inverse, aka keeping the link, but rewriting it to "RSS feed link for [term]" and keeping it visible for everyone would be less confusing than "Subscribe to", but then we're making some assumptions about the user knowing what an RSS feed is, and we're still providing a link that can be activated (possibly by accident) and presenting the user with those browser dialogs or XML, which may seem broken to them.

Couple of questions:

- Is this a matter of trying to let the user know that there's a RSS feed available for the page in question?
- Is this something that we can hide for now, but if there's a resurgence in RSS apps or browsers that can handle these links, we can revisit it in the future? 

My biggest concern is just how RSS/XML links are handled these days. Even if we present the link in the clearest terms possible, browsers simply aren't playing nice with them.

joewhitsitt:
Equal access in my opinion would be leaving it as-is but here I am commenting on an issue that wasn't groomed/planned by the team. Keep in mind, I didn't rule out removing it as long as it stayed in the <head>:

>but have you tried to hide it higher up while maintaining it in the head info? (e.g. template, view config)

I don't like the CSS solutions in general because as a developer I will see that it is supposed to be rendering based on the code/config but will have to dig into the theme to find out why it isn't.

Chrome and Opera show the plaintext which is what I would expect. Doesn't break anything, doesn't hurt anything, might make a user go, "hmm," but the back button is still an option.

FWIW, this is my Safari experience which I don't think is broken either:
Screen Shot 2022-01-25 at 12 22 35 PM

quamsta:
@joewhitsitt I hear you. Today's discussion during the grooming/mid-sprint demo was my attempt at getting visibility on this after I brought it up in @mark-bennett-uiowa's issue (#4435)  a couple of weeks ago, but I'm going back to the drawing board after today's discussion.

Leaving this PR as a draft, and we can continue discussion over on the issue? I have some ideas after today's meeting. Thanks for your help working through this with me.

https://github.com/uiowa/uiowa/issues/4695

quamsta:
Closing in favor of https://github.com/uiowa/uiowa/pull/4714

PR #4702, opened 2022-01-25
UI Press book content layout and design

### LINKED ISSUE #4657: UI Press - Implement layout and design for book content type
# Motivation

After #4636 is merged and #4522 has been approved by the client, then we'll need to do some work to actually implement the design.

# Proposed solution

- [x] Enable Layout Builder on the content type (but not per item).
- [x] Build a layout to match the mockups (and existing best practices, ~such as a content moderation section at the top~).
  - I think they want a five column grid of books across the bottom of the page
  - We will skip implementing the tabs in https://github.com/uiowa/uiowa/issues/4522#issuecomment-1016517171 until we can get customer approval. They can just be implemented as stacked areas for now.
- [x] Create any necessary templates to get fields look right.
- [x] Write custom CSS and JS as necessary (try to use existing classes before writing new CSS)
  - The "Praise" section uses the blockquote component without the borders and margins: https://github.com/uiowa/uids/issues/624.

e-marie-w:
Resolves #4657 

## How to test ##
Pull down branch
`drush @uipress.local uli`
Trying adding a book content type. Use a book from the current site to make it easier/allow for proper testing of cart options.
Test the Add to Cart and View Cart buttons. Make sure the formats added (ebook, paperback, etc) show up as options to add to cart. 
Test that the sales price and sale code display when within a sale date.


e-marie-w:
Ran into this error when trying to view our latest updates:
`TypeError: Argument 1 passed to Twig\Template::display() must be of the type array, null given, called in /var/www/html/vendor/twig/twig/src/Environment.php(418) : eval()'d code on line 52 in Twig\Template->display() (line 376 of /var/www/html/vendor/twig/twig/src/Template.php).`

e-marie-w:
Looks like the new permissions related to intranet got set as removed, will need to change

joewhitsitt:
> Looks like the new permissions related to intranet got set as removed instead of added, will need to change them back.

Was the Intranet split turned on when the export ran?

e-marie-w:
> > Looks like the new permissions related to intranet got set as removed instead of added, will need to change them back.
> 
> Was the Intranet split turned on when the export ran?

No, so I should just delete those lines of code instead, right? Because they're already set correctly in default, this is just extending default as needed for ui press.

joewhitsitt:
Yeah, you can manually remove those, but it would be good to figure out why it is happening.

e-marie-w:
I figured I pulled or pushed when I should've done the other first? I know we'll be exporting config again, we can see what happens then.

joewhitsitt:
when you are pulling are you running composer install and cim on the site?

e-marie-w:
No. Should I be every time I pull?

joewhitsitt:
depends on what comes with a pull (configuration permission updates, updated composer dependencies), if you see an .install modification, run updb. these are done during a sync so if you aren't syncing (trying to preserve local content), you probably should.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/src/Plugin/Block/CartButtons.php
```diff
@@ -0,0 +1,52 @@
...
+      $href = 'https://cdcshoppingcart.uchicago.edu/Cart/ChicagoBook.aspx?ISBN=' . $isbn . '&PRESS=iowa';
+      $add_to_cart_button =
+        '<a class="bttn bttn--primary bttn--caps" target="none" href=' . $href . '>
+        Add to Cart <span class="fa-arrow-right fas"></span>
+      </a>';
+      $view_cart_button =
+        '<a class="bttn bttn--secondary bttn--caps" target="none" href="https://cdcshoppingcart.uchicago.edu/Cart/Cart?PRESS=iowa">
+          View Cart <span class="fa-arrow-right fas"></span>
+      </a>';
+      $markup = '<div class="layout-builder-block">' . $add_to_cart_button . $view_cart_button . '</div>';
```
pyrello: I would really like to see this moved into a template. This block type should have associated templates that could be overridden.

pyrello:
@e-marie-w FYI, I changed the terminology in the issue summary from "Tracks" to "Resolves" because it connects the PR with the issue. "Tracks" does not do the same thing. Even if you have not completed the work in a PR so that it resolves the related issue, you can still go ahead and use the "Resolves" terminology and people will understand that means that is what the PR will eventually do.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/block--cartbuttons-block.html.twig
```diff
@@ -0,0 +1,14 @@
+{% if content['#href'] != '' %}
+  <div class="layout-builder-block">
```
pyrello: Why are we wrapping this div around this?

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/block--cartbuttons-block.html.twig
```diff
@@ -0,0 +1,14 @@
...
+    <a class="bttn bttn--primary bttn--caps" href="{{ content['#href'] }}">
+      Add to Cart <span class="fa-arrow-right fas"></span>
+    </a>
+    <a class="bttn bttn--secondary bttn--caps" href="https://cdcshoppingcart.uchicago.edu/Cart/Cart?PRESS=iowa">
+      View Cart <span class="fa-arrow-right fas"></span>
+    </a>
+  </div>
+{% else %}
+  <div class="layout-builder-block">
+    Cart buttons block
```
pyrello: Probably we should just display nothing rather than this?

e-marie-w: I think that was left over from testing? Or me needing to check that an else wasn't required for every if in twig? Regardless I'll get it removed.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/block--cartbuttons-block.html.twig
e-marie-w: Not sure I follow. Do they not need to be in a div? Or is the div something that should be added in php instead of twig?

pyrello: They may need to be in a `<div>`. I'm not sure if the templating inheritance results in them be wrapped already or not. I guess the thing I was asking about is the use of the `.layout-builder-block` class.

config/sites/uipress.uiowa.edu/config_split.config_split.site.yml
```diff
@@ -14,6 +14,8 @@ module:
 theme: {  }
 complete_list:
   - config_split.config_split.site
+  - core.entity_view_display.node.book.default
+  - core.entity_view_display.node.book.teaser
```
pyrello: I think that these should not be necessary to add, as they should be covered by `node.type.book`.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/src/Plugin/Block/CartButtons.php
```diff
@@ -0,0 +1,44 @@
...
+    $node = \Drupal::routeMatch()->getParameter('node');
+    if ($node) {
+      $pid = $node->get('field_book_type')->getValue()[0]['target_id'];
+      $paragraph = Paragraph::load($pid);
+      $isbn = $paragraph->get('field_book_isbn')->getValue()[0]['value'];
+      $href = 'https://cdcshoppingcart.uchicago.edu/Cart/ChicagoBook.aspx?ISBN=' . $isbn . '&PRESS=iowa';
+    }
+    else {
+      $href = '';
+    }
```
pyrello: This can be replaced by setting `$href = '';` before the if statement. Then it is an empty string by default and only gets set if `$node` exists.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/field--node--field-book-type.html.twig
```diff
@@ -0,0 +1,49 @@
...
+ * - field_type: The type of the field.
+ * - label_display: The display settings for the label.
+ *
+ *
+ * @see template_preprocess_field()
+ */
+#}
+{{ attach_library('uids_base/accordion') }}
+
+{% set id = "accordion-book-type" %}
```
pyrello: Since `id` is only used once in this file, I don't think it needs to be defined as a variable. You can just add it as a literal string value where it is currently being called as `{{ id }}`

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/field--paragraph--field-book-sale-code--book-type--default.html.twig
```diff
@@ -0,0 +1,5 @@
+<div{{ attributes.addClass('alert alert-info') }}>
+  Please use promo code <strong>{% for item in items %}
+  {{ item.content }}
+  {% endfor %}</strong> when prompted by shopping cart
```
pyrello: Let's try replacing the whole `{% for ... endfor %}` block with `{{ items.0.content }}`

docroot/sites/uipress.uiowa.edu/modules/uipress_core/templates/paragraph--book-type--default.html.twig
```diff
@@ -0,0 +1,69 @@
...
+{% block paragraph %}
+  <div{{ attributes.addClass(classes) }}>
+    {% block content %}
+      {% set label = 'heading' ~ paragraph.id() %}
+      {% set id = 'pid' ~ paragraph.id() %}
+
+      <h2 {{ attributes.addClass(['accordion__heading']) }} id="{{ label }}">
+        <button class="accordion__button" role="tab" aria-selected="false" aria-expanded="false" aria-controls="{{ id }}">
+          {% set ownership = content.field_book_ebook_ownership.0 %}
+          {{ content.field_book_type.0 }}{{ ownership ? ', ' }}{{ ownership }} {{ ownership ? 'ownership' }}
```
pyrello: I would like to move the setting of these variables into the appropriate preprocess function.

docroot/sites/uipress.uiowa.edu/modules/uipress_core/src/Plugin/Block/CartButtons.php
e-marie-w: Else statement removed, `$href = '';` moved to before the if statement.

dnkemper:
I can look at the magic behind this 🥳

bspeare:
Screen Shot 2022-02-09 at 9 53 02 AM

Should we hide the "Subtitle" field until we know whether we are going to merge it with the title field?

bspeare:
I think we can probably hide the "Summary" and "Review" field titles?

Screen Shot 2022-02-09 at 9 56 08 AM

dnkemper:
1. The date on the v1 site requires a month and year (not day), does that matter?
2. Can’t select related reads that don’t exist - does that matter?
3. URL = first word of book title, not entire title. Does that matter?

bspeare:
The mockup (https://user-images.githubusercontent.com/1036433/145111995-b90d243e-2291-4579-a52d-294633782922.png) contains a horizontal rule dividing the buttons and the summary content, do you want to add this horizontal rule to the button block?  `<hr class="element--spacer-thin" />`

bspeare:
Should we change the "Interior Images" to use Photoswipe?

Screen Shot 2022-02-09 at 10 05 35 AM

docroot/sites/uipress.uiowa.edu/modules/uipress_core/sass/book.scss
```diff
@@ -0,0 +1,8 @@
+.field--name-field-book-file-uploads {
```
bspeare: I think you can get rid of this CSS and possibly the workspace and add a two-column grid classes (`.grid--twocol--50-50`).   I think we want evenly displayed columns for this.  Using flexbox here means that you can display an infinite amount of images in a row:

Screen Shot 2022-02-09 at 10 08 22 AM

bspeare:
Also getting some errors if I go to the "Test" book or save the test book. 

Screen Shot 2022-02-09 at 10 14 33 AM

bspeare:
Also, should we hide the buttons and ebook accordion if there is no content?  They currently are displaying on https://uipress.uiowa.ddev.site/test. Screen Shot 2022-02-09 at 10 27 53 AM

bspeare:
@e-marie-w let us know when this is ready to be reviewed again.   @dnkemper and I are both waiting for the above issues to be resolved.  Maybe move this back to draft?

pyrello:
> Also, should we hide the buttons ebook accordion if there is no content? They currently are displaying on https://uipress.uiowa.ddev.site/test.

This comment is a little confusing. I'm not sure about hiding the accordion. If someone has created a book type and not filled it in, then should it not be displayed. We may want to check in with the customer for this.

bspeare:
@pyrello Yes, I updated it to include the missing "and" in the comment.

e-marie-w:
> I think we can probably hide the "Summary" and "Review" field titles?

Done! I also hid the labels for the other stacked fields (Awards, Table of Contents, etc). 

> do you want to add this horizontal rule to the button block? `<hr class="element--spacer-thin" />`

Also done!

> Also getting some errors if I go to the "Test" book or save the test book.
> Also, should we hide the buttons ebook accordion if there is no content? 

Requiring the ISBN seems to fix those errors. It also takes care of the accordion not having any content. I'll push up the change shortly.

> URL = first word of book title, not entire title. Does that matter?
> Should we hide the "Subtitle" field until we know whether we are going to merge it with the title field?

We should probably set up the URL to take the exact title of the book (aka not drop 'The' all the time). So I'm thinking we have a separate subtitle field, but the title and subtitle display as one. Thoughts?

> The date on the v1 site requires a month and year (not day), does that matter?

Will check with customer.

> Can’t select related reads that don’t exist - does that matter?

I believe this is the expected behavior. A book has to be added for it to be selectable.

> Should we change the "Interior Images" to use Photoswipe?
> I think you can get rid of this CSS and possibly the workspace and add a two-column grid classes (.grid--twocol--50-50). 

Still working on these two.

pyrello:
> Also getting some errors if I go to the "Test" book or save the test book.

I'm assuming this case occurred because you did not save a book type with an ISBN, @bspeare? @e-marie-w We'll need to revisit the code for the cart buttons to see how it needs to be adjusted in case there are no ISBN's. We may need to adjust the template to only display the "Add to cart" button if there is a `content.href` available (which we took out before because I didn't understand why it was necessary).

pyrello:
>> The date on the v1 site requires a month and year (not day), does that matter?

> Will check with customer.

I believe this is a limitation with how we can set up the field in D9. I'm not sure it is possible for us to not include the day. If that is the case, then we can have the customer put in the first day of the month and then not display the day in the output, if it is being used like that.

pyrello:
Based on the removal of the custom SCSS file, you can also remove the following:
* `uipress_core.libraries.yml`
* `package.json` in `uipress_core` module
* Entry added to root `package.json` file referencing the above file

e-marie-w:
ISBN field set to required based off customer's response. The .scss file is in use again so I left all that alone. Left date display as-is for now. URL pattern now set to `/books/[node:title]`. It still drops 'the', 'of', 'a', etc but I'm leaving that as-is for now too. Subtitle is still just hidden, not removed completely. Tabs and updated view cart button will be handled in their own PRs. If I haven't missed anything I'll mark this as ready for review again.

dnkemper:
I can test this! @bspeare want to bring visibility to you because you're testing skills are top tier

dnkemper:
Talked with Marie. Looked at publication details. It seems to be rendering labels even when content isn't added to the fields. What's super strange is that changing the view settings, doesn't change the rendering? CURIOSITY IS KILLING THIS CAT
Screen Shot 2022-02-14 at 4 19 19 PM

bspeare:
Here are the two things that need to be adjusted in the view in order to hide the empty views fields:

**Choose to show content over fields:**

Screen Shot 2022-02-15 at 8 26 02 AM

**And force using fields:**

Screen Shot 2022-02-15 at 8 25 50 AM

joewhitsitt:
>Here are the two things that need to be adjusted in the view...

That is a better solution than what I was going to offer up. Admissions has a couple views that do a bunch of empty field filter checks...

Background: Core doesn't handle the empty very well https://www.drupal.org/project/drupal/issues/953034

bspeare:
I just checked my solution again and I don't think it's working as expected - @joewhitsitt do you want offer up your suggestion?

joewhitsitt:
If you are settled on views fields instead of a view mode and doing the striptags/trim stuff in twig then https://admissions.uiowa.ddev.site/admin/structure/views/view/areas_of_study_block is an example where we had to add a bunch OR checks on these fields as the hide on empty (field) and hide if block is empty (view) checkboxes weren't working.

Screen Shot 2022-02-15 at 8 42 50 AM

dnkemper:
Another way could be to rewrite the field as custom text.. add for instance `<div class="field__label">Publication Date</div> {{ field_book_publication_date }}`

dnkemper:
Also, Marie was hoping to pair on this part after standup 😬

dnkemper (approved):

dnkemper:
@richardbporter @joewhitsitt would it be easier to merge this before or after https://github.com/uiowa/uiowa/pull/4741 is merged?

joewhitsitt:
🤷

richardbporter:
This

Issue #4651, opened 2022-01-13
Fix dynamic block caching issues


joewhitsitt:
# Problem

Our dynamic blocks can display stale data because there is no Drupal action to purge the corresponding Varnish page caches when the data changes. Current workarounds are: 
- edit and save the page(s) the block is on to trigger a purge.
- clear the varnish via AC interface or Drush (requires developer)

Otherwise, data will refresh naturally when the page cache max-age is reached. This is currently 32 days although longer is recommended: https://git.drupalcode.org/project/purge/-/blob/8.x-3.x/modules/purge_ui/purge_ui.module#L35

https://www.drupal.org/docs/drupal-apis/cache-api/cache-max-age#s-limitations-of-max-age

# Proposed Solutions

1. Set max-age: 0 on block render arrays.
  a. Blocked by upstream issue https://www.drupal.org/project/drupal/issues/2352009.
  b. Have to use in combo with https://www.drupal.org/project/cache_control_override for now. Module only supports max-age: 0, not 5/10/15/etc. minutes.
2. Re-write block to get and render data in JS.
3. Use a custom cache tag and clear regularly via cron.
  a. This would be unreliable since cron is currently web-based and only runs every 3 hours. 
  b. Additional performance considerations needed for running cron more frequently.
4. Lower page cache value so pages expire more frequently.
  a. https://docs.acquia.com/cloud-platform/performance/varnish/drupal/#cache-settings-in-drupal-9 
  b. Additional performance considerations needed for more web server hits.
5. Refactor events functionality and content hub to allow posting events thus relying on Drupal's regular purging mechanisms to handle clearing caches on site where event was created/edited.

### uiowa_hours
The block is placed on many, probably high-traffic, pages across the RecServ site. Option 1 will essentially disable caching on those pages which is not great for performance.

Option 2 is ideal because it will keep caching enabled with a long TTL for all of the parts of the page that don't change often. Regular Drupal purging will invalidate those pages as necessary.

a. Create a route that takes a resource ID, start and end parameters, essentially just like HoursApi::getHours, that then calls that method and returns an AJAX response of hours data. One question here is whether the route should/could return formatted markup or just raw data. Formatted data would probably easier but HoursFilterForm::hoursRender might need to move to a service to be used in the route callback. The form would then not load hours by default and trigger the callback on page load and datepicker.

[...]

PR #4717, opened 2022-01-28
Hotfix: render uiowa_hours via JS.

richardbporter:
Related to #4651 

Hotfix to get caching old hours resolved ASAP. The ideal scenario is probably what I described in #4651 as setting up a new route, controller, services, etc. and then making the AJAX call directly in JS instead of autosubmitting the form. 

# How to test

- `blt sync --site recserv.uiowa.edu`
- See the update hook runs.
- Visit building pages and see hours loaded via JS.
- Log in and verify that the hours form is not submitted while editing the page in Layout Builder and that placeholder text is visible instead.
- Double check the https://recserv.uiowa.ddev.site/hours page works correctly.

Also pushed to dev for Varnish testing: https://recserv.dev.drupal.uiowa.edu


joewhitsitt:
I plan to take a look at this shortly.

joewhitsitt (approved):
looks better than a couple minutes ago! i'd like to keep working towards a more sustainable solution for dynamic blocks

docroot/modules/custom/uiowa_hours/src/Form/HoursFilterForm.php
```diff
@@ -86,18 +90,35 @@ public function buildForm(array $form, FormStateInterface $form_state, $config =
...
+      '#type' => 'submit',
+      '#value' => $this->t('Submit'),
+      '#ajax' => [
+        'callback' => '::dateFilterCallback',
+        'wrapper' => 'results-container',
+        'disable-refocus' => TRUE,
+      ],
+      '#attributes' => [
+        'class' => [
+          'element-invisible',
```
richardbporter, line 118: @joewhitsitt is this the "right" thing to use here?

joewhitsitt: if the submit button is only there for the js to use, then it should probably be a display: none instead but i don't think we have a utility class for that available and i wonder if the js would still be able to act on it.

richardbporter: Did a quick non-JS test and things aren't working so I might just go with this for now. Thanks.

PR #4719, opened 2022-01-28
Prevent media revisions

### LINKED ISSUE #4641: Prevent new media revisions from being created
Epic: https://github.com/uiowa/uiowa/issues/4338

## Issue

Media revisions create problems as it relates to garbage collection and our replace functionality.

## Proposed Solution

- Prevent new media revisions from being created on edit/save automatically or toggled by editors and see how that affects other entities.

## Notes

Administer media is the permission that allows an editor to override "create new revision" option, but I don't think that is working. Look into https://www.drupal.org/forum/support/post-installation/2018-05-01/hide-or-disable-revision-information and removing access programmatically.

joewhitsitt:
Resolves #4641 

Changes config to prevent automatically creating a new revision for each media save and removes revision information fieldset from form since there is no granular permission to control it.

# Story

- As an editor, I can can create/edit/delete media items but can not see revision functionality.
- As a developer, editing a media entity does not create new revisions.
- As a developer, saving a layout builder page with a media entity reference does not change media revisions.
- As a developer, when a media entity is deleted through the UI and the entity was not reference by any other content, I will see usage and revision information for the entity be removed and the actual file marked with a status of 0.

**Note**: file_managed entries with a status of 0 will be removed from the table and the files directory on cron after six hours

# To Test
- Sync any site and confirm config import has run.
- Open db in the database visualization software of your choice. Keep media_revisions, file_usage and file_managed tables open. Take note of the current counts.
- As any role with media permissions, add a file. Save. take note of the db tables increasing by one.
- Edit the file, replace the file with overwrite checked. notice no change to the db tables count.
- Edit the file, replace the file with overwrite unchecked. notice file_managed and file_usage increase but that media revisions does not.
- delete the file. notice that media revisions goes down, file usage goes down and the related rows in file_managed have a status of 0.


cory-skeers:
> * Edit the file, replace the file with overwrite unchecked. notice file_managed and file_usage increase but that media revisions does not.

For this step, I'm seeing the file_managed increase, but not the file_usage (and not the media_revision). But I think that's the expected behavior? The previous fid->mid usage drops off but gets replaced with the new fid->mid keeping the overall count the same, the file_managed gets an additional row for the new file, and media_revision doesn't get a new revision created.

cory-skeers (approved):
Reviewed and tested, and all looks good / functioning as expected.

Nice steps toward mending some frustrating file behavior!

PR #4723, opened 2022-01-31
Intranet Permissions

### LINKED ISSUE #4243: When all content is private, ability to assign access/role based on AD group
# Story

As an owner of a site where all content is private, I would like the ability to grant site access based on AD group(s).

# Additional Info
Parent issue #4241 

One or more AD groups should be able to be configured so that if a user is part of the specified AD group(s), an account will be created for that user with the mapped role.

If a user should be removed from an AD group, the next time they attempt to login that role would be removed from their account and they would no longer have that level of access.

Website management role access (webmaster, publisher, editor, etc.) can continue to be managed as they currently are.

# Questions

What role should be used for access to these authenticated sites? Do we need a new role? Requestor role, viewer role? I don't think we will want to use the authenticated user role.

Do Webmasters have the ability to configure these AD groups or is that something an administrator would need to do?

In the past these AD groups have been required to be Universal and cannot be nested. Assuming this still applies?

# Proposed solution

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4243

What is interesting is that the authenticated user role losing permissions forces explicit permission updates for higher roles.

# Notes

- Anonymous and Authenticated User roles **need** the `view published content` permission in order to get to the restrict ip redirection.
- /admin/* **must** also be included in the allowed paths to get to the restrict ip redirection.
- Rick set the appropriate private files directory for each multisite when generating local settings files.
- Additions to the default role config were all additions that they already had via authenticated user, so sites _should_ accept them unless they are already removing them. thanks CS2 :)

# Stories
- as a viewer role, I can see Intranet content.
- as a viewer role, unlike public sitenow v3 websites I can't view backend pages, unpublished content and can't provide revision comments.
- as an authenticated user (no other roles), I can login but I still can't see any Intranet content and the resulting page explains this to me.

The act of setting up AD groups to map to the viewer role or higher is a site by site solution and stored in configuration.

# To Do
- [x] Too many redirects issue when authenticated user goes to /admin pages. There must be some permission they need to then be properly handed off to restrict_ip for proper redirect.
- [x] Some of the permission changes are not showing in config and I suspect it is because they are not technically set in default (authenticated has permission and the rest of the roles inherit it without it actually being set). So default permissions need to be set first? then addressed in sitenow_intranet OR the permissions config needs to be changed to a complete split item.

# To Test
blt drupal:install --site default
drush en sitenow_intranet
drush cim
create a user with only the authenticated role and a user with the viewer role. masquerade as both.

Authenticated with no role:
Screen Shot 2022-01-31 at 1 40 29 PM

Viewer sees toolbar, can logout ("manage" doesn't expand anything):
Screen Shot 2022-01-31 at 4 46 06 PM


joewhitsitt:
something is causing too many redirects for the authenticated user for admin pages

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
```diff
@@ -44,18 +44,32 @@ function sitenow_intranet_restrict_ip_access_denied_page_alter(&$page) {
...
-  $page['main']['continuation_message'] = [
-    '#markup' => t('Continue by logging in with your HawkID.'),
-    '#prefix' => '<p class="text-align-center">',
-    '#suffix' => '</p>',
-  ];
+  // Display a different message to authenticated users with no other roles.
+  $roles = \Drupal::currentUser()->getRoles();
+  if (count($roles) === 1 && in_array('authenticated', $roles)) {
+    $page['main']['access_denied_message'] = [
+      '#markup' => t('You do not have sufficient privileges to access this page. Please contact an administrator.'),
```
pyrello, line 51: Since the whole point of the intranet is that there is no public-facing part of it, would it make sense to change this into a friendlier "Please login to access this page" message?

joewhitsitt: Yeah, this message is a bit different. A user would encounter this if they have an account on the site but don't have any roles (stripped by being removed from an AD group or not properly added by a webmaster). Line 57 is the start of the anonymous visitor message where it is suggested they login.

pyrello: Got it.

richardbporter:
>Rick snuck some stuff into this PR that I assume are good things to have 🤷

https://github.com/uiowa/uiowa/issues/4242#issuecomment-1011432303

I set the appropriate private files directory for each multisite when generating local settings files.

pyrello:
Pulling to test.

pyrello (approved):
Reviewed and tested. Looks good!

PR #4736, opened 2022-02-01
Hotfix - UIowa Hours - Get today's date before submit

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/4651

get today's date with offset set as value before mousedown.

going to leave my window open over night to test an anonymous refresh tomorrow.

### DIFF
--- a/docroot/modules/custom/uiowa_hours/js/filter_form.js
+++ b/docroot/modules/custom/uiowa_hours/js/filter_form.js
@@ -10,6 +10,8 @@
           $('.uiowa-hours-container', this).html('<p>Placeholder for ' + hoursConfig[0] + ' data</p>');
         }
         else {
+          let today = new Date().toLocaleDateString('en-CA', {timeZone: "America/Chicago"});
+          $('input[type="date"]', this).val(today);
           $('input[type="submit"]', this).mousedown();
         }
       });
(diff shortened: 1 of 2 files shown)


joewhitsitt:
This worked as expected on that overnight browser window.

docroot/modules/custom/uiowa_hours/js/filter_form.js
```diff
@@ -10,6 +10,8 @@
           $('.uiowa-hours-container', this).html('<p>Placeholder for ' + hoursConfig[0] + ' data</p>');
         }
         else {
+          let today = new Date().toLocaleDateString('en-CA', {timeZone: "America/Chicago"});
```
richardbporter: Shouldn't we use `en-US` here instead of Canada?

richardbporter: Nit: single quotes on `America/Chicago`. :)

joewhitsitt: my understanding is that en-CA outputs the yyyy-mm-dd format (which the input expects) and en-US outputs mm/dd/yyyy https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_tolocalestring_date_all

richardbporter: Oh I see. Thats annoying. Also found this https://www.drupal.org/project/drupal/issues/2875131.

richardbporter (approved):

PR #4743, opened 2022-02-02
Hours hotfix: prevent tabbing to submit button which is autosubmitted.

richardbporter:
I need to get a release out to fix the caching issue but noticed a keyboard navigation issue late on. Tabbing to the submit leads to wonky behavior. This should probably be done in another way but there are other various a11y issues we should address holistically later.

# How to test

- Code review.


pyrello (approved):
Reviewed, not tested.

PR #4752, opened 2022-02-03
Potential path forward with not showing country if US

### LINKED ISSUE #4569: Suppress US in office address fields
Users have reported that when they create an address for an office that they would like to remove the United states from it. The current workaround is that they are using the phone number field to list their office

example here:
https://org-iowajpec.prod.drupal.uiowa.edu/node/21/edit

## Proposed solution

Enable the [address display module](https://www.drupal.org/project/address_display) and use it to allow users to check on or off the display of the country.

dnkemper:
Closes #4569 

# How to test

1. blt ds --site=sandbox.uiowa.edu
2. drush @sandbox.local uli
3. Add a person with contact information - if country = US - country shouldn't display
4. ^^if country is != US, country should display


cory-skeers:
Visually I like this as a light-weight path forward, but there ends up being empty markup on the page.

image

Instead of replacing the value with an empty space, would it work to unset the field in this instance? Not sure if that would lead to issues elsewhere.

dnkemper:
@cory-skeers I'll double check, but from what I remember (if accurate haha) is that it removed all address fields if I unset the country field entirely? Since the address depends on that having a value. BUT will double check!

dnkemper:
@cory-skeers ah, yes. It then displays '%placeholder' w/ errors. Hmmm we could add a class that sets the display to none? Not sure if that's ideal - but would be more accurate for screen readers?

docroot/modules/custom/sitenow_people/sitenow_people.module
```diff
@@ -699,7 +699,17 @@ function sitenow_people_preprocess_paragraph(&$variables) {
...
         }
       }
+      break;
 
+    case 'uiowa_contact':
+      $admin_context = \Drupal::service('router.admin_context');
+      if (!$admin_context->isAdminRoute()) {
+        if ($paragraph->hasField('field_uiowa_contact_address')) {
+          if ($paragraph->field_uiowa_contact_address->country_code == 'US') {
+            $variables["content"]["field_uiowa_contact_address"][0]["country"]["#value"] = ' ';
```
pyrello: Can we switch these double quotes to single quotes, please?

Also, if I'm not mistaken, what the logic is doing here is checking the country code of the first entry of the multi-value `field_uiowa_contact_address` (via magic method) and then setting the value of the country field for the first entry to an empty string. It appears that this would not prevent a second entry from showing the country if the country code was `US`.

pyrello:
> I'll double check, but from what I remember (if accurate haha) is that it removed all address fields if I unset the country field entirely?

Would setting `#access` => `denied` on the country work?

dnkemper:
> > I'll double check, but from what I remember (if accurate haha) is that it removed all address fields if I unset the country field entirely?
> 
> Would setting `#access` => `denied` on the country work?

Sadly, no. Tried that :(

docroot/modules/custom/sitenow_people/sitenow_people.module
cory-skeers: Interestingly, it does seem to prevent multiple entries from hiding the US in various combinations of US and non-US addresses.

image

dnkemper: hmm It works for me when setting first entry = US and second to a different country?

cory-skeers:
> @cory-skeers ah, yes. It then displays '%placeholder' w/ errors. Hmmm we could add a class that sets the display to none? Not sure if that's ideal - but would be more accurate for screen readers?

I think that'd be closer to ideal. There's the question why we'd hide something from visual display that would be read for screen readers, but I think the request stems primarily from a visual design, so maybe that's not a concern here.

docroot/modules/custom/sitenow_people/sitenow_people.module
dnkemper: Op. Sorry, Cory! That works as desired then! 

> > @cory-skeers ah, yes. It then displays '%placeholder' w/ errors. Hmmm we could add a class that sets the display to none? Not sure if that's ideal - but would be more accurate for screen readers?
> 
> I think that'd be closer to ideal. There's the question why we'd hide something from visual display that would be read for screen readers, but I think the request stems primarily from a visual design, so maybe that's not a concern here.

I can try going about it the class route? @pyrello any objections?

dnkemper: why... did it add that comment here 🙃

pyrello: Maybe I'm misreading the code.

pyrello:
> I think that'd be closer to ideal. There's the question why we'd hide something from visual display that would be read for screen readers, but I think the request stems primarily from a visual design, so maybe that's not a concern here.

@bspeare Thoughts?

docroot/modules/custom/sitenow_people/sitenow_people.module
dnkemper: No, the way you're reading it would make perfect sense... the address field is just incredibly wonky

dnkemper:
can push up a potential change for @bspeare to look at 😇

pyrello:
@briand44 Do we want to keep this old PR around?

briand44:
@pyrello I am ok with this going away if we think a different approach (https://github.com/uiowa/uiowa/issues/4569#issuecomment-1332749738 or something else)makes more sense at this point in time.

If we are still not sure how we want to solve this it may be worth keeping until we know our path forward.

PR #4763, opened 2022-02-04
Added ical and ics to media file types

### LINKED ISSUE #4680: Add .ical and .ics file types to file media type
# Story

As an editor, I would like to provide a calendar event file for visitors to download and add to their calendars as a reminder of my upcoming event.

This should not be limited to the event content split but generally available to place within a WYSIWYG.

# Proposed Solution

- [x] Update the file media type configuration to allow .ical and .ics files

makurj:
Resolves #4680 

How to test:

1. Pull down branch
2. `blt ds --site=default.uiowa.edu`
3. Upload ics and ical files.
4. Can you access them from content blocks?


richardbporter (approved):
Worked for me.

PR #4766, opened 2022-02-04
International Programs migration tweaks

### LINKED ISSUE #4685: Review/Test International Programs News Migration
It's been awhile since we've done this work. Let's make sure all is still working before we schedule a date to run it in prod.

https://github.com/uiowa/uiowa/issues/3375

# Proposed Solution

Follow test steps on the PR https://github.com/uiowa/uiowa/pull/3716

cory-skeers:
* Remove dependency. Not all articles are migrated (older than 2015 are skipped), and without adjusting the query, it will never be "complete"

# How to test

Code review should be sufficient--check that the redirects migration definition does not have the articles migration as a dependency.

### DIFF
--- a/docroot/sites/international.uiowa.edu/modules/international_migrate/config/split/migrate_plus.migration.international_article_redirects.yml
+++ b/docroot/sites/international.uiowa.edu/modules/international_migrate/config/split/migrate_plus.migration.international_article_redirects.yml
@@ -34,6 +34,4 @@ process:
   status_code: constants/status_code
 destination:
   plugin: 'entity:redirect'
-migration_dependencies:
-  required:
-    - international_articles
+migration_dependencies: { }


richardbporter (approved):

richardbporter:
I'm realizing after the fact that this will require running the migrations one-by-one in order instead of `drush mim --all`, right? Just wanted to mention that before I forgot.

cory-skeers:
> I'm realizing after the fact that this will require running the migrations one-by-one in order instead of `drush mim --all`, right? Just wanted to mention that before I forgot.

It will, but this is one of the migrations that had memory issues in the node migrate, and needs to be run in batches with `--limit=X` already, in place of `mim --all`.

richardbporter:
Gotcha. This all sounded familiar so I just double checked old PRs and found https://github.com/uiowa/uiowa/pull/3912/commits/29ac0d7acd87187d2ad3733285c8018d3f7c205d which got around it slightly differently. Seems like both ways work. 👍🏻

PR #4770, opened 2022-02-07
Feature: intranet search

### LINKED ISSUE #4244: When all content is private, ability for users to search private content
# Story

As an owner of a site where all content is private, I would like the ability for viewer roles and above to search private content.

# Additional Info

Parent Issue #4241 

# Proposed solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/4244. 

This pr:

- Turns on Search API for sitenow_intranet
- Adds a custom search view for sitenow_intranet
- Adds a search region to `uids_base` and a corresponding search block for `uiowa_search`
- Makes the custom search form use placeholder text from the "Search label" set in the form

# How to test
```
blt frontend
blt drupal:install --site default
drush en sitenow_intranet
drush cim
```
- Test that the search block takes you to `/search` which is a custom view search page. 
- Run cron and verify that the site search works by searching for `ITS`.


docroot/themes/custom/uids_base/templates/layout/header--homepage.html.twig
```diff
@@ -6,11 +6,12 @@
...
-        {% include '@uids_base/uids/global/search-toggle.html.twig' with {
-            search_form: uiowa_search_form
-        } %}
-    {% endif %}
+  {% if page.search %}
+    {% include '@uids_base/uids/global/search-toggle.html.twig' with {
+      'search_form': page.search,
+      'under_search': page.under_search,
+    } %}
+  {% endif %}
```
pyrello, line 14: It's hard to tell for sure, but the indentation seems off for this section compared to the sections above and below it.

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -183,15 +183,6 @@ function uids_base_preprocess_page(&$variables) {
...
 
-  // Hide the search form by default.
-  $variables['uiowa_search_form'] = FALSE;
-
-  if (\Drupal::service('module_handler')->moduleExists('uiowa_search')) {
-    if (\Drupal::config('uiowa_search.settings')->get('uiowa_search.display_search')) {
-      $variables['uiowa_search_form'] = \Drupal::formBuilder()->getForm('Drupal\uiowa_search\Form\SearchForm');
-    }
-  }
-
```
cory-skeers: Do we need to keep this in, but update it to match the page.search format instead of uiowa_search_form used in the template later? (Looking at the uiowa_search test failure.)

bspeare: Yes, I was thinking we need to add that back after the conversation this morning!

cory-skeers: Pushed up a quick change--might need some alterations yet to play nicely with everything else, but should fix the build.

bspeare: Thanks @cory-skeers !

richardbporter:
I can replicate the test failure locally with `ddev phpunit docroot/modules/custom/uiowa_search/`. The search form is not in the header in the test screenshots. Still trying to understand what is causing it...

richardbporter:
Whoops, missed the previous conversation. Anyway, looks like the test started failing again after c3d1cbbd25a54c9d13ff693ac8da1b9ab52eb889.

bspeare:
Thanks @richardbporter, I think I need to check the templates first.   I hadn't noticed that missing.

richardbporter:
> Whoops, missed the previous conversation. Anyway, looks like the test started failing again after [c3d1cbb](https://github.com/uiowa/uiowa/commit/c3d1cbbd25a54c9d13ff693ac8da1b9ab52eb889).

I think the test will need to be updated to place that new block since that is controlling the render of the form. Not sure how to do that exactly.

bspeare:
> I think the test will need to be updated to place that new block since that is controlling the render of the form. Not sure how to do that exactly.

Do we want to alter the test or keep this line to make the test pass?   https://github.com/uiowa/uiowa/blob/b38a5504fe06ce834fbeb6f30dd6e0223b551f93/docroot/themes/custom/uids_base/uids_base.theme#L192

bspeare:
I think this is ready to review; however, we still need to decide what the plan is with https://github.com/uiowa/uiowa/pull/4770#issuecomment-1034210857.

cory-skeers:
> > I think the test will need to be updated to place that new block since that is controlling the render of the form. Not sure how to do that exactly.
> 
> Do we want to alter the test or keep this line to make the test pass?
> 
> https://github.com/uiowa/uiowa/blob/b38a5504fe06ce834fbeb6f30dd6e0223b551f93/docroot/themes/custom/uids_base/uids_base.theme#L192

I have the test partway working with the block...Now it's failing on a later part!

I can keep plugging away at it while the rest of the functionality gets tested, though.

cory-skeers:
> > > I think the test will need to be updated to place that new block since that is controlling the render of the form. Not sure how to do that exactly.
> > 
> > 
> > Do we want to alter the test or keep this line to make the test pass?
> > https://github.com/uiowa/uiowa/blob/b38a5504fe06ce834fbeb6f30dd6e0223b551f93/docroot/themes/custom/uids_base/uids_base.theme#L192
> 
> I have the test partway working with the block...Now it's failing on a later part!
> 

Current testing status (when removing the above line from `uids_base.theme`.

In the test setup, search toggle button correctly appears and is interactive, but the actual search field is not appearing. Somewhere along the line, the search_form isn't getting set for the template (page.search in the variables), and so ends up missing. The line in `uids_base.theme` makes sure it's added in, but masks the disconnect. Should be something that needs to get added to the test that isn't, but still unsure where the disconnect is.

image

bspeare:
It looks like https://github.com/uiowa/uiowa/blob/b38a5504fe06ce834fbeb6f30dd6e0223b551f93/docroot/themes/custom/uids_base/uids_base.theme#L192 is causing the search toggle button to appear when logged out of the site.   If I remove that line, then it disappears for anonymous users but still appears for authenticated users.

cory-skeers:
> It looks like
> 
> https://github.com/uiowa/uiowa/blob/b38a5504fe06ce834fbeb6f30dd6e0223b551f93/docroot/themes/custom/uids_base/uids_base.theme#L192
> 
> is causing the search toggle button to appear when logged out of the site. If I remove that line, then it disappears for anonymous users but still appears for authenticated users.

Dang.

Removed the line and added a null check to the test as a temporary workaround. Will want to remove it when we track down how to get the test working properly, but for now it allows us to remove that line, keep the rest of the test running, and have things in place for a followup fix.

joewhitsitt:
pulling down to test

joewhitsitt:
Looking through issue/pr comments I was curious why it was set to match the whole word? db size concerns? As a user, I was getting frustrated that "herky" didn't work but "Herky" did.

Update: changing it to match partial word doesn't seem to help. "herky" still returns no results but "erky" does...

bspeare:
@joewhitsitt Those are good questions!  Do you want to make suggestions or changes that work better?

joewhitsitt:
yeah, i am poking at it to learn more. first timer here.

joewhitsitt:
There might be some other processors we may want to add in the future https://www.drupal.org/docs/8/modules/search-api/getting-started/processors

joewhitsitt (approved):
Tested the test. Reviewed the search. Skimmed the code. Checked a non-intranet site. Added a readme. Looks good.

bspeare:
It would be nice to see a Percy snapshot on this before merging so I am going to move this back to draft until our plan resets on Monday.

Issue #4208, opened 2021-09-28
As a webmaster, I would like to embed TikTok videos with a preview thumbnail


makurj:
"Is it possible to embed/post TikTok videos in our Drupal system? (With the preview thumbnail?)
I’m trying to do just that on this page and can’t quite get there: https://bam.sites.uiowa.edu/videos"

Robert R Cargill

quamsta:
@mjoneill, here's my exploration on this in the PR:

https://github.com/uiowa/uiowa/pull/4784

joewhitsitt:
marking as blocked, needs discussion on whether we pursue this further based on https://github.com/uiowa/uiowa/pull/4784#issuecomment-1054709426

quamsta:
Example TikTok embed code that's provided by TikTok in case we need to investigate alternate routes/allowing iframes, script tags or other elements and attributes:

```
<blockquote class="tiktok-embed" cite="https://www.tiktok.com/@uiowa/video/7082749025522404650" data-video-id="7082749025522404650" style="max-width: 605px;min-width: 325px;" >
   <section> <a target="_blank" title="@uiowa" href="https://www.tiktok.com/@uiowa">@uiowa</a> Our favorite place. 💛🖤 <a title="uiowa" target="_blank" href="https://www.tiktok.com/tag/uiowa">#uiowa</a> <a title="iowa" target="_blank" href="https://www.tiktok.com/tag/iowa">#iowa</a> <a title="collegelife" target="_blank" href="https://www.tiktok.com/tag/collegelife">#collegelife</a> <a target="_blank" title="♬ original sound - UIowa" href="https://www.tiktok.com/music/original-sound-7082748952600234794">♬ original sound - UIowa</a> </section>
</blockquote>
<script async src="https://www.tiktok.com/embed.js"></script>
```

joewhitsitt:
darn, is there just an iframe option without the additional JS?

quamsta:
Not really. It's technically possible to get the generated iframe that's built via that embed.js script and place it back in CKEditor's view source mode, but that's a pretty big ask for the content editor.

joewhitsitt:
Then the "allow this iframe domain" approach is not going to work for TikTok :(

joewhitsitt:
Is this the nail in the coffin for this work? https://www.iowaregents.edu/news/board-news/statement-from-president-mike-richards-on-governor-reynolds-directive-to-ban-access-to-tiktok

[...]

PR #4784, opened 2022-02-09
Added TikTok to the allowed video providers

quamsta:
Relates to https://github.com/uiowa/uiowa/issues/4208

This config edits adds TikTok to the allowed video providers

# How to test

- Add a TikTok video to a local or test site, like the one below:

https://www.tiktok.com/@quamsta/video/7032277729961757999

-  Enjoy the spinning cat on your website


joewhitsitt:
I think this is involved too: https://github.com/uiowa/uiowa/blob/main/config/default/oembed_providers.bucket.video.yml

Have you tested each of the remote video sizes with and without captions? uids_base treats remote video with a 16:9 assumption https://sandbox.prod.drupal.uiowa.edu/media

noticed your video example is horizontal but I thought vertical was more popular.

quamsta:
> I think this is involved too: https://github.com/uiowa/uiowa/blob/main/config/default/oembed_providers.bucket.video.yml
> 
> Have you tested each of the remote video sizes with and without captions? uids_base treats remote video with a 16:9 assumption https://sandbox.prod.drupal.uiowa.edu/media

Will check on these! Thanks!

quamsta:
Hmmm, there's some strangeness happening with `Permission-Policy header` when I test this locally.

I added TikTok to `config/default/oembed_providers.bucket.video.yml` but I get this message in my browser when trying to render the iframe: 

``Error with Permissions-Policy header: Unrecognized feature: 'interest-cohort'.``

And it seemingly prevents the actual iframe from loading despite a seemingly successful embed (thumbnail is loaded, media is added to media library, etc). Is TikTok trying to do something... not great, in regards to tracking that we don't allow on our sites?

quamsta:
> Hmmm, there's some strangeness happening with `Permission-Policy header` when I test this locally.
> 
> I added TikTok to `config/default/oembed_providers.bucket.video.yml` but I get this message in my browser when trying to render the iframe:
> 
> `Error with Permissions-Policy header: Unrecognized feature: 'interest-cohort'.`
> 
> And it seemingly prevents the actual iframe from loading despite a seemingly successful embed (thumbnail is loaded, media is added to media library, etc). Is TikTok trying to do something... not great, in regards to tracking that we don't allow on our sites?

Actually, after investigating this further, it  may be an issue with the local environment + oembed not working correctly, will need to test this on one of our test environments.

quamsta:
Just did some testing on dev and got the same result on remote servers. Did some more research and this is due to Drupal blocking Google's Federated Learning of Cohorts (FLoC).

This PR blocked until some discussion around this can occur:

["Permissions-Policy header is added by default to disable a method of browser-based user tracking" ](https://www.drupal.org/node/3213197)

https://en.wikipedia.org/wiki/Federated_Learning_of_Cohorts
```
/**
 * Add Permissions-Policy header to disable Google FLoC.
 *
 * By default, Drupal sends the 'Permissions-Policy: interest-cohort=()' header
 * to disable Google's Federated Learning of Cohorts feature, introduced in
 * Chrome 89.
 *
 * See https://en.wikipedia.org/wiki/Federated_Learning_of_Cohorts for more
 * information about FLoC.
 *
 * If you don't wish to disable FLoC in Chrome, you can set this value
 * to FALSE.
 */
# $conf['block_interest_cohort'] = TRUE;

```

Is this something we'd feel comfortable as a group disabling?

quamsta:
Other browsers also seem to hate TikTok embeds for different, but likely (tracking)related reasons:

Safari:
Screen Shot 2022-02-28 at 4 16 22 PM

Firefox:
Screen Shot 2022-02-28 at 4 16 46 PM

quamsta:
More TikTok investigation: it looks like the embed that gets generated via oembed requires the attribute `allow` on the `<iframe>` element, the following embed taken from an oembed testing suite actually works just fine if you paste it into the view source mode in a page editor and allow "Full HTML": 

```html
<iframe allow="autoplay; encrypted-media" allowfullscreen="" frameborder="0" scrolling="no" src="https://www.tiktok.com/embed/6881684188173585670" style="top: 0px; left: 0px; width: 100%; height: 100%; position: absolute;"></iframe>
```

Whereas the generated `<iframe>` from the Drupal/Remote Video embedding process "doubles up" the iframes and puts an iframe within an iframe and eventually omits the `allow="autoplay; encrypted-media"` attribute/properties entirely: 

```html
<iframe src="/media/oembed?url=https%3A//www.tiktok.com/%40uiowa/video/7070915486774660398%3Fis_copy_url%3D1%26is_from_webapp%3Dv1&amp;max_width=0&amp;max_height=0&amp;hash=kBG6WUyUDp-33nxsI5IwbTjkt601Uta-xKMvio9nySM" frameborder="0" allowtransparency="" width="100" height="100" class="media-oembed-content" title="#1 in our hearts. @therealherky #mascots #uiowa"></iframe>
```

which eventually leads to this iframe:

```html
<iframe name="__tt_embed__v34668357545413820" src="https://www.tiktok.com/embed/v2/7070915486774660398?lang=en-US&amp;referrer=https%3A%2F%2Fdefault.uiowa.ddev.site%2Fmedia%2Foembed%3Furl%3Dhttps%253A%2F%2Fwww.tiktok.com%2F%2540uiowa%2Fvideo%2F7070915486774660398%253Fis_copy_url%253D1%2526is_from_webapp%253Dv1%26max_width%3D0%26max_height%3D0%26hash%3DkBG6WUyUDp-33nxsI5IwbTjkt601Uta-xKMvio9nySM" style="width: 100%; height: 648px; display: block; visibility: unset; max-height: 648px;"></iframe>
```

Is there a way to inject or add an `allow` attribute and properties to the resulting iframe?

quamsta:
Confirmed that adding `allow="autoplay; encrypted-media"` to a copied/pasted `<iframe>` element actually allowed the embed to be rendered:
Screen Shot 2022-03-09 at 8 36 45 AM

joewhitsitt:
Technically, a Filter plugin could be created similar to `docroot/modules/custom/uiowa_core/src/Plugin/Filter/FilterIcon.php` which is added to our text formats.

I'd like to understand more about how/why `allow` is getting stripped.

New territory for me, but it sounds like Google doesn't like `encrypted-media` and Drupal uses Google Permission-Policy
https://www.chromium.org/Home/chromium-security/deprecating-permissions-in-cross-origin-iframes/

joewhitsitt:
Posted a question to the community here https://drupal.slack.com/archives/C1VLTJJD8/p1646929324550269

joewhitsitt:
Not sure if it is related but someone pointed out https://www.drupal.org/project/drupal/issues/3225184

pyrello:
Given the recent ban on TikTok for U of I hardware, wondering what the status of this should be?

joewhitsitt:
I think it can be closed because we closed the parent issue for similar reasons https://github.com/uiowa/uiowa/issues/4208

PR #4790, opened 2022-02-10
revert timeline updates

GaryRidgway:
my work for timelines has rendered them unusable, it needs to be removed.

# How to test

test that you can create timeline items and that you can add items to timelines


dnkemper (approved):
can add, edit, save

PR #4818, opened 2022-02-16
Custom time-based cache tag

### LINKED ISSUE #4786: Create custom cache tag for sitenow_events block render
First slice of #4651. Should be investigated first to make sure it is doable. 

https://www.drupal.org/docs/drupal-apis/cache-api/cache-tags

### Proposed solution
Add `time:hours_3` cache tag in `sitenow_events_preprocess_block` to block render. Push to dev and test Varnish cache is cleared when tag is purged manually with `drush @mysite.dev pinv tag time:hours_3` for pages that the events view is placed on. Inspect response headers to verify cache clear. Bonus points for testing an update to an event is displayed after tag is purged.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4786

- Make sure the latest commit makes it out to uiowa DEV
- Checkout https://default.dev.drupal.uiowa.edu/about and see a cached event.
- Edit https://content.dev.drupal.uiowa.edu/node/82386/edit and change the title.
- Give it a moment and run `drush @default.dev pinv tag time:hours_3`
- Reload and see your updated title
- ??
- Profit


joewhitsitt:
deploying to dev

joewhitsitt:
deploying to dev again

richardbporter (commented):
I'm still not seeing it update. I will revisit tomorrow.

joewhitsitt:
fwiw i refreshed this AM and saw the update that was made "by me"

richardbporter:
I was able to see see the update from an anonymous, cached page after ~5 minutes by first invalidating the tag via `drush @default.dev cache:tag time:hours_3` and then `drush @default.dev pinv tag time:hours_3`. Seems like both are necessary.

richardbporter:
Testing the single event page...

joewhitsitt:
then if both of those are necessary, should we remove the max-age again?

richardbporter:
> then if both of those are necessary, should we remove the max-age again?

Yea we could try that. I can't remember if that had an impact on seeing the updated data while logged in or not. If not, then it doesn't seem necessary to set max-age to 0.

richardbporter:
> Testing the single event page...

Same result. 👍🏻

joewhitsitt:
I think max-age is still needed if we go this route https://default.dev.drupal.uiowa.edu/about

richardbporter:
> I think max-age is still needed if we go this route https://default.dev.drupal.uiowa.edu/about

Thanks, I will get back to this later.

richardbporter (approved):
I tested 60 seconds and that worked as well. So our hook_cron implementation will need to invalidate and purge the tag it looks like.

richardbporter:
> I tested 60 seconds and that worked as well. So our hook_cron implementation will need to invalidate and purge the tag it looks like.

Maybe this could be set to 5 minutes like the other caches but lets go with this for now. I'm sure all this will need to be adjusted as we get further into it.

PR #4819, opened 2022-02-16
update core from 9.3.5 to 9.3.6

joewhitsitt:
No update hook, no obvious config changes when syncing from STAGE.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4319,16 +4319,16 @@
         },
         {
             "name": "drupal/core",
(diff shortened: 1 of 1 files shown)


joewhitsitt:
since 9.3.5 isn't in PROD yet, testing should be done with a sync from STAGE

richardbporter (approved):
https://git.drupalcode.org/project/drupal/-/compare/9.3.5...9.3.6?from_project_id=59858

PR #4821, opened 2022-02-16
Set full_html instances to filtered_html (grad.admissions AoS content)

### LINKED ISSUE #4808: Admissions - Migration caused some fields to be uneditable
# Problem

Migration caused some fields to receive the wrong format and editors don't have access to them.

- Estimated Cost of Attendance
- Application Deadlines

# Proposed solution

Loop through and re-save the fields with the correct filter format?

Some are marked as full_html (node__field_area_of_study_deadlines).

cory-skeers:
Resolves #4808 

# How to test

`git checkout gradadmissions-full-html && git pull`
`blt ds --site=grad.admissions.uiowa.edu`
Watch for a successful update hook
`drush @admissionsgrad.local uli`
Masquerade as a webmaster
Edit a node from the update hook output, check that you are able to access all fields
Spot check any areas of study not listed in the update hook output, and if they do not have any unaccessible fields for a webmaster; Can also look for any instances of `full_html` within the database field tables
Sync a site other than grad.admissions.uiowa.edu, and check that the update hook doesn't run.


mark-bennett-uiowa:
This may break several Twitter and Facebook feeds that are out there. For instance, the only way i can get a Facebook feed to work on a frontpage, is to set a textarea widget to Full HTML and enter the following:

`<iframe allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share" allowfullscreen="true" frameborder="0" height="900" scrolling="no" src="https://www.facebook.com/plugins/page.php?href=https%3A%2F%2Fwww.facebook.com%2Fiowabiosciencesacademy&amp;tabs=timeline&amp;width=500&amp;height=900&amp;small_header=true&amp;adapt_container_width=true&amp;hide_cover=false&amp;show_facepile=false&amp;appId" style="border:none;overflow:hidden" width="100%"></iframe>`

makurj:
We also have ScheduleOnce and a few other iframes embedded by changing the text area to Full HTML by an admin on Tippie's D9 and Tippie's Current Students D9 sites.

cory-skeers:
> This may break several Twitter and Facebook feeds that are out there.

> We also have ScheduleOnce and a few other iframes embedded by changing the text area to Full HTML by an admin on Tippie's D9 and Tippie's Current Students D9 sites.

This should only touch text fields on area of study content entities on grad.admissions.uiowa.edu, but those are good points--we should double check and confirm that there aren't intentional uses of full_html in these specific fields.

joewhitsitt:
pulling down to test

joewhitsitt (approved):
Thanks for cleaning this up. I have compared to troublesome nodes on .prod masquerading as Scott.

Fairly confident this will not trigger on other sites since it is contained within the site's module

cory-skeers:
Thanks for testing @joewhitsitt!

PR #4829, opened 2022-02-17
Removed colorbox from GA config (#4796)

### LINKED ISSUE #4796: Remove colorbox from GA config
Colorbox was removed and does not need to be in: 
https://github.com/uiowa/uiowa/blob/0cc4a2456dff8659354efb067fc98ea44413d5c4/config/default/google_analytics.settings.yml#L22

anantrp:
Fixes #4796


richardbporter:
Thanks for this! Unfortunately, our status checks don't seem to work with forks so our team will have to fix this.

anantrp:
That's alright

PR #4834, opened 2022-02-18
Headline alignment options for blocks

### LINKED ISSUE #2427: Implement headline refactor from UIDS into uids_base
# UIDS Headline refactor 

A headline refactor and background component was added in https://github.com/uiowa/uids/pull/392 and these changes will need to be implemented into `uids_base`. 

- [x] Update references to LBS background color names in component `scss` files.
- [x] Update class names to remove legacy classes like `bold-headline`.
- [x] Update Headline template to reflect class name refactor: https://github.com/uiowa/uids/blob/i309-event-card-variant/src/components/headline/headline.twig

### LINKED ISSUE #4402: As a site owner, I can align my block headline to the left or center
# Story

As a site owner, I can align my block headline to the left or center so that site owners don't have to create separate text areas to center their headings. 

Currently, I'm seeing a lot of site owners create separate text areas just to center a headline style.  Adding an alignment option to the Headline would help us reduce the technical debt of having custom text areas with custom headline classes when they could be using the built-in Headline field.  

# Proposed solution

- Add a "Headline alignment" select option to the Headline, these files are the main headline files to update:             
  - https://github.com/uiowa/uiowa/blob/75d37e08e6fd5de44f5dd3174db9d21286f7babe/docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
  - https://github.com/uiowa/uiowa/blob/c05630523a8d365227edb66a78d40195da167a2e/docroot/modules/custom/uiowa_core/src/Element/Headline.php
  - Possibly https://github.com/uiowa/uiowa/blob/c05630523a8d365227edb66a78d40195da167a2e/docroot/modules/custom/layout_builder_custom/src/Plugin/Field/FieldType/UiowaHeadline.php
- Use `headline--center` as modifier class to center headline.   The left default option does not need a class. 
- Pass the headline alignment option to the template https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/uids/headline.html.twig

Screen Shot 2021-11-03 at 1 24 40 PM

e-marie-w:
Closes #4402 
Closes #2427

# Story

As a site owner, I want to align block headlines to the center as well as to the left. 
Pre-existing blocks should have Left (default) set as their alignment.

# How to test

- Test a fresh site install and make sure default content can be modified and saved.
- Add new blocks and make sure switching between alignments works fine.
- Also check that switching alignments works on pre-existing blocks.
- `blt ds --site=grad.uiowa.edu` to test the Academic Dates block
- `blt ds --site=recserv.uiowa.edu` to test the Hours block.
- Blocks with the alignment option:
  - Collection 
  - Events
  - Featured Content
  - Image Gallery
  - RSS Feeds
  - Slider
  - Timeline
  - Text Area
  - Articles Listblock
  - People Listblock
  - Academic Dates (MAUI)
  - Hours
  - Sign Up Form (Sitenow Dispatch)
  - Thank You Form (Sitenow Dispatch)


cory-skeers:
Thought about it a bit more, and we still need an update hook for non-list blocks, but it shouldn't need to be as complicated as the whole swaperoo we needed for text types, since we're only adding a property, not switching field types.

This should work as a reference--might be some differences in 9, but it's pretty lightweight.
https://www.drupal.org/docs/drupal-apis/update-api/updating-database-schema-andor-data-in-drupal-8#s-adding-a-new-column

bspeare:
Here's the Percy error:

Screen Shot 2022-03-02 at 9 05 41 AM

joewhitsitt:
Pulling down to test

Update: Gah! Sry. Pulling down now 14:11

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -75,12 +76,15 @@ function uiowa_core_theme($existing, $type, $theme, $path) {
 function template_preprocess_uiowa_core_headline(&$variables) {
   if ($variables['hide_headline'] == TRUE) {
     $style = 'sr-only';
+    $alignment = 'sr-only';
```
joewhitsitt: Though this might solve the error, I am not sure this is how we should go about it. Going to poke at this a bit

joewhitsitt: the `if` check only related to hiding the headline which doesn't change what we do with alignment. moved it in https://github.com/uiowa/uiowa/pull/4834/commits/4006eba21d74017ca75bae6c58bbc5349fd131e1

e-marie-w: AWESOME, thank you Joe!!

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
```diff
@@ -40,8 +40,18 @@ public static function getHeadingSizeUp($size) {
...
+      'headline_bold_serif_underline' => 'headline headline--serif headline--underline block__headline',
+    ];
+  }
+
+  /**
+   * Get a list of valid heading alignments.
+   */
+  public static function getHeadingAlignment() {
+    return [
+      'default' => ' ',
```
joewhitsitt, line 59: So this `' '` causes extra spaces in the class attribute but `''` returns `default`. Is there a reason why we aren't using `headline--left.` I assume we don't have a uids class and it would just be ignored but not be spaces either. This isn't breaking anything but just something to think about I guess.

joewhitsitt:
Not positive, but I don't think `layout_builder_custom_theme()` is needed anymore.

```
/**
 * Implements hook_theme().
 */
function layout_builder_custom_theme($existing, $type, $theme, $path) {
  $theme = [
    'uiowa_headline_widget' => [
      'variables' => [
        'headline' => NULL,
        'heading_size' => NULL,
        'hide_headline' => NULL,
        'headline_alignment' => NULL,
        'child_heading_size' => NULL,
      ],
    ],
  ];
  return $theme;
}
```

I think this might be a remnant from the move of this functionality from layout_builder_custom to `uiowa_core_theme()`. Again, doesn't break anything but we might want to clean up later.

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
e-marie-w: `headline--left` didn't ever occur to me, brb need to try it. It would be great to not have extra spaces or a non-existent 'default' class.

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
```diff
@@ -40,8 +40,18 @@ public static function getHeadingSizeUp($size) {
   public static function getStyles() {
     return [
       'default' => 'headline block__headline',
-      'headline_bold_serif' => 'headline bold-headline bold-headline--serif block__headline',
-      'headline_bold_serif_underline' => 'headline bold-headline bold-headline--serif bold-headline--underline block__headline',
+      'headline_bold_serif' => 'headline headline--serif block__headline',
+      'headline_bold_serif_underline' => 'headline headline--serif headline--underline block__headline',
```
pyrello, line 47: This isn't strictly related to this functionality, but if there are any classes that we add every time (like `headline` and `block__headline`), we should maybe just add those when the headline is rendered rather than having them as part of these options.

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
e-marie-w: `headline--left` works great with new blocks and pre-existing listblocks. For some reason I'm having trouble with other pre-existing blocks, even after setting the default value in the update hook to `headline--left`.

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
joewhitsitt: Added a todo for another day. https://github.com/uiowa/uiowa/pull/4834/commits/d75ece34830db5ea0662dec1a6790882ce074dfd Does that work for you @pyrello ?

docroot/modules/custom/uiowa_core/src/HeadlineHelper.php
joewhitsitt: Added a todo for another day.

joewhitsitt (approved):
I think this work revealed other work unrelated to alignment. Added some todos for another day. I tested a big sync of sites and there were no errors with updb/cim. Additionally, I tested updating existing blocks, adding new, trying uiowa_hours, academic dates maui, text, slider, articles list block, uiowa_events and a couple more I can't think off right now. children still got updated based on the parent heading size and visually hidden still worked for new and existing. Overall I didn't encounter any visual regressions, particularly where the headline component moved inside the content wrapper.

default_content, i think there are some unrelated issues (children headings) but that might be solved by https://github.com/uiowa/uiowa/issues/4849

e-marie-w:
I just figured out the `headline--left` stuff, do we still want to add it to this or should it wait for another PR?

joewhitsitt:
>I just figured out the headline--left stuff, do we still want to add it to this or should it wait for another PR?

If you think it can be done and doesn't need more testing, go for it. I am tested out!

e-marie-w:
> > I just figured out the headline--left stuff, do we still want to add it to this or should it wait for another PR?
> 
> If you think it can be done and doesn't need more testing, go for it. I am tested out!

I'm not 100% confident in it so I'll just merge this without it. Thank you for the awesome thorough testing!

PR #4836, opened 2022-02-18
Find Text hotfix for Education

cory-skeers:
Issue cropped up for Find Text on education.

To reproduce: `blt ds --site=education.uiowa.edu`
`drush @education.local uli /admin/find-text`
Search for something like 'programs'
See no results, and no "no results found" message.

# How to test

Follow the reproduce steps above, only get results.
Check a non-education site, perform a text search, see that it still works.


cory-skeers:
Putting this back to draft. Forgot `node_field_data` is used to fetch the title, so it can't be dropped without breaking other functionality.

Glad we had a test in place!

cory-skeers:
I think we should create a followup task to catch issues like this elsewhere, but this can get education's search working again without impacting title search.

joewhitsitt:
Pulling down to test

joewhitsitt (approved):
Approved knowing that a follow-up issue can be created to generalize this solution more.

joewhitsitt:
Searching for "programs" on the education site made me wonder at what point do we say "too many results" or time out

PR #4853, opened 2022-02-23
Update uids to 3.5.0

richardbporter:
Resolves #4771, maybe.

# How to test

- Check out branch.
- Run `blt frontend`
- Ensure node_modules/@uiowa/uids/package.json says 3.5.0
- Test other stuff.
- yeehaw

PR #4855, opened 2022-02-23
Feature: quick fix to update extends to be includes

bspeare:
I missed a few files that needed to be updated in https://github.com/uiowa/uiowa/pull/4853, this pr fixes those that needed to be updated but did not create console errors.


bspeare:
Looks like this is not urgent as there is a fallback.  These updates will happen in https://github.com/uiowa/uiowa/pull/4667.

Issue #3420, opened 2021-04-27
Consider rabbit hole for "tags" vocabulary


joewhitsitt:
## Issue

Originally tags were pitched as an internal form of organization for use in finding content and creating filtered views. Though the appearance of content will be addressed in https://github.com/uiowa/uiowa/issues/3358, many don't know that these pages exist and think these pages should not be public. However, as it stands now, some site owners do know about them and are using the feeds that are generated.

## Proposed Solution

- [ ] See if the RSS feeds with the tags vocabulary is affected by rabbit hole.
- [ ] Use rabbit hole taxonomy to hide the tags vocabulary.

richardbporter:
Engineering reached out asking if they could sort these pages. Workaround was to use person list block but now they'd like to hide the tag pages. +1 to this.

joewhitsitt:
Leaning towards an opt-in rabbit hole, but not sure where to provide site owners control to toggle this functionality and have it set rabbit hole on the vocab

joewhitsitt:
sounds like eventual "double" validation as it relates to #4480

"tags are displayed on the following content types: article, ... please hide those before making this change"

"tags are hidden from visitors. please make these visible before displaying them on this content type"

PR #4858, opened 2022-02-24
Feature: tags in content types

### LINKED ISSUE #4480: As a site owner, I can show tags and related content at the bottom of every article
# Story

As a site owner, I can show tags and related content at the bottom of every article content type to help direct readers to related content. 

# Proposed solution

Screen Shot 2021-11-22 at 10 00 54 AM

@dnkemper brought this up recently, and if we used the view created for the `topic_page` feature we could display tags and links to other articles for each page. 

- Utilize `block_2` from https://github.com/uiowa/uiowa/blob/main/config/features/topic_page/views.view.topic_page_browse_by_tag.yml.

bspeare:
# Summary 
Resolves https://github.com/uiowa/uiowa/issues/4480.

(this is @quamsta, adding to this PR a bit for clarity)

This PR allows both tags and related content to be shown as a view block in the pre-footer region in the following content types:

- Page
- Article
- Person

Other content types currently do not have the ability to display tags and related content. That may have to be done during a different slice.

Screenshots of the view block in action (**note: tag button styles may not be finalized!**):

- [Screenshot of tags and related content on a page](https://user-images.githubusercontent.com/472923/156051972-52eb05eb-3ebe-47ad-848f-c9b36722a1fb.png)
- [Screenshot of only tags on a page](https://user-images.githubusercontent.com/472923/156051982-aa656cd4-8322-43ee-84ae-b2fa0589dd37.png)
- [Screenshot of only related content on a page](https://user-images.githubusercontent.com/472923/156051991-639eacae-682a-4f52-8100-e845523a3258.png)

Each content type is configurable via their respective settings pages in 

-  `admin/config/sitenow/uiowa-pages` ([Screenshot](https://user-images.githubusercontent.com/472923/156046333-1393ec87-fe1b-4973-9bd2-248164f7d40c.png))
- `admin/config/sitenow/uiowa-articles`([Screenshot](https://user-images.githubusercontent.com/472923/156046344-9bfe460f-820f-4c17-baf3-cb3e7f9e9c71.png))
- `admin/config/sitenow/uiowa-people`([Screenshot](https://user-images.githubusercontent.com/472923/156046357-baf1f1df-1171-4a30-8d03-53288669498d.png))

# How to test

1.  Ensure you've tagged some pages, articles, or people
2. Pull down this branch
3. `drush @default.local csim`
4. Enable the desired tag display via the config options in SiteNow Pages, Articles, and/or People (see summary above)
5. Visit the pages you tagged in step 1 to see those tags.

# To do: 
- [x] Exclude currently visited page from related content
- [x] Resolve issue with uiowa_people.settings being reverted during a `drush csim`
- [x] Tag style finalization
- [x] Other things @bspeare had in mind for this PR 
- [x] Decouple tags and related content functionality


bspeare:
@quamsta some theming thoughts:

- Can we reuse any of the styles we added for https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/studentlife_topics/sass/topic-collection.scss?  If so, maybe pull those styles out into a more general `tags.scss` file? 
- We have some existing tag styles on https://admissions.uiowa.edu/colleges/college-liberal-arts-sciences. Do we want to go with that type of styling?  The code for it is at https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/views/view-taxonomy-term-colleges.scss.  Or maybe use that as an example for implementing something like https://user-images.githubusercontent.com/2752004/153955757-ff2f79b8-5836-4372-928c-db66faa19bd6.png.

quamsta:
@bspeare

Just pushed up some initial styling options, but this is the general idea I have so far:

Screen Shot 2022-02-25 at 4 47 44 PM

One major change I'm considering illustrated in the screenshot above is the ability to show **both** tag buttons and related content. In the previous mockups in #4480, I only allowed either one or the other.

Showing both is a notion that I had while trying to make the block displays switch back and forth, but may need to be explored further.

I have a couple more ideas that I'm implementing, and willing to hear feedback and ideas. SCSS is a bit messy and I want to further "componentize" it and make some of the classes reusable on Topic Collection and Areas of Study. But this should get us started?

Thanks for your help with this @bspeare @joewhitsitt !

quamsta:
Updates: 

- Added config_ignore option for new sitenow_people.settings
- Excluded node you're currently on from the "Related" view

docroot/modules/custom/sitenow_pages/src/Form/SettingsForm.php
```diff
@@ -129,6 +130,25 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+          ->t('Do not display tags'),
+        'tags' => $this
+          ->t('Display tag buttons'),
+        'related' => $this
+          ->t('Display related content'),
+        'tags_and_related' => $this
+          ->t('Display tag buttons and related content'),
+      ],
+      '#default_value' => $tag_display_type ?: 'do_not_display',
+    ];
```
pyrello: This doesn't have to happen in this PR, but I think this should be split out into two separate checkboxes. This is based on the idea that we are going to probably introduce other settings related to how each of these two separate features work and the checkboxes could serve as the basis for showing additional settings that are hidden otherwise.

joewhitsitt: hey that was my idea from this AM's check-in. great minds...

quamsta: I have some ideas about this and I'm gonna record them here... soon..

quamsta: "soon" is probably going to be tomorrow. Some unplanned stuff fell in my lap during planning, ironically.

quamsta: After thinking about this a bit more, I'm going to dedicate some time to decoupling the tags and related functionality (currently "Tags List" and "Related" are two view modes on the same view, which may not be ideal) in addition to decoupling the settings fields (into checkboxes as suggested). 

I believe decoupling these two pieces of functionality will allow for easier later expansion on both sets of features. This could allow for separate view modes for related content (cards, lists of content, etc) without being intrinsically tied to tag buttons.

I think this is a better way to approach these two separate, yet "related" features and I want to make sure we reduce complexity in future updates (e.g., reducing the need for update hooks to rearrange settings) by putting our best foot forward and not pushing up something just to get it up there.

quamsta: I'm still not 100% sold on checkboxes, but I do think the functionality needs to be separated to allow for more options down the line. @joewhitsitt, @pyrello, what do you think of two dropdowns that provide more options as they become available in the platform?

Mockup with a fictional-yet-possible alternate view for related content:

Screen Shot 2022-03-04 at 1 17 38 PM

Screen Shot 2022-03-04 at 1 22 08 PM

joewhitsitt: So, I think when it comes to "related content," yeah it makes sense to split this off. If you are saying that the headings/bulleted list (grouped by tag) option is no longer to be associated with tags and instead act as an either/or to content as cards with images, then I am fine with this change. Keep in mind that splitting off the "public display of tag information" into all of these buckets makes the future of hiding tags site-wide more complicated: https://github.com/uiowa/uiowa/issues/3420

quamsta: @joewhitsitt Thanks for bringing up the hiding tags option. I'd heard about "rabbit hole" here-and-there from dropping into meetings but didn't have a sense of what it was 'til now.

That's definitely something I'm going to keep in mind going forward. I think we can work together on making sure a future option to disable the taxonomy term views doesn't conflict with the options presented in this PR. I'm more than willing to make adjustments to the views and the logic associated with the settings form if someone has the taxonomy term view disabled.

pyrello: The reason I suggested checkboxes is because each checkbox can serve as the trigger to show more options related to each feature. If the box is checked, those other options display. If not, they don't.

I am imagining something like:
```
[ ] Display tags

[ ] Display related content
```

If the user checks the first box, it would look more like:
```
[x] Display tags

Display tags as
---
Cards with images
Headings with bulleted lists
---
(^ this is a select list)

[ ] Display related content
```

bspeare:
One thing that we didn't document here is the style change for the tags.  We started with the "pill" tag design like the screenshot below:

Screen Shot 2022-03-07 at 4 09 45 PM

And we moved to the gray background style because it is a little easier to read.   The pill design is a bit busier because of the border.  Also, the line thickness is similar to the font weight and the two blend together visually. 

Screen Shot 2022-03-07 at 4 10 38 PM

docroot/modules/custom/sitenow_pages/src/Form/SettingsForm.php
quamsta: @pyrello, sorry I didn't see this 'til now. Was trying to truck along with this PR a bit and may have resolved this conversation a bit too quickly. 

I get the spirit of this and I agree 100%. Although I think I prefer to use two dropdowns instead of two sets of fields (a checkbox set and a conditional select list) per option.

I think this is a better user experience because the user gets all of the display options laid out in front of them upon clicking one of the dropdowns as opposed to being required to make a decision first ("check 'Display tags'" -> Get more options) and you can only see those additional options after you've made a decision.

GaryRidgway (approved):
After some trial and error getting everything set up, this all looks great!

Also tested with global pre-footers and it works with them as well, being placed before the pre-footer.

Well done all! Very cool.

PR #4867, opened 2022-02-28
Reverse column order for 2, 3, 4 and Page w/sidebar layouts in mobile

### LINKED ISSUE #2737: As a site owner, I can adjust the layout column order for mobile through a layout builder style
# Story

As a site owner, I can adjust the layout column order for mobile so that I can adjust the order/display on how the content stacks for mobile.  

# Proposed solution

- Create a layout builder style for sections to `Reverse column order on mobile` which would apply `flex-direction: row-reverse;` to the container.

### LINKED ISSUE #4782: Switch column order for Admissions events in mobile
# Story

As an Admissions site owner, I would like to see event registration information display higher in mobile order on events to help get more registrations.

# Proposed solution

- Create a layout builder style called something like "Reverse column order for mobile" which would which would apply class like `.section-order__reversed`.
- In `global.scss` create the class and style that will be applied in the layout builder style like:

```
.section-order__reversed {
  // two column
  &.layout--twocol {
    .layout__spacing_container {
      display: grid;
      @include breakpoint(sm) {
        display: inherit;
      }

      .layout__region--first {
        order: 2;
        @include breakpoint(sm) {
          order: inherit;
        }
      }

      .layout__region--second {
        order: 1;
        @include breakpoint(sm) {
          order: inherit;
        }
      }
    }
  }
}
```
- Apply layout builder style to the event content type config here https://github.com/uiowa/uiowa/blob/main/config/features/event/core.entity_view_display.node.event.default.yml#L81.

dnkemper:
Closes #4782 
Closes #2737
# How to test

1. `blt ds --site=admissions.uiowa.edu`
2. `blt frontend`
3. `drush @admissions.local cr`
4. `drush @admissions.local uli`
5. Go to https://admissions.uiowa.ddev.site/event/next-step-iowa-admitted-student-days-0 and minimize window to mobile. Column on R side should display above column on L side. 
6. When you go to add a multicolumn layout (2, 3, 4, page w/sidebar), you should have option to select the reverse mobile display and it should work accordingly 
7. `blt ds --site=sandbox.uiowa.edu`
8. `drush @sandbox.local uli`
9. go to https://sandbox.uiowa.ddev.site/reverse-column-order-mobile and add the 'Reverse column order on mobile' style to the section
10. minimize the window to a mobile breakpoint - make sure highest column number is at top of section


docroot/themes/custom/uids_base/scss/global.scss
```diff
@@ -316,6 +316,7 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled
 
       .layout__region--second {
         order: 1;
+        margin-bottom: 2rem;
```
joewhitsitt: Should `$mobile-height-gutter` be used instead? Paging @bspeare

bspeare:
@makurj I think you want to add the 'block_margin_bottom' layout builder style to the layout builder template for events instead of adding CSS.  This is probably an issue for all of our layouts and we might want to look at making a global change here.

bspeare:
Should we be adding the layout builder style globally? #2737 would add this feature for the platform.

makurj:
Adding `block_margin_bottom` here /admin/config/content/layout_builder_style/section_order_reversed/edit?destination=/admin/config/content/layout_builder_style didn't fix it. It's back to not having enough spacing.

dnkemper:
@bspeare crap.. did I misunderstand the proposed solution steps? Cooouuulllddd we just do the whole shebang? Nothing left in planned for the sprint 😇

bspeare:
No, I'm just wondering if the style should live in the default config if we plan on implementing this for the service in the future?

dnkemper:
LET'S DO IT

makurj:
So what am I changing?

bspeare:
> Adding `block_margin_bottom` here /admin/config/content/layout_builder_style/section_order_reversed/edit?destination=/admin/config/content/layout_builder_style didn't fix it. It's back to not having enough spacing.

Maybe go with the original CSS solution for now.

docroot/themes/custom/uids_base/scss/global.scss
bspeare: Yes, let's use that.

makurj:
The config now resides in both events and default?

dnkemper:
haha you must have added that comment seconds before I pushed up commit?

makurj:
Pulled the latest branch down. The spacing still seems to be working okay.

bspeare:
@makurj @dnkemper I moved the style back into the events feature until https://github.com/uiowa/uiowa/issues/2737 happens.   If we put it in the default config, the style option would appear for every two column layout.  Right now it will only appear for every two column layout within the admissions site.

dnkemper:
Alright! We were going to talk about this in parking lot and I def misunderstood https://github.com/uiowa/uiowa/pull/4867#issuecomment-1055696254. Super pumped!

bspeare:
@dnkemper No, I misunderstood when I made the comment.  I thought there might be a way to implement it without it being widely available.    Keeping this scoped to admissions seems like a good initial slice.

dnkemper:
I think that is an excellent idea! 🥳
From: bspeare ***@***.***>
Sent: Wednesday, March 2, 2022 8:44:58 AM
To: uiowa/uiowa ***@***.***>
Cc: Wisnousky, Danielle N ***@***.***>; Mention ***@***.***>
Subject: [External] Re: [uiowa/uiowa] Started adding the section-order__reversed style (PR #4867)


@dnkemper<https://github.com/dnkemper> No, I misunderstood when I made the comment. I thought there might be a way to implement it without it being widely available. Keeping this scoped to admissions seems like a good initial slice.

bspeare:
I think we need to update the event config since https://github.com/uiowa/uiowa/pull/4881 was merged in?

bspeare:
One thing to note on this is that if a site has the "Event" split enabled outside of admissions, they will see the option to "Reverse column order" on all of their two-column layout sections within layout builder.

pyrello:
> One thing to note on this is that if a site has the "Event" split enabled outside of admissions, they will see the option to "Reverse column order" on all of their two-column layout sections within layout builder.

If we are going to limit this to just Admissions, we should move the layout builder style to that site split rather than the event feature split. Probably there aren't any other sites using the event split, but adding the style to the event feature split doesn't really make any sense to me.

Also, can we change the title of the PR to reflect the actual scope of work?

dnkemper:
But Admissions is the only site to have this feature enabled? Thought I'd get to work on that 'support issue' that follows this and moves it to default... where all two columns will be able to use this style. Also thought was going to do both issues in this PR - apologies for the title.

dnkemper:
THANK YOU, @bspeare

joewhitsitt (approved):
Works for me. Nice feature. Maybe down the road "page with sidebar" could default with it so that menus are at the bottom instead of overpowering the top.

bspeare:
Thanks @joewhitsitt! I added a final commit to make sure the mobile view works if the `grid--1-3` classes are used on a four column grid.

PR #4881, opened 2022-03-01
Admissions event display permanent fix

pyrello:
Fixes an issue where events on the Admissions site were displaying their entire LB template twice. I already fixed it in prod, so this just makes the fix permanent.

# How to test

Code review is fine.


bspeare (approved):
No more duplicates!

PR #4907, opened 2022-03-07
Tags added within Relationship text input are hidden from view

### LINKED ISSUE #4042: Tags added within Relationship text input are hidden from view
# Problem
When you have multiple tags added within a Relationship they end up being hidden if they are longer then the width of the text input field. It is hard to even add a new tag because you have to scroll all the way over to the right which is then covered up by the search icon.

Screen Shot 2021-08-30 at 12 52 27 PM

# Proposed solution
Use a textarea where tags wrap to a new line if it is too long.

dnkemper:
When you have multiple tags added within a Relationship they end up being hidden if they are longer then the width of the text input field. It is hard to even add a new tag because you have to scroll all the way over to the right which is then covered up by the search icon.

Closes #4042 

# How to test

1. blt ds --site=sandbox.uiowa.edu
2. add article, page and people tags - should now see the default tag behavior (has handles - 'add another' button etc) like https://user-images.githubusercontent.com/4663676/131385625-c265b2eb-8474-4482-aa74-0c0932602770.png


joewhitsitt (approved):
Works as expected.

I can create a follow up issue for event tags and consider the same for the media types. I think Grad has one site that might want to adopt this widget type for consistency.

dnkemper:
Good thinking. Thank you, @joewhitsitt

PR #4931, opened 2022-03-10
Hotfix for BLT inspector change and refactor UME to only notify on error.

### LINKED ISSUE #4929: Too many Slack messages for AC cron start/stops
### Problem
Now that cron is running is AC prod via `blt ume`, we get notifications every time it starts/stops. This is multiple times per hour. The messages clutter the drupal-status channel and have lost their value.

### Proposed resolution
Change `blt ume` from notifying start/stop to only if a site encounters an error running the specified command.

### LINKED ISSUE #4930: BLT Detector::isDrupalInstalled is now misleading
### Problem
A change in BLT's Detector::isDrupalInstalled will now return FALSE if the site is installed but actually just throwing an error preventing it from bootstrapping. This problematic for a few reasons including:
- our install command could then return that site to be re-installed. 
- our override of `blt auda`, which runs on AC code deployments, skips sites that are not installed to reduce running time avoid reporting unnecessary errors. This change prevents the `drupal:update` process from running on an error-ing site which we utilize to `cr` in an `updb`.

### Proposed resolution
Temporarily use our own isDrupalInstalled method using the previous code in Detector::isDrupalInstalled.

richardbporter:
Resolves #4930 
Resolves #4929 

# How to test

On main, sync these sites:
```
multisites:
  - default
  - admissions.uiowa.edu
  - brand.uiowa.edu
```

Drop admissions, i.e. replicate a uninstalled site:
`ddev drush @admissions.local sql:drop`

Create an error in brand - add a bogus `$foo` argument to `brand.uiowa.edu/modules/brand_core/src/Commands/BrandCoreCommands` constructor:
```
  public function __construct(AccountSwitcherInterface $accountSwitcher, DateFormatterInterface $dateFormatter, MailManagerInterface $mailManager, ConfigFactoryInterface $configFactory, $foo) {
...
```

Run `ddev blt umi` and see that `brand.uiowa.edu` is reported as being uninstalled when actually an error is preventing BLT inspector from running Drush status.

Check out this branch and run `ddev blt umi` again. See that brand.uiowa.edu is not reported since our method uses a SQL query.

Checkout main again and run `ddev blt auda -v`. See that the `drupal:update` process is only triggered on the default site even though `brand.uiowa.edu` is installed. Check out this branch and re-run. The `drupal:update` process should be triggered on default and brand.uiowa.edu, not admissions, and the error regading `$foo` is thrown. We want this to happen so sites throwing errors in AC get an update hook to run.

#### Cron
From the previous steps, i.e. on this branch with the `$foo` error:

- DM me for a Slack webhook URL.
- ddev ssh
- export SLACK_WEBHOOK_URL=xxx
- blt ume status

See that no start/stop messages are posted to Slack but rather an error message is posted for brand site. Fix `$foo` error and re-run in same terminal session. See that no Slack messages are posted.

## Post-deploy tasks
- [x] Update [AC variable](https://docs.acquia.com/cloud-platform/manage/variables/) to new webhook URL

### DIFF
--- a/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
+++ b/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
@@ -21,12 +21,15 @@ use GuzzleHttp\Exception\ClientException;
 use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Yaml\Yaml;
+use Uiowa\InspectorTrait;
 use Uiowa\Multisite;
 
 /**
  * Global multisite commands.
  */
 class MultisiteCommands extends BltTasks {
+
+  use InspectorTrait;
 
   /**
    * A no-op command.
@@ -161,7 +164,7 @@ class MultisiteCommands extends BltTasks {
         continue;
       }
 
-      if (!$this->getInspector()->isDrupalInstalled()) {
+      if (!$this->isDrupalInstalled($multisite)) {
         $uninstalled[] = $multisite;
       }
     }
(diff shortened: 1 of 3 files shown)


joewhitsitt:
Testing

joewhitsitt (approved):
Test steps reproduced. I find it amazing that you are able to find these issues. Nice

joewhitsitt:
Wonderful instructions btw. Helped me understand what I was doing more.

PR #4934, opened 2022-03-10
Implement hook_cron in uiowa_core for hourly tag invalidation.

### LINKED ISSUE #4842: Create hook_cron to invalidate and purge custom cache tag used in sitenow_events
Blocked by #4841.

### Proposed solution
Rename sitenow_events cache tag created in #4818 from `time:hours_3` to `time:hourly`. 

Implement `hook_cron()` in uiowa_core.module. Check for existence of of [State API](https://www.drupal.org/docs/8/api/state-api/overview) key called `uiowa_core_hourly_last` or similar. If set, subtract the value from the request time. If that is greater or equal to an hour (3600 seconds), then invalidate the tag `time:hourly` via `Cache::invalidateTags();` and queue a purge for the tag as documented in the [purge README](https://git.drupalcode.org/project/purge#queueing). Set the State API key `uiowa_core_hourly_last` to the current request time.

#### Testing
This can be tested similarly to #4818 but instead of running the commands, just wait for cron to run assuming it has been set up on dev in #4841.

##### Story
As an anonymous visitor to a SiteNow website, I can see updates to events after one hour.

richardbporter:
Resolves #4842

# How to test

The default dev site sitenow_events.settings have been changed to use the content.dev.drupal.uiowa.edu API. Check with `drush @default.dev cget sitenow_events.settings`.

- Open https://default.dev.drupal.uiowa.edu/about in an anonymous session to cache a response in Varnish.
- Edit https://content.dev.drupal.uiowa.edu/test-4 like the title or something.
- Wait an hour for cron to run. Cron is running on the hour for uiowa.dev
- See if the https://default.dev.drupal.uiowa.edu/about page refreshes.


richardbporter:
[74b3260](https://github.com/uiowa/uiowa/pull/4934/commits/74b32600dffa720a347ba07289fbeb52cb191ad8) is a bad commit message - max-age is in seconds (60), not minutes.

richardbporter:
This isn't working in dev for me. I don't see the change until I refresh the page in an authenticated session. Then after that, the Varnish page cache seems to invalidate and the change is visible.

richardbporter:
> This isn't working in dev for me. I don't see the change until I refresh the page in an authenticated session. Then after that, the Varnish page cache seems to invalidate and the change is visible.

I'd be curious if any interaction with the site might do it. The queue isn't processed right when cron runs. Maybe we could run cron more often or try [bypassing the queue](https://git.drupalcode.org/project/purge#invalidation-without-queue). Docs say not recommended though.

richardbporter:
> > This isn't working in dev for me. I don't see the change until I refresh the page in an authenticated session. Then after that, the Varnish page cache seems to invalidate and the change is visible.
> 
> I'd be curious if any interaction with the site might do it. The queue isn't processed right when cron runs. Maybe we could run cron more often or try [bypassing the queue](https://git.drupalcode.org/project/purge#invalidation-without-queue). Docs say not recommended though.

Actually, scratch that. We have purge_processor_cron enabled so I would expect the queue to be processed. I would recommend trying to hook_module_implements_alter uiowa_core hook_cron() to run before purge_processor_cron hook_cron(). I'm surprised that doesn't already do that.

joewhitsitt:
testing/waiting for cron to run since deploying that implements_alter change

joewhitsitt:
Still not sure if the order of operations is working but I decided to give cron a bit more time to register/run against the time check. I checked the last run timestamp to see that it last ran at 4:00:23 which if the next run is any more precise will mean that the 3600 second check will fail and it will wait till the next run. Travis failed to get the change [ee3f4ed](https://github.com/uiowa/uiowa/pull/4934/commits/ee3f4ed926f9b4447aa5b54f77a80422dbffc6f5) out to DEV on time. I will revisit tomorrow

joewhitsitt:
Poking more at `purge_processor_cron` today by trying to enable it locally enough to see the hook_cron order and queue. Works at hour 2. So as-is max cache for events would be two hours and 59 minutes.

joewhitsitt:
This is working as expected now (at least for me). Looking for another tester

joewhitsitt:
To make testing easier for folks, I cached https://default.dev.drupal.uiowa.edu/about this morning and then made an edit to https://content.dev.drupal.uiowa.edu/node/82386 that changes the title to 8:38am 3/17/22. Shortly after 9 AM a refresh of https://default.dev.drupal.uiowa.edu/about should reload the block with the updated title. I might hijack standup...

makurj (approved):
https://default.dev.drupal.uiowa.edu/about displaying changes in incognito at 10 for a 9:45 AM change.

PR #4950, opened 2022-03-11
SPPA - Removed unnecessary LB override from page display.

pyrello:
It looks like this change got added during the update to 9.3.x, but it is not actually necessary.

# How to test

Code review is sufficient.


joewhitsitt (approved):
Reviewed, not tested

PR #4951, opened 2022-03-11
HOTFIX - fixed existing blocks with grid settings

### LINKED ISSUE #4949: Some events blocks are requiring a grid selection due to the recent layout builder styles update
# Problem

On https://education.uiowa.edu/student-experience/baker-teacher-leader-center/upcoming-btlc-events:

The layout builder group for grid columns is showing `-select-` on https://education.uiowa.edu/student-experience/baker-teacher-leader-center/upcoming-btlc-events:

Screen Shot 2022-03-11 at 9 23 26 AM

If you create a new events block the settings look like:

Screen Shot 2022-03-11 at 9 23 41 AM

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4949

# How to Test

- Sync a bunch of sites to test an update hook that runs on the following blocks:
  - 'inline_block:uiowa_articles'
  - 'inline_block:uiowa_events'
  - 'inline_block:uiowa_people'
  - 'inline_block:uiowa_aggregator'
  - 'views_block:article_list_block-list_article'
  - 'views_block:people_list_block-list_card'
- There should be no update errors.
- Additionally you should be able to go into an existing block that is displaying as a list, change it to grid and see that the 3-column option is set. Switch back to list and save the block. The markup should not include the 3-column class.
- On a new block with grid display option, see that the 3-column option is set and that there is not an empty option.
- You should be able to save and re-save blocks and get the expected display settings regardless of what you choose.


bspeare:
I can test this.

bspeare:
`The markup should not include the 3-column class.`

Is this new?  I think the 3-column `grid--threecol--33-34-33` class has always been appearing on blocks no matter the selection.

joewhitsitt:
Basically, we moved where it was getting unset

We had submit code that was unsetting it:
```
// Unset grid options if list format is selected.
  $display_option = $form_state->getValue('layout_builder_style_list_format');
  if (isset($display_option) && $display_option == 'list_format_list') {
    $form_state->setValue('layout_builder_style_grid_columns', '');
  }
```

To keep that but get rid of the empty option and keep it required, we wanted to store the value, but unset before render:
```
// Unset grid_columns classes if display as list.
  if (isset($variables['elements']['#attributes']['class'])) {
    $is_list = array_search('inner--list', $variables['elements']['#attributes']['class']);
    if ($is_list) {
      foreach ($variables['elements']['#attributes']['class'] as $key => $style) {
        if (str_starts_with($style, 'grid--')) {
          unset($variables['elements']['#attributes']['class'][$key]);
        }
      }
    }
  }
```

bspeare:
OK, I'm still seeing `views-element-container inner--list grid--threecol--33-34-33 contextual-region block block-views block-views-blockarticle-list-block-list-article` as the classes on a block set to be a list.   Is that correct?

joewhitsitt:
Shouldn't be... After save? Taking a look.

joewhitsitt:
I wonder if this is listblock specific. Doesn't seem to be affecting it, but probably something we want to clean up for consistency.

bspeare:
Yes, I added CSS a while ago to make the blocks work with the class.  Should I make a follow-up issue for this?

joewhitsitt:
If this PR fixes the immediate problem, I'd say a follow-up issue would be good.

bspeare (approved):
Tested on education, admissions, and sandbox and it fixed the issues I noticed on the prod sites!

joewhitsitt:
Thanks for the quick review!

PR #4956, opened 2022-03-11
Feature: horizontal nav dropdowns

### LINKED ISSUE #4823: As a site owner, I can use a "Horizontal" version of the menu block that displays the menu depth in dropdown menus
# Story

As a site owner, I can use a "Horizontal" version of the menu block that displays the menu depth in dropdown menus

# Proposed solution

View my latest comment below for progress notes.

### LINKED ISSUE #5061: As a developer, I can integrate the `accessible-menu` library into uids_base to use with our secondary menus
# Story

As a developer, I can integrate the `accessible-menu` library into uids_base to use with our secondary menus

In order to unblock https://github.com/uiowa/uiowa/issues/4823, this work needs to be done. 

# Proposed solution

- Work with https://github.com/NickDJM/accessible-menu/pull/168 that is in https://github.com/uiowa/uiowa/pull/4956. 
  - There are a series of checkboxes on the `accessible-menu` pull request documenting the work to be finished.

bspeare:
Resolves #4823. 
Resolves #5061.

# To test
1. Setup:  
    a. `composer install`
    b. `blt frontend`
    c. Test on Tippie and Sandbox `blt ds --site=tippie.uiowa.edu` and `blt ds --site=sandbox.uiowa.edu --define drush.aliases.remote=sandbox.dev`
    d. `drush @tippie.local uli iowa-mba/curriculum-iowa-mba` and `drush @sandbox.local uli media`
    e. Go to the above pages to test out dropdowns. 
2. Tab through until you get to the menu (Shift+tab traverses backwards). 
    a.  If you hare focusing a menu dropdown indicator, space should toggle it open and closed.
    b.  If you hare focusing a menu dropdown indicator, enter should toggle it open and closed.
    c.  If a menu is open, the escape key should close it.
    d. All of the above should also work with nested menus.
3.  Test that the above works on mobile, and that the mobile version generally feels like it functions normally.
5. Test that alternative background colors still have reasonable button contrast.
6. Test that having that having the initial visibility level follow the active menu items, and both of the options therein, do not break functionality.
7. Test that stuffed menus work just fine. Fill them up as much as you can!
8. Examine `aria` 
9. test that <Nolink> links do not function oddly. 

<details>

# todo
- [x] Attach js based on `horizontal` selection possibly using https://git.drupalcode.org/project/layout_builder_styles/-/blob/8.x-1.x/src/EventSubscriber/BlockComponentRenderArraySubscriber.php#L96 as a way to do it. 
- [x] Add toggles to menu markup.
- [x] Add mobile menu toggle `<button id="main-menu-toggle" aria-label="Toggle main menu">Menu ☰</button>` (https://github.com/NickDJM/accessible-menu/blob/3.x/docs/basics/two-level-menu-with-toggle.md)
- [x] Deploy to dev and share with `NickDJM` for feedback on third-level
- [x] Browser test
- [x] Test keyboard accesssibility

**Note:** I moved our initial prototype to Codepen for reference: https://codepen.io/bspeare/pen/ZEvymNR. 

### Pull in feature branch
This needs to be run the first time you work on it or anytime you run `composer install`:
* `cd docroot/libraries/accessible-menu && git checkout top-link-disclosure-menu`
* If you make a change to the `src` JS files, do the following:
* `npm install`
* `npm run build`

Docs: https://accessible-menu.netlify.app/

</details>


richardbporter:
Added `composer install` to setup in PR description.

pyrello:
A few updates:
* We Adapted the existing implementation script to add the usual wrapper to use `Drupal.behaviors`.
* The library is only being attached when a horizontal menu is present.
* We figured out steps (added to the IS) to checkout the `accessible-menu` feature branch for local development and contribution back to the project.
* We ran into an issue where we were getting console errors that we traced back to our `<nolink>` functionality. This was resolved by updating our implementation code (RTD!) via f943ba3.
* It turns out that the library does not handle adding toggle links and those need to be present to be acted upon. @bspeare and I decided to add via JS so that they are not visible and non-functional if the JS breaks for some reason: 6a492cc

bspeare:
I added a few more items to the todo list.

bspeare:
After testing the keybinding and hover options for the `accessible-menu`, here are a few recommendations:

- Left to Right keyboard arrows don't even work with Superfish, so we could set `optionalKeySupport` to `false` until that is fixed and deliver it as enhanced accessibility when it is finished. 
- I don't think we need to make `hover` work for the initial release.  I think that would be a nice feature enhancement to turn on when it is ready, so `hoverType` should be set to `off`. 
- That leaves us with enter/space toggle.   It currently opens the menu but will not toggle it close.  This item should be prioritized before we release this functionality. 

@pyrello what do you think?

pyrello:
Added a comment on the PR: https://github.com/NickDJM/accessible-menu/pull/168#issuecomment-1088872943

GaryRidgway:
As I have researched this I have found that the escape key is the intended way to exit the second level menus. Additionally, I have found multiple sources saying that this is the correct way to implement accessible menu navigation. Here are a few sources:
https://www.evinced.com/blog/a11y-nav-menus/
https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/#keyboard-and-focus-behavior

So it actually seems like the menu is functioning correctly in accordance to accessibility Guidelines. My initial impression is that it is supposed to work this way to keep interaction consistent across menus, modals, etc.

GaryRidgway:
I am noticing some contrast issues on black backgrounds
Screen Shot 2022-04-13 at 12 07 31 PM


Additionally, the `<nolink>` functionality is a little wonky right now. If you tab backwards to a `<nolink>`, it opens the menu:
0c364709b84a53bc7c81f7accb08b69f

However if you tab forwards through another link to a `<nolink>`, it opens the link you passed through:
d79eafbc9137a007a73396702f98303a

bspeare:
It looks our blocking issue is not with the `accessible-menu` library but with these lines of code https://github.com/uiowa/uiowa/blob/d06878a3062264b55d9fd8fa2490af780c991a80/docroot/profiles/custom/sitenow/sitenow.profile#L913-L916 which  is adding `tabindex` to all `<span>` tags because we need to be able to tab through superfish dropdown menus that use `<nolink>`.

richardbporter:
> It looks our blocking issue is not with the `accessible-menu` library but with these lines of code
> 
> https://github.com/uiowa/uiowa/blob/d06878a3062264b55d9fd8fa2490af780c991a80/docroot/profiles/custom/sitenow/sitenow.profile#L913-L916
> 
> which is adding `tabindex` to all `<span>` tags because we need to be able to tab through superfish dropdown menus that use `<nolink>`.

I'm poking around and it seems like this code could be moved to https://github.com/uiowa/uiowa/blob/2f6e9aa0a208b32eed0d93a5b2bd581d5ef03f2b/docroot/themes/custom/uids_base/uids_base.theme#L1044-L1056

richardbporter:
>Test that the above works on mobile, and that the mobile version generally feels like it functions normally

Is this in dev somewhere?

I'm not seeing that weird nolink behavior that @GaryRidgway indicated earlier. 👍🏻

richardbporter:
Random thoughts:

Maybe its just my monitor, but the little indicator set ::before a nested link is so faint I thought it was a spec of something on my screen. 😆 
Screen Shot 2022-04-19 at 3 39 28 PM

On mobile, the dropdown arrow does not seem to be vertically-aligned with the link text for nested dropdowns:
Screen Shot 2022-04-19 at 3 54 33 PM

Deploying to dev now...

richardbporter (approved):
I don't think https://github.com/uiowa/uiowa/pull/4956#issuecomment-1103172468 is a blocker. This seems to work really well and look good.

bspeare:
Thanks @richardbporter! I think that we should fix those points you surfaced.  I can make those updates later today.

bspeare:
@richardbporter I updated the pr to address https://github.com/uiowa/uiowa/pull/4956#issuecomment-1103172468 if you want to take another look.

richardbporter:
> @richardbporter I updated the pr to address [#4956 (comment)](https://github.com/uiowa/uiowa/pull/4956#issuecomment-1103172468) if you want to take another look.

Yep, looks good.

PR #4970, opened 2022-03-14
admissions - re-export with person status field placed

joewhitsitt:
Admissions never got the person status field and it is now having issues since the field_tags change on the form.

# To Test
Sync admissions.uiowa.edu, no errors


cory-skeers (approved):
Reviewed and tested, was able to get a clean sync with the update.

PR #4972, opened 2022-03-15
SPPA - make sure private field stays split

joewhitsitt:
Not sure how this wasn't captured in configuration before. But this locks in the private field to these default content types.

### DIFF
--- a/config/sites/sppa.uiowa.edu/core.entity_form_display.node.article.default.yml
+++ b/config/sites/sppa.uiowa.edu/core.entity_form_display.node.article.default.yml
@@ -1,0 +1,202 @@
+uuid: 61298539-c20b-451e-86eb-03384dbd98f9
+langcode: en
+status: true
+dependencies:
+  config:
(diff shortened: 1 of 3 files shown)


pyrello:
These forms should all be able to be added to the `partial_list`.

joewhitsitt:
@pyrello i can't get this to split partially. i've tried `core.entity_form_display.*` and specifying them directly. is it because the diff related to a module that is in the complete split?

pyrello:
I'm not sure. I'll give it a try in a little bit.

pyrello (approved):
Thanks for the fix!

pyrello:
> @pyrello i can't get this to split partially. i've tried `core.entity_form_display.*` and specifying them directly. is it because the diff related to a module that is in the complete split?

I guess this is a limitation of the distinction between `complete_list` and `partial_list` when there is a module involved. To me it seems like a bug with the logic in CS2.

PR #4985, opened 2022-03-16
adding pretty ids for accordions

GaryRidgway:
This is just an extension of https://github.com/uiowa/uids/pull/658 to make sure everything works in uids_base. please check to see if the things listed in that issue work the same here.


joewhitsitt:
Does {{ accordion_hash|clean_class }} reduce your special characters stuff? https://www.drupal.org/docs/theming-drupal/twig-in-drupal/filters-modifying-variables-in-twig-templates#clean_class

GaryRidgway:
@joewhitsitt I'm not seeing a way for it to work sadly https://twigfiddle.com/k40wrh

joewhitsitt:
Hmm it should, but probably not in that twigfiddle since it is a Drupal specific filter that leverages https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AgetClass/9. We use it quite a bit in the code base. https://github.com/uiowa/uiowa/search?l=Twig&q=clean_class

GaryRidgway:
clean_class worked very well locally! thanks @joewhitsitt

bspeare:
@GaryRidgway any concern about the underscore it adds for something like this?

Screen Shot 2022-03-18 at 3 19 37 PM

GaryRidgway:
well, i dint think it will cause a linking problem, but it is definitely not good looking. I think that we should go back to the custom filter I had before so that it doesn't strip everything so harshly. Dangit `clean_class`!

bspeare:
I'm fine with `clean_class`, this is probably not going to be a common issue?

bspeare:
I'm fine with `clean_class`, this is probably not going to be a common issue?

GaryRidgway:
No almost certainly not, and it doesnt change too much anyways. I will add a card to PL for monday to have some other opinions weigh in.

bspeare:
I tested it out and everything worked as expected.  The only weird thing as noted above is what it does to numbers and special characters.

bspeare:
I tested it out and everything worked as expected.  The only weird thing as noted above is what it does to numbers and special characters.

pyrello:
Discussion in standup: `clean_class` is good for now. We will revisit if needed based on customer feedback.

joewhitsitt (approved):

PR #4987, opened 2022-03-17
Uipress tabs

### LINKED ISSUE #4759: UI Press - Implement tabs on book page
Fields that will be individual tabs:

- Reviews/blurbs
- Contributors
- Table of contents
- Excerpt
- Awards

Fields are going to have to be placed in a view to allow placement as a single item in Layout Builder.

Tabs section will go before Related Reads in the book content layout. 

- Implement tabs css and js from: https://github.com/uiowa/uids/pull/637
- Create a tabs library in `uids_base` like: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.libraries.yml#L80-L84
- Create a tabs template and possibly use existing tab template: 
  - https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/paragraphs/paragraph--uiowa-collection-item--tab.html.twig

cory-skeers:
Resolves #4759 

~This work currently also undoes the removal of tabs from collections--they need to be finished and tested as well, or the line disabling them added back in.~ The code has been re-added.

# How to test

## Testing UI Press books
`blt frontend`
`git checkout uipress-tabs && git pull`
`drush @uipress.local uli`
Create new or visit existing book pages
Check out the tabs at the bottom of the book page layout.
Try combinations of empty and non-empty fields, see that things don't break

## Testing collections (not necessary to move this PR forward)
In docroot/modules/custom/layout_builder_custom/layout_builder_custom.module, removing lines 263-265 will enable the use of tabs in existing and new collections blocks. The templates and code are in place, but doing so has not been vetted.

Edit a page layout builder layout
Add a new collection, select tabs, save
Edit existing collection, change view type to tabs, save
Experiment--see if it works or breaks


cory-skeers:
This is mostly working, with a few updates yet to go:

There's a weird extra line that looks like it part of the sidebar accordion above (ebook, perpetual in this example):
image

Either a check for empty fields needs to be added to not include the field in the tabs if there is no content, or a content rewrite should be added to the view so there's not just empty space.

The default selected tab is hardcoded to use the reviews tab, and this behavior won't work if it gets updated to not include empty fields.

Otherwise, general testing.

cory-skeers:
I have also re-added the tab disabling snippet so that it won't get merged in by accident until it has been thoroughly tested. This PR was originally intended only to do the UI Press layout updates.

cory-skeers:
> There's a weird extra line that looks like it part of the sidebar accordion above (ebook, perpetual in this example):

Looks like this may be a collision with the tabs styling, and the button is getting both the intended accordion styling, as well as some of the tabs styling that is targeted to `role="tab"`.

image

bspeare:
@cory-skeers this commit https://github.com/uiowa/uids/pull/659/commits/0d57c683d0b94b2273757c8d7f4e6752e00d425b should fix the collision issue.

cory-skeers:
> this commit [uiowa/uids@0d57c68](https://github.com/uiowa/uids/commit/0d57c683d0b94b2273757c8d7f4e6752e00d425b) should fix the collision issue.

Working like a charm from what I've seen so far. Thanks!

cory-skeers:
@e-marie-w or @pyrello: Question about if there isn't content for one or more of the tabbed fields:
If for instance there isn't an Excerpt, should that tab not appear at all, or should it appear but display something like "No excerpt available." My inkling was the former.

Also, the tab headings I'm currently using are: "Reviews | Contributors | Table of contents | Excerpt | Awards", but they're set up in the View, so can be adjusted pretty easily (for instance, if there may be more than one Excerpt, might want to make that plural as well).

e-marie-w:
I know back in January the customer specifically asked if the tabs would appear when empty. Since we hadn't built them yet I said it was still being figured out. I took that conversation to mean ideally the tabs shouldn't display when empty. 

It's also what makes the most sense to me, but if it ends up being complicated I can ask if it's necessary or not.

cory-skeers:
Awesome--I'll move forward with them only appearing if there's content. It shouldn't be much more complicated. Thanks @e-marie-w!

pyrello:
I would concur that we don't want the tabs to appear if there is no content.

bspeare:
@cory-skeers I updated this with the latest release and removed that section in the "How to test" instructions.   The tabs look good!

e-marie-w:
I started testing this a little earlier and will continue now! So far it's working great, I just want to try some more combinations.

e-marie-w (approved):
Haven't tried every possible field combo, but tested combos of 1, 2, 3, 4, and 5 (all tabs) at least once and rotated which fields were included. They all worked as expected, on newly added and existing books. Adding or removing a field after creation also works fine. I tested a book with no tabs too, that also worked as expected.

I would wait on adding the tabs to collections, though. I tested it a little and ran into issues with adding a new collection of type tab. The Configure Block options aren't dismissed when I click Add Block. If I close out and Save Layout my new tabs will show up. If I don't Save Layout, the next time I go to Layout Builder I get a white screen with `Error: Call to a member function get() on null in uids_base_preprocess_block_content__uiowa_collection__tab()` Converting existing collections to tab worked fine, though, as long as I don't add another tab while doing so. Similarly, editing existing tabs works fine, it's just adding a new one that starts breaking stuff.

cory-skeers:
> Haven't tried every possible field combo, but tested combos of 1, 2, 3, 4, and 5 (all tabs) at least once and rotated which fields were included. They all worked as expected, on newly added and existing books. Adding or removing a field after creation also works fine. I tested a book with no tabs too, that also worked as expected.
> 
> I would wait on adding the tabs to collections, though. I tested it a little and ran into issues with adding a new collection of type tab. The Configure Block options aren't dismissed when I click Add Block. If I close out and Save Layout my new tabs will show up. If I don't Save Layout, the next time I go to Layout Builder I get a white screen with `Error: Call to a member function get() on null in uids_base_preprocess_block_content__uiowa_collection__tab()` Converting existing collections to tab worked fine, though, as long as I don't add another tab while doing so. Similarly, editing existing tabs works fine, it's just adding a new one that starts breaking stuff.

Thanks for all the testing! I've added your notes to the issue on enabling tabs so that we have it as a starting point there.

PR #4989, opened 2022-03-18
Remove banner block title requirement

### LINKED ISSUE #4750: As a site owner, I can choose to visually hide the banner title in the banner block
# Story

As a site owner, I can choose to visually hide the banner title in the banner block so that I can create content with a background image and text area only.   

Screen Shot 2022-02-03 at 1 47 48 PM

# Proposed solution
- [x] Make the banner title field not required
- [x] Add validation logic to provide a form error if neither the title nor link text are included.
- [ ] Use `#states` to set the required property on the link text field (if possible)

dnkemper:
Closes #4750 

# How to test

1. `blt ds --site=tippie.uiowa.edu`
2. `drush @tippie.local uli`

## Tests
1. Add a banner with a title and a link with text. No form error. Link shows as a button and is linked.
2. Add a banner with a title and a link without text. No form error. Link shows as a pseudo-button and title is linked.
3. Add a banner without a title and a link with text. No form error. Link shows as a button and is linked.
4. Add a banner without a title and a link without text. Form error displays with an error message below the link title field.


dnkemper:
Noticed that link doesn't appear without title

pyrello:
> Noticed that link doesn't appear without title

Probably we need to update the logic in the generic banner template. It looks like we currently aren't handling the case where a link exists but there is no banner title:
https://github.com/uiowa/uiowa/blob/2fa9185d4911b566595ed9966e1e398a3bcfaa02/docroot/themes/custom/uids_base/templates/uids/banner.html.twig#L9-L16

bspeare:
@pyrello do we wrap the entire banner in a link for that case?

pyrello:
> @pyrello do we wrap the entire banner in a link for that case?

I would have to revisit the logic a bit more closely to understand it better. However, I think that would be consistent with what we are doing with cards. When we worked on the updates to the banner, I seem to recall thinking that this segment of code didn't seem quite right.

GaryRidgway:
i can pull and test this

GaryRidgway (approved):
Looks great to me! Testing instructions evaluated true.

pyrello:
Thanks for testing, @GaryRidgway!

PR #5008, opened 2022-03-23
Admissions - Community Colleges

### LINKED ISSUE #4976: Create community colleges taxonomy and view block
- Term Name - this will be the official name of the Community College
- Locations - Not sure what we want to call this but my understanding is the District name is likely the official name but we will want a place to store/list specific CC's in that district (e.g. Eastern Iowa Community College District would be official name but Clinton Community College, Muscatine Community College and Scott Community College would be colleges in that district). We'll likely want to output these locations in lists since students will identify that easier. Plan info would be the same for all locations in that district.
- Contact Info - text area since some are links and others are individuals contact info

Need to reference what Iowa majors are available for each CC. Not sure how the team wants to architect that reference or where.

Here is a link provided by Admissions to some info about the fields needed here https://iowa-my.sharepoint.com/:w:/g/personal/katmills_uiowa_edu/EaQbdfwQQIJDgKED1dvJ7jQB143-R7ErxZPSZ8B019J6gA?e=j1aece. Provided 3/30/22

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4976

## To Test

- `ddev composer install && ddev blt frontend && ddev blt sync --site admissions.uiowa.edu --define drush.aliases.remote=admissions.test`
- There is a new community colleges taxonomy that webmasters should have access to add/edit/delete terms within. The term is public to visitors.
- On that taxonomy should be a description field (intro), contact information (long text), locations (multi text), "unsure major" (long text with default value).
- On render the contact information field label should prepend the term name.
- List of majors will be handled in a follow-up.
- Webmasters should have the ability to place the community colleges block in layout builder which displays the colleges and the locations in the list.
- Looks like Percy is struggling with lazy loaded images again...

Screen Shot 2022-04-01 at 1 37 04 PM
Screen Shot 2022-04-04 at 8 58 23 AM


bspeare:
Looks like Percy is having trouble loading the images for the baseline again.  I'll approve the Percy build and see if the baseline continues to change.

bspeare:
Just read your comment @joewhitsitt

joewhitsitt:
I added a note to parking lot. We might have to do something different based on what I read here https://docs.percy.io/docs/capturing-lazy-loading-images

e-marie-w:
I can test this!

e-marie-w:
On the taxonomy page where you can add new terms or reorder existing ones, should 'Reset to alphabetical' be hidden from non-admins? When masquerading as a webmaster I get a 'You are not authorized to access this page' message. It's not just the community colleges, though, it happens on the other taxonomies too. 

Also, for the Community Colleges block, should we set a default number for 'Items to display'? It doesn't throw an error if I don't enter a number, though, is a default already set somewhere?

joewhitsitt:
>It's not just the community colleges, though, it happens on the other taxonomies too.

I think that is what it is. I don't think there is a separate permission controlling the visibility of that operation and the route it goes to. Maybe there is an upstream issue for that.

>should we set a default number for 'Items to display'?

Items to display defaults to all if not set. I think `0` also has the same effect. I might look into removing the option from editors entirely since I can't think of the use-case where they would only want to show some of the results.

e-marie-w (approved):

joewhitsitt:
Thanks for the approval. Going to demo to customer tomorrow before merging.

joewhitsitt:
This initial slice has been approved by the Admissions team

PR #5063, opened 2022-03-30
Removed unused code in text area template.

pyrello:
I noticed while looking into how headlines are themed that this section of code was not being used in the text area block template. It was referencing a field name that is not in use on that block type.

# How to test

Code review or other testing to confirm the headline still shows.


GaryRidgway:
I will look at this!

GaryRidgway (approved):
Headline still works, REMOVE THE CODE

pyrello:
Thanks for testing, @GaryRidgway!

PR #5069, opened 2022-03-31
Feature: Updated banner pre-title font weight

bspeare:
This is a quick fix to update the banner pre-title font weight to 500 instead of 700. 

# How to test

Code review is fine.


GaryRidgway (approved):
Reviewed, not tested, but I'm sure it's fine :)

PR #5089, opened 2022-04-04
set max image size for uploads going forward

### LINKED ISSUE #4863: Set max image dimensions restriction on image fields
## Problem

Though we currently limit file uploads to under 10 MB, our image processing still seems to choke on larger dimension images. From D7, this max seemed to be around 3840px and our image docs state that our largest processed image style is 2952px wide.

## Proposed solution

- For the image media type, restrict max dimensions to a sensible limit. May need to test rendering in DEV to know for sure what the server limit is or fallback to a conservative number like 2952px.
- Update our https://sitenow.uiowa.edu/documentation/media/images-and-files documentation with information about this change.

## Additional notes

Something to keep in mind is that there might be images out there that were uploaded before this that might choke like https://github.com/uiowa/uiowa/issues/4856 or https://github.com/uiowa/uiowa/issues/4691#issuecomment-1018656724 so this could be a support issue for a long time.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4863

To Test

- `ddev composer install`
- Sync or install a site and see if you can upload an image larger than 2952x2952px. You will see a status message that the image has been resized to be under the max
- Check /media/add/image, wysiwyg upload, card image upload

Note: I will publish https://sitenow.uiowa.edu/node/326/latest once this is merged/deployed


e-marie-w:
I can test this too.

e-marie-w:
> see if you can upload an image larger than 2952x2952px

Is the expected behavior that any image over that is resized to fit?

joewhitsitt:
new images should be rejected by form validation if they exceed those dimensions

e-marie-w:
Silly question but this is for uploading an image anywhere, right? WYISWYG in the body, a text area, a card, etc? I'm going to try it on a different site, that might be my problem.

joewhitsitt:
All image fields and functionality related to images (wysiwyg image upload) should be using the media type image which this change has been made to.

e-marie-w:
On sandbox I'm still able to upload any size image. I tried putting the max_resolution value in single quotes and then syncing again but that didn't make a difference. Maybe someone else should take a look at this? I feel like I'm missing something simple.

joewhitsitt:
Sorry. My bad. Looks like the functionality has changed since D7.

>Status message - The image was resized to fit within the maximum allowed dimensions of 2952x2952 pixels. The new dimensions of the resized image are 2952x1661 pixels.

It does upload it. Reduces it.

e-marie-w:
> It does upload it. Reduces it.

Whew okay, I'll continue testing then!

e-marie-w:
When I upload an image through `admin/content/media`, I get the resize message. And when I inspect that image on a page it has the dimensions it was resized to. But, if I upload media from the media library in a WYSIWYG, it isn't resized. On the page it looks identical to the version of it that was resized. But I can inspect it see and it's still over 3000 px.

joewhitsitt:
Gosh darn, you are right. Nice testing. Switching to draft and I will update the issue.

https://www.drupal.org/project/drupal/issues/3008292

https://www.drupal.org/project/drupal/issues/3016649

joewhitsitt:
This is testable again. I've updated test instructions to composer install before hand

e-marie-w (approved):
It works! Images are resized correctly, whether added via `media/add/image`, the WYSIWYG, or the add media button on a block. I like the documentation update too.

PR #5100, opened 2022-04-06
Hotfix: Patch `smart_date` module to fix issues with end date not being set correctly

### LINKED ISSUE #5056: Grad - More Thesis Defense issues
# Problem
When creating a Thesis Defense, if you keep the same date but adjust the time, you get the same error as in #5050. Additionally, now the events are showing as a time range of 5 hours, even though no end time was put in.
Screen Shot 2022-03-28 at 5 34 44 PM

# Steps to reproduce
Follow steps in #5050 

# Proposed solution
Filed issue at https://www.drupal.org/project/smart_date/issues/3272338. TBD based on maintainer response.

pyrello:
Resolves #5056 

# How to test
- `ddev composer install`
- `ddev blt ds --site grad.uiowa.edu --define  drush.aliases.remote=grad.test`
- `ddev drush @grad.local uli node/add/thesis_defense`
- Test the following scenarios, observe no form errors, and observe that the time displays as a point in time rather than a duration:
  - Future date, leave default time.
  - Future time, leave default date.
  - Change both date and time to future values.
  - Change both date and time to past values.
- `ddev blt ds --site admissions.uiowa.edu --define drush.aliases.remote=admissions.test`
- `ddev drush @admissions.local uli node/add/event`
- Test date functionality in local events and observe no unexpected behavior. When you change the start time, you can observe the end time update based on the duration. What displays on the page is what you selected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -276,6 +276,9 @@
             "drupal/restrict_ip": {
                 "Parameter must be an array or an object that implements Countable": "https://www.drupal.org/files/issues/2022-01-11/2957482-11.patch"
             },
+            "drupal/smart_date": {
+                "Issues with time in Smart Date inline range widget": "https://www.drupal.org/files/issues/2022-04-06/3272338-3.patch"
+            },
             "drupal/jquery_ui_touch_punch": {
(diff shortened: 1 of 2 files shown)


richardbporter (approved):
Didn't notice any wonky behavior.

PR #5117, opened 2022-04-08
Prevent Purge errors from showing on every Drush command.

### LINKED ISSUE #5053: Purge warnings/errors locally when syncing sites and running other commands
# Problem
After a recent change, possibly #4942 , errors and warnings started printing on the CLI when running any Drush commands locally. This is because we don't have any purge integration locally (yet). This makes it very hard to parse console output.

# Steps to reproduce
Run `blt sync` to see multiple errors/warnings for purge diagnostics.

# Proposed solution
Initially, I thought splitting the config for the local environment would work. I tried that in https://github.com/uiowa/uiowa/pull/5052 but it doesn't really solve the problem because syncing from other environments rightfully has purge configured to log those items.

TBD -> https://github.com/uiowa/uiowa/issues/5053#issuecomment-1093326577

richardbporter:
Resolves #5053 

Also removes a committed patch to drupal/purge and enables purge_ui by default.

# How to test

- `git checkout purge-errors`
- `ddev blt drupal:install`
- See no Purge error.
- `ddev blt sync --define drush.aliases.remote=default.test`
- See no Purge error.
- `ddev drush cron -v`
- See no errors related to code checking for acquia_purge: https://github.com/uiowa/uiowa/blob/d02866b24fdfeede321345bd01068d914ddbb6b3/docroot/modules/custom/uiowa_core/uiowa_core.module#L28-L53
- `ddev drush --filter p`
- See Drush Purge commands are still available after removing patch.
- BONUS
- `ddev drush @default.local uli /admin/config/development/performance`
- See that the "Browser and proxy cache maximum age" now correctly shows "1 month" as opposed to erroneously saying "no caching" like https://default.stage.drupal.uiowa.edu/admin/config/development/performance.
- Click the Purge tab on that page.
- See that the Acquia Cloud status section is grey and reports info of "Acquia Purge only works on your Acquia Cloud environment and doesn't work outside of it."


GaryRidgway:
I will pull this and check it

GaryRidgway (approved):
tested all the things and they all work as rick said!

Issue #4294, opened 2021-10-11
Review default permissions


joewhitsitt:
https://sitenow.uiowa.edu/documentation/user-management-and-roles

Site owners are confused about how the publisher role extends the editor role. Decide if we should continue to treat publisher as an add-on role or provide publisher role all the same permissions as editor but extend where needed.

Depending on the route we choose to go, update documentation to be more clear.

pyrello:
This will be a much easier proposition after #4137 since we won't have to sync up default permissions with all the splits that have permissions.

joewhitsitt:
For example, publisher lacks layout permissions by itself, must be paired with the editor role to be useful for "publisher" type activities.

joewhitsitt:
Working on https://github.com/uiowa/uiowa/issues/4243 is highlighting this issue a bit as it relates to the authenticated role. Not sure if we should continue to treat roles with the cascading mindset (publisher should also have editor role for basic permissions) and instead move to having each role contain all the permissions they need.

mark-bennett-uiowa:
I assumed the publisher role was a stand-alone role separate from editor. I dont necessarily like that editor has access to Layout Builder because of some of the editors disregard for standards/policies/etc. My vote would be publisher has all the same roles as Editors and extends to higher level permissions, and remove LB access to editors. Just my 2 cents.

joewhitsitt:
removing layout builder from editor role would be a huge disruption to v3 unless all existing would get upgraded to something else in order to repurpose the editor role. I could see a "writer" role being introduced that would nest nicely in the writer > editor > publisher > [administrator > developer](https://github.com/uiowa/uiowa/pull/5447) hierarchy.

we also need to think about the transition from v2 to v3. i'd like to avoid any big role changes until after v2 editors (paragraphs "layout" permissions) are transitioned to v3 editors (layout permissions)

cory-skeers:
Adding a note to consider as part of this:

Publisher asked about their lack of permissions to create redirects. Currently that is limited to webmasters, but creating redirects seems like it may be in a Publisher's purview.

PR #5126, opened 2022-04-11
As a site user, I can add an image block

### LINKED ISSUE #1931: Let's have an image block
Currently we do not have a custom block that is just used for inserting an image. If someone needs to insert an image, they are just supposed to use the text area and insert it via the WYSIWYG. This is a decision that was made a while back, but it occurs to me that we should revisit it now that Layout Builder is going to be the shared solution.

makurj:
Addresses #1931 

### How to test:

1. `blt frontend`
2. `blt ds --site=default`
3. `drush @default.local cim`
4. As a webmaster/publisher/editor, add an image block to a page.
5. Try adding/removing a link to make sure the image is linked or unlinked properly.
6. Try out all 3 image styles available, you should see these visually change in aspect ratio but otherwise span full width of the column it is placed in.
7. See how the block interacts with other blocks around it. Does it need padding/margin styles?


makurj:
Not getting the form display option for the block. Could someone test that out so I can figure out if it's my local environment?

bspeare:
@makurj have you tried adding these lines to your template?  https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-banner.html.twig#L46-L48

makurj:
Trying it now. Thanks, @bspeare

makurj:
Didn't work :(

joewhitsitt:
@makurj can you provide test instructions and/or user stories for this?

makurj:
How's that?

joewhitsitt:
>How's that?

I think there are questions around this that haven't been captured related to permissions, behavior of the block itself and with alongside other blocks, but if you are satisfied I am not going to tell you how to prepare your PR.

makurj:
Could you clarify what questions are remaining "related to permissions, behavior of the block itself and with alongside other blocks"? If there's work remaining not captured on the issue, I'm happy to move this to draft.

joewhitsitt:
I updated the test instructions instead of trying to explain

makurj:
https://github.com/uiowa/uiowa/issues/1931#issuecomment-1102752842

joewhitsitt:
looking at the git diff, publisher needs the add/edit/delete permissions for `uiowa_image block content`

makurj:
Checked and publisher doesn't have those for the other blocks. Is my environment out of date? 
image

quamsta:
A few additional thoughts about images within anchor `<a>` tags without any text:

- We see a lot of image links without text to link to events from images of event flyers or other "lead-in" images
- According to W3C's Web Accessibility Initiative, these are called "functional images" and are technically okay as long as the alt text provides context about the images:
https://www.w3.org/WAI/tutorials/images/functional/

joewhitsitt:
Nope sorry. I guess we aren't doing this. This module (set of permissions) could probably go away once v2 is gone.

quamsta:
A few additional thoughts about images within anchor `<a>` tags without any text:

- We occasionally see image links without text on our sites. These sometimes link to events from images of event flyers or serve as other "lead-in" images that take people to additional information.
- According to W3C's Web Accessibility Initiative, these are called "functional images" and are technically okay as long as the alt text provides context about the images instead of describing the image itself:

_Screen readers will typically announce the image file name, the image URL, or the URL for the link destination, which is unlikely to help users understand the action that will be initiated by the image._

https://www.w3.org/WAI/tutorials/images/functional/

- SiteImprove also says something similar:

_Functional images are used to initiate actions rather than to convey information. They are used in buttons, links, and other interactive elements. The text alternative for the image should **convey the action that will be initiated** (the purpose of the image) rather than a description of the image._

https://help.siteimprove.com/support/solutions/articles/80000863904-accessibility-image-alt-text-best-practices

- We could encourage users to describe the action in the alt text of the image or override the image's alt text with a field to something specific and provide examples in the Layout Builder form like: "Learn more about the event"

makurj:
@quamsta What about the link text? How is that handled by screen readers? For this block, link text is required if there's a URL.

joewhitsitt:
Yeah, I am not sure I understand what you are bringing up here Dustin. The image requires alt text on upload. That image could be used anywhere on the site and in different contexts. For the image block, we are requiring text if a link is add to provide contextual information about the link to a screen reader through aria-label: https://github.com/uiowa/uiowa/pull/5126/files#diff-beb0e606a12c374bc5e66103950609ff1e756c375dce472eb2d1377c27cd29b6R45

A link around an image is not "named" and is flagged without an aria-label or text between.

To make this information available to more, maybe the title attribute should be filled out with this as well to provide a tooltip.

joewhitsitt:
pulling to test

joewhitsitt:
WebAIM thinks the aria-labelled link around the image is fine for screen readers. My SiteImprove widget wasn't working for some reason.
Screen Shot 2022-04-19 at 4 21 31 PM

That said, I think visual users could benefit from a title attribute:
Screen Shot 2022-04-19 at 4 24 58 PM

The hover and focus effects are a nice touch IMO:
Screen Shot 2022-04-19 at 4 32 01 PM

It seems like the image block can be used for the grid instead of text areas:
Screen Shot 2022-04-19 at 4 22 50 PM

However, I wonder if padding styles should be available:
Screen Shot 2022-04-19 at 4 23 09 PM

Fun fact, publisher doesn't need block permissions because publish doesn't have access to the layout tab. Another reminder that this is probably in order: https://github.com/uiowa/uiowa/issues/4294

joewhitsitt (approved):
Not sure how folks feel about providing the title attribute for visual users or providing padding options, but otherwise, this seems to accomplish what the issue describes.

docroot/themes/custom/uids_base/scss/global.scss
```diff
@@ -741,3 +741,16 @@ body:not(.page-node-type-article, .page-node-type-page).layout-builder-disabled
...
+    width: 100%;
+  }
+  a:focus-visible img {
+    outline: 1px solid blue;
+  }
+  a:focus-visible img,
+  a:hover img {
+    opacity: .6;
+  }
+}
```
pyrello: Instead of this going here, there should be a library added for the image block in `uids_base.libraries.yml`. A separate SCSS file should be created for the image block, following how we handle adding CSS for other block types.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
```diff
@@ -0,0 +1,51 @@
...
+%}
+{% set content_rendered = content|render %}
+{% if content_rendered is not empty %}
+  <div{{ attributes.addClass(classes) }}>
+    {{ title_prefix }}
+    {% if label %}
+      <h2{{ title_attributes }}>{{ label }}</h2>
+    {% endif %}
+    {{ title_suffix }}
+    {% block content %}
```
pyrello: `{% block %}` and `{% endblock %}` are used for overriding a section in another template or to provide sections of a template that can be overridden by another template that is implementing it. It doesn't seem like the former case is happening here and the latter case seems unlikely, so I think these can be removed.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
```diff
@@ -0,0 +1,51 @@
...
+      {% if content.field_uiowa_image_link[0]['#url'] is not empty %}
+        <a href="{{ content.field_uiowa_image_link[0]['#url']|render }}" aria-label="{{ content.field_uiowa_image_link[0]['#title']|render }}">{{ content.field_uiowa_image_image }}</a>
+      {% else %}
+        {{ content.field_uiowa_image_image }}
+      {% endif %}
+    {% endblock %}
```
pyrello: This is the `{% endblock %}` that should be removed per previous comment.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
```diff
@@ -0,0 +1,51 @@
...
+{% if content_rendered is not empty %}
+  <div{{ attributes.addClass(classes) }}>
+    {{ title_prefix }}
+    {% if label %}
+      <h2{{ title_attributes }}>{{ label }}</h2>
+    {% endif %}
+    {{ title_suffix }}
+    {% block content %}
+      {% if content.field_uiowa_image_link[0]['#url'] is not empty %}
+        <a href="{{ content.field_uiowa_image_link[0]['#url']|render }}" aria-label="{{ content.field_uiowa_image_link[0]['#title']|render }}">{{ content.field_uiowa_image_image }}</a>
```
pyrello: Are we indicating to the editor that this is how the link text is going to be used? I'm not sure I would have an expectation that link text would be used this way.

pyrello (dismissed):

docroot/themes/custom/uids_base/scss/global.scss
joewhitsitt: In a huddle with @bspeare, we thought it was light enough at the time to just include it in global.scss and if it was developed further we would move it out. That was when it was just `width: 100%` so I wonder if that line has been crossed

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
joewhitsitt: I think help text could be improved. Are we doing that for banner and card as well?

pyrello: We don't require link text in either of those cases. If link text is added, then it always shows up in the form of text in a button link. So, while I think that the field help text could be improved in those cases, I think this case is different.

docroot/themes/custom/uids_base/scss/global.scss
pyrello: If we are not going to create a library for this block here, then I think a follow-up task needs to be created to document that it needs to be done. Otherwise, in the future when people are looking for how to change stuff related to the block, it is not going to be obvious where it needs to be done. Having individual libraries per block makes it clear where future updates need to go.

makurj: I can move this to uids_base, if someone can help guide me a bit.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
joewhitsitt: i thought link text was required if banner title was empty.

docroot/themes/custom/uids_base/scss/global.scss
joewhitsitt: i can help but i need to multitask

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
pyrello: You are correct about this recent change.

joewhitsitt:
in trying to resolve that [merge conflict](https://github.com/uiowa/uiowa/pull/5126/commits/ab88be430e72709fde8b3c34126145d9579f51d2), image viewmode switching should be tested for image block, card block and make sure the accessible menu stuff still works for horizontal menu.

makurj:
Card and image blocks are fine.

makurj:
For me, tabbing through is skipping every other menu item and the hover line color is different.

f6a6ba958ee324dff885f8c95f8de886

joewhitsitt:
> For me, tabbing through is skipping every other menu item and the hover line color is different.

Pulling down to check this out

UPDATE: This is expected because of the nolinks and the lack of depth set on this menu from PROD

docroot/themes/custom/uids_base/scss/global.scss
joewhitsitt: This was moved to a separate library

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image.html.twig
makurj: This has been removed.

joewhitsitt:
Created potential follow-up issue for spacing https://github.com/uiowa/uiowa/issues/5155 and another one for title attribute https://github.com/uiowa/uiowa/issues/5156

PR #5135, opened 2022-04-14
Iowa JPEC article migration

### LINKED ISSUE #5059: News migration for Iowa JPEC.
Anything from 2019 forward.  No need to migrate news content prior to 2019.

They would like to launch by May 1.

Contacts: 
- Westlake, Casey N <casey-westlake@uiowa.edu>
- Miller, Darren <darren-miller@hawkeyesports.com>

cory-skeers:
Resolves #5059 

Extra considerations for this migration:

- Only articles from 2019 on
- Include tags
- Skip previously migrated articles (12 in total)
- Create redirects
- Use the extra date field on the existing site to alter the "created" date on the new to match "publication" dates

# How to test

Sync www.iowajpec.org locally in dev desktop from uiowa702 (db name is `www_iowajpec_org`)
`git checkout iowajpec-migration && git pull`
`blt ds --site=iowajpec.org`
`drush @org-iowajpec.local en iowajpec_migrate -y && drush @org-iowajpec.local uli /admin/config/sitenow/sitenow-migrate`
Complete the form following the help text
`drush @org-iowajpec.local ms`
`drush @org-iowajpec.local mim --all`
Check there are no errors, check migrated content looks alright and that the above considerations are accounted for (such as publication dates match old site's, didn't duplicate manually moved articles, etc.)


richardbporter:
There is one article imported from 2018 but only because of the discrepancy between the created query alter vs the date field. It can just be deleted post-migration IMO.

Screen Shot 2022-04-18 at 3 31 23 PM

I am seeing `<p>` tags and other markup in the summary for migrated articles though:
Screen Shot 2022-04-18 at 3 34 36 PM

Everything else seem to check out.

cory-skeers:
> There is one article imported from 2018 but only because of the discrepancy between the created query alter vs the date field. It can just be deleted post-migration IMO.

Nice catch! Agreed re: deleting it post-migration.

> I am seeing `<p>` tags and other markup in the summary for migrated articles though

Taking a look.

richardbporter:
Seems like only a problem when we use the source summary in `getSummaryFromTextField`. We aren't running source summary values through the `extractSummaryFromText` which does all the filtering. Maybe we should?

cory-skeers:
> Seems like only a problem when we use the source summary in `getSummaryFromTextField`. We aren't running source summary values through the `extractSummaryFromText` which does all the filtering. Maybe we should?

I think so--it helps clean things up depending on what the D7 setup was, and gives an additional option to further truncate, that might be desirable in some migrations.

cory-skeers:
Also noticed they added at least one more article to production...may need to update the `$to_skip` list, and might look to either doing that at the last moment before we're ready to deploy/run, or just realize we're going to need to do some minor post-migrate cleanup of duplicates.

richardbporter (approved):
That did it. I'm not noticing anything else. 🤞🏻

richardbporter:
> Also noticed they added at least one more article to production...may need to update the `$to_skip` list, and might look to either doing that at the last moment before we're ready to deploy/run, or just realize we're going to need to do some minor post-migrate cleanup of duplicates.

We could also run `drush users:toggle` to block everyone ahead of time.

PR #5139, opened 2022-04-15
Limit number of upcoming dates displayed on the MAUI block

### LINKED ISSUE #5072: Ability to limit number of upcoming deadlines displayed
As an editor using the MAUI upcoming dates feature, I would like to limit the number of items displayed, so that I can place a limited number of items on the homepage of my site and link to another page or the Registrar's website for additional upcoming dates.

e-marie-w:
Resolves #5072 

# Story
(Taken from the linked issue): As an editor using the MAUI upcoming dates feature, I would like to limit the number of items displayed, so that I can place a limited number of items on the homepage of my site and link to another page or the Registrar's website for additional upcoming dates.

# How to test
1. Sync down a site. 
2. If the MAUI module isn't already enabled (aka you're not syncing down Grad,) `ddev drush @yoursite.local en uiowa_maui`. Then, on `/admin/structure/types/manage/page/display`, make sure 'Allow all existing & new MAUI blocks', found under 'Blocks available for placement (all regions & placements)', is selected.
4. Add a MAUI block somewhere, test that the new options for limiting the number of deadlines displayed works as intended. 
5. A lot of the description text and default values are set up to match the articles and events listblocks, but some is custom. Check it makes sense.
6. Check articles and events listblocks still work as intended, as a function used by them was moved out of listblock.php and to its own file, so that it could also be called by the MAUI block.

Bonus points: sync down grad and make sure this doesn't mess up any existing blocks, such as on https://grad.uiowa.edu/academics/deadlines


joewhitsitt:
Just a thought that since a function is moving from list block to somewhere else, should this be run off-hours? Do you agree @richardbporter ?

docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php
```diff
@@ -223,9 +305,13 @@ public function build() {
       '\Drupal\uiowa_maui\Form\AcademicDatesForm',
       $config['session'],
       $config['category'],
-      $child_heading_size
-    );
+      $child_heading_size,
+      $config['items_to_display'],
+      $config['display_deadlines'],
+      $config['display_more_link'],
+      $config['display_more_text'],
```
richardbporter: I'm getting some errors locally related to to this code: `Notice: Undefined index: items_to_display`. I think each of these should provide a default value if the config key is not set. For example:
`$config['items_to_display'] ?? 10`

docroot/modules/custom/uiowa_maui/src/Form/AcademicDatesForm.php
```diff
@@ -116,8 +116,11 @@ public function buildForm(array $form, FormStateInterface $form_state, $session_
     if (!empty($data)) {
       $form['dates-wrapper']['dates'] = [
         '#theme' => 'uiowa_maui_session_dates',
-        '#data' => $data,
+        '#data' => ($display_deadlines == 1) ? array_slice($data, 0, $items_to_display, TRUE) : $data,
```
richardbporter: I had trouble understanding this code initially. I think it's because of the name of the variable: `$display_deadlines`. Would you be open to renaming this? How about `$limit_dates` - I think MAUI refers to them as dates as not everything is a deadline necessarily.

richardbporter (commented):
I plugged in an internal link for the "More link" and it rendered as what I've posted in the screenshot so something isn't right there. I haven't nailed that down yet.

Screen Shot 2022-04-26 at 3 17 39 PM

Also added some code comments.

richardbporter:
> Just a thought that since a function is moving from list block to somewhere else, should this be run off-hours? Do you agree @richardbporter ?

I'm unable to produce an error related to this locally so I'm leaning towards off-hours being unnecessary.

docroot/modules/custom/uiowa_maui/src/Form/AcademicDatesForm.php
e-marie-w: Yep, I can get all instances of that changed. `$limit_dates` sounds good to me.

docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php
e-marie-w: Nice catch, I will get those changed too. How did you produce the errors?

richardbporter: I synced grad.uiowa.edu, logged in and went to https://grad.uiowa.ddev.site/academics/deadlines - the warnings should display to the page. You would probably need to sync grad.uiowa.edu fresh to reproduce it. 

Another option would be to write an update hook to set the config for each block but IMO using a default value is fine, especially considering there aren't many instances of this block yet.

docroot/modules/custom/uiowa_maui/src/Form/AcademicDatesForm.php
```diff
@@ -50,7 +50,7 @@ public function getFormId() {
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $form_state, $session_prefilter = NULL, $category_prefilter = NULL, $child_heading_size = NULL) {
+  public function buildForm(array $form, FormStateInterface $form_state, $session_prefilter = NULL, $category_prefilter = NULL, $child_heading_size = NULL, $items_to_display = NULL, $display_deadlines = NULL, $display_more_link = NULL, $display_more_text = NULL) {
```
richardbporter: I think that it is not necessary for us to pass in the last two link-related arguments here. The more link could be rendered in `AcademicDatesBlock::build` right after the `$build['form']` element. For example, check if `$display_more_link` is true and if so, construct a link using https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Link.php/class/Link/8.2.x. Then the template markup could be removed.

richardbporter: I can help out with this if you'd like.

e-marie-w: Okay! I could definitely use help with this.

richardbporter:
> I plugged in an internal link for the "More link" and it rendered as what I've posted in the screenshot so something isn't right there. I haven't nailed that down yet.
> 
> Screen Shot 2022-04-26 at 3 17 39 PM
> 
> Also added some code comments.

This seems to be fixed now. 👍🏻

richardbporter:
Noticing an issue where its not possible to limit dates but not provide a more link, which I think we wanted to allow. I pushed up a failing test for this.

I think we'll want to create an update hook similar to sitenow_update_9041 in a new file, uiowa_maui.install, which sets default config values for all maui blocks. Then we can drop our `??` checks and assume values exist when accessing them. After that, I think we'd want to check if a more link was provided in addition to limiting dates before rendering that more link.

e-marie-w:
> Noticing an issue where its not possible to limit dates but not provide a more link, which I think we wanted to allow. I pushed up a failing test for this.

Okay! I'm starting on this now.

> I think we'll want to create an update hook similar to sitenow_update_9041 in a new file, uiowa_maui.install, which sets default config values for all maui blocks. Then we can drop our `??` checks and assume values exist when accessing them. After that, I think we'd want to check if a more link was provided in addition to limiting dates before rendering that more link.

Sounds good! I may need some help with this too.

richardbporter:
> > Noticing an issue where its not possible to limit dates but not provide a more link, which I think we wanted to allow. I pushed up a failing test for this.
> 
> Okay! I'm starting on this now.
> 
> > I think we'll want to create an update hook similar to sitenow_update_9041 in a new file, uiowa_maui.install, which sets default config values for all maui blocks. Then we can drop our `??` checks and assume values exist when accessing them. After that, I think we'd want to check if a more link was provided in addition to limiting dates before rendering that more link.
> 
> Sounds good! I may need some help with this too.

Ok. I pushed the start of an untested update hook. I'm in meetings so go for it! Will touch base when I can.

richardbporter (approved):
I made one last change which is to return booleans for those two checkboxes. I think that makes it easier to read and do comparisons. I tested it out one last time and it works great! If someone else could try it as well to make sure I didn't introduce regressions, that'd be appreciated.

e-marie-w:
I tested this on default and grad and haven't seen any regressions. We decided it didn't need an off-hours release, correct? So I can go ahead and merge it in?

richardbporter:
Yep, thanks!

e-marie-w:
Um slight issue I only noticed when demoing to housing: right now it displays one less than whatever number you set the limit to. Like, I set it to four and get three, set it to eight I get seven. Should we just fix it as part of the issue to enable uiowa_maui for housing?

e-marie-w:
Nevermind, I figured it out. Sometimes two deadlines exist on the same day so there's five dates listed but actually six deadlines listed.

richardbporter:
> Nevermind, I figured it out. Sometimes two deadlines exist on the same day so there's five dates listed but actually six deadlines listed.

Good catch. I think we'll want to pass the limit_dates and number_of_items to the theme function and do the array_slice there. It needs to happen after the grouping code in https://github.com/uiowa/uiowa/blob/7316a41d05e056441cf76c9e57c55d155612f472/docroot/modules/custom/uiowa_maui/uiowa_maui.module#L49-L60

I think we can just add this to #5167

PR #5154, opened 2022-04-20
Initialize pmplusbeqaa.lab.uiowa.edu multisite on uiowa03

GaryRidgway:


joewhitsitt:
this one seems odd. should we do a bit more follow-up on this?

i don't understand how pmplusbeqaa.lab = Young adults as community mental health workers in humanitarian settings

long site name btw.

GaryRidgway:
just resent an email asking for other possible names and url. I will keep this PR open just in case they are adamant

PR #5158, opened 2022-04-21
Admissions - Entity Print - Image Block config diff

joewhitsitt:
# To Test

- `git checkout main && git pull`
- `composer install && blt frontend`
- should fail due to config diff.
- `git checkout entity_print_image_block`
- should succeed.


cory-skeers (approved):
Tested--it was failing and with this it does not!

PR #5159, opened 2022-04-22
Limit number of slides to 4 for slider

### LINKED ISSUE #4645: Unlimited slider blocks create unusable slider that doesn't slide
# Problem
It appears that the number of slider blocks you can add is unlimited and creates an unusable slider if there are more than 4 items. 

# Steps to reproduce
https://careers.prod.drupal.uiowa.edu/jobs

https://user-images.githubusercontent.com/1036433/149164741-d8855116-c8aa-464f-a855-4a22d34b995d.mov

# Proposed solution
Limit amount of slides to 4

### LINKED ISSUE #4857: Slider using one slide needs `position:relative` to prevent overflow
# Problem
Slider using one slide in edge-to-edge container needs `position:relative` to prevent overflow.

Screen Shot 2022-02-24 at 11 44 09 AM

# Steps to reproduce
Add only one slider slide with a lot of headline text. 

# Proposed solution
- Add `position:relative` to the banner container, possibly scoped to the slider. 
- Require at least two slides in the slider? https://github.com/uiowa/uiowa/issues/4645

Screen Shot 2022-02-24 at 11 44 22 AM

makurj:
Resolves #4645 and #4857

# How to test

4645:
1. `blt ds --site=default`
2. `drush @default.local cim`
3. Add a slider. You should not be able to add more than 4 slides. 
4. Is the help text helpful?
5. Does it work fine with 4 slides?

4857:
1. `blt ds --site=default`
2. `blt frontend`
3. Add an edge-to-edge section.
4. Add a slider with one slide with a lot of slide headline text.
5. Does the text fit within the slide?


joewhitsitt:
pulling down to test

joewhitsitt:
So this didn't work how I expected. I thought the validation of four would happen on block save. instead the form hides any additional slides from the form and saves/displays with just the four. This setup prevents the editor from cleaning up their additional slides. After block save, cron and deleting orphans, the slide(s) still exist in the db. I don't love that.

I wonder if instead we should just have a custom validation function to handle this instead. This will prompt editors to rearrange, remove on existing blocks. Keep the help text to hopefully prevent folks from adding more than four before being flagged. Would love other thoughts on this.

As for the position fix, this worked as expected for modern browsers

makurj:
Went by https://github.com/uiowa/uiowa/issues/4645#issuecomment-1068362527

joewhitsitt:
>Went by https://github.com/uiowa/uiowa/issues/4645#issuecomment-1068362527

I understand, but this isn't quite working the way we thought. There still won't be an update hook. I will comment on the issue

joewhitsitt:
Created sliders to test against https://sandbox.prod.drupal.uiowa.edu/node/471

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -500,6 +500,16 @@ function _layout_builder_custom_block_validate(array &$form, FormStateInterface
       }
       break;
 
+    // Limit slider to 4 slides. Display an error if more than 4 on save.
+    case 'inline_block:uiowa_slider':
+      $block_form = $form_state->getValue('settings')['block_form'];
+      // Removing one for "Add more".
+      $number_of_slides = count($block_form['field_uiowa_slider_slides']) - 1;
```
joewhitsitt: Instead of assuming add more exists, this code should check for it first

joewhitsitt (approved):
Thanks for making these changes

PR #5168, opened 2022-04-25
Initial unique ID functionality. 

### LINKED ISSUE #3567: Add ID field to all block types
Epic https://github.com/uiowa/uiowa/issues/3311

# Story

As a V3 SiteNow site owner I want to be able to set a unique id on my blocks so that I can take advantage of functionality that existed in V2.

# Proposed solution
- [x] Review https://github.com/uiowa/uiowa/issues/3567#issuecomment-1103864040 below. Determine whether patch from https://www.drupal.org/project/drupal/issues/3015152 will be necessary.
- [x] If block `third_party_settings` will work, follow example in https://codimth.com/blog/web/drupal/adding-custom-fields-config-block-entity-using-thirdpartysettingsinterface to add `third_party_settings` for ID to block content block types.
- [x] Update block templates to output as id attribute on wrapping div.
- [x] Account for child IDs on collections, accordions, tabs.
- [x] Should list blocks be accounted for in this?

richardbporter:
Resolves #3567.

Todo:
- [x] Create follow-up issue to remove phpstan ignore comments and update SectionComponent::get/set method calls after https://www.drupal.org/project/layout_builder_styles/issues/3188763 lands.
- [x] Determine if other site/feature splits need updating and update, if so.

# Setup
```
git fetch
git checkout unique-id
ddev composer install
ddev blt sync
ddev drush @default.local uli
```

Add or edit any and all types of LB blocks on new and existing pages. Set a unique ID. See that each one renders appropriately.


richardbporter:
Force pushed this to reset changes made to deprecated SectionComponent method calls. That breaks other stuff and is dependent on a TBD update hook in https://www.drupal.org/project/layout_builder_styles/issues/3188763. The build will fail on the deprecation notices.

joewhitsitt:
pulling down to test

joewhitsitt:
So V2, there is a paragraph presave just in case the editor doesn't click out of the unique id field to clean up and save the input. Do we want to do the same here? sitenow_paragraphs_entity_presave. Also should the help text explain that it will be cleaned up?

Went through all the default blocks (including more) and confirmed they exist, however, blocks that don't have lb_styles (display options) the unique id field displays below the add/update block button.

Screen Shot 2022-04-27 at 9 53 39 AM

Works for prefooter blocks. Doesn't affect placed blocks like footer contact info.

>Determine if other site/feature splits need updating and update, if so.

At first I thought this would be brutal but I think with CS2, splits will just accept this new item unless the split needs to remove this (can't think of that use-case)

richardbporter:
>So V2, there is a paragraph presave just in case the editor doesn't click out of the unique id field to clean up and save the input. Do we want to do the same here?

Yes good call. Could probably be aded to _layout_builder_custom_block_validate.

> Also should the help text explain that it will be cleaned up?

Yes that sounds good.

Thanks for adjusting the weights. 

For the splits, we do have those custom BLT commands we could try. It exported a lot more than I need last time but I selectively committed what I expected.

richardbporter:
>So V2, there is a paragraph presave just in case the editor doesn't click out of the unique id field to clean up and save the input. Do we want to do the same here?

Yes good call. Could probably be aded to _layout_builder_custom_block_validate.

> Also should the help text explain that it will be cleaned up?

Yes that sounds good.

Thanks for adjusting the weights. 

For the splits, we do have those custom BLT commands we could try. It exported a lot more than I need last time but I selectively committed what I expected.

richardbporter:
The other thing I wasn't sure about was if in the SectionComponentSubscriber we wanted to set the ID only if one did not already exist. That is, do we want to respect other code that  had already set an ID? I don't know how/where that would happen off the top of head though.

joewhitsitt:
>do we want to respect other code that had already set an ID? I don't know how/where that would happen off the top of head though.

Historically, blocks haven't had an id. It is actually a problem we have run into in the past with views block pagers, etc.

Would that be something worth logging if it does occur somewhere and then consider a custom wrapper div in the future if it is an issue? Moving where it renders doesn't matter much but providing something reliable is important I think.

joewhitsitt:
>Could probably be aded to _layout_builder_custom_block_validate.

Not sure if I handled that correctly. https://github.com/uiowa/uiowa/pull/5168/commits/401ce85b4a0ae519283a2ce195dc50fd098234b3

joewhitsitt (approved):
I am okay with this, but made a decent-sized change at the end here if someone else wants to evaluate that.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -450,6 +471,13 @@ function _layout_builder_custom_block_validate(array &$form, FormStateInterface
   /** @var \Drupal\layout_builder\SectionComponent $component */
   $component = $form_state->getFormObject()->getCurrentComponent();
 
+  // Run unique_id clean for when the AJAX callback is not triggered.
+  if ($form_state->getValue(['unique_id_wrapper', 'unique_id'])) {
+    _layout_builder_custom_unique_id($form, $form_state);
```
richardbporter: Seems like we could just run this through `Html::cleanCssIdentifier`. The ajax callback was specifically for the form element and if it changes in the future, it might break this. Maybe something like: 
```
if ($unique_id = $form_state->getValue(['unique_id_wrapper', 'unique_id'])) {
  $form_state->setValue(['unique_id_wrapper', 'unique_id'], Html::cleanCssIdentifier($unique_id));
}

richardbporter:
>Determine if other site/feature splits need updating and update, if so.

I'll try running those split commands.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: Good call

richardbporter:
The feature split command caught some additional related configs to update. Site split did not.

I'm happy with this. 👍🏻

PR #5196, opened 2022-04-29
Creating Housing site split & enabling MAUI block

### LINKED ISSUE #5167: Enable uiowa_maui for Housing
As a Housing site owner, I would like the uiowa_maui module enabled so that we can take advantage of #5139 and also to test/use additional MAUI features as they become available.

# Proposed Solution
- [ ] Setup a site split: https://github.com/uiowa/uiowa/blob/21f0a7681674c89c2e5f3dfb0d32504ed669e069/config/sites/grad.uiowa.edu/config_split.config_split.site.yml#L16

e-marie-w:
Closes #5167 

Can't really be tested since it is a site split, so code review it is. 

Not sure Purge UI lines in `config_split.patch.core.entity_view_display.node.page.default.yml` are needed, but its what happened when exporting the site config so I left it for now.


richardbporter (approved):
I think the Purge UI stuff was from #5053. I just resaved the basic page default entity view display settings and exported: `drush @default.local cex`. Then I re-exported the housing site split which removed the Purge UI config. In other words, its just something we needed to add to the default config.

When this is deployed, we'll want to run `drush @housing.prod config:import --source ../config/sites/housing.uiowa.edu --partial` and then `drush @housing.prod cim`.

e-marie-w:
> I think the Purge UI stuff was from #5053. I just resaved the basic page default entity view display settings and exported: `drush @default.local cex`. Then I re-exported the housing site split which removed the Purge UI config. In other words, its just something we needed to add to the default config.

AWESOME, that was going to continue to bug me otherwise. Thank you @richardbporter!

PR #5199, opened 2022-04-29
2 plus 2 continued

### LINKED ISSUE #5157: Link from an AoS transfer tips card to a list of community colleges that participate in a 2 plus 2 plan for that major
On an AoS within the transfer tips card, display a link if at least one 2plus2 major references the AoS. This would link to a page similar to https://2plus2.admissions.uiowa.edu/majors/majors/accounting.htm

## Proposed Solution

- [x] Preprocess transfer tip on AoS to check if current AoS is referenced in any of the 2plus2 majors. If so, display a link to a /majors/majors/{aos} views page that lists out all of the 2plus2 majors but labeled as their referenced community college.

cory-skeers:
Resolves #5157  

As a prospective student interested in a 2 plus 2 plan, I want to see a list of participating colleges for a given major, so that I can more easily explore my options.

As a prospective student learning more about my area of study, I want to see all of my options (including transfer options) if they are available, so that I can compare all my options.

As an admissions site owner, I want the relevant connections between 2 plus 2 plans, areas of study, and majors to be built automatically, so that after creating my content the structure and connections are already in place with little room for user error.

# How to test

- `git checkout 2-plus-2-continued && git pull`
- Create some Community College taxonomy terms
- Create some Major (2 plus 2) content pieces connecting community colleges to areas of study
- Check the AoS page, and if there is a Transfer Tips card present, see that it is linked to 2 plus 2 view
  - The view should have a listing of all (published) community colleges which offer majors related to this area of study
  - Each listed college link should direct to the specific major page related to that AoS and community college
- On a transfer tips page, if there is a community college which offers a major related to the area of study, there should be a link to the above view page
- Check that various page titles, links, url paths, appear relevant across throughout the above connections
- Check that the following AoSs are usable by creating associated majors and checking their corresponding views pages
   - https://admissions.uiowa.ddev.site/academics/nursing-rn-bsn
   - https://admissions.uiowa.ddev.site/academics/resilience-and-trauma-informed-perspectives-certificate
   - https://admissions.uiowa.ddev.site/academics/interscholastic-athleticactivities-administration

## Note

~There is a slight inconsistency between pages of some areas of study (possible in 25/150 of them that have "of" or "and" in their title). In some of these cases, the URL path is sometimes constructed as "/asian-languages-literature/" and sometimes "/asian-languages-and-literature/". We may want to create a followup task for this specifically, but it touches content already in production such as Transfer Tips, and needs more exploration.~

Test links across the content structure for functionality, but some of the inconsistencies may be a result of paths being created both by pathauto and user-generated. https://github.com/uiowa/uiowa/pull/5199#issuecomment-1118684449

## Follow-up

Breadcrumb to be updated after a landing page is created. To come later.


cory-skeers:
Might have broken something. Getting WSODs on the 2plus2 majors pages, and not sure why...

image

joewhitsitt:
👀  Is majors in the views list? Otherwise, I might not have tested the no results stuff. 😬 

`admissions_core_taxonomy_term_view()` is where this is probably breaking

cory-skeers:
> 👀 Is majors in the views list? Otherwise, I might not have tested the no results stuff. 😬
> 
> `admissions_core_taxonomy_term_view()` is where this is probably breaking

Ah! I'd ended up with the majors view deleted locally. I was careful about not committing the deletion up, then started working on other pages, and forgot to restore.

cory-skeers:
View needs some styling yet, but is functional. One thing I'm seeing is that setting the title in the view config using a token seems to properly update the title on the page, but not the title in the meta. I think this could be handled in code pretty easily, but also want to understand if there's a views option that needs to be changed/updated for this before going that route.

image

cory-skeers:
Couple to-dos:

- [x] Figure out why there's the weird extra space in the intro paragraph on the view (after the aos name, before the period) and fix it
 - [x] Update path to use aos name, and not nid
 - [x] Finalize details like page title/meta, link title, etc and update
 - [x] Link on the transfer tips card (on the aos page) with a unique icon as well as the transfer tips content

bspeare:
@cory-skeers you may have mentioned this already, but `{{ title|striptags|trim }}` worked to remove the extra space.

cory-skeers:
> @cory-skeers you may have mentioned this already, but `{{ title|striptags|trim }}` worked to remove the extra space.

Thank you! I'd used `{{ title|trim }}` but without the `striptags`. That does the trick.

cory-skeers:
Looking into https://www.drupal.org/project/views_url_path_arguments (thanks @joewhitsitt!) either as an option, or guidance on how to get the url path working with an entity title instead of id.

cory-skeers:
The paths seem to be working now using the aos title, w/out extra modules (for example, https://admissions.uiowa.ddev.site/2plus2/majors/international-business). Links made to this from transfer tips content or card need to be updated for this.

docroot/sites/admissions.uiowa.edu/modules/admissions_core/admissions_core.module
```diff
@@ -696,6 +704,36 @@ function admissions_core_preprocess_node(&$variables) {
...
+        // We only really need to know if there areas of study,
+        // and not which or how many, because the link will just be
+        // based on the aos id.
+        $count = $query->count()->execute();
+        if ($count > 0) {
+          $area_of_study_url = \Drupal::service('path_alias.manager')->getAliasByPath('/node/' . $area_of_study_id);
+          // Get the link to the 2 Plus 2 community college aggregator page
+          // using the last segment of the aos url, which is constructed
+          // from the node title.
+          $url = \Drupal::service('path_alias.manager')->getAliasByPath('/2plus2/majors/' . basename($area_of_study_url));
```
cory-skeers: This feels like a "works 95% of the time" type solution here. There might be some cases with titles like "Asian Languages and Literature" which get handled slightly differently in how the path alias gets constructed and the contextual filter query parameter gets taken in...Need to check that.

An alternative would be to grab the node title, lower-case it, and replace spaces with hyphens (following the method that the contextual filter should be using).

cory-skeers: Update: it does break on those examples.

joewhitsitt: Would it be too bold to assume we could use pathauto's alias cleaner to get the same path? We are using `pathauto.alias_cleaner` in several configuration forms (SiteNow Articles, People, etc.)

joewhitsitt: oh maybe i read this wrong. it isn't just matching the aos path component but massaging it for contextual filtering...

cory-skeers: Yep...the contextual filtering is based on the title and not the path, which also means that the 2 plus 2 view page has a path a la "2plus2/majors/asian-languages-and-literature" where the aos has "academics/asian-languages-literature". Which is an annoying discrepancy, as well as potentially problematic in spots.

cory-skeers: I'm wondering if it's possible to contextually filter on path alias rather than title..?

joewhitsitt: so where is the views contextual filtering doing their conversion? could we leverage the same code to craft the URL?

cory-skeers: In the views options, it's set up for the argument as
```
          path_case: lower
          transform_dash: true
          break_phrase: false
```
I'm not sure where the code executes and if there's a way to hook in there. But in the view options themselves, I believe the only available options are to lower case, hyphen in place of spaces, and altering how multiple filters are handled.

joewhitsitt: docroot/core/modules/views/src/Plugin/views/argument/StringArgument.php

cory-skeers:
w/regards to the url paths, this is working based on the title with some inconsistencies.

The area of study pages drop stop words like "and":
image

The current pathing for the Majors (2 plus 2) doesn't drop these:
image

The current views setup follows the latter of these, and all link creation should work correctly this way based off the `pathauto.alias_cleaner` `cleanString` method.

If full consistency is desired and the Majors pathing gets updated, one quick workaround would be to create aliases from the "asian-languages-and-literature" style paths to the "asian-languages-literature" style.

--

I'd also realized I'd originally set the view up to link to the community colleges and not the related majors page (labelled by the community college). That has been updated.

cory-skeers:
The primary to-dos have been completed for this. May want to get feedback regarding the URL path structure, but open for technical reviews.

joewhitsitt:
It looks like path might not be reliable enough to work with. I was scratching my head why it was inconsistent. Turns out some of the AoS have custom aliases:

Screen Shot 2022-05-05 at 10 14 37 AM

The Admissions team could change these at anytime without thinking about the consequences.

cory-skeers:
> It looks like path might not be reliable enough to work with. I was scratching my head why it was inconsistent. Turns out some of the AoS have custom aliases:
> 
> Screen Shot 2022-05-05 at 10 14 37 AM
> 
> The Admissions team could change these at anytime without thinking about the consequences.

Ah! Okay, then maybe there isn't need for any actual followup, and this can just be a point of feedback.

The links to the views pages should be tested, but should work regardless of the path, since it's based only on the title forming the query. And otherwise, the other links are pulled dynamically and not constructed, so we should be good to go there.

joewhitsitt:
Pulling down to test

joewhitsitt:
This needs no results behavior (e.g. https://admissions.uiowa.ddev.site/2plus2/majors/asdf)

Also registering https://admissions.uiowa.ddev.site/academics/interscholastic-athleticactivities-administration as a 2plus2 showed no results

joewhitsitt:
Do I recall the transfer card link was supposed to be "2 plus 2 plan"

cory-skeers:
> This needs no results behavior (e.g. https://admissions.uiowa.ddev.site/2plus2/majors/asdf)
>
> Also registering https://admissions.uiowa.ddev.site/academics/interscholastic-athleticactivities-administration as a 2plus2 showed no results

Good catches!

I'm assuming the latter issue is a result of the "/" character in the title...will have to look into how that gets handled.

cory-skeers:
> Do I recall the transfer card link was supposed to be "2 plus 2 plan"

Yep--I was going off my memory, but I do have in my notes "2 plus 2 plan."

cory-skeers:
Updated the Transfer Tip card link title, and added a no results views page behavior.

Looking into how to handle the `/` behavior.

cory-skeers:
Setting back to draft until the `/` behavior is accounted for.

cory-skeers:
related d.o issue: https://www.drupal.org/project/drupal/issues/672606

So...likely to cause issues with the following AoS:
https://admissions.uiowa.ddev.site/academics/nursing-rn-bsn
https://admissions.uiowa.ddev.site/academics/resilience-and-trauma-informed-perspectives-certificate
https://admissions.uiowa.ddev.site/academics/interscholastic-athleticactivities-administration

joewhitsitt:
Sorry to mention this late. Is there a reason you created a separate view instead of using the majors view?

cory-skeers:
> Sorry to mention this late. Is there a reason you created a separate view instead of using the majors view?

Was probably a result of mixing up terminology of majors v. areas of study.

There might be a little extra validation that needs to go on in some of the preprocessing if they're combined, but I don't think there's anything else preventing that.

cory-skeers:
I believe this is ready for review again.

Couple updates since past reviewing: 

- The view page has been added to the existing `majors` view rather than new, and relevant templates/preprocesses scoped to the specific view display. 
- No results behavior has been added
- Query alter has been added with 3 overrides. Todo comment has been added if/when the core issue moves forward. Otherwise, I think this is the likely the best approach, as it is lightweight and while would require possible dev involvement should AoSs get updated or added, previous discussions suggested that is not a common occurrence 
- Removed the 'all' exception for the contextual filter, as using that would need further setup that can happen separately if needed

joewhitsitt:
taking another look

joewhitsitt (approved):
Thanks for making these changes.

>Query alter has been added with 3 overrides. Todo comment has been added if/when the core issue moves forward. Otherwise, I think this is the likely the best approach, as it is lightweight and while would require possible dev involvement should AoSs get updated or added, previous discussions suggested that is not a common occurrence

This is unfortunate but I think will be much then handling all of this on the content end of things (entering paths, create pages, etc.). Thanks for figuring out this puzzle.

PR #5225, opened 2022-05-03
Investigate webp again.

### LINKED ISSUE #5121: As a website visitor, I can load SiteNow websites quickly and save bandwidth with nextgen image formats.
Related #5119 

PageSpeed Insights reports this as the biggest opportunity for mobile and desktop. We initially tried using webp in #1341 but had to roll it back in #1531 to fix #1476.

## Proposed solution

- Clearly define what the problem is with webp and lazy.
- Compare Lighthouse reports with webp vs lazy.
- Reach out on d.o Slack if anyone has experienced this.

richardbporter:
Relates to #5121 

# How to test

```
git checkout webp
ddev composer install
ddev blt sync
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -174,6 +174,7 @@
         "drupal/views_ical": "^1.0@alpha",
         "drupal/views_tree": "^2.0@alpha",
         "drupal/webform": "^6.0",
+        "drupal/webp": "^1.0@beta",
         "drupal/weight": "^3.2",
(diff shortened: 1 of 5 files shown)


richardbporter:
Looks like Percy isn't loading any images. TravisCI must not have webp configured by default. :(

richardbporter:
Goodbye webp. 😢

PR #5226, opened 2022-05-03
Added presave hook to clear tags list and related content conditionally if those blocks are enabled.

### LINKED ISSUE #5181: Bug: Tags and related content don't display if tags are added to a node after it's been saved without them
# Problem
Tags and related content don't display on a page, article, or person if tags are added to a node after it's been saved without them. If you add tags and re-save the node, they won't appear on the published node.

You have to re-save the Article/Page/Person configuration settings in order for them to appear.

# Steps to reproduce
1. Enable the display of tags and/or related content in SiteNow Pages/Articles/Person configuration, 
ex: `/admin/config/sitenow/uiowa-pages`. (I'll just use Page as an example since I believe this bug affects all nodes the same way).
2. Create a new Page.
3. Save the Page.
4. Edit the Page and add a tag in the "Relationships" setting .
5. Save the Page again.
6. Note that the newly added tag isn't showing.

## Workaround:

I noticed if you return to `/admin/config/sitenow/uiowa-pages` and simply re-save this setting, the tags will appear on the newly created page. This may have something to do with caching since I believe we're flushing the cache upon saving `uiowa-pages`.

# Proposed solution

Unsure what's going on here, but we may need to flush out a cached node in a way that affects the display of the Prefooter region (where tags and related content currently reside) upon saving Pages/Articles/Persons.

It seems like that Prefooter region isn't getting the hint that it needs to be added when editing a Page/Article/Person and adding a tag that should trigger its appearance.

quamsta:
Resolves #5181

This PR invalidates the caches of the view blocks `tags_list` `related_content` upon saving a node only if tags and/or related content are enabled in the site's config.

# How to test

1. Enable the display of tags and/or related content in SiteNow Pages/Articles/Person configuration, 
ex: `/admin/config/sitenow/uiowa-pages`. 
2. Create a new page.
3. Save the page.
4. Edit the page and add a tag in the "Relationships" setting .
5. Save the page again.
6. Note that the newly added tag IS showing on the page, preferably test this by viewing the page in a private/incognito window as well
7. Repeat for articles and people as well


joewhitsitt:
I am interested in pulling this down soon to test, but scanning the code it looks like if there is a non-default content type (topic collection, area of study), these variables will be undefined when checked

quamsta:
@joewhitsitt  Good call. I had the invalidate methods running inside the switch cases, but there was too much repetition, so I moved them outside. But this will break on non default content types. Will refactor. Thank you!

quamsta:
non default content types should now be accounted for!

joewhitsitt:
I am going to try merge into DEV to see how this works with varnish

joewhitsitt:
Yes this clears the caches (tested locally and on dev). @quamsta, would you be open to condensing the code a bit similar to how [we handle featured image](https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L481)

```php
/**
 * Implements hook_ENTITY_TYPE_presave().
 */
function uiowa_core_node_presave(Node $node) {
  // Clear related views blocks as they don't get cleared as part of the node.
  $tag_display_config = \Drupal::config("sitenow_{$node->getType()}s.settings")->get('tag_display');
  $related_display_config = \Drupal::config("sitenow_{$node->getType()}s.settings")->get('related_display');

  if ($tag_display_config && $tag_display_config !== 'do_not_display') {
    $tagsListView = Views::getView('tags_list');
    $tagsListView->storage->invalidateCaches();
  }
  if ($related_display_config && $related_display_config !== 'do_not_display') {
    $relatedView = Views::getView('related_content');
    $relatedView->storage->invalidateCaches();
  }
}
```

quamsta:
@joewhitsitt I think I considered doing something like `sitenow_{$node->getType()}s.settings`, but the catch is that the module for person content type is named `sitenow_people` not `sitenow_persons` and the content type is `person` (I think?)

joewhitsitt (approved):
Ahhh. Okay. Good eye.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -457,3 +459,40 @@ function uiowa_core_preprocess_block(&$variables) {
...
+
+    case 'person':
+      $tag_display_config = \Drupal::configFactory()->getEditable('sitenow_people.settings')->get('tag_display');
+      $related_display_config = \Drupal::configFactory()->getEditable('sitenow_people.settings')->get('related_display');
+      break;
+
+    default:
+      $tag_display_config = 'do_not_display';
+      $related_display_config = 'do_not_display';
+  }
```
pyrello: This could be replaced with:
```php
$node_type = $node->getType();
$type_map = [
  'article' => 'articles', 
  'page' = 'pages', 
  'person' => 'people'
];
$tag_display_config = (isset($type_map[$node_type])) ? \Drupal::configFactory()->getEditable('sitenow_{$node_type}.settings')->get('tag_display') : 'do_not_display';
$related_display_config = (isset($type_map[$node_type])) ? \Drupal::configFactory()->getEditable('sitenow_{$node_type}.settings')->get('related_display') : 'do_not_display';
```

pyrello: I changed `$nodeType` => `$node_type` because I thought it was the standard inside `.module` files, but that is not the case anymore. However it is recommended that we use consistent naming pattern for variables and most use snake case in `.module` files currently: https://www.drupal.org/docs/develop/standards/coding-standards#s-functions-and-variables

quamsta: Working on this. Should update the PR soon after a meeting. Thanks @pyrello

quamsta: @pyrello This has been refactored: https://github.com/uiowa/uiowa/pull/5226/commits/03d014a28e250d99844aff74f144c0c7d741e5c4

I had to adjust the lines setting `$tag_display_config` and `$related_display_config` to use the new`$type_map` and changed the quotes to double quotes so the variables inside the quotes are parsed correctly (single quoted strings can't read variables): 

```php
  $tag_display_config = (isset($type_map[$node_type])) ? \Drupal::configFactory()->getEditable("sitenow_$type_map[$node_type].settings")->get('tag_display') : 'do_not_display';
  $related_display_config = (isset($type_map[$node_type])) ? \Drupal::configFactory()->getEditable("sitenow_$type_map[$node_type].settings")->get('related_display') : 'do_not_display';
```

This seems to work, okay to resolve?

quamsta: Refactored to https://github.com/uiowa/uiowa/pull/5226/commits/02fdb400cdce2c93153aaad05f64fabbc0f44633 based on a Slack conversation with @pyrello , tested with all three content types, seems to work great.

quamsta: Also tested on non-standard content types just to make sure saving doesn't break there.

joewhitsitt:
I think this is much more what I was going for in terms of compactness. I tested this again locally with default and custom content types. Thanks for the suggestion/changes.

PR #5230, opened 2022-05-05
Remove duplicate profiles `<title>` element.

richardbporter:
# How to test

- Go to https://education.dev.drupal.uiowa.edu/directory and inspect meta tags like title, description, canonical and schema as you navigate between profiles and back to the directory listing.
- Compare the source code, not inspector, of the following URLs:
  - https://education.uiowa.edu/directory/john-achrazoglou
    - Notice the duplicate, empty `<title>` element.
  - https://education.dev.drupal.uiowa.edu/directory/john-achrazoglou
    - Confirm that the dev URL shows the correct meta tags: one title, description, canonical and schema while viewing source.
- Also, you can run some dev profile URLs through https://tamethebots.com/tools/fetch-render?utm_source=pocket_mylist and confirm those metatags are seen by bots. You need to uncheck "Honour robots.txt" and "Flatten DOM".

PR #5197, opened 2022-04-29
Render profile metatags server-side so robots see them consistently.

richardbporter:
Resolves #5185 

# How to test

- Go to https://education.dev.drupal.uiowa.edu/directory and inspect meta tags like title, description, canonical and schema as you navigate between profiles and back to the directory listing.
- Compare the source code, not inspector, of the following URLs:
  - https://education.uiowa.edu/directory/john-achrazoglou
  - https://education.dev.drupal.uiowa.edu/directory/john-achrazoglou
  - Confirm that the dev URL shows the correct meta tags: title, description, canonical and schema while viewing source.
- Also, you can run some dev profile URLs through https://tamethebots.com/tools/fetch-render?utm_source=pocket_mylist and confirm those metatags are seen by bots. You need to uncheck "Honour robots.txt" though since dev/stage are set to deny all.

See #5185 for details.

PR #5233, opened 2022-05-05
Create initial uiowa_profiles test suite to cover meta tags functionality.

richardbporter:
Follow up from #5230 and #5197. 

# How to test

- Review the tests, check the TravisCI status/logs.
- Its probably worth trying to enable the Profiles split and importing config to make sure the config schema change didn't impact that.


GaryRidgway (approved):
Looks good to me! Test looks good, config looks good after looking at the split.

PR #5242, opened 2022-05-06
UI Press - Moved subtitle to be visible in node form

pyrello:
Resolves request by customer for subtitle field to be visible.

# How to test
Code review is sufficient.


joewhitsitt (approved):
reviewed

PR #5245, opened 2022-05-06
Hotfix profiles for sitemap change.

richardbporter:
# Problem
The uiowa_profiles sitemap route is current broken:

https://dentistry.uiowa.edu/directory/sitemap.txt
https://education.uiowa.edu/directory/sitemap.txt

## Solution

This changes the `Client::request` method to throw 503s instead of 404s when an error is encountered. It can be misleading when a 404 is returned when there is an actual error. However, I'm ambivalent about this because it looks like requests for /directory/nonexistent-person return a 500 from the profiles API so we can't detect that and still throw 404s for those.

This also changes the sitemap route to `/directory/sitemap` from `/directory/sitemap.txt`. I can't get the `.xml` extension to work so I left it off for now. I suppose we could leave it as `.txt` but that is misleading as it is no longer plain text. No extension gives us more flexibility but does Google need it?

# How to test

- Inspect https://education.dev.drupal.uiowa.edu/directory
  - Make sure the usual profiles functionality works.
-  Also the sitemap loads at https://education.dev.drupal.uiowa.edu/directory/sitemap
- See errors are now thrown instead of 404s for URLs like https://education.dev.drupal.uiowa.edu/directory/nonexistent-person


mjoneill:
> No extension gives us more flexibility but does Google need it?

Hmmmm...  good question.  We might be in uncharted territory. I'll see if they mention this, but we might need to try to point to this as a sitemap to really know.

richardbporter:
Still can't figure out the extension bit. We can push this out and re-submit them as a trial - I suppose that is better than them not working at all at present.

joewhitsitt (approved):
reviewed. 🤞

richardbporter:
We should return the same response code as the API gives us. That way, they could potentially set a 404 for non-existent people vs. an actual error.

mjoneill:
> We should return the same response code as the API gives us. That way, they could potentially set a 404 for non-existent people vs. an actual erro

Agree.  Next meeting with Profiles team I'll ask about this (or see if @briand44 can ask if I can't be there)...  we'd also want a mechanism to deploy the GTM and/or GA tag. I presume that they would deliver that to us with the content of the response code, etc...

mjoneill:
Looks like it's no dice.  

image
image

mjoneill:
Looks like our sitemap is missing some information: 

Standard Drupal sitemap:
```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-stylesheet type="text/xsl" href="/sitemap.xsl"?>
<!--Generated by the Simple XML Sitemap Drupal module: https://drupal.org/project/simple_sitemap.-->
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">
 <url>
  <loc>https://education.uiowa.edu/</loc>
  <lastmod>2022-05-10T10:39:15-05:00</lastmod>
  <changefreq>daily</changefreq>
  <priority>1.0</priority>
 </url>
```

Our People Directory Sitemap:
```
<urlset>
  <url>
    <loc>https://education.uiowa.edu/directory/adam-harry</loc>
    <changefreq>daily</changefreq>
    <priority>1.0</priority>
    <lastmod>
    ...
```

mjoneill:
Might be less about the lack of extension and more about declaring this as an XML file and a sitemap file.  

Google's reference: 
```
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url>
    <loc>http://www.example.com/foo.html</loc>
    <lastmod>2018-06-04</lastmod>
  </url>
</urlset>
```

briand44:
>This changes the Client::request method to throw 503s instead of 404s when an error is encountered. It can be misleading when a 404 is returned when there is an actual error. However, I'm ambivalent about this because it looks like requests for /directory/nonexistent-person return a 500 from the profiles API so we can't detect that and still throw 404s for those.

This has been resolved by the profiles team and now requests for `directory/nonexistent-person` will return a 404.

Issue #4314, opened 2021-10-14
As a developer deleting a D9 site, I can delete files through an automated process.


richardbporter:
# Story

As a developer deleting a D9 site, I can delete files through an automated process. 

Right now, this is a manual and error prone process. 

# Proposed solution

Copy the automation added in #4283 and add to the `uiowa:multisite:delete` command for each application environment.

briand44:
Would this just be scoped to D9 then?

richardbporter:
Yes. Updated.

richardbporter:
https://github.com/uiowa/uiowa/blob/028bb73cb09d68d666e5ffcb9db8013a232d271d/blt/src/Blt/Plugin/Commands/MultisiteCommands.php#L970-L979

briand44:
Distance and iowacreates are D7 sites so they are not relevant to this issue, correct?

richardbporter:
> Distance and iowacreates are D7 sites so they are not relevant to this issue, correct?

Correct, my mistake.

briand44:
Copy code from transfer command. May need some adjustment? Use to remove referenced sites.

briand44:
Update v2/v3 deprovision docs

PR #5271, opened 2022-05-12
Move files deletion task to take simulate option into account for `umd` command.

richardbporter:
Follow up from #4314 

The files deletion task should not run if the `simulate` option is specified. 😬 

# How to test

Code review.


pyrello (approved):
Good catch!

PR #5276, opened 2022-05-16
iFrame Title Filter module investigation

### LINKED ISSUE #3393: Consider iFrame Title Filter
## Issue

SiteImprove and other a11y tools have has previously flagged our embeds for not having a "title." Since we don't directly control the iframe markup this has been out of reach.

## Proposed Solution

- [x] Audit our current embed options for title attributes and see if there are SiteImprove issues (local extension).
- [x] Install and configure https://www.drupal.org/project/iframe_title_filter to see if it remedies any of the embeds that are missing title attributes and see if SiteImprove (local extension) no longer flags the embed for this issue.

makurj:
Resolves #3393 

# How to test
1. `blt ds --site=default`
2. `drush cim`
3. On a page, embed Youtube, Google maps, Instagram, Candid Career, Panopto, etc. and see if you can see a <title> attribute for each one.


bspeare:
Looks like this is messing with our classes within filtered_html?

Screen Shot 2022-05-16 at 2 30 39 PM
Screen Shot 2022-05-16 at 2 30 31 PM

config/default/filter.format.filtered_html.yml
```diff
@@ -71,20 +73,20 @@ filters:
...
...> <strong> <cite> <blockquote cite class=""> <code> <ul type class=""> <ol start type class=""> <li> <dl> <dt> <dd> <s> <sup> <sub> <pre class=""> <span class=""> <hr class=""> <table border summary id role aria-label aria-labelledby class=""> <caption id> <tbody class=""> <thead class=""> <tfoot class=""> <th id align valign colspan rowspan scope class=""> <td align valign colspan rowspan headers class=""> <tr align valign class=""> <drupal-media data-entity-type data-entity-uuid data-align data-caption data-view-mode title alt> <br> <mark> <small class=""> <iframe src width height allowfullscreen title loading seamless sandbox> <a href hreflang title id target aria-label rel data-* class=" bttn bttn--caps bttn--primary bttn--secondary"> <p class=" uids-component--bold-intro uids-component--light-intro"> <h2 id class=" headline headline--serif headline--underline"> <h3 id class=" headline headline--serif headline--underline"> <h4 id class=" headline headline--serif headline--underline"> <h5 id class=" headline headline--serif headline--underline"> <h6 id class=" headline headline--serif headline--underline"> <div class=" alert alert-success alert-info alert-warning alert-danger">'
```
bspeare: Why are things like `uids-component--bold-intro uids-component--light-intro` these being hard-coded here?

bspeare: Same for `headline headline--serif headline--underline`

makurj: 🤷‍♀️  That's no good. I'd say the module...

cory-skeers: I had this issue with a separate PR. Any time a change gets made and re-exported, I think it tries to consolidate and restrict.

There's probably a way to do it without it adding extra changes, but for mine I ended up just manually editing the file and adding the html needed.

makurj: Thanks, Cory. Doing just that. Only adding what's needed.

makurj: Reverting, cr didn't fix the issue.

makurj: I looked at files changed here (https://github.com/uiowa/uiowa/pull/5276/commits/5cd5b0ee6c5d430a9ddad3ec85f9ad4fba6bfdc7) and that filtered_html is not being set by the module.

makurj:
Pulled the latest changes down but not seeing the fix visually: 
image

config/default/filter.format.filtered_html.yml
makurj: Cory's commit and importing config has resolved the issue.

joewhitsitt:
pulling to test

joewhitsitt (approved):
The lead on /about is looking like a lead again.

Saw `Embedded content from uicapture.hosted.panopto.com`, `Embedded content from www.google.com` so I call this an improvement

Issue #5106, opened 2022-04-06
Accordions - Role not inside the required context


joewhitsitt:
## Problem

SiteImprove is flagging our accordions with "Role not inside the required context" related to the `role=tab` attribute.

>Make sure the element is contained by another element that has the required context role.

Lighthouse is also flagging this and pointing to these:

https://web.dev/aria-required-parent
https://web.dev/aria-required-children

## Proposed Solution

- Investigate why it is being flagged given that we have `role=tablist` wrapper. Too much separation in nesting divs perhaps?

briand44:
Can we remove this card in blocked/waiting now that we have an issue for this? https://github.com/orgs/uiowa/projects/1#card-76793651

briand44:
Thanks for creating this by the way!

joewhitsitt:
Ah.. I thought it looked familiar. Yes

briand44:
Example

https://teachonline.uiowa.edu/resources/new-semester-icon-setup
This was mentioned in the #redesign-teachonline channel

joewhitsitt:
Semi related for v2: https://github.com/uiowa/uiowa/issues/3567#issuecomment-1098081334

briand44:
>Siteimprove reports the inverse issue as well: according to the report, the TABS in that section are not associated with the TABLIST. I’m also seeing this issue in other sites, so there may be something in the accordion widget that wants adjustment. Stay tuned.

Not sure if this is helpful or not but something TM mentioned in relation to this issue.

briand44:
I reached out to TM to see if he has any recommendations to address this.

GaryRidgway:
@briand44 has there been any response from TM on this? I would like to start it but I believe it is blocked until he responds?

e-marie-w:
I started poking at this. Not sure yet if this is the problem, but https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-20 and https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example both mention/use the additional `role=tabpanel` alongside `role=tab` and `role=tablist`

briand44:
>has there been any response from TM on this?

No

>I would like to start it but I believe it is blocked until he responds?

I think we could dig into it and see if there is something we find that could improve this from an accessibility standpoint.

joewhitsitt:
Though this is still an issue, closing in favor of #4004

Issue #4004, opened 2021-08-19
Implement UIDS 4.x accordion changes


GaryRidgway:
## Stories
Implement the changes to the accordion component from UIDS 4

## Proposed solution
- Address uiowa/uids#961
- Update template
- Review any custom CSS related to accordions in SiteNow
- Search for instances of `'@uids_base/uids/accordion-item.html.twig'` to see where this is getting implemented. Review these to see if we need adjust our implementations.

<details>
<summary>Previous information</summary>
Story: As a website visitor, I can interact with accordions that are as accessible as possible.
Proposed solution:
Change the template for accordions to utilize the details tag like below.

https://www.w3schools.com/tags/tag_details.asp
<details>
<summary> Apparently details tags work in markdown EXACTLY as they do in HTML.</summary>
<div>I knew they worked in Github, but not that they were exactly the same! Pretty neat!</div>
</details>
</details>

joewhitsitt:
Based on what was discovered in https://github.com/uiowa/uiowa/pull/5277, I think it would be better to address https://github.com/uiowa/uiowa/issues/5106 as part of this issue, by simplifying the nesting and not negating the a11y of the headings.

bspeare:
I'm not seeing the `<details>` in use on https://www.w3.org/WAI/ARIA/apg/patterns/accordion/.   I also don't see it on https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/.   Do we really want to convert to `<details>` or should we just compare our version with https://www.w3.org/WAI/ARIA/apg/patterns/accordion/?

bspeare:
Looks like they are aware of the issue: https://github.com/w3c/aria-practices/issues/1856

joewhitsitt:
I believe the origin of this was a tip from TM that `<details>` was accessible out of the box and that the show/hide functionality was free of charge. The accordion example on w3 doesn't have all of the tablist role information which is what is causing the flag in https://github.com/uiowa/uiowa/issues/5106

The tablist was also something I picked up from TM a long time ago related to collapsible menus.

If we want to use this issue to revisit what an accordion should be I think that can be in scope, but we should go the extra distance to get agreement from others on campus.

bspeare:
Great points @joewhitsitt, I'm fine with either direction.

mcking65:
[...]

PR #5277, opened 2022-05-16
Accordion accessibility

GaryRidgway:
Relates to https://github.com/uiowa/uiowa/issues/5106
Added accessible label to accordion header to make sure that it was properly added into the aria mapping. Ran lighthouse, increased accessibility score.

## To test

- sync down sandbox
- go to this page https://sandbox.uiowa.ddev.site/really-cool-accordions
- run a lighthouse scan (This may only work in chrome??) (inspect > click the arrows in the row with the elements tab > Lighthouse > Generate report)
- under the accessibility section check that we are not getting dinged for `Make sure the element is contained by another element that has the required context role.`
- PROFIT


joewhitsitt:
Does this work without the role presentation? Seems weird to me to mark a heading as presentational

>Writing `<h2 role="presentation">Democracy Dies in Darkness</h2>` removes the heading semantics of the `<h2>`element, making it the equivalent of `<div>Democracy Dies in Darkness</div>`. The heading role semantics are removed, but the content itself is still available. - https://developer.mozilla.org/en-US/docs/Web/accessibility/ARIA/roles/presentation_role

e-marie-w:
> Does this work without the role presentation? Seems weird to me to mark a heading as presentational

At the moment the presentation role is the fix. It seems like the nesting is causing the problems, with the `role=tab` items not being children of `role=tablist`. `role=presentation` keeps the heading content available but doesn't make it part of the accessibility tree, allowing the `role=tab` items to be seen as children instead of grandchildren. 

I based this off of https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/#test-cases, passing example 3. That example is using list and list-items instead of tablist and tabs, but it seems to work the same. 

Removing the heading semantic but leaving the content is weird, but I don't think it causes problems. Let me know if I'm wrong, I don't completely follow if there's consequences of that part. 

We could try the `aria-owns` attribute used in passing example 5 instead?

joewhitsitt:
>We could try the aria-owns attribute used in passing example 5 instead?

Might be tricky to get all of the child information for that up to the parent.

The role presentation on a heading essentially removes the work we did with the headline helper to get the correct headings for child items and makes the document more navigable by screen readers.

So while this might fix one a11y issue, it sends us back a step for an a11y feature.

I'd like to find some time to pull this down

e-marie-w:
> The role presentation on a heading essentially removes the work we did with the headline helper to get the correct headings for child items and makes the document more navigable by screen readers.

> So while this might fix one a11y issue, it sends us back a step for an a11y feature.

Oh shoot, thanks for catching this! If I have time when you pull this down I'd be interested in pairing.

joewhitsitt:
I am not going to have time today.

If nesting is the issue, should we be trying to find a way to bring role tab to the heading element, not the button within it? I am sure there is a javascript reason for the button, but how easy would it be to change that besides revisiting in UIDS?

Or is this an issue we address later in https://github.com/uiowa/uiowa/issues/4004

PR #5281, opened 2022-05-17
IISC - Remaining content updates

### LINKED ISSUE #5261: IISC - Finish project and article content migrations
# Proposed solution

Finish the work started in #5258 to fully implement a migration for project and article content.
- Check that logic for setting article source works.
- Check existing fields to make sure they are functional.
- Add an academic year field to project content type (I missed this in the previous PR). Apparently there is some code in the housing site that might be relevant to this.
- Add any remaining fields that need to be added.

pyrello:
Resolves #5261 

# How to test

- `ddev blt ds --site iisc.uiowa.edu`
- Ignore error messages
- `ddev drush @iisc.local cim -y`
- `ddev drush @iisc.local uli admin/config/sitenow/sitenow-migrate`
- Fill in form
- Append `$settings['http_client_config']['timeout'] = '90';` to the end of `docroot/sites/iisc.uiowa.edu/settings/local.settings.php` - This should prevent a timeout related to the very large PDF file that is attached to a project node.
- `ddev drush @iisc.local ms`


pyrello:
For the "Academic year" field, I decided to go with a list (text) field using an allowed values callback and a default value callback. It is a little tricky because we want the values to change over time (e.g. 5 years from now, it lists options that weren't available before). Here is how I set it up:
* Static start year of 1950 (to what should we set this?)
* Showing 10 additional years past the current year (maybe this should only be 5 or 3?)
* It lists all options in reverse chronological order (e.g. 2022-2023, then 2021-2022)
* It uses the academic year starting with the current year as the default (2022-2023 currently)

Here is the actual code. Let me know what you think.
https://github.com/uiowa/uiowa/blob/acab3dbd8dbfca202a265c1045b43b4aec7ba933/docroot/modules/custom/uiowa_core/uiowa_core.module#L629-L659

joewhitsitt:
This looks cool. Is there a "standard" format for academic year? Also for the default value, should we check it against the academic year range or is actual year enough?

pyrello:
A couple notes:
- The D7 site uses media for files, so some of the information about the files is not available in the file record. This means that all our existing file/media code only partially works. I was thinking that I might be able to alter the query for affected entity bundles to join on the relevant database tables, but I have not been able to figure out where the media data lives in database :|
- I'm noticing that the tags work that @quamsta did is showing up on the project and partner content types, so it looks like there is a condition that is not quite right for how that code decides when to add those elements.

pyrello:
> I have not been able to figure out where the media data lives in database :|

Okay, found it. It's in the `file_managed` table.

cory-skeers:
In working on the media field process plugin, I simplified it a bit and moved the media filename check up to the first check. In the cases where a file might be used for multiple spots it avoids a lot of the checks. This assumes that we're always using the USE_EXISTING method of file copying though, and also ignores the case where the same filename could have been used in different subdirectories (since it's only based on the filename itself). Also not sure how it'll handle the human readable filenames and how those match up or not, and it doesn't pass in any alt/title info...

BUT in the most basic sense and with the unique filename assumption, it's essentially working now.

cory-skeers:
We can also use an `extract_summary` process plugin for the `body_summary` processing stuff we've been copying over all the time! Need to test that the length parameter is being passed in correctly yet.

pyrello:
> In working on the media field process plugin, I simplified it a bit and moved the media filename check up to the first check. In the cases where a file might be used for multiple spots it avoids a lot of the checks.

I saw this commit come through and thought it made a lot of sense. Also prevents more code indentation!

Thanks for all your work pushing this forward! I'll take a look again on Monday.

pyrello:
@cory-skeers Looks like you made some fantastic updates on this PR. Thanks for taking a look at it!

pyrello:
I'm running into a very strange issue here. I was hoping it was just part of the noise that can happen sometimes when running a migration, but I was able to reproduce it on subsequent migration imports.

There is a record that appears to get created twice. The first one seems to be incomplete. Not all the fields are filled in and in the Colleges and Departments field, there is a reference to a term called "c". The second one appears to be fully filled in and matches the original, which is here: https://iisc.uiowa.edu/cedar-prairie-trail-pedestrian-bridge-replacement

Looking at the taxonomy terms, there is an entry for "c".

I had not run the migration all the way through before, so I'm not clear if this is somehow related to the updates that we made last week or if this has been a problem all along.

pyrello:
Okay. So it turns out that the weird issue above is because the customer had already started entering content on the new prod site and the things I was seeing are an artifact of that.

pyrello:
Here are my notes with regard to what is not working as of now:
**People**
- [x] Image field is working inconsistently. In the 5 records I checked, it worked 1/5 times.

**Partners**
- [ ] Same issue with image field

**Projects**
Image field seems to be okay here. I spot checked about 20-25 items.
- [x] Deliverables are not using the human readable filename.

**Articles**
- [x] Source link is not populating
- [x] Reference fields are not working
- [x] Tags are not used on the D7 site, so I removed it from the migration

pyrello:
I noticed the following with the summary generation for an article:
```
Each work day brings a variety of projects to Washington&rsquo;s Chamber of Commerce director Michelle Peiffer and Debbie Stanton, director of the Washington Public Library.
```
Looks like we need to handle fancy quotes. Here is the relevant article: https://iisc.uiowa.edu/ui-partners-city-washington-iowa

pyrello:
> Source link is not populating

The previous version of the logic was not accounting for the fact that the involved fields are represented as multi-value, even though they are not. I refactored the code to account for that.

pyrello:
> Reference fields are not working

I added some initial code to start handling this. However, I realized that my next step was going to be to use the migration lookup process plugin, but I don't think that will work for a field whose values span multiple migrations. This may be the final nail in the coffee of this approach of combining three fields into one.

cory-skeers:
I'd run the migrations locally and was not seeing any issue with the images, but now after re-syncing and deleting images downloaded previously, am seeing the issue. I'm assuming it's an issue with the file copying process then, and I wasn't seeing it because the files were in my local directory from past tests.

Going to dig into it.

cory-skeers:
> I'd run the migrations locally and was not seeing any issue with the images, but now after re-syncing and deleting images downloaded previously, am seeing the issue. I'm assuming it's an issue with the file copying process then, and I wasn't seeing it because the files were in my local directory from past tests.

I take it back--I am seeing the images working for people and partners. Not sure if there was another update along the way that has fixed those, or maybe there's some other difference locally that's impacting this.

Do you know if the issue was consistently happening with the same content on multiple runs, and do you have some examples of nodes that didn't work to try and debug it? Otherwise, I think I have the reference fields (articles) working, and the deliverables (projects) using the machine readable filename.

pyrello:
> Do you know if the issue was consistently happening with the same content on multiple runs, and do you have some examples of nodes that didn't work to try and debug it?

@cory-skeers Unfortunately, I do not seem to have made any notes besides what I put in the comment above. One thing to clarify that is not clear there is that what I was seeing is that the image would be missing, but the record appeared intact. It behaved as though there was a media item referenced, but the image was not displaying. At first I thought it was an issue with Stage File Proxy, but the settings there seemed correct. So I think all of the records looked like the field was filled in, but when you would view the actual page, there was the missing image rendered.

pyrello:
Testing again.

Here is what I am seeing:
- No issues spot checking half a dozen people records.
- Images seem mostly okay for partners, but https://iisc.uiowa.ddev.site/partners/jackson-county is showing a missing image. It is the only one I found out of about 20 reviewed.
- For projects, I am still seeing the deliverables displayed using their raw filename instead of the human readable one.
- Articles seem to be working as expected.

cory-skeers:
> * For projects, I am still seeing the deliverables displayed using their raw filename instead of the human readable one.

Do the media entities have the human readable filename set as their name? When I was testing I only browsed media and checked node edit forms, but didn't check the actual displayed project page. I'm wondering if it's a display update that's needed or the migration.

pyrello:
I actually can't figure out how to get it to display the media name but link to the actual file. However, since I was able to see that the media name was getting transferred over, I'm good with moving forward with this. I'm going to add this branch to `develop` and run the migrations there to allow the customer to review them.

cory-skeers (approved):
Any additional updates can be handled in a follow-up PR after testing in dev.

Woohoo!

pyrello:
Going to merge this as soon as I get an approval on it. We will do any follow up work in an additional PR

Issue #5309, opened 2022-05-24
Provision latham.uiowa.edu on uiowa03


mark-bennett-uiowa:
Provision latham.uiowa.edu on uiowa03 to update existing [https://latham.uiowa.edu](https://latham.uiowa.edu) D7 site.

pyrello:
Closed by #5311

PR #5311, opened 2022-05-26
Security updates

pyrello:
# How to test

- `ddev blt ds`
- Check if there are config import issues.
- Check embed functionality.


pyrello:
Not sure how I got @mark-bennett-uiowa's updates from #5309 in here.

briand44 (approved):
Approving so we can be secure and deploy other code.

Issue #3878, opened 2021-07-29
P2LB to-dos for site self-transition phase


e-marie-w:
# Problem #
#3792 revealed some bugs to fix before the p2lb module can be rolled out to customers wanting to self-transition their sites to v3

## Add
- [x] #6680
- [x] #6681
- [x] #5633
- [ ] P2LB - Create action to revert to V2 version of page
- [ ] Run P2LB to see what's what
- [ ] Investigate outstanding issues to see if they are still issues

Check on Brand
```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local uli admin/content/sitenow-converter
```

Check on HMB
```
ddev blt ds --site hawkeyemarchingband.uiowa.edu && ddev drush @hawkeyemarchingband.local config-split:activate p2lb -y && ddev drush @hawkeyemarchingband.local uli admin/content/sitenow-converter
```

## Fix/Update
- [ ] Carousel items without captions are not being converted but do not throw an error/inform you that carousel items need a caption to be converted. 
- [ ] Cards - Videos - added in v2 text field not captured when page is converted. 
- [ ] Cards - Images - needs to convert to image block if card is image only
- [ ] Cards previously did not have buttons.  Are we ok with all P2LB cards having buttons?
- [ ] Background images/banners: Do we need to convert these to image blocks if there is no text area in use?  The default gradient gets in the way. 
- [ ] ListBlock (Articles and People): Update `sitenow_p2lb_process_paragraph` to set tags in new block based on tags in paragraph settings. Added todo comments in #5613.
- [ ] Sidebar menus: All pages need the `Page w/ left sidebar` layout with the `Main navigation` block set to `Vertical` that sidebar menus will work for converted pages. 
- [ ] Events block: Audit again to account for recent updates (eg date range, changes to how past/future events are selected, etc)
- [ ] if https://github.com/uiowa/uiowa/issues/5025 is done, update field_teaser to body summary as part of this transition.

### Done
- [x] #5254
- [x] #5317
- [x] #5318
- [x] Articles: `This block is broken or missing. You may be missing content or you might need to enable the original module`
- [x] Background images/banners: V3 draft does not display background images
- [x] Blockquote: V2 blockquotes need to be mapped to new blockquote blocks now available. Previously these were mapped/styled into text area blocks.
[...]

PR #5316, opened 2022-05-31
Place for some p2lb updates

### LINKED ISSUE #5254: After conversion, V2 version of page still showing on View tab
After converting a node using the conversion tool, v3 drafts and published pages currently showing v2 content when you view them. You can only see the changes when viewing content in Layout Builder.

This may be related to the tempstore or needing a cache clear.

# Proposed solution
Investigate why this is happening. Does it happen to all nodes or just some? Is it because some nodes have data in the tempstore already?

### LINKED ISSUE #5318: Incorrect block title/headline values are being set after after P2LB conversion
# Issue

Incorrect block title/headline values are being set after after P2LB conversion

- Events
  -  display/hide value of v3 headline set wrong (it displays if v2 title is set to hide and hides if v2 title set to display)
- People
  - display/hide value of v3 headline set wrong (it displays if v2 title is set to hide and hides if v2 title set to display)
- Featured Content: 
  - display/hide value of v3 headline set wrong (it displays if v2 title is set to hide and hides if v2 title set to display)
- Image Gallery
  - display/hide value of v3 headline set wrong (it displays if v2 title is set to hide and hides if v2 title set to display)
- Webform: 
  - V2 title field not being captured. It displays on the v3 block, but there’s no way to edit the text or set it to hidden.
-  Accordion: 
    -  V2 title field not being captured by v3 headline field.
- Carousel/Slider:
  -   V2 title field not being captured by v3 slider headline field

Example from Featured Content

Live site

Screen Shot 2022-05-31 at 2 24 15 PM

Local 

Screen Shot 2022-05-31 at 2 24 05 PM

# Proposed solution

?

cory-skeers:
Resolves https://github.com/uiowa/uiowa/issues/5254. 
Resolves #5318 

# How to test
(hawkeyemarchingband or brand are good v2 candidates to use, but feel free to use any v2 site)
`drush @hawkeyemarchingband.local en sitenow_p2lb`
`drush @hawkeyemarchingband.local config-split:activate p2lb`

If testing the conversion of ListBlock blocks (People, Articles), either the v2 split needs to then be disabled, or the following code needs to be updated locally to avoid the v2 check: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L1240 (eg manually set `$is_v2` to `FALSE` before the check).

Visit the P2Lb config page (/admin/config/sitenow/p2lb) or the converter page (/admin/sitenow-converter) and convert some existing pages or newly created test pages, and check their conversions.

## Updates to test:
- Articles
  - Articles block created is the views ListBlock version of an articles block, not the old Articles block
  - V2 title field is included in the v3 block, and display/hide values match
  - All default styles are set. You should be able to edit the block and immediately save without errors.
  - Pager use properly maps (hidden if hidden on v2, displayed if previously used), and does not error if the pager is used.
- Events
  - display/hide value of v3 headline matches display/hide value of v2
- People
  - display/hide value of v3 headline matches display/hide value of v2
  -   - Pager use properly maps (hidden if hidden on v2, displayed if previously used), and does not error if the pager is used.
- Featured Content:
  - display/hide value of v3 headline matches display/hide value of v2
- Image Gallery
  - display/hide value of v3 headline matches display/hide value of v2
- Webform:
  - v2 title is included as a v3 text area w/headline, placed above the webform (and display/hide value of the v3 headline matches the display/hide value of the v2 title)
- Accordion:
  - V2 title field is included in the v3 block, and display/hide values match
- Carousel/Slider:
  - V2 title field is included in the v3 block, and display/hide values match

If other issues are spotted, check that their is a corresponding note and/or issue in #3878 to account for it, and if not, it can be either addressed in this PR or added there for later followup.


cory-skeers:
Looks like ListBlock might be missing or not applied correctly when p2lb is enabled on a v2 site, leading to the People and Articles lb list blocks to not function correctly (they're available, and content can be converted into it, but certain options are incorrect or missing, leading to issues if any conversions are run).

Not sure what is missing or conflicting from having v2 and v3 turned on at the same time that's causing the issue, but it at least seems to be the root cause of a lot of down-the-chain issues I'd been seeing while working on this.

joewhitsitt:
This might have something to do with it... https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L1240

I forget what the issue was but could be traced back via git blame

cory-skeers:
> This might have something to do with it... https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L1240
> 
> I forget what the issue was but could be traced back via git blame

Looks like that would do it! Taking a look at why it was added, if it's still needed, etc., but at least it gives a way to properly test the conversion process.

cory-skeers:
Updated testing instructions and added a note to #3878 re: how to get the ListBlock blocks to work correctly. I'm not sure what issues were present that required that limitation be added, but as long as we're aware of it and it's on the roadmap for getting p2lb ready, I don't think it needs to hold up this PR in particular.

I think with that, this is ready for testing.

e-marie-w:
While testing the headline for the image gallery I noticed that the masonry style option, when set in v2, is not being set in v3. Not sure if we want to tackle it here or add it to the epic.

joewhitsitt:
@e-marie-w i think that would be covered by https://github.com/uiowa/uiowa/issues/5317, so maybe just make a note of it there.

e-marie-w:
> @e-marie-w i think that would be covered by #5317, so maybe just make a note of it there.

Wait this PR is supposed to resolve 5317. When I get done testing the rest of this I can look into fixing it.

joewhitsitt:
The commits of this PR don't really indicate much work has been done on https://github.com/uiowa/uiowa/issues/5317 unless they are in fact up to date already. Looks like @bspeare originally made that addition to the PR description and had a commit related to it (https://github.com/uiowa/uiowa/pull/5316/commits/604702b31c8dd38d753bfc776cc8c21540fe2cc1). Have the other blocks (besides banner, image gallery) been audited?

e-marie-w:
Headlines are working fine for the listblocks, but there's some non-headline issues with the v3 listblocks I want to capture.

For the article listblock, no defaults are set for display options (content alignment, display list as, image format, image position, image size). So, you can't save any changes to the block until you've selected something for all these options. Also, images will not display until the options are set. And, if you set `display list as` to list, you will still get a grid column error. So, you need to set display list as grid, select a grid column option, and then change display list back to list and then save. 

Affecting both the articles listblock and the people listblock are issues related to the pager. v3 is adding a pager when it is not set in v2. And if it is already set in v2, the v3 page will give a white screen and 'The website encountered an unexpected error.'

cory-skeers:
> The commits of this PR don't really indicate much work has been done on #5317 unless they are in fact up to date already. Looks like @bspeare originally made that addition to the PR description and had a commit related to it ([604702b](https://github.com/uiowa/uiowa/commit/604702b31c8dd38d753bfc776cc8c21540fe2cc1)). Have the other blocks (besides banner, image gallery) been audited?

In either case, they'll likely need to be audited again, as the defaults may have changed (I'm thinking button alignment style options in particular--I don't think they existed yet when these defaults were last updated).

I think it might be worth taking #5317 off of being resolved by this PR, and letting the full styles audit get prioritized separately.

cory-skeers:
Setting back to draft to get some more updates in.

cory-skeers:
Removed the style updates issue from this, and it can have a directed effort to audit prioritized later. Updated the missing default styles that led to errors, though.

@e-marie-w I believe I've taken care of all the v3 listBlock issues you came across. Thank you!

e-marie-w:
Default display options on the Articles listblock are now working great! One question though: do we want the images to default to circle format and small size? It looks like this wasn't adjustable in v2, but widescreen format and medium size more closely mimic what v2 had and are the default article listblock options anyways.

cory-skeers:
> Default display options on the Articles listblock are now working great! One question though: do we want the images to default to circle format and small size? It looks like this wasn't adjustable in v2, but widescreen format and medium size more closely mimic what v2 had and are the default article listblock options anyways.

That's a good point! I'd based it off the people listblock, which in retrospect isn't the best reference for articles defaults. I'll update it to match the default articles options.

cory-skeers:
> Default display options on the Articles listblock are now working great! One question though: do we want the images to default to circle format and small size? It looks like this wasn't adjustable in v2, but widescreen format and medium size more closely mimic what v2 had and are the default article listblock options anyways.

Got this updated. 😄

e-marie-w (approved):
Awesome work! The headlines and listbock pagers now work as expected. Also, Articles is correctly converted to a listblock and uses the default display options. Bonus points for the code comments explaining stuff like why the the headline logic needs to be flipped, or why a text area is created to hold a webform headline, they really helped me understand what was happening where.

cory-skeers:
Thanks @e-marie-w for all the testing and re-testing!

PR #5330, opened 2022-06-01
Feature: Theming residence hall content type

### LINKED ISSUE #5294: Theme residence hall pages
As a Housing site owner, I would like the residence hall pages to look like the mockup.

# Additional info
Mockup created in https://github.com/uiowa/uiowa/issues/5290

https://sandbox.prod.drupal.uiowa.edu/catlett-hall

- We will need to Create a `Border` layout builder styles option and map to a `block--enclosed` class which will eventually be used elsewhere: (https://github.com/uiowa/uiowa/issues/5179). 
- Need to set in `card.scss` something like `.card:not([class*=bg-]) .bttn--sans-serif {background: #fff;}` so that you can see the button when it is on top of a pattern. 
- Attach https://github.com/mpetroff/pannellum `js` and `css` for panoramic image field.

jQuery from d7 for panorama image field:
<details>

```
(function($) {
  Drupal.behaviors.pano = {
    attach: function(context, settings) {
        $('.floorplan-collapse').on('shown.bs.collapse', function() {
            var panoID = $(this).data('pano-id');
            var panoImage = $(this).data('pano-img');

            pannellum.viewer(panoID, {
                'type': 'equirectangular',
                'autoLoad': true,
                'panorama': panoImage
            });
        });
    }
  };
})(jQuery);

```

</details>

### LINKED ISSUE #5373: Integrate UIDS accordion a11y changes into `uids_base`
# Proposed solution

- Create a new release on https://github.com/uiowa/uids/commits/3.x
- Update templates to use `bg--white` class: https://github.com/uiowa/uids/blob/972fad8d4d8715a2e13427a7b55a1e558d541ec5/src/components/accordion/accordion.twig#L40

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5294. 
Resolves https://github.com/uiowa/uiowa/issues/5373.

## To test:
- `ddev composer install`
- `ddev blt frontend`
- `ddev blt ds --site=housing.uiowa.edu`
- you can use the existing Catlett Hall content that already has fields filled in (https://housing.uiowa.ddev.site/residence-halls/catlett-hall)
- note: for images, floorplans, and 3D view fields you need to upload local files. Using existing media library images will not work correctly.
- Appearance/layout should resemble the mockup (https://sandbox.prod.drupal.uiowa.edu/catlett-hall) for the most part. Notable differences: 
-- The static map is not yet available for the 'Getting around' section. 
-- Section of accordions has changed - Floor plans and 3D view section combined floorplans and room types accordions from before. 
-- Bathrooms accordion replaced by the restroom type section that has its own accordions with photoswipe galleries. 
-- Emergency policy section now vertically displays fields instead of horizontally.

## todo 

- [x]  This is probably going to need to be split patched for sites like hr and admissions
- [x] Create issues for `@todos`
- [x] Add comment that `fivecol` class is meant for images
- [x] Update a few fields to make them required
- [x] `aria-labelledby` on accordions are not being set
- [x] 436 is the nid, but see if you can add the field delta
- [x] fontawesome implementations need role="presentation" 
- [x] Alias pattern be [node:menu-link:parent:url:path]/[node:title] instead of residence-halls/[node:title]


joewhitsitt:
Based on https://github.com/uiowa/uiowa/pull/5330/commits/406b01a7b476c259e2a93f694ed04fafabe81b80, this is probably going to need to be split patched for sites like hr and admissions

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image-gallery.html.twig
```diff
@@ -30,6 +30,7 @@
     'block',
     'block-' ~ configuration.provider|clean_class,
     'block-' ~ plugin_id|clean_class,
+    'block--image-gallery',
```
pyrello: What is the story with this class name? It seems kind of Drupal-specific (block) for something that is potentially a component we would want to have in UIDS.

docroot/themes/custom/uids_base/templates/uids/accordion-item.html.twig
```diff
@@ -24,7 +24,7 @@
 </{{ accordion_h_size }}>
 <div id="{{ accordion_item_id }}"
      aria-labelledby="{{ accordion_hash }}"
-     class="accordion__content"
+     class="accordion__content bg--white"
```
pyrello: Do I remember that we were talking about allowing folks to set the background color for accordions?

bspeare: I think that was for the button/heading toggle color to change.   I think we always want the content area to be white.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-image-gallery.html.twig
bspeare: Thanks for pointing that out.  I was considering changing the image block styles to use in this pr, but I think that needs to be separate work in the future.

joewhitsitt:
>going to do a big dsa to check “This is probably going to need to be split patched for sites like hr and admissions”

update: dsa completed after fixing admissions, hr, sppa and previously sitenow_v2. a bit confused why a couple other sites didn't have an issue (transportation, whitman, policy.clas).

joewhitsitt:
>Add fivecol grid class... https://github.com/uiowa/uiowa/pull/5330/files#diff-6f4cf2342cac6d62454737dcc66e426bae521213b45ff29a4b26362dc5e492f4R41

whoa

Should we comment that this class should only be used for images?

joewhitsitt:
Noticing some `@todo` in this PR. I know it is still a draft working agreement, but should those be turned into to issues?

bspeare:
Yes, I just added a todo checkbox to make sure those exist before this is finished.

joewhitsitt:
moving on to testing

joewhitsitt:
- Should more fields be required? For example, neighborhood. Is there a use-case for not setting east/west. if so, should that option be captured as well? Aka, can we help guide the editor to fill out the minimum amount of content for these by providing required inputs/validation?
- For the paragraph fields, should a default exist and be open (required fields within)? I assume these will be filled out 90% of the time so maybe not a big deal.
- Should the alias pattern be `[node:menu-link:parent:url:path]/[node:title]` instead of `residence-halls/[node:title]` if providing a menu link is allowed.

Some a11y issues:
- `aria-labelledby` on accordions are not being set
- fontawesome implementations need role="presentation" or the resulting svg will default to role="img" and be flagged.
- `content-floor-plans-and-3d-view-436` id is not unique and repeated. 436 is the nid, but see if you can add the field delta to these instead.

bspeare:
@joewhitsitt I think most of the issues have been addressed if you want to take another look.   Regarding the required fields, I think they have already created all of this content on housing.prod.drupal.uiowa.edu already so perhaps tweaking this is not as important since they only have 10 or so pages of content?

joewhitsitt:
@bspeare if you could just show me a screenshot of the markup fixes, the rest of the commits look good too me. i can't get this local right now and don't want to slow it down

bspeare:
Here's the floorplan markup showing unique ids and the accordion hash being set:

Screen Shot 2022-06-23 at 9 58 56 AM

joewhitsitt (approved):
Checks out. Thanks for the fixes

PR #5346, opened 2022-06-07
Created a boolean field for card link button display

### LINKED ISSUE #4788: As a site editor, I can choose to hide the circle arrow button on the card block
# Story

As a site editor, I can choose to hide the arrow button on the card block if I don't want to have an arrow display below the card. 

Scott has mentioned that the arrows are useful, but it would be handy to turn them of in some situations.  Like on this "Contact Us" page: https://admissions.uiowa.edu/contact-us.

Screen Shot 2022-02-09 at 4 01 42 PM

# Proposed solution

- [x] - Introduce a separate settings page for layout builder block specific settings (code can live in layout_builder_custom for now).
- [x] - Similar to Featured Image settings, provide a global Show/Hide setting for card link visual indicators with help text explaining the functionality.
- [x] - We change the card field to be an options list: "Use site default", "Show", "Hide". We default this to "Use site default". Help text needs to be specific with the functionality here. Same #states logic would exist: If the card link field contains a url and the link title field is empty show a checkbox option to hide the circle button.
- [x] - Our update hook to update existing card blocks would set this field to the same default value "Use site default".

makurj:
Resolves #4788 

### How to test:
1. `blt ds --site=default`
2. Head to https://default.uiowa.ddev.site/admin/config/sitenow/uiowa-pages
3. Check that the "Display card arrow button" is already on. Save.
4. Create a new card and see if "Show button" is defaulting to "Use site default." (The dropdown will only show if there's a link but no link text.)
5. Check that the arrow button is showing on the card.
6. Test the 2 conditions (show and hide) for the "Show button" dropdown on the card and see if they behave as expected.
7. Change the "Display card arrow button" to unchecked https://default.uiowa.ddev.site/admin/config/sitenow/uiowa-pages. Save.
8. Head back to the card page and refresh.
9. Verify that the button has now disappeared, following the site default option.
10. Test the 2 conditions (show and hide) for the "Show button" dropdown on the card and see if they behave as expected.
11. `blt ds --site=hawkeyemarchingband.uiowa.edu`
12. https://hawkeyemarchingband.uiowa.ddev.site/admin/config/sitenow/uiowa-pages You should not see the option to set the card button for the site as this is a V2 site.


joewhitsitt:
follow-up from yesterday on the boolean getting `true` to pass to the uids template. looks like most of the boolean use within our codebase preprocesses and turns a if == 1 into a true variable, else false and passes that variable to to twig (https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L845). so, i guess the current method Makur implemented yesterday is okay but it is inconsistent with our uses. also, we have gone the field route with a couple of these already (e.g. field_uiowa_banner_autoplay, field_automatically_open) so i think the field approach is fine.

makurj:
If I should change the field method to something else, I'm okay with that for the sake of consistency. If we do decide to change it, some guidance would be helpful.

joewhitsitt:
I think you can leave it as is for now. If you wanted to switch it, it would be similar to
https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L845
and
https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/paragraphs/paragraph--uiowa-collection-item--accordion.html.twig#L49

but within a hook_preprocess_block() function

joewhitsitt:
still needs an update hook to set existing cards (e.g. https://sandbox.prod.drupal.uiowa.edu/cards)

makurj:
Tested. Seems to work as expected for both new and existing cards.

image

bspeare:
I think this functionality is broken: https://github.com/uiowa/uiowa/blob/eeeacca1af78335785f49fee9674b7cf162922cf/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L642-L651.    It is rendering the card image as `card__media--small`.   When that line above should be forcing `card__media--large` when `Stacked` is selected.

makurj:
https://github.com/uiowa/uiowa/blob/eeeacca1af78335785f49fee9674b7cf162922cf/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L613-L631 was added above it to get the "show button" to work.

joewhitsitt:
Sorry, I didn't test the previous commit before committing my own changes. If all of that is not necessary, feel free to remove.

joewhitsitt:
To make things more complicated, this is going to affect default content regardless of true/false

bspeare:
@makurj I switched the order of the statements and the images seem to be working now.

makurj:
Thanks, Ben! I'll test it out.

joewhitsitt:
Apologies for being dismissive of this before. Untested but I think case statements are preventing the second from firing whatever the order. We originally ran into this and flipped the order with the idea we would test the other later. The existing case statement for media handling should just have an inner if statement for card and the button logic should go in there.

makurj:
Thanks, @joewhitsitt!

makurj:
Tested by commenting out https://github.com/uiowa/uiowa/blob/eeeacca1af78335785f49fee9674b7cf162922cf/docroot/profiles/custom/sitenow/sitenow.install#L1954-L1964 and `updb` and `blt ds` both failed. Uncommented out. `updb` worked.

makurj:
The hide functionality doesn't seem to be working with any setting on the SN page for site-wide default, for either new or old cards. The only time the arrow disappears is when there is no URL or link text.

joewhitsitt:
taking a look see

joewhitsitt:
The update hook config import was failing due to the format of the file. Came across https://www.drupal.org/project/drupal/issues/2944144#comment-12481021 as the proper way to handle this because:

>The important part is "$entity = $storage->createFromStorageRecord($config_record);", the config structure and config entity structure doesn't always match 1:1.

Also, similar to the featured image (https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L401), we need to deal with the "none" or "select an option" by repurposing the site default as `_none`
Screen Shot 2022-06-23 at 3 57 23 PM

makurj:
"- None -" had disappeared when we made the field required. Will export config again. Unless we want to keep it for a reason.

joewhitsitt:
Not sure we want to "require" the field since it isn't always in-play if a card isn't linked. I thought since the featured image was doing something similar, it would make sense to keep it consistent. If you think requiring it is the better path, my apologies for the commits. Keep in mind that [4c517b8](https://github.com/uiowa/uiowa/pull/5346/commits/4c517b8cd0786b11157e1191dcccfef8ba0dc80d) undos a help text change i made which was unrelated.

makurj:
Yep, I noticed you'd referenced "none" in your work, but since I hadn't exported the config, I wanted to get that in. I will make it not required, export and push up.

joewhitsitt:
Your local/config was outdated after work was continued on it. The export occurred because you didn't cim the latest.
[4c517b8](https://github.com/uiowa/uiowa/pull/5346/commits/4c517b8cd0786b11157e1191dcccfef8ba0dc80d) is the opposite of https://github.com/uiowa/uiowa/pull/5346/commits/27c0c85a53854454da4d5713c7c675e3f2865e14#diff-e5fa029cdc7b4eb53e443662b2f183e8053bdd6bf3b24a8d01501f2a80cb9783 and https://github.com/uiowa/uiowa/pull/5346/commits/67bad4ed340fd2677e065524deeea27f2258b5e0#diff-9485115118e1c661822d9d70c68cd51b16714a7bf08e677e8e28e99b4170332b

makurj:
Yep, noticed that. Will revert and `cim`.

joewhitsitt:
if null here ([b799f2f](https://github.com/uiowa/uiowa/pull/5346/commits/b799f2fad1b394779a489587ec268a7d2d572811)) is the case for new blocks, then `sitenow_update_9044` shouldn't be needed at all for existing blocks.

Actually this paired with the update hook now breaks the functionality for existing card (default homepage top two cards)

bspeare:
I just tested this with turning off the arrows on https://admissions.uiowa.ddev.site/contact-us.   Does unchecking the box on https://admissions.uiowa.ddev.site/admin/config/sitenow/uiowa-pages supposed to stop the arrows from displaying?

Update: It looks like the "Use site default" option only works if the block is re-saved.   Is that the intent?  I thought we wanted people to be able to turn the buttons off for their sites.

makurj:
Here is my understanding: Unchecking the box on the Sitenow pages is supposed to hide the arrows everywhere, both old and new cards.

GaryRidgway:
it seems as though a saved card will have the 'Use site default' value be `NULL`, and a non saved card will have it be `'_none'`. I think I will check against both of these, does that sound reasonable?

pyrello:
> it seems as though a saved card will have the 'Use site default' value be NULL, and a non saved card will have it be '_none'. I think I will check against both of these, does that sound reasonable?

I have not reviewed the code in question, but having to account for both cases to me seems like something is not quite right. If we need to massage the values, then we might need to add something to a validation handler that converts the value if we can ensure that it is set correctly another way.

GaryRidgway:
> > it seems as though a saved card will have the 'Use site default' value be NULL, and a non saved card will have it be '_none'. I think I will check against both of these, does that sound reasonable?
> 
> I have not reviewed the code in question, but having to account for both cases to me seems like something is not quite right. If we need to massage the values, then we might need to add something to a validation handler that converts the value if we can ensure that it is set correctly another way.


We actually fixed this in the most recent code iteration, so no worries anymore!

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -629,6 +631,26 @@ function _layout_builder_custom_block_submit(array &$form, FormStateInterface $f
...
+        )[0]['title'];
+        if ($link_title) {
+          $form_state->setValue(
+            [
+              'settings',
+              'block_form',
+              'field_uiowa_card_button_display',
+            ], ['value' => 1]
+          );
+        };
```
pyrello, line 652: This could be moved up to below line 615, if line 615 was moved to be before line 614.

config/default/sitenow_pages.settings.yml
```diff
@@ -1,4 +1,4 @@
 featured_image_display_default: large
 tag_display: do_not_display
 related_display: do_not_display
-show_teaser_link_indicator: false
+show_teaser_link_indicator: true
```
pyrello, line 4: Just checking why this was changed, since it seems like an unrelated change.

Also, it looks like this is missing the config value that is added in `sitenow_pages.settings.yml` inside the module config folder below.

docroot/modules/custom/sitenow_pages/config/install/sitenow_pages.settings.yml
```diff
@@ -2,3 +2,4 @@ featured_image_display_default: medium
 tag_display: do_not_display
 related_display: do_not_display
 show_teaser_link_indicator: false
+card_link_indicator_display: true
```
pyrello, line 5: I think this needs to be added to our platform config file (above) too.

docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-card.html.twig
```diff
@@ -53,7 +53,7 @@
   'headline_level': content.field_uiowa_card_title[0]['#size'],
   'title_prefix': title_prefix,
   'title_suffix': title_suffix,
-  'visible_button': true,
+  'visible_button': field_uiowa_card_button_display,
```
pyrello, line 56: This appears to still be incorrect because the value of this field can be one of multiple strings which will always evaluate to true when used as a boolean.

pyrello: Okay... nevermind this comment, I see below that the value is being artificially set.

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -679,6 +679,43 @@ function uids_base_preprocess_block(&$variables) {
...
+
+      // Else if it's Hide...
+      elseif ($button_display_field_value == 'Hide') {
+
+        // Set to false.
+        $display_button = FALSE;
+      }
+
+      // Set the variable for the card to display the button equal to what we calculate earlier.
+      $variables['field_uiowa_card_button_display'] = $display_button;
```
pyrello: I think that we should set a different variable because this is confusing when looking at the template to see what is happening.

Also, I think this whole section should be in `sitenow_pages` instead of the theme, since we are just doing preprocessing. It is usually our practice to do preprocessing related to fields in the module that is relevant to them.

pyrello:
> We actually fixed this in the most recent code iteration, so no worries anymore!

I saw the code change and I'm saying that having two potential values be produced that mean the same thing seems like there is an error in the logic somewhere that should be fixed rather than the workaround you suggested and implemented.

bspeare:
Does this need to move back to draft?

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
makurj: Working on these changes.

GaryRidgway:
> > We actually fixed this in the most recent code iteration, so no worries anymore!
> 
> I saw the code change and I'm saying that having two potential values be produced that mean the same thing seems like there is an error in the logic somewhere that should be fixed rather than the workaround you suggested and implemented.

The generation of the `_none` value was coming from the update hook that was implemented before, and that was removed. Additionally, the workaround had been removed because there was no need anymore. 

https://github.com/uiowa/uiowa/pull/5346/files#diff-6503c7a78c0e38bcff75de02412a7be653a23bbe691eb237eb36d0b78f2cfe25R53-R59

Does this cover what you are talking about?

joewhitsitt:
I can't test at the moment, but doesn't moving that logic break the card media handling again? https://github.com/uiowa/uiowa/pull/5346/commits/bb99ba33eeb9bc6f569ae5eee7dbb8db332b5c9f

https://github.com/uiowa/uiowa/pull/5346#issuecomment-1156537727

i would make sure that is tested as part of this

bspeare:
I can confirm that https://github.com/uiowa/uiowa/pull/5346#issuecomment-1156537727 is broken again.  Move back to draft?

bspeare:
Everything seems to be working great on this.   Any concern about this checkbox showing up for `v2` sites?  

https://hawkeyemarchingband.uiowa.ddev.site/admin/config/sitenow/uiowa-pages

Screen Shot 2022-06-29 at 3 28 25 PM

joewhitsitt:
>Any concern about this checkbox showing up for v2 sites?

Yeah, should probably wrap this in the same ` if (!$is_v2)` that the teaser visual indicator stuff is in.

makurj:
image

makurj:
The site-wide setting is no longer available for V2 sites, however, per card block options are. So that logic will need to be adjusted, right, as there's no "site default" for V2 sites to set?

image

bspeare:
@makurj the card block only appears for v3 sites.   So we don't need to worry about that part.

joewhitsitt:
pulling down to test

joewhitsitt:
bad news, since this isn't captured in default content, it is getting set incorrectly for site installs.

makurj:
Okay, back to draft we go :( Would we follow this? https://wiki.uiowa.edu/display/drupaldev/Creating+Default+Content

joewhitsitt:
i am poking at this to see if we can do a quick fix until we fix default content

joewhitsitt:
On site install, default_content is creating these cards with the value `Use site default` not leaving it as null. Technically one in the same, but moving on.

So doing an export of this block to try to get the right value of `null`, nothing is registered automatically in the default content yml...

Manually, we could add:

```
  field_uiowa_card_button_display:
    -
      value: ''
```
Or see if `field_uiowa_card_button_display: null` works. However, this would be a manual step each time.

Eating my hat, the simplest solution would be to maintain the extra logic to cover default content:

`if ($button_display_field_value == NULL || $button_display_field_value == 'Use site default')`

joewhitsitt:
based on planning discussion:

todo
- [x] update the logic to check for null or Use site default
- [x] add a comment saying the `Use site default` is needed for default_content
- [x] add a todo linking to an issue where we could clean this up if we could automatically capture field = null in our default_content yaml without manual manipulation OR if we switch away from repurposing the null as the site default and instead remove the "select option" by requiring the field all the time or conditionally when link is filled.

joewhitsitt (approved):

joewhitsitt:
waiting to merge

PR #5355, opened 2022-06-10
Test role permissions for adding, editing, and deleting default content types.

### LINKED ISSUE #4000: Write tests for default content type stories
# Motivation

When we make changes to permissions (as in #3986), there is the possibility of introducing regressions related to the basic functionality around our 3 standard content types. We could automate testing this to prevent the need to write testing instructions and manually test them.

# Proposed solution

Write tests (@richardbporter suggested Behat) that include testing/execution of these stories:

- As an editor, I can create an article
- As an editor, I can create a page
- As an editor, I can create a person
- As an editor, I can edit an article
- As an editor, I can edit a page
- As an editor, I can edit a person
- As a publisher, I can create an article
- As a publisher, I can create a page
- As a publisher, I can create a person
- As a publisher, I can edit an article
- As a publisher, I can edit a page
- As a publisher, I can edit a person
- As a publisher, I can publish an article
- As a publisher, I can publish a page
- As a publisher, I can publish a person
- As a webmaster, I can create an article
- As a webmaster, I can create a page
- As a webmaster, I can create a person
- As a webmaster, I can edit an article
- As a webmaster, I can edit a page
- As a webmaster, I can edit a person
- As a webmaster, I can publish an article
- As a webmaster, I can publish a page
- As a webmaster, I can publish a person

Please double-check my work on the above list.

e-marie-w:
Closes #4000 

As a developer, I want our automated testing for site installs to catch any accidental regressions related to user role permissions for default content. 

# How to test
If your `blt/local.blt.yml` does not already have tests set up, add 
```
# Configure Drupal tests to run on ddev.
tests:
  drupal:
    web-driver: false
    sudo-run-tests: false
    mink-driver-args-webdriver: ''
    simpletest-db: ''
    simpletest-base-url: 'https://web'
    phpunit:
      -
        # The directory to scan for tests. Change to what you want to test.
        directory: '${docroot}/profiles/custom/sitenow'
        # Use the core PHPUnit config file.
        config: '${docroot}/core/phpunit.xml.dist'
        # Filter by a specific test type, class, name, etc.
        filter: PermissionsTest
```
If it is already set up, make sure to change directory to `'${docroot}/profiles/custom/sitenow` and filter to `PermissionsTest`

Run `ddev blt tests:drupal`. It should output `OK (4 tests, 72 assertions)`

Now, in `/config/default/`, change some of the `user.role` permissions that are being tested. For example, add `'use editorial transition publish'` to `user.role.editor.yml`, and/or remove it from `user.role.publisher.yml` and/or `user.role.webmaster.yml`. You could also remove the create or edit permission for article, page, or person for any of the three user roles we're testing. 

Run `ddev blt tests:drupal` again. It should fail, and the number of failures should match however many permissions you changed. The output this time should tell you exactly what assertions it failed on too.


joewhitsitt:
Sorry, what was the reasoning in removing the delete checks?

e-marie-w:
> Sorry, what was the reasoning in removing the delete checks?

We originally planned to test for publishing differences. I didn't understand the publish override thing and thought deletions were the new and only difference between editors and publishers/webmasters. I can add them back if we think they add value.

joewhitsitt:
I think because there is a difference between roles that the deletion check adds value. I'd like to know if editors could suddenly delete content they aren't supposed to.

e-marie-w:
I'll get them added back in then, thanks!

joewhitsitt (approved):
reviewed, not tested. thanks

PR #5356, opened 2022-06-10
Webform/Honeypot Updates

### LINKED ISSUE #4028: Hide allowed file extensions for webform file document component
# Story

As a webform editor, since I can't modify allowed file extensions for my file component, I don't want to show all of the allowed extensions to users as it might be overwhelming or misleading.

Screen Shot 2021-08-25 at 1 11 41 PM

# Background

We are disabling the ability for webmasters and below from modifying the allowed file extensions on a component by component basis. This is because they could add additional extensions themselves. There is no granular permission or restriction.

https://github.com/uiowa/uiowa/blob/main/docroot/profiles/custom/sitenow/sitenow.profile#L483

# Proposed solution

- [ ] Alter the description text for `edit-file--description` file webform component to not include the allowed extensions for document. Validation message will correct the user if the wrong type is attempted.

joewhitsitt:
Webform [6.1.2 to 6.1.3](https://git.drupalcode.org/project/webform/-/compare/6.1.2...6.1.3?from_project_id=59750)

Honeypot [2.01 to 2.1.1](https://git.drupalcode.org/project/honeypot/-/compare/2.0.1...2.1.1?from_project_id=53761)

Fixes #4028

## To Test
- For the general module update, there were some configurations to repair, but just poke around normal webform config for issues. This stable release has been out since February.
- As a webmaster, I can add a file component to my webform limit the list of allowed file extensions without developer interaction, however if I try to add a non-allowed extension like exe, makerbot, billmurray, I won't be able to save the component and will get a message as to why. The component's help text provides the original list of allowed extensions should the editor wish to revert back.
- Limiting the list of allowed extensions, try the form as a user and try uploading a file that is not allowed. The error message should pop up and be limited to your limited list.
- To test honeypot, just confirm honeypot and antibot are being loaded for anonymous users on a webform. You can do this by inspecting the page looking for `honeypot` and `antibot` bits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -113,7 +113,7 @@
         "drupal/google_analytics": "^4.0",
         "drupal/google_tag": "^1.3",
         "drupal/heading": "^1.4",
-        "drupal/honeypot": "^2.0",
+        "drupal/honeypot": "^2.1",
         "drupal/iframe_title_filter": "^1.1",
         "drupal/imagecache_external": "^3.0",
         "drupal/imagemagick": "^3.0",
--- a/composer.json
+++ b/composer.json
@@ -176,7 +176,7 @@
         "drupal/tvi": "^1.0@RC",
         "drupal/views_ical": "^1.0@alpha",
         "drupal/views_tree": "^2.0@alpha",
-        "drupal/webform": "^6.0",
+        "drupal/webform": "^6.1",
         "drupal/weight": "^3.2",
         "drupal/workbench_email": "^2.2",
         "drush/drush": "^11.0.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "cf44e547aa2387da12cee522c5e50f6c",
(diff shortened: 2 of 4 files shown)


joewhitsitt:
Putting this up for review before I get too distracted with more webform issues

joewhitsitt:
Unfortunately by grouping it with honeypot, this should be deployed off hours to register the new service that was introduced

pyrello:
I tested this using sitenow.uiowa.edu and it seems good to me. However, I'm not sure that I know how to test a webform well, so I'm hesitant to give full approval. Happy to do so if you think this is good, @joewhitsitt.

joewhitsitt:
Thanks for taking a look @pyrello and the adding the trim(). Generally, webform is still accepting submissions for me and allowing me to create components, configure them as a webmaster. I won't say there aren't outstanding bugs with webform or honeypot, but scanning the issue queues, I am not seeing any major regressions caused by these recent releases. The recent honeypot release supposedly fixes an issue with webform. There is overall concern that Honeypot/Webform combo is not preventing spam like it used to, especially forms with file fields, but that is a separate and on-going battle. Both modules are seeing active development, so this won't be the last update.

pyrello (approved):
Tested and found no issues.

PR #5360, opened 2022-06-13
Updating to most recent version of Drupal.

pyrello:
# How to test

Code review + whatever testing you think is appropriate. I'm good with that it passes build.


joewhitsitt (approved):
thanks for putting this together

PR #5363, opened 2022-06-13
Switch from sitenow_find_text to drupal/find_text

### LINKED ISSUE #5282: Give sitenow_find_text back to the community?
# Story

As a drupal site builder, I can leverage the sitenow_find_text module on my non-university Drupal 9 website.

# Question

Is this module something we want to generalize and give back to the community? Is it possible or is it too specific to our codebase? 

# Proposed solution

- [ ] Audit and generalize code so that it works for any Drupal website's fields, blocks.
- [x] Rename to https://www.drupal.org/project/find_text and see if Mark would hand over the project to us.

cory-skeers:
Resolves #5282 

Moving from a custom module in our codebase to a shared module on drupal.org (https://www.drupal.org/project/find_text). This PR uninstalls sitenow_find_text, installs find_text, and updates config/permissions.

A followup PR will be needed to remove the sitenow_find_text module from docroot/modules/custom after find_text is in place.

# How to test

- Check that Travis completes a successful build
- `git checkout relocate-find-text && git pull`
- `ddev composer install`
- Sync a site, watch that it completes without error
- Log in, test that Find Text is available in the toolbar, and is functionable
  - double check a search that includes a uiowa headline field, and that it appears in the search results
  - bonus points: masquerade in various roles and check permissions. `webmaster` should have access, and only that role at this time


GaryRidgway:
I will pull and test this

GaryRidgway:
ITS WORKING

As I was testing this, I found myself often times trying to search by pressing the enter button after typing something into the search box. Currently, this does not execute the search query, I would love to see this added in the future. That would really add a known paradigm for not only cited users, but keyboard only users as well. 

Otherwise, this is working well!

GaryRidgway (approved):

cory-skeers:
Thanks for testing, and the feedback @GaryRidgway! That does sound like a great incremental improvement--I'll get an issue added to the project.

joewhitsitt:
Since things are moving around, does this need to be run off hours?

cory-skeers:
> Since things are moving around, does this need to be run off hours?

That's a good question...I think for the most part it should be fine, but I wonder how things would be handled if a deployment happened while someone was using the tool? 

I think it could be unnoticeable, or cause a blip where results temporarily didn't display and the page would need to be refreshed (the page path is the same, so only the routing of which module is called is updated)...or would the tool being actively used cause a hiccup in the deployment disabling `sitenow_find_text`? I'm not sure how that works if code gets deployed while a user is actively working on a site.

joewhitsitt:
Historically, that is why we have done off-hour deployments, to avoid the temporarily unavailable errors. If the functionality only gets called on those routes and you think usage is low enough, I am fine with not scheduling one.

pyrello:
The problem will be caused by the fact that various things such as route and service definitions as well as PHP classes get cached in the database. If someone hits the Find Text page, they will be potentially be requesting a PHP class that no longer exists and I think they will get a 505 or WSOD. The biggest risk is for the sites that have domain names later in the alphabet because there is the greatest amount of time between when the code is changed and when their cache is eventually flushed. I'm not totally clear if bootstrapping the site requires that all the requested classes be present or if it only happens on the page where the class is invoked, but there is a possibility that it would affect the entire sites functionality.

Probably we should do it off-hours.

cory-skeers:
> Probably we should do it off-hours.

Thanks for the added clarity and input! Also good to consider the whole alphabet-ness that we've run into in the past. Hadn't thought of that part, and was just thinking of the immediate code deployment impact.

PR #5382, opened 2022-06-16
Move US States Allowed Values Function

### LINKED ISSUE #2764: Move US States Allowed Values Function to be shared
# Story

As a developer, I can make use of a shared list of US States with the ability to alter it to add/remove options without recreating the wheel each time.

# Proposed solution

Move admissions_core_state_options function in [admissions_core](https://github.com/uiowa/uiowa/blob/9472ebe354cc7b50f9158a90eb11e15025cacc3a/docroot/sites/admissions.uiowa.edu/modules/admissions_core/admissions_core.module#L1074-L1132) to uiowa_core module, rename it to something like uiowa_core_state_options, and add an alter similar to [sitenow paragraphs](https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_paragraphs/sitenow_paragraphs.module#L336) that allows [modification](https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_articles/sitenow_articles.module#L268). Update the fields that currently use admissions_core_state_options to instead pull from uiowa_core (name change in config files).

IISC recently added a similar Iowa counties options function to uiowa_core.

An update hook in uiowa_core or sitenow to flush caches might be needed for a clean deployment. There are examples of these types of updates already in those .install files.

makurj:
Resolves #2764 

# How to test

1. `blt ds --site=admissions.uiowa.edu`
2. Create student profile content
3. Does the state field still work?
4. Repeat for SPPA


joewhitsitt:
Looks like https://github.com/uiowa/uiowa/blob/main/docroot/sites/sppa.uiowa.edu/modules/sppa_core/sppa_core.module#L14 could be updated to use this as well. Might need to compare lists to make sure they are the same and if not, use the alter to add/remove.

joewhitsitt:
I am going to remove these unrelated/messed up permission changes, and see about getting sppa wired up as well

joewhitsitt:
IMO, this is ready for testing on admissions and sppa. Since option keys are the same, there shouldn't be any issue with lost values.

e-marie-w:
This is successfully working! I can add a state to a student profile on Admissions, and add multiple states to an opportunity on SPPA. Existing student profiles and opportunities still have the right states listed. 

The original issue for this mentioned the possibility of needing an update hook. Is that still needed, or this ready for final approval?

makurj:
@joewhitsitt Does this still need an update hook to flush caches?

joewhitsitt:
Hmm. Forgot about that. Only Admissions/SPPA would be affected during deployment and Admissions is going to get cleared pretty quickly (alpha sorted). @pyrello do you have concerns that this field could be broken for SPPA during the deploy process?

pyrello:
The update hook to flush caches doesn't run until the the process gets to that site, so there is there is still the period of time before that where the some things might be broken.

One option is to just flush the caches for those sites when the deployment starts to run, especially since there are only two of them.

makurj:
Not sure I know how to flush caches through the sitenow.install file for specific sites.

pyrello:
> Not sure I know how to flush caches through the sitenow.install file for specific sites.

This would be running `drush cim` manually against the appropriate sites once the deployment had started.

joewhitsitt:
I think what Sean is suggesting is that when we start a deployment, to manually run drush cr for those two sites right away.

makurj:
Ah, got it. Thanks @joewhitsitt I was typing a message asking about that. Thanks @pyrello. Do we have to wait for deployment on the specific app the site(s) is on?

e-marie-w (approved):

PR #5268, opened 2022-05-11
proof of concept concept3d static map field

joewhitsitt:
relates to https://github.com/uiowa/uiowa/issues/5238

todo:

- more/better validation (validate maps.uiowa.edu domain, check for marker fragment, if link -> require everything else)
- better URL parsing (i think i am assuming the url ends with the marker id)
- map aspect ratio options (similar image style ratios)?
- lazy load (placeholder bg image with fade-in)?
- clean up what I am sure is overboard drush generate crap
- make it more form api friendly?

PR #5390, opened 2022-06-22
Static map as media entity

### LINKED ISSUE #5367: As a prospective student, I can view a residence hall map on housing.uiowa.edu
# Story

As a prospective student, I can view a campus map on each residence hall page on housing.uiowa.edu

# Proposed Solution

- From grooming on 6/14/22 , it sounded like we thought moving forward with treating this as a media entity makes sense. There are some notes on the PR linked below about some things to keep in mind (lazy loading, etc)

work was started in https://github.com/uiowa/uiowa/pull/5268 but sounds like we may go a different direction now.

e-marie-w:
Resolves https://github.com/uiowa/uiowa/issues/5367
<details>
<summary>Previous development details</summary>

Quick summary of where I'm leaving this at: 
- I haven't exported config, so when you pull this branch you'll need to go to https://housing.uiowa.ddev.site/admin/structure/media and create the static map media type (make sure to set Media Source as Static Map).
- Media Library and rendering is not working yet, you can test just adding a static map link on https://housing.uiowa.ddev.site/admin/content/media
- The 2 constraint/validation files will be needed but for now are commented out as they aren't working correctly
- the `docroot/modules/custom/sitenow_media_wysiwyg/css/media-overrides.css` and `docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.libraries.yml` are most likely wrong, as nothing was rendering at all I delayed figuring out the right way to format them
- most of the static map files are identical to our panopto files, as I was trying to get just the basics working to start. `docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/Field/FieldFormatter/StaticMapUrlFormatter.php` and `docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/media/Source/StaticMap.php` borrow from other media types, though, and are probably why the media library doesn't work. When I change them to resemble panopto, though, I start getting `Plugin not found` errors.
</details>

# How to test

1. Pull down, composer install, all that jazz
2. Sync default site
3. create new page, use any wysiwyg and add a new media item
4. choose static map
5. paste the share link of any building on the uiowa map in the link field. (ex: https://maps.uiowa.edu/?id=1890#!m/579256)
6. fill out the other fields, (Tip: 17 is a good base zoom level)
7. verify thumbnail is created
8. Verify that a map is created
9. Verify that hovering over it shows a tooltip with the alt text
10. Verify that the aria-label of the link says that it will take you to maps.uiowa.edu
11. Verify that the `img` tag has alt text.
12. Edit media within WYSIWYG allows changing between Original (Medium Square), widescreen and ultrawide.
13. Test bad URLs (not maps.uiowa.edu base url, no id, no marker, incorrect marker). With dblog enabled check recent messages for an unable to create thumbnail message.


pyrello:
@e-marie-w and @GaryRidgway were looking at this today and invited me into check a few details regarding the work going in this PR. The issue I am noticing is that using a `StaticMapUrl` field type seems to indicate that the user is going to just paste in a URL to the field. After that, the user can add a label through a field added to the media type. 

The display of the map is handled via the `StaticMapUrlFormatter`, which can access the URL, but not the label, which we want to be able to set based on what the user inputs. As far as I can tell, without trying load in more entity information, there is no access to the fields that are set on the media entity when using the `StaticMapUrlFormatter`. So, it seems that the formatter cannot handle generating the map as current set up. 

When the prototype was created for this in https://github.com/uiowa/uiowa/pull/5268/files, it used a different type of field formatter that allowed for the different fields that we want to be able to use to get from the user that affect the map display. Based on what I saw today, it seems like an approach like that might work better, since we want the map to be the result of multiple values as part of a compound field.

GaryRidgway:
Thank you again for the consult and your thoughts @pyrello!

joewhitsitt:
Is the suggestion to use the custom field type (compound) within a media entity instead of extending Link or abandon the media entity approach?

pyrello:
From the quick look and little debugging we did yesterday, I don't think that extending Link will allow including a custom label unless the label is added to the URL that the user pastes in, which it can then be parsed out of. But requiring a user to understand how to do that seems a bit much.

pyrello:
FWIW, that last commit seems to have removed some config files that would not be considered theme files.

e-marie-w:
> FWIW, that last commit seems to have removed some config files that would not be considered theme files.

Ack forgot about those. We do want them removed but yes they're not theme, good catch.

joewhitsitt:
```
FILE: ...stom/sitenow_media_wysiwyg/src/Plugin/media/Source/StaticMap.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 96 | ERROR | Doc comment short description must start with a capital
    |       | letter
 97 | ERROR | Doc comment short description must be on a single line,
    |       | further text should be a separate paragraph
----------------------------------------------------------------------
FILE: ...itenow_media_wysiwyg/src/Plugin/Field/FieldType/StaticMapUrl.php
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
----------------------------------------------------------------------
 25 | ERROR | [x] Expected 1 blank line before function; 0 found
 33 | ERROR | [x] Functions must not contain multiple empty lines in
    |       |     a row; found 2 empty lines
 49 | ERROR | [x] Functions must not contain multiple empty lines in
    |       |     a row; found 2 empty lines
----------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
FILE: ...ysiwyg/src/Plugin/Field/FieldFormatter/StaticMapUrlFormatter.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 23 | ERROR | Doc comment short description must start with a capital
    |       | letter
 24 | ERROR | Doc comment short description must be on a single line,
    |       | further text should be a separate paragraph
----------------------------------------------------------------------
FILE: ...edia_wysiwyg/src/Plugin/Field/FieldWidget/StaticMapUrlWidget.php
----------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
----------------------------------------------------------------------
 24 | ERROR | [ ] Missing short description in doc comment
 39 | ERROR | [ ] Missing short description in doc comment
 57 | ERROR | [x] Use null coalesce operator instead of ternary
    |       |     operator.
 61 | ERROR | [x] Functions must not contain multiple empty lines in
    |       |     a row; found 2 empty lines
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

joewhitsitt:
i think i figured out how to get imagecache_external and media view modes to work. scope creep. not sorry.

scale is hardcoded at 2 (the middle size) and comes over as a square. so that is the original right now. added medium widescreen and medium ultrawide to make use of those options. allows us to start capturing scale later (small, large) with the same shape options.

joewhitsitt:
Screen Shot 2022-07-06 at 4 52 20 PM

Spotted an issue. Though we hide the name field, it looks like we need that default name code. Entity is being saved without a name

docroot/modules/custom/sitenow_media_wysiwyg/config/schema/sitenow_media_wysiwyg.schema.yml
```diff
@@ -0,0 +1,13 @@
...
+  type: config_object
+  label: 'Media entity static map settings'
+  mapping:
+    link:
+      type: uri
+      label: 'Link'
+    zoom:
+      type: integer
+      label: 'Zoom'
+    label:
```
joewhitsitt: this doesn't match up with the actual fields. was this pulled and modified from the remote image module? is this needed?

docroot/modules/custom/sitenow_media_wysiwyg/css/media-overrides.css
```diff
@@ -1,3 +1,9 @@
 .media-library-add-form__source-field .image-widget .image-preview {
     display: block;
 }
+
+.field--type-static-map .static-map {
```
joewhitsitt: this is no longer needed i believe

docroot/modules/custom/sitenow_media_wysiwyg/config/schema/sitenow_media_wysiwyg.schema.yml
e-marie-w: It might've been pulled from a different media entity. At the time I thought it was needed to get the fields to appear. It clearly wasn't in the end, I'll remove it.

joewhitsitt: i think you need to remove the entire file if empty for travis to be happy

e-marie-w: I thought it already existed but looking back I apparently did add it, removing now.

pyrello:
We probably want to have a default zoom value.

docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/Field/FieldFormatter/StaticMapUrlFormatter.php
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $elements = parent::viewElements($items, $langcode);
+    $values = $items->getValue();
+    foreach ($elements as $delta => $entity) {
+      $location = str_replace('!m/', '', parse_url($values[0]['uri'], PHP_URL_FRAGMENT));
+      $label = $values[0]['alt'];
```
pyrello: One tiny nit: Can we change this variable to `$alt`?

pyrello:
@e-marie-w, just a quick note about your test instructions: Since there are no changes to `composer.lock`, there is no reason for the tester to run `ddev composer install`. Not a big deal, but probably a good idea to keep your instructions focused on the tasks that are strictly necessary so that the tester doesn't have to take extra time to test your PR :)

pyrello:
Pulling down to test.

docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/Field/FieldFormatter/StaticMapUrlFormatter.php
pyrello: I'll fix this.

pyrello:
A couple things I'm noticing that seem not quite right:
- When clicking the "Edit media" button in the WYSIWYG editor, there is a caption checkbox that doesn't seem to do anything. I just checked it against an image and the behavior is the same. Not sure if there is already an issue for this, but we probably want to make a note of it as a bug if there isn't one.
- When viewing the static map media entity on the media listing page, it has a not very friendly looking "name":
  Screen Shot 2022-07-07 at 3 59 18 PM
  This probably isn't a deal-breaker, but we probably want to have the alt text show up there? We might need to slightly refactor how we are referring to the alt text so that people understand it will be the label too.

pyrello:
Outside of the above issue with how the static map appears in the media library list, this looks good and I'm ready to approve. The only reason I am not is that I think we are probably better off fixing the above issue before rolling this out rather than in a follow-up issue.

joewhitsitt:
>When clicking the "Edit media" button in the WYSIWYG editor, there is a caption checkbox that doesn't seem to do anything. I just checked it against an image and the behavior is the same. Not sure if there is already an issue for this, but we probably want to make a note of it as a bug if there isn't one.

The caption should appear below the element and allow you to add text.

Screen Shot 2022-07-07 at 4 28 22 PM

e-marie-w:
> When clicking the "Edit media" button in the WYSIWYG editor, there is a caption checkbox that doesn't seem to do anything. I just checked it against an image and the behavior is the same. Not sure if there is already an issue for this, but we probably want to make a note of it as a bug if there isn't one.

For me checking the caption checkbox allows me to add a caption, once I've saved and exited that edit media menu. I assumed this was expected behavior?
 
> When viewing the static map media entity on the media listing page, it has a not very friendly looking "name":
> This probably isn't a deal-breaker, but we probably want to have the alt text show up there? We might need to slightly refactor how we are referring to the alt text so that people understand it will be the label too. too.

So this was actually on purpose. Panopto and some other media types also have that auto-naming structure for files. This might be wrong, but I think the reasoning was that it isn't obvious the name field is naming the file instead of adding a title for the displayed content. 

> Since there are no changes to composer.lock, there is no reason for the tester to run ddev composer install.

I didn't actually write these test instructions but thank you for the explanation of when composer install is actually needed! Turns out I have been running it when it wasn't needed.

Edit: Once again Joe is faster than me

joewhitsitt:
>When viewing the static map media entity on the media listing page, it has a not very friendly looking "name"

>So this was actually on purpose. Panopto and some other media types also have that auto-naming structure for files. This might be wrong, but I think the reasoning was that it isn't obvious the name field is naming the file instead of adding a title for the displayed content.

We do have the marker id, the alt text and the zoom level. Since we are hiding the name field, I can be on board with crafting one: media:static_map:marker-579168:zoom-17

Based on our discussion alt text didn't seem like a good administrative label

pyrello:
> Since we are hiding the name field

Is the name field being hidden to avoid people being confused about that vs the alt text?

What would it take to actually display a thumbnail of the map?

joewhitsitt:
> > Since we are hiding the name field
> 
> Is the name field being hidden to avoid people being confused about that vs the alt text?

Yes with the addition of simplicity and consistency

> What would it take to actually display a thumbnail of the map?

Similar to Panopto, we could grab separately grab a 512px version of the map making sure we have a fallback in case it fails.

This makes me think of two things:
- We aren't checking the concept3D response currently to make sure an image returns.
- Panopto thumbnails are never cleaned up.

joewhitsitt:
I'd be curious to see what happens for the media creation if the image URL didn't resolve (perhaps disable validation temporarily and pass a bad formed link in to simulate their service being down or something). If it handles poorly, we may need to use GuzzleHttp to ping it first and return error if there is an issue.

e-marie-w:
> We do have the marker id, the alt text and the zoom level. Since we are hiding the name field, I can be on board with crafting one: media:static_map:marker-579168:zoom-17

I've figured out how to grab the marker value, because that's coming from the URL. But the zoom level is set on the same window where the name is already generated. Right now I'm not seeing how we can grab the zoom value to add to the name field. Am I missing something obvious. or should the name just be `media:static_map:marker-584524` instead?

joewhitsitt:
ah makes sense. i think using the marker id is an improvement.

pyrello:
Per this comment, do we want to add the "provider" (maps.uiowa.edu) to the name to make these easily distinguishable from other providers once that becomes a thing?

e-marie-w:
> Per this comment, do we want to add the "provider" (maps.uiowa.edu) to the name to make these easily distinguishable from other providers once that becomes a thing?

Since other providers are not a thing yet, I feel like it's not needed at this time. But I'm not against it if we think it's necessary.

I did notice that when someone adds a static map via Content->Media->Add media the user will see the description `A static map image from maps.uiowa.edu.` Inserting from Media Library just has `Static map URL` on the first screen. Only on the second screen does it have text specifying maps.uiowa.edu. 

So, regardless of what we decide for the name, I think we should change that `Static map URL` label. We could do it like UICapture, where it specifies `Panopto URL`. Or we could do it like Remote video, which has `Add Remote video via URL` and then text below the input box specifying `Allowed providers: University of Iowa, Vimeo, YouTube`.

pyrello:
> Since other providers are not a thing yet, I feel like it's not needed at this time. But I'm not against it if we think it's necessary.

I haven't looked closely at what is involved in setting up the display name. The only reason to do this now vs later is if it would require an update hook or create more work down the road to change it from what it is now beyond just adding that information.

joewhitsitt:
I just tested to see if a media item is created on that first form (media_library_add). Nothing new is added to the database at that step. It says "The media item has been created but has not yet been saved." I am not sure what that actually means because this is what is called docroot/core/modules/media_library/src/Form/AddFormBase.php updateFormCallback and it is just an ajax replace command setting up the next form and setting focus. It is basically a simple input to fill in the next form (media entity). Looking at the other media_entity_* modules we have, none of them have validation at that step.

As I said before, the likelihood of an issue with the hardcoded the zoom during the get_headers validation is next to none.

e-marie-w:
So, can I remove that `'#required = TRUE'` I added to the media_library_add then? If none of the other media types are checking it seems unnecessary.

joewhitsitt:
> So, can I remove that `'#required = TRUE'` I added to the media_library_add then? If none of the other media types are checking it seems unnecessary.

That should stay. I am talking about the custom validation at that point

joewhitsitt:
I noticed the `$response` variable in the thumbnail_url section is unused. Might want to just go through that file one more time to clean up things that aren't needed. I think the fallback if fine if the thumbnail doesn't exist.

Added a test instruction about passing in a bad URL to test the thumbnail fallback log message.

I am going to create a follow up issue to look into garbage collection of panopto and static map thumbnails.

joewhitsitt:
Garbage collection issue: https://github.com/uiowa/uiowa/issues/5461

e-marie-w:
This is going to sound weird but if I delete the unused $response variable and put an invalid URL in the media library add it will not let me continue to the second form. It won't give me an error at that moment, but if I close the media library and reopen it then it will say:

`'Failed to fetch file due to error Client error: GET https://staticmap.concept3d.com/map/static-map/?map=1890&loc=58&scale=1&zoom=17 resulted in a 400 Bad Request response...`

So, it's like removing the $response variable somehow adds the validation from the second form to the first somehow.

joewhitsitt:
I guess `$this->client->request('GET', $thumbnail_url);` is needed for the try/catch but assigning it to a variable that is never used is probably not needed.

config/default/filter.format.filtered_html.yml
```diff
@@ -73,7 +73,7 @@ filters:
...
+      allowed_html: '<em> <strong> <cite> <blockquote cite class=""> <code> <ul type class=""> <ol start type class=""> <li> <dl> <dt> <dd> <s> <sup> <sub> <pre class=""> <span class=""> <hr class=""> <table border summary id role aria-label aria-labelledby class=""> <caption id> <tbody class=""> <thead class=""> <tfoot class=""> <th id align valign colspan rowspan scope class=""> <td align valign colspan rowspan headers class=""> <tr align valign class=""> <drupal-media data-entity-type data-entity-uuid data-align data-caption data-view-mode title alt> <br> <mark> <small class=""> <iframe src width height allowfullscreen title loading seamless sandbox> <a href hreflang title id target aria-label rel data-* class=""> <p class=""> <h2 id class=""> <h3 id class=""> <h4 id class=""> <h5 id class=""> <h6 id class=""> <div class="">'
```
pyrello: Could you provide some context about these changes?

pyrello:
Pulling this down to take a look.

pyrello:
Is this a message we are expecting related to the marker being wrong?
Screen Shot 2022-07-18 at 4 50 24 PM

pyrello:
I'm satisfied with all the elements of this except the last step, checking the error handling. 

In the step where I paste the URL, I am able to paste valid URL's (e.g. https://google.com) that are not going to be accepted on the next screen. This is a little confusing if you enter a URL that will not actually work, but you don't get alerted to it until you get to the second step, fill in the rest of the information, then get an error message.
Screen Shot 2022-07-18 at 4 55 16 PM
Screen Shot 2022-07-18 at 4 55 43 PM
Screen Shot 2022-07-18 at 4 56 29 PM

config/default/filter.format.filtered_html.yml
joewhitsitt: we needed to allow static map on the filtered_html config and when exported, these were changed. from my previous inspection, it moved the `<a>` tag stuff and changed all of the `class` attributes to include `=""`, however now that you have pointed it out I do see some regressions (table role, iframe). not sure what we can do to get these to stick.

e-marie-w:
I need to look into the filtered html and the invalid marker message. But I know the URL not giving an error until the second page is how many of the other media entities work, which is one reason we left it that way.

joewhitsitt:
>Is this a message we are expecting related to the marker being wrong?

That is not what is expected.. Do you know when and under what conditions that happened?

Update: oh did you pass in a valid URL like google.com? Since the response is valid, it would try to get the thumbnail and fail later on.

config/default/filter.format.filtered_html.yml
cory-skeers: Whenever I've made an addition to the `allowed_html`, I've been manually adjusting YML file. Handling it through an export seems to lead to lots of undesired changes. My hunch is that the export tries to consolidate to what it thinks is the most concise, restricted allowable list that matches what you've entered, but that the process has some issues (at least with our setup). That's a hunch I haven't dug into too far, though, since it doesn't come up too often.

tl;dr: Easiest way I've found around this issue is to manually adjust the YML in the codebase (and double and triple check I've only changed what I've intended).

pyrello:
> > Is this a message we are expecting related to the marker being wrong?

> That is not what is expected.. Do you know when and under what conditions that happened?

I used the share URL from this: https://maps.uiowa.edu/?id=1890#!m/545690?s/kinnick?ct/51247. Apparently, the 360 degree view creates an invalid marker.

joewhitsitt:
Made a note on this issue https://github.com/uiowa/uiowa/issues/5427 regarding unintended filtered_html config export result

joewhitsitt:
The file_mdm error seems to be related to limited exif data? https://www.drupal.org/project/file_mdm/issues/3229460

`sitenow_media_wysiwyg_file_presave` is called in our process and I am wondering if our custom code doesn't like the image in this situation. need to dive deeper here.

update: it pretty much skipped past the meat of the logic located there. this error doesn't cause blockers for the functionality or the user so I wonder if we can just ignore this here and follow that upstream issue?

joewhitsitt:
As for no validation on the media_library add form. I think leave as is for now. Like Marie said, we are following the patterns set by the other media_entity modules
Screen Shot 2022-07-19 at 4 16 27 PM

pyrello:
I'm testing this again.

pyrello (approved):
Looks good. Thanks for sticking with it!

e-marie-w:
Thank you for all the testing @pyrello!

PR #5391, opened 2022-06-22
update core to 9.3.17, updated core patches, update to htaccess

joewhitsitt:


joewhitsitt:
not very familiar with the htaccess patching we have implemented. looks fine, but it keeps generating a .orig version which is tracked in git. i forgot how to keep the change but get rid of the .orig...

joewhitsitt:
Ugh. Might be easier if I wasn't trying to update to 9.4.1... Closing and starting over.

PR #5401, opened 2022-06-27
UI Press updates

### LINKED ISSUE #5400: UI Press - Create sale books page
Original: https://www.uipress.uiowa.edu/sale-books

pyrello:
Resolves #5400 

# How to test
- `ddev blt ds --site uipress.uiowa.edu`
- `ddev drush @uipress.local uli /`
- Visit https://uipress.uiowa.ddev.site/seasons and observe 3 columns and typography change.
- Visit https://uipress.uiowa.ddev.site/series and observe 2 columns and typography change.
- Visit https://uipress.uiowa.ddev.site/subjects and observe 3 columns and typography change.
- Visit https://uipress.uiowa.ddev.site/books/sale. Spot check books to make sure that they are displaying a sale price and that they should be on sale.


pyrello:
I'm noticing what appears to be a discrepancy between the books that are showing and that some are not showing a sale price. This seems to indicate a mismatch between the logic that determines which books show up on the page (via Views) and when a sale price is shown in a book type paragraph (`uipress_core_preprocess_paragraph__book_type`). I suspect this might be because of the set timezone not being used to calculate the `$today` variable in the preprocess logic. However, I haven't been able to get it working yet.

cory-skeers:
Checkin' this out.

cory-skeers:
The series page on prod (https://uipress.prod.drupal.uiowa.edu/series) includes a trimmed description as well, which isn't included in the update--was that intentional? Otherwise, the pages check out, and I'm moving on to checking out the sale books.

pyrello:
> The series page on prod (https://uipress.prod.drupal.uiowa.edu/series) includes a trimmed description as well, which isn't included in the update--was that intentional? Otherwise, the pages check out, and I'm moving on to checking out the sale books.

Yeah, we changed it to use fields instead of the teaser specifically to prevent the term description from showing up.

cory-skeers:
Is the ordering on https://www.uipress.uiowa.edu/sale-books randomized on each page load? That's kinda neat. Without that, I would have expected it sorted by publication date, title, or author, but the content authored date is maybe closer to the original's intent.

Also seeing the timezone discrepancy (sale starting today is showing up on the book, but the book isn't appearing in the view. Moving the sale date to starting yesterday or prior, and it appears). My guess is we could maybe do a query alter hook and manually adjust the timecode on one end or the other to compensate whatever timezone weirdness is probably going on, but that feels like a bandage that shouldn't be necessary. 

The sale code is required for a book to appear, but the code isn't required in order to enter a sale. Should that be sale price that's required instead? (you can also get onto the sales view by entering a sale code but no sale price).

Multiple sales on a book (eg sales for both paperback and ebook) result in multiple appearance in the sales view.

cory-skeers:
Pushed an update to aggregate the results for the case of multiple concurrent sales. Other than that, things are looking good and ready to go other than the sale code requirement question and start time discrepancy.

I was thinking about poking at the time thing, but that could also be a followup task if it was important to get this up and available on the prod site.

cory-skeers:
> Also seeing the timezone discrepancy (sale starting today is showing up on the book, but the book isn't appearing in the view. Moving the sale date to starting yesterday or prior, and it appears). My guess is we could maybe do a query alter hook and manually adjust the timecode on one end or the other to compensate whatever timezone weirdness is probably going on, but that feels like a bandage that shouldn't be necessary.

Should debug to find the exact time adjustment needed, but replacing "today" in the inequality in the view with something like "-6 hours" works. Still a bit of a bandage, but at least it's straightforward and self-contained in the view config. Not sure if 6 hours is exactly right, but assuming it is timezone weirdness, it should be 5 or 6 hours for cdt/cst.

joewhitsitt:
For D7 timezone offset, we have used a helper function on a couple of sites (ITS, Content Hub, etc.) which handles the standard/daylight issue.

```
/**
 * Custom function to convert string to time with proper offset.
 *
 * @param string $time
 *   The string date/time to convert.
 *
 * @return string
 *   Formatted time string.
 */
function content_hub_core_tz_offset($time) {
  $time = strtotime($time) + date_offset_get(new DateTime($time));
  return $time;
}
```

pyrello:
> Also seeing the timezone discrepancy (sale starting today is showing up on the book, but the book isn't appearing in the view. Moving the sale date to starting yesterday or prior, and it appears). My guess is we could maybe do a query alter hook and manually adjust the timecode on one end or the other to compensate whatever timezone weirdness is probably going on, but that feels like a bandage that shouldn't be necessary.

I'm surprised that this is an issue. I would have expected Views to be incorporating the timezone by default. I assumed that is why I was previously seeing a discrepancy between what was showing up and the sale price sometimes not being displayed because we were not account for timezone in the logic to display sale price. If anything, if what you are saying is correct, then I would have expected to never discover the timezone issue with the sale price display in the first place because I think everything would have lined up.

Using -6 hours is incorrect based on my reading about handling timezones because the timezone adjustment can change based on the time of year. It might work correctly in our situation, but it is not considered a best practice.

> The sale code is required for a book to appear, but the code isn't required in order to enter a sale. Should that be sale price that's required instead? (you can also get onto the sales view by entering a sale code but no sale price).

This seems like maybe the logic around these is not quite right. Having a sale date without a sale code doesn't seem right. I'm going to check with the customer about it. If we need to, we could potentially add another filter group and set an or condition where either the code or the price are filled in.

cory-skeers:
> Using -6 hours is incorrect based on my reading about handling timezones because the timezone adjustment can change based on the time of year. It might work correctly in our situation, but it is not considered a best practice.

I think Joe's helper function should be accurate and account for the changes throughout the year, and using it in a query alter hook. Otherwise just doing -6 or -5 hours gets it close, but it won't be 100% accurate.

I'm also surprised it's even an issue with Views, and haven't looked for any drupal.org issues on it yet. Maybe there's something there rather than us sorting it out.

cory-skeers:
Debugging and looking at the query, the time it is using for 'today' is '2022-06-29T19:00:00'. Not sure where that value is being provided. By the time the comparison gets made, that value gets chopped to just '%Y-%m-%d', which is the same format and only value that is stored at the field level.

pyrello:
> Debugging and looking at the query, the time it is using for 'today' is '2022-06-29T19:00:00'. Not sure where that value is being provided. By the time the comparison gets made, that value gets chopped to just '%Y-%m-%d', which is the same format and only value that is stored at the field level.

When you say "Debugging and looking at the query," do you mean the views query? If so, the 'today' value is what I set as the relative date value. I was under the (possibly mistaken) impression that that would give me something like 'Y-m-d 00:00:00'.

cory-skeers:
> When you say "Debugging and looking at the query," do you mean the views query? If so, the 'today' value is what I set as the relative date value. I was under the (possibly mistaken) impression that that would give me something like 'Y-m-d 00:00:00'.

Yep, the views query. I added a query alter hook with a breakpoint, and was looking at the conditions before they're sent to the database using xdebug. It starts as 'Y-m-d 00:00:00' in the Sql.php Views plugin, with the correct date, but somewhere along the line between that and the final query condition, it's ending up as 'Y-m-d 19:00:00 with yesterday's date. I assume, then, that the timezone offset is being applied twice, since the end date is 5 hours behind, which would match CDT if it adjusted (correctly) and then adjusted an extra time.

pyrello:
And here I thought this was going to be so simple, I could just sneak it in as unplanned work :|

From what you are saying it appears like it is applying the TZ offset after setting the date to Y-m-d 00:00:00, which is not what we want. Essentially, for this we want it to ignore the timezone offset because it incorrectly changes our "relative" date.

pyrello:
It seems like this might be at play here: https://www.drupal.org/project/drupal/issues/2647292. I might poke at this a bit later.

pyrello:
To recreate the issue of views relative filter + timezone issue:
1. Follow setup instructions above
2. Add a book: https://uipress.uiowa.ddev.site/node/add/book
3. Set a sale price and code
4. Set the sale start date as the current date
5. View the sale page: https://uipress.uiowa.ddev.site/books/sale
6. Not sure it matters, but just to make sure its not the reason: `ddev drush @uipress.local cr`
7. We are expecting the new book to not show up

cory-skeers:
I'm not sure if the patch effects it or not, but using 'now' as the relative time seems to work correctly, but 'today' still has the strange offset. Still debugging to see if there's a straightforward way to fix this, but when using 'now' the Date.php views filter assigns `$this->calculateOffset = FALSE;` so that it doesn't do any timezone adjustment later, but this doesn't happen when using 'today'.

pyrello:
> I'm not sure if the patch effects it or not, but using 'now' as the relative time seems to work correctly, but 'today' still has the strange offset. Still debugging to see if there's a straightforward way to fix this, but when using 'now' the Date.php views filter assigns `$this->calculateOffset = FALSE;` so that it doesn't do any timezone adjustment later, but this doesn't happen when using 'today'.

@cory-skeers When you say it works, do you mean it works in that the offset doesn't happen? Or that it works in terms of the view works the way it is supposed to?

cory-skeers:
> @cory-skeers When you say it works, do you mean it works in that the offset doesn't happen? Or that it works in terms of the view works the way it is supposed to?

Both. It doesn't apply the offset and keeps an accurate datetime, which gets compared to a date field that doesn't have the time, so should always be treated as midnight for the sake of the <= comparison. So I _think_ it works for all cases, but doesn't answer the question of why 'today' is getting an offset applied to a midnight datetime.

pyrello:
> Both. It doesn't apply the offset and keeps an accurate datetime, which gets compared to a date field that doesn't have the time, so should always be treated as midnight for the sake of the <= comparison. So I think it works for all cases, but doesn't answer the question of why 'today' is getting an offset applied to a midnight datetime.

I'm cool with not having to solve the underlying issue here if we can sidestep it for now. I'll check this again without the patch to see if that if affecting anything.

joewhitsitt (approved):
Brief review. Approval based on Cory's previous testing and Sean's satisfaction

PR #5402, opened 2022-06-28
Attempt to get default content working again

### LINKED ISSUE #5321: Can't edit/replace images on new default homepage template
# Problem
I can't edit/replace any of the images in the banner and card components on the new default homepage setup. 

When you try to add a new field or choose an existing file and press the button "insert selected" you get the get the loader icon and then nothing happens.

Everything is fine if you create a new banner or card but does not work on existing components.

Current site we are editing. https://powwow.prod.drupal.uiowa.edu/

https://user-images.githubusercontent.com/731952/171436009-14f78b85-b001-47f6-ba17-be4c306b0cb5.mov

# Notes from Grooming 6/14/22

Could test if this problem exists on the showcase page... Cory thinks he checked but it's been awhile. Since this is a regression, we can review the changes to see if we notice anything that stands out.

could use helper update function to fix once we identify the problem

button styles not set in default content, could add those if makes sense or create a follow up issue if there is not one already

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5321

Based on https://github.com/uiowa/uiowa/issues/5321#issuecomment-1170227279, https://github.com/uiowa/uiowa/issues/5321#issuecomment-1170408565, I moved forward with repurposing the component showcase to contain the "new homepage" content, taking the opportunity to properly set section and block config along the way. The homepage grid now contains image blocks instead of text areas.

Downside is that the component showcase no longer exists and we still don't know _why_ this was even an issue.

Upside is that the initial bug is fixed, default content has been updated to be lean and mean with little manual tweaks, got rid of a virtual_tour remnant.

# To Test

- Delete the contents of the `uiowa/docroot/sites/default/files`
- `ddev drush @default.local sql-drop && ddev drush @default.local cr`
- `ddev blt drupal:install --site=default`
- `ddev drush @default.local uli`
- The front page should be the /home
- The footer contact block should be present
- Going into to layout builder, you can modify existing media references without encountering the ajax error (new or existing media replacements)
- Content has the proper classes in the wysiwyg, etc.

# ToDo

- Reset percy baseline.


joewhitsitt:
i think i broke the footer at some point in this PR

pyrello:
I'm going to use reviewing this PR to poke at #5321 just a tiny bit more.

composer.lock
```diff
@@ -11000,54 +11004,6 @@
...
-                    "name": "sd9121",
-                    "homepage": "https://www.drupal.org/user/3513915"
-                }
-            ],
-            "description": "Provides an field formatter to Virtual Tour effect on an image.",
-            "homepage": "https://www.drupal.org/project/virtual_tour",
-            "support": {
-                "source": "https://git.drupalcode.org/project/virtual_tour"
-            }
-        },
```
pyrello, line 11050: This is removing the `virtual_tour` module, which I'm assuming was unintentional, @joewhitsitt?

pyrello: Or maybe not, since I'm not seeing it in the `composer.json` file.

pyrello: So, now I'm remembering that we didn't go with `virtual_tour` in favor of doing our implementation, so this is fixing that, I guess.

joewhitsitt: >got rid of a virtual_tour remnant.

Yeah, this is still in main and didn't get cleaned up properly.

pyrello:
I tested this and everything is as expected, except I'm not sure what this means:
> Content has the proper classes in the wysiwyg, etc.

Holding off on approving until I get clarification on that, but otherwise this is good to go.

joewhitsitt:
Good question. By recreating the default content, I was able to save the blocks and such with the proper lb styles and wysiwyg markup including things like https://github.com/uiowa/uiowa/pull/5402/files#diff-9882502a03b6505ceb37063d92665d1034522c9408dc855e21629292bb81d056R27 where we switched the intro/lead style class

pyrello:
Looking into this a bit more, I think we made a change in UIDS a while back that we failed to pull through to our configuration. In @joewhitsitt's link above, the class is being changed from `.element--light-intro` to `.uids-component--light-intro`. This is right based on what we have set in our configuration right now: https://github.com/uiowa/uiowa/search?q=uids-component--light-intro&type=code. However, it is actually our intention to start using `.element--` style classes instead of `.uids-component--`. These classes may change again as part of the updates happening in UIDS 4, so I would suggest that we don't go to the trouble of changing them just yet, so we don't have to potentially change them in an update hook twice. I think everything still works and looks correct with the `.uids-component--` classes, so we let this go ahead for now and fix it in a future issue.

pyrello:
In reviewing the visual changes in Percy, I'm noticing that we seem to be reverting to having our example article be called "Keep Moving Forward!" instead of "Demo Article". I don't have a particular feeling about one versus the other, but this seems like an unintended change based on the PR summary.

The rest of the visual changes seem fine.

bspeare:
I'd like to see this to remain "Demo Article" https://github.com/uiowa/uiowa/blob/db80dfd0cf9c163b837de8afdad647bf6d083fa9/docroot/profiles/custom/sitenow/content/node/b61431c8-dcd6-4a06-9932-78f3bae55c4e.yml#L23.   I think it was a good change that Alex introduced before he left.  I can update that file.

pyrello (approved):

pyrello:
With change back to "Demo Article" this looks good. We have identified some additional work to be handled via #5429 and #5418, so I think this is ready to go.

PR #5440, opened 2022-07-08
Iisc deliverables

### LINKED ISSUE #5406: IISC - Display project deliverable file using human readable name
# Motivation

On the current IISC site, when a deliverable is attached to a project, the user can set a human-readable name and that is what displays. On the new site, we figured out how to bring that human-readable name over in the migration, but I cannot figure out how to get it to display.

# Proposed solution

Figure out the appropriate configuration that will allow us to use the human-readable name when displaying a file media entity reference field.

cory-skeers:
Resolves #5406 

This does also enable the 'description' field for File type media entities on the platform, allowing the description to be used in places where a media field is being displayed. If there is no description (default for any pre-existing media entities with this update), the display will continue to use the filename when displaying the rendered entity.

# How to test

`git checkout iisc-deliverables && git pull`
`ddev blt ds --site=iisc.uiowa.edu`
Create a project with a deliverable which uses a description in the media entity
Check the project display, and see that it displays the description, not the filename

Run the migrations up to at least a few projects
Check a migrated project's display, and see that it displays the human readable name, not the filename

Try syncing some custom split sites like admissions, and check that it completes without config errors
Try opening and editing an existing media entity, and adding new media entity, and check that they save properly with or without a description


cory-skeers:
Gah...accidentally built this off of the uipress branch.

docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/CreateMediaTrait.php
```diff
@@ -83,11 +83,13 @@ public function createMediaEntity($fid, array $meta = [], $owner_id = 1) {
         case 'application':
         case 'document':
         case 'file':
+          // Set the description as either the title or filename.
+          $description = (isset($meta['title'])) ? $meta['title'] : $meta['filename'];
```
joewhitsitt: Can the description just be blank if a title doesn't exist? This is not a required field and media's own logic will display the filename unless a description is provided.

```
if (isset($meta['title'])) {
  $media_entity['field_media_file']['description'] = $description;
}
```

cory-skeers: Hrm...In most cases you're right, and it should be handled like that, but for IISC in particular, the human readable name actually comes in as the filename. That worked out alright when it was just setting a name that was needed anyway, but this is a little different...

cory-skeers: I updated this in...a way? It feels very hackey, but it "works." If we have an actual title, we set it to the description for use in displays (in addition to the name for use in the media browser), and if we don't then we check if the filename gathered in the meta is different from the actual filename, to use that instead. And if it is the same, then we don't bother setting the description.

```
          // If we have a title,
          // go ahead and set it as the description
          // so it can be used in displays.
          if (isset($meta['title'])) {
            $media_entity['field_media_file']['description'] = $meta['title'];
          }
          else {
            // If we didn't have a title, check if we had
            // a human readable filename to use for the description
            // (that doesn't match the true filename, which
            // would be used in displays with an empty description.
            if (!str_ends_with($meta['file_uri'], $meta['filename'])) {
              $media_entity['field_media_file']['description'] = $meta['filename'];
            }
```

joewhitsitt: Hrm. You lost me, but I trust you. Can you update the test instructions to surface one of these instances?

joewhitsitt:
pulling down to test

joewhitsitt (approved):
I got confused in the terminology for title and filename. Looks good.

Tested a migration too. No sites split this config.

Screen Shot 2022-07-11 at 4 18 50 PM

cory-skeers:
> I got confused in the terminology for title and filename. Looks good.
> 
> Tested a migration too. No sites split this config.

Thanks for testing Joe, and for the confirmation on config splitting!

PR #5444, opened 2022-07-11
Move LB styles to their own library, override smooth scrolling

### LINKED ISSUE #5441: Smooth scrolling behavior causes excessive bouncing when editing blocks in Layout Builder
Related: https://github.com/uiowa/uiowa/issues/4647 and https://www.drupal.org/project/drupal/issues/3152561

# Problem
When performing some actions on a block (e.g., removing a block) in Layout Builder, an editor will occasionally get bounced around the page excessively. ~~This is likely due to the introduction of `scroll-behavior: smooth` in [https://github.com/uiowa/uiowa/issues/3688.](https://github.com/uiowa/uids/pull/581).~~ This is more likely an upstream thing with Layout Builder, but our implementation of `scroll-behavior: smooth` may be exacerbating/highlighting the issue.

### Removing a block

https://user-images.githubusercontent.com/472923/178061768-6f75feb8-9d4d-49b2-920d-32888244e1a8.mov

### Adding a section 

https://user-images.githubusercontent.com/472923/178064836-4afea861-ade2-40e5-a2df-877c7dff6c25.mov

### Editing a block

https://user-images.githubusercontent.com/472923/178065185-2b53c188-ac93-4350-9c04-afafc5d8bf27.mov

# Steps to reproduce
1. Remove a block that's far enough down the page that scrolling can occur, watch the page bounce up and down

OR 

1. Configure a block that's far enough down the page that scrolling can occur

OR 

1. Add a section far enough down the page that scrolling can occur

# Proposed solution
It appears that Layout Builder isn't expecting `scroll-behavior: smooth` to be enabled when jumping to several anchors in a row as part of its process when moving or removing a block. If this `scroll-behavior` CSS was overridden or disabled while an editor is logged in and/or editing a page, the site editor wouldn't see the actual scrolling to/from anchors, and their browser would seemingly be placed instantaneously where it should be, as was the case before `scroll-behavior: smooth` was added

While at the same time, a typical user who wasn't logged in could see the desired smooth scrolling behavior.

This could be accomplished by adding a CSS class that disables this behavior to `<html>` via some conditional Javascript that executes only when a user is logged in.

Testing solution via the web inspector tool reduces some of the bouncing around, but the scrolling to the top is a bit jarring. The above solution might be a partial remedy.

pyrello:
Related to #5441 but does not completely resolve it. Includes a refactoring of the code that sets a variable indicating whether LB is active for fragments and nodes.

Resolves https://github.com/uiowa/uiowa/issues/5441. 

# How to test
- `ddev blt ds`
- `ddev drush @default.local uli /`
- Use inspector to check that `layout-builder-enabled` class is added to the `body` element.
- Use dev tools Network tab to check that `layout-builder.css` file _is not_ loaded.
- Click the "Layout" tab.
- Use dev tools Network tab to check that `layout-builder.css` file _is_ loaded.
- Visit https://default.uiowa.ddev.site/news and confirm that `layout-builder-disabled` class is added to the `body` element.


pyrello:
One thing I noticed that I could not immediately figure out was that there is extra padding being added to the `li` elements in the local task menu after the changes in this PR. It seems like there is some additional class setting that I am not finding that needs to be moved from `layout-builder.scss` to `admin.scss`. Feel free to tell me what I'm missing :)

quamsta:
@pyrello Checking on this..

quamsta:
@pyrello .. Not seeing the extra padding that may be coming from `layout-builder.scss`:

(This video compares the local actions on a local dev site and the prod Mental Health site after pulling this branch and running all of the frontend commands)

https://user-images.githubusercontent.com/472923/178507812-6b877a9c-e99d-428a-88b2-d4a593bff1f6.mov


Can you show a screenshot of the extra padding you're seeing?

I'm also not seeing the `layout-builder-disabled`/`layout-builder-enabled` classes applied to the `<html>` element, but I am seeing `layout-builder-enabled` applied to the `<body>` tag no matter what (even if you're logged out), but that may be something you're still working on.

bspeare:
@pyrello can you post a screenshot of "extra padding being added to the li elements" because I'm not seeing that either.

pyrello:
> @pyrello can you post a screenshot of "extra padding being added to the li elements" because I'm not seeing that either.

I'm not in a position to switch back to this PR at the moment. It seems like maybe there is some some of mismatch in what I was doing locally and what I wrote in the test instructions. It may be the case that there is no issue with the local tasks and there was an issue with my local CSS.

pyrello:
Okay. Not seeing the padding on local tasks now. Also, looks like I misidentified the element that the LB classes should be added to, which is in fact the `body` element. In looking at the demo site in production, I'm seeing the same behavior where `layout-builder-enabled` is getting added even in cases where it is not active, so I'll look into that.

pyrello:
In reviewing the code that adds the `layout-builder-enabled` variable, it seems that how we are checking if LB is enabled may not be meaningful in the same way it was a long time ago. It is checking for whether LB is enabled for the content type rather than whether the actual LB interface is being loaded. So, it currently loads for all nodes no matter what because all our content types have LB enabled, even though they don't all allow per node customization. Views pages end up with the `layout-builder-disabled` class.

I can't really remember what we set this up for, so maybe this is okay and working as expected. It certainly hasn't changed from what is going on in production at least. @bspeare Do you have any additional insight?

bspeare:
We are using the variable here to set different container classes: https://github.com/uiowa/uiowa/blob/b2684f186643f02b3cdfd64e3894d41664a5fec1/docroot/themes/custom/uids_base/templates/layout/page.html.twig#L53  The `page__container` class is our fixed container class that will set all views pages to have the same default container width.   The `content__container` class is set up to target the different sizes of containers set within layout builder.   It looks like it is still working as expected.

pyrello:
So, I guess the thing that remains to be done with this PR is to find a condition for loading our `layout-builder` library only when the LB tool is active.

pyrello:
> So, I guess the thing that remains to be done with this PR is to find a condition for loading our layout-builder library only when the LB tool is active.

I found a way to do this.

bspeare:
I can finish reviewing this today.

bspeare (approved):
Tested all the steps and this is working as expected.

bspeare:
Took a deep dive into this to find out what was causing https://github.com/uiowa/uiowa/issues/5441 and I think I found the CSS that was creating this problem in https://github.com/uiowa/uiowa/pull/5444/commits/d5e423de682fd7d8cc0184a4807efb6c85f9b0f6.   Marking this as a draft so we can get this reviewed for to resolve https://github.com/uiowa/uiowa/issues/5441 .

quamsta:
@bspeare Tested this branch and we're no longer bouncing around at all! Fantastic.

@pyrello Sorry for setting up a wild goose chase here, `scroll-behavior` was sort of a red herring...

bspeare:
This can probably be merged unless anyone else wants to test that this resolves https://github.com/uiowa/uiowa/issues/5441.   I have tested it and @quamsta has also tested it.

joewhitsitt:
This is a HUGE improvement. Thanks! Still going to hope for further core improvements to this editing experience.

bspeare:
@joewhitsitt you are right that there is a slight jitter still.  I might take a final look and see if we could remove that as well.   
It should be as smooth as this umami demo:

https://user-images.githubusercontent.com/1036433/179093766-c7cf61f3-f572-40c6-8c61-d37f8951eb5a.mov

bspeare:
I think we can merge this in now.  I looked a little further into this and we could override the dialog transition as well if we want to remove any movement and just have the tray appear. 

```
.dialog-off-canvas-main-canvas {
    transition-duration: 0s !important;
    transition-delay: 0s !important;
}
```

PR #5452, opened 2022-07-12
Temporary hotfix for entity_print + dompdf

pyrello:
# How to test

- `ddev blt ds --site admissions`
- `ddev drush @admissions.local uli kiosk`
- Spot check AOS PDFs for consistency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ addons:
     - svn-14671.prod.hosting.acquia.com
     - svn-29892.prod.hosting.acquia.com
     - svn-37631.prod.hosting.acquia.com
+    - git.drupal.org
 
 before_install:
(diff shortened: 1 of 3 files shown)


pyrello:
I guess how I have this setup isn't going to work because Drupal expects keys or authentication from an issue fork branch.

pyrello:
I thought I had found a way around Travis not being able to get this, but it seems like we need SSH keys set up that would allow Travis to clone the repo. There are existing SSH keys, but they are already saved and encoded in Travis. Rick set this up, so I'm not sure if we passed the actual keys on to anyone or if they were set up and then discarded because they were never meant to do anything but provide the connection for Github.

bspeare:
Tested out the PDFs and they look the same.

bspeare (approved):
Tested and PDFs look the same.

PR #5479, opened 2022-07-22
UI Press sale updates

### LINKED ISSUE #5468: UI Press - Sale price logic is still not quite right
Sale price should be showing up on this book but is not: https://uipress.prod.drupal.uiowa.edu/books/american-pipe-dream

cory-skeers:
Resolves #5468 

# How to test

`git checkout uipress-sale-update && git pull`
`drush @uipress.local uli /books/american-pipe-dream`
^ The paperback should show up with a sale price


cory-skeers:
@pyrello I think this should take care of the odd missing sale prices, but doesn't do anything to address possible page caches. Want to try and fit that in here too?

pyrello:
> @pyrello I think this should take care of the odd missing sale prices, but doesn't do anything to address possible page caches. Want to try and fit that in here too?

I would be interested in seeing if it could be solved relatively easily. Maybe invest a little time into that and then check in with the team about whether to keep going? For launch, we can probably work around it

docroot/sites/uipress.uiowa.edu/modules/uipress_core/uipress_core.module
```diff
@@ -141,25 +142,31 @@ function uipress_core_preprocess_paragraph__book_type(&$variables) {
...
         // If neither condition is true unset display of sale price.
         if (!$in_range) {
           unset($variables['content']['field_book_sale_price'], $variables['content']['field_book_sale_code']);
         }
+        // If we are in range of a sale, invalidate the paragraph cache
+        // to make sure we're updating the display.
+        if ($in_range) {
+          $tags = $paragraph->getCacheTagsToInvalidate();
+          Cache::invalidateTags($tags);
+        }
```
cory-skeers: Would love to go over this or other options @joewhitsitt. I think this'd get the job done (make sure the paragraph is being rendered fresh if a sale has started), but feels off to me.

pyrello: I'm a little uncertain about this approach. I'm trying to reason about when it will be triggered and it seems like it would be triggered when the cache for the paragraph is already invalidated, since the preprocess function should only run when the paragraph needs to be re-rendered because the cache is invalid. I may be wrong about this assumption.

cory-skeers: Eh...I think you're right about that. I'd already felt off about it, so I think it definitely needs to be at least moved elsewhere, if not handled differently.

cory-skeers: What I was/am hoping to avoid is invalidating the cache for all books that may or may not have sales, even if it's somewhat mitigated by being handled through a cron job. But the more I think about it, the more I'm unsure about how to guarantee that the sale logic will trigger this invalidation. Maybe I'm just overthinking it and getting confused.

joewhitsitt: Cron is going to be your trigger. That is pretty much all we have right now, however, it can be a uipress specific hook_cron (e.g. every morning) that does a entityQuery to only get nodes that meet the condition and clear those.

cory-skeers: > Cron is going to be your trigger. That is pretty much all we have right now, however, it can be a uipress specific hook_cron (e.g. every morning) that does a entityQuery to only get nodes that meet the condition and clear those.

Okay! That makes a lot of sense, and sounds like what I was trying to get to but couldn't quite grok.

joewhitsitt: Have a look at https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_core/uiowa_core.module#L38. It technically runs every time, but you can set a different time interval and instead of the hourly cache tag you can queue up a list of nodes to clear instead which should cascade down and clear blocks.

cory-skeers:
I think this has been updated to handle the cache clearing. It:

- Triggered by cron, completes once a day
- Finds all nodes that have a sale active today, or that has ended yesterday (so that if a sale is no longer active today, but was yesterday, we clear it as well)
- Invalidates the nodes' tags and resets their cache
- Queues the nodes' tags to be purged (where `acquia_purge` is enabled)

Running it locally, it appears to work correctly, BUT seems to create a cache tag for every node on the site EXCEPT for the ones that are invalidated, which...does not seem good.

joewhitsitt:
>seems to create a cache tag for every node on the site EXCEPT for the ones that are invalidated, which...does not seem good.

not sure i understand

looking at this so far, I suspect resetCache does too much in this situation. 

>Content entities have both an in-memory static cache and a persistent cache. Use this method to clear all caches.

The only other place we are using resetCache in our code has to do with unmanaged files and if I recall correctly was a last resort

joewhitsitt:
invalidateTagsOnSave (e.g. node save) seems to just Cache::invalidateTags and getCacheTagsToInvalidate fires a couple times getting the paragraph tags, node tags and moderation tags:

Screen Shot 2022-07-26 at 11 33 42 AM
Screen Shot 2022-07-26 at 11 34 09 AM
Screen Shot 2022-07-26 at 11 35 24 AM

So if it is an issue that `node:123` is not enough, maybe the paragraph stuff should be included as well?

joewhitsitt:
Alternatively, I wonder if the preprocess_paragraph could use Cache::mergeTags() to include the `node:123` so that the paragraph preprocess is triggered.

cory-skeers:
> > seems to create a cache tag for every node on the site EXCEPT for the ones that are invalidated, which...does not seem good.
> 
> not sure i understand

Little misuse of terms there on my part, I think. When testing I was looking at data in the `cache_entity` table in the database before and after triggering cron, and the nodes that were desired to be cleared no longer had data there, but every other node not included in that list then had data in that table, despite the vast majority of them not being in there prior to triggering cron.

> looking at this so far, I suspect resetCache does too much in this situation.
> 
> > Content entities have both an in-memory static cache and a persistent cache. Use this method to clear all caches.
> 
> The only other place we are using resetCache in our code has to do with unmanaged files and if I recall correctly was a last resort

That is very likely too much then, and came from some of my previous efforts trying to clear up memory in migrations.

cory-skeers:
> Running it locally, it appears to work correctly, BUT seems to create a cache tag for every node on the site EXCEPT for the ones that are invalidated, which...does not seem good.

I think this is more to do with where I'm trying to look to see what is cached or not. I tried a fresh sync without the new cron hook, and triggering cron has the same behavior but without having removed the nodes of interest.

joewhitsitt:
ah yes, i see now. yeah, i just ran cron on another branch and got the same thing. you beat me to it

joewhitsitt:
i think this should be pushed to DEV with last run set to null and that resetCache removed. then we can test cache hits on sale nodes mirroring PROD to see if that is enough

joewhitsitt:
@cory-skeers if you add the last run back in, I will approve and we can live test

cory-skeers:
Got it added back in--thanks @joewhitsitt!

joewhitsitt (approved):
From what I have tested, this is an improvement and sets a foundation for proper cache clearing. What is unknown is whether or not it does enough to clear caches when needed. Since develop is in a weird state, testing while live seems like a good step forward.

PR #5497, opened 2022-07-28
Events block date range

### LINKED ISSUE #4320: Add smart date module to default platform
# Motivation

We have smart date as a dependency for multiple feature splits. This means that we have multiple copies of the smart date formats that are used in those feature splits. Adding this module to the default platform would allow us to deduplicate those formats and prevent possible configuration collisions between different features.

# Proposed solution

* Enable smart date module
* Update default config

### LINKED ISSUE #5187: Add a date range option for events block
# Story

As a site editor I want to specify a date range to pull events based on in addition to displaying past or upcoming events.

# Proposed solution

- Duplicate `past` view display and name it `reverse` on content.uiowa.edu (heads up, might be packaged in features)
- Create a select or radio that allows Sort to be set to Descending or Ascending (defaults to Ascending for upcoming)
- Add a select, "Events query type"? for specifying upcoming (default), past or specific date range which if selected shows the specific dates field.
- https://github.com/uiowa/uiowa/issues/4320
- Add a date field (collect start and end) with date picker - see issue comments about using smart date, getting values, limiting date picker range.
- Update hook to take the past checkbox boolean value on existing to set "events query type" to past and the sort field to descending.
- Validate on all this to prevent dates that extend beyond our min and max (01-01-1970, 19-01-2038).
- Based on query type, take the dates or type and convert/format them to work as parameters.
- Based on sort, change from events to reverse.

# Follow-up

Remove the past boolean field

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5187
Resolves https://github.com/uiowa/uiowa/issues/4320

# Test

- `ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu`
- Inspect https://sandbox.uiowa.ddev.site/node/686 to make sure they are showing upcoming and past blocks accurately and that min and max attributes on the input elements are set (which limits the date picker).
- Sync admissions and grad to make sure event and thesis defense content types import correctly and the smart date functionality still works.
- The past date checkbox on the v3 events block should be hidden and there should now be three fields to replace it (timeframe, sort, date range (hidden until date range is selected from timeframe).
- Check that the date range field is appropriately displayed with no overlap of the elements.
- Field validation should occur on the date range where it requires values if timeframe:date range is selected and that the values themselves do not extend the min and max.
- Help text should make things relatively clear though I understand Asc/Desc is complicated in a upcoming/past/mix context.
- Inspect code to confirm that it is limited to v3 functionality and does not affect v2 paragraphs logic.

# Todo

- [x] Validation - require and within safe limits
- [x] Limit date picker to min and max. Seems to require min/max attributes being present on the widget.
- ~~Adjust the logic associated with the past date update hook. - It works in situations where the field is set and has a value, but we thought we should check for instances where the field is not set (neither true/false). If not set, it should also be set to upcoming. https://github.com/uiowa/uiowa/pull/5497/files#diff-fead566a578ea17efbda9b82e2a399962f8d578c39998fedf5bb87ceebd7d3f6R2139~~ _since we are checking for 1, and covering 0 and null with the else, no separate check is necessary._
- [x] Set date range field during update hook or set min/max attributes if not present.
- [x] Form styles for date range field - right now they don't display with enough spacing between elements. @makurj 
- [x] Help text for the new fields (sort, timeframe and date range)
- [x] Clean up smart date config from feature and site splits - check for sites that have smart date split, remove the declaration from the split, remove the smart_date configuration unless it is different than what is in default. if different, a partial export is needed.
- [x] Remove out of date `@todos`, convert any to separate issues that can't be removed.
- [x] If https://github.com/uiowa/uiowa/pull/5501 is merged first, remove smart date from iowa now site split.

# Follow-up
- https://github.com/uiowa/uiowa/issues/5521
- https://github.com/uiowa/uiowa/issues/5518


joewhitsitt:
The update hook issue appeared to be a gremlin in the db and flipped logic (0 instead of 1) for setting the timeframe and sort fields.

Should check for whether past field exists/not set.

e-marie-w:
I pulled down the branch and did the test instructions, the min and max values are showing up correctly for me!

Edit: I do still have to save the block configuration with a date range, and then go back in and inspect to see these values. Unsure if this affects what we're trying to do?

joewhitsitt:
@e-marie-w I saw your comment Friday night and debugged this AM. Looks like it traces to docroot/core/lib/Drupal/Core/Datetime/Element/Datetime.php L248 where it looks for a date before adding those attributes...

So setting a default value (config and update hook), validating on the min max and requiring when selected should cover _most_ of our situations. An editor could gut the values, change selection and save it to later run into a limitless datepicker, but the validation would still catch it.

Alternatively, we could see when layout_builder_custom runs and add them there if they aren't already present.

makurj:
Pulling down.

makurj:
Still working on styling (need to make it responsive). Wanted to push up what I had.

joewhitsitt:
@makurj docroot/themes/custom/uids_base/scss/admin/_off-canvas.scss has other form element manipulation like the statistic prefix/suffix stuff. Maybe you should follow your first hunch and move it there. https://iowaweb.slack.com/archives/C0164FNN1PV/p1659384419161259

GaryRidgway:
Pulling down to test

joewhitsitt:
Just realized there is a small task to remove smart date split from now.uiowa.edu since it merged first. That should not affect the rest of testing however. I'll see if I can do that soon.

GaryRidgway (approved):
Looks great!

PR #5500, opened 2022-07-29
adding button block starting point

### LINKED ISSUE #2873: Create button block

GaryRidgway:
Tracks https://github.com/uiowa/uiowa/issues/2873

##To test
Check out branch
Sync a site
Make a button block in Layout builder
Try all the styles
Profit????


docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -794,6 +797,12 @@ function _layout_builder_custom_process_element(array $element, FormStateInterfa
 
       break;
 
+    case 'inline_block:uiowa_button':
+      // Add some things here.
+      $element["field_uiowa_button_link"]["widget"][0]["title"]["#maxlength"] = 25;
+      $foo = 'bar';
```
bspeare: Remove `$foo` here?

GaryRidgway: not anymore!

bspeare:
Something like https://github.com/uiowa/uiowa/issues/2670 would be useful for these blocks to prevent stacking. 

Screen Shot 2022-08-01 at 2 44 01 PM

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -794,6 +797,10 @@ function _layout_builder_custom_process_element(array $element, FormStateInterfa
 
       break;
 
+    case 'inline_block:uiowa_button':
+      $element["field_uiowa_button_link"]["widget"][0]["title"]["#maxlength"] = 25;
```
pyrello: Please change to single quotes :)

docroot/themes/custom/uids_base/templates/uids/button.html.twig
```diff
@@ -0,0 +1,10 @@
+{% set attributes = attributes ?: create_attribute() %}
+<div {{ attributes }}>
+  {% block button_content_start %}
+    {{ title_prefix }}
+    <a {{ button_classes }} href="{{ button_link_url }}">
+      {{ button_link_title }}
```
pyrello: I'd be interested in moving away from having `button_`-prefixed variables names, at least in the context of this file.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
GaryRidgway: can do! I'm sorry lol, when i cope th path from PHPStorm it does doubles!

docroot/themes/custom/uids_base/templates/uids/button.html.twig
GaryRidgway: I will try and refactor!

GaryRidgway: This hsa been achieved!

GaryRidgway:
> Something like #2670 would be useful for these blocks to prevent stacking.
> 
> Screen Shot 2022-08-01 at 2 44 01 PM

I totally agree, any opinions on proceeding? probing now? requesting sprint prioritization? waiting and asking?

bspeare:
I added a `block` class to the wrapping `div` on [this ](https://github.com/uiowa/uiowa/pull/5500/commits/e05c3f818167fe364af61a5707d5bf339451b437) commit so that it gets this property applied to it: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss#L40.  Ideally, we should just pick a class we can use for all of our blocks so that we can reduce this line of code and use it in future work like https://github.com/uiowa/uiowa/issues/2670#issuecomment-1201694625.

makurj:
Pulling down to test.

makurj:
Should I test all content blocks with all button styles, or 1 content block with all button styles?

GaryRidgway:
> Should I test all content blocks with all button styles, or 1 content block with all button styles?


Just the button block
Add block > button

makurj (approved):
Seems to work as expected. Tried all styles with a few different background and section options.

Issue #5474, opened 2022-07-21
Transition Iowa Now to v3


briand44:
# Story

As an Iowa Now site owner, I would like to transition to SiteNow v3 and create a faculty/staff information hub so that we can share information on the latest SiteNow platform and serve as the go-to resource to for faculty/staff in addition to existing audience.

# Proposed solution

- [x] https://github.com/uiowa/uiowa/issues/5335
- [x] https://github.com/uiowa/uiowa/issues/5283
- [x] https://github.com/uiowa/uiowa/issues/5403
- [x] #5485
- [x] #5524
- [x] #5562 
- [x] #5575
- [x] #5626
- [x] Person content type development to support migrating "Person and Contact" content
    - [x] https://github.com/uiowa/uiowa/issues/5675
    - [x] https://github.com/uiowa/uiowa/issues/5670
- [x] #5821
- [x] #5507
- [x] https://github.com/uiowa/uiowa/issues/3561
- [x] #6025
- [x] #6147
- [x] #6150 
- [x] #5887
- [x] https://github.com/uiowa/uiowa/issues/6093
- [x] https://github.com/uiowa/uiowa/issues/6196
- [x] https://github.com/uiowa/uiowa/issues/6197
- [ ] ~Communication with campus and/or other tasks in regards to https://github.com/uiowa/uiowa/issues/6093~ - Iowa Now folks will handle this communication.
- [x] #6423
- Not planning to launch until Fall so they can shop it around

pyrello:
> Migrate photo galleries as basic pages with image block link to the article using the related content. Maybe use first gallery image as the featured image for teaser references?

I think I have mentioned this before, but I am a fan of the idea of either a) changing the current field_image field on the article content type to allow multiple images or b) adding a new field for additional images. In either case, I would then present the additional images as a thumbnails under the featured image. When clicking the thumbnails, whatever our current lightbox solution is would pop up and show the larger image and controls to view the next image.

It would be very similar to the current experience: https://now.uiowa.edu/2022/07/community-assists-iowa-research-during-bioblitz. Also I think it would be a good addition to the entire platform.

bspeare:
I'd like to see something done within the Articles page for a photo gallery as well.   Doing something like what @pyrello suggests above or turning on layout builder to use the image block seem like they would provide better value/UX.

pyrello:
> Doing something like what @pyrello suggests above or turning on layout builder to use the image block seem like they would provide better value/UX.

[...]

Issue #2691, opened 2020-12-09
Issue with zero duration smart date range field


pyrello:
# Problem
Tracking an issue that was filed at https://www.drupal.org/project/smart_date/issues/3187574

# Steps to reproduce
* Add a Smart Date Range field to a content type
* Set "Allowed durations increments" to just one row with the value of 0.
* Set "Default duration" to 0.
* Save the field settings
* Create a piece of content using the content type

# Proposed solution
Wait for update at the linked issue.

joewhitsitt:
Is this still an issue then @pyrello as it relates to https://github.com/uiowa/uiowa/pull/5014?

pyrello:
I don't think it is framed correctly, but it is tracking an issue that is still open and is actually still an issue. Should we ever decide to adopt smart date for other things, we may need to fix it. Currently it only seems to affect Grad College.

joewhitsitt:
Resolved by https://github.com/uiowa/uiowa/pull/7078

PR #5501, opened 2022-07-29
now.uiowa.edu Article Type site split

cory-skeers:
Relates to the first slice of #5474 

# How to test

`git checkout now-news && git pull`
`blt ds --site=now.uiowa.edu`
`drush @now.local en now_core smart_date`
`drush @now.local cim --source ../config/sites/now.uiowa.edu --partial`
`drush @now.local cim`
`drush @now.local cr`

Try adding some new articles with the various article types. Article Type should be required, defaulting to News Feature.

- `In the News` adds an "Original Publication Date" field
- `Achievement` currently matches the default Article type
- `News Feature` adds an "Embargo Message" field
  - If there is content in the Embargo Message field, the news feature should not be able to be published.


cory-skeers:
Setting back to draft for a couple updates, accounting for using smart_date, hide embargo message field on display, possibly update original publication date display.

bspeare:
@cory-skeers tried this out and got this message at the top of the node add screen: `Notice: Undefined index: #type in Drupal\Core\Form\FormHelper::processStates() (line 211 of core/lib/Drupal/Core/Form/FormHelper.php).`

cory-skeers:
> @cory-skeers tried this out and got this message at the top of the node add screen: `Notice: Undefined index: #type in Drupal\Core\Form\FormHelper::processStates() (line 211 of core/lib/Drupal/Core/Form/FormHelper.php).`

Thanks for the heads up! I think some form config didn't get updated correctly, so then it was trying to apply `#states` to disabled fields. Should be fixed for now, but will need to check for that again after updating the original publication date field.

cory-skeers:
> > @cory-skeers tried this out and got this message at the top of the node add screen: `Notice: Undefined index: #type in Drupal\Core\Form\FormHelper::processStates() (line 211 of core/lib/Drupal/Core/Form/FormHelper.php).`
> 
> Thanks for the heads up! I think some form config didn't get updated correctly, so then it was trying to apply `#states` to disabled fields. Should be fixed for now, but will need to check for that again after updating the original publication date field.

This is still an issue on import. The form display config looks correct on export on initial skim, but on import the new fields are disabled and need to be manually updated.

joewhitsitt:
>This is still an issue on import. The form display config looks correct on export on initial skim, but on import the new fields are disabled and need to be manually updated.

I am not seeing config related to 'core.entity_form_display.node.article.*' in this PR. is this just on your local?

cory-skeers:
> I am not seeing config related to 'core.entity_form_display.node.article.*' in this PR. is this just on your local?

`config_split.patch.core.entity_form_display.node.article.default.yml`? I'd set it up as a partial split, thinking it would only need to adjust weights and add the new fields.

joewhitsitt:
~~right but the resulting config (patch or full) is not present in the PR~~

Update. sry. gosh. my bad.

https://github.com/uiowa/uiowa/pull/5501/files#diff-a18715ef3d4dacfebb5337ca35b0aaca8baa88a3e7c3f8b14eec6ab1f4902fc4

joewhitsitt:
@cory-skeers i made some tweaks and updated the test instructions. had to run cim again after the partial import

cory-skeers:
> @cory-skeers i made some tweaks and updated the test instructions. had to run cim again after the partial import

Thank you @joewhitsitt! Tried it out with your adjustments and worked like a charm.

bspeare:
@cory-skeers is it best practice to have a `.install` file in the module?  I notice some sites have it and some don't.

config/default/core.entity_form_display.node.article.default.yml
```diff
@@ -113,7 +114,9 @@ content:
     type: metatag_firehose
     weight: 6
     region: content
-    settings: {  }
+    settings:
```
bspeare, line 117: What does this new addition add?  Just curious.

cory-skeers:
> @cory-skeers is it best practice to have a `.install` file in the module? I notice some sites have it and some don't.

That's a good question...I tend to include it as a personal practice, but not sure if there is a best practice. I might do some digging around d.o to see if there's a common community practice to follow.

bspeare (approved):
All the testing steps worked for me!

briand44:
> > @cory-skeers is it best practice to have a `.install` file in the module? I notice some sites have it and some don't.
> 
> That's a good question...I tend to include it as a personal practice, but not sure if there is a best practice. I might do some digging around d.o to see if there's a common community practice to follow.

I doubt it's needed unless there is something happening during install but that is just my guess.

config/default/core.entity_form_display.node.article.default.yml
joewhitsitt: this was just a re-export of default. 🤷

joewhitsitt:
@cory-skeers i saw the smart date unset stuff. is that really needed? grad college has a thesis defense date inline widget with duration 0

Screen Shot 2022-08-03 at 3 10 08 PM

I guess we never solved https://github.com/uiowa/uiowa/issues/2691 before adding more implementations...

cory-skeers:
> @cory-skeers i saw the smart date unset stuff. is that really needed? grad college has a thesis defense date inline widget with duration 0
> 
> Screen Shot 2022-08-03 at 3 10 08 PM
> 
> I guess we never solved #2691 before adding more implementations...

It's a little goofy, but the way I went about it for this case where only a date is needed, was to use the date only range widget and hide the end date (and the extra "to" that gets displayed with it). 

image

I don't think it's absolutely necessary, but with the current implementation of smart_date, it seems like a cleaner editor interface without preventing us from switching widgets later.

I'll add a `@todo` that references the issue so it doesn't get lost.

Issue #5505, opened 2022-08-01
SSL certs for uiowa03


bspeare:
# Certs for uiowa03

- *.its.uiowa.edu / its.uiowa.edu
- svp.uiowa.edu
- bijou.uiowa.edu 
- scope.uiowa.edu 
- dancemarathon.uiowa.edu 
- lectures.uiowa.edu 
- offcampus.uiowa.edu
- archaeology.uiowa.edu

bspeare:
I have updated this issue to be a general issue for new certs needed for uiowa03.

briand44:
This is complete

PR #5519, opened 2022-08-02
Feature: 8/2/22 provisions

bspeare:
# How to test


joewhitsitt (approved):
Number of files checks out. All look in order. Will you capture the SSL tasks for these? We've dropped the ball regarding IMU site certs in the past.

bspeare:
Thanks @joewhitsitt!  I updated https://github.com/uiowa/uiowa/issues/5505 with all the SSL stuff.

Issue #4767, opened 2022-02-07
As a site owner, I can have a row of buttons that appears fixed to the bottom of the browser


bspeare:
# Story

As a site owner, I can have a row of buttons that appears fixed to the bottom of the browser when scrolled

As seen on https://viewbook.admissions.uiowa.edu/. 

<img width="1730" alt="Screen Shot 2022-02-07 at 8 29 09 AM" src="https://user-images.githubusercontent.com/1036433/152807179-a57a2b9b-f664-4b61-8960-e3c786257793.png">

# Proposed solution

## First slice
- [x] Blocker: #2873 
- [x] Add a "After content" region to our theme to handle the placement of these region fragments. The exact placement of this is still up for debate so it may change, but we'll move forward with the assumption that this new region will be place in between the end of the content and the Pre-footer region.
- [ ] Add a set of reference fields (one per region that can be overridden) that directly references the region fragment type. We could also potentially allow people to create alternate region fragments inline.
- [x] Update the SiteNow Regions settings page to account for multiple regions.
- [ ] Update the RegionContentBlock class to check if a page is a node and has override fields set and not empty, in which case it would pull the appropriate fragment from the node instead.
- [x] Place an instance of the RegionContentBlock in our new theme region.
- [x] Set up custom onecol layout
- [x] https://github.com/uiowa/uiowa/issues/4622
- [x] Set "Default" as the default container size. 

## Second slice
- [ ] Add a global setting to allow displaying the new region as a fixed area at the bottom of the view port.
- [ ] Implement final design.
- [ ] Optional depending on how the previous slice gets rolled out: Update hook to allow setting the global setting to false for any sites that started adding region fragments to this new area.

<details>
<summary>Original proposed solution</summary>
- [x] Investigate possible solutions for this, documenting the pros/cons of each solution for both the developer and the customer. - completed 06/29/2022
- Create a new button row block that can only be placed in one column sections?
- Allow user to check a box to make the block sticky?
</details>

joewhitsitt:
- Concerns about taxonomy (tag) driven solution for this.
- This doesn't seem like it should fit within the normal layout interface since it is fixed bottom (needs to break out of main content container). Also how to prevent multiple?
- Mobile concerns (disappear, condense, only show first, etc.?)
- We do have a button block in our backlog but not necessarily a button row - https://github.com/uiowa/uiowa/issues/2873
- Probably want to limit the number of buttons, little can do about button text length
[...]

PR #5545, opened 2022-08-05
Admin pre-footer column options

### LINKED ISSUE #4622: Two and three column layouts available in prefooter are broken
# Problem
As pointed out by @bewoldt Admin users are able to add 2 and 3 column layouts to the prefooter, however, none of the section settings or styles are working. 

# Steps to reproduce
Try adding a 2 or 3 column layout to a prefooter and adjust the section container or spacing. 

Screen Shot 2022-01-05 at 11 29 06 AM

# Proposed solution
Fix the layouts so that they work or remove them as an option.

GaryRidgway:
Fixes Admin pre-footer column options, you can now properly use spacing and styles on them.

## To test
Sync sandbox
You cant add any new sections other than one column? Even as an admin?
This means this is working


bspeare:
This works great! Nice that it was just a name change issue. 

I did notice one issue that previously existed. Once an admin adds a two or three column layout, then an editor/webmaster can go and add unlimited two and three column layouts.  Unless we have a permissions workaround for this, I think we should just remove two and three column layouts from the pre-footer region now that we know that this isn't a blocker for https://github.com/uiowa/uiowa/issues/4767.

bspeare:
It looks like there would be no data loss either if we removed the two and three column options.   If an admin had created those layouts they will continue to be able to edit those.

makurj:
Pulling down to test.

makurj (approved):
As an admin, I can add only a one-column section to an existing pre-footer.

PR #5552, opened 2022-08-05
Article gallery

### LINKED ISSUE #5524: Article photo gallery
As a sitenow editor, I would like to be able to add multiple images to an article and display them in an image gallery so that visitors can easily navigate all related photos without having to go to a separate page.

# Proposed Solution

- [ ] Create another image field in addition to the featured image that allows unlimited number of images
- [ ] Output the images below the article body using existing gallery styles with a link to the anchor (https://github.com/uiowa/uiowa/issues/5524#issuecomment-1203211250).
  - Use either these styles (https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/image-gallery-block.scss#L12-L35) and refactor to use a more generalized class name
  - Or use https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/grid.scss#L58-L67
- [ ] Get feedback on placement

cory-skeers:
Resolves #5524 

# How to test

- `git checkout article-gallery && git pull`
- `blt frontend`
- Sync a site, check an existing article display, and see that there is no gallery image link or any other changes when viewing the page
- Add an article/edit an existing article, and add some gallery images
- Check the display of the page, try the gallery jump link, play with the gallery, etc.

## Additional config checks

Some sites are splitting Articles and need to be tested that 1) the new field appears in the edit form and 2) the gallery and its jump link appear correctly on display and 3) nothing from their respective splits is broken by these additions.

[The following sites may be splitting Articles in some way, and should be tested that nothing breaks or is out of place](https://github.com/uiowa/uiowa/issues/5524#issuecomment-1205370370)

- IISC (check our changes)
- SPPA (check our changes)
- HR (check that no changes are needed)
- Admissions (check that no changes are needed)
- Now (check that no changes are needed)
- others?

(for now.uiowa.edu, the split is not yet enabled in prod, so the following steps are needed to enable its split:
```
blt ds --site=now.uiowa.edu
drush @now.local en now_core
drush @now.local cim --source ../config/sites/now.uiowa.edu --partial
drush @now.local cim
drush @now.local cr
```
)


docroot/modules/custom/sitenow_articles/sitenow_articles.module
```diff
@@ -556,6 +556,12 @@ function sitenow_articles_preprocess_block(&$variables) {
 
       break;
 
+    // @todo Move this to a more appropriate spot, if we may
+    //   want to use the gallery image field for other content.
+    case 'field_block:node:article:field_gallery_images':
+      $variables['attributes']['class'][] = 'grid--fivecol--20 photoswipe-gallery';
+
+      break;
```
cory-skeers: Wondering if others have thoughts on this. I'd begun setting up the gallery image field as pretty generic, where it may be reused in other content types (similar to `field_image`), but I'm not sure where best to place things like this. `layout_builder_custom` seems wrong, but so does this.

cory-skeers:
Slight deviation from the mockup that we were looking for some feedback on (@bspeare and others). Including the gallery link inline with the org, author, etc., felt a little off in some cases, where you might have an author and an organization listed (author and their UI department, for instance), or in cases where you have a source org and a link. In these cases you end up having multiple elements that are all related, and then the gallery. It's less noticeable in cases where you only have one of these three elements, and the gallery link sharing space.

For now we've simply let it fall to its own space below the author/org byline:
image

Would otherwise love to get some more eyes on the work done in general, as well.

docroot/themes/custom/uids_base/templates/field/field--field-gallery-images.html.twig
```diff
@@ -0,0 +1,5 @@
+{% set title_attributes = title_attributes.addClass('h2 headline headline--serif headline--underline') %}
```
bspeare: Try adding the `block__headline` class in here to get some more spacing.

bspeare:
I'm getting an error going to `/admin/structure/types/manage/article/display/default/layout`, are you all getting that?

cory-skeers:
> I'm getting an error going to `/admin/structure/types/manage/article/display/default/layout`, are you all getting that?

We hadn't before, but I am now...Also getting a warning when viewing an article with the gallery.
```
Warning: array_search() expects parameter 2 to be array, string given in layout_builder_custom_preprocess_block() (line 1071 of modules/custom/layout_builder_custom/layout_builder_custom.module).
```

GaryRidgway:
An interesting thing here is that apparently a lot of class definitions on attribute render arrays are not arrays. For example, many of ours are `class => 'some-cool-class'`, and we were getting errors/warnings because the function wanted something similar to `class => ['some-cool-class']`. It seems we have this in a lot of places, so maybe we should be going through and fixing those to prevent future problems?

cory-skeers:
> > I'm getting an error going to `/admin/structure/types/manage/article/display/default/layout`, are you all getting that?
> 
> We hadn't before, but I am now...Also getting a warning when viewing an article with the gallery.
> 
> ```
> Warning: array_search() expects parameter 2 to be array, string given in layout_builder_custom_preprocess_block() (line 1071 of modules/custom/layout_builder_custom/layout_builder_custom.module).
> ```

I believe both of these warnings/errors have been fixed by putting the custom block's class assignment into an array.

bspeare:
I think we want to remove this left and right negative margin for this articles block only. 

Screen Shot 2022-08-11 at 10 33 27 AM

bspeare:
For inline fields with icons, we are adding a little extra margin after the icon.  Can you add this much spacing on the icon: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/fields/fields.scss#L48?

bspeare:
I added a possible fix in https://github.com/uiowa/uiowa/pull/5552/commits/c2d5c83c480d836180c25c92219065152fba1d2f.  Feel free to remove.  Maybe we can do something like this and create a follow-up issue for all of the stuff we'd like to do?

cory-skeers:
> I added a possible fix in [c2d5c83](https://github.com/uiowa/uiowa/commit/c2d5c83c480d836180c25c92219065152fba1d2f). Feel free to remove. Maybe we can do something like this and create a follow-up issue for all of the stuff we'd like to do?

That does work like a charm for Articles!

image

bspeare:
I created https://github.com/uiowa/uiowa/issues/5566 as a follow-up issue if you want to add any of your thoughts there!

cory-skeers:
- [x] IISC 
  - Quick test, the gallery field's there, but not appearing on display.
- [x] SPPA
  - Field not appearing on article content
- [x] HR
  - Quick test, everything looks good.
- [x] Admissions
  - Quick test, everything looks good.
- [x] Now
  - Quick test, everything looks good.

GaryRidgway:
these are odd issues....

cory-skeers:
We believe this is ready for review.

@joewhitsitt and @pyrello and others, if you have ideas of other sites that may be splitting Articles that may need config updates and/or testing, please feel free to add them to the list.

docroot/modules/custom/sitenow_articles/sitenow_articles.libraries.yml
```diff
@@ -0,0 +1,4 @@
+gallery-jump-link:
+  css:
+    theme:
+      css/gallery-jump-link.css: {}
```
bspeare: I like the idea of this but it might be more useful to move this to https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/content/node--type--article.scss so we can keep all of the article-related css together?

cory-skeers: I think I could go either way on it. The block it targets is defined in the module, but keeping all article-related css together makes sense to me.

GaryRidgway: I can refactor this, if we are eventually moving this anyways, then we can just deal with all of the, "where should it be put", questions when we finally do that re-factoring.

cory-skeers: That sounds good to me. Thanks Alan!

docroot/modules/custom/sitenow_articles/sitenow_articles.module
```diff
@@ -556,6 +573,12 @@ function sitenow_articles_preprocess_block(&$variables) {
 
       break;
 
+    // @todo Move this to a more appropriate spot, if we may
+    //   want to use the gallery image field for other content.
+    case 'field_block:node:article:field_gallery_images':
+      $variables['attributes']['class'][] = 'grid--fivecol--20 photoswipe-gallery';
```
joewhitsitt: Based on https://github.com/uiowa/uiowa/pull/5552#issuecomment-1212128967 should these also be separated as individual classes instead of a string of multiple?

cory-skeers: Yep--it wouldn't run into quite the same issue that brought it to our attention since it's at least within an array, but for consistency and possibly other issues, it should be split apart.

joewhitsitt:
pulling down and can fix the conflict while testing

docroot/modules/custom/sitenow_articles/src/Plugin/Block/GalleryJumpLink.php
```diff
@@ -0,0 +1,46 @@
...
+
+use Drupal\Core\Block\BlockBase;
+
+/**
+ * An image gallery jump link block.
+ *
+ * @Block(
+ *   id = "galleryjumplink_block",
+ *   admin_label = @Translation("Gallery Jump Link Block"),
+ *   category = @Translation("System")
```
joewhitsitt: Is there a reason why we didn't use the `Site custom` category here?

cory-skeers: When we were setting it up, `System` felt like the closest fit for a block that's intended for use by developers creating layouts, and not for per-page custom layouts. We'd done a fair amount of going back and forth when deciding, and could be swayed.

docroot/modules/custom/sitenow_articles/sitenow_articles.module
```diff
@@ -206,6 +206,23 @@ function sitenow_articles_preprocess_node(&$variables) {
             }
           }
         }
+        if (in_array($variables['view_mode'], ['default', 'full'])) {
+          // Check if we don't have any gallery images.
+          if (empty($node->get('field_gallery_images')->getValue())) {
```
joewhitsitt: This logic will WSOD during deployment for any traffic to articles.

To reproduce, checkout main, sync sandbox, checkout this branch, uli and try to hit https://sandbox.uiowa.ddev.site/news/2020/11/blue-flower-now-listed-our-site

cory-skeers: Ah, speaking of defensive coding practices...

cory-skeers: Gonna take a look at this.

docroot/modules/custom/sitenow_articles/src/Plugin/Block/GalleryJumpLink.php
joewhitsitt: `Site custom` is what we have used for all site defined blocks. This includes placed blocks for AoS, UIPress and is then restricted using layout builder restrictions. Alternatively, we have a few blocks that use the `Restricted` category (Region Item, Search).

I feel like `System` should be left for core.

docroot/modules/custom/sitenow_articles/sitenow_articles.libraries.yml
joewhitsitt: This has been resolved correct with https://github.com/uiowa/uiowa/pull/5552/commits/8d9947591c290cd4fd20b99dca39bb79d9e4fd21?

docroot/modules/custom/sitenow_articles/sitenow_articles.module
cory-skeers: Should avoid the WSOD now.

docroot/modules/custom/sitenow_articles/src/Plugin/Block/GalleryJumpLink.php
cory-skeers: That's fair re: leaving `System` to core.

`Site custom` to me sounds like blocks on an individual site like admissions, and not for blocks that may be used across the platform. Maybe `Restricted` then?

docroot/modules/custom/sitenow_articles/sitenow_articles.libraries.yml
cory-skeers: Yep! Should be good to go here.

docroot/modules/custom/sitenow_articles/src/Plugin/Block/GalleryJumpLink.php
joewhitsitt: +1 Restricted (forgot that this was default and not site specific)

cory-skeers: Sold. Moved it over.

joewhitsitt:
Added a doc issue: https://github.com/uiowa/uiowa/issues/5586

joewhitsitt (approved):
I think this looks good. Thanks for the quick adjustments.

cory-skeers:
> I think this looks good. Thanks for the quick adjustments.

Awesome. Thanks for the testing and feedback, and for creating the doc issue!

PR #5567, opened 2022-08-12
Now Spotlight article type; auto-tagging

### LINKED ISSUE #5562: Iowa Now Spotlights
# Story

As an Iowa Now website editor, I would like to be able to place a listing of articles that are spotlights on our SiteNow v3 site similarly to other article types.

# Additional Information
Currently "Spotlights" are "In the News" content that have a source url containing `uiowa.edu`.

When migrating from D7 we will want to migrate any "In the News" content that has a source url containing `uiowa.edu` to be granted the "Spotlight" type.

# Proposed Solution

- [x] Add article type for "Spotlight" that contains the same fields (hidden/shown/required) as "In the News" type.
- [x] Programmatically create tags for articles based on the Article Type (In the News, News Feature, Achievement, Spotlight) when saved.

cory-skeers:
Resolves #5562 

# How to test

Sync Iowa Now and make sure the site split's enabled:
```
git checkout now-spotlight && git pull
blt ds --site=now.uiowa.edu
drush @now.local en now_core
drush @now.local cim --source ../config/sites/now.uiowa.edu --partial
drush @now.local cim
drush @now.local cr
```

- Add some articles, including the newly added "spotlight" article type.
  - The Spotlight type should enable the "original publication date" field
- When an article is created, check that the associated tag is added
- Try different combos of including tags when creating an article, and check that a duplicate tag doesn't get added, that other tags aren't removed, etc.
- Edit an article and change the article type
  - After save, the new article type should be represented by a tag
  - After save, the previous article type should no longer be represented by a tag**
- Check that you can't delete tags associated with this auto-tagging (`achievements`, `featured`, `in-the-news`, `ui-spotlight`), but that you can delete other tags

** The current approach for this only removes tags from the previously selected article type, and not ALL other possible article types that may have been added manually. This allows a user to manually add additional tags that correspond to article types (eg 'feature', 'in-the-news', etc) in addition to the auto-tags and not lose them on save, while allowing the auto-tags to replace the old article type's tag with the new type's tag if it gets changed.


joewhitsitt:
If we are going this tag route, we should probably make sure that we remove the ability for folks to delete these tags (tids). An example of this is with the front page functionality:

`sitenow_form_node_confirm_form_alter()` and `sitenow_form_node_delete_multiple_confirm_form_alter()` within docroot/profiles/custom/sitenow/sitenow.profile

cory-skeers:
> If we are going this tag route, we should probably make sure that we remove the ability for folks to delete these tags (tids). An example of this is with the front page functionality:
> 
> `sitenow_form_node_confirm_form_alter()` and `sitenow_form_node_delete_multiple_confirm_form_alter()` within docroot/profiles/custom/sitenow/sitenow.profile

Good call! I've updated to prevent deletion using the `confirm_form_alter`; I believe we don't have the option to delete multiple tags, so that should cover it.

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
```diff
@@ -85,3 +138,61 @@ function now_core_node_presave($entity) {
...
+  // Check/prevent are auto-tags from being deleted on single delete.
+  // Only need to alter the delete operation form.
+  if ($form_state->getFormObject()->getOperation() !== 'delete') {
+    return;
+  }
+  $tag = $form_state
+    ->getFormObject()
+    ->getEntity();
+
+  if (in_array($tag->id(), [
```
joewhitsitt: Sorry to nitpick but can we use _article_type_tag_map() to get this list of ids so we only maintain the list in one place?

cory-skeers: Absolutely--pushed up a slight refactor for this.

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
```diff
@@ -85,3 +138,65 @@ function now_core_node_presave($entity) {
     );
   }
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function now_core_form_taxonomy_term_confirm_form_alter(&$form, FormStateInterface $form_state) {
+  // Check/prevent are auto-tags from being deleted on single delete.
```
pyrello: This sentence doesn't quite make sense to me.

cory-skeers: Doesn't make sense to me either. Updated to "  // Check/prevent tags from being deleted on single delete if they are in our list of tags used to associate article types."

joewhitsitt:
pulling down to test

joewhitsitt (approved):
I wouldn't close the door on extending the ListBlock to filter by this field sometime in the future, but this functionality seems to work just fine and there is no data loss for a fallback later on. It would be good to capture our reservations related to that method.

Adjusted the help text for the field and updated the default value.

PR #5569, opened 2022-08-12
Feature: Blockquote field updates with new basic text format

### LINKED ISSUE #5536: Add very limited HTML markup options to blockquote block `field_uiowa_quote_content` and `field_uiowa_quote_footer` fields
# Story

As a content editor I want to be able to italicize or bold text in the blockquote content or source fields so that have more control over how they look.

This is a request from UI Press.

# Proposed solution

- Maybe we need to consider a new text format in between plain text and minimal to add just the most basic styling options.
- Implement this new option on both fields in the blockquote block type `admin/structure/block/block-content/manage/uiowa_quote/fields`

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5536. 

# How to test

`blt frontend`

- Sync a site with a lot of blockquotes `blt ds --site=tippie.uiowa.edu` check out https://tippie.uiowa.ddev.site/undergraduate/admissions for blockquotes. 
- Sync a site with some temp storage and verify that the changes have been made: `blt ds --site=sandbox.uiowa.edu` and go to https://sandbox.prod.drupal.uiowa.edu/node/656/layout and edit the first blockquote and verify that the word "President" has been removed from the Source field. 
   - ~One issue I've noticed is that the item in temp storage in this example has `<p>` tags around it which should not be allowed~ https://github.com/uiowa/uiowa/pull/5569/commits/c0d1eced8e298bdb6b55e263d4eb09d6dea20e32. 
- Verify that update hook runs and changes the Content and Source fields to text areas with `basic` html
- Review the new `basic` text format/permissions and let me know if I am missing any files or entries or splits. 
- Review the update hook.


bspeare:
I'm not sure if this needs the `needs_scheduled_deploy` tag.

cory-skeers:
Going to start taking a look at this

cory-skeers:
> I'm not sure if this needs the `needs_scheduled_deploy` tag.

Wanna do some more looking/poking around this, but fwiw, I hit a local site at various points

1. with the codebase changes but without the update
2. after the update but before config imports and
3. after config imports

and didn't have any issues. Kind of informal testing this way, but at least nothing stood out as problematic.

pyrello:
> > I'm not sure if this needs the `needs_scheduled_deploy` tag.
> 
> Wanna do some more looking/poking around this, but fwiw, I hit a local site at various points
> 
> 1. with the codebase changes but without the update
> 2. after the update but before config imports and
> 3. after config imports
> 
> and didn't have any issues. Kind of informal testing this way, but at least nothing stood out as problematic.

If this was going to be an issue, I would expect it to show up on the blockquote block form, as that's where the mismatch between current and updated config would be encountered in situations where there is already data. That being said, I don't necessarily expect that that would be the case, especially since I don't think we have any custom code that is specifically counting on the form field format to be set a certain way.

cory-skeers:
> > > I'm not sure if this needs the `needs_scheduled_deploy` tag.
> > 
> > 
> > Wanna do some more looking/poking around this, but fwiw, I hit a local site at various points
> > 
> > 1. with the codebase changes but without the update
> > 2. after the update but before config imports and
> > 3. after config imports
> > 
> > and didn't have any issues. Kind of informal testing this way, but at least nothing stood out as problematic.
> 
> If this was going to be an issue, I would expect it to show up on the blockquote block form, as that's where the mismatch between current and updated config would be encountered in situations where there is already data. That being said, I don't necessarily expect that that would be the case, especially since I don't think we have any custom code that is specifically counting on the form field format to be set a certain way.

Content doesn't get displayed for blockquotes during the time when there is a mismatch in config
image
And if the block gets edited during this time, there's a slight mismatch with the format being available for changing and not yet having the basic_html option (for admins). For permissions below admin, this just means the content gets locked.
image
image
For new blockquotes created during this time, users are able to select filtered or minimal and save, leading to some weirdness
(pre config import)
image
(post config import)
image
In the above new blockquote instance of mismatched formatting, post-config import also requires dev involvement to fix, because any lesser permissions don't have access to edit the field and change the format.

So, nothing to break and cause an outage, but there are some complications that can arise depending on if a page is being viewed and/or edited during the time between update hook running and the final config import (if I'm right on the ordering of the deployment process..if not all the better).

Mismatches between old/new config and the codebase doesn't lead to any issues--it's only if there's a point between the updb and the config import.

pyrello:
> So, nothing to break and cause an outage, but there are some complications that can arise depending on if a page is being viewed and/or edited during the time between update hook running and the final config import (if I'm right on the ordering of the deployment process..if not all the better).

It seems like the weirdness above is enough reason to do an off-hours deployment even without the risk of an outage.

bspeare:
@cory-skeers can you share your technique for testing those steps listed on https://github.com/uiowa/uiowa/pull/5569#issuecomment-1218427647?

cory-skeers:
> @cory-skeers can you share your technique for testing those steps listed on [#5569 (comment)](https://github.com/uiowa/uiowa/pull/5569#issuecomment-1218427647)?

- `git checkout main && git pull`
- `blt ds --site=tippie.uiowa.edu`
- `git checkout feature_basic_markup_html && git pull`
- (play with things)
- `drush @tippie.local updb -y`
- (play with things)
- `drush @tippie.local cim -y`

The first round of testing in there would be once the codebase updates and nothing else has happened, and I didn't experience any issues. The second round is then between the update hook running and config changes importing, and is when those behaviors were observed.

bspeare:
Thanks @cory-skeers!

cory-skeers:
It's a small detail, but is there something small we could do to avoid the "About text formats" being added when users aren't allowed other options?
image

I vaguely remember something like this showing up and being addressed (or not?) but I'm not sure.

cory-skeers:
Other extremely minor detail--any opposition to reordering the formats so that `Basic HTML` appears below `Minimal HTML`? 
image

cory-skeers:
> I vaguely remember something like this showing up and being addressed (or not?) but I'm not sure.

It's just in the form display--not sure what I was thinking of that had me thinking it was more complicated.

cory-skeers:
Had the few small notes/questions about the form display, but otherwise I've gone through the testing with this and think it is good to go (off-hours for the possible editor collisions).

I also reviewed the update hook, and think everything looks good there. I haven't explicitly stepped through it, but it seems to catch the few gotchas that might exist for certain sites, like not having a revisions table for the fields yet.

I'd give it my 👍 . Leaving it unapproved for now since we'll want to plan for its deployment, and leaving it up at least through next PL to see if any other reviewers want to do their own testing/reviewing as well.

joewhitsitt:
pulling to test

joewhitsitt:
Sorry if I missed a discussion about this. Do we want to be consistent and offer plain text for these fields too? For example, card excerpt has minimal and plain.

bspeare:
@joewhitsitt I added plain text.  I think we should be consistent.

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -2147,3 +2149,143 @@ function sitenow_update_9046() {
...
+  $table_mapping = \Drupal::service('entity_type.manager')
+    ->getStorage('block_content')
+    ->getTableMapping();
+  $storage_definitions = \Drupal::service('entity_field.manager')
+    ->getActiveFieldStorageDefinitions('block_content');
+  foreach ($fields as $field_name => &$field) {
+    $field_definition = $storage_definitions[$field_name];
+    if ($field_definition) {
+      $field['table'] = $table_mapping->getDedicatedDataTableName($field_definition);
+      strip_tags($field['revision_table'] = $table_mapping->getDedicatedRevisionTableName($field_definition));
```
joewhitsitt: Not sure what `strip_tags` does since it is fetching a table name

bspeare: Adding it there removed some `<p>` tags that were being rendered around the field.   I'm open to other suggestions on how to strip the tags from the revisions.

joewhitsitt: >Adding it there removed some `<p>` tags that were being rendered around the field.

i just don't understand how this does this. `$table_mapping->getDedicatedRevisionTableName($field_definition)` returns a table name string.

Screen Shot 2022-08-18 at 4 01 00 PM

I haven't traced down that example of `<p>` tags being included but I would imagine stripping the tags on the data $rows and $revision_rows would be where this happens.

Screen Shot 2022-08-18 at 4 02 28 PM

pyrello: Either way, I would think that strip tags should go around whatever is being evaluated on the right side of the `=`. If for example the `=` condition were to evaluate to `FALSE`, then I'm not sure this is something that `strip_tags` could handle.

bspeare: I looked for the `<p>` tag in xdebug too but couldn't find it there and just saw the string.   I moved the `strip_tags` to the right side and it is still working.

joewhitsitt: I might need to have you run through this in a huddle tomorrow. I just removed strip_tags from the update hook and synced the sandbox site. I am not seeing `<p>` tags either.

Also if this is doing what it says it does, why is only the revision_table addressed? Shouldn't we do this for table too?

bspeare: Sounds good. Let's huddle after standup if you are available.

joewhitsitt:
We found that strip_tags wasn't working as placed and that tempstore is the issue where it isn't being targeted as part of this update hook. We determined it was not worth the squeeze to switch out a field and try to fix serialized data when a block save corrects the visual issue.

joewhitsitt (approved):
How about scheduling this for 3 AM Monday (Friday afternoon setup)?

PR #5572, opened 2022-08-12
Feature: Articles page layout fix

### LINKED ISSUE #5542: SiteNow Articles Page Flex Adjustments
# Story

As a visitor to a SiteNow Articles view page with a large monthly archive, I don't have to scroll all the way to the bottom to see the pager.

https://international.uiowa.edu/news

# Proposed solution

- [ ] Adjust the flex styles on this page.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5542. 

# How to test

- yarn workspace uids_base build`
- `ddev blt ds --site=international.uiowa.edu`
- Test https://international.uiowa.ddev.site/news against https://international.uiowa.edu/news.
- Test on a page without a pager like https://sandbox.uiowa.ddev.site/news
- Test browser breakpoints, shrink desktop slowly to mobile.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/views/view-display-id-page_articles.scss
+++ b/docroot/themes/custom/uids_base/scss/views/view-display-id-page_articles.scss
@@ -2,48 +2,32 @@
 @import "uids3/assets/scss/_utilities.scss";
 
 .view-display-id-page_articles {
-  display: flex;
-  flex-wrap: wrap;
   @include margin($top: $lg);
-  .view-header {
-    width: 100%;
+  @include breakpoint(sm) {
+    display: grid;
+    grid-template-columns: repeat(3, 1fr);
+    grid-template-rows: repeat(3, 1fr);
+    grid-column-gap: 0px;
+    grid-row-gap: 0px;
   }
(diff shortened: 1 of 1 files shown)


cory-skeers:
The monthly archive feels a little scrunched (on large/desktop screens), particularly with longer month names (eg September). I'm not sure I would feel that way if I hadn't just recently looked at the current setup and noticed the difference.

image

Existing for comparison:
image

The RSS Feed also ends up alone on large/desktop screens (still down where the pager formerly sat). I'm fine with it being down there on its own, but wonder if others might want to see it brought up with the pager as well?

bspeare:
Thanks @cory-skeers!  I think your points are correct.  I've updated it to give the sidebar some more space and moved the feed back under the pager.

cory-skeers:
> Thanks @cory-skeers! I think your points are correct. I've updated it to give the sidebar some more space and moved the feed back under the pager.

Thanks! I'll check out the updates.

cory-skeers (approved):
Thanks for the extra updates @bspeare--this looks great!

PR #5576, opened 2022-08-15
Add after content region options

### LINKED ISSUE #4767: As a site owner, I can have a row of buttons that appears fixed to the bottom of the browser

GaryRidgway:
tracks #4767

# How to test
- checkout branch
- git pull
- composer install
- blt frontend
- sync default

## Permissions
- As an Editor, I can access the region configuration page and create/edit region items but I can't set the global region item. 
  -  Do we want these fields disabled or removed?  I thought it might be nice for Editors to know the fields exist, but not available to use. 
- As a Publisher, I can't access the region configuration page. 

## Region settings page
- on `https://default.uiowa.ddev.site/admin/config/sitenow/region-settings` you should be able to see two areas, after content and pre-footer.
- You should be able to create both after content and pre-footer region items, and assign them only to their specific region based on type.
- saving should store these values

## After content region
- The after content region should not be configurable as a webmaster
- The after content region should only allow buttons
- The after content region should have gold as a default background color
- full width buttons should span the entirety of the after content region
- The after content region should only allow you one content block.

## Pre footer content region
- Creating a new pre footer region item should only allow you to have one block.

## Multi-value button

Screenshots of the multi-value button within layouts.  When the "Equal Width" checkbox is checked, here's how the button will behave:

One column layout:
- The buttons will flex equal widths in a single row.

Page w/left sidebar, Two, Three, Four column layouts
- The buttons will stack and display full-width of the layout column. 

<details>
Screen Shot 2022-08-24 at 3 28 37 PM
Screen Shot 2022-08-24 at 3 28 26 PM
Screen Shot 2022-08-24 at 3 28 20 PM
Screen Shot 2022-08-24 at 3 28 13 PM
Screen Shot 2022-08-24 at 3 28 06 PM
</details>


bspeare:
We might want to add `Add small top spacing` and `Add small bottom spacing` options to our container settings so that we can go from the defaults:

Screen Shot 2022-08-22 at 11 35 16 AM

To something smaller:

Screen Shot 2022-08-22 at 11 36 16 AM

bspeare:
# Summary of current discussion points

Using the shared/same fragment type as the pre-footer has some downsides and benefits:

## Benefits 

- Ease of saving. we have set this up in the past for region items because we wanted to keep things flexible for the future.
- `region items` can be used in any `region item` block.
- Continuing to use region items allows us strong control over the overarching config of all of them.
- Using region items means we don’t have to manage two types of region-item-esque entities.
- Using region items allows us to maintain locked down single column section without extra code.
- Using region items allows us to use region_item and not create a separate fragment type, therefore keeping a consistent editing experience.
- Using region items allows us to open up after_content to other layout builder components in the future.

## Downsides

- One layout to rule both regions introduces extraneous block types in the after-content region.  
  - This is not ideal because we only want to have the button block available in the after-content region (in the future we might want to allow more).
  - Styling for region items that are sticky in the after-content gets real wonky real quick. What happens if there are text areas that we don't want to stick?
    - Determining what behavior makes region items stick or not stick is messy.  A field for controlling this would go where? 

### Possible proposed solution

1. ?
2. Continue to use one fragment and set some custom code that sits on top of Layout Builder Restrictions to limit blocks used in after-content region.  
3. Create different fragment types to represent each region.

If we went with option 3, we could:

- Create a new fragment type for `region_aftercontent`.
- Rename the current fragment `/admin/structure/fragment-types` to `region_prefooter` and name the new fragment `region_aftercontent`.
- Use the key value from `uiowa_core.settings` (`/admin/config/development/configuration/single/export`) such as `region_prefooter` or `region_aftercontent` and concatenate it to https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_core/src/Form/RegionSettings.php#L102.  
- Adjust region settings page `/admin/config/sitenow/region-settings` to have two separate sections, one section for pre-footer and another section for after-content. 
- This would require an update hook

joewhitsitt:
so multi-button has been removed? i missed that bit. flex layout makes more sense i guess in that situation

bspeare:
Yes, multi-button has been removed in favor of the flex layout.

GaryRidgway:
things to keep in mind as we are working on this after deciding on the flex layout. This list will grow.

- [x] Full width toggle on buttons for after content is useless, and needs to be removed through access or code. If this is not removed it can mess with our `after_content` specific styles
- [x] layout builder page has some odd styles for single buttons because we are scoping single buttons to not flex based on `:last-child:first-child`, but there is also the "Add block" button that makes it not `:last-child:first-child`, but only on the layout builder edit page

bspeare:
@joewhitsitt after working a bit with the flex layout, we decided to go back to the multi-button.   Providing the multi-button will create value for users on the platform and create less overhead for restricting some button style options, like the issues mentioned in https://github.com/uiowa/uiowa/pull/5576#issuecomment-1225869423.

bspeare:
Maybe we should keep this sticky at all times?   It doesn't look great on smaller pages and in combination with the prefooter. 

Screen Shot 2022-08-25 at 3 09 00 PM

GaryRidgway:
Leaving a comment here for easy access to code that I am removing:

https://github.com/uiowa/uiowa/compare/3691d94...b9c03fa

joewhitsitt:
pulling down as well

cory-skeers:
~It looks like the button block in the After Content region doesn't have a character limit. I tried adding a button block in a regular page layout, and it was limited. Edit: Also tried in the pre-footer content region, and hit the character limit.~

Update: Was a local issue.

joewhitsitt:
1. Needs to account for entity print on admissions: https://github.com/uiowa/uiowa/issues/3204 as it relates to core.entity_view_display.fragment.region_item_after_content.default
2. Should the region settings permission portion of this issue be addressed? https://github.com/uiowa/uiowa/issues/5609
3. Should the human readable terminology for prefooter be changed?

Screen Shot 2022-08-26 at 1 09 45 PM

cory-skeers:
I wonder what other people's thoughts are on the after content item appearing when editing page layouts?

My initial thought was to ask if we could hide it because it's distracting, but

1. Having it appear when editing the layout is more accurate in the WYSIWYG approach to layout builder, and how the final page will look
2. As an "eat our own dog food" concept, it's a way for editors to experience what the sticky content item experience will be like for the end user

joewhitsitt:
>I wonder what other people's thoughts are on the after content item appearing when editing page layouts?

Previously talking with Alan, I think we thought that it could get in the way of block save, add components, etc. A lot of functionality to account for there that isn't on the rendered page.

cory-skeers:
~Both after content and pre-footer content pieces are appearing in both of the content management corrals. (The limitation to only be able to add an after content to that region or pre-footer to that region is working as intended.)~

Update: Was a local issue.

joewhitsitt:
@cory-skeers are you sure you local is up to date? so far I haven't been able to reproduce the max length issue and I don't see the content in the wrong view

Screen Shot 2022-08-26 at 1 51 51 PM
.

composer install is needed for the fragment views patch (filter by fragment type)

cory-skeers:
> @cory-skeers are you sure you local is up to date? so far I haven't been able to reproduce the max length issue and I don't see the content in the wrong view
> 
> Screen Shot 2022-08-26 at 1 51 51 PM
> 
> .
> composer install is needed for the fragment views patch (filter by fragment type)

Hrm...I did another composer install and checked that the config was up to date, cleared cache, etc., and am still seeing both issues. I did have some funkiness going on with my local this morning, though, so not ruling it out. I'll try stepping back and re-setting things up from the beginning and see if it persists.

Update: Was able to sync a different site and have things work correctly. Dropped default, re-synced, and now am seeing things work as intended.

cory-skeers:
> > I wonder what other people's thoughts are on the after content item appearing when editing page layouts?
> 
> Previously talking with Alan, I think we thought that it could get in the way of block save, add components, etc. A lot of functionality to account for there that isn't on the rendered page.

That makes sense. And the amount of functionality that needs to be accounted for could always increase over time, or constantly need to be checked against, as updates to core and our system come out.

I'm seeing it when editing layouts--does that still need to be addressed? After other parts turning out to be local issues, I'm not sure if it's just me 😬

joewhitsitt:
I see it too. Originally, I think Alan had the mobile effect of being in the region instead of floating while in layout builder. So far, I think the z-indexes are all okay.
 
Screen Shot 2022-08-26 at 3 18 45 PM

cory-skeers:
I was able to create too long button texts in other places with the multi-button option, so maybe it wasn't (just) my local:
image

to recreate this time: Add a button block, and the first button added correctly gets the character limit. Additional buttons I have been able to insert as much text as I want, and save the block. After saving, I am able to go back in and rearrange the buttons to save with one of the long texted buttons in the first slot, but upon editing a third time, the validation seems to apply to all buttons.

On a separate note: Once a button is added, is it possible to remove it? (eg I add two buttons in a block, but now only want one.)

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -840,6 +841,18 @@ function _layout_builder_custom_process_element(array $element, FormStateInterfa
 
     case 'inline_block:uiowa_button':
       $element['field_uiowa_button_link']['widget'][0]['title']['#maxlength'] = 25;
```
cory-skeers: I think this line needs to be updated to account for multiple buttons (deltas 1 and 2 in widget, if they exist) to ensure all buttons have a max length.

cory-skeers (changes requested):

cory-skeers:
Checking Equal Width button block with a single button in a multi-col layout stretches the button height to fill the container.

Example next to a multi-button, equal width component.
image

Example next to cards.
image

Two buttons in the block with Equal Width checked.
image

~Edit to add: From working on a page that already had equal width buttons in a multi-col layout (stretched to fit the width of the column), I had the expectation that checking Equal Width for a single button should make it full width. I know it doesn't quite fit for the meaning of "equal width," but behavior-wise, it made sense. Could we either hide the Equal Width checkbox if only one button is present, or add clarifying help text?~

Edit to the edit: Equal width is working exactly how I expected; it was just the weird stretchy that was interfering.

bspeare:
@cory-skeers thanks for finding the stretchy buttons.   I just pushed up a fix that should fix that.

joewhitsitt:
Stubbed out a docs issue: https://github.com/uiowa/uiowa/issues/5620

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
cory-skeers: Tested the update, and working great. Thanks @bspeare!

bspeare:
@cory-skeers I added some description text to the "Equal width" checkbox: https://github.com/uiowa/uiowa/pull/5576/commits/28ec881015c8f10acae7c4d270eba56fdcb55dc4.   Do you have any suggestions to improve?

cory-skeers:
> I added some description text to the "Equal width" checkbox: [28ec881](https://github.com/uiowa/uiowa/commit/28ec881015c8f10acae7c4d270eba56fdcb55dc4). Do you have any suggestions to improve?

Adjusted a bit to: "If selected in single column layouts, multiple buttons will appear inline and equally sized. If selected in layouts with multiple columns, the buttons will stretch to fill the column and stack."

Also tested out the character limit validation on multiple buttons, and it's working great!

bspeare:
>  As an "eat our own dog food" concept, it's a way for editors to experience what the sticky content item experience will be like for the end user.

I like this idea, but perhaps we should hide it since we aren't showing the header, prefooter, or footer in the layout edit screen?

bspeare:
> On a separate note: Once a button is added, is it possible to remove it? (eg I add two buttons in a block, but now only want one.)

It would be nice to have a remove button.  The banner works the same way in that you have to delete the text from the field to remove unwanted button(s).

cory-skeers:
> > On a separate note: Once a button is added, is it possible to remove it? (eg I add two buttons in a block, but now only want one.)
> 
> It would be nice to have a remove button. The banner works the same way in that you have to delete the text from the field to remove unwanted button(s).

Okay! I'm fine with that as a way to remove it--a remove button might be a nice addition, but not necessary like it is with collections, because it needs to remove the actual paragraph.

I'd tried removing a button like that before and it didn't work, but I think it was related to the character length validations. Tried it out following the update there, and no issues.

bspeare:
@joewhitsitt I addressed items 1 and 3 in https://github.com/uiowa/uiowa/pull/5576#issuecomment-1228815530.   What's the best practice on the permissions part - where do we set that?

joewhitsitt:
Right now, the fragments permission is controlling access to the settings form. In addition to making sure editors have the correct fragment type permissions, the region settings access can be changed here to a webmaster only type permission along the same lines: docroot/modules/custom/uiowa_core/uiowa_core.routing.yml

cory-skeers:
I can test the region settings permissions if they get updated, but otherwise I believe I've gone through all the listed testing, my own testing, and tested the subsequent updates, and could give it my 👍

bspeare:
I've got some questions on the permissions over at https://github.com/uiowa/uiowa/issues/5609#issuecomment-1230709325, so I'm going to hold off on the permissions change until we can discuss those.

bspeare:
We discussed the permissions issue at mid-sprint demo today and we decided that the work here was blocked until we can meet tomorrow to discuss these issues further:

- Editors
  -  Editors will potentially be able to access the page override to set the after content overrides.  
    -  The plan is that there will be a reference field on the page content type to reference an existing after content item or click a button to create a new after content item. 
    - They should probably still have access to `/admin/config/sitenow/region-settings`.
      - Do we just remove access to the reference field on that page?

- Publishers
  - Do not have access to layout builder.
  - After content and Pre footer require layout builder access to create.
  - Should Publishers  have access to the region settings page?

joewhitsitt:
Adding my thoughts here. Not going to make the meeting.

> Editors will potentially be able to access the page override to set the after content overrides.

I was thinking that this would be a form disabled access check and help text if we decided that editors would not be allowed.

> [Publishers] Do not have access to layout builder.

As I said on https://github.com/uiowa/uiowa/issues/5609#issuecomment-1230870083, it was originally setup as a stacking role meaning to get the full use of the role, they would have both the editor and publisher role assigned. Because of the confusion, I am a proponent of moving away from the stacked mindset in https://github.com/uiowa/uiowa/issues/4294. For this issue though, I don't see why that should should tie up this issue. 

>Should Publishers have access to the region settings page?

Yes, I think this role should have the necessary permissions to work with the region functionality.

bspeare:
After our discussion this morning, I have added these updates:

# Editors

Editors can access the Region settings page but will find the reference fields disabled with a message displayed at the top of the page:

Screen Shot 2022-08-31 at 10 45 43 AM

# Publishers 

For Publishers, I changed the access for the Region settings page to based on layout builder https://github.com/uiowa/uiowa/blob/f96b405363bf5e6b3f386987bff0ddfa6cf19596/docroot/modules/custom/uiowa_core/uiowa_core.routing.yml#L14 so this option no longer appears on configuration page.   If we grant Publishers access to Layout Builder in the future, they will be able to use this page. 

Screen Shot 2022-08-31 at 10 45 00 AM

bspeare:
@cory-skeers I added some new testing instructions for the permission updates if you have time to take a look at this again.

cory-skeers:
> I added some new testing instructions for the permission updates if you have time to take a look at this again.

Pulling it back down. Thanks for the heads up!

cory-skeers:
> * Do we want these fields disabled or removed?  I thought it might be nice for Editors to know the fields exist, but not available to use.

I think it's good to still show them. It also makes it clear which items are the active item, in case there are multiple available in the items list.

cory-skeers:
Tested the permissions updates, and it's all working as expected.

My one point of pause is that an editor role can make changes to a region item that has been placed already, without any further check or review. This is all working within how permissions line up, but we might want to be aware of it and discuss possibly locking down active region items, adding a content workflow to them, or something else. That can come at a later point, though.

cory-skeers:
After this gets merged and deployed, the doc updates (#5620) can be published.

cory-skeers:
Did a code review after the merge and things looked good, and was doing a quick pass through testing again.

The links on the buttons seem to be broken, and not processed into a proper link for the end user. The example below is with an after content item, but it's the same for any button using an internal reference link.

image

bspeare:
Thanks @cory-skeers, I'll take a look!

bspeare:
@cory-skeers the button link should be working now.

cory-skeers:
> the button link should be working now.

Thanks! Checked it out and it works perfectly.

cory-skeers (approved):
I think this is ready to go! 🎉

PR #5583, opened 2022-08-16
Remove field_uiowa_events_past

### LINKED ISSUE #5518: Remove field_uiowa_events_past from v3 events block
# Reference

https://github.com/uiowa/uiowa/pull/5497/files#diff-22455fd1cfa159b58a0289784a91064192c27738f06b23e252be3a46271afb4cR857-R861

# Proposed Solution

- [ ] Remove field/data from events block/db.
- [ ] Remove logic from layout_builder_custom module and sitenow_events preprocess block.
- [ ] Make sure sitenow_events_paragraph (v2) past functionality is not affected.

# Blocked

- Blocked until https://github.com/uiowa/uiowa/pull/5497 is merged/deployed.

cory-skeers:
Resolves #5518 

Removes the deprecated `field_uiowa_events_past` boolean field, and its related logic in `layout_builder_custom`.

# How to test

- `git checkout remove-events-past-bool && git pull`
- Sync a site, watch that the config update works and completes successfully
- Test an existing events block and create a new one, check that they still work
- Sync a v2 site, double check that the events paragraph still works
- Double check that the db no longer contains data for the old field
  - Data would normally be stored in `block_content_revision__field_uiowa_events_past` and `block_content__field_uiowa_events_past`, both of which should be removed upon removal of the field from config
- Double check for other custom code that might be tied to this field (`field_uiowa_events_past`)


makurj:
Everything checked out except for finding 2 instances of `field_uiowa_events_past:`

config/default/diff.plugins.yml
image

docroot/profiles/custom/sitenow/content/block_content/e4d3ec12-11e9-4ed7-ad97-002c4c8beecd.yml
image

cory-skeers:
> docroot/profiles/custom/sitenow/content/block_content/e4d3ec12-11e9-4ed7-ad97-002c4c8beecd.yml
> image

Ahh...I think this is part of the default content for new installs?

makurj:
https://github.com/uiowa/uiowa/pull/5583#issuecomment-1218450058

Okay, should we remove that somehow?

What about the other instance? config/default/diff.plugins.yml

makurj:
The default site one looks good now. docroot/profiles/custom/sitenow/content/block_content/e4d3ec12-11e9-4ed7-ad97-002c4c8beecd.yml

cory-skeers:
I think I've accounted for the field changes for the default content (removing `field_uiowa_events_past` and setting the new timeframe and sort options), but that'll need to be tested

The diff.plugin one I think is good too. It hadn't been re-exported in awhile, so there are some changes there that aren't specifically tied to this PR, but that should match what the expected setup should be for a site at this time.

makurj:
> I think I've accounted for the field changes for the default content (removing field_uiowa_events_past and setting the new timeframe and sort options), but that'll need to be tested

Editing an existing new block to date range seemed to work fine, and also the past one seems to be working okay still, as well as new blocks with past and date range options. 

I'm not seeing any change in diff.plugins.yml even though when I pulled the branch and `cim`-ed, it showed changes. Maybe the file hasn't refreshed yet?

makurj:
That did it! Not seeing `field_uiowa_events_past` in diff.plugins.yml anymore.

cory-skeers:
> That did it! Not seeing `field_uiowa_events_past` in diff.plugins.yml anymore.

Thanks! Yeah, guess it was a bad assumption on my part, and the first re-export made additions but not deletions.

makurj (approved):

PR #5587, opened 2022-08-17
Feature: container query polyfill

bspeare:
A quick experiment using the updated container query poyfill.  

# How to test

- `blt frontend`
- Go to https://sandbox.uiowa.ddev.site/four-column-card-list and collapse the browser and the card lists should display as stacked when you get below 1230px.


bspeare:
Since container queries are now supported in Chrome and Safari, I don't think it is worth pursing the polyfil.

PR #5619, opened 2022-08-26
Expand run first list of websites

### LINKED ISSUE #5558: Run campus-critical websites first during a deployment
# Story

As a web team, we can have a mechanism whereby campus-critical websites are the first to recover during a deployment issue like we experienced on 8/5/22.

# Additional Comments

Outages/problems will happen.  What can we do to lessen the impact/speed up recovery when/if they DO happen? 

Probably a number of different things we _could_ do here.  Let's figure out something that would make a material difference.

# Proposed Solution

- [ ] Define list of campus critical websites - This should include colleges, public safety, emergency, admissions, uiowa.edu. Want to keep this list short with specific reasoning, so we don't have to deal with everyone wanting to be on this list.
- [ ] Store ordered list in config and then add those to the beginning of the site list so they run first during deployment similar to how we are currently handling uiowa.edu in https://github.com/uiowa/uiowa/blob/c0b937b6123e3b88498c1b9fb6322c8dcc8b0cb3/blt/src/Blt/Plugin/Commands/ReplaceCommands.php#L29-L35

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5558

# To Test

- Checkout main, `ddev blt composer install` and run a `ddev blt dsa` on a couple websites that also include uiowa.edu, admissions, grad.admissions where they aren't first in the list.
- Run `ddev blt auda` and see that uiowa.edu is run first followed by whatever is next.
- Checkout `run_first_sites` and run `ddev blt auda` again. See that uiowa.edu admissions and grad.admissions in that order are run first.


joewhitsitt:
actually, i want to make another change

bspeare:
Pulling to test.

bspeare (approved):
`ddev blt auda` ran the `run_first` sites first.

Issue #5575, opened 2022-08-15
Support referenced content for articles


briand44:
As a SiteNow site editor creating an article, I would like to be able to choose existing related articles to display to site visitors so that they can more easily find similar articles they are interested in.

# Proposed Solution 

- [x] Add field_content_reference (avoids conflict with HR field_reference) to article content type with the ability to reference other articles.
- [x] Display as stacked cards in a three column grid.
- [x] Adjust site splits (iisc, sppa, hr, admissions)
- [x] ~Update default_content?~

# Follow-up

- Global settings with other display options

joewhitsitt:
I am curious to hear from @quamsta about how this new field could/should interact with the related by tag functionality.

cory-skeers:
> * Integrate related content field with the related content by tag functionality (admin/config/sitenow/uiowa-articles) where referenced items will take precedence over tags or is it a separate option from the dropdown to show those instead of tag based?

For the majority of cases, I would think referenced items first, then related by tags.

In my experience in the past (uihc, library) there've been some cases where it was desired to display only manually related content, and in other cases allow for both manually related and auto/tag-related. It was generally sensitive areas (eg medical, public policy) where it was limited to manual only.

If that is the case, I would lean toward providing the option to do referenced items only, or referenced items then tags, defaulting to the latter. (And with an option, use a form of our 'site-wide default' or override method.)

joewhitsitt:
>And with an option, use a form of our 'site-wide default' or override method.

Yeah, at this point I think establishing a "related content" sidebar section for this, help text and the reference field would be a good idea and by doing so highlights the existing functionality more.

bspeare:
I'd vote to display the related content in a card grid by default and for a future slice, develop different card display options like the screenshots below. 

<img width="550" alt="Screen Shot 2022-08-16 at 1 54 47 PM" src="https://user-images.githubusercontent.com/1036433/184959045-d751c602-582c-4a2a-9d0a-02e29dd54e0b.png">
<img width="484" alt="Screen Shot 2022-08-16 at 1 55 35 PM" src="https://user-images.githubusercontent.com/1036433/184959291-d8021346-3952-4094-9608-abff26a4d5ef.png">

briand44:
I can get on board with having configuration here if the plan is to have different display options outside of "Display" or "Do not display".  If we only have the card display option, my thought is to just display it if there is referenced content otherwise do not display. No need to add additional config.

quamsta:
[...]

PR #5623, opened 2022-08-29
Article content reference field

joewhitsitt:
Relates to #5575 

# Todo
- [x] stacked cards - taking this opportunity to refactor how classes are implemented for these? https://github.com/uiowa/uiowa/pull/5623#issuecomment-1230811205
- [x] hr split
- [x] iisc https://github.com/uiowa/uiowa/issues/5657
- [x] sppa

# How to test
- Review the code.  Any suggestions to make the machine names better like `card_grid` and `related_manually_referenced_content`?  Is there any extra code that could be removed?
- `blt frontend`

## Test the update hook
- Go to https://sandbox.uiowa.ddev.site/admin/config/sitenow/uiowa-articles and verify the "do not display" option is gone from "Display related content" and that this site is set to "Grouped by tag".  Adjust it to display "Manually referenced content" and save. 
- Test on another site that may have had "Do not display" set. 

## Test the functionality
- Go to https://sandbox.uiowa.ddev.site/node/546/edit and find the "Related content" field in the sidebar and select some content to reference.  Read the help text and provide suggestions to help make the text better. 
- Edit or create a new article and use the new "Related Content" field under the "Relationships" tab. 
- A related content section should display a grid of cards at the bottom of the article in the "stacked" card design. 

## Test the splits
- HR - `ddev blt ds --site=hr.uiowa.edu` - it should work on v2 but just look different https://github.com/uiowa/uiowa/pull/5623#issuecomment-1238383764
- IISC - `ddev blt ds --site=iisc.uiowa.edu` it should not work here
- SPPA -`ddev blt ds --site=sppa.uiowa.edu`-  it should work!


bspeare:
We are setting the card defaults for the teaser view mode at https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L308-L312.  Can we unset those and provide new defaults for article teasers using this reference field?  Or, do we need a card view mode?

joewhitsitt:
Don't we need to keep that around until v2 goes away? Not sure I understand the ramifications of unsetting those there. Would that be a lot of teaser template updates?

joewhitsitt:
Input overflow
Screen Shot 2022-08-30 at 6 50 56 AM

Grid stacked enclosed with hidden descriptions using existing logic.
Screen Shot 2022-08-30 at 6 51 07 AM

joewhitsitt:
pulling down to look at the update hook and view

joewhitsitt:
Looks like when this was first installed (sitenow_articles_update_9301), it was set to an option that doesn't exist which left things unset for many sites. It now accounts for do_not_display as well as unset.

joewhitsitt:
i'd like to pretend I understand how Views works but I don't. required the relationship, removed the contextual filter and switched the render to content. 🤷 

Update: nevermind. I must have broke something

bspeare:
@joewhitsitt what needs to be updated for splits like (https://github.com/uiowa/uiowa/blob/main/config/sites/hr.uiowa.edu/config_split.config_split.site.yml#L22) or default content?

joewhitsitt:
default_content is probably nothing since we aren't setting a default value.

for the HR split, I think it would be removing the storage base field from the split and adding a complete split item for the field instance located on page. making sure the storage base file is removed from git.

bspeare:
It looks like https://github.com/uiowa/uiowa/blob/9e164473d42499b22e61755361d8279014c28fe7/docroot/themes/custom/uids_base/uids_base.theme#L305-L311 is not working on the `sppa` site.

joewhitsitt:
taking a look at SPPA

bspeare:
This feature does work for V2 sites but has a different visual display - are we good with that?

Screen Shot 2022-09-06 at 11 26 54 AM

bspeare:
I think this should be ready to test.   @joewhitsitt anything else you want to add to the testing instructions?

joewhitsitt:
I think the test instructions are good

makurj:
Pulling down to test.

makurj:
> Any suggestions to make the machine names better like `card_grid` and `related_manually_referenced_content`?

 I kinda like these as they are very clear terms.

makurj:
I'm okay with: 

> Reference similar content. Manage the display of these items at the SiteNow Articles [page](https://sandbox.uiowa.ddev.site/admin/config/sitenow/uiowa-articles).

If needed, we could change "similar" to "related" but as this is under "Related Content", that would be redundant. We could also do "Reference other articles. Manage the display of these items at the SiteNow Articles."

makurj:
To confirm, we are now setting “Display related content” for all sites to "manually referenced"? For ones that were on “Do not display”, this shouldn’t make a difference unless they go into each article and enter some content to reference, correct?

briand44:
>To confirm, we are now setting “Display related content” for all sites to "manually referenced"?

We should be setting "manually referenced" as the default going forward and updating existing sites that have "Do not display" set. Sites that are displaying tags wouldn't be changed.

joewhitsitt:
>Sites that are displaying tags wouldn't be changed.

Correct.

makurj:
For IISC, I see the "Related Content" field in the sidebar on articles but when I enter some text, it's not referencing the article. Is that what is meant by "it should not work here" in the test instructions?

briand44:
https://github.com/uiowa/uiowa/issues/5657

makurj:
Everything checks out.

makurj (approved):

bspeare:
Does this need the off hours deployment tag?

joewhitsitt:
Created a docs issue https://github.com/uiowa/uiowa/issues/5665

pyrello:
> Does this need the off hours deployment tag?

I reviewed the code from this perspective and I didn't see anything concerning in this regard. It doesn't:
- Introduce new dependencies which are required by code updates
- Update existing dependencies with classes being renamed or moved
- Move existing template files

The config for the view won't exist until after config import happens, but won't be called by anything until the update hook runs. There is maybe the possibility of a very short outage based on this. I think ideally, we would manually import the views config during the update hook so that it is there when the update hook finishes running and sites that did not have related content turned on get switched to using the related content field by default. However, I'm not sure the length of the possible outage necessarily warrants the additional effort.

cory-skeers:
> I think ideally, we would manually import the views config during the update hook so that it is there when the update hook finishes running...

Even if we don't go back and add it here, that's a good point, and this is a good consideration going forward for future update hooks. Similar 'defensive coding' type practice.

PR #5662, opened 2022-09-07
Update field value naming for embargo check (now.uiowa)

cory-skeers:
Machine names were updated during some previous work, and we missed a spot.

# How to test

Sync now.uiowa.edu, enable the now split (`drush @now.local config:import --source ../config/sites/now.uiowa.edu --partial`)
Create a new article or edit an existing article.
Make the article type "News Feature"
Fill something in the Embargo Info field
Try to publish the page
Page should be set to Draft instead


bspeare (approved):
It works!

Issue #4839, opened 2022-02-21
Set per-site Drush cache config


richardbporter:
Blocked by #4789 

Create a new PR using #3383 as a guide.

The PR should:
- [x] Remove env variables being set in BLT tasks [591f8f9](https://github.com/uiowa/uiowa/pull/3383/commits/591f8f9b6bb717731ed287d56e9484269f7c6915)
- ~~Create a BLT update hook to set site-specific config for all multisites [91788d9](https://github.com/uiowa/uiowa/pull/3383/commits/91788d9bad7a66d7ce0fb4f370a5aeea6b18eadb)~~
- ~~Ensure the same file is created for new multisites in `blt umc` command.~~
- ~~Add a filesystem test for said file.~~
- [x] Remove unused drush config [51f32f7](https://github.com/uiowa/uiowa/pull/3383/commits/51f32f753f53060b3c71c9f3426dbec9dd27b9b7)
- [x] Investigate BLT environment detector usage and determine if there is a way to set AH vars in DDev without breaking stuff. That is, it'd be nice to expand the environment variables used in the drush config files locally. [e16569d](https://github.com/uiowa/uiowa/pull/3383/commits/e16569da74711e1a65c52cb5e689e130436ec7d6)
- [x] Deploy to develop and test code deployments while running Drush commands like `blt ume cr`

~~It would good to do this while also pausing on any new site provisions until it is merged.~~

richardbporter:
>Investigate BLT environment detector usage and determine if there is a way to set AH vars in DDev without breaking stuff. That is, it'd be nice to expand the environment variables used in the drush config files locally. [e16569d](https://github.com/uiowa/uiowa/pull/3383/commits/e16569da74711e1a65c52cb5e689e130436ec7d6)

This might not be possible: https://github.com/uiowa/uiowa/pull/3383#issue-861508526

joewhitsitt:
@richardbporter should I start holding on site provisions?

richardbporter:
> @richardbporter should I start holding on site provisions?

No. These files might not be necessary at all. Go ahead.

richardbporter:
>Investigate BLT environment detector usage and determine if there is a way to set AH vars in DDev without breaking stuff. That is, it'd be nice to expand the environment variables used in the drush config files locally. [e16569d](https://github.com/uiowa/uiowa/pull/3383/commits/e16569da74711e1a65c52cb5e689e130436ec7d6)

[...]

PR #4167, opened 2021-09-22
Clear each sites plugin cache to prevent missing layouts.

richardbporter:
Marking as resolves #3585 but it doesn't really. I'd still be ok with this closing #3585 - its still there for future reference.

### Testing
- `blt auda`
- Should see something like the below for each site:
```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/uiowa/vendor/bin/drush @self cc plugin --define='drush.paths.cache-directory=/tmp/.drush-cache-local/local/default' --no-interaction --ansi in /var/www/uiowa/docroot
 [success] 'plugin' cache was cleared.
```

PR #5676, opened 2022-09-09
Re-add drush cc plugin

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/3585

Following the git history of this file, I think the work originally done in #4167 can be simplified to what it is in this PR because we upgraded to Drush 11 and this work: https://github.com/uiowa/uiowa/issues/4839

# To Test
- `ddev blt auda`
- Should see something like the below for each site:
```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/html/vendor/bin/drush @self cc plugin --uri=admissions.uiowa.edu --no-interaction --ansi in /var/www/html/docroot
 [success] 'plugin' cache was cleared.
```


bspeare (approved):
This ran `cc plugin` during `ddev blt auda` on all the sites.

PR #5682, opened 2022-09-12
Events block - add checks after defensive coding meeting

joewhitsitt:
Similar testing from https://github.com/uiowa/uiowa/pull/5497 is needed to make sure date range functionality works on events blocks in v3


pyrello:
After looking at this and seeing all the `$block->get('field_whatever')` calls we are doing, I'm wondering if maybe we should doing something where we create an array of all the field names at the beginning and check if they all exist and then throw an error if they don't. That way we don't have to clutter up our code with a bunch of `if` checks throughout.

pyrello:
This would also be a good case for a test.

PR #5684, opened 2022-09-13
add language perms and defaults

### LINKED ISSUE #5543: Specifying language on News articles
<meta http-equiv="content-language" content="en"/>

is the wrong language on  

https://diversity.uiowa.edu/news/2022/06/isandra-martinez-marrero-nombrada-como-la-nueva-directora-de-participacion-cultural-y

## Proposed Solution

- Adjust the `Access Content Language in Advanced` permission which should make this available on a piece by piece basis. Default should be english.

GaryRidgway:
tracks https://github.com/uiowa/uiowa/issues/5543

## How to test
- sync a site (Probably diversity)
- try creating some content
- edit SEO settings > advanced > content language, putting a language code in.
- Does the rendered content have a content language meta tag, and does it say what you told it to?
- nice.
- Additionally, CEX changed the names of the `'Related'` headers for related content. should we keep that?

The only pitfall i can see here is that if someone messes up the langcode, it will still output it. For example, if someone puts `eng` instead of `en`, the rendered language will still be `eng`, even if no screen reader or crawler will recognize that as a valid language code.

### PLEASE PAY SPECIAL ATTENTION TO THE CONFIG FILES THAT I CHANGED IN THIS PR, I DID THEM MANUALLY SO HAVING THEM DOUBLE/TRIPLE CHECKED WOULD BE GOOD.


joewhitsitt:
pulling down to test

joewhitsitt:
Config looks fine, but I have a question out for discussion: https://github.com/uiowa/uiowa/issues/5543#issuecomment-1245985981

joewhitsitt (approved):
Approving to send this out for now. Alan is going to create a separate issue to investigate a long term solution to setting the lang attribute when this meta value is removed from Metatag module

PR #5688, opened 2022-09-13
phpcs - some fixes

joewhitsitt:
- Removed unused var
- adjusted use of var within `t()`
- changed `access administration pages` permissions check to `administer basic site settings` which only webmaster and above roles have
- keep it under 80

# To Test

- Checkout `main` and pull
- Run `ddev blt validate:phpcs` - see warnings.
- Checkout `random-phpcs-fixes`
- Run `ddev blt validate:phpcs` - see solutions.
- Sync a site and go to Region Settings and see that the h3s are still rendering as part of each fieldset.


docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -324,8 +324,9 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
               // it can be placed below the admin label.
               $form['settings']['layout_builder_style_menu_orientation'] = [
                 '#weight' => -10,
-                  // We set #tree to FALSE because this value will override the
-                  // value from this will override the original in the form state.
+                  // We set #tree to FALSE because
+                  // this value will override the value from this
+                  // will override the original in the form state.
```
cory-skeers: This fixes the warning, but I'm also confused by the wording. Is it supposed to be "We set the #tree to FALSE because this value will override the original in the form state" (without the extra '...will  override the value from this...")?

cory-skeers (changes requested):

cory-skeers:
Checked things out, saw the warnings get removed, and checked that the h3s are still rendering. Thanks for cleaning things up!

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: Looks like it originated here https://github.com/uiowa/uiowa/pull/4667. @GaryRidgway do you recall this?

GaryRidgway: I don't recall it right now, so my context is fuzzy. cory seems to have wording that makes sense to me though

cory-skeers (approved):
Thanks @joewhitsitt and @GaryRidgway!

PR #5689, opened 2022-09-13
Preserved link message feature

### LINKED ISSUE #5626: Broken link message feature for articles
As a SiteNow editor, I have a way to toggle a message to appear at the top or at the bottom of old news articles so that visitors are made aware of old links.

# Additional Info

The message should be editable and default to say something like, "Links in this article are preserved for historical purposes, but the destination sources may have changed."

This was addressed previously for the D7 Iowa Now website in https://github.com/uiowa/uiowa/issues/4079

# Proposed Solution

e-marie-w:
Resolves #5626 - As a SiteNow editor, I have a way to toggle a message to appear at the top of old news articles so that visitors are made aware of old links.

# How to test

- Pull down branch, sync a site. 
- Edit an existing article, check the Preserve Links box now available under the body field (do we like it there?)
- Save and view article, the default "Links in this article are preserved for historical purposes, but the destination sources may have changed." message should appear.
- Create a new article, see that the preserve link box is not available. Save and then edit it, it should now appear.
- While editing as a webmaster or administrator, click the link (that should open in a new tab) to change the message that displays. Change the placeholder text, save, check that the article now displays your custom text. You can also go back and empty that field, save, and see that the default message will reappear on articles. 
- Masquerade as editor or publisher, you should no longer see the link to change the message.
- Check that turning off the checkbox removes the message
- Repeat this process for iisc.uiowa.edu and sppa.uiowa.edu


docroot/modules/custom/sitenow_articles/src/Plugin/Block/PreservedLinksMessage.php
```diff
@@ -0,0 +1,41 @@
...
+    $markup = '<div>Links in this article are preserved for historical purposes, but the destination sources may have changed.</div>';
+
+    return [
+      '#markup' => $markup,
+      '#attributes' => [
+        'class' => [
+          'alert alert-info',
+        ],
+      ],
+    ];
```
cory-skeers: Can we move the `alert` and `alert-info` into separate, individual entries in the `class` array? I think it should work either way, but just for consistency.

I also think we could drop the `<div>...</div>` around the `$markup` message, because the block will be rendered in a div with the 'alert alert-info' classes.

cory-skeers (dismissed):

cory-skeers:
I think if the default message is set in config, we'll need to set it in an update hook--sitenow_articles.settings is config ignored since it gets used to set site-specific settings. That's why it wasn't showing up as config differences or ever getting imported, whereas the old 'preserved_links...' message was, because it was set as part of the settings form save.

So we'd want to set the default variable in there for new sites, but for existing sites we'll have a short update hook to set that value. (And continued defensive coding practices for the instances where the code may be there but the update hook has not run to set the config value.)

Edit: Update hook will look something like this:
```
/**
 * Set the newly created footer.login_link theme setting to TRUE.
 */
function sitenow_update_9034() {
  $config = \Drupal::configFactory()->getEditable('uids_base.settings');

  if (!$config->get('footer.login_link')) {
    $config->set('footer.login_link', TRUE)
      ->save();
  }
}
```

docroot/modules/custom/sitenow_articles/sitenow_articles.module
```diff
@@ -231,6 +231,16 @@ function sitenow_articles_preprocess_node(&$variables) {
               }
             }
           }
+          // Check if preserve links checkbox is checked.
+          if ($node->hasField('field_article_preserve_links') && ($node->field_article_preserve_links->value == 0)) {
+            // If it is, do not display the broken links message.
+            foreach (array_reverse($variables['content']['_layout_builder'][1]['content'], TRUE) as $key => $block) {
```
e-marie-w: This code is based on the gallery jump link block, which uses array_reverse because it expects the gallery link to be last. I'm not sure if that is the case here, or how I can tell.

cory-skeers: Only really found out via xdebug, but I believe as blocks get added to the node display, they get added to the end of the list. I'd expect this one to be in the last space, and now the gallery link to be second to last.

It's a small optimization where reversing the array and then traversing it should be slightly more efficient if the blocks are at the end, but if/when more blocks get added, it'll be less and less of a difference.

e-marie-w: Oh cool, thanks for figuring it out! I'll leave the code as-is then.

e-marie-w:
Follow up question about the update hook, does it need to be deployed off-hours? I know the ones in sitenow.install do, but this is in sitenow_articles.install instead so I don't know.

cory-skeers:
> Follow up question about the update hook, does it need to be deployed off-hours? I know the ones in sitenow.install do, but this is in sitenow_articles.install instead so I don't know.

Will want to take a closer look during the review to make sure, but from what we worked on before, it shouldn't need an off-hours deployment. In the time that the code is there but the update hook hasn't run, if the `$config->get('...')` code is hit, it'll just return a `NULL`, but nothing will break. Worst case, I believe, is someone would turn on the preserved link functionality on a page without setting a custom message, and the default message just won't display.

bspeare:
Does the text area on this need to be `Minimal HTML`?  This seems like `Basic HTML` might be a better place to start.

bspeare:
How do we want to handle these issues with the `sppa` and `iisc` splits?  
- `sppa` - checkbox does not appear on node edit screen.
- `iisc` - it does appear on form edit screen but checking it does not make the message appear on the saved page.

e-marie-w:
> Does the text area on this need to be `Minimal HTML`? This seems like `Basic HTML` might be a better place to start.

Yeah, I wanted the one with the least options so the user can't change the styling. It looks like that is actually Basic and not Minimal like I thought, will change.

> How do we want to handle these issues with sppa and iisc?

...forgot about the site splits, will investigate.

docroot/modules/custom/sitenow_articles/src/Plugin/Block/PreservedLinksMessage.php
```diff
@@ -0,0 +1,82 @@
...
+
+    return [
+      '#markup' => $markup,
+      '#attributes' => [
+        'class' => [
+          'alert',
+          'alert-info',
+        ],
+        'style' => [
+          'margin-bottom: 1.25rem',
```
bspeare: Instead of adding an inline style here, you should add a layout builder style to the block config https://github.com/uiowa/uiowa/blob/80be9cccfed3d754f994a2ddf812269a3cc42208/config/default/core.entity_view_display.node.article.default.yml#L232 like in https://github.com/uiowa/uiowa/blob/main/config/sites/admissions.uiowa.edu/core.entity_view_display.node.area_of_study.default.yml#L180-L181.

joewhitsitt:
Just a thought based on Mark's question at mid-sprint. Do we want to hide this on `new` and just show the field on `update`? It could still be populated during migrations but it would be less confusion for a new article being crafted

e-marie-w:
> Just a thought based on Mark's question at mid-sprint. Do we want to hide this on `new` and just show the field on `update`? It could still be populated during migrations but it would be less confusion for a new article being crafted

Oooh I like that idea! Going to change this back to draft again while I try to get that working.

bspeare (dismissed):

docroot/modules/custom/sitenow_articles/src/Plugin/Block/PreservedLinksMessage.php
e-marie-w: Now fixed! Took me a little too long to realize I needed to run `cim` to get it to take effect but I got there eventually!

bspeare: Thanks!

cory-skeers:
I'm wondering about the difference in formatting for the two config values (default and manually created preserved link messages), and where one is nested and the other not:

```
default_preserved_links_message_display: 'Links in this article are preserved for historical purposes, but the destination sources may have changed.'
preserved_links_message:
  display: ''
```
It's the difference between using `preserved_links_message.display` with the dot and `default_preserved_links_message_display` without the dot. 
Can we either do them both without the dot, or do them both nested under a parent like `preserved_links_message_display`? (For the latter, I think it would be something like `preserved_links_message_display.default` and `preserved_links_message_display.user_created` so in config it would be like:
```
preserved_links_message_display:
  default: 'default message and such'
  user_created: 'the editor-chosen message'
```
Or if we think there might be other config options in the future, it could be
```
preserved_links_message:
  display:
    default: 'default message and such'
    user_created: 'the editor-chosen message'
```
It's a small detail, but might make things easier in the future. If we add extra automated testing to the module, we'll probably have to update the config schema for this, too, but I think that can be a future slice as needed.

config/default/field.field.node.article.field_article_preserve_links.yml
```diff
@@ -0,0 +1,23 @@
...
+dependencies:
+  config:
+    - field.storage.node.field_article_preserve_links
+    - node.type.article
+id: node.article.field_article_preserve_links
+field_name: field_article_preserve_links
+entity_type: node
+bundle: article
+label: 'Preserve Links'
+description: 'Displays message that article links may be broken and are preserved for historical purposes. <a href="/admin/config/sitenow/uiowa-articles" target="_blank">Customize message</a>'
```
bspeare: Add `rel="noopener"` here too? https://github.com/uiowa/uiowa/pull/5696#issuecomment-1253998530

cory-skeers:
Gone through the testing and it's working great! (Also tested out on iisc and sppa).

> * Create article or edit existing article, check the Preserve Links box now available under the body field (do we like it there?)

I'm not sure about the placement. It feels a little out of place, but I'm not sure there's really a better spot for it.

With the link to the SiteNow Articles config page, I'm wondering if it's clear enough that the settings (preserved link message, but also the other settings) are the site-wide defaults? I think it's clear if you're used to the system, but I could see a user clicking the link and changing things here not realizing they're changing the site-wide settings and defaults, thinking they're changing them for only the page.

Also, the link is available to editors who don't have access to the page, leading to an Access Denied. Not sure if it's worth the extra work to do an access check beforehand to only display the link to those who have access to customize the message.

> * Note I personally don't love the placeholder text and would rather leave it empty, with the default message in the description text for the field. Or figure out that ghost text thing.

I think for the ghost text placeholder, it's not working because it's a formatted text area, and ckeditor would need to handle it in order to place it into the wysiwyg field. But I think this works alright--someone going to the page sees there's a message there, which matches what the expected behavior would be on the article end.

e-marie-w:
> It's the difference between using `preserved_links_message.display` with the dot and `default_preserved_links_message_display` without the dot. 

This is the code I tried to refactor that one day and got nowhere with. :joy: But the suggestions look promising! To me it'd be best without the dot but uh that's how I broke it last time, so we'll see.

>Also, the link is available to editors who don't have access to the page, leading to an Access Denied. Not sure if it's worth the extra work to do an access check beforehand to only display the link to those who have access to customize the message.

If it's not too complicated I think it's worth it.

>With the link to the SiteNow Articles config page, I'm wondering if it's clear enough that the settings (preserved link message, but also the other settings) are the site-wide defaults? 

I can specify that in the description with the link and see if that seems clearer.

config/default/field.field.node.article.field_article_preserve_links.yml
e-marie-w: Yep, can do!

cory-skeers:
> This is the code I tried to refactor that one day and got nowhere with. 😂 But the suggestions look promising! To me it'd be best without the dot but uh that's how I broke it last time, so we'll see.

I forgot about that! I think I came in at the end of that troubleshooting. I updated it to `preserved_links_message_display` (alongside `default_preserved_links_message_display` that I didn't change), and it seems to be working as expected. I'm guessing there was something else going on before that's since been changed where this works now.

e-marie-w:
> > This is the code I tried to refactor that one day and got nowhere with. 😂 But the suggestions look promising! To me it'd be best without the dot but uh that's how I broke it last time, so we'll see.
> 
> I forgot about that! I think I came in at the end of that troubleshooting. I updated it to `preserved_links_message_display` (alongside `default_preserved_links_message_display` that I didn't change), and it seems to be working as expected. I'm guessing there was something else going on before that's since been changed where this works now.

IT WORKS thank you @cory-skeers!! Not sure what was going on before but I don't even care I'm just glad it works now.

joewhitsitt:
Pulling down to test

joewhitsitt:
In terms of defensive coding (code firing before update hook and config import), this needs some adjustments:

`Warning: Attempt to read property "value" on null in sitenow_articles_preprocess_node() (line 246 of modules/custom/sitenow_articles/sitenow_articles.module).`

joewhitsitt:
Though I am still reviewing (sorry, been distracted), moving this back to draft until we can fix https://github.com/uiowa/uiowa/pull/5689#issuecomment-1259797195

joewhitsitt:
gosh darn this is my first priority

joewhitsitt:
Am I the only one concerned about this functionality being used for other things?
Screen Shot 2022-09-30 at 10 23 10 AM

Would we combat this by providing some kind of heading or footer label describing what this is?

joewhitsitt (approved):
Besides my concern here https://github.com/uiowa/uiowa/pull/5689#issuecomment-1263715196, this PR looks good to me.

pyrello:
> Am I the only one concerned about this functionality being used for other things?

Counterpoint: maybe we are just being too limited in our thinking about what we are creating. Maybe this is useful for other things besides talking about old links and we just haven't thought it through enough.

docroot/modules/custom/sitenow_articles/config/schema/sitenow_articles.schema.yml
```diff
@@ -14,3 +14,9 @@ sitenow_articles.settings:
     show_teaser_link_indicator:
       type: boolean
       label: Show arrows linking to pages from lists/teasers
+    preserved_links_message_display:
+      type: text
+      label: Specified preserved links message
+    default_preserved_links_message_display:
```
pyrello: I'd like to see this be renamed to `preserved_links_message_display_default`

pyrello:
I'd like to make a few minor changes to this.

joewhitsitt:
>Maybe this is useful for other things besides talking about old links and we just haven't thought it through enough.

Keep in mind that https://github.com/uiowa/uiowa/issues/5726 will be layered on top.

bspeare:
If we fast-tracked https://github.com/uiowa/uids/issues/770, we could have the permanent link icon attached to this with a heading?

bspeare:
I agree, without a heading and an icon this field could be repurposed.

bspeare:
Or we could set a custom design like I mentioned when this first started: https://github.com/uiowa/uiowa/issues/5626#issuecomment-1247279899

pyrello:
> I agree, without a heading and an icon this field could be repurposed.

I guess I'm just not that concerned about it. Maybe some creative individual will help me understand the error of my thinking.

pyrello:
@joewhitsitt I made some changes. I ran through test instructions again and it looks good to me. Do we want another run through on testing this or is it good?

joewhitsitt:
Good catch on the isset. I have reviewed the code changes. I trust ya.

PR #5692, opened 2022-09-14
update better normalizers

joewhitsitt:
Resolves `ddev composer update --lock` issue on the main branch after updating to drupal 9.4.5 - https://iowaweb.slack.com/archives/C0164FNN1PV/p1663163812322409

I _think_ this is used for default content. I don't think there is much to test here. It is just aware now that HAL has move from core to contrib.


pyrello (approved):
Reviewed, not tested.

PR #5696, opened 2022-09-14
Added general_help_text option to the list block and added some default help text to list blocks.

### LINKED ISSUE #5672: Links to manage people from the people block
# Story

As a user of SiteNow, I'd like to be able to manage people or add a person directly from the People block. I've been in several trainings with partners with a lot of people content (student government, multicultural programs) where a clear path to manage person content while adding or modifying a person block would be extremely useful. 

We've heard direct feedback that it's currently a bit cumbersome and confusing to users that they can't directly add or modify instances of the person content type when editing a person block.

# Proposed solution

A simple solution would be adding a bit of descriptive text in the People block that would link to: `/admin/content?title=&type=person&status=1` and `/node/add/person` :

Screen Shot 2022-09-09 at 10 51 31 AM

quamsta:
Resolves #5672

This draft PR adds some general help text to the list blocks' configuration panes that I think could be useful for user experience and also will make navigating between editing the blocks and the content within the blocks a lot quicker. 

Screen Shot 2022-09-14 at 11 58 11 AM

Screen Shot 2022-09-14 at 11 58 27 AM

We've bumped up against folks not having a clear idea about how to manage local people and articles that show up in list blocks. The most common area for confusion is that the user can't directly add a person or post when they edit a block containing these items.

I think this confusion arises from the fact that they're clicking on a post or a person and they expect to be able to edit these items in the same way that clicking a text area  (or any other block like a card, etc) allows them to edit text.

See:

Screen Shot 2022-09-14 at 11 54 58 AM

vs

Screen Shot 2022-09-14 at 11 54 48 AM

While this PR doesn't give them the ability to actually manage people or other content from the block itself, it does provide a front-and-center path to getting to those areas of Drupal and allows them to quickly add a person or article if they're attempting to edit them from the block itself.

Open to feedback or other ideas to improve this.

# How to test

1. Check out this branch
2. On a local site of your choosing, import the configuration: `drush @sandbox.local csim`
3. Add or edit an article list block and note the help text
4. Add or edit a person list block and note the help text


pyrello:
One issue I see here is that when the user sees these links, they are in the midst of a form (list block) inside a form (layout builder). We probably would need to ensure that the link opens up in new tab/window so that their progress is not lost. Opening a new tab/window is not totally ideal either because then the person ends up with two tabs/windows and they have to remember to go back to the first one to finish what they were doing. It seems like it would increase the possibility of someone losing some of their work by accident if they closed the first tab before finishing their update to the block. It wouldn't be much work lost - their previous changes to Layout Builder would be saved in the temp store, but any work they had done on the block would be lost.

quamsta:
> One issue I see here is that when the user sees these links, they are in the midst of a form (list block) inside a form (layout builder). We probably would need to ensure that the link opens up in new tab/window so that their progress is not lost. Opening a new tab/window is not totally ideal either because then the person ends up with two tabs/windows and they have to remember to go back to the first one to finish what they were doing. It seems like it would increase the possibility of someone losing some of their work by accident if they closed the first tab before finishing their update to the block. It wouldn't be much work lost - their previous changes to Layout Builder would be saved in the temp store, but any work they had done on the block would be lost.

Great points. I think opening the links in new tabs would likely be enough to mitigate losing changes. These blocks aren't often heavily modified, and if the user's trying to add or modify the posts contained within them, they're not likely to start messing with the block's options since they're actually trying to edit the items contained within the lists. 

Basically, they're trying to do one thing or the other, not both.

However, to prevent closing tabs when the configuration pane is open in a general sense, would investigating something like the `beforeunloadevent` (https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) to provide a warning message before closing the tab be worthwhile?

quamsta:
It looks like there is some precedent for having links (that open in new tabs) in these configuration panels: 

Screen Shot 2022-09-16 at 9 24 10 AM
Screen Shot 2022-09-16 at 9 23 05 AM

joewhitsitt:
To be fair, we don't have open in new tab for several help text links (non-core like the above)

pyrello:
> To be fair, we don't have open in new tab for several help text links (non-core like the above)

My understanding is that this is a UX best practice to prevent losing data already entered into the form if the page is redirected and that it overrules the accessibility best practice of opening links in the same window/tab to prevent someone from being launched into a new window/tab context. @mjoneill Does this sound correct?

If that is the case, I would consider the existing help text links that don't open in a new window to not meet our best practices in this regard and we should consider opening issues to fix them.

quamsta:
> > To be fair, we don't have open in new tab for several help text links (non-core like the above)
> 
> My understanding is that this is a UX best practice to prevent losing data already entered into the form if the page is redirected and that it overrules the accessibility best practice of opening links in the same window/tab to prevent someone from being launched into a new window/tab context. @mjoneill Does this sound correct?
> 
> If that is the case, I would consider the existing help text links that don't open in a new window to not meet our best practices in this regard and we should consider opening issues to fix them.

@pyrello I've opened an issue containing some of the links I've found in the `/config` directory here: https://github.com/uiowa/uiowa/issues/5715

In the meantime, I've added `target="_blank" rel="noopener"` to these links and I believe this PR is ready for review! Thanks all.

docroot/modules/custom/layout_builder_custom/src/Plugin/Display/ListBlock.php
```diff
@@ -41,6 +41,8 @@ public function optionsSummary(&$categories, &$options) {
       'configure_filters' => $this->t('Customize filters in block'),
       // Add use_more option summary.
       'use_more' => $this->t('Display more link'),
+      // Add general_help_text field.
+      'general_help_text' => $this->t('General help text'),
```
pyrello, line 45: I think this should be an empty string since the default value could slip through somewhere.

bspeare:
@quamsta what's your thoughts on this from https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener?

Screen Shot 2022-09-21 at 10 35 34 AM

bspeare:
Should we change the link behavior of https://github.com/uiowa/uiowa/blob/main/config/default/views.view.people_list_block.yml#L981 too?  It seems weird to have different behavior on the links within this block.   I understand you have marked it as future work on https://github.com/uiowa/uiowa/issues/5715, but maybe we make the change here to have a consistent experience for this block?

quamsta:
> @quamsta what's your thoughts on this from https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener?
> 
> Screen Shot 2022-09-21 at 10 35 34 AM

@bspeare 

It's good that newer versions of browsers implicitly provide the same effect without needing `rel="noopener"`, but the idea behind putting it there anyway is that we can't guarantee that the user has an up-to-date browser that does this by default. In the past it's been best practice to include `rel="noopener"` for users with older or out of date browser versions, aka versions of browsers  below the version numbers listed in this table:

Screen Shot 2022-09-21 at 12 12 50 PM

Safari 12.1, for instance came out in 2019, so it's really not totally infeasible for a user to be stuck on an older version of a browser, although it'll become more of an edge case as time goes by. 

I say better safe than sorry! The extra attribute doesn't really harm anything, imho.

quamsta:
> Should we change the link behavior of https://github.com/uiowa/uiowa/blob/main/config/default/views.view.people_list_block.yml#L981 too? It seems weird to have different behavior on the links within this block. I understand you have marked it as future work on #5715, but maybe we make the change here to have a consistent experience for this block?

@bspeare

Yeah! That sounds completely reasonable to me. I'll make that adjustment and update the PR

quamsta:
@bspeare PR updated with the `more_link_text` given the `target="_blank" rel="noopener"` attributes.

quamsta:
@bspeare, I also added  `target="_blank" rel="noopener"` to article_list_block's more_link_help_text as well since it's also in the same file as the general help text.

bspeare:
I've tested this out and I'm happy to approve it unless anyone else wants to review it too.

joewhitsitt:
A thought from demo that might aid this issue (not replace this) is to make contextual links work for those listblock items. Right now I think it contexual links the media entity but not the node. That way when they see a typo in the first article they can go directly there to fix it.

quamsta:
> A thought from demo that might aid this issue (not replace this) is to make contextual links work for those listblock items. Right now I think it contexual links the media entity but not the node. That way when they see a typo in the first article they can go directly there to fix it.

Yes! Getting into the individual nodes via the block is exactly the sort of thing I'm aiming for with this. Just didn't know what was possible with those contextual links or even if such a thing was possible.

bspeare (approved):
Thanks for making this improvement!

PR #5711, opened 2022-09-16
z-index fix for the sticky header for uiowa.edu

bspeare:
This pr moves the sticky header to `z-index: 3` so that it sits above the banner play button. 

# How to test

- `blt frontend`
- `ddev blt ds --site=uiowa.edu`
- Scroll down/scroll up on the home page until the banner play button overlaps with the sticky header. 

**Before**

Screen Shot 2022-09-16 at 3 17 01 PM

**After**

Screen Shot 2022-09-16 at 3 18 58 PM

### DIFF
--- a/docroot/themes/custom/uids_base/package.json
+++ b/docroot/themes/custom/uids_base/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "ISC",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


e-marie-w (approved):
Works great, banner play button no longer appears on top of sticky header. And I followed the UIDS commit to see that the only change is the z-index being set to 3.

bspeare:
Thanks @e-marie-w !

PR #5611, opened 2022-08-24
PHP 8.1 Upgrade

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/5315
Resolves https://github.com/uiowa/uiowa/issues/5708

# To do

<details>
<summary>History</summary>

- [x] profiles/custom/sitenow @joewhitsitt
- [x] themes/custom/uids_base @pyrello
- [x] modules/custom/engineering_migrate
- [x] modules/custom/grad_custom @pyrello 
- [x] modules/custom/layout_builder_custom @pyrello 
- [x] modules/custom/sitenow_articles @joewhitsitt
- [x] modules/custom/sitenow_dcv @joewhitsitt - _no changes_
- [x] modules/custom/sitenow_dispatch @joewhitsitt
- [x] modules/custom/sitenow_events @joewhitsitt
- [x] modules/custom/sitenow_find_text @joewhitsitt - _no changes_
- [x] modules/custom/sitenow_intranet @joewhitsitt - _no changes_
- [x] modules/custom/sitenow_media_wysiwyg @joewhitsitt  
- [x] modules/custom/sitenow_migrate @joewhitsitt
- [x] modules/custom/sitenow_p2lb @joewhitsitt
- [x] modules/custom/sitenow_pages
- [x] modules/custom/sitenow_paragraphs @joewhitsitt
- [x] modules/custom/sitenow_people @joewhitsitt
- [x] modules/custom/sitenow_seo @joewhitsitt - _no changes_
- [x] modules/custom/studentlife_migrate @joewhitsitt - _no changes_
- [x] modules/custom/studentlife_topics @joewhitsitt - _no changes_
- [x] modules/custom/uiowa_aggregator @joewhitsitt
- [x] modules/custom/uiowa_alerts @joewhitsitt
- [x] modules/custom/uiowa_apr @joewhitsitt  - _no changes_
- [x] modules/custom/uiowa_area_of_study @joewhitsitt
- [x] modules/custom/uiowa_book @joewhitsitt - _no changes_
- [x] modules/custom/uiowa_core @joewhitsitt
- [x] modules/custom/uiowa_entities @joewhitsitt 
- [x] modules/custom/uiowa_graph @joewhitsitt 
- [x] modules/custom/uiowa_hours @joewhitsitt 
- [x] modules/custom/uiowa_maui @joewhitsitt 
- [x] modules/custom/uiowa_profiles @joewhitsitt 
- [x] modules/custom/uiowa_search @joewhitsitt - _no changes_
[...]

PR #5713, opened 2022-09-19
Allow page node overrides for regions

### LINKED ISSUE #5669: As a site owner, I can have a row of buttons that appears fixed to the bottom of the browser and configure it per page
#4767, https://github.com/uiowa/uiowa/pull/5576

Make "after content" configurable on a per-page basis. This is for "Page" content type only, and only for "After Content".

- [x] Add a set of reference fields (one per region that can be overridden) that directly references the region fragment type. We could also potentially allow people to create alternate region fragments inline.
- [x] Update the RegionContentBlock class to check if a page is a node and has override fields set and not empty, in which case it would pull the appropriate fragment from the node instead.
- [ ]  ~~Add a global setting to allow displaying the new region as a fixed area at the bottom of the view port.~~
- [ ] ~~Optional depending on how the previous slice gets rolled out: Update hook to allow setting the global setting to false for any sites that started adding region fragments to this new area.~~
- [x] Preprocess to set a class on the body/footer if there is an after content block to set extra bottom padding https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/footer.scss#L8

### Next Slice
- [ ]  https://github.com/uiowa/uiowa/issues/5685
- [x] https://github.com/uiowa/uiowa/issues/5693

GaryRidgway:
tracks https://github.com/uiowa/uiowa/issues/5669

# How to test
1. `ddev blt ds`
2. `yarn workspace uids_base build && ddev drush @default.local cr && ddev drush @default.local uli`
4. Go to the region settings page and create two region items per region.
5. Check that on the page node edit form that the region overrides are in the sidebar in a section called "Region overrides" under "Relationships", and that the help text links you to the region settings page.
6. In all combinations, set global and per page region items and determine that the overrides work, and that if you have no globals and there are overrides, that the regions will show on only that specific page.
7. Additionally, whenever you don't have an after content, you SHOULD NOT see extra padding on the `.footer` element, and if you do have an after content item, you SHOULD see extra padding.
8. In addition to the additionally, make sure that the back to top button works with all combinations.
9. Check the above combinations logged out.


bspeare:
@GaryRidgway this is something we missed with the last round.  Do you want to test this with the back to top button and position the button similarly to https://viewbook.admissions.uiowa.edu/?

GaryRidgway:
@bspeare I will check that, thanks!

bspeare:
@GaryRidgway I'm noticing that the markup on an overridden after content region is different.  

Global region:

Screen Shot 2022-09-21 at 11 25 38 AM


Override:

Screen Shot 2022-09-21 at 11 27 37 AM

bspeare:
I also set up a bunch of button rows for testing that you can find on https://sandbox.prod.drupal.uiowa.edu/admin/config/sitenow/region-settings.

GaryRidgway:
Thanks ben! I'll take a look at these

GaryRidgway:
I'm having trouble grabbing an entire region's render array in code, so I am a little stumped :(

GaryRidgway:
Okay so after a bit more consideration on this I was able to match the markup a bit deeper, however one of the benefits of the way that this was being constructed was that this was cache busting the cached rendered version of the global after content. If i place this any deeper, it will not bust the cached version, and i think the display is still just as good, so maybe this is good enough?

docroot/themes/custom/uids_base/templates/layout/html.html.twig
```diff
@@ -33,6 +33,16 @@
     layout_builder_enabled ? 'layout-builder-enabled' : 'layout-builder-disabled',
   ]
 %}
+{# If we have after content that is renderable... #}
+{% set after_content_rendered = page.after_content | render %}
+{% if after_content_rendered|striptags('<embed><hr><iframe><img><input><drupal-render-placeholder>')|trim %}
+
+  {# Add the after content spacing class to the footer_classes. #}
+  {% set body_classes = body_classes|merge([
```
bspeare: Are we able to do this in https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L75 like how we are adding the other body classes?

config/default/field.field.node.page.field_after_content_override.yml
```diff
@@ -0,0 +1,29 @@
...
+  config:
+    - field.storage.node.field_after_content_override
+    - fragments.fragment_type.region_item_after_content
+    - node.type.page
+id: node.page.field_after_content_override
+field_name: field_after_content_override
+entity_type: node
+bundle: page
+label: 'After content override'
+description: 'Override the sitewide sticky <i>After content</i> region for this page. <a href="/admin/config/sitenow/region-settings">Manage or create <i>After content</i> items.</a>'
```
bspeare: Do we want to put `target="_blank" rel="noopener"` here?

GaryRidgway: Good idea, I was concerned about the items not populating, but then i remembered that it is ajax fetching the items, so it's all good!

GaryRidgway: Just added this

docroot/themes/custom/uids_base/templates/layout/html.html.twig
GaryRidgway: perhaps, and maybe that would make the footer template, which has similar functionality, better as well.

GaryRidgway: I will check after lunch

GaryRidgway: Wow, this was much more complicated than we expected, but we got it done! Yeehaw! Hope that works for you, it was weird.

GaryRidgway:
IT'S ALIVE

makurj:
Just noticed in demo: Help text for pre footer has the word "sticky" so that needs to be removed as pre footer is not sticky.

makurj:
Following steps from https://github.com/uiowa/uiowa/pull/5611 to test this again.

makurj:
Looks good to me. I'd ask someone else to test this too.

pyrello:
I'm reviewing this from a code perspective.

pyrello:
@GaryRidgway The code updates for this are spread out over `uiowa_core.module`, `sitenow.profile`, and `uids_base.theme`. Is there are reason why this can't all be contained in `uiowa_core.module`?

GaryRidgway:
No real reason why not, I think I just found the similar functionality in those places so they seemed like the place to add them at the time. I can refactor if we want.

pyrello:
Testing this.

pyrello:
I did a bit of refactoring here to try to further simplify the footprint of this feature update. Since the RegionContentBlock class is our own, I think it makes sense for us to encapsulate all the logic about how the block gets rendered there rather than only handling the global setting case and then using a preprocess function to handle the override case.

> Okay so after a bit more consideration on this I was able to match the markup a bit deeper, however one of the benefits of the way that this was being constructed was that this was cache busting the cached rendered version of the global after content. If i place this any deeper, it will not bust the cached version, and i think the display is still just as good, so maybe this is good enough?

@GaryRidgway I'm curious about the mechanism of how this is getting cache busted and how you were verifying that? Normally, I would expect to see something manipulating some cache artifact as part of this update for that to be the case.

@joewhitsitt As our resident expert on caching, I would love your input on how we should be testing this to make sure that it is working as expected for a logged out user viewing a version of the page served by the load balancer.

pyrello:
Okay, there was a problem with caching in my solution. 

I think I understand now how the cache busting was working because the section of the render array where the caching information for the block was stored was getting overwritten when there was an override. Obviously if we are handling all the rendering in `RegionContentBlock::build()`, that doesn't happen.

To resolve this, I found a module that has a similar use case to ours (block displays content based on a config setting or something about the current node) and stole the cache-related code: https://www.drupal.org/project/block_node. It seems to work now. I would still like to have @joewhitsitt validate my testing assumptions.

GaryRidgway:
This looks good from a code perspective and an intent perspective, @pyrello has some super great points for the move. I have not tested this on a site yet though s some more testing should be done before approval

joewhitsitt:
>As our resident expert on caching, I would love your input on how we should be testing this to make sure that it is working as expected for a logged out user viewing a version of the page served by the load balancer.

First off. Lies. Second, to be sure, we usually deploy to DEV so we can confirm that Varnish doesn't get in the way.

Do you want me to review this PR in that way?

pyrello:
> we usually deploy to DEV so we can confirm that Varnish doesn't get in the way.
>
> Do you want me to review this PR in that way?

@joewhitsitt I can merge this into DEV and I'll let you know when that is done.

pyrello:
> @joewhitsitt I can merge this into DEV and I'll let you know when that is done.

This is done and in the process of building and deploying.

joewhitsitt:
I will take a look at this in DEV

joewhitsitt:
https://sandbox.dev.drupal.uiowa.edu/about-1 with a after/pre-footer override set has a cache age of 0 and varnish is a MISS.

https://sandbox.dev.drupal.uiowa.edu/child-page with a global after/pre-footer is a HIT for both varnish and drupal.

Something with the page override is causing it never to be cached. We don't want this.

joewhitsitt:
Going to pull this down locally to see if we can make tweaks to the caching here.

pyrello:
> Something with the page override is causing it never to be cached. We don't want this.

This is probably related to the caching logic that I added to the custom block class.

joewhitsitt:
>This is probably related to the caching logic that I added to the custom block class.

I think for the most part, this was a good addition. Not sure if `route` is the best context here though. Is that `entity.node.canonical` with the nid passed? I wonder if `url` would be better since it has the unique identifier baked in.

I am starting to believe that my original test was not properly cached because I had a webform as a pre-footer (not cachable due to spam protection). I am getting some cache age now: https://sandbox.dev.drupal.uiowa.edu/about

With that said, I am having troubles with general content edits triggering the necessary purging for an anonymous session. Just making an edit to the body field is not returning what I expect navigating between pages. Instead I need to hit the refresh button to get it to fetch.

joewhitsitt:
Aside from caching, I suggest adding to the help text on Region Settings that the global settings can be overridden on a per page basis.

pyrello:
> Instead I need to hit the refresh button to get it to fetch.

Is that something we have control over? Or is that a browser caching thing?

joewhitsitt:
>Is that something we have control over? Or is that a browser caching thing?

I assumed that making an edit to a page, clears site cache for the page and through the purge stuff sends a note up to varnish that the next time the page is hit, it will bypass varnish and the cleared drupal cache and request the page. This doesn't always seem to be the case. Maybe we have a regression with our recent purge module update

joewhitsitt:
Disregard my comment about switching route to url. Good point made here: https://www.drupal.org/project/fieldblock/issues/2646070#comment-10992417

pyrello:
> > Is that something we have control over? Or is that a browser caching thing?
> 
> I assumed that making an edit to a page, clears site cache for the page and through the purge stuff sends a note up to varnish that the next time the page is hit, it will bypass varnish and the cleared drupal cache and request the page. This doesn't always seem to be the case. Maybe we have a regression with our recent purge module update

It is my understanding the browsers cache assets in a way that would prevent a new version of the asset from being retrieved without a hard refresh. I've run into this before with redirects. I don't know that this would be the case in this situation, but wondered if it might.

pyrello:
I can merge these conflicts.

joewhitsitt:
Reviewing one more time as a whole to get this to done.

joewhitsitt:
- I don't think this is a blocker but potentially a follow-up issue. The back to top button ends up being half way up the page with a large footer and sticky after content. Those that don't notice it on the way down might confuse it with a prefooter cta. Could it ride on the top footer line?

joewhitsitt (approved):
I'd like to see some in-PROD testing of the caching but the rest seems to check out. Shouldn't merge until we are ready for the scheduled deployment

PR #5742, opened 2022-09-28
Iowa now contact

### LINKED ISSUE #5675: Ability to list a contact for articles and people on Iowa Now
# Story

As an Iowa Now site editor, I would like the ability to reference an individual contact or organizational contact when creating an Achievement or News Feature article type and also when creating a person.

# Additional Information
They do not want to migrate existing contacts, they will start from scratch with these.

Organizational contact should allow for a contact name, email, phone. Not sure if dept is needed.

Existing examples:
https://now.uiowa.edu/2020/08/longevity-awards-march-2020
https://now.uiowa.edu/person/karim-abdel-malek

# Proposed solution

- Create content type to handle organizational contact
    - Name, email, phone
- Add ability to reference this new content type as well as a person from Article types Achievement and News Feature
- Add ability to reference this new content type as well as a person from person content type
- Render on article node as a link to entity
- Render on person node as a link to entity

### LINKED ISSUE #5741: IowaNow - Embargo field errors within generateSampleValue
Screen Shot 2022-09-27 at 11 39 36 AM

The line in question is `$values['value'] = $random->word(mt_rand(1, $field_definition->getSetting('max_length')));`

_Originally posted by @GaryRidgway in https://github.com/uiowa/uiowa/issues/5729#issuecomment-1259788951_

```
ValueError: mt_rand(): Argument #2 ($max) must be greater than or equal to argument #1 ($min) in mt_rand() (line 75 of /mnt/www/html/uiowa03/docroot/core/lib/Drupal/Core/Field/Plugin/Field/FieldType/StringItem.php) #0 /mnt/www/html/uiowa03/docroot/core/lib/Drupal/Core/Field/Plugin/Field/FieldType/StringItem.php(75): mt_rand(1, NULL)
```

# How to reproduce
- Go to Iowa Now's article display settings

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/5675
https://github.com/uiowa/uiowa/issues/5741

## How to test
- Sync now.uiowa.edu
- your config should import
- You should be able to add a contact, either a person or a contact, to a person or an article
- When you add a new person from the IEF, you should see a name in the table
- yeehaw


config/sites/now.uiowa.edu/config_split.patch.core.entity_view_display.node.article.default.yml
```diff
@@ -1,6 +1,83 @@
...
+  dependencies:
+    config:
+      - field.field.node.article.field_article_type
+      - field.field.node.article.field_contact_reference
+      - field.field.node.article.field_embargo_information
+      - field.field.node.article.field_original_publication_date
+  third_party_settings:
+    layout_builder:
+      sections:
+        -
```
pyrello: I think we need to change the split type of this config to be full. Because we don't have UUID's for sections, there is no way to guarantee that this will work correctly. See https://www.drupal.org/project/drupal/issues/3208766#comment-14284826

GaryRidgway:
A couple questions that Brian brought up today.

1. First, Our current person content type has a "Contact info" field that is a paragraph field, and then adding this "Contact" field in addition to that might cause confusion. Should the field be renamed? Should we only allow one with states?
Screen Shot 2022-09-30 at 10 58 50 AM

2. Second, should the "Contact" field be an Inline Entity Form? Currently it references both people and contact content types, not sure how that would work with this, but that was a question that might be needed to be answered.

joewhitsitt:
I think `Point of Contact` would be an option for differentiating from the person contact field.

IEF would be consistent with the Author field and there shouldn't be an issue with the difference between person and contact references. Contact will probably need a "minimal" form setup similar to person (e.g. basic detail and publish state)

GaryRidgway:
looking at the way the field gets set up, i would have to choose one of these options when doing an IEF

Screen Shot 2022-09-30 at 2 53 25 PM

GaryRidgway:
i don't like this help text message when we are removing the reference, we are deleting people instead of removing them.
Screen Shot 2022-09-30 at 3 06 41 PM

GaryRidgway:
~~Oh my, the field ordering Q_Q~~


<details>
<summary>Solved with moving the moderation state field within the form display and creating a minimal form view!</summary>

</details>

GaryRidgway:
Blargh, when creating a contact and using a "Contact" content type, it uses the table headings for "Person"

Screen Shot 2022-09-30 at 3 11 35 PM

cory-skeers:
Pulling this down to do some testing

cory-skeers:
Are we concerned if points of contact are added and then the article type is switched to something that shouldn't have a point of contact? We added some logic before to handle changing tags if the type is changed, and this could be extended to remove points of contact if the type doesn't match.

Example:
Create a new article that is a 'news feature', and add a point of contact.
Switch the article type to 'in the news', and the point of contact field is no longer visible.
Save.
See the originally added point of contact on the final node display.

The workaround is to change the article type back to news feature or achievement, remove the point of contact, then change the article type back to whatever other type is needed.

@briand44 thoughts if we need to account for this?

cory-skeers:
I updated the article form display config to a partial split, because I don't think it needed to be a complete split (unlike the display). In testing, it still seems to work correctly, but if there was a reason I'm missing that it needed to be complete, I can change it back.

briand44:
@cory-skeers My first reaction is that I am not too concerned if they add a contact and switch to another type. Aren't the other types linking directly to a source anyway so they would never get to the full node?

cory-skeers:
> @cory-skeers My first reaction is that I am not too concerned if they add a contact and switch to another type. Aren't the other types linking directly to a source anyway so they would never get to the full node?

Ooh, that's a good point. It'd be the 'in-the-news' and 'ui-spotlight' article types, so yes, I think you're right and they're intended to link elsewhere.

Thanks!

cory-skeers:
> I updated the article form display config to a partial split, because I don't think it needed to be a complete split (unlike the display). In testing, it still seems to work correctly, but if there was a reason I'm missing that it needed to be complete, I can change it back.

Ditto with the Person form displays. I think they can be partial splits, and updated them.

cory-skeers:
Holding off on giving an official 👍 in case there was other work in progress, but from my testing I think this is ready to go.

briand44:
I wouldn't mind someone giving me a quick demo

cory-skeers:
Couple notes from demoing with @briand44:

When creating a new Person contact, the form doesn't yet have a Title to display
image
Once the node is saved, the newly created person gets a node title, and upon re-editing the page, will display their Contact Name.
One option that was brought up here was to change the fields displayed here to include the name fields, even if they're blank for referenced Contacts. Though, I just realized, that in that case you'd have the following possible displays:

- A Contact will always have Contact Name filled and empty First, Last Names
- A New Person will have no Contact Name filled, but filled First, Last Names
- An existing Person will have Contact Name filled with the title (First Name Last Name), as well as the First, Last Name fields

Without getting into some funky processing, I think we might be stuck deciding between a temporarily empty Contact Name for New Person additions, or an overly full Contact, First, Last Name fields for existing Person additions. The latter's not bad, it's just not very elegant.

**Other Note:**
When adding a Contact you have the Name, Phone, and Email fields. When adding a Person, you only have the First, Last names. Can we add the Phone Email fields here as well, as they are the primary fields that would be wanted when adding a new point of contact. (It's okay if this also adds these fields to the Add an Author form as well.)

GaryRidgway:
I will pull down and start working on these enhancements

GaryRidgway:
Just added an alter for the form table, so now it creates a title from the individual fields that matches the title that is created. My code works, but I am worried about its implementation, could anyone add some ideas for refactoring it to be more robust? Additionally, I added the email and phone fields for new people

cory-skeers:
> Just added an alter for the form table, so now it creates a title from the individual fields that matches the title that is created. My code works, but I am worried about its implementation, could anyone add some ideas for refactoring it to be more robust? Additionally, I added the email and phone fields for new people

Tested generally, and it looks great! Still want some more thorough testing on it than I could provide yet, and check out the code, but initial check it looks promising.

joewhitsitt:
Pulling down to test

joewhitsitt (approved):
Looks great. Made a tweak or two.

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
```diff
@@ -41,6 +79,30 @@ function now_core_form_alter(&$form, FormStateInterface $form_state, $form_id) {
...
+ *   The fields, keyed by field name.
+ * @param array $context
+ *   An array with the following keys:
+ *   - parent_entity_type: The type of the parent entity.
+ *   - parent_bundle: The bundle of the parent entity.
+ *   - field_name: The name of the reference field on which IEF is operating.
+ *   - entity_type: The type of the referenced entities.
+ *   - allowed_bundles: Bundles allowed on the reference field.
+ *
+ * @see \Drupal\inline_entity_form\InlineFormInterface::getTableFields()
```
pyrello: Is this necessary? Or could this be `{@inheritdoc}`?

joewhitsitt: Shortened it to

```
/**
 * Implements hook_inline_entity_form_table_fields_alter().
 */
```

Does that work?

joewhitsitt:
Test comment

Issue #4393, opened 2021-11-03
Provide print.css for general browser printing


joewhitsitt:
# Story

As a site visitor I want to be able to print a page on SiteNow v3 and have that output be fairly clean and readable.

Though entity print has been made available for certain content, a general print style can serve most other content.

# Proposed solution

- [ ] Establish print.css to be available on all pages.
- [ ] Style certain content to have a decent print style fallback. (header, footer, body fields, etc.)
- [ ] Hide other elements that might not be necessary for print (background images/colors, etc.)

bspeare:
Can we close this?  I think https://github.com/uiowa/uiowa/pull/5753 forgot to add the resolve label.

joewhitsitt:
yeah

PR #5753, opened 2022-09-29
Feature print.css for browser printing

bewoldt:
For issue #4393 

Adds styles for print.

Notes:

- removes unnecessary elements like navigation, pagination, iframes, search, etc...
- hides all remote videos since they don't print
- all background colors are removed to use less ink
- all borders are #7c7c7c so they are easier seen when printing
- simplified header and footer
- url paths are shown on links

# How to test

Most browsers have a way to emulate a print stylesheet but will not be an exact match for print. Instructions on browser emulators here: https://css-tricks.com/can-you-view-print-stylesheets-applied-directly-in-the-browser/.

# Resources
https://www.smashingmagazine.com/2018/05/print-stylesheets-in-2018/

# Before

# After


joewhitsitt:
I haven't had a chance to test this yet, but is the plan to start with this foundation and then cover more complex elements in follow-up issues? (layouts, region content, timeline block, etc.)

joewhitsitt:
pulling to test

bewoldt:
I've gone through most out-of-the-box elements to make sure they are readable. There's always going to be things to improve and other custom templates I haven't covered.

The portrait mode when printing uses a breakpoint that's below 768 so everything uses the mobile stylesheet which can make elements like cards really large.

joewhitsitt:
I think this is a good initial step for v3.

Should more be done with link color? Related to the underlines?
Screen Shot 2022-10-12 at 10 29 32 AM

Should more `break-inside: avoid;` be added for blocks?
Screen Shot 2022-10-12 at 10 29 54 AM
Screen Shot 2022-10-12 at 10 47 12 AM

bspeare (approved):
This looks like a great foundation for us to continue to improve upon.   Thanks @bewoldt for taking this on!  Let's get this deployed to the service so we can improve the printed experience.

PR #5754, opened 2022-09-30
Careers split for aggregator expire setting

### LINKED ISSUE #5614: Purge careers aggregator items after 1 month rather than 1 year.
# Story

As an owner of the Career Center website, I would like to only keep aggregator items for 1 month rather than the current setting of 1 year, so that the number of items in our feeds doesn't get overwhelming and start causing performance issues and pages to quit working.

# Additional Info

Career Center is using feeds aggregator functionality to display job opportunities for students. They have ~25 different feeds being consumed, some of which have had as many as 7,000 items stored in our db. This has caused a page to break on their site because of memory issues.

A potential long term solution would be to discard/purge items that no longer exist in the feeds source but that is not currently an available option with the aggregator module.

law requested these be kept for 1 year rather than our original setting of 6 months to handle feeds that contain an archive (https://apps.its.uiowa.edu/dispatch/archives/past/--iowa-law---september-25-lxj144j).

# Proposed Solution

- [x] Create a site split for Careers
- [x] Adjust config to limit aggregator items to 1 month
- [x] Truncate aggregator items if it chokes.
- [ ] #5807

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5614

# To Test

- Checkout `careers_split`
- `ddev drush @careers.local uli admin/config/services/aggregator`
- See lots of items.
- `ddev drush @careers.local config:import --source ../config/sites/careers.uiowa.edu --partial && ddev drush @careers.local cim && ddev drush @careers.local cron`
- See less items after refresh. https://careers.uiowa.ddev.site/admin/config/services/aggregator

### DIFF
--- a/config/sites/careers.uiowa.edu/.htaccess
+++ b/config/sites/careers.uiowa.edu/.htaccess
@@ -1,0 +1,27 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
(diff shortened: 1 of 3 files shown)


cory-skeers (approved):
Looks good, and worked as expected. 👍

PR #5770, opened 2022-10-03
P2LB Paragraph ID -> Block ID

### LINKED ISSUE #5246: P2LB - Map paragraph ID to block ID for custom block types in LB
With #3567 done, we now have the ability to map the unique IDs from paragraphs to blocks.

# Proposed solution
Modify the p2lb module to set the unique block ID based on the corresponding paragraph ID.
We will address list blocks, sections, accordion items, etc in a follow up issue.

cory-skeers:
Resolves #5246 

# How to test

## Main testing
`git checkout p2lb-ids && git pull`
`blt ds --site=hawkeyemarchingband.uiowa.edu` (or another v2 site)
`drush @hawkeyemarchingband.local config-split:activate p2lb`
`drush @hawkeyemarchingband.local uli`
Try converting some pages that have unique ids set at the paragraph level (not the section level). (Links to the converter can be found in various places; direct URL for hawk band: https://hawkeyemarchingband.uiowa.ddev.site/admin/content/sitenow-converter)
Check that the corresponding blocks have the matching unique id from their previous paragraph

Note: This should be applicable for paragraphs that have a specific mapping to a block, such as text area, card, etc. It does not yet apply for sections or interior pieces, like individual accordion items (for instance, with accordions, unique ids applied to the overall accordion should be present on the collection block after converting, but unique ids applied to individual accordion items are not transferred).
Webforms, which were a single paragraph in v2, are converted into two blocks in v3: a text block for the title, and the webform block. The unique id, if available from the v2 paragraph, should be attached to the text block title

## Faux banner behavior testing
There is a process that converts sections which contained a section image a text area in the first paragraph position (more paragraphs may or may not exist below it) into a banner. This is done through a slightly different process and needs to be tested separately:

On a page, make a section with an attached section image and at least one text area, with a unique id attached to the text area.
Convert the page
Check the corresponding banner block, and see that it has the unique id from the text area

## OOP changeup in section breaking
Create a section with 3 items, the first two with 'Fluid' widths set, and the last with 'Full' set.
Convert the page
Check that all three items are available in the layout (the first two should be put into one section, and the third in a section on its own)
In other testing, if a block is ever not visible in the layout, note the paragraphs in its section along with their set widths

## Extra testing
While working on the site, check that you properly get the "This page can be converted..." and "This page has been converted..." messages on the edit screen, in an appropriate location at the top (ie not repeating within paragraph edit forms)


cory-skeers:
Capturing here: If it's a quick fix, can add it in here, but if it needs some investigation, should open a follow-up issue to address later.

When opening/collapsing paragraphs editing a v2 page with the p2lb split enabled, the "This page can be converted..." message can sometimes end up duplicated in various spots on the page.

Example (there's another repetition further down the page, not shown here):
image

cory-skeers:
Setting back to draft to check on a few things.

cory-skeers:
Found a bug in cases where there are multiple fluid items in the same section as a full width item; taking a look at it.

joewhitsitt:
pulling to test

joewhitsitt:
checking again (post computer crash) but the id for my image gallery didn't come over. it did for text, blockquote and card.

confirmed: image gallery isn't working

cory-skeers:
Hrm...I was able to convert a page and see the id successfully added to an image gallery that I added.

Were you converting an existing page I could test on (just to know we have all the same settings)? I assume there's something in the section setup or combo of field settings that's not playing nice, but am not sure what might be causing that.

joewhitsitt:
I synced law down and added an image gallery using existing images to the homepage. because of stage file proxy, the image styles are broken, but it seemed like it was a successful conversion until I inspected the unique id

cory-skeers:
I added a quick gallery to law, and am seeing the id on it after conversion:
image

On an unrelated note to IDs, this might need some looking at 😬 (banner created out of a text and section image. The pieces are fit together in their right spots, but there's some funky encoding from moving between text fields)
image

joewhitsitt:
okay. i can try a fresh sync here shortly. might just be me. i did notice that other issue, figured it was out of scope and potentially already captured somewhere.

joewhitsitt (approved):
It worked this time. I think what happened was I was quick to add it and the check/sanitize stuff didn't finish before i clicked save in v2. didn't inspect that it actually saved in v2 and then converted it to v3.

PR #5813, opened 2022-10-04
Iowa Now experts list

### LINKED ISSUE #5670: Iowa Now experts list
# Story

As an Iowa Now website owner, I would like to be able to create a person list that has an exposed filter for areas of expertise so that media and other site visitors can quickly identify an expert by their area of expertise.

# Additional Information

Existing page - https://now.uiowa.edu/experts-list

Would like to be able to display the areas of expertise in the teaser view. They do not need the ability to link to each area of expertise. It sounded like we would prefer they use the website field to enter the url for the person's canonical profile page and then link the teaser to that page rather than viewing the full node on this site? If we go that route we would need to be able to display photo, first, last, email, phone, positions, areas of expertise and contact in the teaser. We may be able to avoid migrating the person's about info since it doesn't seem to be used on the staff listings https://now.uiowa.edu/staff-list and I doubt we want to output that in a teaser. They would lose the ability to link to a related article on these experts profiles.

Do we need a person type to differentiate between an expert and Iowa Now staff or all experts faculty? Could we just make this determination based on if the person has areas of expertise listed?

# Proposed Solution
- Custom view display to expose filter for research areas and output research areas in teaser display.
- Move "Research Areas" to display in the teaser at /admin/structure/types/manage/person/display/teaser.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5670

# How to test

- `ddev blt ds --site=now.uiowa.edu`
- Add a research area to https://now.uiowa.ddev.site/admin/structure/taxonomy/manage/research_areas/overview
- Add a person profile with Title/position, email, phone, research areas and add the tag "expert"
- Go to https://now.uiowa.ddev.site/experts-list and verify that the above fields are listed in a card teaser format
- Try searching for something that is unavailable and check out the error message.


cory-skeers:
Should people without any Research Area associated with them show up on this list as well (prior to filtering to any specific research area)?

Also, it's working correctly, but is there a separate setting to change it from an error to only displaying the no matching results message instead when a non-area is entered? (In this example, "Things" is the only available Research Area, but did a quick type/enter for "biology")
image

bspeare:
Thanks @cory-skeers!  The view is updated showing only people with the tag "expert" and I switched the exposed filter to select instead of autocomplete because that error message is pretty bad.  I didn't find a setting for changing that yet.

cory-skeers:
> Thanks @cory-skeers! The view is updated showing only people with the tag "expert" and I switched the exposed filter to select instead of autocomplete because that error message is pretty bad. I didn't find a setting for changing that yet.

Thanks! Pulled it down and tested, and I think everything looks good! I'd found one setting that I thought was related to the error message, but either I was wrong and it's about something else, or changing it doesn't work... I think the select list works great, though.

cory-skeers (approved):
👍

bspeare:
@cory-skeers I changed it back to autocomplete and added `hook_form_views_exposed_form_alter` to alter the error message.  Let me know if you have any suggestions to improve that!

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
```diff
@@ -41,6 +42,16 @@ function now_core_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   }
 }
 
+/**
+ * Implements hook_form_views_exposed_form_alter().
+ */
+function now_core_form_views_exposed_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $view = $form_state->getStorage('view');
+  if ($form_id == 'views_exposed_form' && in_array($view['view']->id(), ['iowa_now_experts'])) {
+    $form_state->setErrorByName('field_person_research_areas_target_id', t('There are no matching results.'));
```
cory-skeers: I think this is marking the form as errored with any input, even if there are matching results. Instead, can we check if there's an error (and probably check what kind of error it is, too), and if it's there, alter the displayed message?

cory-skeers (changes requested):

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
bspeare: Thanks @cory-skeers!

docroot/themes/custom/uids_base/uids_base.info.yml
```diff
@@ -1,6 +1,6 @@
 name: UIDS Base
 type: theme
-base theme: stable
+base theme: classy
```
pyrello: lol

joewhitsitt:
@bspeare do you want help resolving the conflicts?

bspeare:
@joewhitsitt thanks, I can fix them quick.

bspeare:
Moving this back to draft.  I'm going to try to update the template a bit this morning.

bspeare:
@pyrello I've updated the view to use the field template files for the icon instead of within the view template file.

GaryRidgway:
I will pull to test

GaryRidgway:
Based on the testing instructions, this works, however the reset button does not reset the field.

GaryRidgway:
okay, tested again, PEBKAC?

GaryRidgway (approved):

docroot/sites/now.uiowa.edu/modules/now_core/templates/views-view-unformatted--iowa-now-experts.html.twig
```diff
@@ -0,0 +1,34 @@
...
+    {% endblock %}
+
+    {% block card_meta %}
+      {% if person_email|striptags|trim is not empty %}
+        {{ person_email }}
+      {% endif %}
+      {% if person_phone|striptags|trim is not empty %}
+        {{ person_phone }}
+      {% endif %}
+    {% endblock %}
```
pyrello: @bspeare I'm wondering if you are doing this to avoid the `<div class="card__details"><div class="card__meta"></div></div>` markup around these elements? Otherwise, see my suggestion for `now_core.module`.

docroot/sites/now.uiowa.edu/modules/now_core/now_core.module
```diff
@@ -258,3 +280,59 @@ function now_core_article_type_tag_map(string $article_type = '') {
...
+      $variables['rows'][$id]['card_options'] = [
+        'attributes' => $variables['rows'][$id]['attributes'],
+        'media_styles' => ['media--circle', 'media--border'],
+        'media_sizes' => ['card__media--small'],
+        'card_image' => $style->getField($id, 'field_image'),
+        'card_title' => $style->getField($id, 'title'),
+        'card_link_url' => $link,
+        'card_subtitle' => $style->getField($id, 'field_person_position'),
+        'person_email' => $style->getField($id, 'field_person_email'),
+        'person_phone' => $style->getField($id, 'field_person_phone'),
```
pyrello: This is related to my comment about the view template. If we don't care about having the extra markup associated with using the existing `{% block card_meta %}`, the I believe that both of these element should already be render arrays that will print out correctly if set as the value of `card_meta`. I'm not exactly sure, but I think you could do something like this:
```suggestion
        'card_meta' => [
          'email' => $style->getField($id, 'field_person_email'),
          'phone' => $style->getField($id, 'field_person_phone'),
        ],
```

bspeare:
I removed our custom code and the reset button still doesn't work.  Do we want to remove the option to reset or move to the select list option is long but doesn't have this issue?

joewhitsitt:
Since it is only one filter field, not sure if the reset button is that helpful/needed. I think we can pitch the select list as an alternative to the Iowa Now team.

bspeare:
I've removed the reset button and updated the template and module to follow @pyrello's suggestion https://github.com/uiowa/uiowa/pull/5813#discussion_r991708892.   Can someone take a look at these changes and approve?

GaryRidgway:
I will pull and test

GaryRidgway (approved):
reviewed again for a little while and checked over code. Everything looks pretty reasonable to me!

cory-skeers (approved):

PR #5824, opened 2022-10-07
Update smart_date

### LINKED ISSUE #5664: Use smart_date min and max field settings for events block's range field.
# Story

As a developer, I can re-configure the smart date range field on the events block to use the new min/max field settings.

# Proposed solution (updated 10/11/2022)

- ~~Create separate issues to address undefined array keys for min/max and next hour default value when max is set. https://www.drupal.org/project/smart_date/issues/3300768#comment-14726415~~ being handled in the same issue.
- [x] Submit patches to the module for the two issues.
- [x] Test and implement next stable release of smart_date with the functionality above.
- [x] Update the the range field on the events block and remove the override code from layout_builder_custom.

### LINKED ISSUE #5804: PHP error for smart_date
`Deprecated function: Calling static trait method Drupal\smart_date\SmartDateTrait::isAllDay is deprecated, it should only be called on a class using the trait in Drupal\smart_date\Plugin\Field\FieldWidget\SmartDateWidgetBase->massageFormValues() (line 285 of modules/contrib/smart_date/src/Plugin/Field/FieldWidget/SmartDateWidgetBase.php).`

# Steps to reproduce
Try setting "original publication date" on a "in the news" article type https://now.prod.drupal.uiowa.edu/news/2022/07/ui-researcher-wins-funding-study-regional-air-quality. 

# Proposed solution
- Update to `3.6.x`?
- https://github.com/uiowa/uiowa/issues/5664

joewhitsitt:
resolves #5804

error was fixed in 3.6, min-max is coming out in 3.6.2 so holding off #5664 for now.

# To Test 

- `ddev composer install`
- Try setting original publication date (in the news) on https://now.uiowa.ddev.site/news/2022/07/ui-researcher-wins-funding-study-regional-air-quality
- event features (admissions), events block (v3).

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,15 +100,16 @@ You can run `composer update package/name` to update additional dependencies. Th
 ### Locked Packages
 The packages below are locked at specific SHAs and will not update using the method described above. They should be periodically checked for new stable releases and updated, if viable.
 
-| Package                               | Reason                   |
-| ------------------------------------- | ------------------------ |
-| acquia/blt-travis                     | No stable release to pair with blt 13.5 |
-| drupal/diff                           | Need php 8.1 compatibility, no stable release |
-| drupal/lb_direct_add                  | No 2.x stable release.   |
-| drupal/media_thumbnails_video         | Need php 8.1 compatible dependencies, no stable release |
-| kartsims/easysvg                      | Need https://github.com/kartsims/easysvg/pull/27 which is not included in a release. |
-| uiowa/block_content_template          | Forked from a deprecated project. |
-| drupal/purge                          | Need php 8.1 compatibility fixes, no stable release |
+| Package                       | Reason                                                                              |
+|-------------------------------|-------------------------------------------------------------------------------------|
+| acquia/blt-travis             | No stable release to pair with blt 13.5                                             |
+| drupal/diff                   | Need php 8.1 compatibility, no stable release                                       |
+| drupal/lb_direct_add          | No 2.x stable release.                                                              |
+| drupal/media_thumbnails_video | Need php 8.1 compatible dependencies, no stable release                             |
+| kartsims/easysvg              | Need https://github.com/kartsims/easysvg/pull/27 which is not included in a release. |
+| uiowa/block_content_template  | Forked from a deprecated project.                                                   |
+| drupal/purge                  | Need php 8.1 compatibility fixes, no stable release                                 |
+| drupal/smart_date             | No stable release for min/max field settings.                                       |
 
 
 # Redirects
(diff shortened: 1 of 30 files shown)


joewhitsitt:
Seeing an issue on Now's article form `Undefined array key "#type" in Drupal\Core\Form\FormHelper::processStates()`

makurj:
Choosing "In the News" and entering an "Original Publication Date" gives this error on save. If I refresh, it goes away, but the moment I go to Edit > Save (even if I'm not making any changes to the 2 fields), the error returns.

image

joewhitsitt:
@makurj I wasn't explicit with the test instructions. did you run `ddev composer install` for this module update PR before starting?

makurj:
Just did it now. Testing again.

makurj:
Now: error disappeared. Works as expected.
Admissions: error disappeared. No issue creating a single event or event series.
Default: event block works. Events block with upcoming, past and date range works.

makurj:
Could someone also test this from a deployment perspective? I'd like to pair when this is done, if possible.

joewhitsitt:
>Could someone also test this from a deployment perspective?

What do you mean?

makurj:
Second reviewer who would code review, like we've been doing recently with PRs.

pyrello (commented):
I code reviewed this with an eye towards potential issues with a deployment causing downtime. Sorry that you were not able to pair on this, @makurj, but here is my assessment in case its helpful:
- I'm assuming that all the configuration changes correspond to things that are changing in the smart_date module.
- Since the smart_date module is already on for the service, the code updates associated with the module update will happen immediately at the beginning of the deployment.
- The associated config updates will not get loaded until each site runs.
- Without specifically reviewing the code changes in the module, if there are any places where the updates need the new configuration and don't gracefully fail, then that could potentially be an issue that would cause a white screen.
- This is potentially mitigated by the code being cached in the database. Until a cache rebuild happens, the old code should be in effect. Although I'm not totally clear on what gets cached in the DB... all the code or only some of it. Generally, it is hard to know how much we can count on this because the cache can be cleared right after a deployment happens, canceling any benefit we would have gotten from it.
- We can potentially test around this by doing the following:
  - `git checkout main && git pull && ddev composer install`
  - Sync the relevant site for testing
  - `git checkout smart_date_update && git pull && ddev composer install`
  - Test the relevant functionality on the site without running any config imports or database updates. This is the state of the application we are worried about encountering. If you don't find anything worrisome in this testing, then I would say that we don't have to worry about scheduling a deployment.

makurj:
> git checkout main && git pull && ddev composer install
Sync the relevant site for testing
git checkout smart_date_update && git pull && ddev composer install

I already did this, so I should be able to go ahead and approve, correct?

joewhitsitt:
>I already did this, so I should be able to go ahead and approve, correct?

I'm biased, but I would say so. ;)

makurj (approved):
Tested functionality.

Issue #5810, opened 2022-10-04
ISWF - Probe using events, event series and people to see what can be reused vs what needs to be custom


pyrello:
# Motivation
As part of #5732, we need to understand what existing pieces we can use that pertain to building out the functionality for the Iowa Summer Writing Festival.

# Proposed solution
- [x] Investigate the Event feature, what could be used and what would need to be modified.
- [x] Investigate the Event Series content type that exists in the Admissions split to see how similar/different it is from from the requirements of ISWF.
- [x] Investigate the Person content type to see what modifications would need to be made to it to use it as the basis for instructors on ISFW.
- [x] Make a recommendation to the team about how to proceed.
- [x] If you make code updates that are potentially useful to your proposed path forward, put them into a PR for future use.

briand44:
Wanting to draw attention to the difference between the online festival and in person festival mentioned in the update in https://github.com/uiowa/uiowa/issues/5732#issuecomment-1263915168

pyrello:
> Wanting to draw attention to the difference between the online festival and in person festival mentioned in the update in [#5732 (comment)](https://github.com/uiowa/uiowa/issues/5732#issuecomment-1263915168)

I made a comment with an idea for how to address the apparent discrepancy: https://github.com/uiowa/uiowa/issues/5732#issuecomment-1273913210

briand44:
Lets keep in mind other potential conference type websites in the future.

joewhitsitt:
Without digging yet, this seems more and more like schema.org's EventSeries > Event setup with "superEvent" and "subEvent" relationships.

Olympic Games (EventSeries) > The Rio 2016 Summer Olympics (Event with superEvent=Olympic Games, subEvent=Rowing). Then Rowing (Event with superEvent=The Rio 2016 Summer Olympics and subEvent=Rowing Semi-Final) and onward. So I think if there can be some flexibility in what we call the buckets, we should be able to capture the content for ISWF as well as several other use-cases.

https://schema.org/EventSeries

mark-bennett-uiowa:
If we have the potential to use this same structure for other conferences, festivals, or super-events on campus, do we want to consider a structure to work with something like a DrupalCon? Where you have the top level conference/festival/super-event, etc. Within that you have a Monday session, a Tuesday session, a Wednesday session, etc each with their own start/end date/time. Within that you have workshops with their OWN start time / end time, but their start date / end date is dependent on the session that contains them, or include a start date / end date for sessions as well?

[...]

PR #5869, opened 2022-10-19
Event feature infused with schema.org

### LINKED ISSUE #5879: SiteNow events parent/child functionality
Capturing from internal-demo:

- [x] - CLAS approval to move forward with single content type parent/child representation and event type classification.
- [x] - Keep reference field to the parent event (superEvent) and removing the child event (subEvent) reference field.
    - [x] Remove add new event?
- [x] - Views block to list out child events chronologically based on parent relationship.
- [x] - Rip out the schema.org subtype field and module (schemadotorg_subtype). Instead move forward with event category to designate festival, session, workshop for now by changing it to a single value/dropdown and consider moving it out from the sidebar. Remove sitenow_events_event_types_options_allowed_values
- [x] - Update Series Link label to Source Link (communicate this change to be more generic with Scott)
- [x] - Make sure admissions, oniowa can accept changes (no config import issues)
- [x]   - Change the schema prefix to `field_` to match our field naming convention.
- [x] - [Alter json-ld](https://github.com/uiowa/uiowa/blob/99fe091eb33c0e3a0ab76591c3a4252331f0fec3/docroot/modules/custom/sitenow_events/sitenow_events.module#L659) to include subEvent info of all child events (view results).
    - [ ] Bare minimum has been done here, is there more to do? 
- [x] - Basic styling based on new content being output (reference to other events, performer, event categories).
- [x] - Update the event path to include the nid to aid with multiple events with the same name. - > Capturing eventID for pathing, e.g. /event/12345/human-readable-title - https://github.com/uiowa/uiowa/issues/5810#issuecomment-1282972378
- ~Currently blueprint modules are on for default. Decide if these can still be here or if they and the related config should move to the event feature for now.~ - https://github.com/uiowa/uiowa/issues/5879#issuecomment-1292291442

Separate issue: making person content type more flexible to single line value name or offer another content type to handle multiple types of event performers.

Scope this to events feature and not iswf split in this slice.

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/5810
Resolves https://github.com/uiowa/uiowa/issues/5879

# To Do

- [x] Remaining styling (sub events enclosed and spanning full width)
- [x] Add subEvent urls to json+ld (https://mysite.uiowa.edu/node/123)
- [x] Parent event field label
- [x] Finalize path
- [x] Check onIowa and Admissions for compatibility
- [x] Change schedule heading to serif bold

# Follow up

- [ ] Deal with 404 caused by path pattern /event/123/event-title ( /event/123 -> /node/123 -> /event/123/event-title)
  - #5910 
- [ ] Add Event>eventStatus field (e.g. Status) for "active", "cancelled", or "rescheduled" events, attach to eventStatus property from schema???
  - #5939
- [ ] Add Thing>eventAttendanceMode field (e.g. Attendance Mode) for "online", "onsite" or "mixed/blended"???
  - #5940
  - consider adding this in preparation of an accepted property decision, considering this would be good to have

# To Test

- `git checkout schema_event`
- `ddev drush @default.local sql-drop -y && ddev drush @default.local cr`
- `ddev composer install && ddev yarn install && ddev yarn workspace uids_base build`
- `ddev blt di --site=default`
- login, enable the event split and `cim` twice.
- For the default site you should be able to create a parent event and then separately create sub events that link to the parent. The parent event should show the sub events in chronological order (nearest first). There is also a Presenter's field that allows referencing multiple people. There should no longer be an event categories field/taxonomy on the website.
- If you were to sync oniowa.uiowa.edu and admissions.uiowa.edu locally, you should see no import errors. For admissions, you should still be able to assign multiple event categories (sidebar) filter event by category on /events and in the events list block on the homepage.
- Admissions should not see the "Part of" or "Presenters" fields on their event form.
- Inspecting the HTML source, you should see a secondary json+ld script tag that contains event information if those fields are filled out. You should see superEvent if parent event is filled out, subEvent if sub events exists and a performer property if presenters is filled out. Additionally, location should include both address and virtual link if provided.
- Copy/pasting the inner source of the schema markup, you should get valid markup on https://validator.schema.org/
- Linking directly to source for events still works as expect (title and card body linked to the same URL on events list block and sub events view, admissions /events and that internal/external links work)


joewhitsitt:
Screen Shot 2022-10-20 at 11 38 27 AM
Screen Shot 2022-10-20 at 11 38 34 AM
Screen Shot 2022-10-20 at 11 40 52 AM

Fixed description and added support for event subtype including adding ISWF specific session and workshop types

GaryRidgway:
There is a problem with our logic for icons in locations that will make them display at all times.
Ex:
Screen Shot 2022-10-26 at 3 54 28 PM
we should probably fix the logic that outputs these to make sure that they only output when there is actual content to display with them.

config/default/config_split.config_split.event.yml
```diff
@@ -18,12 +18,12 @@ complete_list:
   - node.type.event
   - pathauto.pattern.event
   - simple_sitemap.bundle_settings.default.node.event
-  - taxonomy.vocabulary.event_categories
   - views.view.events
   - views.view.events_list_block
   - 'core.entity_form_display.node.event.*'
   - 'core.entity_view_display.node.event.*'
   - 'field.storage.node.field_event_*'
+  - 'field.storage.node.event_*'
```
pyrello: Wondering if this is necessary? Not seeing any files that match this pattern. Seems like it might be a typo, since fields always get prepended with `field_`.

joewhitsitt: yeah, that should probably be removed.

bspeare:
I think for the icons usage, we can use some of font awesome's built-in utility classes so it could look like:

```
<div class="card__meta fa-ul">
<div class="field"><span class="fa-li"><i role="presentation" class="fas fa-fw fa-map-marker-alt" aria-hidden="true"></i></span>Location field</div>
<div class="field"><span class="fa-li"><i role="presentation" class="fas fa-fw fa-desktop" aria-hidden="true"></i></span>Virtual field</div>
</div>
```

This allows to get rid of `card__meta--icon` and `card__icon` classes.  Also note the `fa-fw` class on the icons to give them a fixed width.

config/default/config_split.config_split.event.yml
joewhitsitt: i think in my first pass with the blueprints module, i forgot to include the field_ and packaged them up that way.

joewhitsitt: https://github.com/uiowa/uiowa/pull/5869/commits/afd49016d557c3111c9fd109a65cf7da07baf7c1

docroot/modules/custom/sitenow_events/sitenow_events.module
```diff
@@ -36,6 +36,29 @@ function sitenow_events_form_alter(&$form, FormStateInterface $form_state, $form
...
+        $contacts = $form['field_event_performer']['widget']['entities'];
+
+        // For each one that exists, make sre it is one of the integer entries,
+        // which is actually an entity.
+        foreach ($contacts as $key => $contact) {
+          if (is_int($key) && $contact['#entity']->getType() === 'person') {
+
+            // Then, get the string of the first and last names.
+            $first = $contact['#entity']?->get('field_person_first_name')?->getString();
+            $last = $contact['#entity']?->get('field_person_last_name')?->getString();
```
pyrello: A couple things here:
- If `$contact['#entity']?->get('field_person_first_name')?->value` gets just the string value, I think that is what we are using in other cases where we are pulling this field value.
- I like to have these chained methods on separate lines like this:
```php
$contact['#entity']
  ?->get('field_person_first_name')
  ?->value;
```

docroot/modules/custom/sitenow_events/sitenow_events.module
```diff
@@ -36,6 +36,29 @@ function sitenow_events_form_alter(&$form, FormStateInterface $form_state, $form
...
+
+            // And set the title, which would have been empty,
+            // to the concatenation of the two.
```
pyrello: I'm not sure I'm going to understand what this comment means in a year when we are going through the code and trying to decide whether this functionality needs to be here or not. It might be good to add a longer comment at the beginning of this whole section describing why the section is being added.

joewhitsitt:
We are trying to get a full-width card for the sub events but the views-element-container div contains a lot of card and media classes that are getting in the way. We can't figure out where these are being set and how best to clean them up _after_ they are used to construct the individual row items. `card--media-left` in particular is the sticking point at which, if removed, we get a full width card.

Screen Shot 2022-10-31 at 12 04 25 PM

pyrello:
> We are trying to get a full-width card for the sub events but the views-element-container div contains a lot of card and media classes that are getting in the way. We can't figure out where these are being set and how best to clean them up _after_ they are used to construct the individual row items. `card--media-left` in particular is the sticking point at which, if removed, we get a full width card.

These would be coming originally from the LB Styles settings on view block in the LB template. They might be getting added to this specific HTML element through additional processing. We went through last week and added this block's ID to quite a few blocks of code being used for events that would need to be reviewed.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -1347,6 +1347,12 @@ function layout_builder_custom_preprocess_views_view_unformatted(&$variables) {
         'card_date' => '',
         'headline_level' => $view->display_handler->getOption('heading_size'),
       ];
+
+      // Only if the events card_list view display is used (sub events).
+      if ($view->id() === 'events' && $machine_name === 'card_list') {
```
bspeare: Can you set this on the block config for the event page content type (https://github.com/uiowa/uiowa/blob/schema_event/config/features/event/core.entity_view_display.node.event.default.yml#L283)? Or is there something else you are trying to do?

bspeare:
> We are trying to get a full-width card for the sub events but the views-element-container div contains a lot of card and media classes that are getting in the way. We can't figure out where these are being set and how best to clean them up _after_ they are used to construct the individual row items. `card--media-left` in particular is the sticking point at which, if removed, we get a full width card.

Yes, there should be no card classes on the views-element-container.

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: Maybe the enclosed but the config options weren't there. Needed this handle this content type specific field mapping though.

joewhitsitt: >but the config options weren't there

Correction: The block config is not currently moving card--enclosed down to individual items
Screen Shot 2022-10-31 at 1 06 18 PM

joewhitsitt:
I guess my main question is why `layout_builder_custom_preprocess_views_view_unformatted` hasn't been an issue for this yet (classes left in)

bspeare:
> I guess my main question is why `layout_builder_custom_preprocess_views_view_unformatted` hasn't been an issue for this yet (classes left in)

I think this could be the first list block that we are placing in a layout for use like this?  If you place this block through layout builder on a Page, these extra classes are not added in.

bspeare:
These classes are being rendered on the card and overridden by the classes being set from layout builder styles: https://github.com/uiowa/uiowa/blob/schema_event/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L1262-L1263.   Perhaps we can remove these two lines?  Inspect the card markup on https://sandbox.prod.drupal.uiowa.edu/people-examples for an example.

bspeare:
I'm noticing these new fields appearing for admissions, do we want to include/exclude these from the admissions split?

Screen Shot 2022-10-31 at 2 55 01 PM

mark-bennett-uiowa:
@joewhitsitt @pyrello @bspeare I was looking for where to add the schema.org addition of subEvent URLs for the child events. On the schema.org page for the event content type, there isnt a property for subEvent URL (even when looking at unmapped properties as well. It looks like `/docroot/modules/contrib/schema_metatag/data/schemaorg-all-http.jsonld` does some mapping for subEvents on line 4535, but `/Users/mbennett/Sites/uiowa/docroot/modules/custom/sitenow_events/sitenow_events.module` has the `sitenow_events_schemadotorg_jsonld_schema_type_entity_alter()` function altering the schema as well. Should I be editing the sitenow_events.module file?

Also, what property do we want to use? There isnt a property for SubEvent URLs that I can find in schema.org.
- [keywords](https://schema.org/keywords) and [location](https://schema.org/location) are URLs in [Event](https://schema.org/Event).
- [additionalType](https://schema.org/additionalType), [identifier](https://schema.org/identifier), [image](https://schema.org/image), [mainEntityOfPage](https://schema.org/mainEntityOfPage), and [sameAs](https://schema.org/sameAs) from [Thing](https://schema.org/Thing) all can hold URLs.

ALSO, while we are in the schema, I noticed we have the [description](https://schema.org/description) and [disambiguatingDescription](https://schema.org/disambiguatingDescription) properties. Should we be adding the main event description text to description, and the short description field text to disambiguatingDescription?? Currently, the short description field is in the description property.

Curious how best to handle this.

joewhitsitt:
@mark-bennett-uiowa we are programmatically putting together subEvents because instead of reference field we are using a view to display these so the suite of modules doesn't wire it up for us. sitenow_events is currently a mix up of code that we have issues to clean up. For now, UI events and the local event content type feature fire from the same module.

https://github.com/uiowa/uiowa/pull/5869/files#diff-af2c7ba26f825b10d8a8016e8191ac2bb8092e13fcf39d415c4996241415ae78R685 is the line where you will want to add the `url` property of the "Event" schema type. The $sub_event object is already available. You can construct the node/id URL from the ID there.  

>short description field is in the description property

This is correct. This is the field_teaser we currently use for teasers and SEO until we move back to using body summary fields (another, clean up/fix task in our backlog)

joewhitsitt:
@bspeare 

>I'm noticing these new fields appearing for admissions, do we want to include/exclude these from the admissions split?

I guess I assumed we would ask if they want to accept new "default" stuff. If not, I would imagine we would update their split to no import these new fields.

joewhitsitt:
>Change "Performers" label

I thought we wanted to start to adopt schema.org terminology for the default here and customize for ISWF if they are unwilling to go along?

pyrello:
> > Change "Performers" label
> 
> I thought we wanted to start to adopt schema.org terminology for the default here and customize for ISWF if they are unwilling to go along?

I'm not sure how I feel about adopting schema.org terminology for user-facing labels. The terminology is sometimes awkward and confusing because of the need for it to be generalized. While there are likely some potential users of events that would find the term "Performers" appropriate, it doesn't seem like that would be the case most of the time.

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/5879

Performers was a discussion point today at mid-sprint. It sounded like we were going to move forward without changing the label in this PR. In the following work, we were going to look at going with a visually hidden label and person teaser displays to be more flexible for the three sites that have this feature. As the use-cases pile up we were going to look into ways to label individuals differently but still map up to performer for schema.org whether that is refactored as another "contact information" type collection. "Performers" that don't have a first name/last name (e.g. musical groups) was brought up again as something we would want to account for in the future if more time was invested into this.

joewhitsitt:
For the `@url` property for subEvents, `Url::fromRoute('entity.node.canonical', ['node' => $nid], ['absolute' => TRUE])->toString();` I think will work

joewhitsitt:
Pulling down to investigate why performer is not showing up in the schema

Update: for it to show, the entity being referenced also needs schema mapping

joewhitsitt:
Tested sync for admissions and oniowa. Moved form access stuff in admissions to configuration (disabled fields on form and default render). Maintain event categories on events list block and titles are still linked on admissions according to the split

docroot/modules/custom/sitenow_events/sitenow_events.module
```diff
@@ -648,3 +660,85 @@ function _sitenow_events_build_options_list(array $tree, $r = 0, $p = NULL, arra
...
+        else {
+          $data['location'] = $virtual;
+        }
+      }
+    }
+  }
+}
+
+/**
+ * Set dynamic allowed values for the publish_options field.
```
pyrello: This comment seems like its incorrect.

joewhitsitt:
@mark-bennett-uiowa looking to you to see if this PR is ready to be reviewed

Created separate issues for the tag bug https://github.com/uiowa/uiowa/issues/5909 and route alter https://github.com/uiowa/uiowa/issues/5910

docroot/modules/custom/sitenow_events/sitenow_events.module
joewhitsitt: This whole function was supposed to be ripped out..

joewhitsitt: https://github.com/uiowa/uiowa/pull/5869/commits/26518d5b9c155b1336f9e425cf794a252959610c

removed

makurj:
Did run into config issues for admissions and oniowa sync. 
image

makurj:
"Event Categories" taxonomy vocabulary still exists on default. The other steps went through fine for default.

image

cory-skeers:
When linking a subevent directly to the source, the card that displays on the parent event

- links to the source on the card itself
- links to the subevent node on the card title

makurj:
ld+json is displaying superEvent on child event and subEvent on parent event fine. Parent and child events passed validation.

makurj:
Despite the config error, admissions allowed me to assign multiple event categories.

joewhitsitt:
@makurj I can't replicate the event categories taxonomy being available on default. config/features/event does not include the taxonomy anymore

mark-bennett-uiowa:
> @makurj I can't replicate the event categories taxonomy being available on default. config/features/event does not include the taxonomy anymore

I also cant see this either in my local version.

makurj:
Interesting. I wonder what's different with my local...

joewhitsitt:
Added a new test instruction related to the source link functionality:

>Linking directly to source for events still works as expect (title and card body linked to the same URL on events list block and sub events view, admissions /events and that internal/external links work)

joewhitsitt:
I am going to put this up for review.

joewhitsitt:
Sorry, I realize there is more to do here. @pyrello's comments have not been addressed and I still see some other remnants from the PR's pivot

makurj:
Started over. "Event categories" taxonomy is no longer available.

docroot/modules/custom/sitenow_events/sitenow_events.module
mark-bennett-uiowa: this is also found in /docroot/sites/now.uiowa.edu/modules/now_core/now_core.module so if we are updating this as well, we might want to update this in the Iowa Now module.

docroot/modules/custom/sitenow_events/sitenow_events.module
mark-bennett-uiowa: this is also found in /docroot/sites/now.uiowa.edu/modules/now_core/now_core.module so if we are updating this as well, we might want to update this in the Iowa Now module.

docroot/modules/custom/sitenow_events/sitenow_events.module
mark-bennett-uiowa: This has been altered.

docroot/modules/custom/sitenow_events/sitenow_events.module
mark-bennett-uiowa: This has been altered.

makurj (approved):
Functionality works as expected based on testing instructions.

docroot/modules/custom/sitenow_events/sitenow_events.module
joewhitsitt: @pyrello can we resolve this?

docroot/modules/custom/sitenow_events/sitenow_events.module
joewhitsitt: @pyrello can we resolve this?

docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
joewhitsitt: @bspeare can this be resolved?

docroot/modules/custom/sitenow_events/sitenow_events.module
pyrello: What I was suggesting beyond making the original comment more clear was that the entire section of code that is being added should have a comment at the beginning that explains why that whole section exists. Right now, you have to step through each piece to understand what is happening and it is not totally clear why.

That being said, I don't want to hold this up any longer.

cory-skeers:
Wrapped up my testing here, and I think everything is good to go!

Snuck in a quick comment for clarity at a spot I was initially confused at when reviewing the code.

docroot/modules/custom/sitenow_events/sitenow_events.module
joewhitsitt: @pyrello I added a more detailed "why" comment to the beginning of this code. I say further comment refinement can happen in the ISWF issue once this PR has been merged.

joewhitsitt:
Thanks everyone for the reviews. Merging once successful

PR #5880, opened 2022-10-20
update search_api module and update config

joewhitsitt:
There were updates to the view to change the query and caching to search_api compatible options. The team believes that this would be best to test in PROD.


joewhitsitt:
It was discussed in stand-up to switch to the search_api tag based caching thinking it would be the comparable option to start testing in PROD with. I will make that change here shortly

bspeare (approved):

Issue #5821, opened 2022-10-07
Migrate Iowa Now people


briand44:
# Story
As an Iowa Now website owner, I would like people migrated from our current site to our new site so that this content doesn't have to be manually created again.

# Additional Info

Fields:
First Name
Last Name
Job Title->Title/Position
Email
Phone
Photo
Areas of Expertise->Research Areas
About->Bio
Type->Tags (Expert, OSC Staff)
OSC Contact->Person reference

# Questions

- Does Iowa Now plan on using person type? - No, we won't map this in the migration and it won't be used to populate experts list or OSC staff list.

- Path - currently at /person/title. Do we keep this the same or move to people/title? Redirects? - Whatever is easiest for us. Seems like that would be to keep the existing path at /person/title?

- How do we handle person reference during a person migration?

# Proposed Solution

- [x] Extend base node migration for people?
- [x] Keep existing url pattern /person/title
- [x] Verify no redirects are needed
- [x] Run in dev so we can get feedback
- [x] Run in prod

joewhitsitt:
>OSC Contact->Person reference

Isn't this dirty data, aka contact name is a single string? Should this initially migrate to the contact type and manually transition certain contacts to person nodes later on?

joewhitsitt:
@briand44 the migration was performed in DEV: https://now.dev.drupal.uiowa.edu/person , https://now.dev.drupal.uiowa.edu/experts-list

Do you want to contact the Iowa Now team to review?

briand44:
Sure I can contact them. The link you provided, https://now.dev.drupal.uiowa.edu/person, is not going to be used correct? That will likely be disabled and they will just have https://now.dev.drupal.uiowa.edu/experts-list and the staff directory?

joewhitsitt:
>The link you provided, https://now.dev.drupal.uiowa.edu/person, is not going to be used correct?

It can be disabled at anytime. Temporarily serves as a content overview of the migration result

joewhitsitt:
Looked at this for two minutes and I am not sure why only some of the images migrated. The email appears to have just been a bad mapping. `/0/email` perhaps might need to be added to the mapping?

cory-skeers:
I think the image issues are due to spaces in filenames (eg "simon balto.jpeg"). The files are listed in the database with the space, and that's what is used to construct the path that gets passed to `file_get_contents` in the ProcessMediaTrait.

cory-skeers:
[...]

PR #5882, opened 2022-10-21
Iowa Now People Migration

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/5821

# Notes

There are ~4 temp person nodes that are created to deal with the OSC contact referencing. Most likely these are referenced entities that no longer exist in the source db. This causes some pretty scary looking errors on the command line, but there should only be a few and it should finish.

# Todo

- ~~Output redirects created~~ Removed in favor of the "sit and wait 404s" method
- Run in DEV for review
- ~~Look into warning: `Undefined array key "date_limiter" BaseNodeSource.php:124`~~ - _@cory-skeers fixed it!_
- ~~Re-wire staff/experts tids~~
- ~~Investigate breaking up job titles by delimiter `;` into multi value position field.~~  - _@cory-skeers fixed it!_

# How to test

- Sync now.uiowa.edu locally in dev desktop from uiowa703 (db name is now_uiowa_edu)
- git checkout now_migrate_person && git pull
- ddev blt ds --site=now.uiowa.edu
- ddev drush @now.local en now_migrate -y && ddev drush @now.local uli /admin/config/sitenow/sitenow-migrate
- Complete the form following the help text
- ddev drush @now.local ms
- ddev drush @now.local mim --all


joewhitsitt:
Trying to separate positions based on `;` but the following is failing for all:

```yml
  field_person_position:
    plugin: explode
    source: field_person_job_title
    delimiter: ;
```

cory-skeers:
> Trying to separate positions based on `;` but the following is failing for all:
> 
> ```yaml
>   field_person_position:
>     plugin: explode
>     source: field_person_job_title
>     delimiter: ;
> ```

Got this part working. Running it, I did end up with 63 failures, though...not sure if that was there before or if I broke something.

joewhitsitt:
>I did end up with 63 failures, though.

Before it was completely successful 177 or whatever. I meant to change the delimiter to semicolon + space. With 63 failures it seems like it might be failing on the ones that don't have a semicolon or entities without the field filled out?

cory-skeers:
Original site has `cristóbal-mckinney` as an alias, but it's `cristobal-mckinney` in the auto-generated alias post-migration. Doing a quick scan through the database, I _think_ this is the only original alias that had an accent and would be different for this reason, so it may be worth skipping the alias migration. Not 100% sure it's the only one, though--depends how good my eyes did at spotting accents.

joewhitsitt:
that is the one i spotted too

cory-skeers:
> > I did end up with 63 failures, though.
> 
> Before it was completely successful 177 or whatever. I meant to change the delimiter to semicolon + space. With 63 failures it seems like it might be failing on the ones that don't have a semicolon or entities without the field filled out?

Hrm...It doesn't look like there are any empty values for the field, and there are 49 entries w/semicolon and 128 w/out. I'll do some more pokin'.

cory-skeers:
Still not sure what was going on with the errors. I tried doing the same thing in prepareRow and it gets the job done without the errors.

joewhitsitt:
Thanks for helping with this @cory-skeers. Unless you notice anything else, I am thinking we should remove the redirects bit and run this is DEV for customer review.

Perhaps we should create a separate task to make the redirects comparison. Until redirect expiration is a thing, anything to keep that table small would be helpful.

cory-skeers:
> Thanks for helping with this @cory-skeers. Unless you notice anything else, I am thinking we should remove the redirects bit and run this is DEV for customer review.
> 
> Perhaps we should create a separate task to make the redirects comparison. Until redirect expiration is a thing, anything to keep that table small would be helpful.

That sounds good to me. Not noticing anything else outside the couple error-ed people, but it's a small enough number to be able to manually clean up. Similar with redirects--I think it's better to just remove it, and if we need to re-look into how that process works for future migrations that have more of a sure need for migration, that'd be a better time.

pyrello (approved):
Reviewed code updates and approving based on successful run in dev and customer feedback.

joewhitsitt:
Putting this up for technical review so it can merge into main and clear the way from further Now migrations. More customer feedback can be collected and implemented before the migration is run in PROD.

PR #5892, opened 2022-10-26
In-progress events

### LINKED ISSUE #5878: in progress events
We want events to be able to pull in events that are currently in progress (ie start date prior to today, end date after today). Currently there is no way to pull these in without having both the start date and end date encompass the event.

# Proposed Solution
Add a new display_id to events api that is a copy of Events from the Events API. Remove the Start Date constraint. Change End Date constraint to events with an end date after the provided date.
Update sitenow_events.module parameter handling in the case that it is upcoming (ie not date_range or past): Remove the start date and set the end date to Today instead of 31-12-2037
Update help text on the block
Let Chris C. know this is happening.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5878. 

**Changes made to content.uiowa.edu**
- Added two new views:
  -  In progress: https://content.dev.drupal.uiowa.edu/admin/structure/views/view/event_instances_api/edit/events_in_progress
  - Past events view for in progress: https://content.dev.drupal.uiowa.edu/admin/structure/views/view/event_instances_api/edit/past_upcoming
    - On both new views, we removed the start date as an exposed filter 
  - Changed end date to "greater than or equal to" in this new view
  

# How to test

- Review code changes
- `ddev drush @sandbox.local cset sitenow_events.settings base_endpoint 'https://content.dev.drupal.uiowa.edu/api/v1/'`
- Go to https://sandbox.uiowa.ddev.site/digital-signage-events and test events from content.dev.drupal.uiowa.edu
   - Test Upcoming Ascending and Descending
   - Test Past Ascending and Descending
   - Test Date Range Ascending and Descending


bspeare:
The current problem that I am seeing with this is that the code changes have allowed us to display in-progress events with a past start date and a future end date, however the block no longer displays any upcoming events that have a single date.   I've added testing instructions above and a link to the adjusted event api view on content.dev.drupal.uiowa.edu.

bspeare:
Actually, I think this is working.  I was momentarily confused by comparing the events on content.uiowa.edu and content.dev.drupal.uiowa.edu.   I'm going to do some more testing to figure out what the remaining tasks are for this issue.

bspeare:
@cory-skeers can you test this out today based on the testing instructions above and let me know if this is working as expected?

cory-skeers:
> @cory-skeers can you test this out today based on the testing instructions above and let me know if this is working as expected?

Can do --thanks for tackling this @bspeare!

cory-skeers:
> @cory-skeers can you test this out today based on the testing instructions above and let me know if this is working as expected?

From my testing, I think this is all working as expected!

_If_ we want to give the option to exclude in-progress events, and only include upcoming as those events that have not started yet, we do have the option of still using the `events` endpoint and providing both start and end date constraints. (I'd think like adding an "exclude in-progress" checkbox that appears on selecting upcoming, and adjusts the query appropriately).

I'm inclined to say we should only do that if it gets specifically requested, and stick to this as the intended function of upcoming, while providing past and date-range as separate modes. And I think that's what we'd settled on in grooming and previous discussions, but wanted to include the note that now that this has been worked out, adding that option should be a relatively small lift since we're keeping the old display as well.

e-marie-w:
> I was momentarily confused by comparing the events on content.uiowa.edu and content.dev.drupal.uiowa.edu.

Sooo I also confused myself with this. Everything was working then I changed departments and events were missing. 

Should I assume dev and prod have different events list and that if I'm seeing at least some past, future, or today events in their correct timeframes it's working correctly?

bspeare:
@e-marie-w yes, they have different events on each and are not synced.   The thing to check is whether you can get the timeframe working.  I adjusted my local page to display all university events so there are enough to test against.

e-marie-w (approved):
Timeframes are working correctly, with upcoming now pulling in-progress events like we planned. 

I don't understand the default/upcoming code, with how there's no start date and the end date is set to today, but it was laid out in the proposed solution and works successfully so I feel comfortable approving. 

But if someone has a minute sometime to explain it that'd be awesome.

e-marie-w:
Pulling down again to test latest changes

e-marie-w:
Timeframes still work great and as expected.

Right now I'm trying to check if a sub-department also shows parent department events, after this was brought up in the mid-sprint demo.

In trying to test this I discovered a couple of error messages that display when an events block has no events to display:
```
Warning: Undefined array key "page" in sitenow_events_preprocess_block() (line 154 of modules/custom/sitenow_events/sitenow_events.module).
Warning: Trying to access array offset on value of type null in sitenow_events_preprocess_block() (line 154 of modules/custom/sitenow_events/sitenow_events.module).
```
These messages appear per empty events block, if that's helpful. So, if I have three empty events blocks I see the errors three times. I think I saw this error message once yesterday but forgot to come back to it.

bspeare:
@e-marie-w I noticed that error message as well and I was trying to recreate it.   Can you outline the steps you took to get the error message?

e-marie-w:
> @e-marie-w I noticed that error message as well and I was trying to recreate it. Can you outline the steps you took to get the error message?

Oh so this isn't happening on newly created events blocks, I see now. It's only when I change an existing block to a department with no events that it happens. (Academic Advising Center has no upcoming events and is at the top of the list)

I also noticed that if I refresh the view the errors will vanish (until I resave the layout). They always persist in layout.

e-marie-w:
Actually forget everything I said in the last comment, these errors are occurring because I have the pager checked and there's no events to display. It can happen on new and existing blocks. As soon as I uncheck pager and save the errors disappear.

bspeare:
@e-marie-w were you able to test out that sub-department also shows parent department events?

e-marie-w:
No, I got distracted with the pager thing. I'll try testing it now.

e-marie-w:
So, it looks like sub-department only shows sub-department events, not parent. Here's a local example: 
Screen Shot 2022-11-01 at 4 00 57 PM
But I created the same page on sandbox prod and saw the same behavior, and now that I think about it this is the intended behavior. I remember working on grad department sites and we'd set their event feed to grad college instead of their specific department for this very reason.

e-marie-w:
> these errors are occurring because I have the pager checked and there's no events to display. It can happen on new and existing blocks. As soon as I uncheck pager and save the errors disappear.

Interesting development: this is not happening on sandbox prod for me, just my local. If we want to get this out sooner rather than later, though, I'm fine with creating a follow-up to fix this error.

cory-skeers:
> Interesting development: this is not happening on sandbox prod for me, just my local. If we want to get this out sooner rather than later, though, I'm fine with creating a follow-up to fix this error.

I was able to recreate it locally too, and I think it's small enough we could slip a fix in.

`sitenow_events.module` Line 142, adding a quick `isset` condition I think does the trick.

```
      // If pager, create it and re-query data based on current page.
      $show_pager = (bool) $block->get('field_collection_pager')->getString();
      if ($show_pager && isset($data['page'])) {
        $count = $data['page']['total_results'];
        static $element = 0;
        $element++;
```

bspeare:
@cory-skeers do you want to add that in?

bspeare:
@e-marie-w can you check to see if Cory's fix removes the errors?

e-marie-w:
Yep! So now I get `Error messageWarning: Undefined variable $element in sitenow_events_preprocess_block() (line 210 of modules/custom/sitenow_events/sitenow_events.module)`.

But I figured it out! Line 208 also has `if ($show_pager)` code, and now that we check `isset()` for the first instance, the $element isn't necessarily defined before we get to the second instance.

We can add another `isset()` - no more errors, although PhpStorm underlines it and warns it still might not be defined. We could also just define $element before the first if ` if $show_pager`, probably after Line 149 where $data is initialized. 

I'm leaning towards the second option but wanted to check before I commit any code.

e-marie-w:
> We can add another `isset()` - no more errors, although PhpStorm underlines it and warns it still might not be defined. We could also just define $element before the first if ` if $show_pager`, probably after Line 149 where $data is initialized.

Wait on second thought maybe we want to do both? Even though it doesn't error for me we still can't display the pager without the isset condition also being true.

cory-skeers:
> > We can add another `isset()` - no more errors, although PhpStorm underlines it and warns it still might not be defined. We could also just define $element before the first if ` if $show_pager`, probably after Line 149 where $data is initialized.
> 
> Wait on second thought maybe we want to do both? Even though it doesn't error for me we still can't display the pager without the isset condition also being true.

Yeah, I think if we aren't able to display the pager anyway, we should skip over adding it to the variables.

I think we could add an `isset($element)` there and php would like it as well as it not erroring/skipping over that variables addition, but in terms of code readability that feels icky to me, since what we're really dependent on is there not being paged results, not that the arbitrary `$element` wasn't defined.

cory-skeers:
Pulled and tested with the additional check and move of `$element` initialization, and am seeing no errors when using pager with no results (and still correct pager behavior when there are results).

bspeare:
@cory-skeers great! Let's merge this in.

Issue #5702, opened 2022-09-15
AD role mappings connect.studentlife.uiowa.edu


briand44:
As an owner of the intranet connect.studentlife.uiowa.edu, I would like to grant access to specific AD groups so that we do not have to manually control access.

Examples from other websites:
https://github.com/uiowa/uiowa/blob/main/config/sites/sitenow.uiowa.edu/config_split.patch.uiowa_auth.settings.yml
https://github.com/uiowa/uiowa/blob/main/config/sites/brand.uiowa.edu/config_split.patch.uiowa_auth.settings.yml

briand44:
blocked until we get the list of AD groups

briand44:
The list of AD groups have been provided. These will need to map to the "Viewer" role. There are a lot.

https://iowa.sharepoint.com/sites/CSI-Drupal2/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FStudent%20Life%2FStudent%20Life%20Connect%2Fstudent%2Dlife%2Dconnect%2Dad%2Dgroups%2Etxt&viewid=bf54b9d9%2D0975%2D4fbc%2D9f1b%2D6377bdc2307c&parent=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FStudent%20Life%2FStudent%20Life%20Connect

briand44:
Grooming conversation: Move forward with this like SiteNow and Brand. Test any potential performance issues in dev. Create a follow up to determine if this is how we want to handle these going forward or if we want to be ignoring this config and only allowing Administrators to change this directly in the UI.

pyrello:
Follow-up issue: #5862

quamsta:
> The list of AD groups have been provided. These will need to map to the "Viewer" role. There are a lot.
> 
> https://iowa.sharepoint.com/sites/CSI-Drupal2/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FStudent%20Life%2FStudent%20Life%20Connect%2Fstudent%2Dlife%2Dconnect%2Dad%2Dgroups%2Etxt&viewid=bf54b9d9%2D0975%2D4fbc%2D9f1b%2D6377bdc2307c&parent=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FStudent%20Life%2FStudent%20Life%20Connect

[...]

Issue #4640, opened 2022-01-11
Intranet - Additional considerations


joewhitsitt:
Not sure if this will become an epic, but we wanted to capture some other aspects of running an Intranet split. Jotting those down here:

- ~~Update content types to not be included in the sitemap and/or look at disabling (but not uninstalling the sitemap functionality). Ultimately, Sitemap.xml should NOT disclose pages that are not accessible without authentication.~~ 
- ~Disable or redirect search result page~
- ~~Remove xml feeds (sitenow articles, taxonomy terms)?~~
- ~~Additional disallow entries to the robots.txt~~
- ~~Remove SEO Settings per content type and/or set meta noindex, nofollow tag set in the meta~~
^^ https://github.com/uiowa/uiowa/issues/4640#issuecomment-1050073341
- ~~See about a certain range of IPs to allow?~~ - https://github.com/uiowa/uiowa/issues/4640#issuecomment-1010431168
- ~~Consider removing header and footer as links and footer content will be public~~ https://github.com/uiowa/uiowa/issues/4640#issuecomment-1050074591
- [x] Custom handling/messages (redirect elsewhere?, Custom Intranet messaging other than Access Denied)
- ~~Update or split off default content to not use public files for this split~~
- ~~Set destination parameter for post-login redirect~~

Pulled bits from comment https://github.com/uiowa/uiowa/issues/4241#issuecomment-952239131, the "later on" todo of https://github.com/uiowa/uiowa/pull/4626 and discussion from our Intranet files meeting

Updated todo:

- [x] Ensure editor/publisher/webmaster do not have access to siteimprove.
- [x] Figure out why the footer login link test isn't failing - it should be. (@richardbporter)
- [x] Set the footer login link config on install of module. (@richardbporter)
- [x] Set config override priority of robotstxt to take precedence over other overrides.

richardbporter:
>See about a certain range of IPs to allow?

Don't think this possible without encountering cache issues.

richardbporter:
I wonder if the "Access Denied" page should actually return a 403? Currently returns 200.

briand44:
Let's probe the items listed in the IS and determine what makes sense to address in additional slices.

richardbporter:
The current login link in the footer sets the destination parameter to redirect back to the access denied page which is really odd UX.

joewhitsitt:
>The current login link in the footer sets the destination parameter to redirect back to the access denied page

[...]

PR #5893, opened 2022-10-27
Updated intranet login button query parameter to send user to the homepage after logging in

quamsta:
Minor tweak to intranet to allow the big login button to drop user off at the homepage:

https://github.com/uiowa/uiowa/issues/5702#issuecomment-1293819260

# How to test

I'm... not actually sure and wasn't able to test this locally. The SAML login button just doesn't work at all locally with or without the query parameter.

The button's generated href does look correct though:

Screen Shot 2022-10-27 at 12 50 31 PM

May need to be deployed to `dev` to fully test?


joewhitsitt:
Yeah, we will want to test in DEV. Travis ain't happy with the old array style and the middle optional portion might need to be included:

https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L218

quamsta:
> Yeah, we will want to test in DEV. Travis ain't happy 

One day I'll learn how to make Travis happy. Ran `phpcbf` and fixed this. 

> the middle optional portion might need to be included:
> https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.theme#L218

I believe I'm already using the second optional argument in `Url::fromRoute` which is `$route_parameters` to populate the query parameters. It looks like the third argument `$options` can also be used to populate the query params. Not sure which one's preferred.

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Url.php/function/Url%3A%3AfromRoute/9.1.x

pyrello:
I'm wondering about making this configurable. Also, if someone is sent to a link that they need to log in to view, then we probably want to handle redirecting to that resource, so they are not just unceremoniously dumped on the user page/homepage.

quamsta:
> I'm wondering about making this configurable. 

Not sure what situation this would be useful in. Can you describe a scenario you're thinking about here?

If you're being sent to `https://intraneturl.uiowa.edu/` do we need to allow admins to configure the site so the user is dropped off at `https://intranet.uiowa.edu/landing-page` upon logging in, for example? If they're expecting to land on the homepage, does a "new homepage" need to be configurable or can we just call the homepage the homepage? Maybe I'm overthinking.

> Also, if someone is sent to a link that they need to log in to view, then we probably want to handle redirecting to that resource, so they are not just unceremoniously dumped on the user page/homepage.

Yes! This would be ideal and after discussing with some stakeholders, this would actually be critical for our leadership sending intranet links to people and having them able to open them properly. But I wasn't sure how possible dropping folks to their original location was in this scenario. I figured the homepage would be better than `/user/`.

I will investigate redirecting to a resource, and may reach out.

joewhitsitt:
I believe the conversation has fragmented. Homepage was the "anything better than the user page" solution. Scraping the original entry before restrict ip redirects to a general 403, adding that to the link seemed like a bigger task. If the idea that route checking isn't done for this and folks could still get dumped to a 404 after login, then the anonymous guessing isn't an issue here.

https://github.com/uiowa/uiowa/issues/4640#issuecomment-1033036113

pyrello:
> > I'm wondering about making this configurable.

> Not sure what situation this would be useful in. Can you describe a scenario you're thinking about here?

I don't currently operate an intranet, so any examples I come up with will be contrived. Someone could potentially want people to end up at `/admin/content` when logged in. I guess my point is that just because having the homepage be the default redirect location after login is your use case doesn't mean that every intranet would want it set up that way. I agree that there is little value in having people end up at their own profile page because we don't really have any valuable functionality enabled there.

I'm not necessarily opposed to this change, but also don't necessarily like doing this in the absence of a broader conversation about what our goals are.

quamsta:
> > > I'm wondering about making this configurable.
> 
> > Not sure what situation this would be useful in. Can you describe a scenario you're thinking about here?
> 
> I don't currently operate an intranet, so any examples I come up with will be contrived. Someone could potentially want people to end up at `/admin/content` when logged in. I guess my point is that just because having the homepage be the default redirect location after login is your use case doesn't mean that every intranet would want it set up that way. I agree that there is little value in having people end up at their own profile page because we don't really have any valuable functionality enabled there.
> 
> I'm not necessarily opposed to this change, but also don't necessarily like doing this in the absence of a broader conversation about what our goals are.

My main goal right now as an operator of an intranet, and trying to launch one in Drupal, is preventing the user from logging in and landing on their profile page, which contains nothing except their HawkID username, which, in my opinion, is unacceptable.

Emails from our VP, newsletters, and other direct messaging methods, provide links to the intranet directly since that's the only way to really get into the intranet due to its non-public nature. Dropping people off from those communications at a page that looks like this:

Screen Shot 2022-10-31 at 10 33 10 AM

...Is not really acceptable and is a bigger problem than just having "little value." IMHO it provides no value and impedes the usability of the intranet itself.

We're currently trying to plan for manually appending query strings to links sent out via these communications, (e.g. we'll be manually sending out links that look like `https://connect.studentlife.uiowa.edu/saml/login?destination=/` or `https://connect.studentlife.uiowa.edu/saml/login?destination=/news/`) but it'd really nice to not have to do this.

It can be safely assumed that a website, even an intranet, will have a homepage, and if we can't easily direct people to their original destination, then the homepage is really the next acceptable option in my opinion. I don't really see the need to add complexity to the codebase by having something configurable here, especially if the ideal scenario and eventually the end goal is directing people to their original destination.

As @joewhitsitt said, this is the 

> Homepage was the "anything better than the user page" solution.

I'm marking this as ready for review for the time being, since it's one of the few things blocking us from launching our intranet aside from re-grabbing some content from our previous, live intranet site. Would love to discuss further, but would also like to get our old Intranet site decomissioned soon.

briand44:
It seems like sending folks to the homepage is an improvement from sending them to their user page. I would be ok with rolling this out as an initial improvement to get the Student Life Intranet launched and we can create a follow up to address routing folks to the appropriate page from a link after log in.

joewhitsitt:
Going to push this out to DEV to test

joewhitsitt (approved):
This worked https://studentlifeconnect.dev.drupal.uiowa.edu/

quamsta:
Thanks all. Appreciate this. A lot.

PR #5899, opened 2022-11-01
IISC: Pre-launch updates

pyrello:
- Changes the teaser display of partnerships to show the image as a rendered entity instead of a label.
- Adds permissions for Colleges and Departments taxonomy similar to existing vocabularies.
- Switches filters to show all the time on projects view.
- Removes some partnership views that the client no longer wants.

# How to test

Code review is sufficient.


cory-skeers (approved):
Reviewed the code changes, and all looks good 👍

PR #5921, opened 2022-11-04
Limiting tag and related content from displaying on all custom content types

### LINKED ISSUE #5909: Related by tag functionality shows on non-default content types
# Problem
The related by tags functionality is showing on the non-default event content type (possibly others?) and there are no controls to turn it off.

Screen Shot 2022-11-02 at 12 47 13 PM

# Steps to reproduce
With the event feature split enabled, add a tag to an event. See the tag show on the rendered entity.

# Proposed solution
- [ ] Check the logic that is used to determine the visibility of this functionality so that content types without control setup don't display these blocks.

e-marie-w:
Resolves #5909 

# How to test
With the event feature split enabled, add a tag to an event. See that the tag does not show on the rendered entity.
Add a tag to one of the node types expected to display (eg articles). See that the tag display still properly follows the site-wide settings


joewhitsitt:
Is this happening for views_block__related_content_block_related_headings_lists as well? If not, what is the difference in the logic?

cory-skeers:
> Is this happening for views_block__related_content_block_related_headings_lists as well? If not, what is the difference in the logic?

Hrm, to be thorough we should probably unset both `views_block__related_content_block_related_headings_lists`, and `views_block__tags_list_tags_list`. What about `views_block__related_manually_referenced_content_block_grid`?

At this point in the code, there's no difference in the logic for the two, and if we don't unset then `views_block__related_content_block_related_headings_lists` ends up being left in the variables. At least for events, further down the line it doesn't end up getting rendered.

cory-skeers:
> Hrm, to be thorough we should probably unset both `views_block__related_content_block_related_headings_lists`, and `views_block__tags_list_tags_list`. What about `views_block__related_manually_referenced_content_block_grid`?

It feels to me like we shouldn't default to unsetting the manually referenced, since that 1) only gets unset in other instances when there's an alternate display method of that content and 2) is less likely to have weird interactions with other functionality and is more likely to just be empty and not display on content pieces that we don't want it (and saves a step if there is a case we do want it).

makurj:
Test this on default?

cory-skeers:
> Test this on default?

Yup! Default works--the only extra step is to turn on the Event split then, or you can test on a site that already has the split turned on (Admissions...not sure of any others).

makurj:
Admissions: With "Do not display tags" on for the site, the tags are displaying on the event but not on the article.

image

image

With "display tags buttons", they do become visible on the article:

image

And with more info on the event:

image

makurj:
Admissions: tags not showing up on either event or article when site setting is set to "do not display tags". For "display tag buttons", tags still not visible on rendered event, but they are showing up on the article. So the article content type is working as expected and following the site-wide default.

Are tags not supposed to show at all on the event irrespective of the site setting? Can’t tell from the test instructions.

makurj (approved):
Tested. Working as expected.

PR #5922, opened 2022-11-04
Search Block - External URLs

### LINKED ISSUE #5875: Law Library InfoHawk+ Search
# Story

As an owner of library.law.uiowa.edu, I would like to have a search box on my website where users can search InfoHawk+.

# Additional Info
Existing code on 703 in /docroot/sites/library.law.uiowa.edu/modules/custom/custom_widgets/plugins/content_types/custom_widgets_search_block.inc.

# Proposed solution

joewhitsitt:
Resolves #5875

# How to test #
- pull branch, sync site
- Add search block with following:
   - endpoint: https://search.lib.uiowa.edu/primo-explore/search
   - query parameter: query
   - query prepend: any,contains,
   - additional parameters: 
       &institution=01IOWA&vid=01IOWA&mode=basic&tab=default_tab&search_scope=default_scope
- Use the block. Search should take you to results thru InfoHawk. 
- Add extra words, special characters, etc to the additional parameters text area. Hitting search should strip the excess out and still return results. 
- Check default search block field values still work as expected.

<details>
<summary>Original description</summary>
<p>For my professional dev time today, I wanted to look into why the block didn't support external URLs.
Relates to upcoming work https://github.com/uiowa/uiowa/issues/5875</p>
</details>


cory-skeers:
Pulling down to test

e-marie-w:
Confirmed collapsible form element code works. In `docroot/modules/custom/uiowa_core/src/Plugin/Block/SearchBlock.php`, I thought I might need to specify ['advanced'] in the build() function. This breaks it, however, so I am now confident code is correct as-is. Marking ready for review again.

cory-skeers:
There's an error if there are repeated query parameters (e.g. `institution` appearing twice in the additional parameters) that happens in the `parseQueryString` method. It fails pretty gracefully, and you actually get directed with a successful search, but going back to the page and refreshing brings it up.

```
Warning: foreach() argument must be of type array|object, string given in Drupal\Component\Utility\NestedArray::mergeDeepArray() (line 327 of core/lib/Drupal/Component/Utility/NestedArray.php).
Drupal\Component\Utility\NestedArray::mergeDeepArray(Array, 1) (Line: 211)
Drupal\Component\Utility\UrlHelper::parseQueryString('&amp;institution=01IOWA&amp;vid=01IOWA&amp;mode=basic&amp;tab=default_tab&amp;search_scope=default_scope&amp;institution=01IOWA') (Line: 68)
Drupal\uiowa_core\Form\SearchBlock->submitForm(Array, Object)
```

It's not a wsod, and the "workaround" is to enter good, unique parameters, so I'm not sure it's a big concern. We could potentially do our own check for duplicates, but that feels like a duplication of efforts of what the `parseQueryString` is supposed to do.

Otherwise, I've tested this with InfoHawk+ and some other basic searches (imdb, google, duckduckgo) with various parameters, and this is working really well! The `prepend` and `additional parameters` options make it really flexible.

As far as security, I think this generally isn't problematic. The main concern I could see would be if an editor added query parameters like uersernames/API keys and potentially leak info that way. It might be worth adding something to the help text w/regards to that, like "Do not include sensitive information such as api keys, usernames, or passwords."

cory-skeers:
> It's not a wsod, and the "workaround" is to enter good, unique parameters, so I'm not sure it's a big concern. We could potentially do our own check for duplicates, but that feels like a duplication of efforts of what the `parseQueryString` is supposed to do.

Dang, it's not a wsod but it can end up visible for non-logged in users.

image

e-marie-w:
Great finds @cory-skeers! I think it's worth checking for duplicates. I'll update the help text too.

joewhitsitt:
Wait, `parseQueryString` is returning back duplicate keys in the array? so we need a key version of array_unique()?

cory-skeers:
> Wait, `parseQueryString` is returning back duplicate keys in the array? so we need a key version of array_unique()?

I think `parseQueryString` is expecting that if there are multiple values for a parameter key, they were included in the same instance, and so get properly appended to the same key.

But if you include duplicate keys separately, the second time the key comes around, it'll try to do an array merge on the separate keys' strings.

It's this part of `UrlHelper.php` lines 210:215, when it gets a hit on the `if` from an already set key.

```
      if (isset($parsed[$name])) {
        $parsed[$name] = NestedArray::mergeDeepArray([$parsed[$name], reset($param)], TRUE);
      }
      else {
        $parsed[$name] = reset($param);
      }
```

cory-skeers:
The first thought I'd had for this would be to do a form validation check for duplicate parameters, and set an error with a message to the user to fix. That way we could still make full use of `parseQueryString` for the actual processing.

I think something like

```
preg_match_all('@&(.*?)=(.*?)@is', $s, $matches)
```

would work as the quick check to get all of the parameter keys (`$s` would be the query parameter string we'd search; matches will get assigned to `$matches` and `$matches[1]` would be an array of all the matched keys). Then check `$matches[1]` for duplicate keys, and if found, set a form error.

cory-skeers:
After it sitting for a moment, rethought the regex a little.

`preg_match_all('@&([^&]+)=[^&]+@is', $s, $matches)` would be a little more concise and likely a bit faster.

https://regex101.com/r/iD49Gk/1 to play around with it

joewhitsitt:
Tried to put this together into a validate. Having troubles with the setErrorByName nested in details...

@cory-skeers is this what you had in mind?

cory-skeers:
> @cory-skeers is this what you had in mind?

Yes! Exactly the kind of thing I was thinking.

cory-skeers (approved):
Reviewed and tested, and looking good! Thanks for the extra additions.

PR #5935, opened 2022-11-09
Events block: Include last_day if end_day not present

### LINKED ISSUE #5916: Multi-day All Day events only show the first date
# Problem
Events that are marked as All Day but span multiple days, when included in Sitenow Events block, are only showing their start date.

# Steps to reproduce

https://content.uiowa.edu/uicb-echizen can be used as an example. (for instance, set up an Events block filtered on Department: Graduate College and Type: Exhibit)

See that the card teaser only shows the first date, though the event spans 10/26/2022 (All day) to 11/5/2022 (All day)
image

# Proposed solution

Revisit `sitenow_events_format_instance` (https://github.com/uiowa/uiowa/blob/6aaf4731ae68ead21919b4b5523328da624744b2/docroot/modules/custom/sitenow_events/sitenow_events.module#L397-L463) which is the logic that prepares the event details for the card. In this instance, it's determining the event to be `$is_same_day && $is_all_day` when it should be `!$is_same_day && $is_all_day`. 

In preliminary digging, I think the start of the method
```
  if (isset($event_instance['end'])) {
    $end_date = strtotime($event_instance['end']);
  }
  else {
    $end_date = $start_date;
  }
```
isn't getting an end date, and need to also check `last_date` from the event data. Some more digging may need to be done in the code that calls this function.

cory-skeers:
Resolves #5916 

Essentially, if an event is marked as All Day and spanning across more than one date, it currently only displays the start date when pulled into a Sitenow Events block.

# How to test

Set up an Events block which features an All Day event over multiple days.
https://content.uiowa.edu/uicb-echizen can be used as an example. (for instance, set up an Events block filtered on Department: Graduate College and Type: Exhibit)

See that rather than only displaying one day,
image

all days in the range are displayed.
image

Note: 
In looking into this, the above Event is not set up in the recommended practice:

image
"If the event spans multiple days, it is recommended to use a date repeat rule instead of different start/end dates. For example, if the event runs all week, select the start date, omit the end date and then add a daily repeat rule that stops on the last day of that week."
But as it's a recommendation and not a requirement, I think this change works as a small fallback.


joewhitsitt:
Can this logic be simplified to just use last_day? I feel like it is always provided if for non all day events:

https://content.uiowa.edu/api/v1/node/106936.json (all-day), https://content.uiowa.edu/api/v1/node/106246.json (one-day specific time), https://content.uiowa.edu/api/v1/node/82456.json (weekend)

cory-skeers:
> Can this logic be simplified to just use last_day? I feel like it is always provided if for non all day events:
> 
> https://content.uiowa.edu/api/v1/node/106936.json (all-day), https://content.uiowa.edu/api/v1/node/106246.json (one-day specific time), https://content.uiowa.edu/api/v1/node/82456.json (weekend)

The last_day only includes the date and not a time, where the instance_start and instance_end include the time as well and are needed for non-all day events.

The other trickiness comes in for repeated events like https://content.uiowa.edu/api/v1/node/98261.json. Each instance_end corresponds to its specific instance, but last_day is the last-most date of any instance.

joewhitsitt:
Ah gotcha. Thanks for clarifying.

joewhitsitt:
pulling down to test

joewhitsitt:
@cory-skeers any tips on how to find some upcoming/in-progress to test against? should i pull the site down and do some view querying?

cory-skeers:
> @cory-skeers any tips on how to find some upcoming/in-progress to test against? should i pull the site down and do some view querying?

Yep, unfortunately that's my only tip 😬

cory-skeers:
I take that back: one other tip
https://studentwellness.uiowa.edu/ and https://mentalhealth.uiowa.edu/ both have events blocks on their home page with some good instances of repeats and different combos of types of durations. Not necessarily comprehensive for testing, but a decent starting point at least.

joewhitsitt (approved):
Ugh. That was not a fun testing experience. I ended up creating events in DEV and changing the base endpoint. Code looks good though. Works as expected.

cory-skeers:
Thank you for going through that, @joewhitsitt!

PR #5961, opened 2022-11-18
GCLID/FBCLID workaround

### LINKED ISSUE #5964: Allow gclid and fbclid to pass through to Webform fields
# Problem
Webform has pre-populate functionality that fills in matching fields with query string parameter values, but there are at least two ids, `gclid` and `fbclid` that don't get passed when Acquia's varnish is in front of the request.

https://support-acquia.force.com/s/article/360004119294-Acquia-s-default-Varnish-VCL-strips-Google-Analytics-and-Facebook-Ads-query-string-parameters

# Proposed solution
- [ ] Using https://github.com/uiowa/uiowa/pull/5961 as a reference
- [ ] Check all webforms for elements with gclid/fbclid machine names with type hidden and prepopulate checked. Attach JS if it exists.
- [ ] prepopulate info is available when the js is attached, however, we don't want the js blindly applying to each element without further checking (multiple forms). capturing info in drupalSettings or setting an attribute on the element to check against with JS would be neat.

# Additional

Working example as of 11/18: https://tippie.dev.drupal.uiowa.edu/form/gclid-test?gclid=123456789123456789aaaaa

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5964

This PR introduces a workaround for Acquia varnish not passing certain click identifiers to a full page load for Drupal Webform's prepopulate functionality to utilize. Instead this attaches JS that is on top of a varnish request that will fire and populate these fields.

# To Test

- In Chrome (not Brave), visit the form with query parameters attached to the URL (e.g. https://sandbox.dev.drupal.uiowa.edu/click-id-form-test?gclid=thisismygclidvalue&fbclid=thisismyfbclidvalue&other_hidden=asdf) and inspect the hidden elements to see that their values are populated.
- If successful, go back and try adjusting configuration for the forms where not all of these conditions match (different element machine names, prepopulate not checked).


joewhitsitt:
oops. assets is not tracked in git

makurj:
If "prepopulate" is unchecked, this error comes up:

image

If a default value is provided, it is being overwritten by the URL param:

image

makurj:
Testing more conditions so maybe hold off on fixing the issues above.

joewhitsitt:
good finds. moving back to draft for now.

makurj:
With 2 forms on the page: 

If `gclid` machine name is given to a non-hidden element, it pre-populates the URL param value for it in the field.

image

Second form with `gclid` and `fbclid` hidden and prepopulated, it works fine.

image

makurj:
I think this is ready to be worked on now.

makurj:
With 2 forms on a page, both with `gclid` and `fbclid` hidden and prepopulated, both are getting the URL param values.

image

image

joewhitsitt:
fixed the undefined error from https://github.com/uiowa/uiowa/pull/5961#issuecomment-1352103785

as for the default value issues. i misunderstood the out of the box webform prepopulate functionality. for webform, the prepopulate checkbox will overwrite the default and we are seeing that while testing locally because there is no varnish. I am going to remove the check on the default value in our workaround to act the same way and we should deploy this out on DEV to actually test so that it isn't confused with the webform php prepopulate stuff.

joewhitsitt:
still need to look into the non-hidden type issue

Update: prepopulate is an option for other elements besides type hidden. So removing the hidden check as well.

joewhitsitt:
I'd like to push out to DEV again for testing.

joewhitsitt:
I think this is ready for review again. I have updated the test instructions to use DEV.

To recap the changes:

- no errors related to undefined prepopulate.
- no longer limited to hidden type elements as prepopulate is available on other element types.
- default can be overwritten if prepopulate is checked. this matches the webform php functionality.

docroot/profiles/custom/sitenow/js/get_clickid.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+      )
+
+      // If there are params matching our target elements, loop through and
+      // populate the value.
+      if (query.length) {
+        let queryString = query.join(',');
+        $('.webform-submission-form', context).once('get_clickid').each(function (index) {
+          context.querySelectorAll(queryString).forEach(function (input) {
+              input.value = params.get(input.name);
```
pyrello:
```suggestion
            input.value = params.get(input.name);
```

makurj:
With all conditions matching:

image

Default value being overwritten by prepopulate:

image


Prepopulate unchecked for gclid, the value is no longer coming through, as expected:

image

No error on prepopulate being unchecked on hidden `gclid`

makurj:
`gclid` machine name given to a non-hidden text field with prepopulate checked is picking up the URL param value, as expected.

image

Without URL params, it's blank:

image

makurj:
Looks good to me. I can approve, unless there's more testing needed.

joewhitsitt:
Short of errors/wsod, I think the real testing will happen in PROD by our customer(s)

makurj (approved):

joewhitsitt:
Thanks!

PR #5967, opened 2022-11-22
Admissions - Config fix

joewhitsitt:
I caused a bad export in https://github.com/uiowa/uiowa/pull/5948 that messes with the lb keys/uuids. This switches things back and also adds dependencies for two hidden fields also added in that PR.

# To Test

- Sync admissions. No import errors.
- Merge to main and test against admissions.stage


e-marie-w (approved):
Synced admissions site successfully, no import errors.

Issue #2813, opened 2021-01-07
As a site owner, I can adjust the layout/placement of a featured image in the article content type


bspeare:
# Story

As a site owner, I can adjust the placement of a featured image in the article content type so that there is an alternative to the widescreen image format at the top of the page. 

# Proposed solution

Create a "Featured image display option" in the "Featured image" edit panel on the node edit screen:

<img width="531" alt="Screen Shot 2021-01-07 at 1 41 40 PM" src="https://user-images.githubusercontent.com/1036433/103937290-2cdc3680-50ee-11eb-92e1-a3c351fd1e33.png">

This select option would control the layout for the article.  
- The `Default` option could be the standard widescreen
- An additional `Sidebar` option could display the article in a two column layout

## First sidebar layout mockup:

- Displays featured image inline with the body text 
- Displays sidebar with date, author, share links

<img width="1878" alt="Screen Shot 2021-01-07 at 12 45 11 PM" src="https://user-images.githubusercontent.com/1036433/103937559-847aa200-50ee-11eb-874a-ecc574794f2d.png">

bspeare:
Possible solution related to https://github.com/uiowa/uiowa/issues/2784 and https://github.com/uiowa/uiowa/issues/2530.

bspeare:
Questions/observations from planning discussion:

**Default layout: Editorial Design**
- Narrow centered column width would make sense for widescreen featured image like https://stories.uiowa.edu/iowa-helps-turn-tide-in-pandemic but only if featured image is displayed. 
- Add banner headline styles for highlight text treatment
- Keep container left aligned?

**Sidebar layout: Press Release Design**
- How to handle different featured image sizes inline?
- Use WYSIWYG instead to upload an image after hiding the featured image?
- Two column layout format with meta information in the sidebar would be useful for the Event content type layout.
- Options for meta information placement and column width?

bspeare:
After further consideration, maybe we should just update the layout to add an option use a sidebar column for metadata to solve the line-length issue?

## Sidebar column option with featured image:

<img width="1738" alt="Screen Shot 2021-01-25 at 2 50 50 PM" src="https://user-images.githubusercontent.com/1036433/105765113-841b3d00-5f1d-11eb-9826-9a935e6df750.png">

## Sidebar column option with featured image hidden:

[...]

PR #5993, opened 2022-12-05
Make Article Captions Captiony Again

### LINKED ISSUE #3561: Make Article Featured Image Caption Render Again
With https://github.com/uiowa/uiowa/pull/3550 we are sweeping a regression under the rug until we can reimagine how a caption on an article should render for all devices or decide to deprecate and remove the caption functionality.

Before we decide to remove/deprecate the caption functionality, keep in mind there is a global caption field and that there have been requests to provide better attribution for media sources.

Known issue:  https://sitenow.uiowa.edu/known-issues#uiowa-issue-3561

# Proposed solution
- Make sure styles are scoped for only featured image caption for the article page. 
- Banner mobile stacked featured image will need to be tested/taken into account. 
- Try https://github.com/uiowa/uiowa/issues/3561#issuecomment-1331166227 first, and then https://github.com/uiowa/uiowa/issues/3561#issuecomment-1331106535
- Search for `core.entity_view_display.node.article.default.yml` on https://github.com/uiowa/uiowa/find/main

cory-skeers:
Resolves #3561 

# How to test

`git checkout article-captions && git pull && ddev blt frontend && ddev blt ds --site=default && ddev drush @default.local uli`
Create an article with a featured image and a featured image caption
Check things out!
Test various combos of featured image, featured image caption, and global caption (Check that the caption doesn't render when there is no featured image)
Test that global captions still apply to images added to a photo gallery
Test on some or all of the splits listed in https://github.com/uiowa/uiowa/pull/5993#issuecomment-1340115714


cory-skeers:
When checking that Page wasn't affected by these changes, found that global captions appear tucked behind the featured image (if the selected image has a global caption):
image
This looks to be the case on `main` and not due to the changes here, but as a followup maybe we can decide what to do with those.

cory-skeers:
Search for article displays on https://github.com/uiowa/uiowa/find/main came up with the following splits that'll need to be checked against
- [x] [config/default/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/default/core.entity_view_display.node.article.default.yml)
- [x]  [ config/features/periodical/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/features/periodical/core.entity_view_display.node.article.default.yml)
- [x]  [ config/sites/now.uiowa.edu/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/sites/now.uiowa.edu/core.entity_view_display.node.article.default.yml)
- [x]  [ config/sites/iisc.uiowa.edu/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/sites/iisc.uiowa.edu/core.entity_view_display.node.article.default.yml)
- [x]  [ config/sites/sppa.uiowa.edu/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/sites/sppa.uiowa.edu/core.entity_view_display.node.article.default.yml)
- [x]  [ config/sites/91stmeridian.iwp.uiowa.edu/core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/sites/91stmeridian.iwp.uiowa.edu/core.entity_view_display.node.article.default.yml)
- [x]  [ config/sites/hr.uiowa.edu/config_split.patch.core.entity_view_display.node.article.default.yml](https://github.com/uiowa/uiowa/blob/main/config/sites/hr.uiowa.edu/config_split.patch.core.entity_view_display.node.article.default.yml)

cory-skeers:
The little border and background for the 'i' works pretty well for dark areas of images:
image
(example found on SPPA site split)

cory-skeers:
Works well with the Now split. Checked it on some of the pages that had migrated in captions. The featured image is set to not display by default, in which case the caption does not render; setting it to display includes the caption as expected.
image

cory-skeers:
I think this is the desired functionality with the overlap with global caption, and is how it currently works:

- Regardless of if there is a global caption or not, if there is a featured image caption listed, that caption is displayed on the featured image
- If there is a global caption but no field image caption, no caption is listed on the featured image
- Global caption gets used in places like photo gallery, and featured image caption does not

The second bullet is the main one that I could see an argument for other functionality, like including the global caption if there is no featured image caption. It makes sense to me as the global caption should be one that is appropriate to the image regardless of its placement, but there is also no way to not include it (whereas if it was still desired, an editor could just use the global caption in the field image caption field as well).

joewhitsitt:
Pulling down to test. Fair warning though, I coming to this without hearing the decisions made around direction, so other reviewers are welcome.

bspeare:
Can we fade in the button instead of animate it?  Perhaps we could animate the actual tooltip instead? See https://codepen.io/neogomo/pen/BjqJzr for examples.

bspeare:
I just realized that we created the circle button in the recent card component work https://github.com/uiowa/uids/blob/4.x/src/components/button/button.scss#L149-L175.  I think we should look at using those styles here instead of rolling our own button style.   I am working on updating this pr with the latest uids4 styles and I'll push those up when I am finished.

bspeare:
I've update the UIDS4 hash. Run `blt frontend` to get the udpated styles.  And here's the markup that you would want to use:

```
<div class="bttn bttn--primary bttn--medium bttn--transparent bttn--light-font bttn--circle bttn--no-text">
<i class="fas fa-info-circle"></i>
</div>
```

bspeare:
Also, turn on the back to top button. I think we want this button placed at the same distance from the edge of the browser. 

Screen Shot 2022-12-07 at 1 12 34 PM

docroot/themes/custom/uids_base/uids_base.libraries.yml
```diff
@@ -354,6 +354,9 @@ node-type-article:
   css:
     component:
       assets/css/content/node--type--article.css: { preprocess: false }
+      assets/css/components/toggletip.css: { preprocess: false }
```
joewhitsitt: Should this be a separate library to only load if caption/global caption is available (within the if check of uids_base.theme) or if this is used on other content types in the future?

docroot/themes/custom/uids_base/assets/js/toggletip.js
```diff
@@ -0,0 +1,77 @@
...
+    var container = document.createElement('span');
+    container.setAttribute('class', 'toggletip-container');
+
+    // Put it before the original element in the DOM
+    toggletipText.parentNode.insertBefore(container, toggletipText);
+
+    // Create the button element
+    var toggletip = document.createElement('button');
+    toggletip.setAttribute('type', 'button');
+    toggletip.setAttribute('aria-label', 'more info');
```
joewhitsitt: I think we should call it what it is. "featured image caption"

joewhitsitt: actually, i think the label should be on the role status span if the intention is that it isn't read until clicked

pyrello: Counterpoint: we should leave it generic and/or potentially use another data attribute to populate the value rather than hardcoding it. To make it easier to apply to other things in the future.

joewhitsitt: did this js get lifted from somewhere? i am wondering from a screen reader perspective if this is good UX. tab to the icon which says "more info" and that it is a button. click/enter creates a new element that is read off (status) but without a change in focus. esc to close or tab to close/move off.

bspeare: I had suggested https://codepen.io/heydon/pen/Vzwdpy as a starting point from https://inclusive-components.design/tooltips-toggletips/.

cory-skeers: The original codepen: https://codepen.io/heydon/pen/Vzwdpy. It's been updated a bit, but that's largely where it originates.
There was a post where this codepen originated as well, but I don't have that handy...
In general, from my understanding this UX pattern is preferable for 'toggletips', where screen readers see it on focus/click, and subsequent clicks essentially re-build it in order to re-read it if necessary. That option to re-read is where the lack of changing focus is supposed to be a benefit.

But even with that intention, the specifics of labels weren't always clear from my reading, and might need re-thinking.

joewhitsitt: Thanks for pointing out the article. I understand the need to handle the mobile touch but it seems to take a step backwards from expected tabbing behavior.

Screen Shot 2022-12-07 at 4 41 25 PM

As a keyboard user, I can tab through the elements and have the information revealed/read.
As a mouse user, I can hover on the elements to reveal the information.
As a mobile touch user, I can tap on the elements to reveal/close the information.
As a screen reader, if I am expected to click/enter something, I want to know what it is that I am clicking into so that I am not wasting time and not worried about getting trapped.

>potentially use another data attribute to populate the value rather than hardcoding it. To make it easier to apply to other things in the future.

I think that is a good idea.

Also as a visual user, I don't know this content exists unless I hover over/tab through the image and notice it. Should it be visible all the time and not fly in?

I am not saying we need to adopt or completely mimic it, but what can we take from Webform's implementation of https://atomiks.github.io/tippyjs/ that seems to address the user stories above.

Screen Shot 2022-12-07 at 4 20 39 PM

docroot/themes/custom/uids_base/uids_base.libraries.yml
cory-skeers: Shifted this around. Thanks! We were going for a generic, reusable approach, so this definitely should be kept separate and attached as needed.

docroot/themes/custom/uids_base/assets/js/toggletip.js
cory-skeers: So, I think it's first a question of if we want it as a tooltip or toggletip (as I've been differentiating it when discussing it, but have also seen it as popuptip as in https://www.nngroup.com/articles/tooltip-guidelines/). Both have a lot of overlap, and generally fill the same role, but function differently on how/when the info displays and how/when it goes away. I'm open to either, but I think that might help between what functionality is an established user pattern for this type that needs to be improved (eg button label), and what might be an argument for one type of pattern over another.

Re: fly-in or appearance, there was some discussion in midsprint demo about its level of prominence and wishing to tuck it away. Here it should appear on hover or focus over anything within the banner in order to 'reveal it' if it's possibly relevant, but maybe there's another way to lessen its prominence without hiding completely? Potentially positioning, a smaller (but still visible) indicator that becomes more prominent/clear upon interaction?

cory-skeers: Adding to previous, I think if we kept it a toggletip, we'd want to update the aria-label to something like "image caption" (better, make it dynamic so that it can be included as image caption for this, but otherwise have different descriptions) so that it is more accurate in the description, but still succinct as intended.

joewhitsitt: Thanks for the library and style change. If this toggletip (new to me) behavior was decided during mid-sprint by the team, I am fine with shipping with this behavior to see how it is received by users. I would still like to see more context in the label wherever this is used.

cory-skeers: Nope! Still not decided. It's what was went with while working on it, and midsprint demo had feedback on the style, but it wasn't a solid decision on whether to use toggletip or tooltip. I think there's still room for discussion there, and whether we change it to tooltip, or roll with toggle and get feedback.

FWIW, toggletip as a concept is new to me too, I've just found various references to it since starting this work.

pyrello: > Re: fly-in or appearance, there was some discussion in midsprint demo about its level of prominence and wishing to tuck it away. Here it should appear on hover or focus over anything within the banner in order to 'reveal it' if it's possibly relevant, but maybe there's another way to lessen its prominence without hiding completely? Potentially positioning, a smaller (but still visible) indicator that becomes more prominent/clear upon interaction?

The thought was that by having it show on hover/focus, we would not have to display it on the image otherwise. This actually potentially allowed us to have the icon be more prominent (not need to use as much transparency and have a solid color icon) since it wouldn't be displaying all the time. When I suggested this, I was thinking of it more as something to try than as a decision per se.

quamsta: Way late to the party on this, but curious to know if an option with the caption always visible below the featured image was considered before landing on a tooltip/toggle tip? Depending on the article and image, the caption text could hypothetically be critical to understanding the context of the image/photo used in the article and may need to be more readily visible.

bspeare:
@quamsta Yes, we considered that option as well.  If you feel strongly about moving away from the toggle tip approach, now is a good time to chime in!

docroot/themes/custom/uids_base/assets/js/toggletip.js
cory-skeers: > Way late to the party on this, but curious to know if an option with the caption always visible below the featured image was considered before landing on a tooltip/toggle tip? Depending on the article and image, the caption text could hypothetically be critical to understanding the context of the image/photo used in the article and may need to be more readily visible.

So, we had brought it up as an option, but it we'd generally thought of that as the second option if the tool/toggletip didn't work out.

After thinking about it while working on it, I've also been wondering if there are some instances where we'd want it displayed directly on the page. One use-case that's been mentioned has been to list photo credits, and that to me seems like one instance where the credit should maybe not be tucked away. (In this case, might not be critical exactly, but good/courteous practice).

We could roll this out and later add the option to always display, but we might want some more discussion on those plans first.

quamsta: > @quamsta Yes, we considered that option as well. If you feel strongly about moving away from the toggle tip approach, now is a good time to chime in!

Definitely don't want to step on work that's already been done, but for things like image credits like @cory-skeers mentioned, they may be important to readily have available instead of putting them behind a user action. For the Mental Health site, I'm always putting image credits after the article text, but it'd be nice to have them up in the top area, where they'd make more sense imho. 

Quick-and-dirty mockup of an ideal credit:

screenshot_2022-12-08

(Would definitely need to be re thought for mobile)

> So, we had brought it up as an option, but it we'd generally thought of that as the second option if the tool/toggletip didn't work out.
> 
> After thinking about it while working on it, I've also been wondering if there are some instances where we'd want it displayed directly on the page. One use-case that's been mentioned has been to list photo credits, and that to me seems like one instance where the credit should maybe not be tucked away. (In this case, might not be critical exactly, but good/courteous practice).
> 
> We could roll this out and later add the option to always display, but we might want some more discussion on those plans first.

Good thinking. An additional option may work. Would it be beneficial to ask content editors or people who work on Iowa Now about the preferred option here?

cory-skeers: Thanks for the additional feedback, @quamsta!

I think the plan will be to push forward with a version similar to your mockup here to get the caption field functional and on display on the page, and create a followup to gather feedback and potentially offer other options.

quamsta: That sounds awesome. Appreciate the flexibility here! Thank you for being receptive to some later-stage feedback.

cory-skeers:
For rendering the caption directly, do we want at the bottom of the banner, or top of the body? I'd started with it directly below the title, but that needs extra accounting-for with the `banner-stacked`, whereas top of the body could be output directly.

image

joewhitsitt:
To me, above body seems like the easiest path forward for now unless the mobile stack thing affects that as well. Since featured image can be small and this field can be 255 characters on top of a title that could be just as long. Also, don't need to be concerned about small text contrast which i think has a higher requirement

However, it would be awesome if we could sprinkle in aria-labelledby for the image then if this caption is present since it would be disconnected in the DOM.

bspeare:
@cory-skeers I like @quamsta idea of putting it under the banner/above the body.  I also wonder if we should put the caption inline with the Image gallery block link?

Screen Shot 2022-12-09 at 9 33 14 AM

cory-skeers:
Quick example getting it in the meta. Not inline with the image gallery or styled yet, but wondering if this is starting to get too far away from the image. Aria labels could help screenreaders, but visually it's getting pretty disconnected if there's additional meta info like a source org, author, etc.

image

Might pivot to putting it above the rest of the meta, but open to other options

bspeare:
@cory-skeers I'd vote to move both above everything and be separated from the rest of the meta information.  We would possibly want to bring it closer to the image than where the current date is sitting.

cory-skeers:
Update with latest changes:

image

pyrello:
> Update with latest changes:
> 
> image

To me this is confusing because it seems like the caption is related to the image gallery link. This would be even more confusing if the caption didn't specifically mention the picture.

bspeare:
Another thought, maybe we introduce a sidebar column with some of the meta information?  https://github.com/uiowa/uiowa/issues/2813#issuecomment-767109876

Currently we have:

- Image Caption (possibly)
- Published Date
- Authors
- Source Org
- Source Org Link
- Image Gallery Link
- Issue (91st Meridian)

Maybe it is time to move a few of these into a sidebar column?  With the addition of the caption, and the publication issue from 91st Meridian work, this article meta information section is getting a bit unwieldy.

If we had the sidebar column, we could place the image caption below the featured image and structure the rest of the information into the sidebar.  

This would also dial back the long line-length width of the article.

quamsta:
Summary from a huddle with @e-marie-w  and @cory-skeers:

We worked through some CSS and were able to get the caption to display nicely on its own under the featured image on desktop-sized viewport, however, the mobile-friendly version of this eluded us due to the fact that the breadcrumbs and page title show directly below the image on mobile and the caption ends up below the article title:

(Rough mockup of the problem, not representative of the CSS work on the caption that's already been done)
Screenshot 2022-12-09 at 3 48 21 PM

We wanted to bring a few options to solve this mobile-friendly caption stacking issue to discuss with the group (if the non-sidebar option is where we're heading with this):

## Option A
On articles that specifically have featured images with captions, move the page title below the image (instead of overlaying it), disable the gradient, and keep the caption directly below the image. Articles without the caption field filled out would have the traditional overlaid text and gradient like they always have.

Example of an article + featured image the caption field filled out:
Screenshot 2022-12-09 at 3 56 40 PM

(also dropping a floated right option in here just for comparison's sake)
Screenshot 2022-12-09 at 4 11 10 PM

Mobile:
Screenshot 2022-12-09 at 4 12 47 PM


This option gives us a couple of advantages over some alternatives:

- It reduces the chance of caption text or the page title overlaying an important part of an image
- We increase a captioned image's visibility by removing the gradient

My reasoning for using this option that I discussed in the huddle is that if an image is important enough to have a caption, it may be important enough to stand on its own without text overlaying it or a gradient affecting its appearance.

This option also wouldn't affect existing articles that don't have the caption field filled out.

## Option B
Put the caption in the body of the text, but then using Javascript, we could move its place in the DOM to be above the page title and breadcrumbs. This sort of thing could maaaaybe be done in CSS, but it may be kind of dicey. The advantages of this may be that we don't have to rewrite a bunch of logic in PHP to accommodate articles with image captions.

## Option C
Overlay the caption into the heading like in @bspeare's mockup: https://github.com/uiowa/uiowa/pull/5993#issuecomment-1344406625

The advantages of this would be that we could do this without changing the markup or the DOM at all, but we may run into issues where too much text gets overlaid on top of the images. This could be alleviated by adjusting the image container size, but we'd also start bumping into the preset sizes of featured images for articles (e.g., the "Small", "Medium", "Large" options available for featured image sizes)

I personally think Option A is the way to go here, but willing to hear other ideas and would love feedback from the group! I'm a little hesitant to put a lot of meta information in the sidebar, but that may not be a bad idea either and this set of options wouldn't necessarily prevent that from happening IMHO

bspeare:
Thanks for putting together the mockups @quamsta!   I vote for option C and we could put work into mocking up future iterations with tooltips or removing the overlay.   It could be useful to be able to change the behavior of the featured image like I described in https://github.com/uiowa/uiowa/issues/2813#issuecomment-767109876.  Maybe we offer overlay, no overlay, and a portrait featured image or something like that?  That would enable users to keep the current behavior and provide new options if their page title is too long or their image is the wrong size.

pyrello:
> For rendering the caption directly, do we want at the bottom of the banner, or top of the body? I'd started with it directly below the title, but that needs extra accounting-for with the `banner-stacked`, whereas top of the body could be output directly.
> 
> image

I'm a fan of this option minus the gold icon. It seems like the field would need to go directly below the featured image in the header layout, which is hard because I think that region stopped rendering in LB at some point. You'll have to find it via inspector and enlarge it so that you can drag the field there. Or maybe you can use the Move option in the block's context menu.

cory-skeers:
Latest updates:

On regular screens:
image

On small screens where the banner stacks:
image

pyrello:
@cory-skeers We came up with basically the exact same way of resolving this without talking to each other.

docroot/themes/custom/uids_base/scss/content/node--type--article.scss
```diff
@@ -149,6 +149,44 @@ article.node--type-article {
...
   }
+
+  .block-field-blocknodearticlefield-image-caption {
+    font-style: italic;
+    color: #fff;
+    z-index: 3;
+    position: relative;
+    background: $dark-gray;
+    padding: 10px 20px;
+    order: 1;
```
pyrello: I'm not 100% about this, but I think it will infer the original order, so I don't think you don't need this.

pyrello:
Since we are moving away from the toggletip option for the meantime, we should probably not include in this PR. But it might be a good thing to explore in UIDS, so opening an issue there and linking to the relevant files in this PR would be a good way to preserve the work done on it in case we decide to use it at some point.

cory-skeers:
> Since we are moving away from the toggletip option for the meantime, we should probably not include in this PR. But it might be a good thing to explore in UIDS, so opening an issue there and linking to the relevant files in this PR would be a good way to preserve the work done on it in case we decide to use it at some point.

Sounds good! I'd been leaving it in here not having decided where to put it, but hoping the work could be used as a reference in the future.

Got an issue created https://github.com/uiowa/uids/issues/827 and (I'm pretty sure) the remaining bits taken out of here.

bspeare:
Screen Shot 2022-12-13 at 8 21 06 AM
Can we place the caption without so much extra space and a without a background color?  It seems like the banner gradient should be sufficient.   On mobile, we can just adjust the font color.

pyrello:
I removed the generic “links” field from all article view displays. I don’t think that links actually shows anything currently. But it could, and that could happen unexpectedly with a module update, so I think we should move away from having it enabled.

We should probably do this for all content types moving forward.

docroot/themes/custom/uids_base/scss/content/node--type--article.scss
```diff
@@ -149,6 +149,38 @@ article.node--type-article {
...
+  .block-system-breadcrumb-block {
+    order: 2;
+  }
+
+  .block-field-blocknodearticletitle {
+    order: 3;
+    @include breakpoint(sm) {
+      order: 2;
+    }
+  }
```
pyrello: I think, but am not totally sure, that we can omit these, since their values will be inferred.

cory-skeers: Yup! Needed a specified `-1` added to the caption, but after that, the rest falls in line.

joewhitsitt:
Sean mentioned the changes to the splits for display settings. Going to sync down a few sites to check for import errors/regressions.

bspeare:
What do people think about changing the weight/order of this markup to:

1. Image caption
2. Breadcrumbs
3. Page title

We would be able to cut down on some of the custom css and negative margins added here for the caption if this order changed and we could rely on the styles provided by the banner.

pyrello:
> What do people think about changing the weight/order of this markup to:
> 
> 1. Image caption
> 2. Breadcrumbs
> 3. Page title
> 
> We would be able to cut down on some of the custom css and negative margins added here for the caption if this order changed and we could rely on the styles provided by the banner.

I thought this was the order that things were in. I manually set weights on a lot of stuff, so I thought I had set it up that way.

bspeare:
In https://github.com/uiowa/uiowa/blob/c52e742b1d57438ca99bba1eed7c4a63b4960b0c/config/default/core.entity_view_display.node.article.default.yml, the weights are listed:

- Breadcrumbs: -10
- Page title: -9
- Caption: 2

Creates the source order of:

1. Breadcrumbs
2. Page title
3. Caption

joewhitsitt:
Screen Shot 2022-12-13 at 12 03 14 PM

Need to account for no featured image (still renders)

pyrello:
> In https://github.com/uiowa/uiowa/blob/c52e742b1d57438ca99bba1eed7c4a63b4960b0c/config/default/core.entity_view_display.node.article.default.yml, the weights are listed:
> 
> * Breadcrumbs: -10
> * Page title: -9
> * Caption: 2
> 
> Creates the source order of:
> 
> 1. Breadcrumbs
> 2. Page title
> 3. Caption

My bad. Whatever gets done to fix this in default, make sure it gets propagated to the splits.

cory-skeers:
> Screen Shot 2022-12-13 at 12 03 14 PM
> 
> Need to account for no featured image (still renders)

Shoot. Think we lost the accounting for this somewhere in the shuffle.

cory-skeers:
> What do people think about changing the weight/order of this markup to:
> 
> 1. Image caption
> 2. Breadcrumbs
> 3. Page title
> 
> We would be able to cut down on some of the custom css and negative margins added here for the caption if this order changed and we could rely on the styles provided by the banner.

I'd be open to it. For both the full display and mobile display?

bspeare:
I think we want the html source order to be as described in https://github.com/uiowa/uiowa/pull/5993#issuecomment-1349491096.  Then we adjust the order for desktop through flex order.

joewhitsitt:
https://github.com/uiowa/uiowa/pull/5993#issuecomment-1349150152

Didn't notice any issues related to the splits.

cory-skeers:
Display order has been updated, and this should be reflected across the various splits.
Cleaned up the scss a bit as a result. There's now minimal flex ordering needed, and only a couple small additions to margins where there are things like an overlap with the title.
Unset the image caption if there is no image to which it should be attached on render

cory-skeers:
Setting back to draft--realized the case that there is an image and a caption, but the image is not set to display has not been accounted for.

cory-skeers:
> Setting back to draft--realized the case that there is an image and a caption, but the image is not set to display has not been accounted for.

Updated this with some logic in `uids_base.theme` so 1) it'll work if we add captions to other content types w/featured images and 2) can bootstrap off of some code we already had that handled checking node and site settings for display. I think there's enough coverage checking that the field exists and is empty that it won't affect anything else, but would appreciate some other eyes.

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -552,6 +552,33 @@ function uids_base_preprocess_layout(&$variables) {
 
         if (!$variables['featured_image_display']) {
           unset($variables['content']['background']);
+          // If we don't have a featured image,
+          // but have a featured image caption set,
+          // find it and unset it.
+          if ($node->hasField('field_image_caption') && !$node->field_image_caption->isEmpty()) {
```
pyrello:
```suggestion
          if ($node->hasField('field_image_caption') && !$node->field_image_caption?->isEmpty()) {
```

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -552,6 +552,33 @@ function uids_base_preprocess_layout(&$variables) {
...
+            // Loop through regions in the layout.
+            foreach (Element::children($variables['content']) as $region) {
+              // Loop through blocks in that region.
+              foreach (Element::children($variables['content'][$region]) as $block_uuid) {
+                if ($region === 'content'
+                  && isset($variables['content'][$region][$block_uuid])
+                  && isset($variables['content'][$region][$block_uuid]['#plugin_id'])
+                ) {
+                  // Check if this block plugin_id matches the pattern for a
+                  // page title block.
```
pyrello:
```suggestion
                  // Check if this block plugin_id matches the pattern for a
                  // field_image_caption block.
```

pyrello:
> > Setting back to draft--realized the case that there is an image and a caption, but the image is not set to display has not been accounted for.
> 
> Updated this with some logic in `uids_base.theme` so 1) it'll work if we add captions to other content types w/featured images and 2) can bootstrap off of some code we already had that handled checking node and site settings for display. I think there's enough coverage checking that the field exists and is empty that it won't affect anything else, but would appreciate some other eyes.

My hunch is that this code should actually be in `uiowa_core` and not the theme since it is more about business logic than theming logic. However, we are not very consistent about how we approach this.

As an aside, I have been wondering throughout this project whether we might not want to consider using a paragraph to represent the image + caption and then reuse that in the banner and other content types. I like the idea of packaging the two things together since they should be considered and acted upon together. However, that will most certainly be a project for another day.

joewhitsitt:
@pyrello were you reviewing this? If not, I can take a look after lunch

pyrello:
> @pyrello were you reviewing this? If not, I can take a look after lunch

Go ahead, @joewhitsitt. I have not gotten to it yet.

config/default/field.field.node.article.field_image_caption.yml
```diff
@@ -10,7 +10,7 @@ field_name: field_image_caption
 entity_type: node
 bundle: article
 label: Caption
-description: 'This field is currently not rendering to visitors and will be addressed in a later update.'
+description: 'Provide a caption for this featured image. A tooltip will render on the image on the page to display the caption.'
```
joewhitsitt: Need to update this comment based on the shift away from tooltip

cory-skeers: Good catch--I think if it renders directly as a caption, there's not much else to add beyond "Provide a caption for this featured image."

joewhitsitt: Resolved by d10baf8abbc443194f31b9400b18b6983388c561

joewhitsitt (approved):
Overall I think what is here is good. Curious if you looked into, "... it would be awesome if we could sprinkle in ~aria-labelledby~  aria-describedby for the image..."

I might open a separate issue to push for that unless you discovered something while working on this where we can't/shouldn't do it.

cory-skeers:
> Overall I think what is here is good. Curious if you looked into, "... it would be awesome if we could sprinkle in ~aria-labelledby~ aria-describedby for the image..."
> 
> I might open a separate issue to push for that unless you discovered something while working on this where we can't/shouldn't do it.

I'd looked into it back when it was a toggletip, but it didn't make it over here. There aren't any blockers for it as far as I'm aware--a followup issue to add that would be great.

PR #6038, opened 2022-12-16
SiteNow Microsite

joewhitsitt:
This feature creates a separate content type that could be governed by groups to allow the creation of a sub-section (microsite) within a website. The content type would be the landing page with the ability to nest an unlimited* number of paragraph entities inside it that captures a page title and alias slug for sub pages at a single level. The paragraph type itself could be expanded to layout builder functionality (similar to region fragments) as well as capture other page information (e.g. SEO values). The ordering of the paragraph entities within the content type could serve as the ordering for a custom menu that is output on the content type and the subpages. Groups could also govern media assets so that users are touching other editors' assets on the site.

Screen Shot 2022-12-16 at 3 07 00 PM
Screen Shot 2022-12-16 at 3 07 53 PM


joewhitsitt:
Started think about microsite structure a bit more. Originally it was content type > paragraph nested and I ported it to this PR. Paragraphs can have layouts, but I can't imagine the option to allow each a customized layout to actually work since there isn't a route dedicated to each paragraph item. Fragments has been used to provide this. Fragments are lighter than nodes and can still provide the same Sign -> Slide relationship that Signage uses but access control will have to be more involved than with Paragraphs. I want to avoid nested paragraphs (v2) since we reduce variations of the same functionality. Paragraphs Layouts and Gutenberg would probably be the same. Worse case, microsite would be a less customized solution for particular use-cases where custom sub page layouts are necessary.

Screen Shot 2022-12-30 at 4 19 48 PM

bspeare:
Just noticed this article https://www.lullabot.com/articles/microsites-drupal.

joewhitsitt:
Yeah, I think their definition is a bit different but an interesting read.

joewhitsitt:
This was my original inspiration for this a long time back: https://www.drupalgovcon.org/sites/default/files/session/slides/2019-07/The%20Single%20Node%20Website%20%285%29_0.pdf

joewhitsitt:
Everyone is trying the microsite thing:

https://fivejars.com/blog/building-microsites-drupal

hansel is so hot right now

Issue #4241, opened 2021-10-01
Require Authentication to view all content on SiteNow site


briand44:
# Story

As a SiteNow v3 site owner, I would like to have all my content behind authentication so it is not available publicly.

# Additional Info
It would be nice if there was some sort of option or split that could be enabled that would automatically make it so authentication was required to view content. That same option or split would also turn on all of the related functionality in future slices.

# Related Issues
#4242 
#4243 
#4244 

# Proposed solution

briand44:
Doesn't need to be a checkbox to enable this, it could be a manual process that needs to happen. If so, we should document the process.

briand44:
Want to avoid the possibility of this accidentally getting turned off/on and causing issues or displaying content that should not be public.

mjoneill:
For consideration for sites that require authentication to view content:

* We need a search solution that doesn't rely on Google indexing public pages
* robots.txt should likely be disallow all bots and agents
* Sitemap.xml should NOT disclose pages that are not accessible without authentication.
* Each page should have the meta noindex, nofollow tag set in the meta.  <--- extra layer of protection in case the pages are made visible by accident

Undoubtedly more as we think through this further...

pyrello:
I haven't read this yet, but it may be relevant: https://robertroose.com/blog/drupal/how-create-groups-within-drupal-which-can-share-nodes-and-files-privately

joewhitsitt:
if the first slice is just stopping folks at the bootstrapped drupal level this looks like an easy step forward that could be separate in a feature split: https://www.drupal.org/project/require_login

cory-skeers:
> if the first slice is just stopping folks at the bootstrapped drupal level this looks like an easy step forward that could be separate in a feature split: https://www.drupal.org/project/require_login

I'm having trouble getting this set up to test locally. Visiting any page (including the login page) redirects to the homepage, and I'm not sure why. If the homepage is excluded from needing to log in, that's the end of it. If it's not excluded, a redirect loop occurs between the home and login pages.

Don't see any issues in the module, and saml routes seem to be properly excluded to where I'd think there wouldn't be an issue there.

cory-skeers:
https://www.drupal.org/project/restrict_ip seems to work well out of the box in initial testing. By blocking all IP addresses (or allowing a select few like local connections) and allowing specific roles to bypass the blocking, it can be used to lock down all or parts of a site except for specific roles, which would work for assigning AD groups in the future.

[...]

PR #6040, opened 2022-12-19
Intranet login route

### LINKED ISSUE #5911: Ability to route user after login to a specific page
https://github.com/uiowa/uiowa/pull/5893#issuecomment-1295052384

Currently when a user authenticates to an intranet, they are routed to the home page (see https://github.com/uiowa/uiowa/pull/5893).

It would be great if someone was able to share a link to https://studentlifeconnect.prod.drupal.uiowa.edu/training-and-development and after the user logs in they would be routed to that page.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5911. 

# To test
- Build passes (all intranet tests pass)
```
ddev blt ds && ddev drush @default.local en sitenow_intranet -y && ddev drush @default.local cim -y
```
- https://default.uiowa.ddev.site/ should show Unauthorized title and message.
- You should see a Hawkid link with https://default.uiowa.ddev.site/saml/login?destination=/ as the URL.
- https://default.uiowa.ddev.site/about Hawkid link should show https://default.uiowa.ddev.site/saml/login?destination=/about as the URL
```
ddev drush @default.local uli admin/people/create
```
- Add a user with no other roles than authenticated with the username `only_auth`.
- Masquerade as users with other roles and click around to ensure that you have access to everything you would expect.
- Log out.
```
ddev drush @default.local uli --name only_auth
```
- You should be able to access the user view and edit tabs.
- Click a link in the menu
- You should see the Access Denied title and message
- You should also see a logout link.


joewhitsitt:
I pushed up a change with some known problems:

- Are we going to duplicate all of the RestrictIPEventSubscriber checks?
- The page login button gets cached.

joewhitsitt:
i think there is enough to move forward with here that doesn't rely on a patch to restrict_ip, but we are essentially adding our own wall in doing so.

>Maybe our use case is narrow enough that we don’t need [restrict_ip] and could just get by with our own event listener logic

>if we build a lite version we need to make sure we can disable it without logging in (drush toggle).

https://iowaweb.slack.com/archives/C8SM82ZFZ/p1671463751429749

pyrello:
I did an investigation to see what it might take to just ditch the `restrict_ip` module in favor of our own solution. The results were promising, but maybe a bit too much to fit into what we are trying to accomplish this sprint.

Here is what I did to get it working:
1. Removed the dependency on `restrict_ip` from `sitenow_intranet` and uninstalled `restricted_ip`.
2. SitenowIntranetSubscriber.php:
```php
<?php

namespace Drupal\sitenow_intranet\EventSubscriber;

use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\HttpKernel\Event\RequestEvent;
use Symfony\Component\HttpKernel\Event\ResponseEvent;
use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
use Symfony\Component\HttpKernel\HttpKernelInterface;
use Symfony\Component\HttpKernel\KernelEvents;
use Symfony\Component\HttpFoundation\RedirectResponse;
use Drupal\Core\Url;

/**
 * Sitenow intranet event subscriber.
 */
class SitenowIntranetSubscriber implements EventSubscriberInterface {

  /**
   * Kernel request event handler.
   *
   * @param \Symfony\Component\HttpKernel\Event\RequestEvent $event
   *   Response event.
   */
  public function onKernelRequest(RequestEvent $event) {
    if (\Drupal::currentUser()->isAnonymous()) {
      $status_code = $event->getRequest()
        ?->attributes
        ?->get('exception')
        ?->getStatusCode();
      if (is_null($status_code) || !($status_code === 401 && $event->getRequestType() === HttpKernelInterface::SUB_REQUEST)) {
        $route_name = \Drupal::request()->attributes->get('_route');
        if (!in_array($route_name, ['user.reset.login', 'samlauth.saml_controller_login'])) {
          throw new UnauthorizedHttpException('Login, yo!');
        }
      }
    }
  }

  /**
   * {@inheritdoc}
   */
  public static function getSubscribedEvents() {
    return [
      KernelEvents::REQUEST => ['onKernelRequest'],
    ];
  }

}
```
3. Added the following to `uiowa_core_preprocess_block()`:
```php
    case 'system_main_block':
      if ($current_route->getRouteName() === 'system.401') {
        $variables['content']['login_link'] = [
          '#type' => 'link',
          '#title' => t('HawkID Login'),
          '#url' => Url::fromRoute('samlauth.saml_controller_login', ['destination' => \Drupal::request()->getRequestUri()]),
          '#prefix' => '<p class="text-align-center">',
          '#suffix' => '</p>',
          '#attributes' => [
            'class' => [
              'bttn',
              'bttn--primary',
              'bttn--caps',
            ],
          ],
        ];
      }
      break;
```

There are other things that would need to be resolved, but this got things to a working state.

pyrello:
So, I just realized that the reason some of the tests are failing is that the terminology is backwards in the current intranet code. It related a `401` exception to being the Access Denied (Forbidden) status code and `403` to the Unauthorized status code.

**401**: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
**403**: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

To avoid future confusion, I think we should swap these around. The code I have written already reflects this as I wrote it having read the above docs. The downstream change is that we are going to be necessary is to swap around the titles and messages for the error pages in the existing intranet projects once this goes out. Since none are live yet and there are only a couple provisioned so far, this shouldn't be a big deal.

pyrello:
This is ready for testing except for that I need to write testing instructions.

joewhitsitt:
pulling down

joewhitsitt:
Maybe it is a gap from before but footer links and social links are showing.

Screen Shot 2022-12-22 at 10 04 45 AM

joewhitsitt:
I am seeing unauthorized message on paths as an anonymous user and the access denied message as authenticated user. Our settings page has this help text:

>The Unauthorized page will be displayed to users that have logged in but lack the proper role/permission to access the site.

>The Access denied page will be displayed to users that have not logged in yet.

Should we update these? admin/config/sitenow/intranet

cory-skeers:
> Maybe it is a gap from before but footer links and social links are showing.

I _think_ this was a gap from before, and was considered unfortunate but acceptable with the 'workaround' being to not include any sensitive labels/paths there.

pyrello:
> I am seeing unauthorized message on paths as an anonymous user and the access denied message as authenticated user. Our settings page has this help text:
> 
> > The Unauthorized page will be displayed to users that have logged in but lack the proper role/permission to access the site.
> 
> > The Access denied page will be displayed to users that have not logged in yet.
> 
> Should we update these? admin/config/sitenow/intranet

Yep.

pyrello:
> > Maybe it is a gap from before but footer links and social links are showing.
> 
> I _think_ this was a gap from before, and was considered unfortunate but acceptable with the 'workaround' being to not include any sensitive labels/paths there.

It's pretty easy to add additional regions to the unset logic. Not sure about the social media icons, but the footer menus should be easy to add to that.

joewhitsitt:
Trying to add an after content fragment button block:

```
An AJAX HTTP error occurred.
HTTP Result Code: 200
Debugging information follows.
Path: /layout_builder/add/block/overrides/fragment.1/0/content/inline_block%3Auiowa_button?_wrapper_format=drupal_dialog.off_canvas&ajax_form=1
StatusText: parsererror
ResponseText: Error: Call to undefined method Drupal\Core\Form\FormAjaxException::getStatusCode() in Drupal\sitenow_intranet\IntranetHelper::getStatusCode() (line 20 of /var/www/html/docroot/modules/custom/sitenow_intranet/src/IntranetHelper.php).
```

Update: Getting this for all layout builder block adds

UPDATE: later commits resolved this.

cory-skeers:
We should make a followup for updating composer to remove `restrict_ip` and its patch, and not make that change in this same PR, right?

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
```diff
@@ -10,97 +10,91 @@
  */
 
 use Drupal\Core\Url;
+use Drupal\sitenow_intranet\IntranetHelper;
 
 /**
- * Implements hook_restrict_ip_access_denied_page_alter().
+ * Implements hook_preprocess_page().
```
cory-skeers:
```suggestion
 * Implements hook_preprocess_block().
```

cory-skeers (dismissed):

joewhitsitt:
taking another look

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
```diff
@@ -10,97 +10,91 @@
...
+  $code = IntranetHelper::getStatusCode();
 
-    // If the user is not authenticated, return a 401.
-    $page['#attached']['http_header'][] = ['Status', '401'];
+  if (IntranetHelper::checkStatusCode($code)) {
+    foreach ([
+      'action_menu',
+      'footer_first',
+      'footer_second',
+      'primary_menu',
```
joewhitsitt, line 91:
```suggestion
      'pre_footer',
      'primary_menu',
      'region_content',
```

what about alerts? hawk alerts and custom alerts are showing

pyrello: Good question. We're kind of getting into new territory here, since this is locking down more than what the old system locked down. If we are going to remove other things from being displayed, then we probably want to do that as a follow up so we can have some discussion around it?

Removing `action_menu` will cause our test to break. We may want to consider updating the test to place the menu in `primary_menu`? Or add a separate test to check that?

joewhitsitt: oh. i didn't intend to remove action_menu. i think github is highlighting the area wrong. just adding prefooter alphabetically, keeping primary menu and adding region_content (after content). These regions were added after Intranet was made. Alerts was a missed region the first pass. "an important announcement (custom alert) to the intranet viewers about not talking to the press about xyz would be visible"

joewhitsitt:
@pyrello mentioned in planning that we will want to contact Intranet customers to coordinate the flip in messaging https://github.com/uiowa/uiowa/pull/6040/commits/47678d3913c8bf34c1a35a4d9dd555053b187fa7

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
pyrello: I'm wondering about whether we might want to consider just suppressing custom messages? Or even adding a setting to suppress custom messages? The hawk alerts seem like they should still be shown. And there could be a valid use case for wanting to display a message in the alert area that 401/403 users should see.

joewhitsitt:
From what I have reviewed so far, the issues I originally raised have been resolved. From a replace restrict_ip and redirect after login, this PR looks good to me. Though not originally in-scope, I am not giving it approval at this time because there is content (after content and alerts regions) that is visible to anonymous visitors that I am not comfortable leaving for a follow-up now that it has been noticed.

In addition to the testing instructions above, I made sure the private file system is working as expected and that the disallow robots.txt is still accessible.

joewhitsitt:
One thing I didn't test was the layers of caching. Might be good to put this out on DEV to test varnish and all.

pyrello:
Re: https://github.com/uiowa/uiowa/pull/6040#discussion_r1055796982

After thinking about it for a bit, I think starting with a more restrictive setup and then introducing some exceptions based on that being requested would make the most sense.

cory-skeers:
@briand44 and I chatted today about this, specifically the alerts region and wanting to hide that content from unauthorized users. Ideally Hawk Alerts would still be visible.

I made a change here to unset the `custom_alerts` section of the `uiowa_alerts_block` for unauthorized users. Originally it was targetting `custom_alerts`, but that felt off to me, on the off-chance that other alert types might be added, or anything changed with that type of wrapper that could introduce a hole in the future. It now targets any content in that block except `hawk_alerts`, which still leaves the hole if other blocks are added to the `alerts` region, but keeps this one a little tighter.

^ This logic could use some review and testing, but otherwise I think the main thing left to do with this PR is to get it into Dev and put it through some cache testing.

pyrello:
Merged into `develop` so we can test caching.

joewhitsitt:
To test caching I was just thinking of hitting different paths to change the login destination (button link) and then visiting page/file paths while logged in to see if they get cached for anonymous visitors.

https://github.com/uiowa/uiowa/issues/4241#issuecomment-1004174535

We may have issues in the future if we try to do ip restriction with our in-house solution - https://github.com/uiowa/uiowa/issues/4241#issuecomment-1010331836

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
```diff
@@ -10,97 +10,105 @@
...
 use Drupal\Core\Url;
+use Drupal\sitenow_intranet\IntranetHelper;
 
 /**
- * Implements hook_restrict_ip_access_denied_page_alter().
+ * Implements hook_preprocess_block().
  */
-function sitenow_intranet_restrict_ip_access_denied_page_alter(&$page) {
-  $config = \Drupal::config('sitenow_intranet.settings');
-  \Drupal::service('renderer')->addCacheableDependency($page, $config);
```
joewhitsitt, line 19: @pyrello do you recall why this was removed/not moved to the preprocess page?

pyrello: Mostly because I wasn't sure if it was necessary. I believe there is a todo message about checking what we need to do for caching.

joewhitsitt: I am a bit confused about this. DEV is showing private uncachable on the requests and unless restrict_ip is doing something else, this was the only thing I saw related to caching that was changed in this PR. That doc says:

`Otherwise, the passed in object must be assumed to be uncacheable, so max-age 0 is set.`

My guess would be to add this back in and go from there. 🤷

joewhitsitt: Rick had done this instead of changing the performance cache setting from 1 month for the split as seen on studentlife.connect PROD

pyrello: I think restrict IP must have been doing something else. Just from looking at this, my guess would be that its effect would be to clear the cache when the config settings are changed.

pyrello: OTOH I don't see any instances in `restrict_ip` where any class or method with the word 'cache' in it.

joewhitsitt: >would be to clear the cache when the config settings are changed

which we also want correct?

in restrict_ip doing a string search, only the readme mentions `cache`. no results for invalidate or other cache terminology.

pyrello: This was re-added in a new place.

pyrello:
I believe this is ready for testing again on https://default.dev.drupal.uiowa.edu

pyrello:
# Testing on dev
- https://default.dev.drupal.uiowa.edu/ should show Unauthorized title and message.
- You should see a Hawkid link with https://default.dev.drupal.uiowa.edu/saml/login?destination=/ as the URL.
- https://default.dev.drupal.uiowa.edu/about Hawkid link should show https://default.dev.drupal.uiowa.edu/saml/login?destination=/about as the URL
```
ddev drush @default.dev uli admin/people/create
```
- Add a user with no other roles than authenticated with the username `only_auth`.
- Masquerade as users with other roles and click around to ensure that you have access to everything you would expect.
- Log out.
```
ddev drush @default.dev uli --name only_auth
```
- You should be able to access the user view and edit tabs.
- Click a link in the menu
- You should see the Access Denied title and message
- You should also see a logout link.

pyrello:
@cory-skeers I just dismissed your review so you don't have to be the one to give it the approving review.

docroot/modules/custom/sitenow_intranet/sitenow_intranet.module
```diff
@@ -10,97 +10,107 @@
...
+      'footer_second',
+      'pre_footer',
+      'primary_menu',
+      'region_content',
+    ] as $region) {
+      if (isset($variables['page'][$region])) {
+        unset($variables['page'][$region]);
+      }
+    }
+    // @todo Do we need to set cache max-age to 0 here?
```
joewhitsitt:
```suggestion
```

remove todo

joewhitsitt:
Confirmed that the custom alerts don't show and that hawk alerts do by changing the hawk alert source to the test endpoint.

joewhitsitt (approved):
_What a Long Strange Trip It's Been_ for this issue, but I am happy with the results. 

Created https://github.com/uiowa/uiowa/issues/6056 to remove restrict_ip

I'd be interested in layering back in some caching similar to https://intranet.oneit.uiowa.edu/ (d7) has once we understand more about how to cache and provide the destination

pyrello:
Thanks @joewhitsitt for all your help testing!

joewhitsitt:
Sites with intranet split on:

intradent.dentistry.uiowa.edu (uiowa)
grantbulletin.research.uiowa.edu (uiowa)
thelightbox.radiology.uiowa.edu (uiowa03)
connect.studentlife.uiowa.edu(uiowa03)

Talked in standup about just doing the swap for the customers upon deployment.

joewhitsitt:
The message swap has been performed on the above websites.

PR #6046, opened 2022-12-21
Updated snapshot to hide hawk alerts

bspeare:
# How to test
If the percy build passes, approve?


pyrello (approved):
Approved based on percy passing.

Issue #6030, opened 2022-12-15
Come up with a better solution for dealing with false positives for empty fields than `|striptags|trim`


pyrello:
# Motivation

Related to #6029. We have a lot of instances where we have adopted a pattern of checking values in twig templates against `|striptags|trim`. If I am remembering correctly, we started doing this because otherwise, those fields would contain debugging HTML markup when rendered in a local environment and that would cause unexpected behavior when working on twig templates. The `|striptags|trim` removes HTML comments and spaces before the `is empty` comparison is made, so we can be sure that something only renders if there is actual content. However, those filters can only be applied to strings, so that means we have ensure that anything we pass to one of those checks has already been rendered. This means that we are adding code to manually do a process that Drupal will automagically do for us if we just pass the render arrays.

It is also worth noting that this is only an issue locally, so we have all this logic in place to work around a problem that doesn't occur in production.

# Proposed solution
Write a service that can be called anywhere in the code (probably via `uiowa_core`) that can pre-check a field to see if it is just HTML comments. Then we can decide whether to pass the field or not in preprocessing. This would potentially allow us to remove most or all of our `|striptags|trim` checks in twig templates.

PR #6053, opened 2022-12-23
People list links directly to site

### LINKED ISSUE #5722: Add ability to have people in lists link directly to first website
# Motivation

On https://uicb.prod.drupal.uiowa.edu/associates, UICB would like to just have the record directly link to an external website to prevent the user from having to click through to the profile then click a link to a website.

This is kind of similar to the "Link directly to source" feature with articles.

# Proposed solution

- Adapt logic for article content type to "link directly to source" to work for person content type teaser display.
- Add boolean field to person content type to "link directly to first website".
- If there are multiple website links, use the first one.

e-marie-w:
Fixes #5722 
# To fix still:

- [x] Anonymous users are not being redirected as expected when they have the node's url
- [x] Need to reset checkbox value to 0 when user removes all websites (otherwise can get 'Visitors are being redirected' message when no websites are linked).

# How to test

- checkout branch and pull, sync a site (I've been using default - `ddev blt ds --site=default`)
- Create or edit existing person nodes, adding websites to a couple and checking the 'Link directly to website' checkbox for at least one.
- Create a page with the people list block.
- For the default people listing (/people) and your list block, click on the people. People with no website and people with website but unchecked box should send you to the regular person node. People with website and checked box should send you directly to the site you added. 
- If you leave the checkbox checked but remove the site(s), save, and check the lists again it should send you to the regular node also. 
- While logged in you can still access the person node for people who link directly to a site. Copy the link for one of these nodes and open in a private/incognito tab. It should redirect you to the site now instead.
- Site splits to test:
  - admissions.uiowa.edu
  - immuno.grad.uiowa.edu 
  - isa.uiowa.edu
  - now.uiowa.edu
  - pharmacy.uiowa.edu
  - sppa.uiowa.edu
  - uipress.uiowa.edu
- Feature splits to test:
  - event
  - grad_person_extended


joewhitsitt:
Several sites have split `core.entity_view_display.node.person.*` and `core.entity_form_display.node.person`. Have those been tested against/prepared to accept this change?

https://github.com/uiowa/uiowa/find/main

e-marie-w:
> Several sites have split `core.entity_view_display.node.person.*` and `core.entity_form_display.node.person`. Have those been tested against/prepared to accept this change?
> 
> https://github.com/uiowa/uiowa/find/main

I started testing, but did not check for those specific files. Moving back to draft since there is a lot.

pyrello:
> Several sites have split core.entity_view_display.node.person.* and core.entity_form_display.node.person. Have those been tested against/prepared to accept this change?

I would be primarily concerned with the `core.entity_view_display.node.person.*`, since those would be split completely. For `core.entity_form_display.node.person.*`, I would expect those changes would primarily be weights, which can self-resolve if we don't explicitly re-export them.

e-marie-w:
So far, having some trouble updating the `core.entity_view_display.node.person.default` splits. I thought there was a way to re-import or re-export. Nothing's worked so far, though, so now I'm thinking I just need to recreate the new field for each. Which makes sense for those sites where the`core.entity_form_display.node.person.default` is also completely split I guess. 

Update: wait nevermind the 'I ask for help and then figure it out 30 seconds later' thing has happened once again.

joewhitsitt:
>Update: wait nevermind the 'I ask for help and then figure it out 30 seconds later' thing has happened once again.

Good! Let me know if you want someone else to look at this with you.

e-marie-w:
I'm having trouble understanding the SPPA split. `core.entity_view_display.node.person.default` and `core.entity_form_display.node.person.default` exist for it, but are not actually listed in the site split. 

If I add `core.entity_view_display.node.person.default` to the complete split list it exports as expected. (The other person view displays already have config_split.patch files and work correctly).

When I save the person form display and export `core.entity_form_display.node.person.default` it updates. But why is the form display completely split? It seems like it shouldn't be, so I tried adding `core.entity_form_display.node.person.*` to the partial split. But it still exports like it's part of the complete list.

docroot/themes/custom/uids_base/templates/content/node--person--teaser.html.twig
```diff
@@ -88,6 +88,12 @@
   {% set headline_level = heading_size %}
 {% endif %}
 
+{% if '1' in node.field_person_website_link_direct.0.value and node.field_person_website.0.uri %}
+  {% set node_link = node.field_person_website.0.uri %}
+{% else %}
+  {% set node_link = url %}
+{% endif %}
```
pyrello, line 95: Wondering if there was a reason to do this in the template vs in a preprocess function?

docroot/themes/custom/uids_base/templates/views/views-view-unformatted--list-card.html.twig
```diff
@@ -42,6 +43,12 @@
     {% endif %}
 
   {% endfor %}
+
+  {% if '1' in row.content['#node'].fields.field_person_website_link_direct.value  %}
+    {% set node_link = view.style_plugin.getField(key, 'field_person_website')|striptags|trim %}
+  {% else %}
+    {% set node_link = view.style_plugin.getField(key, 'view_node')|striptags|trim %}
+  {% endif %}
```
pyrello, line 51: Same note as above. Also wondering about why `|striptags|trim` is being used here?

e-marie-w: I thought in midsprint demo we decided to keep it in the template. Partially due to concerns the template was being used by more than just sitenow people, and partially because of the UIDS card work. There was something about the card work involving refactoring anyways, so this would be temporary and could be kept as simple as possible. 

`|striptags|trim` was added because at one point the link wouldn't work due to html tags being added. I will double-check if this is still happening/still necessary.

pyrello: > I thought in midsprint demo we decided to keep it in the template.

Thanks for reminding me of this. Yes, that makes sense and let's keep it this way for now. Maybe add a comment to the effect that it intended to be temporary so that it is clear for the next person looking at it that they should not use it as an example of how to do things.

> `|striptags|trim` was added because at one point the link wouldn't work due to html tags being added. I will double-check if this is still happening/still necessary.

Okay, I see now. Thanks for the explanation.

joewhitsitt:
@e-marie-w the WSOD traces back to sitenow_people_person_type_statuses_allowed_values and how it appears that none of the person types for uipress allow "former." For lb display settings to generate sample values, there needs to be an option for it. A temporary work around for this PR would be to give one of the person types the former option here https://uipress.uiowa.ddev.site/admin/structure/person-type. That will allow you to load the display settings. I will briefly look into a better solution to this or open an issue.

joewhitsitt:
i started a big dsa to start testing and admissions is tripping up on core.entity_form_display.node.person.minimal and core.entity_view_display.node.person.token because of `field_person_type_status` being removed from the hidden section. is that expected?

update: prod has the field hidden for both. i will see about correcting that config.

e-marie-w:
> i started a big dsa to start testing and admissions is tripping up on core.entity_form_display.node.person.minimal and core.entity_view_display.node.person.token because of `field_person_type_status` being removed from the hidden section. is that expected?

So, I was seeing changes around that when I exported but I didn't understand what we actually wanted so I uh ignored those changes and just committed the ones related to the default displays. If it's what we want I can commit those changes too.

Edit: you beat me to it, thanks!

joewhitsitt:
Kept trying to get uipress config_split.patch.core.entity_form_display.node.person.default to work but it kept wanting to remove `region: content` config and was also messing up the disabled fields, credentials, research areas, teaser, education. Trying complete split. https://github.com/uiowa/uiowa/pull/6053/commits/7ba3a54afd1b9519348bffb3ea25849334aaadfe

docroot/themes/custom/uids_base/templates/views/views-view-unformatted--list-card.html.twig
e-marie-w: Good news! 1. I remembered to go back and check this, and 2. `|striptags|trim` is no longer needed, it works as it should without it.

e-marie-w: Wait I was wrong I looked at the /people view instead of the People listblock. It is still needed for the listblock to work.

docroot/themes/custom/uids_base/templates/content/node--person--teaser.html.twig
joewhitsitt: @pyrello based on https://github.com/uiowa/uiowa/pull/6053#discussion_r1064710950 can this be resolved?

docroot/themes/custom/uids_base/templates/views/views-view-unformatted--list-card.html.twig
joewhitsitt: Can we leave striptags|trim in until we address them in https://github.com/uiowa/uiowa/issues/6030? If so, based on https://github.com/uiowa/uiowa/pull/6053#discussion_r1064710950 can this also be resolved?

joewhitsitt:
Though biased, I am going to pull and test this

Update: After I fix my local...

joewhitsitt:
Sorry it took so long for me to notice but I mentioned to Marie that the rabbit hole stuff is not quite right and referenced the work we did for AoS and how they need the permission to get the rh stuff on the form to alter/manipulate on save: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_area_of_study/uiowa_area_of_study.module

Beyond that, I think everything else checks out.

bspeare:
Is the pre-save function the only thing left to test on this? If so, what are the steps to test there?

e-marie-w:
> Is the pre-save function the only thing left to test on this? If so, what are the steps to test there?

Actually, Joe noticed a few extra things Areas of Study has that would be useful for this as well (mostly messages about the rabbit hole redirects), so I'm currently trying to get those added. I forgot to change it back to draft but it is now.

e-marie-w:
Sooo when I fixed the issue where the person node was redirected to site link regardless of checkbox status, I apparently broke it in a different way: right now visitors can see the node if they have the url even when the checkbox is checked.

joewhitsitt:
@e-marie-w just saw your comment. is this still an issue? is it because the config for the node type still has the token?

https://github.com/uiowa/uiowa/pull/6053/files#diff-3386385cb6c3e53d9f486a269f6956f124147f76a62541b7ac519748f0fd0d83 should be more like https://github.com/uiowa/uiowa/blob/main/config/features/areas_of_study/rabbit_hole.behavior_settings.node_type_area_of_study.yml

joewhitsitt:
>Need to reset checkbox value to 0 when user removes all websites (otherwise can get 'Visitors are being redirected' message when no websites are linked)

I think this should be handled during the presave once https://github.com/uiowa/uiowa/pull/6053#issuecomment-1404334136 is resolved: https://github.com/uiowa/uiowa/pull/6053/files#diff-5664e4af7ce2e39c8ff0fd1279ab7e35fb7ae57ff31e9d710a52db2a6528ef12R397 meaning the redirect won't be set if source doesn't have values (yes the box will still be checked while hidden but won't have an effect)

joewhitsitt:
There seems to be some noticeable caching when changing the website on an active rabbit hole redirect (e.g. changing the first link from google.com to yahoo.com). I don't think it is a bit problem as the data is being saved correctly and it is just the token/redirect taking a bit to clear up. This will be easier to test once we know the varnish impact on the remotes.

joewhitsitt (approved):
@e-marie-w thanks for sticking with this. looks good.

Issue #6042, opened 2022-12-20
Probe making all articles use a narrow container


briand44:
From https://github.com/uiowa/uiowa/issues/2784

# Story
As a member of the web team, I would like to understand any negative effects on existing articles when switching to a narrow container so that we can determine if we can make this switch for all articles or if we need to add an option where folks can choose to turn this on.

# Proposed solution

joewhitsitt:
Knowing that this is similar to a page LB narrow section with a filtered_html text area, if I were to pick this up, I'd start with a kitchen sink dump of content (all the wysiwyg buttons, media types and their variations, allowed html tags and properties that don't have a wysiwyg button) in one of those to see how things squish. We should be preventing the use of style attributes (e.g. width, height) so there shouldn't be issues with fixed widths. If there is, we have text areas in four column sections in lb that will have to be addressed.

I think asking about it in demo is a good idea if we haven't previously done so.

bspeare:
I think if we do this, we should either use the 67/33 layout or follow Tippie's design lead and center the featured image text and page title and remove the breadcrumbs: https://tippie.uiowa.edu/faculty-research/research-expertise-interests/research-organizations/experimental-economics.

Other good examples of narrow article design with centered page titles:

- https://www.marylandhall.org/classes/studios/
- https://www.stonehill.edu/news/details/supporting-skyhawks-sacramentally/
- https://www.kenyon.edu/news/archive/leaving-home-is-never-easy/
- https://home.dartmouth.edu/news/2022/12/dartmouth-experts-weigh-threats-nh-primary

bspeare:
Here is a quick mockup of what the article could look like with a centered page title:
- https://sandbox.prod.drupal.uiowa.edu/test-narrow-article
- https://sandbox.prod.drupal.uiowa.edu/test-narrow-article-w-image
- https://sandbox.prod.drupal.uiowa.edu/test-narrow-article-w-gray-background

briand44:
@bspeare in the second example can you provide thoughts on moving the created date and organization to the bottom?

bspeare:
@briand44 that is just an artifact of using the replicate function.  It jumbles up the order of blocks.  It should be updated now.

briand44:
Thanks for putting those together!

cory-skeers:
[...]

Issue #2784, opened 2021-01-04
Create a narrow option for articles


pyrello:
# Story

As a website visitor I would like articles to have narrow columns so that it is easier to read and more in line with my expectations based on other sites.

This functionality was requested for the new Writing University website.

Example other sites: https://medium.com/mind-cafe/einsteins-formula-for-a-happy-life-b29aff61a9c7, https://www.nytimes.com/2021/01/04/technology/slack-down-outage.html

# Proposed solution
Create a new site-wide setting that applies to articles and periodicals to allow the site owner to display these in a narrow container.

cory-skeers:
Interested in this. Also curious to look into some research and best practices surrounding this, might inform whether this should be a universal update or open to options. I've read some on it before, but don't know what kinds of margins there are around "optimum" widths for readability, if there's a clear best practice or if going off expectations is adequate.

cory-skeers:
Sample resource from Smashing Magazine: https://www.smashingmagazine.com/2014/09/balancing-line-length-font-size-responsive-web-design/

**tl;dr version:** column width that accommodates around 45 to 85 characters (with the 65 characters as the optimum previously determined for print resources).

Grabbing random full-length lines from the examples:

https://medium.com/mind-cafe/einsteins-formula-for-a-happy-life-b29aff61a9c7: 77 characters,
https://www.nytimes.com/2021/01/04/technology/slack-down-outage.html: 65 characters,
https://sitenow.uiowa.edu/news/2020/12/changes-look-and-feel-person-page (unrestrained column size, so highly variable, but based on what my viewport happened to be upon loading): 140 characters

And the counterargument: http://web.archive.org/web/20170918212943/http://psychology.wichita.edu/surl/usabilitynews/72/LineLength.asp
Suggests that longer lines aren't as detrimental in digital media (they only tested up to 95 characters, though). Part of the reason the "optimum" above is bumped up to 85 rather than 75.

joewhitsitt:
Related: https://github.com/uiowa/uiowa/issues/1609

cory-skeers:
~Per backlog grooming discussion, looks like the plan is to go forward with adjusting the column width platform-wide for all articles using the current templates, prior to #1609 updates.~

cory-skeers:
Scoped to Writing University only for now, with further discussion later.

quamsta:
Might have made a duplicate issue here: https://github.com/uiowa/uiowa/issues/2924

[...]

PR #6064, opened 2022-12-30
Article Narrowification

cory-skeers:
Relates to #6042 

# How to test

Either sync a site with articles that include various combinations of inline media (especially some that are left- or right-aligned and large), or `git checkout main` first and create some sample articles
Checkout, cim, cr, and run the update hook
Visit various article examples, check for weirdness.
In articles that have large left- or right-aligned media, check that
1.  Those media, but not others, have been adjusted to center aligned
2. A new revision has been created with a custom revision message


pyrello:
Closing this based on https://github.com/uiowa/uiowa/issues/2784#issuecomment-1371357572.

PR #6096, opened 2023-01-13
Hotfix for views exposed form reset link

bspeare:
Test the reset button https://iowasummerwritingfestival.prod.drupal.uiowa.edu/events. 

`ddev blt frontend`
Test https://iowasummerwritingfestival.uiowa.ddev.site/events in a private window.  Reset button should be placed under search and not above.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/views/views.scss
+++ b/docroot/themes/custom/uids_base/scss/views/views.scss
@@ -8,6 +8,7 @@
   min-height: 20px;
   padding: 19px;
   margin-bottom: 2rem;
+  position: relative;
 
   [type=checkbox], [type=radio] {
     height: 1.2rem;


cory-skeers:
Tested, and works well!

On a note unrelated to the reset button: Should there be a "No results" text added to the view? I don't remember if there was a reason to specifically not include it or not.

cory-skeers (approved):
Approving for the specific change this addresses. Extra question in https://github.com/uiowa/uiowa/pull/6096#issuecomment-1382186469 that could be snuck in here if desired, but not necessary.

bspeare:
Thanks @cory-skeers! Do you want to sneak that in?

cory-skeers:
> Thanks @cory-skeers! Do you want to sneak that in?

Er, after looking at it more, it should probably be updated in the Event feature split for the default view. I don't think that would be an issue for the few sites that have it turned on, but I also don't want to hold this up for an unrelated change. I'll open a separate issue.

PR #6113, opened 2023-01-19
Update README.md

joewhitsitt:


joewhitsitt:
Ben recent encountered some drush errors similar to `PHP Fatal error: Uncaught TypeError: Drush\Preflight\PreflightSiteLocator` when working with newly provisioned websites. This was because the local settings files were not yet setup and couldn't bootstrap. Running `ddev blt bis` solves the issue but at a great cost. ALL multisites local.settings are deleted and recreated and it wipes out any custom overrides you have there. The way around it is to store those overrides in local.blt.yml within the site directory and they are written to the local.settings file while regenerated. This change occurred back with https://github.com/uiowa/uiowa/pull/4673 but the readme was not updated.

cory-skeers (approved):
Thanks for updating this @joewhitsitt!

Issue #5432, opened 2022-07-06
Consider dropping responsive image support?


pyrello:
# Motivation

Based on this article (https://medium.com/@julienetienne/responsive-images-in-2020-vision-10caa8e9b993), it seems like responsive images may not be as useful as they once were, especially given the complexity and overhead of implementing and maintaining their usage. Focusing instead on using a technology like WebP might be a better use of our bandwidth ;)

# Proposed solution

- [ ] Remove config related to responsive image styles.
- [ ] Uninstall responsive images module

pyrello:
This seems relevant: https://gorannikolovski.com/blog/drupal-92-will-support-webp-images-out-box

pyrello:
Lullabot has an article about webp where they seem to be indicating that they still believe in responsive images, although they don't go into much detail: https://www.lullabot.com/articles/speed-drupals-front-end-webp-images

They refer to their original 2017 article about responsive images: https://www.lullabot.com/articles/fundamentals-of-responsive-images

bspeare:
~I like the idea of trying out webp and moving away from responsive images~.   I did a comparison of https://uiowa.edu/sites/uiowa.edu/files/styles/ultrawide__2592_x_1111/public/2020-07/2019_08_19-Iowa%20EDGE%20Teambuilding%20jatorner%20-0645%20%281%29.jpg converted to webp and it went from `1mb` to `584kb` for our largest `2592_x_1111` crop.   As comparison, the `768 x 329` version of that same image is `189kb` : https://uiowa.edu//sites/uiowa.edu/files/styles/ultrawide__768_x_329/public/2020-07/2019_08_19-Iowa%20EDGE%20Teambuilding%20jatorner%20-0645%20%281%29.jpg.  Perhaps if we focused on some of the further optimizations listed on https://medium.com/@julienetienne/responsive-images-in-2020-vision-10caa8e9b993 we can make up the difference?

edit: we should keep our responsive image support and try some additional optimizations.

bspeare:
Actually, comparing some `jpeg` to `webp` from tippie.uiowa.edu, some of the `webp` files became larger.   

Our mobile page speed reports are terrible (https://pagespeed.web.dev/report?url=https%3A%2F%2Ftippie.uiowa.edu%2F), the top recommendation is to "Preload Largest Contentful Paint image" and it looks like https://www.drupal.org/project/responsive_image_preload might be useful for us to try.  

We might want to keep our responsive image setup and make these additional changes:

[...]

PR #6120, opened 2023-01-23
Add inline size classes to images in WYSIWYG

### LINKED ISSUE #3614: Image editing functionality in CKEditor fix
# Problem
When you edit an image in the wysiwyg editor you get a certain amount of options in the Display dropdown. It's not clear to the user what these options mean. When I pick small, I get a 100px wide image. When I pick Small - Square I would expect to get a 100px wide square image but instead I could (based on screen width) get a 2592x2592px image with a width="768" set on it.

The "Small - Square" and "Large - Square" images options end up being the same size.

I created some demo pages with examples of images.
https://studentlife.prod.drupal.uiowa.edu/image-test-small
https://studentlife.prod.drupal.uiowa.edu/image-test-medium
https://studentlife.prod.drupal.uiowa.edu/image-test-large

# Steps to reproduce
Screen Shot 2021-06-03 at 4 53 44 PM

# Proposed solution
Ideally there should be a way for the content editor to select a customize width/height for the image but I assume this would require a different plugin. In the mean time, I'd like to see the Square, Widescreen, and Ultrawide options be consistent. So, if Medium is 220px than Medium - Square should be 220px but cropped square.

### LINKED ISSUE #3725: Remote Video/Panopto "Original" with captions are not ideal sizing
# Problem

WYSIWYG and rendered markup is different for drupal-media depending on what view mode is selected the alignment and whether a caption is provided or not.

We have made the functionality usable in https://github.com/uiowa/uiowa/pull/3711

Original alignment with a caption is not well reflected in the WYSIWYG, and Original shrinks down when a caption is added.

# Steps to reproduce

- https://sandbox.prod.drupal.uiowa.edu/media-panopto/panopto-original-captions
- https://sandbox.prod.drupal.uiowa.edu/media-panopto/panopto-original-no-captions

# Workaround
- If a full-width video is needed, do not use the caption option.
- If a caption is needed, select a view mode other than original.

# Proposed solution
- Unless the markup becomes more targetable (Drupal 10 due to backwards compatibility?) there is little we can do here. For the WYSIWYG `<drupal-media>` needs more info about what type of media is within, for render the wrapping `<figure>` needs more info about what type of media and size is within.

### LINKED ISSUE #5631: White space gap between sections
# Problem
Removing bottom spacing on a multi-column section that has cards will show a white space gap before the next section begins. Only happens on mobile screen sizes. 

Screen Shot 2022-08-30 at 4 59 20 PM

example pages:
https://sandbox.prod.drupal.uiowa.edu/card-bg-color
https://admissions.uiowa.edu/first-year-visit-options

# Steps to reproduce

1. Create a new 2,3, or 4 column section with a background color.
2. Remove the bottom spacing from the section
3. Create a card to that section with the image position of left or right
4. Create a new section below with a background color and add any component

This might occur with components other than cards but haven't tested those.

### LINKED ISSUE #6074: Implement container query work for image styles and other inline elements in uids_base
# Story

As a SiteNow content editor, I would like image sizes and other inline elements (callout) to be defined based on the size of their container so that I can have more consistent styling of images in different column widths.

# Proposed Solution

- Update to a commit of UIDS 4 that includes updates in uiowa/uids#845.
- Figure out how to add a relevant sizing class based on the responsive image option that is selected.
- [x] How do we deal with `Small`, `Medium`, and `Large` image styles/view modes?
    - [x] Investigation
    - [x] Implementation
- [x] After completing the `<figure>` refactor [here](https://github.com/uiowa/uids/issues/844#issue-1519630880), rectify the changes in uiowa
- [x] Figure out how to refactor `align-ORIENTATION` to `inline--align-ORIENTATION`
- [x] Move inline sizing classes to the figure element. This may be done with something similar to `filter_caption_preprocess()`
- [x] [Contact tippie about pages like this](https://tippie.uiowa.edu/melissa-pepper-story-iowa-mba), perhaps see if Mara can convert these pages for tippie before the update?
- [x] Add new aspect ratio classes `media--16-9` and `media--21-9`.
    - [x] [Is this needing to be refactored?](https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_core/src/Plugin/Filter/FilterIframe.php#L86-L101)

<details>

- Example `@container (min-width: 768px) and (max-width: 855px) { .media--view-mode-large__* { width: 600px/65%}};`
- You can view an example of a container query like this at https://codepen.io/bspeare/pen/QWBKxzO (view in Safari or Chrome for it to work)
- Test this out in all the layout columns

</details>

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/5631
Resolves https://github.com/uiowa/uiowa/issues/3725

This PR adds the container query work to inline images. Additionally, future elements with the relevant classes will also query their containers for size adjustments.

This can be merged once the container query work is merged in uids4, but the hash will have to be changed.

## how to test.

`ddev blt frontend` to get UIDS updates. 
___
1. Sync sandbox
2. Check all the pages in the "Inline sizing" menu link and ensure they look reasonable.  
___
1. Make a new page. (You may want to do this on the prod site, so that others can use it later)
2. Make multiple columns in a section.
5. Make a text areas in the different columns with large, medium, and small sized images, videos, etc added in with different alignments, and with and without captions.
6. Inspect different variations to see if they are adjusting themselves based on their container both in the layout builder edit page, and on the view page.  
___
1. Do a thorough code review of both this, and the "[Callout component](https://github.com/uiowa/uids/pull/845)" PR.
2. Look for things you think may break, and test that they don't. Report back with results.  
___
1. I specifically changed the Panellumn PAnorama for housing. ensure that it is not broken.
___


docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -134,6 +134,20 @@ function template_preprocess_uiowa_core_headline(&$variables) {
 
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function uiowa_core_preprocess_media(&$variables) {
+  $view_mode = $variables['view_mode'];
+  $accepted_view_modes = ['large', 'medium', 'small'];
+  foreach ($accepted_view_modes as &$value) {
```
pyrello:
```suggestion
  foreach ($accepted_view_modes as $value) {
```
Since `$value` is not being modified in the loop, there is no need to pass it by reference.

GaryRidgway: Ah, thank you

pyrello:
@GaryRidgway I'm noticing that this is failing the percy check. It looks like the job title on the person page is changed for some reason. Can you look into why this is happening?

GaryRidgway:
> @GaryRidgway I'm noticing that this is failing the percy check. It looks like the job title on the person page is changed for some reason. Can you look into why this is happening?

I'm on it

pyrello:
I just realized the logic that is in place for adding inline size classes to media does not have an appropriate constraint to ensure that it is only runs for media in WYSIWYG text areas. Setting back to draft.

GaryRidgway:
As I am going through this, I am having a hard time knowing what exactly should be scoped. To update old images, we would need to preprocess the media, but once we are on the viewing side, there is no way to know they came from a WYSIWYG. Its just text area, or description, or body, etc.

I think performing the  preprocess on all media is still fine, the more that I think about it. its not going to affect anything that is not defined by our set of view modes.

Any thoughts on this?

pyrello:
> I think performing the preprocess on all media is still fine, the more that I think about it. its not going to affect anything that is not defined by our set of view modes.

This will add an unnecessary class to media in a wide variety of contexts, such as cards and banners. While we can probably mitigate for this via CSS, it is really a problem that we should solve via code.

When you say there is no way of knowing where it came from, does the `$variables` object in the preprocess function not contain any reference to the entity that the media field is being embedded in?

GaryRidgway:
Looking in to some issues right now with percy, and I am seeing that `container-type: inline-size` is changing the way margins are measured within containers. 

Margins can currently overlap each other, but when we set the container with that rule, they start stacking. How do we feel about this?

I can get an example tomorrow

bspeare:
@GaryRidgway it looks like the issue that you brought up is fixing this issue: https://github.com/uiowa/uiowa/issues/5631.  The extra card spacing that you are seeing is from the cards, which have a bottom margin set on them.

GaryRidgway:
> @GaryRidgway it looks like the issue that you brought up is fixing this issue: #5631. The extra card spacing that you are seeing is from the cards, which have a bottom margin set on them.

Should we disregard for now then, or also fix that bug?

GaryRidgway:
Additionally, I think this inline sizing issue could use some love as well. I am observing issues for images that have a small pixel width and height. There is some float weirdness, and additionally some width/height weirdness.

<details>
<summary>
Float example
</summary>


</details>

<details>
<summary>
Width / Height example
</summary>


</details>

GaryRidgway:
This might be fixed by changing out container queries to be `max-width: NUMBERpx` instead of `width: NUMBERpx`

I am going to experiment with that now

bspeare:
> Should we disregard for now then, or also fix that bug?

Yes, disregard.  We can say this work resolves that bug.

bspeare:
Here's an example that could be an issue.  Compare https://tippie.uiowa.ddev.site/melissa-pepper-story-iowa-mba with https://tippie.uiowa.edu/melissa-pepper-story-iowa-mba.

bspeare:
I'm noticing the classes are not being applied to the "Full" image style, is that intentional?

bspeare:
I think https://github.com/uiowa/uiowa/pull/6120#issuecomment-1405375127 has been resolved by https://github.com/uiowa/uiowa/pull/6120/commits/b47b81a9eef170bee34ef72300c7a99c1836a57a, but we will want to test that again.

pyrello:
> I'm noticing the classes are not being applied to the "Full" image style, is that intentional?

There is no class to apply based on the work done in UIDS.

bspeare:
Leaving out the core Drupal view modes is also not ideal.  This is the "Large" image. 

Screen Shot 2023-01-30 at 11 20 01 AM

pyrello:
> Leaving out the core Drupal view modes is also not ideal. This is the "Large" image.

Maybe we need to update the definitions of the legacy SNV2 media view modes? Our intention is to remove them eventually, so limiting the amount of work we do to account for them would be great!

bspeare:
Another one using `Small - no crop`: https://pharmacy.uiowa.edu/news/2022/04/ernst-named-diversity-equity-and-inclusion-director. 

Screen Shot 2023-01-30 at 11 29 58 AM

GaryRidgway:
> > Leaving out the core Drupal view modes is also not ideal. This is the "Large" image.
> 
> Maybe we need to update the definitions of the legacy SNV2 media view modes? Our intention is to remove them eventually, so limiting the amount of work we do to account for them would be great!

Is that just creating an update hook to change all the SNV2 ones to a similar SNV3 view mode?

pyrello:
> Is that just creating an update hook to change all the SNV2 ones to a similar SNV3 view mode?

It would require figuring out what we would switch them to first. We haven't done it yet because it is not a straightforward problem.

bspeare:
For issues like https://github.com/uiowa/uiowa/pull/6120#issuecomment-1407095092 - I think that we can advise switching to the "Full" image style to achieve adjacent width/height images.  That page is using the "Medium - Square" crop and would be fixed by adjusting to the "Full - Square" image style.

GaryRidgway:
Reviewing again, I was mistaken about the WYSIWYG class issue in my last commit comment

GaryRidgway:
After making test pages, I am going to start making changes to preprocess functions to try and catch our relevant cases that are not currently being preprocess correctly for alignment and sizing.

After that, I will start on refactoring remote video issues.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -138,33 +139,88 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+
+function uiowa_core_inline_viewmode_map($view_mode): array {
+  $accepted_view_modes = ['large', 'medium', 'small'];
+
+  foreach ($accepted_view_modes as $value) {
+    if (str_starts_with($view_mode, $value)) {
+
+      return [
+        "library" => 'uids_base/inline',
+        "class" => 'inline--size-' . $value,
```
pyrello:
```suggestion
        'library' => 'uids_base/inline',
        'class' => 'inline--size-' . $value,
```

GaryRidgway: FIXED

cory-skeers:
I'm pulling this down to take a look, but as there's a lot here, anyone else feel free to pull down and review/test in tandem.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -134,6 +135,131 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+  // Create a new array to be used later for the associative media attributes data.
+  $associative_drupal_media_attributes_strings = [];
+
+  // For each of the media attributes strings...
+  foreach ($drupal_media_attributes_strings as $value) {
+
+    // Separate them in to two parts, clean them up, and add them to the associative array.
+    $attribute__value = explode('=', $value);
+    $associative_drupal_media_attributes_strings[$attribute__value[0]] = str_replace('"', '', $attribute__value[1]);
+  }
```
cory-skeers: At first I was thinking we could use an HTML parser to pull out the attributes for this, but I think that's over the top, needing to convert things back and forth.

But I was wondering if we could do something like
```
  // Retreive the drupal media attributes.
  preg_match_all('%\s?([^"]*)="([^"]*)"%is', $drupal_media_attributes_string, $matches);
  $associative_drupal_media_attributes_strings = [];
  for ($i = 0; $i < count($matches[0]); $i++) {
    $associative_drupal_media_attributes_strings[$matches[1][$i]] = $matches[2][$i];
  }
  ```
  The preg_match there grabs pairs of not-in-quotes and in-quotes values (and drops the leading space if it's there), and then a similar pairing up to make the associative array.

I'd guess the preg_match_all would be technically faster since it can do the search and cleanup in one go, but am open to preferences because 1) the strings that'll be parsed here shouldn't be long enough for processing to be that intense and 2) I'll admit that regexing everything doesn't always leave the most readable code for future maintenance.

File: docroot/modules/custom/uiowa_core/uiowa_core.module [hunk 4]
```diff
@@ -134,6 +135,131 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+  foreach ($accepted_view_modes as $value) {
+    if (str_starts_with($view_mode, $value)) {
+
+      // We return a string reference to the inline library and a modified view mode class.
+      return [
+        'library' => 'uids_base/inline',
+        'class' => 'inline--size-' . $value,
+      ];
+    }
+  }
```
cory-skeers: This (`^`) checks for something at the start of the search string, which we can check for each of the values in one go and if found, passes it on for the return.

```suggestion
  if (preg_match('%^(' . implode('|', $accepted_view_modes) . ')%i', $view_mode, $value)) {
    return [
      'library' => 'uids_base/inline',
      'class' => 'inline--size-' . $value[0]
    ];
  }
```

docroot/modules/custom/uiowa_core/uiowa_core.module
cory-skeers:
```suggestion
  // Get the '<drupal-media>' tag string.
  $drupal_media_markup = strval($variables['node']);
  
  // Define some $replacement_strings to remove.
  $replacement_strings = ['<drupal-media ', '></drupal-media>'];

  // Remove the $replacement_strings.
  $drupal_media_attributes_string = str_replace($replacement_strings, '', $drupal_media_markup);

  // Retreive the drupal media attributes and separate
  // them into an array of attribute strings.
  // Ex: ['foo="bar"', 'ban="baz"'].
  preg_match_all('%\s?([^"]*)="([^"]*)%is', $drupal_media_markup, $matches);
  $associative_drupal_media_attributes_strings = [];
  for ($i = 0; $i < count($matches[0]); $i++) {
    $associative_drupal_media_attributes_strings[$matches[1][$i]] = $matches[2][$i];
  }
```

pyrello: What does `$variables['node']` contain here? I would usually expect this to be the node object (although it would usually be listed as `$variables['#node'] in that case).

pyrello: While this is well-commented in terms of explaining what is happening from a code perspective, it does not provide any context as to what is happening generally or why. Despite the comments, if I were reading this, I would have no idea why this piece of code existed if I was later evaluating whether it was still needed (for example, when we upgrade to CKEditor 5 and have to revamp all our WYSIWYG code).

cory-skeers:
I'm not seeing the center align work for anything that doesn't have a caption (it does work for those with captions). Left/right align are working as expected. Not sure where the issue is coming in there yet, though.

docroot/modules/custom/uiowa_core/uiowa_core.module
cory-skeers: > What does $variables['node'] contain here? I would usually expect this to be the node object (although it would usually be listed as `$variables['#node'] in that case).

It has the media entity, like `<drupal-media data-entity-type="media" data-entity-uuid="ffc45612-f7e6-46c3-b9e2-17ce7861be49" data-view-mode="medium__no_crop"></drupal-media>`. I'm not sure if there are times when it has other things--so far it's been the only thing I've seen come through it.

pyrello:
I am pulling this to test a few things I'd like to understand better.

pyrello:
@cory-skeers I was renaming variables and tried to implement your regex changes. Can you take a look and see if I managed to get it right?

GaryRidgway:
Going to work on the `inline--size-full` and the `inline--align-center` issues after lunch.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -134,6 +135,141 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+  // Detect if this is a figure and if it is,
+  // it has already been processed for inline sizing and alignment.
+  $is_figure = FALSE;
+
+  if (isset($variables['attributes']['data-figure'])) {
+    $is_figure = filter_var($variables['attributes']['data-figure'], FILTER_VALIDATE_BOOLEAN);
+  }
+
+  // Check to see if the media element has been processed, or if it is an embed.
+  if (!$is_figure && isset($variables['elements']['#embed']) && $variables['elements']['#embed']) {
```
cory-skeers: Do we think there'll ever be a situation where we have a figure that was processed, but might still need further processing in this?

If not, can we shorten this to just exit if it's a figure, and if we hit the check further down, we already know it's not a figure.
```suggestion
/**
 * Implements hook_preprocess_media().
 */
function uiowa_core_preprocess_media(&$variables) {

  // Detect if this is a figure and if it is,
  // it has already been processed for inline sizing and alignment.
  if (isset($variables['attributes']['data-figure']) &&
    filter_var($variables['attributes']['data-figure'], FILTER_VALIDATE_BOOLEAN) === FALSE) {
    return;
  }

  // Check to see if the media element has been processed, or if it is an embed.
  if (isset($variables['elements']['#embed']) && $variables['elements']['#embed']) {
```

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -1950,3 +1951,116 @@ function sitenow_update_9054() {
...
+          // Loop through the fields we defined before.
+          foreach ($field_check as $name => $definition) {
+            // If the field has a value (or values).
+            if (!$entity->get($name)->isEmpty()) {
+              $updated_values = [];
+              $field_values = $entity->{$name}->getValue();
+
+              foreach ($field_values as $field_value) {
+
+                // @todo Any special handling for text_with_summary fields?
```
cory-skeers: There shouldn't need any special handling for `text_with_summary`. The regular value is still in `['value']`, and if there does happen to be something in the `['summary']` (copy/pasting?), it won't be rendered, and there's a good chance it'd be truncated and broken anyway.

I think the main chance there'd be for something ending up in there would have been migrations, but we have special handling for those cases that _should_ have prevented that.
```suggestion
```

cory-skeers: On the flip side, it'd be possible to check the string format before processing, and only proceed with those that have a format that allows for embeds.

But those cases would just end up with an `$i` of `-1`, and processing would get skipped anyway, so 🤷

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -1950,3 +1951,116 @@ function sitenow_update_9054() {
...
+                  $old_dm_instance = $dm_instances->item($i);
+                  $new_dm_instance = $old_dm_instance;
+
+                  $view_mode = $old_dm_instance->getAttribute('data-view-mode');
+                  $is_relevant_view_mode = in_array($view_mode, $media_view_mode_map_keys);
+                  if ($is_relevant_view_mode) {
+                    $new_dm_instance->setAttribute('data-view-mode', $media_view_mode_map[$view_mode]);
+
+                    $old_dm_instance->parentNode->replaceChild($new_dm_instance, $old_dm_instance);
+                  }
```
cory-skeers: Should the `$needs_save = TRUE;` be moved up here? I think this is the point where we know a change is needed, and if we don't hit this, it's possible we had an entity with the right fields, values in those fields, but no embeds to be updated.

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -1950,3 +1951,116 @@ function sitenow_update_9054() {
...
+                $updated_values = $updated_values[0];
+              }
+
+              $entity->{$name} = $updated_values;
+              $needs_save = TRUE;
+            }
+          }
+          if ($needs_save) {
+            $entity->save();
+          }
```
cory-skeers: Do we want to set a revision/revision message for this?

On a similar note, do we want/need to account for revisions in this update, in case someone reverts to a previous revision after the update hook runs, and ends up with the old view mode?
```suggestion
          if ($needs_save) {
            $entity->setNewRevision(TRUE);
            $entity->revision_log = 'Updated entity embed view modes.';
            $entity->save();
          }
```

bspeare:
Is there a trick to regenerate the image styles locally? For instance like on this page https://sandbox.uiowa.ddev.site/inline-image-caption-alignment?

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -134,6 +135,155 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+  foreach ($accepted_view_modes as $value) {
+    if (str_starts_with($view_mode, $value)) {
+
+      // Return the class that needs to be added.
+      return "inline--size-$value";
+    }
+  }
+
+  return NULL;
+}
```
cory-skeers:
```suggestion
function uiowa_core_get_inline_size_class_from_view_mode(string $view_mode): ?string {
  // The view mode sizes which have a matching class. Defining them here
  // will define what we use in the regex matching pattern.
  $accepted_view_modes = ['default', 'large', 'medium', 'small'];

  // Look for any of our accepted view modes by checking the
  // start of the view mode string, and return the matching
  // inline size class string.
  // Regex101 explanation: https://regex101.com/r/CHe1tD/1
  if (preg_match('%^(' . implode('|', $accepted_view_modes) . ')%i', $view_mode, $value)) {
  
    // Return the class that needs to be added.
    return 'inline--size-' . $value[0];
  }

  return NULL;
}
```

pyrello: @cory-skeers This suggestion is returning an `?array` instead of `?string`.

cory-skeers: Oh shoot. I must have missed a change in there somewhere.

cory-skeers: Updated to only return the string.

docroot/profiles/custom/sitenow/sitenow.install
pyrello: Looking at the update hook in a little closer detail, I'm wondering if this should be re-written to access the database tables directly to pull the field value and update it. My impression is that the entity load is going to be a much heavier process. We might need to consider batching this anyway for sites that have a very large number of nodes/blocks, but using the direct database method might make that a moot point. That would also allow us to not have to load an entity to discover whether it contains a value that we need to work on, as that could just be reflected in the query. Finally, we could use logic like https://git.drupalcode.org/project/drupal/-/blob/c540e9ae4917193f75bdc539ce66858fa3f1e49e/core/modules/layout_builder/layout_builder.install#L106-113 to programmatically determine whether revisions need to be considered. 

And to answer that question, yes, we need to account for revisions. The alternative is to remove old revisions, which is not desirable. We essentially need to set the update up to make it as though the old view modes never existed.

cory-skeers: > Finally, we could use logic like https://git.drupalcode.org/project/drupal/-/blob/c540e9ae4917193f75bdc539ce66858fa3f1e49e/core/modules/layout_builder/layout_builder.install#L106-113 to programmatically determine whether revisions need to be considered.

Also just remembered the tempstore will have to be considered (which is part of update hook linked above, in the `key_value_expire` section).

GaryRidgway:
@cory-skeers The centering issues should now be solved!

cory-skeers:
> @cory-skeers The centering issues should now be solved!

Awesome! I'm doing a bit of digging/thinking about the update hook, but will re-check that as part of it.

docroot/modules/custom/uiowa_core/uiowa_core.module
cory-skeers: Suggestion no longer applies. Could still switch to a regex approach, but would still need to be a bit more detailed. Resolving this conversation--might revisit later.

cory-skeers:
I took a stab at refactoring the update hook to 1) go directly to the database for the updates to skip entity loads/saves and 2) as part of that, handle both the regular and revision tables. I _think_ it's working correctly, and the actual value updating is handled using the same logic as before, but it could use review and testing.

I'm planning on setting up a local example for the tempstore and working on that part this afternoon, unless anyone gets a wild hair to tackle it in the meantime.

cory-skeers:
> @cory-skeers The centering issues should now be solved!

Checked on this, and it is working! 💯

cory-skeers:
I set up a small test page with some tempstore data in it including embedded images to test that part of the update hook: https://sandbox.prod.drupal.uiowa.edu/inline-sizing-tempstore-test
After syncing/update hook, going into the layout editor of that page should have a few components with (updated) images.

I'm getting a serialize error, so something's not quite right yet, but looking into it.

cory-skeers:
I think the update hook is fully working now, but definitely needs a good lookover and testing.

cory-skeers:
Removed my assignment for clarity--was going to step away from this for a bit. I'd focused on reviewing code and testing the update hook based on the example pages on sandbox. I'd also done some experimenting on my own, but the multiple columns in a section has been the least tested by myself, and was going to be the next stop on my path through this if that's helpful for other potential testers.

joewhitsitt:
if we are getting rid of these uids_base classes completely, we _do_ need to fix the docroot/modules/custom/uiowa_core/src/Plugin/Filter/FilterIframe.php test which is failing.

joewhitsitt:
Is there an easy way to look through config as well? https://sandbox.uiowa.ddev.site/news (SiteNow Articles/People views header filtered_html text area). Did we touch these when we switched to media library?

joewhitsitt:
Screenshot 2023-02-27 at 3 18 58 PM

remote video original aligned (no caption) not getting `inline--size-full` similar to the caption versions so it is still aligned but with no padding.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -134,6 +135,162 @@ function template_preprocess_uiowa_core_headline(&$variables) {
...
+  foreach ($accepted_view_modes as $value) {
+    if (str_starts_with($view_mode, $value)) {
+
+      // Return the class that needs to be added.
+      return "inline--size-$value";
+    }
+  }
+
+  // If we have a `default` view mode, set full.
+  if (str_starts_with('default', $value)) {
```
joewhitsitt:
```suggestion
  if ($view_mode === 'default') {
```
As it relates to https://github.com/uiowa/uiowa/pull/6120#issuecomment-1447144862

joewhitsitt:
I think this resolves https://github.com/uiowa/uiowa/issues/3725 by making both original and original with captions full width so they are consistent and not super small in the case of original with captions.

https://sandbox.uiowa.ddev.site/media-panopto/panopto-original-captions

joewhitsitt:
Ugh. accidentally deleted my comment.

I think this would resolve https://github.com/uiowa/uiowa/issues/3725 so that original and original with captions are consistent and full-width. See https://sandbox.uiowa.ddev.site/media-panopto/panopto-original-captions

joewhitsitt:
>I specifically changed the Panellumn PAnorama for housing. ensure that it is not broken.

Ope. It broken.

**Local**
Screenshot 2023-02-27 at 4 03 55 PM

**Prod**
https://housing.uiowa.edu/residence-halls/catlett#double
Screenshot 2023-02-27 at 4 05 04 PM

joewhitsitt:
>I specifically changed the Panellumn PAnorama for housing. ensure that it is not broken.

It isn't broken exactly, but it is no longer the width of the accordion which makes viewing the pano even less desirable.

**Local**
Screenshot 2023-02-27 at 4 14 39 PM

**Prod**
https://housing.uiowa.edu/residence-halls/catlett#double
Screenshot 2023-02-27 at 4 14 19 PM

joewhitsitt:
I think all this logic in uiowa_core.module this is fine, but I am wondering if we should have created a WYSIWYG filter similar to FilterIframe where it is also setting classes based on element details and attaching a library.

Update: Oh wait. Is this because this fires and is needed outside of WYSIWYG? If so, disregard.

docroot/sites/housing.uiowa.edu/modules/housing_core/templates/field--paragraph--field-housing-floor-plan-pano--housing-floor-plan.html.twig
```diff
@@ -39,7 +39,8 @@
 #}
 
 {% for item in items %}
-  <div class="embed-responsive embed-responsive-16by9">
+  {{ attach_library('uids_base/media') }}
+  <div class="media--16-9">
```
joewhitsitt:
```suggestion
  <div class="media media--16-9">
```
Seems to resolve https://github.com/uiowa/uiowa/pull/6120#issuecomment-1447197349

joewhitsitt:
Is the communication plan a slice to this? If so, has a draft been started?

docroot/modules/custom/uiowa_core/uiowa_core.module
joewhitsitt: I agree with this change

joewhitsitt:
In investigating https://github.com/uiowa/uiowa/issues/6213, I noticed that simply switching to a larger viewmode in Tippie's situation won't work because of default inline margin top and bottom regardless of alignment. Switching the text area to an image block and removing the default top margin allowed the image to fill the vertical space.

Screenshot 2023-03-01 at 10 24 26 AM

Attn: @makurj 

Update: I've updated the communication message to suggest this as an option.

joewhitsitt:
>Ship it. - https://github.com/uiowa/uiowa/issues/6074#issuecomment-1450439978

Waiting for a date, but in the meantime can we do the following or is it too early?

"This can be merged once the container query work is merged in uids4, but the hash will have to be changed."

joewhitsitt:
>Waiting for a date

Communication sent yesterday. Planning a scheduled deployment for 3/9

pyrello:
Converted to draft to prevent merging prior to site launches today. Can be marked as ready again once those are complete.

bspeare:
I think we need to hold off on merging this.  I just noticed that we are swapping UIDS3 Media for UIDS4 which is causing this regression.  I think I added a fix for this in one of the card prs that I will try to add here. 

Screenshot 2023-03-08 at 12 32 35 PM

bspeare:
https://home.uiowa.ddev.site/ grid is also messed up.  I'm going to spend some time this afternoon testing this and seeing if I can fix.

bspeare:
Screenshot 2023-03-08 at 12 48 30 PM

The images within the grid are getting default margins now.  Do we just want to override the new inline sizing margins on images within this grid class?

bspeare:
Noticing a lot of image regressions on brand.uiowa.edu.  Do we want to give them a heads up to change things before we push this out?

bspeare:
For example, here's a before and after:

Screenshot 2023-03-08 at 1 24 35 PM
Screenshot 2023-03-08 at 1 24 30 PM

joewhitsitt:
wait so a 180px original is stretching full-width beyond it's size? that seems like a bug

sorry i missed all this...

joewhitsitt:
>The images within the grid are getting default margins now. Do we just want to override the new inline sizing margins on images within this grid class?

Can those be converted to image blocks with the margin removed?

bspeare:
> > The images within the grid are getting default margins now. Do we just want to override the new inline sizing margins on images within this grid class?
> 
> Can those be converted to image blocks with the margin removed?

That would probably need to be update hook territory with the amount of grids out there.

joewhitsitt:
We sent out the email detailing that changing to an image block might be necessary

bspeare:
> wait so a 180px original is stretching full-width beyond it's size? that seems like a bug
> 
> sorry i missed all this...

The css changed from:
```
.field--name-field-media-image img {
    max-width: 100%;
    height: auto;
}
```
To 
```
.field--name-field-media-image img {
    width: 100%;
    height: auto;
}
```

bspeare:
> We sent out the email detailing that changing to an image block might be necessary

I'd rather override those images with CSS than have to update and find every page using those on our top level sites.  Uiowa.edu, Education, Admissions has generous use of those sprinkled through their page designs.

pyrello:
> We sent out the email detailing that changing to an image block might be necessary

Image block is only available on V3, so would not be available to brand, correct?

joewhitsitt:
I am fine with the CSS override for image grids, but the image width for original seems like a bigger issue because we've told folks if they want a specific size to upload it as such and select original. .gifs for example.

Correct, the image block solution is not available to brand but is not why i referenced that solution (layout builder grid thingy)

joewhitsitt:
@GaryRidgway was this crucial for this PR or is there a tweak around it?

https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460740254

https://github.com/uiowa/uiowa/commit/71376d2960cae92b9d7ef3ff65eeb535974bc2b3

bspeare:
Seeing lots of before/afters like this screenshot from https://writinguniversity.org/.  I understand that we have put some guidance out there but really what percentage of site users will really take action on it or have the time to take action?  

Shouldn't it be up to us to run an update hook to swap these out so that the final result is barely noticeable?

Screenshot 2023-03-08 at 1 50 12 PM

Screenshot 2023-03-08 at 1 50 16 PM


Screenshot 2023-03-08 at 1 55 07 PM


Screenshot 2023-03-08 at 1 55 14 PM

makurj:
> Shouldn't it be up to us to run an update hook to swap these out so that the final result is barely noticeable?

I'd agree.

GaryRidgway:
@joewhitsitt 
The addition of the `width 100%` was for https://github.com/uiowa/uiowa/pull/6120#issuecomment-1447177999. We were assuming any original to be full width. That was what I remember us deciding.

joewhitsitt:
>Shouldn't it be up to us to run an update hook to swap these out so that the final result is barely noticeable?

I didn't think there was a way to pinpoint all of the potential instances. For example this is a small image that was the right size to fill the space. In other cases, small was intentionally selected to **not** fill the container. If we determine that we should swing the opposite direction apologize for the small instances that should have stayed small then I can be on board with that.

bspeare:
Based on what I'm seeing on [writinguniversity.org](https://org-writinguniversity.uiowa.ddev.site/) there are **_many_** images to update there.  I imagine there are quite a few other sites that will have the same issues.

joewhitsitt:
@GaryRidgway that was related to remote video. not images

GaryRidgway:
> > wait so a 180px original is stretching full-width beyond it's size? that seems like a bug
> > sorry i missed all this...
> 
> The css changed from:
> 
> ```
> .field--name-field-media-image img {
>     max-width: 100%;
>     height: auto;
> }
> ```
> 
> To
> 
> ```
> .field--name-field-media-image img {
>     width: 100%;
>     height: auto;
> }
> ```

can you give me a commit hash for this? I'm having trouble finding it, if you have it off the top of your head that is.

bspeare:
I guess this is beyond the image formats, this has to do with the percentage widths we are adding to the images. 

Should a layout that was designed like this:

Screenshot 2023-03-08 at 2 02 44 PM

Really need to switch to this?

Screenshot 2023-03-08 at 2 02 54 PM

bspeare:
Layout builder preview on https://org-writinguniversity.uiowa.ddev.site/node/106/:

Screenshot 2023-03-08 at 2 06 17 PM

Published page:

Screenshot 2023-03-08 at 2 06 03 PM

GaryRidgway:
> writinguniversity.org

When they use WYSIWYGS for each one, probably? That is what we want the WYSIWYG to do, yeah?

joewhitsitt:
yeah to be fair, they are using a lot of components in a way we didn't intend

bspeare:
A lot of sites try to make cards out of text areas.  I think that's a reasonable use of the component.

GaryRidgway:
> A lot of sites try to make cards out of text areas. I think that's a reasonable use of the component.

We could change our lower bound for small inline sizing so that it is more in line with a 3 card format width?

bspeare:
I'm going to use this comment for posting before/after images in a details. 

<details>
From https://pharmacy.uiowa.ddev.site/about/diversity_equity_inclusion. 
Screenshot 2023-03-08 at 2 19 53 PM
Screenshot 2023-03-08 at 2 19 59 PM

From https://labbyrne.uiowa.ddev.site/

Screenshot 2023-03-08 at 2 30 30 PM
Screenshot 2023-03-08 at 2 30 35 PM

Screenshot 2023-03-08 at 3 38 27 PM
Screenshot 2023-03-08 at 3 38 33 PM



Screenshot 2023-03-08 at 3 39 43 PM
Screenshot 2023-03-08 at 3 39 57 PM

Screenshot 2023-03-08 at 3 44 58 PM
Screenshot 2023-03-08 at 3 45 07 PM

Screenshot 2023-03-08 at 3 48 05 PM
Screenshot 2023-03-08 at 3 48 12 PM

Screenshot 2023-03-08 at 3 53 32 PM
Screenshot 2023-03-08 at 3 53 26 PM


Screenshot 2023-03-08 at 3 56 31 PM
Screenshot 2023-03-08 at 3 56 22 PM

Screenshot 2023-03-08 at 4 10 20 PM
Screenshot 2023-03-08 at 4 10 28 PM

Screenshot 2023-03-08 at 4 25 22 PM
Screenshot 2023-03-08 at 4 25 10 PM

Screenshot 2023-03-08 at 4 29 49 PM
Screenshot 2023-03-08 at 4 29 56 PM

Screenshot 2023-03-08 at 4 34 03 PM
Screenshot 2023-03-08 at 4 34 14 PM

Screenshot 2023-03-08 at 4 35 42 PM
Screenshot 2023-03-08 at 4 35 52 PM

</details>

GaryRidgway:
<details>
<summary>This is what that 100% width solves If I remember correctly?</summary>

<i>Disclaimer: this GIF was made almost two weeks ago</i>
</details>

EDIT: [This is in conjunction with this comment.](https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460865152)

GaryRidgway:
This could also be the case because we are currently assuming that the default view mode is full. 

<details>
<summary>You will notice that this is happening for at least a few of the examples that ben is showing:</summary>


</details>

joewhitsitt:
Trying to capture the uncovered issues into bullet points. Let me know if I missed something or if there is another solution:

- Image grids in v3 affected -  should override with CSS or identify and help transition to image blocks.
- Original _images_ are scaling past their original dimensions - look to exclude these images from container queries? Keep remote video original as full width.
- Selected responsive small image sizes are no longer filling the column as originally intended - look to identify these (e.g., three column grid text area with responsive small image) and update them to a larger view mode assuming editors intent or blanket update all responsive small image sizes to a larger responsive image style and communicate that some folks will want to size down after the update.

bspeare:
I'm going to try to summarize a list of things that need to be fixed as well.  I won't have time to do this until later today.

bspeare:
## Here's a list of issues from a few hours of testing yesterday

- Solve and find more random issues like https://engineering.uiowa.edu/college/office-dean - see https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460814901 for links and screenshots. 
- brand.uiowa.edu site owners have built out their pages with pixel specific images and they should be consulted on how their site would change.  
- Layout builder preview is not showing - should it?
- WYSIWYG preview not showing - should it?
- Address issues noted on https://org-writinguniversity.uiowa.ddev.site/, see the update hook suggestion in the fourth slice as a possible path forward on this.  
- Embedded iframes need styling - see https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460814901 for links and screenshots. 
- Horizontal menus z-index broken - see https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460814901 for links and screenshots. 
- Remove width: 100% for images to avoid major stretching - see https://github.com/uiowa/uiowa/pull/6120#issuecomment-1460814901 for links and screenshots.   
- Two YouTube videos set to “original” are getting different inline sizing on https://dentistry.uiowa.ddev.site/education - why?
- All images now have margins with the inline style classes.  Plenty of pages were designed to display images without margins. 
 - Image grids in v3 affected - should override with CSS or identify and help transition to image blocks?

Seeing all of these issues, I think we should slow down on this work and slice it up and reconsider some of our options. 

# Possible slices

1. **First slice: deliver callout code from UIDS.**
2. **Second slice: get rid of fixed width and height on responsive images.**
3. **Third slice - get rid of legacy S, M, L image sizes.**

### Outdated slices
<details>
**Second or third slice:** 
- Investigate reconsidering our responsive image solution (https://github.com/uiowa/uiowa/issues/5432) 
- Or reconsider the breakpoint groups image sizing. 
  - For instance, does “small – square” really need a 2592 x 2592 version?
  - "Small - square" renders as 768 x 768 on a full page content area.   Is 768 x 768 really the expected size when someone selects “small – square”?  I’d say that I would expect probably half of that. 
    - If you look at how this image renders in the inline sizing pr, it does shrink down to be around have that size.  
    - We could investigate making this change within our responsive image settings and avoid the inline sizing classes.  


**Fourth slice (if necessary): inline sizing w/container queries**
  - Investigate update hook to map images to appropriate styles. 
    - For example, If an image is set to “small – square” and it is within a 3 or 4 column layout, then the image should be update-hooked into “large – square” which would fix an issue like we’ve seen on https://org-writinguniversity.uiowa.ddev.site/. 
    - Find other specific update hook examples like the one suggested above.  


After thinking through the slicing of this work I think I would rather invest in the slices 1-3 and wait for w3c guidance on responsive images and container queries that is being actively discussed as noted on https://www.stefanjudis.com/notes/should-responsive-images-work-with-container-queries/.   Since this is an active discussion and problem that they are trying to solve, I don't think we should roll our own solution.  

</details>

cory-skeers:
> * Update these to “original” instead of an image style?
> * Updating M to small-no crop does not seem like a good match.
>   
>   * Look into mapping these to smaller responsive image styles that actually render as “small” and map update hook to this image style?

Originally the update hook did a straightforward small to small, medium to medium (or something like that). Do we want to add in additional logic to check the original image size as well? Like, if the image size in the file database is something like 300px, set it to original, if it's 800px, setting to small might be appropriate?

Still wrapping my head around the changes in the styles.

bspeare:
A few more thoughts.  :) 

Keep responsive images for the image block only and for other custom blocks like banners and sliders. 

For the WYSIWYG, remove responsive images adjust image styles to correspond to the described size, for example:
- Small - Square = 384 x 384
- Medium - Square = 768 x 768 
- Large Square = 1024 x 1024

I think that would be a great user improvement so that users are seeing the actual size that they are using and avoid the confusion described in https://github.com/uiowa/uiowa/issues/3614.   We could also switch to webp for these image styles and get great performance savings.

bspeare:
I did a little more digging into this and realized why all of our responsive images aren't working as expected.  A while back, we implemented fixed width and heights on our images to help with cumulative layout shift: https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/.   However, as noted in that article it is not ready for use in `<srcset>`, and because of this if you look at a page like https://sandbox.prod.drupal.uiowa.edu/image-sizes you will see that they are all sized the same.   But, if you look at this https://codepen.io/bspeare/pen/ExeQzRx example that removes the width and height attributes you can see the images behaving as expected.

pyrello:
> A while back, we implemented fixed width and heights on our images to help with cumulative layout shift

Specifically, this is related to lazy-loading, right? If you don't pass the height and width, the browser doesn't know how much space to allot for the image.

bspeare:
@pyrello yes, and if we want our images to render correctly we need to figure out a way for the responsive images to appear correctly and avoid cumulative layout shift.   https://web.dev/optimize-cls/ recommends one possibility is to remove the width and height attributes and set the `aspect-ratio` in CSS to the browser knows what size to render.

Issue #6041, opened 2022-12-19
Classrooms - Remaining Fields


briand44:
# Additional Info

- Tile Details
    - multiple items - text field
    - when we migrate we will need to combine the existing two fields into one
    - There does not need to be a filter on the search page for this
- Design Details
    - multiple items - text field
    - when we migrate we will need to combine the existing two fields into one
    - There does not need to be a filter on the search page for this
- Classroom Furniture
    - multiple items - text field
    - when we migrate we will need to combine the existing two fields into one
    - There does not need to be a filter on the search page for this
- Scheduling Region - comes from MAUI room data API https://api.maui.uiowa.edu/maui/api/pub/registrar/courses/AstraRoomData/ajb/e105
    - "SC SW" is an example from the room in the API link above
    - From the regionList
    - This sounds to me like a taxonomy similar to the features where they will need to populate the list of terms, have control over how it outputs on the page vs how it is in the API (API Name field).
    - There will be a filter on the search page to filter by scheduling region.
    - https://classrooms.prod.drupal.uiowa.edu/university-classroom-scheduling-regions

# Proposed solution

joewhitsitt:
would three text areas (instead of multi textfields) be an option here too or is that inviting too much inconsistency with styling, content length?

briand44:
Possibly... These fields all look like text fields currently so I don't think we will run into any issues migrating the content over. They haven't asked for any changes with how that info is handled and based on the styling in the mockup (https://sandbox.prod.drupal.uiowa.edu/art-building-west-240) it seems like multi text fields will better allow styling similar to other room features

PR #6122, opened 2023-01-23
Classrooms - Room list view

### LINKED ISSUE #6098: Develop view and filters for Classrooms search page
# Story

As a Classrooms website owner, I would like the ability to see all rooms and have the ability to search and filter to quickly find the room I am interested in.

# Additional Info

There is a mockup that lives at https://classrooms.prod.drupal.uiowa.edu/find-university-classroom.

**Search**
Search by title (building + room)

**Filters**
Building
Room Type
Room Features
Technology Features
Capacity
Furniture
Scheduling Region

# Proposed Solution

- [x] Build out views and filters
- [x] Theme filters and teasers

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6098

# To Test
```
ddev composer install && ddev blt frontend && ddev blt ds --site=classrooms.uiowa.edu
```

- Room list should be available for placement within basic page layout builder
- Filters form #6098 should all be accounted except for furniture.
- Teasers should look similar to mockup.
- Programmed Classrooms can be added to the site, but do not show in this room list. (e.g. https://api.maui.uiowa.edu/maui/api/pub/registrar/courses/AstraRoomData/BLB/235)
- Unit taxonomy pages are linked from the full node and if content is filled out, looks okay with a grid of rooms.
- Building link on the full node should link to a 404 /buildings/{building_id} in preparation for the next slice.
- Scheduling regions should link to node/151 instead of the term page.
- If you enable the person type filter for SiteNow People, you should still see a dropdown of person types.


bspeare:
It looks like we need this patch or upgrade the module to get the min/max fieldset to appear within the "Filters" details: https://www.drupal.org/project/better_exposed_filters/issues/3175500#comment-14297189.   @joewhitsitt what do you think?

joewhitsitt:
i'd vote upgrade unless unless our patch from https://www.drupal.org/project/better_exposed_filters/issues/3103626 no longer applies...

joewhitsitt:
Upgrading worked @bspeare. Nice find!

Screen Shot 2023-01-25 at 1 29 11 PM

cory-skeers:
Focus keeps getting placed on maximum/max occupancy while interacting with the form. Example below:

https://user-images.githubusercontent.com/49074566/214964407-eb22d2b8-62ba-4ebb-912d-78bb935ce6a1.mov

The scrolling is jarring, and like in the example above while using the text field, if there's a pause in typing the filtering starts, and the focus shift moves the cursor as well.

cory-skeers:
It looks like we have a patch related to the focus already (https://www.drupal.org/project/better_exposed_filters/issues/3103626), but that it's not working entirely as expected with multiple text and select boxes.

One adjustment I tried locally was moving the combined text field to the last filter applied (visually still first), and in that case any form submission shifts focus to it. It's still not 100% desirable, but in the cases of using a select box or typing in that field already, it helps.

joewhitsitt:
Bummer. I forgot to test that after seeing it apply successfully. I can poke around a bit

joewhitsitt:
Updating the patch alone did not work 😞

The issue seems to make the point that no.12 only worked for textfield only exposed filters. We are adding a bunch to that and we are already relying on a module update to treat the max occupancy range input like everything else. Removing that field seems to make it work as expected... Given that, I will probably commit the patch update anyway but... now we need to figure out a workaround for maximum occupancy.

cory-skeers:
Couple other notes:

- Furniture type filter isn't on there. I kinda remember that might have been consolidated as part of "room features," but not sure.
- Should the capacity (max occupancy) be inside the "Filters" group? Currently if you collapse filters you get the free text search as well as occupancy
- The "Room Details" button in the teasers don't seem to be linked

I primarily got through testing that all the filters worked as expected, excluded things they were supposed to, etc., and had just gotten into testing functionality of the links and such. Going to unassign myself, but can re-test things when you get the occupancy stuff working @joewhitsitt.

joewhitsitt:
>Furniture type filter isn't on there. I kinda remember that might have been consolidated as part of "room features," but not sure.

Good eye, though I don't think we want to add this as a filter even though it was part of the IS. This is a multi-text field so it would probably be added to the combined text filter instead of as a separate filter. @briand44 do you think we can get away with not adding that one?

>Should the capacity (max occupancy) be inside the "Filters" group? Currently if you collapse filters you get the free text search as well as occupancy

It _was_ at one point. I am re-creating that filter though so I will make sure it is back in there.

>The "Room Details" button in the teasers don't seem to be linked

I think that is our faux-button technology. It is a feature that it doesn't appear linked but the whole card is a giant link.

joewhitsitt:
Just to note. Composer install was needed to update BEF initially to allow the max occupancy field (range) to be added to the details fieldset. I had to change the filter anyway though to work with the focus stuff.

Screen Shot 2023-01-27 at 10 56 55 AM
Screen Shot 2023-01-27 at 10 57 02 AM

joewhitsitt:
>Furniture type filter isn't on there. I kinda remember that might have been consolidated as part of "room features," but not sure.

>>Good eye, though I don't think we want to add this as a filter even though it was part of the IS. This is a multi-text field so it would probably be added to the combined text filter instead of as a separate filter. @briand44 do you think we can get away with not adding that one?

I think we should hold off on this for now.

cory-skeers:
> > The "Room Details" button in the teasers don't seem to be linked
> 
> I think that is our faux-button technology. It is a feature that it doesn't appear linked but the whole card is a giant link.

I went back and re-checked this, and it looks like the faux-button is working as expected on initial load, but after the teasers are reloaded from an AJAX call, only the card headline is linked and the rest of the cards are no longer linked.

I'm wondering if it might be a bad interaction with the focus stuff? If so, it might be fixed by the recent work (I haven't pulled the newer commits yet).

joewhitsitt:
>I went back and re-checked this, and it looks like the faux-button is working as expected on initial load, but after the teasers are reloaded from an AJAX call, only the card headline is linked and the rest of the cards are no longer linked.

Ooh interesting. Bummer to the max

briand44:
There does not need to be a filter for Furniture. https://github.com/uiowa/uiowa/issues/6041#issue-1503576829

Apologies for having it incorrectly in the IS.

config/sites/classrooms.uiowa.edu/views.view.room_list.yml
```diff
@@ -0,0 +1,1086 @@
...
+            identifier: field_room_max_occupancy_value
+            optional: true
+            widget: select
+            multiple: false
+            remember: false
+            default_group: All
+            default_group_multiple: {  }
+            group_items:
+              1:
+                title: '25 or less'
```
cory-skeers: Tiny nit if you don't mind 😬 

```suggestion
                title: '25 or fewer'
```

cory-skeers (changes requested):

joewhitsitt:
document.querySelectorAll from docroot/themes/custom/uids_base/uids3/assets/js/click-a11y.js grabs a static list from my five minutes of research on Google. For it to grab a new list after AJAX call, we have to do some stuff different?

This appears to be an issue with ajax pagers too out on prod. after the first page, the linking is lost.

https://stackoverflow.com/questions/54157042/cant-access-new-elements-after-ajax-call-vanilla-js

pyrello:
> document.querySelectorAll from docroot/themes/custom/uids_base/uids3/assets/js/click-a11y.js grabs a static list from my five minutes of research on Google. For it to grab a new list after AJAX call, we have to do some stuff different?
> 
> This appears to be an issue with ajax pagers too out on prod. after the first page, the linking is lost.
> 
> https://stackoverflow.com/questions/54157042/cant-access-new-elements-after-ajax-call-vanilla-js

We may not be appropriately wrapping the click ally script so that it is trigger upon a document fragment being added to the DOM. I think we are maybe just including it as is.

Good find, @joewhitsitt! We have ourselves a bug!

bspeare:
Do we want to change "Apply" to "Search"?  I think we are using "Search" on most of our exposed forms.

bspeare:
Should we switch back to horizontal search now that the occupancy field has changed?

Screen Shot 2023-01-27 at 4 31 25 PM

joewhitsitt:
Screen Shot 2023-01-27 at 4 42 00 PM
ipad air (1180px) still kind of seems smooshed. i am fine with leaving it as-is, but if we only made the change because of max occupancy being a range input then we can switch it back. I felt like we sold the sidebar when we pointed out the pattern was being used on other big sites.

bspeare:
The sidebar seems more smushed too.  Maybe we switch to a 33-67 layout if we don't go to the horizontal?
<details>
Screen Shot 2023-01-27 at 4 53 20 PM
</details>

bspeare:
Also, if we stick with the sidebar maybe move to a card list?  The stacked get a bit tall. 

<details>
Screen Shot 2023-01-30 at 8 37 33 AM
</details>

joewhitsitt:
Alright, twist my arm. Seems like horizontal will be the lesser at this point and match the originally approved mockup. I'll try to make that change here shortly.

pyrello:
I can resolve these conflicts.

pyrello:
> I can resolve these conflicts.

Or @joewhitsitt can do it, I guess.

joewhitsitt:
Sorry. didn't see the notification

joewhitsitt:
I _think_ it is ready to review again. Update the PR instructions. Ignoring https://github.com/uiowa/uiowa/issues/6138 for now.

cory-skeers:
Pulling down to test again

docroot/themes/custom/uids_base/scss/views/view-bef.scss
```diff
@@ -20,6 +20,40 @@
...
+    }
+    .bef--secondary {
+      .details-wrapper {
+        @include breakpoint(sm) {
+          display: block;
+        }
+      }
+    }
+  }
+
```
cory-skeers, line 56: Is this left over from the previous sidebar view, or is this part of the overlap with other work?

bspeare: This is a modifier style that can be used in the future if we want to use this bef form in a sidebar.

cory-skeers: Cool--and judging from previous testing it worked out well, so I'd say it's good to approve without further testing at this point.

bspeare: If we switch back to the range fields as mentioned in https://github.com/uiowa/uiowa/pull/6122#issuecomment-1407161632, we will need this.

cory-skeers (approved):
Re-tested going through updated instructions, and everything looks good to go! 👍

e-marie-w:
Also working as expected for me so I unassigned myself.

PR #6123, opened 2023-01-24
Update Grad program list

### LINKED ISSUE #6083: Grad College - Update program names list
The GC has a list of program names that are in use for their Thesis Defense content type. The list is out of date with changes that have occurred since it was created and it needs to be re-generated. I created a python script that pulls MAUI data and compares it to a list provided by @arjenkin to handle this previously. I think that we can use the script again with a new input file. The output from that script can then be used to update the allowed values function in code that provides the list.

We are not concerned about old records as they get removed after 6 months.

# Proposed solution
- Using an update input file, run the python script.
- Paste the output into the correct file.

pyrello:
Resolves #6083 

# How to test

```
ddev blt ds --site grad.uiowa.edu && ddev drush @grad.local cr && ddev drush @grad.local uli node/add/thesis_defense
```
In the "Degree major" drop-down, check for:
- "Criminology" (newly added PhD option)
- "Mass Communication" (name change from "Mass Communications"
- Consult the list of changes to do any additional spot checking you feel like: https://github.com/uiowa/uiowa/pull/6123/files. You are looking for items is `'PhD'` listed under `'objectives'`.


cory-skeers (approved):
Followed the test instructions and everything looked 👍 

Also spot-checked a few not listed that were removed, had PhD removed from objectives, or were added or had PhD added, and everything checked out.

joewhitsitt:
Should the script be stored in the repo for future use?

pyrello:
> Should the script be stored in the repo for future use?

I added it here for now: https://github.com/uiowa/uiowa/issues/6083#issuecomment-1403889319

Issue #4678, opened 2022-01-18
Card block titles overflow below the container breakpoint


bspeare:
# Story

As a SiteNow user, I can use cards with long titles in 3 and 4 column grids without the titles overflowing the cards. 

As a follow-up to https://github.com/uiowa/uiowa/pull/4674, we identified the need to adjust the grid layout gutter below the container breakpoint to allow more space for the components on smaller devices and avoid orphan column breaks often seen with the three column grid. 

# Proposed solution

```[tasklist]
### Tasks
- [ ] Implement Container Queries in `uids` - https://github.com/uiowa/uids/issues/888
- [ ] Review comments on https://github.com/uiowa/uiowa/issues/6386
- [ ] Review work in progress at https://github.com/uiowa/uiowa/pull/6412
```

bspeare:
Chatted with @pyrello about this and we discussed a few options:
- Investigate implementing CSS hyphens https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens with noting https://github.com/uiowa/uiowa/issues/5946#issuecomment-1315847858. 
- Set up a CSS rule for cards in four column layouts where we force them into a "stacked" orientation if they are set to align right or left within a column.   This would be something we could put in place before we implement container queries. 
- We could have the four column layout adjust to two-columns at a smaller breakpoint or anything below 1100 pixels, for example.

pyrello:
Eventually we should invest in container queries to improve this.

bspeare:
If we are going to fix this bug, maybe we should implement container queries as outlined in https://github.com/uiowa/uiowa/issues/6386?

bspeare:
I updated this with a task list linking to two separate issues describing the work necessary for this bug to be fixed.

bspeare:
Here are some UX improvements that we might want to make:

## Card block

1. If `Media position` `right` or `left` is selected add an inline message `Note: left or right media positions will adjust to the stacked media position when the browser is at smaller widths`.  This option would require container queries. 
2.  If a card is placed within a 4 column layout, or a 3 column 25% width column, remove the `Media position` option and set to `stacked` as a default. 

## Card list blocks

1. If `Display list as` `Grid` is selected, remove the `Media position` option and set to `stacked` like in the current Featured Content block. 

## Container queries

1. Implement container queries approach as well in UIDS for cards.

briand44:
[...]

PR #6135, opened 2023-01-27
Block Grid Layout Overflow

### LINKED ISSUE #5946: Blocks in grid layouts overflow layout columns at smaller column widths
# Problem
As you can see with the four-column grid on https://csd.uiowa.edu/research, at smaller breakpoints the grid is no longer equal-width. 

Screen Shot 2022-11-14 at 1 59 03 PM

# Proposed solution
Items within grid layouts need a min-width to maintain the grid: https://css-tricks.com/preventing-a-grid-blowout/. 

- Add `min-width: 0` to https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss#L54. 

Once that is set, the grid widths are maintained but the cards will still overflow.  See details for more information:

<details>

Screen Shot 2022-11-14 at 2 01 49 PM

This line is causing the overflow: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss#L16.   Removing this line will cause words to break:

Screen Shot 2022-11-14 at 2 04 37 PM

Some options to fix that:
- Investigate implementing https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens
- We could adjust this to use CSS clamp and container queries to adjust the sizing of these headings at these smaller widths to maintain 
- Break the grid into three columns earlier to maintain a certain width. 
- 
</details>

# Follow up tasks
- [x] https://github.com/uiowa/uiowa/issues/4678

makurj:
Resolves #5946 

### How to Test:
1. `blt ds` csd.uiowa.edu and default.
2. `yarn workspace uids_base build`
3. Check https://csd.uiowa.ddev.site/research at smaller breakpoints to make sure the grid (cards) is equal-width.
4. On default, create a four-column section with cards and see that the grid stays equal width.
5. If the text is overflowing, don't worry about it. That will be worked on in a [follow-up issue](https://github.com/uiowa/uiowa/issues/4678).


makurj:
After 816ffe8 commit:

Chrome:
image

Safari:
image

bspeare:
@makurj I updated the code.  The code line link that I had provided in the proposed solution had changed since the issue was created.

bspeare:
Tested this with the code below and this fix still appears to work. 
```// This chooses the proper container for rendered content vs building in layout builder.
.fragment .layout__region,
.node__content .layout__region,
.layout-builder .layout-builder-block {
  container-type: inline-size;
  container-name: column;
}

```

bspeare (approved):
This looks good! Thanks @makurj!  I removed the line of code that was from the bad link in the proposed solution.

PR #6141, opened 2023-01-30
Now migration - link source based on redirects

### LINKED ISSUE #6021: Now Migration: Handle 'redirect nodes'
"Items that appear as news features that were entered as redirects to another website don’t have a link. However, there are very few of these that if this is a difficult fix, I’m not sure it’s worth it. Some examples:"

There are some nodes in the source that are empty and serve only as redirects to another website.

Check if a node has an empty body, and if so, query for redirects (`query()` in the migration source plugin can be used for an example of constructing `node/nid` patterns to get node aliases as part of the redirect migration). After finding the redirect destination, add this into the source link, and mark `link directly to source` as `True`.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6021

# To Test
Normal migration setup. In devdesktop setup the now website locally from uiowa703
`ddev blt ds --site=now.uiowa.edu`
`ddev drush @now.local en now_migrate && ddev drush @now.local uli /admin/config/sitenow/sitenow-migrate`
fill in the migration data (now_uiowa_edu)
Add in the query code from https://github.com/uiowa/uiowa/pull/6141/commits/3009d0ee351e4c28ec7c7c13642cda42711782aa based on the findings from https://github.com/uiowa/uiowa/issues/6021#issuecomment-1409085398

`drush @now.local ms` should present you with 54 nodes to test against for the `now_news_feature` instead of testing the +6000k total nodes.

`drush @now.local mim now_news_feature`

As it runs, it should output the original node id and redirect to check now.uiowa.edu against. There should be 51 log messages of the 54 items migrated as three were skipped based on not being external URLs (based on the findings from https://github.com/uiowa/uiowa/issues/6021#issuecomment-1409085398).


bspeare:
Pulling this to test.

joewhitsitt:
Definitely don't recommend running the full migration to test this. The nodes in question all come towards the later half.

The code in https://github.com/uiowa/uiowa/commit/3009d0ee351e4c28ec7c7c13642cda42711782aa (minus the `isset()`s) will really help limit the testing.

bspeare (approved):
Testing steps all worked out.  Too bad some of these source links are old: https://1day.uiowa.edu/s/1773/1day/19/home.aspx, https://magazine.foriowa.org/story.php, etc.

bspeare:
Do we need to add the "preserve links" checkbox on these too? :)

joewhitsitt:
Our articles functionality only redirects teasers, so yes, folks could still get to the full node and see the full node with the bad source link.

In adding the logger to this, I was thinking we could provide a report to the customer. Maybe they can clean some up before it is run in PROD.

joewhitsitt:
Added a report to give to the customers https://github.com/uiowa/uiowa/issues/6021#issuecomment-1409477251

PR #6143, opened 2023-01-30
Programmed Classrooms View

### LINKED ISSUE #6131: Programmed Classroom page and filters
# Story

As a classrooms website owner, I would like to see a separate page that lists programmed classrooms so that they are not confused with University Classrooms.

# Additional Info

- We should only show rooms where the Instructional Room Category is Programmed Classroom.
- Page will be linked from the main menu under Spaces -> Find a Department Classroom.
- They would like the ability to filter by building (requires `uiowa_core_config_entity_filter_select()`).
- They would like to group the display of rooms by building, sorted alpha.
- Can use some design elements from https://sandbox.prod.drupal.uiowa.edu/university-classrooms-and-specialty-spaces if we want.
- Fields to output
    - Room number
    - Building
    - Capacity
    - Seating Type
    - Room Name
    - Responsible Unit.
- Will link to room node.

e-marie-w:
Resolves #6131 

# How to test
-  Code review 
  - Room list block and Programmed list block now use the same code
- `ddev blt ds --site=classrooms.uiowa.edu`
- Add a new page and add the "Room List" and "Programmed List" blocks
  - Test out programmed list filters and verify that these fields display:
    -  Building (Should always display)
    -  Room name (only displays for rooms that have it)
    -  Max Occupancy (Should always display)
    -  Responsible Unit (Should always display)
    -  Seating Type (only displays for those that have it)
 - Verify that the "Room list" displays only Image, Max Occupancy and Room type only. 
 - Check helper function in uiowa_core that allowed removal of a few field template files. 
   - Good naming and usage?
-  On a room page, check that that the request info buttons only display if the room is not a "Programmed Classroom"


docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
```diff
@@ -594,3 +643,93 @@ function classrooms_core_preprocess_views_view(&$variables) {
...
+      foreach ($rows as $id => $row) {
+        // Set remaining row_item_classes as attributes on each row.
+        $variables['rows'][$id]['attributes']->addClass($variables['row_item_classes']);
+
+        // Get node path to link card with.
+        $nid = $view->result[$id]->_entity->id();
+        $link = \Drupal::service('path_alias.manager')->getAliasByPath('/node/' . $nid);
+
+        // Combine fields to display.
+        $text_combined = '<strong>Room name:</strong> ' . $style->getField($id, 'field_room_name') . "<br /><strong>Capacity:</strong> " . $style->getField($id, 'field_room_max_occupancy') . "<br /><strong>Responsible Unit:</strong> " . $style->getField($id, 'field_room_responsible_unit');
```
joewhitsitt: untested, but i wonder if there are a couple places where the following could be used to generate regular field markup instead of crafting custom labels and adding spacing:

```php
$style->getField($id, 'field_room_name')->view(['label' => 'inline', 'type' => 'string']);

or

$style->getField($id, 'field_room_building_id')->view(['label' => 'visually_hidden', 'type' => 'entity_reference_entity_id']); 
```

e-marie-w: Worth trying! It seems like there has to be a simpler way than all the custom stuff.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
```diff
@@ -210,11 +213,51 @@ function classrooms_core_node_presave(EntityInterface $entity) {
...
+
+    case 'field_room_type':
+      if ($variables['element']['#view_mode'] == 'teaser') {
+        classrooms_core_preprocess_field_set_teaser_classes($variables, 'fa-info-circle');
+      }
+      break;
+
     case 'field_room_type':
       $variables['items'][0]['content']['#url'] = Url::fromRoute('entity.node.canonical', ['node' => 206]);
       break;
```
cory-skeers:
```suggestion
    case 'field_room_type':
      if ($variables['element']['#view_mode'] == 'teaser') {
        classrooms_core_preprocess_field_set_teaser_classes($variables, 'fa-info-circle');
      }
      $variables['items'][0]['content']['#url'] = Url::fromRoute('entity.node.canonical', ['node' => 206]);
      break;
```

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
```diff
@@ -594,3 +693,174 @@ function classrooms_core_preprocess_views_view(&$variables) {
...
+
+        // Get the seat taxonomy terms from the features taxonomy.
+        $taxonomy = $node->get('field_room_features')->referencedEntities();
+
+        // Loop through the taxonomy and get items that start with "Seats".
+        $seats = [];
+        foreach ($taxonomy as $term) {
+          if (strpos($term->getName(), 'Seats') === 0) {
+            $seats[] = $term->getName();
+          }
```
cory-skeers:
```suggestion
          if (str_starts_with($term->getName(), 'Seats')) {
            $seats[] = $term->getName();
          }
```

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
bspeare: Just pushed it up, thanks!

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
bspeare: Got it, thanks!

cory-skeers:
> On a room page, check that that the request info buttons only display if the room is not a "Programmed Classroom"

Just checking that I'm thinking of the right buttons--the three at the bottom of the Additional Info area?

image

cory-skeers:
I think `config/sites/classrooms.uiowa.edu/rabbit_hole.behavior_settings.node_type_person.yml` got added as part of a config export that isn't needed (it's a copy of what is already in `config/default`). Wanted to make sure there wasn't an intended change to the person rabbit hole that didn't make it in before taking it out, though.

bspeare:
Thanks @cory-skeers, I did not intend to add that in https://github.com/uiowa/uiowa/commit/91aa94ee3960c2a6eb1230d5c9984f67ac59146b.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -875,3 +875,54 @@ function uiowa_core_config_entity_filter_select(array &$form, $entity_id) {
...
+ *   The classes to be added to the block/field's title attributes.
+ */
+function uiowa_core_set_title_attributes(array &$variables, $title_classes = []) {
+  if (!empty($title_classes)) {
+    if (!isset($variables['title_attributes']['class'])) {
+      $variables['title_attributes']['class'] = [];
+    }
+    $variables['title_attributes']['class'] = [$title_classes];
+  }
+}
```
cory-skeers: Do we need to have the `isset()` checks in here if we're setting it below? Or instead of setting it, should we be adding onto existing values?

Like in
```
  if (!empty($title_classes)) {
    if (!isset($variables['title_attributes']['class'])) {
      $variables['title_attributes']['class'] = [];
    }
    $variables['title_attributes']['class'] = [$title_classes];
  }
  ```
Should we be adding `$title_classes` to existing classes if they're there, or are we wanting to always set them to only the provided `$title_classes`? If the latter, I don't think we need the check/sets.

Comment (bspeare) [reply docroot/modules/custom/uiowa_core/uiowa_core.module]: I think we want it to only be adding `$title_classes` to existing classes if they're there.

Comment (cory-skeers) [reply docroot/modules/custom/uiowa_core/uiowa_core.module]: Alright, so I think they need to be updated to something like 
```  
if (!empty($title_classes)) {
    if (!isset($variables['title_attributes']['class'])) {
      $variables['title_attributes']['class'] = [];
    }
    $variables['title_attributes']['class'] = array_unique(array_merge($variables['title_attributes']['class'], [$title_classes]));
  }
```
so we don't lose the existing, and don't get duplicates?

And something like this for the string?
```
$variables['field_label_icon'] = implode(' ', [$variables['field_label_icon'], $icon_classes]);
```

Comment (cory-skeers) [reply docroot/modules/custom/uiowa_core/uiowa_core.module]: Unless we want to also check that we're not adding duplicate classes if they overlap what's already been set...not sure if that gets resolved somewhere later down the line or not.

File: docroot/modules/custom/uiowa_core/uiowa_core.module [hunk 5]
```diff
@@ -875,3 +875,54 @@ function uiowa_core_config_entity_filter_select(array &$form, $entity_id) {
...
+ *   The block/field variables array.
+ * @param string $title_classes
+ *   The classes to be added to the block/field's title attributes.
+ */
+function uiowa_core_set_title_attributes(array &$variables, $title_classes = []) {
+  if (!empty($title_classes)) {
+    if (!isset($variables['title_attributes']['class'])) {
+      $variables['title_attributes']['class'] = [];
+    }
+    $variables['title_attributes']['class'] = [$title_classes];
```
pyrello, line 926:
```suggestion
    $variables['title_attributes']['class'][] = $title_classes;
```

docroot/modules/custom/uiowa_core/uiowa_core.module
pyrello: I didn't look at this very carefully before. I think we want to think through this a little bit more. We are doing this type of thing all over the place, so the helper functions would be very useful.

It would be nice to have the class that gets passed in be able to be a string or an array. So, if you just need to add one class, you can pass `'my-class'`, but if you need to add multiple classes, you can pass an array.

cory-skeers: I took a pass at refactoring these a bit:

```
/**
 * Helper function to set the block/field's title attributes.
 *
 * @param array $variables
 *   The block/field variables array.
 * @param array|string $title_classes
 *   The class(es) to be added to the block/field's title attributes.
 */
function uiowa_core_set_title_attributes(array &$variables, array|string $title_classes = []): void {
  if (empty($title_classes)) {
    return;
  }
  if (!isset($variables['title_attributes']['class'])) {
    $variables['title_attributes']['class'] = [];
  }
  if (!is_array($title_classes)) {
    $title_classes = [$title_classes];
  }
  $variables['title_attributes']['class'] = array_unique(array_merge($variables['title_attributes']['class'], $title_classes));
}
```

pyrello: I think you could potentially remove the `= []` in the function argument for the classes and then remove the section related to it being empty. This would mean that PhpStorm would be able to tell us if we are not passing anything in so we can correct that. We really shouldn't be using this function if we are going to pass anything for the second argument.

cory-skeers: @pyrello Good point!

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.module
```diff
@@ -546,20 +576,65 @@ function classrooms_core_theme($existing, $type, $theme, $path) {
...
+            'preRenderSelect',
+          ],
+          [
+            '\Drupal\Core\Render\Element\RenderElement',
+            'preRenderGroup',
+          ],
+        ];
+      }
+      break;
   }
```
cory-skeers: Do we need to keep these as separate cases? They seem to end up with the same result in both cases, but in the first we're using an existing field to build it, and the second we explicitly define it.

Instead of a switch, maybe an `if in_array(...)` still to cover our bases?

cory-skeers:
```suggestion
  if (in_array($block_id, ['block_rooms', 'block_programmed_rooms_list'])) {
    $building_filter = 'field_room_building_id_target_id';
    if (isset($form[$building_filter])) {
      uiowa_core_config_entity_filter_select($form[$building_filter], 'building');

      // Change the label from default `- Any -`.
      $form[$building_filter]['#empty_option'] = t('Building');

      // Hack to further treat this field as a select as it relates to BEF.
      // Use the `field_room_type_target_id` if it is present
      // on the form, assuming it is a select.
      if (isset($form['field_room_type_target_id'])) {
        $form[$building_filter]['#process'] = $form['field_room_type_target_id']['#process'];
        $form[$building_filter]['#pre_render'] = $form['field_room_type_target_id']['#pre_render'];
      }
      // If the `field_room_type_target_id` field isn't there,
      // build it from scratch, but this is dependent on BEF
      // select fields not changing in structure.
      else {
        $form[$building_filter]['#process'] = [
          [
            'Drupal\Core\Render\Element\Select',
            'processSelect',
          ],
          [
            'Drupal\Core\Render\Element\Select',
            'processAjaxForm',
          ],
          [
            'Drupal\inline_form_errors\RenderElementHelper',
            'processElement',
          ],
          [
            '\Drupal\Core\Render\Element\RenderElement',
            'processGroup',
          ],
        ];
        $form[$building_filter]['#pre_render'] = [
          [
            'Drupal\Core\Render\Element\Select',
            'preRenderSelect',
          ],
          [
            '\Drupal\Core\Render\Element\RenderElement',
            'preRenderGroup',
          ],
        ];
      }
    }
  }
```

bspeare: Good question @cory-skeers, @e-marie-w or @joewhitsitt did you work on this part?

joewhitsitt: Yeah that will work. Originally, we were able to target the existing fields to use for block_rooms, but another select was not available on block_programmed_rooms_list to use. The warning at the time is that if BEF or something changes how these are processed, we may need to update them.

cory-skeers: Ah, that makes sense.

> if BEF or something changes how these are processed, we may need to update them.

Think it's worth leaving part of the field copying in place for the case where it's there? If a BEF change would break it then it would only break one and we'd have an easy example on how to fix it.

Something like 
```
if (isset($form['field_room_type_target_id'])) {
  ...set the process and prerender in a copy
}
else {
  ...explicitly define
}
```

Update: I updated the suggested change in https://github.com/uiowa/uiowa/pull/6143#discussion_r1103252906 to reflect this option.

cory-skeers (approved):
Reviewed and tested--looking great!

PR #6144, opened 2023-01-31
Updated to latest version of acquia/blt

pyrello:
# How to test
- [Review code changes](https://github.com/uiowa/uiowa/pull/6144/files) - most of them are from PHPCS errors that popped up after switching to the new rules.
- Confirm build completes
- Run `ddev blt ds` to test a sync.
- Run any other blt commands you care to test out.
- You should still see twig debug info on your local sites.


pyrello:
As part of this (and out of curiosity) I started comparing this project with https://github.com/acquia/drupal-recommended-project, which is the current iteration of Acquia's project template (replace their BLT project template, which is now archived). In that project template, they use a different PHPCS standard than we are using. I tried implementing it and (with some fixes applied) it appears to work. I'm wondering if we have any qualms about moving to this standard instead of what we were using before (which was based on the old BLT project template)?

joewhitsitt:
>I'm wondering if we have any qualms about moving to this standard instead of what we were using before (which was based on the old BLT project template)?

🤷

pyrello:
> > I'm wondering if we have any qualms about moving to this standard instead of what we were using before (which was based on the old BLT project template)?
> 
> 🤷

Let's try it out and revert back if we run into problems. We were pretty much going with Acquia's previous standard solution, so this update is in line with that.

joewhitsitt:
pulling to test

joewhitsitt:
Related tidbit. PHPStorm can be set to put use statements in alpha order like the changes in this PR. https://blog.jetbrains.com/phpstorm/2019/04/reworked-imports-code-cleanup/#sort_use_statements

pyrello:
There is a `Code > Optimize Imports` option I've used before to put them in order and cleanup unused references.

joewhitsitt:
Looks like blt.development.services.yml was removed which then gets removed during site provision or `bis`. We are using that to turn on twig debug (docroot/sites/blt.development.services.yml)

no indication about where the commit comes from https://github.com/acquia/blt/commit/3f5044e34eebdc819faebb4d12f83137f4f83bed

looks like it was removed and should be changed to development.services.yml
https://drupal.slack.com/archives/CBH49JN2H/p1663971742636319

joewhitsitt (approved):
looks good to me. tried a dry-run deploy, no-commit/no-db site provision, initialize settings, git clean

also tested replicate as that changed slightly here: docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php

someone should probably just check my commit though https://github.com/uiowa/uiowa/pull/6144/commits/816ea12309b515c77a0b2548c283548bc80830c4

pyrello:
> someone should probably just check my commit though https://github.com/uiowa/uiowa/commit/816ea12309b515c77a0b2548c283548bc80830c4

I reviewed this. I'm not sure it is actually necessary, since I think that is a template file that doesn't get removed and isn't actually called anywhere, but still a good update in terms of housekeeping. Thanks!

joewhitsitt:
the template file is called on `umc` and `bis` for every site.

PR #6151, opened 2023-02-01
Acquia Connector - update config post-update

joewhitsitt:
Follow up from https://github.com/uiowa/uiowa/pull/6146

Acquia connector no longer tracks this stuff in config and instead uses the State API.

This PR updates the config for the default (new sites) and imports the leaner config around config_ignore for existing sites. Also removes config override we were using to specify a site name for the acquia_connector settings.

~~I think there is potential of misfires when this goes out where module code fires before config is set but I doubt anything would cause interruptions.~~

>Because of the removal of the classes, there is a chance of 500 error due to cache needing to be clear because the classes are stored in the DB - Sean

# To Test

- Push out to PROD to see if we still "get the greatest benefit from Cloud Platform."


config/default/acquia_connector.settings.yml
```diff
@@ -1,163 +1,5 @@
...
-    - fast_404
-    - enabled
-  allow_insecure_uploads:
-    - system.file
-    - allow_insecure_uploads
-  http_response_debug_cacheability_headers:
-    - container_parameter
-    - http.response.debug_cacheability_headers
-banner_service: 'https://insight.acquia.com/system/acquia-banner'
+third_party_settings: null
```
pyrello:
```suggestion
third_party_settings: {  }
```

joewhitsitt:
I'd like to schedule this for deployment, so I am still looking for approval.

e-marie-w (approved):
Approving based on previous discussions of concerns and decision to move forward with a scheduled deployment of it.

PR #6152, opened 2023-02-01
Random set of config fixes

joewhitsitt:
Fixes pathauto for publication and adds no results text to two views.


joewhitsitt:
There were more related to https://github.com/uiowa/uiowa/pull/6053 (core.entity, role, etc.) but I am planning on holding off until that is deployed all the way to PROD.

joewhitsitt:
Forgot about this. Small enough to go as-is

e-marie-w (approved):
The building view blocks work as expected, and the pathauto pattern file name now matches the node type, as does its id and label.

PR #6162, opened 2023-02-02
Hotfix: Updated to new config entity filter function

pyrello:
Resolves an issue with people list not loading because of a function being called that doesn't exist.

# How to test

```
ddev blt ds && ddev drush @default.local uli node/1/layout
```
- Confirm that you can add a block and that the person types filter works.


cory-skeers (approved):
Looks good! Reviewed and tested, and can successfully add/edit person blocks again.

PR #6167, opened 2023-02-03
Article subhead

### LINKED ISSUE #6147: Add "subhead" field to default article content type
- Outside the featured image
- Use intro-light class for styling
- Plain long text
- Need to consider blending into RSS feeds as one body

### LINKED ISSUE #6150: Now migration - keywords are not being migrated due to typo
# Problem
The migrations have not been moving over keywords to tags as mentioned in https://github.com/uiowa/uiowa/issues/5887 due to a typo, however, given that keywords has 15k terms, we may have to change our approach to migrating these to reduce memory consumption.

# Steps to reproduce
Notice an absence of 15k rows: https://now.dev.drupal.uiowa.edu/admin/structure/taxonomy/manage/tags/overview
Notice 151 pages of keywords here https://now.uiowa.edu/admin/structure/taxonomy/keywords

# Proposed solution
- Try fixing the typos:
  - https://github.com/uiowa/uiowa/blob/main/docroot/sites/now.uiowa.edu/modules/now_migrate/src/Plugin/migrate/source/Achievement.php#L65
  - https://github.com/uiowa/uiowa/blob/main/docroot/sites/now.uiowa.edu/modules/now_migrate/src/Plugin/migrate/source/InTheNews.php#L122
  -  https://github.com/uiowa/uiowa/blob/main/docroot/sites/now.uiowa.edu/modules/now_migrate/src/Plugin/migrate/source/NewsFeature.php#L44

OR

- Refactor term migrations so that the migration is still able to run.

Refactoring would only be required if there are potential memory issues with holding the mapping table in memory during migration. Including all keywords, there are ~15k terms. How the migration is currently set up, and how it's run in past migrations, ALL migrated terms are held in memory in order to provide a mapping from `name => tid` in order to avoid extra queries and consolidate duplicates with existing terms.

If memory usage is a concern, we should refactor to not hold a static mapping, and instead do streaming lookups to move it to a process-intensive rather than memory-intensive process. For each tid the new process would be:

1. query the source database to get the taxonomy name
2. query the destination database for taxonomy terms matching that name
  a. if found, grab that tid
  b. if not found, create a new term, save, and get the saved term's new tid
3. add the new tid to the tids array and attach to source properties (final step in the current process as well)

Overall, the process is very similar to the steps we have now, but the destination database querying would have to be handled on an individual basis (either once per term or slightly more efficient once per node) rather than once at the start of each migration run.

cory-skeers:
Resolves #6147 
Resolves #6150 

# How to test

`git checkout article-subhead && git pull && ddev blt ds --site=default && ddev drush @default.local uli`
Add a new article or edit an existing article
Test out the new subhead field
View the page source and make sure there's only one schema script appearing

Repeat testing on various splits that are built off the article such as periodical feature, now.uiowa.edu, etc. (https://github.com/uiowa/uiowa/find/main).

For Now migration,
`ddev blt ds --site=now.uiowa.edu && ddev drush @now.local en now_migrate -y && ddev drush @now.local uli /admin/config/sitenow/sitenow-migrate`
Complete form
`ddev drush @now.local mim now_news_feature --limit=200`
Check some of the migrated News Feature articles, see that the subhead has been filled in
Check out the Tags and either spot-check a few articles against their source, or check that there are a significant number of Tags including audiences and generic keywords, and that they aren't duplicating.

Transparency: I also updated Point of Contact reference field labels from "Point of Contact(s)" to "Point(s) of Contact" and removed an extra 'point of contact' block from the person display in the now.uiowa.edu site split. Code reviews for these parts are probably adequate.


bspeare:
Should we set limit on the amount of characters available?

Screenshot 2023-02-08 at 2 05 54 PM

cory-skeers:
> Should we set limit on the amount of characters available?

Good question!

255? Matches the image caption limit, and seems like it should be on the high end where less would be better, but still usable.
image
Basically takes up a full iPhone SE screen, but does fit:
image

I think it also might be good to adjust the form to 2 lines instead of 3, which was already down from the default 5, to help demonstrate visually that it should be a shorter entry.

bspeare:
@cory-skeers that sounds like a good idea!

cory-skeers:
> I think it also might be good to adjust the form to 2 lines instead of 3, which was already down from the default 5, to help demonstrate visually that it should be a shorter entry.

Changed a little from this original intention. Changed it from a `string_long` to a `string` since we're including a limit, and part of that change also changed the form field from a textarea to a textfield, so it's a single line that's longer than the title field. I think this matches the intent of the field well:
image

e-marie-w:
Pulling down to test.

pyrello:
Making visible that this was work that was affected by the issue with LB sections not having UUIDs. If we already had a solution for that, we probably would not have had to update all the article EVD splits.

cory-skeers:
Adjusted the approach for #6150 to check for tag's existing on the destination for each node, rather than the bulk grab and store in memory. In testing with a lower memory allocation, it got as low as exiting every 5-10 nodes. Making this change got it back to 100+ without any noticeable change in speed by increasing the amount of queries.

Going to update the other migrations as well.

e-marie-w:
The subhead works as expected on default and the site splits (91stMeridian, IISC, Iowa Now). The 'Point(s) of Contact' and other now.uiowa.edu changes work and look great too.

The only thing right now is the removal of config/features/periodical/config_split.patch.schemadotorg.schemadotorg_mapping.node.article.yml. It fixes the two schema scripts problem, but the reason we added it was for the isPartOf property that links the article to an issue. 

Assuming we still want that schema, we could create a follow-up issue to fix it, and for now just leave it removed. I might poke at it some. I see an earlier commit that created a schemadotorg.schemadotorg_mapping.node.article.yml file in default, and I'm curious if that helps.

cory-skeers:
> The only thing right now is the removal of config/features/periodical/config_split.patch.schemadotorg.schemadotorg_mapping.node.article.yml. It fixes the two schema scripts problem, but the reason we added it was for the isPartOf property that links the article to an issue.

Shoot! That was my mistake, and should stay there. We can see about cleaning things up and consolidating in the future, but for now I think it's more important to keep the `isPartOf` than clean up the doubling up.

I'd originally been adding the schema mapping to default and meant to roll that back, and went too far in cleaning things up.

e-marie-w (approved):
Subhead is working great! Thanks for adjusting the schema. I did code review only of the migration files, but based on earlier conversations that's all that's needed at this point. They make sense and look good so I'm approving.

PR #6172, opened 2023-02-08
Ignore development.services.yml

pyrello:
After #6144, we don't have a separate local `.services.yml` file. It appears that best practice is to not include this file and modify it locally.

# How to test

Code review.


joewhitsitt (approved):
Thanks for putting this together. Capturing an extra bit for the future folks.

>If core updates the file, we will need to make sure everyone gets the changes as it could be overwritten during composer install. It was last changed six years ago: https://git.drupalcode.org/project/drupal/-/commits/10.1.x/sites/development.services.yml

pyrello:
To get twig debugging working again after this file has been overwritten locally (which might happen when we update Drupal core) you need it to look like this:
```yml
# Local development services.
#
# To activate this feature, follow the instructions at the top of the
# 'example.settings.local.php' file, which sits next to this file.
parameters:
  http.response.debug_cacheability_headers: true
  # Enable twig debugging.
  twig.config:
    debug: true
    auto_reload: true
    cache: false
services:
  cache.backend.null:
    class: Drupal\Core\Cache\NullBackendFactory

```

PR #6054, opened 2022-12-28
Card 4.x updates

pyrello:
**Issues:**

Resolves #5970 
Resolves #5043
Resolves #5971 
Resolves #5044
Resolves #5041
Resolves #5034
Resolves #5040
Resolves #5232
Resolves #6291

**Bugs:**

Resolves #4953
Resolves #6016
Resolves #6094
Resolves #5631
Resolves #5398
Resolves #6334
Resolves #5565

# How to test
**Note:** all stories start with _"As a content editor/webmaster, ..."_
Run once/each time switching to this branch:
```
ddev blt frontend && ddev composer install
```
### Card block
```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli debug-one-card-block
```
- Confirm that all the card block meets all these criteria: https://github.com/uiowa/uiowa/issues/5971#user-content-block-card
- _If I set a title and link text for a card, I can have an `aria-describedby` attribute added to the title link and and `id` attribute added to the pseudo-link `div`._
- _I can set the heading size for a card._
- _I can set the following layout builder styles and see that they work correctly individually and with each other_:
  - Background
    - Test setting a section background that is both the same and different than the card background.
  - Content alignment
  - Media position
  - Media format
    - _When I change the media format, I see that the image styles that get loaded for the responsive image match._
  - Media size
  - Remove default top margin
  - Align button to bottom
  - No padding
    - Test the effect of adding 'No padding' to the above background settings.
  - Add padding to media
- Spot check to compare cards https://sandbox.uiowa.ddev.site/cards vs https://sandbox.prod.drupal.uiowa.edu/cards to ensure that update hook made the correct updates.

### Event block
- Confirm that all the event block meets all these criteria: https://github.com/uiowa/uiowa/issues/5971#user-content-block-event

### RSS feed (aggregator) block
- _I see the appropriate heading size for feed items depending on whether I added a headline to the block or not._
- _I can set layout builder styles and see the my choices reflected in feed item cards in the Layout Builder preview and in the saved layout._

### Article list block
- Test existing and updated LB styles

### People list block
- Test existing and updated LB styles

### UI Events block
```
ddev blt frontend && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli node/1366
```
- Story: _I can see that the block defaults are reflected on save._
- Story: _When I change the media format, I see that the image styles that get loaded for the imagecache_external responsive image match._
[...]

PR #6173, opened 2023-02-08
Add layout with card sidebar

pyrello:
Pulls the new layout from #6054 so that it can be used in other PR's that are in progress, such as #6143.

# How to test

@bspeare is familiar with how this works already, so I'm going to let him test it.


pyrello:
I guess the changes between 3.x and 4.x cards make this too much of a lift to try to separate out.

PR #6194, opened 2023-02-16
Hotfix: Added style fix to adjust author to display separate lines for articles_narrow

bspeare:
This splits the authors on to separate fields. 

# How to test
- Code review or
- `ddev blt frontend`
- `ddev blt ds --site=ilr.law.uiowa.edu`
-  `ddev drush @lawilr.local config-split:activate articles_narrow`
-  Go to https://lawilr.uiowa.ddev.site/news/2023/01/ai-taxation-and-valuation and verify authors are on separate lines. 

Before:

Screenshot 2023-02-16 at 1 43 22 PM

After:

Screenshot 2023-02-16 at 1 43 07 PM

### DIFF
--- a/docroot/themes/custom/uids_base/scss/content/node--type--article.scss
+++ b/docroot/themes/custom/uids_base/scss/content/node--type--article.scss
@@ -170,6 +170,12 @@
     font-weight: 400;
   }
 
+  .article--meta {
+    .field__item {
+      display: block;
+    }
+  }
+
   .field--name-field-article-subhead {
     margin: $gutter 0;
   }


pyrello:
I haven't tested, but wondering if this accounts for all the different ways that author/source org can be displayed?

bspeare:
This is just for the article page.  All the other presentations of those fields remain the same.

pyrello:
> This is just for the article page. All the other presentations of those fields remain the same.

I mean, there are different ways that those same fields get displayed on the article page. Just wondering if it works with all the combinations of that?

bspeare:
Since on this page display for articles_narrow everything is stacking in that column and we don't have to worry about all the combinations using "|" and "-" like the current article page where we are trying to make everything appear inline.  Is that what you mean?

pyrello:
> Since on this page display for articles_narrow everything is stacking in that column and we don't have to worry about all the combinations using "|" and "-" like the current article page where we are trying to make everything appear inline. Is that what you mean?

Ah... I see. I was not paying close enough attention. That is what I was referring to.

pyrello:
Pulling to test.

pyrello:
Just wondering if we considered using a comma instead of separate lines?

bspeare:
I think the comma will create strange line breaks.

pyrello (approved):
Reviewed, not tested

PR #6198, opened 2023-02-20
UIowa Hours - Change js local date creation

joewhitsitt:
From its-web:

>The hours are displaying on Firefox but not Google Chrome or Edge. Can you look into this?
https://recserv.uiowa.edu/

# How to test

- Before checking out this branch, pull recserv down locally and try a chrome browser. You will see Dec 1969 results. In Firefox, you will see today's date results.
- Checkout `hours_today_hotfix` and run `cr`
- In all browser you should see the today's results on load and be able to pick a different date to other results.

>The *60000 is indicating the UTC -6 which is CST

### DIFF
--- a/docroot/modules/custom/uiowa_hours/js/filter_form.js
+++ b/docroot/modules/custom/uiowa_hours/js/filter_form.js
@@ -10,7 +10,10 @@
           $('.uiowa-hours-container', this).html('<p>Placeholder for ' + hoursConfig[0] + ' data</p>');
         }
         else {
-          let today = new Date().toLocaleDateString('en-CA', {timeZone: 'America/Chicago'});
+          let date = new Date();
+          let today = new Date(date.getTime() - (date.getTimezoneOffset() * 60000 ))
+            .toISOString()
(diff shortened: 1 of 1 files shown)


joewhitsitt:
I grabbed the cleanest offset code I could find without requiring moment.js. Open to other solutions.

makurj:
I'm not seeing Dec. 1969 results in Chrome (Version 108.0.5359.124 (Official Build) (arm64)) on main before pulling the feature branch, but will continue testing:

image
image

joewhitsitt:
I believe the issue was introduced in 110 which was just released. That is why it was reported yesterday. I saw it in Brave which adopts chromium stuff pretty quickly.

https://chromereleases.googleblog.com/2023/02/stable-channel-desktop-update_16.html
https://brave.com/latest/

makurj:
Was able to reproduce in Brave. Tested, working as expected.

makurj (approved):

joewhitsitt:
Thanks @makurj

PR #6209, opened 2023-02-27
Timeline Bug

GaryRidgway:
## Problem

Timeline was not contained to only its element, and would extend the line very far if not alone in a section. This was especially egregious if there was a headline. This may not be a perfect solution, but it is probably good enough.

## Solution

Fix the css. Code review and potential visual review will suffice. Try both with and without this PR.


cory-skeers (approved):
Updates look good, and tested it out with the timeline on https://newstudents.uiowa.edu/. With this update, it does not extend into the next section.

PR #6236, opened 2023-03-03
Hotfix: Added articles_narrow to config ignore file

bspeare:
# How to test

- Make sure this is the correct entry for the `articles_narrow` feature split.


joewhitsitt (approved):
matches

PR #6239, opened 2023-03-06
Block ip htaccess part deux

joewhitsitt:
Make sure nothing is lost from our htaccess patch file:

https://github.com/uiowa/uiowa/blob/d0c2095edd975eb840b0f786fbdac64b52afd638/patches/core_htaccess.patch

# To Test

`ddev composer install` should only add the IP block and not remove any of the redirects


pyrello (approved):
Reviewed, not tested.

PR #6258, opened 2023-03-13
Add setting for displaying search option.

### LINKED ISSUE #6231: Search functionality for off-brand sites
# Story

As a site owner of a SiteNow v3 website using an off-brand theme, I would like to still have the ability for folks to use the site search so that they can search for content on our site.

# Questions

Does anything need to happen for this?
Should we address [the favicon](https://github.com/uiowa/uiowa/issues/6232) here? That will be addressed in the linked issue.

GaryRidgway:
This  tracks #6231

As a user with an off brand theme, I want to remove the "Search all University of Iowa for <i>TERM</i>" link from the search form.

# How to test

1. Sync a site
2. Go to `/search?terms=yeehaw`
3. Observe that nothing is broken.
4. Go to `/admin/config/sitenow/uiowa-search`
5. Toggle off the new setting.
6. return to `/search?terms=yeehaw`
7. Observe that the "Search all University of Iowa for <i>TERM</i>" link is gone.

### DIFF
--- a/docroot/modules/custom/uiowa_search/src/Controller/UiowaSearchResultsController.php
+++ b/docroot/modules/custom/uiowa_search/src/Controller/UiowaSearchResultsController.php
@@ -38,14 +38,18 @@ class UiowaSearchResultsController extends ControllerBase {
       'site' => 'default_collection',
     ];
 
-    $build['search'] = [
-      '#type' => 'link',
-      '#title' => $this->t('Search all University of Iowa for @terms', ['@terms' => $search_terms]),
-      '#url' => Url::fromUri('https://search.uiowa.edu', ['query' => $search_params]),
-      '#attributes' => [
-        'target' => '_blank',
-      ],
-    ];
+    $display_search_all_uiowa = $config['display_search_all_uiowa'] ?? TRUE;
+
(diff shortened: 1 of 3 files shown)


bspeare:
Tested this out and it seems to be working great.  Are we to assume that each of these sites will use their own Search Engine ID?

bspeare:
Noticed that search gives you uiowa related suggestions, but won't return the results if "search to this site" is checked. 

Screenshot 2023-03-13 at 2 47 27 PM

joewhitsitt:
>Are we to assume that each of these sites will use their own Search Engine ID?

Ideally they should IMO but there are instances where they are not.

GaryRidgway:
I Have verified that the suggestions are based on the CSE ID, so that would have to be a user solution if desired.

joewhitsitt:
>Tested this out and it seems to be working great.

With the information discovered about using separate IDs to avoid auto suggest, is this ready to be approved @bspeare or is another review needed?

I know PHP logic is the way it is, but is it worth capturing default config for site installs? config/default/uiowa_search.settings.yml

joewhitsitt (approved):

joewhitsitt:
i pulled it down to test as well

PR #6304, opened 2023-03-30
Brand - remove button per request

joewhitsitt:
Brand is reducing collaborator access on the site. Here are a few code changes they requested.

Button removed from https://brand.uiowa.ddev.site/lockup-system


joewhitsitt:
waiting on confirmation from Sondra and for anything else that should be updated that is tied to config/code

joewhitsitt:
Doing something different

Issue #6279, opened 2023-03-22
Migrate classrooms building and room id


briand44:
# Story

As a Classrooms website owner, I would like existing University Classroom rooms migrated from our existing D7 website so that we do not have to manually transition all of the manually controlled content.

# Additional Info

We've decided to scope this first slice to migrating the building and room id. This should get migrated to the dev site so that classrooms folks can verify that the appropriate rooms are getting pulled in.

We should only be migrating existing classrooms that are currently published

The second slice would be to migrate the remaining room information.

# Proposed Solution

PR #6320, opened 2023-04-04
Classrooms Room migrate

cory-skeers:
Relates to #6279 

# How to test

Sync D7 site in devdesktop (uiowa703)
Sync classrooms in D8: `ddev blt ds --site=classrooms.uiowa.edu`
`ddev composer install` if you haven't in awhile, or have been working on other branches
Enable the migrate module: `ddev drush @classrooms.local en classrooms_migrate`
Log in and visit the migrate config page: `ddev drush @classrooms.local uli /admin/config/sitenow/sitenow-migrate`
Complete the page using the form help text and devdesktop info
Check the migration status: `ddev drush @classrooms.local ms`
Import the rooms: `ddev drush @classrooms.local mim classrooms_room`
Go into the site, check out some of the newly migrated rooms. 
For this slice, we're primarily concerned with the Building and Room ID fields importing, though for bonus measure, you can check that the "imported" data appears accurate.
Additionally, we should only be getting those rooms which are published on the current site; check both that we have all the published rooms, and that we don't have the unpublished rooms

(one line version, minus the composer install, up to filling in the config page)
`git checkout classrooms_room_migrate && git pull && ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local en classrooms_migrate -y && ddev drush @classrooms.local uli /admin/config/sitenow/sitenow-migrate`

### DIFF
--- a/docroot/sites/classrooms.uiowa.edu/modules/classrooms_migrate/classrooms_migrate.info.yml
+++ b/docroot/sites/classrooms.uiowa.edu/modules/classrooms_migrate/classrooms_migrate.info.yml
@@ -1,0 +1,8 @@
+name: Classrooms Migrate
+type: module
+description: 'Migrate classrooms from Drupal 7 site to Drupal 9.'
+package: Classrooms
+core: 8.x
+core_version_requirement: ^8 || ^9
+dependencies:
+  - sitenow_migrate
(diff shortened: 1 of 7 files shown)


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Looks good. Ran successful and spot checked some published/unpublished nodes. A 
note for the next slice though is that we either need to account for redirects or 
update the destination's path alias to match. Customer input is probably needed for 
this.

cory-skeers:
> Looks good. Ran successful and spot checked some published/unpublished nodes. A
> note for the next slice though is that we either need to account for redirects or
> update the destination's path alias to match. Customer input is probably needed for
> this.

Thanks @joewhitsitt! I added this as a note to the next slice re: the paths.

PR #6324, opened 2023-04-06
Fix Travis

joewhitsitt:
>Is this just a temporary xenial issue or should we stick with focal going forward?
>>I would suggest sticking with Focal going forward since it seems the recent Chrome versions have issues i.e. some dependencies mismatch on xenial.

# To Test

See passing build: https://app.travis-ci.com/github/uiowa/uiowa/builds/262101045

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 os: linux
-dist: xenial
+dist: focal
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Good for Travis, it's good for me.

PR #6332, opened 2023-04-06
Site rename

pyrello:
Renamed site to remove year.

# How to test
Code review.


joewhitsitt (approved):
looks correct. kinda thrown that github didn't treat blt.yml like a move. since you manually did this instead of a deprovision, will you manually delete the initial dbs in AC and create the new ones?

pyrello:
> since you manually did this instead of a deprovision, will you manually delete the initial dbs in AC and create the new ones?

I haven't checked to see if they got deleted or not. I will delete if they weren't.

PR #6340, opened 2023-04-10
Classrooms migration fields update

### LINKED ISSUE #6179: Migrate remaining classroom room data from existing D7 site
# Story

As a Classrooms website owner, I would like existing University Classroom rooms migrated from our existing D7 website so that we do not have to manually transition all of the manually controlled content.

# Additional Info

We only need to migrate University Classrooms (from classrooms.uiowa.edu) and not Programmed Classrooms (registrar.uiowa.edu). Programmed Classrooms will be manually created by Classrooms team by entering Building and room number as all the rest of the content comes from MAUI.

https://github.com/uiowa/uiowa/issues/6279 was created to address migrating the Building and room id.

**Fields**
- Building -> Building - I believe the D9 select list has a key of building code so hopefully we don't need some lookup or something tomigrate building code to building name
- Room ID -> Number
- Image Gallery -> Images - multiple
- Accessibility -> Don't need to map because this field didn't previously exist
- Tile Details -> Tile Details - combine name and details field collection items into string - multiple values
- Design Details -> Design Details - combine name and detail field collection items into string - multiple values
- Furniture -> Classroom Furniture - combine type and details field collection items into string - multiple values
- Body -> Don't need to map because this field didn't previously exist - Can this go away?
- Technology Guide -> Link - The source field allows multiple, the destination does not. If there are instances with multiple values, we may need to adjust the destination field. (`field_room_link`)
- Seating Chart -> Seating Chart
- Video -> Video
- Map -> lets not do anything with this. We can ask our Classroom partners to manually add these.

# Proposed Solution

- [x] #6279
- [x] Second slice - add additional fields to migration

Additionally:
> we either need to account for redirects or
update the destination's path alias to match. Customer input is probably needed for
this.

cory-skeers:
Resolves #6179 

# How to test

One line option to take you up through opening the migrate config page
`git checkout classrooms-migrate-update && git pull && ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local en classrooms_migrate -y && ddev drush @classrooms.local uli /admin/config/sitenow/sitenow-migrate`

Sync the site in devdekstop on uiowa703
Pull the branch
Sync classrooms.uiowa.edu
enable the `classrooms_migrate` module
Log into the site and visit /admin/config/sitenow/sitenow-migrate
Complete the form according to the help text, matching the data in devdesktop and setting the "Base URL" path to https://classrooms.uiowa.edu

`ddev drush @classrooms.local ms` to check that the migration registered properly. Should see over 300 unprocessed waiting
`ddev drush @classrooms.local mim classrooms_room`
Check the migrated rooms, and compare with the source room pages.

Check that the URL alias pattern matches that of the existing site (/buildingAbbreviation-roomNumber, such as /abw-110)

Basic description of the mappings:

- Building -> Building - I believe the D9 select list has a key of building code so hopefully we don't need some lookup or something tomigrate building code to building name
- Room ID -> Number
- Image Gallery -> Images - multiple
- Accessibility -> Don't need to map because this field didn't previously exist
- Tile Details -> Tile Details - combine name and details field collection items into string - multiple values
- Design Details -> Design Details - combine name and detail field collection items into string - multiple values
- Furniture -> Classroom Furniture - combine type and details field collection items into string - multiple values
- Body -> Don't need to map because this field didn't previously exist
- Technology Guide -> Link
- Seating Chart -> Seating Chart
- Video -> Video
- Map -> lets not do anything with this

# Deployment

We need to either remove all Room nodes in prod, or at least ensure that there is no info in the "Seating Chart" field for deployment to succeed. For further coverage, should probably remove past revisions for Room content type as well.

Alternatively, no worries if all Room content can be removed prior to release.


cory-skeers:
There's an additional note regarding aliases/redirects https://github.com/uiowa/uiowa/issues/6179#issuecomment-1505813971, but the rest of this should be ready for testing.

cory-skeers:
Couple notes: 

- Update tech guide title to "Technology Guide" for all
- update Seating Chart field to media instead of file and make corresponding change to migration
- Update room URL pattern to match abbr-num like current site

cory-skeers:
Updated the logic for the Seating Chart, and now am back to this little issue where I'm having trouble getting the value inline with the icon. With the current styles, because it's a a rendered field -> media -> file link, the chain from `field__label` to `field__item` gets broken. If the media div had `field__item` as well, it gets the needed `display: inline`, but I'm not seeing a way to add that without adding another template or targeted css or something.

image

@bspeare, think you could point me in the right direction? 😬

cory-skeers:
Okay, I'm confused. I actually had been able to add `field__item` to the media successfully, but thought it wasn't working because it's not displaying as expected on initial page load. After going through the trouble of adding a targeted css property, I'm still seeing the same behavior.

On initial page load, the text is not inline. After removing and then reapplying the property, it is.


https://user-images.githubusercontent.com/49074566/232095748-425e32e2-6780-4053-8f66-2d24912842db.mov

bspeare:
@cory-skeers Just seeing this.  Is there an easy way to test this? Can I just pull this branch and try adding a file to the content type?

cory-skeers:
> @cory-skeers Just seeing this. Is there an easy way to test this? Can I just pull this branch and try adding a file to the content type?

I just pushed up what I have locally, so you should be able to just pull and add a file. It has both parts--adding `field__item` to media and a targeted css.

bspeare:
@cory-skeers I added a fix to get it to display inline.   Can you verify that it looks good on your machine?

cory-skeers:
> @cory-skeers I added a fix to get it to display inline. Can you verify that it looks good on your machine?

Yes! Looking great--thank you :)

cory-skeers:
I believe this is now ready for testing.

joewhitsitt:
working on the merge conflict so i can review this

joewhitsitt:
Initial thoughts:

The merge was a little weird for some of the classes. Hope it is okay.
The migration seems really slow. My computer has been really slow in general today. Each node is taking a few seconds.

cory-skeers:
> The merge was a little weird for some of the classes. Hope it is okay.
> The migration seems really slow. My computer has been really slow in general today. Each node is taking a few seconds.

A quick look through the files changed, and things look 👍 after the merge.
A little bit slower than a typical migration should be expected--as each node is created it needs to go through the post-save process to fetch the imported data. A few seconds each node seems slower than I was seeing though. 🤔

joewhitsitt:
Migration seemed successful. Need to audit the fields still. One thing that has stuck out though is that the gallery is not a grid like in the mockups. Perhaps a card_4x regression?

Update: might have been a blt frontend issue

joewhitsitt (approved):
nice work @cory-skeers. i didn't find any videos on the current site after noticing no videos on the new site. everything else seemed to match up in spot checking

cory-skeers:
Thanks @joewhitsitt!

PR #6346, opened 2023-04-12
Now migration: remove excess empty tags

cory-skeers:
# How to test

Code review


joewhitsitt (approved):
reviewed regex. not tested. Cory has done some local testing

PR #6354, opened 2023-04-17
Updated IISC projects view to sort by created date first

### LINKED ISSUE #6071: IISC - Investigate fixing publish dates on migrated content
# Motivation

When we migrated content for IISC, it appears that the published date was set to the date/time that the item was migrated. They would like to be able to have projects listed in reverse chronological order by the published date, but this isn't possible because those dates are all wrong.

# Proposed solution

The custom migration module for that site is still turned on. This means that we still have migration mapping tables that we can reference as well as the connection information to connect to the D7 database. I think the easiest thing to do would be write an update hook that goes through all the project nodes that were migrated, look up the original node published date, then update the current node record's publish date. I think this could all be accomplished via the database layer instead of needing to load nodes and save them. It might require a batching mechanism because I think the nodes would need to be looked up one at a time because of the data being stored in multiple databases and I'm not sure it would wise to run a query with a WHERE IN clause that passed in several hundred node IDs.

### LINKED ISSUE #6072: IISC - Update projects list block to display projects by date added
This is a request from IISC. See example page at https://iisc.uiowa.edu/projects

# Proposed solution
Update the view and re-export.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6071. 
Resolves https://github.com/uiowa/uiowa/issues/6072. 

# How to test

- `ddev blt ds --site=iisc.uiowa.edu`
- Go to https://iisc.uiowa.ddev.site/projects
- Verify that default sort of the view is created date first, then alphabetical. 
- Verify that the Published date and academic Year display in the teaser view. 
- Apply filters and verify that sort remains the same. 
- Go to full project page node view and verify that published date is not appearing in layout. 

Updated screenshot:

Screenshot 2023-04-19 at 12 55 22 PM


makurj (approved):
Tested, date sort working as expected. Saw only one example where alphabetical sort came into play after date. As a user who doesn't have access to edit the project and see the date it was created, they wouldn't know what the projects are being sorted by, as there's no date being displayed on the teaser or the project.

bspeare:
Should we reach out to Travis and ask him if he wants to display the published date on each project?

bspeare:
Travis likes the idea of the dates.  Any objections with moving forward with showing published date and academic year on these?

Screenshot 2023-04-19 at 9 42 12 AM

joewhitsitt:
^^ looks good. would the view sort change then to academic dates (string?) instead of created date?

bspeare:
@joewhitsitt Not sure.  The published date is there to let the user know how the list is being sorted by default while there is the additional filter to sort by academic date.    Perhaps this should be re-reviewed to examine these scenarios.

bspeare:
This is ready to be tested again.

makurj:
Pulling down to test.

makurj (approved):
Tested, working as expected.

PR #6370, opened 2023-04-20
Classrooms Room import command

### LINKED ISSUE #5823: Display Facilities Management building info from API
# Story
As a Facilities Management website owner, I would like to see building information from our APIs displayed on building pages.

# Additional Info

**Data Feeds**
https://bizhub.facilities.uiowa.edu/bizhub/ext/buildings
https://bizhub.facilities.uiowa.edu/bizhub/ext/building?bldgnumber=0004

**Existing building**
https://www.facilities.uiowa.edu/building/0003

**Mockups**
https://sandbox.prod.drupal.uiowa.edu/chemistry-building-mockup
https://sandbox.prod.drupal.uiowa.edu/pappajohn-biomedical-discovery-building

# Scope

This is just the first slice of these building pages which includes creating structure, getting info from API and outputting that information for each building page. Additional information and theming can come in future slices. The information in scope is:

Building Name - I believe we already have this from the API and it should be using `buildingCommonName`
Building Number - I believe we already have this from the API and it should be using `buildingNumber`
Building Abbreviation - I believe we already have this from the API and it should be using `buildingAbbreviation`
Building Address - Structure needed and should come from `address`
Area - Structure needed and should come from `grossArea ` - displayed in `sq. ft.`
Year Built - Structure needed and should come from `yearBuilt `
Ownership - Structure needed and should come from `owned `
Photo - Structure needed and should come from `imageUrl` - no need to override this like on the existing site
Named Building - Structure needed?? The link text can come from `namedBuilding` or from the named building entity "last, first". The actual link needs to come from the named building entity. There is a `building id` on the named buildings.
Single user restrooms men - Structure needed and should come from `singleUserRestroomsMen`
Single user restrooms women - Structure needed and should come from `singleUserRestroomsWomen`
Multi user restrooms men - Structure needed and should come from `multiUserRestroomsMen`
Multi user restrooms women - Structure needed and should come from `multiUserRestroomsWomen`
Gender Neutral restrooms - Structure needed and should come from `singleUserRestrooms`
Lactation rooms - Structure needed and should come from `lactationRooms`
Accessibility links - Structure needed?? The links for Accessibility Map, Hearing Loops, Gender Inclusive restrooms and Lactation rooms will be hard coded. Waiting for updated links for some of these.

# Remaining info for future slices

Contact info - Primary and alt building coordinators
Project Info
Map - want to use Concept 3D map but need to discuss how to handle this. Update, they now want to use FM Sims map.
Artwork
Drush command creates buildings, we need to wire this up to a scheduled job.

# Questions

When/how are the buildings created?
Named Building - do we want to structure this? If so, would it be a reference field.
Accessibility links - is there any reason to structure this?
Schema - https://github.com/uiowa/uiowa/issues/5823#issuecomment-1335431828

# Notes

For buildings that are named buildings, we could potentially get a description and the map url from the named building content but that would only work if it is a named building and that info is captured.

# Proposed Solution

### LINKED ISSUE #6180: Classroom room scheduled update
# Story

As a Classrooms website owner, I would like our room information to update daily from MAUI so that our users are getting up to date information that doesn't require us to manually save all rooms. 

# Additional Information

Need to setup periodic import via node_save. Ideally, it would be adjusted to check for a diff before running the full save. I believe the facilities building import process does this check. Once a day should be sufficient.

# Proposed Solution

- [x] Move presave processing to a separate process that can get called separately from a save trigger
- [x] Write a drush command to process and update all rooms; do not double up in presave check

cory-skeers:
Resolves #6180
Resolves #5823 

# How to test

## Classrooms

`ddev blt ds --site=classrooms.uiowa.edu`
The migration has been run in DEV, and you can sync from there to have more buildings to test with. Alternatively, you can use DEV as reference to make individual classroom nodes to test with.

To run the command:
`ddev drush @classrooms.local classrooms-rooms`
Optional `--batch=X` to change the batch size and experiment with that.

If you just run this command, it should result in 0 updates, as anything created and saved recently should have up-to-date info from MAUI.

To put it through its paces:
Use a sql interface or `ddev drush @classrooms.local sqlc` for the CLI to adjust the database
For example:
Delete a row
`DELETE FROM node__field_room_technology_features WHERE revision_id=9154 AND delta=3;`
Update a row
`UPDATE node__field_room_max_occupancy SET field_room_max_occupancy_value = 20 WHERE revision_id=9154;`
Insert a row
`INSERT INTO node__field_room_features VALUES ('room', 0, 241, 9154, 'en', 4, 526);`
(in this case 241 is the entity id, 9154 is the entity revision id, 4 is the delta, and 526 is an example room feature target id. These values may need to be adjusted per your testing)
Alternatively, temporarily remove the node presave data import and field disabling, adjust import data, and then run the command.

After making one or more changes to the database or imported fields, run the import:
`ddev drush @classrooms.local classrooms-rooms`
The message should match the overall number of rooms that needed updates. Eg if you make 10 changes to one room, the message should still be that one room has changed, whereas one change each to 10 rooms should message that 10 updates were made.

Check that the updates undid your changes, and that the altered rooms now again match data pulled from MAUI.

Create a new room, and after saving, check that the data import worked properly.

Additionally, code review of the dependency injection updates (in particular any changes within `getBuildings` in `ClassroomsCoreCommands.php`) would be appreciated.

## Facilities
```
ddev blt ds --site facilities.uiowa.edu && ddev drush @facilities.local fm-buildings && ddev drush @facilities.local uli admin/content
```
- Observe the number of records processed (should be "385 created") when the `fm-buildings` command runs above.
- Spot check some Building records to see that values are getting filled in for the new fields (refer to https://github.com/uiowa/uiowa/issues/5823 for the list of fields added).
- While you are spot checking, remove or update values in the newly added fields. Delete a building. Add a new building. Keep track of how many records you create, modify, delete.
```
ddev drush @facilities.local fm-buildings
```
- Observe the output from the command and compare this to your changes. Any buildings you deleted should be created anew. Any buildings you modified should show as updated. Any buildings you created should show as deleted. The rest should show as skipped. Note that the total number of records should equal 385 + any buildings you created, since those will be deleted and are not part of the count of records coming from the API.
- ???
- Profit!


cory-skeers:
EoD checkin update: Current iteration is reliant on the node postSave which leads to a second call to the API for every update. Instead, can update this to do direct inserts--would miss out on revision, but otherwise keeps things a little tidier and avoids duplicate API calls.

cory-skeers:
I think this could be tidied up further with some helper functions for the database stuff, but at the same time, it's a pretty particular command, and the updates/insertions vary just enough to make it feel to me like more work than benefit.

That said, this is ready for review, and any suggestions or changes to clean up the code are very welcome.

joewhitsitt:
going to pull this down as-is to start testing. Cory is looking into the batch processing. I might look at ways of reducing duplication

joewhitsitt:
Keeping it as an db insert, cache needs to be invalidated manually.

cory-skeers:
> Keeping it as an db insert, cache needs to be invalidated manually.

Good call. That'd explain some weird behavior I was just seeing that I'd chalked up to my attempt at batching.

joewhitsitt:
I might be deleting content wrong but I thought this would and instead I am getting an error:

`Integrity constraint violation: 1062 Duplicate entry '556-0-7-en' for key 'PRIMARY'`

Steps to reproduce:

- Comment out the guts of presave and the disabled the fields in form_alter.
- Open a node and remove a few entity references from the bottom but i have also tried other deltas.
- Save the node.
- Go to revisions and delete previous revisions.
- Run the import.

Alternatively,

I modified `DELETE FROM node__field_room_technology_features WHERE revision_id=9154 AND delta=3;` to `DELETE FROM node__field_room_technology_features WHERE
revision_id=9771 AND delta=7;` and ran into the same issue.

cory-skeers:
> Integrity constraint violation: 1062 Duplicate entry '556-0-7-en' for key 'PRIMARY'

I think the original way room features and room tech features were being updated. I pushed up a change if you want to pull and test again @joewhitsitt.

Edit: I think there was something off with how the row delete/update/insert was being handled and it wasn't always using the correct delta. Instead, the new method just wipes all current features/tech features on the node and then inserts all the new ones. Especially since the deletion can be handled with one go, I don't think we're losing anything here by not re-using and updating existing delta rows.

joewhitsitt:
unfortunately i am getting the same issue after a fresh sync and running `DELETE FROM node__field_room_technology_features WHERE revision_id=9771 AND delta=7;`

`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1996-0-0-en' for key 'PRIMARY'`

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
```diff
@@ -0,0 +1,322 @@
...
+            sort($tech_features);
+            if ($node_tech_features !== $tech_features) {
+              $updated = TRUE;
+              // Remove all current features before adding new,
+              // to avoid possible constraint violations
+              // during database insertions later.
+              \Drupal::database()
+                ->delete('node__field_room_technology_features')
+                ->condition('revision_id', $node->getRevisionId(), '=')
+                ->condition('delta', $delta, '=')
```
joewhitsitt:
```suggestion
```

This is no longer needed correct?

cory-skeers: Ooh, yep. Definitely not.

joewhitsitt:
Using this method:

> Comment out the guts of presave and the disabled the fields in form_alter.
> Open a node and remove a few values.
> Save the node.
> Run the import.

I am noticing that Instructional Room Category, Responsible Unit, Room Type and possibly other single select/references are not getting updated/replaced with the value from the data. Been trying to debug using a smaller sample (prod) but I am running into issues locally with debugging.

joewhitsitt:
I think it is because those are trying to update where the row doesn't even exist

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
```diff
@@ -0,0 +1,321 @@
...
+            \Drupal::database()
+              ->update('node__field_room_max_occupancy')
+              ->fields([
+                'field_room_max_occupancy_value' => $data[0]->maxOccupancy,
+              ])
+              ->condition('revision_id', $node->getRevisionId(), '=')
+              ->execute();
+          }
+        }
+      }
```
pyrello: I'm trying to understand the rationale for doing this vs. just using `$node->set(...); $node->save();`. Is the intention here to piggyback on the previous node revision and not trigger another revision?

cory-skeers: The way things are currently set up, there's a presave that does a lot of this in a similar manner for when rooms are created (or if, say, they NEED to trigger a room to re-import for any reason). Setting and saving the node sets off the presave process again. The piggybacking off the previous revision is a side benefit, but not necessary.

Originally this was set up to call the API, check for differences, and if any were found, set a new revision and save the node. Then the presave handled the actual updates. The shift to db inserts came from a desire to not double-up on all the API calls and value checks in two places.

joewhitsitt: https://github.com/uiowa/uiowa/pull/6370#issuecomment-1518344910

cory-skeers:
> I think it is because those are trying to update where the row doesn't even exist

We could probably changes those to upserts instead. I don't usually go that route because I'd prefer to be explicit, but this is still a pretty specific row we're updating or inserting.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
joewhitsitt: the insert work isn't going as smoothly as I thought and we ultimately have similar node->save() code in the presave().

a node->save in a drush command passes through presave. i was trying to think of a way to abort full save if coming from drush and nothing has changed.

or maybe it should move out of the presave and instead be a local task to trigger an import for that node using a single version of the drush code. we would have to tell webmasters that the workflow changes slightly.

cory-skeers:
Setting this back to draft for clarity, while things get discussed/worked out.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
cory-skeers: Could we so something like check the user of the revision in presave, and if it's uid 1, skip the extra api processing?

joewhitsitt: As I said in EoD, I'd like to avoid using uid 1 as our check. We have logged in as that user to do things for clients and I don't want to add conditionals to that.

`$_SERVER["argv"]` exists when coming from a `$node->save()` in the batch process and could be used as an identifier within presave of what is saving it.

Screenshot 2023-04-26 at 4 53 44 PM

pyrello: Also noting the other idea of doing something like `$node->set('batch_update', TRUE)` in the batch process and then checking that value later in presave before doing additional processing.

cory-skeers:
I made some updates to the classrooms `RoomProcessor`, and think I feel pretty good about where it's at. There's some specific field processing outside of the `parent::process` that is either dependent on the entity or maps one field to two, but outside of those, things are moved into a `processRecord` that occurs prior to the basic field mapping, and I think works as an adequate model. Further additions to `EntityProcessorBase` could be good, but don't all need to happen now, and may be better suited for followup work.

cory-skeers:
Ready for testing. Facilities testing instructions to be updated, but general setup is there. Classrooms testing instructions updated.

pyrello:
Updated with testing instructions for facilities.

joewhitsitt:
pulling to test

docroot/modules/custom/uiowa_core/src/EntityItemProcessorBase.php
```diff
@@ -0,0 +1,33 @@
...
+   */
+  protected static $fieldMap = [];
+
+  /**
+   * Process an individual entity.
+   */
+  public static function process($entity, $record): bool {
+    $updated = FALSE;
+    foreach (static::$fieldMap as $to => $from) {
+      // @todo Add a message if a node doesn't have a field.
```
joewhitsitt: Should this be captured as a follow-up task?

docroot/modules/custom/uiowa_core/src/EntityProcessorBase.php
```diff
@@ -0,0 +1,278 @@
...
+
+    return $this->entityIds;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process() {
+    if (!$this->getData()) {
+      // @todo Add a logging message that data was not able to be returned.
```
joewhitsitt: Should this be a follow-up task?

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/RoomItemProcessor.php
```diff
@@ -0,0 +1,212 @@
...
+    // Grab MAUI room data.
+    $building_id = $entity
+      ->field_room_building_id
+      ?->target_id;
+    $room_id = $entity
+      ->field_room_room_id
+      ?->value;
+    /** @var \Drupal\uiowa_maui\MauiApi $maui_api */
+    $maui_api = \Drupal::service('uiowa_maui.api');
+    // @todo Need to handle $building_id or $room_id being NULL.
```
joewhitsitt: Is this handled below? If so, should this be removed?

cory-skeers: Yep, I think this should be handled alright by the `return $results[0] ?? [];` and the todo can be removed.
```suggestion
```

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
```diff
@@ -0,0 +1,74 @@
...
+    ]);
+
+    foreach ($nodes as $node) {
+
+      if (!$node instanceof FieldableEntityInterface) {
+        continue;
+      }
+      if ($node instanceof NodeInterface) {
+        $record = $room_processor->getRecord($node);
+        $updated = $room_processor->process($node, $record);
```
cory-skeers:
```suggestion
        $updated = $room_processor->process($node, $room_processor->getRecord($node);
```

joewhitsitt: i think this is missing a parentheses at the end

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/BatchRooms.php
```diff
@@ -0,0 +1,73 @@
...
+      '@id' => $batch_id,
+    ]);
+
+    foreach ($nodes as $node) {
+
+      if (!$node instanceof FieldableEntityInterface) {
+        continue;
+      }
+      if ($node instanceof NodeInterface) {
+        $updated = $room_processor->process($node, $room_processor->getRecord($node);
```
joewhitsitt:
```suggestion
        $updated = $room_processor->process($node, $room_processor->getRecord($node));
```

cory-skeers: Oops! Thanks @joewhitsitt!

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/RoomItemProcessor.php
pyrello: I don't think this is currently handled.

https://github.com/uiowa/uiowa/blob/7f48860680cf1bb3d85d35a48e03a97bef5a361b/docroot/modules/custom/uiowa_maui/src/MauiApi.php#L297-L299

Possibly we need to handle it in the above snippet instead, but since it is not currently handled there, if either `$building_id` or `$room_id` was `NULL`, it would fail ungracefully.

cory-skeers: Shoot. It does error in the case of `building_id` being NULL and `room_id` being filled. As long as `room_id` is null (or null as well as `building_id`, it's a successful request that just doesn't end up with data, and so the other catch works fine.

I'd only checked the case if both were NULL.

pyrello: When `NULL` is (automatically) cast to a string, does it become `''`? Or `'null'`?

joewhitsitt:
```
ddev drush @classrooms.local classrooms-rooms
>  [notice] Running Batch "1"
>  [error]  Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[01000]: Warning: 1265 Data truncated for column 'tid' at row 1: INSERT INTO "taxonomy_index" ("nid", "tid", "status", "sticky", "created") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3, :db_insert_placeholder_4); Array
> (
>     [:db_insert_placeholder_0] => 276
>     [:db_insert_placeholder_1] => 1) Programmed Classrooms - Level 2
>     [:db_insert_placeholder_2] => 1
>     [:db_insert_placeholder_3] => 0
>     [:db_insert_placeholder_4] => 1675704060
> )
>  in taxonomy_build_node_index() (line 345 of /var/www/html/docroot/core/modules/taxonomy/taxonomy.module).

```

@cory-skeers wondering if you had tried a fresh sync from DEV when this was refactored? i just gremlin wiped my local db to see if it was me but something persists.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/RoomItemProcessor.php
cory-skeers: > When `NULL` is (automatically) cast to a string, does it become `''`? Or `'null'`?

it becomes `''`

cory-skeers:
> wondering if you had tried a fresh sync from DEV when this was refactored? i just gremlin wiped my local db to see if it was me but something persists.

Yes, but closer to the beginning of the refactor. I'll pull down, re-sync, and take a look.

cory-skeers:
> > wondering if you had tried a fresh sync from DEV when this was refactored? i just gremlin wiped my local db to see if it was me but something persists.
> 
> Yes, but closer to the beginning of the refactor. I'll pull down, re-sync, and take a look.

@joewhitsitt I think this is cleared up now.

In the case where there were scheduling regions in the API record that didn't map to any set in the region taxonomy, the data from the API call was getting set directly into the entity reference field, and it was then trying to create terms with some funky data.

joewhitsitt:
Nice find with the scheduling regions. Noticed this now as I sync from PROD but not from the DEV site. Is this because of dummy content?

```
ddev drush @classrooms.local classrooms-rooms
>  [notice] Running Batch "1"
>  [warning] Attempt to read property "maxOccupancy" on array EntityItemProcessorBase.php:24
>  [warning] Attempt to read property "roomName" on array EntityItemProcessorBase.php:24
>  [warning] Attempt to read property "roomType" on array EntityItemProcessorBase.php:24
>  [warning] Attempt to read property "acadOrgUnitName" on array EntityItemProcessorBase.php:24
>  [warning] Attempt to read property "regionList" on array EntityItemProcessorBase.php:24
>  [notice] Message: 13 results updated. That is neat.
> 
 [notice] Update batch operations ended.
```

Update: I wanna say that this is because of the non-existent ABW 123 room.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/src/RoomItemProcessor.php
```diff
@@ -0,0 +1,218 @@
...
+        foreach ($terms as $term) {
+          if ($api_mapping = $term->get('field_api_mapping')?->value) {
+            if (in_array($api_mapping, $record->regionList)) {
+              // If we found a mappable region, set it.
+              $region = $term->id();
+              break;
+            }
+          }
+        }
+        $record->regionList = $region;
```
pyrello, line 186:
```suggestion
        $record->regionList = NULL;
        foreach ($terms as $term) {
          if ($api_mapping = $term->get('field_api_mapping')?->value) {
            if (in_array($api_mapping, $record->regionList)) {
              // If we found a mappable region, set it.
              $record->regionList = $term->id();
              break;
            }
          }
        }
```
I think this should work without creating a variable.

pyrello:
> `[warning] Attempt to read property "maxOccupancy" on array EntityItemProcessorBase.php:24`

FWIW, the specific warning seems to be indicating that its trying to read `$record->maxOccupancy` from an array, which might indicate that we need to cast something before it gets to that point. Or that something is being cast and it shouldn't be. Or our typehint is wrong.

joewhitsitt:
Was `imageUrl` from Facilities supposed to be captured in this slice or not?

pyrello:
> Was `imageUrl` from Facilities supposed to be captured in this slice or not?

D'oh! Yes, it was.

cory-skeers:
> Nice find with the scheduling regions. Noticed this now as I sync from PROD but not from the DEV site. Is this because of dummy content?
> 
> ```
> ddev drush @classrooms.local classrooms-rooms
> >  [notice] Running Batch "1"
> >  [warning] Attempt to read property "maxOccupancy" on array EntityItemProcessorBase.php:24
> >  [warning] Attempt to read property "roomName" on array EntityItemProcessorBase.php:24
> >  [warning] Attempt to read property "roomType" on array EntityItemProcessorBase.php:24
> >  [warning] Attempt to read property "acadOrgUnitName" on array EntityItemProcessorBase.php:24
> >  [warning] Attempt to read property "regionList" on array EntityItemProcessorBase.php:24
> >  [notice] Message: 13 results updated. That is neat.
> > 
>  [notice] Update batch operations ended.
> ```

Ah, yes...this looks to be what happens when there is no data from the source. One node on prod has a dummy Art Building West room 123 that doesn't have any data, and the entity processor is trying to read a property from an empty array.

pyrello:
Setting back to draft because I overlooked the `imageUrl` field. Going to work on that.

joewhitsitt:
Overall this is looking good. Just a little further tidying and I will approve.

cory-skeers:
> Nice find with the scheduling regions. Noticed this now as I sync from PROD but not from the DEV site. Is this because of dummy content?

Pushed up a small check for this. Ideally I think we'll want to update these processes with some logging, but for now it won't come up with the errors (or try to save a node that wasn't/shouldn't be updated with no data).

docroot/modules/custom/uiowa_core/src/EntityItemProcessorBase.php
```diff
@@ -0,0 +1,33 @@
...
+      if ($entity->hasField($to) && $entity->get($to)->value != $record->{$from}) {
+        $entity->set($to, $record->{$from});
+        $updated = TRUE;
+      }
```
cory-skeers: List fields need a little different handling, and with this current set up will always end up with a value of `null` being compared to the record's value. Currently with how classrooms is set up, it's leading to the process thinking every node has been updated (though after the data is set and a save attempted, it's not recording a revision because there are no actual changes being made).

pyrello: We probably need to check if the `$record->{$from}` is an array or object and process it differently if so. What is the field that this pertains to?

cory-skeers: I was a bit off on what type of fields. We might need to do some work with multi-value fields as well, but specifically here it's entity reference fields that are having an issue. The `->value` getter returns null, and we either need to get each delta and `->value` on that, or use `getValue()` and work with the resultant array.

I pushed up the little bit that I set up to get it working with classrooms, where it checks if it's an entity reference field and handles differently if so. In this case, it's still doing straight comparisons even in the case of arrays (which is the reason for the odd `array_column(..., 'target_id')` stuff to get it a plain array).

cory-skeers: For classrooms, it's
`field_room_responsible_unit`, `field_room_features`, `field_room_technology_features `, `field_room_type`, and `field_room_scheduling_regions` that it effects.

pyrello: The way I imagined this happening is that in the `processRecord` step, the value on `$record` gets replaced with a `target_id`.

pyrello: Ah, I'm getting the issue now. I think for the field, we can check the field definition that is attached to it and see if its an ER field. So more forked if/then logic...

cory-skeers: Yep... I took a pass at it in https://github.com/uiowa/uiowa/pull/6370/commits/d299e5211efe4426b1f856a55690bb5c80dd3382, and it's working for classrooms. But it feels messy.
```
    foreach (static::$fieldMap as $to => $from) {
      if (!$entity->hasField($to)) {
        // @todo Add a message if a node doesn't have a field.
        continue;
      }
      // Check if it's an entity reference field, and manipulate
      // if necessary.
      if ($entity->getFieldDefinition($to)->getType() === 'entity_reference') {
        if (array_column($entity->get($to)->getValue(), 'target_id') != $record->{$from}) {
          $entity->set($to, $record->{$from});
          $updated = TRUE;
        }
      }
      // It's not an entity reference field, so we can grab
      // the value from it directly.
      elseif ($entity->get($to)->value != $record->{$from}) {
        $entity->set($to, $record->{$from});
        $updated = TRUE;
      }
    }
```

pyrello: Cool. I'm going to look at this and maybe make a few minor adjustments.

pyrello:
Added `imageUrl` field for facilities. Just the work to account for ER fields is now necessary to get this out of draft again.

docroot/modules/custom/uiowa_core/src/EntityItemProcessorBase.php
pyrello: It seems like we still might need to account for multiple values differently, but I think the code we have now is working for both use cases as is.

joewhitsitt (approved):
Nice work on this! Thanks for clearing up those todos.

Tested facilities again, with deletions, updates, etc. Image is there and again after removing it.

Classrooms, I tested off a sync from dev and prod (no notices). Temporarily disabled the form disables and tried to throw curveballs at the fields. Everything updated correctly.

pyrello:
Thanks for testing, @joewhitsitt!

Issue #5931, opened 2022-11-08
ISWF View Listing


joewhitsitt:
# Issue

Create searchable view listing for events by date, genre, and presenter.

# Proposed Solution

- [x] Try to use existing view display otherwise create a ISWF specific view, adding it to the site split.
  - Just filtered events that are not referenced as a parent. So, if there are multiple types this applies to, the view might need to look at certain relationships ("lowest layer")
- [x] As a visitor, I want to pick a date or date range (depending on what we discover while building) that I am available and see events that meet that criteria (on that date or in-progress).
- [x] As a visitor, I want to filter by genre
- [x] Filtered dropdown or autocomplete of person published (not hide on lists) that have been referenced by an event. This would require an entity reference view display to filter so that not all person content on the website is available for selection.
- [x] Ensure consistency of different event listing displays. Excludes event list block because that has an explicit option to change how it looks so that is handled differently.
- [x] If the "Performer" person type is not necessary to create the above views, remove it from the ISWF site split.
- [x] Based on today's conversation (12/05/2022), add an image field for visual conformity. To do this, you’ll want to use the default site and then turn on the event feature.
- [x] Look into genre dropdown opening when "Apply" is clicked.
- [x] Date filter should show events that have date ranges including the date selected.

briand44:
Ensure consistency of different event listing displays. Excludes event list block because that has an explicit option to change how it looks so that is handled differently.

joewhitsitt:
Should this be blocked until https://github.com/uiowa/uiowa/pull/5925 merges in? I think we assumed it would be done before this was pulled in.

joewhitsitt:
From standup, it sounds like this work can be added to #5925 and is not blocked

pyrello:
I made [some updates](https://github.com/uiowa/uiowa/pull/5925/commits/6c30f6f28e595ac4bead9d7def27ae0cc6a35999) and updated the checked boxes in the IS to reflect my changes.

pyrello:
Did we decide that we did not want the base event feature to include the ability to search by date or presenter?

pyrello:
Looks like we are going to need Better Exposed Filters to get a datepicker for the when field: https://www.drupal.org/docs/contributed-modules/smart-date/smart-date-and-views-including-fullcalendar-view#s-adding-a-date-picker-exposed-filter

pyrello:
[...]

PR #6372, opened 2023-04-21
Event Feature - Event Status

### LINKED ISSUE #5939: As a site editor, I can designate the EventStatus of a local event as Active, Cancelled, or Rescheduled.
# Story

As a site editor, I can designate the EventStatus of a local event as Active, Cancelled, or Rescheduled to inform viewers of the status of an event.

[https://schema.org/eventStatus](https://schema.org/eventStatus) - This is the schema.org property defined for the status of the event.

Screen Shot 2022-11-10 at 11 27 31 PM

# Proposed solution

- Add a new field_event_status field to the local event type at the default level.
- Attach field value to schema.org property "eventStatus".
- ~Does this need to carry over to all "child" or "subEvents" within the updated event???~ - No
- Output badge for rescheduled and cancelled using a single color badge on teasers and entity

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/5939

Screenshot 2023-04-21 at 3 23 28 PM
Screenshot 2023-04-21 at 3 22 16 PM
Screenshot 2023-04-21 at 3 22 28 PM

# How to test

## Test update hook

```
ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli node/7166/edit
```
Check that is has a value of Scheduled for the required field. The update hook itself should only run for sites that have the event split enabled. Feel free to test against another site like uiowa.edu

- Test default (event split), iswf, oniowa, admissions for event status functionality. Defaults to "scheduled" and is rendered on the full event node as well as the teaser. Inspect the full node source for event status schema information.
Confirm that https://github.com/uiowa/uiowa/pull/6372/commits/3075ff7639375484b17b08aa72d5ffd051c1cb95 should be added to allow articles/people the list/grid option on Admissions.


joewhitsitt:
Having troubles with the Admissions export. It wants to delete the complete item and replace it a split file containing some unfamiliar bits like `config_split_sequence_3qhidpsnp2a996q5`:

```yaml
adding:
  dependencies:
    config:
      - field.field.node.event.field_event_category
  third_party_settings:
    layout_builder:
      sections:
        -
          layout_settings:
            layout_builder_styles_style:
              - section_order_reversed
          components:
            config_split_sequence_3qhidpsnp2a996q5:
              weight: 2
            config_split_sequence_qvegmgm51mppsmnc:
              weight: 3
            config_split_sequence_asle1hi593u2nvkj:
              region: first
              configuration:
                layout_builder_styles:
                  - content_alignment_left
                  - list_format_list
...
```

joewhitsitt:
Hoping @pyrello can ease my mind a bit with this Admissions split. Is this from https://www.drupal.org/node/3279331 and the fact that Admissions has like four splits turned on?

pyrello:
> Hoping @pyrello can ease my mind a bit with this Admissions split. Is this from https://www.drupal.org/node/3279331 and the fact that Admissions has like four splits turned on?

Yes. Not sure why it is insisting on a partial split instead of a complete split though. I think I ran into something similar with events the last time I was working on it, debugged it, and then ended up either updating the the config manually or manually removing elements that shouldn't be there.

Maybe I opened this issue at the time? https://www.drupal.org/project/config_split/issues/3349737

e-marie-w:
I forgot, is the /events page supposed to not show super events?

joewhitsitt:
>I forgot, is the /events page supposed to not show super events?

I don't recall. I don't think this PR touches any of that functionality.

joewhitsitt:
#5931 looks like the idea of not showing parent events were an ISWF feature.

>Just filtered events that are not referenced as a parent. So, if there are multiple types this applies to, the view might need to look at certain relationships ("lowest layer")

Not sure what was decided in the end.

e-marie-w (approved):
Event status appears on node and teaser, as well as in schema information. Tested for default, ISWF, OnIowa, and Admissions. Also, grid option is now working for article and people listblocks on Admissions. 

I was worried about existing events, but they still work and will just require a status upon next edit. I have a few questions about the events feature in general, but they're unrelated to this functionality so I'm approving.

pyrello:
> I was worried about existing events, but they still work and will just require a status upon next edit. I have a few questions about the events feature in general, but they're unrelated to this functionality so I'm approving.

@joewhitsitt Is there a reason we are not doing an update hook on this to get our data matching the expected schema?

e-marie-w:
I assumed because there's only one existing event on a live site and only seven on internal prod domains.

joewhitsitt:
>@joewhitsitt Is there a reason we are not doing an update hook on this to get our data matching the expected schema?

No particular reason. 

>I assumed because there's only one existing event on a live site and only seven on internal prod domains.

This is probably part of it. Also that no other code is acting on whether or not it is set. Didn't think too much about it.

I will put something together.

e-marie-w:
>  looks like the idea of not showing parent events were an ISWF feature.

Okay cool I missed that happening. It's also part of default but can be changed if needed, because Admissions shows all events. 

Other not-event-status-but-event-feature thing that confused me is that ongoing, in-progress events don't display in the events listblock. I'm going to double-check I can recreate it and then create an issue for bug smash if so.

pyrello:
> This is probably part of it. Also that no other code is acting on whether or not it is set. Didn't think too much about it.
>
> I will put something together.

I didn't realize there were not very many records. Its probably not worth it. Just double-checking.

joewhitsitt:
I think it is a good habit to get into to. Going to see if I can lift and repurpose something real quick.

joewhitsitt:
Updated the test instructions for the update hook

e-marie-w:
Pulling down to test update hook.

e-marie-w (approved):
Update hook works as expected, thanks for adding that!

pyrello:
🤣 An update hook with batching!

PR #6367, opened 2023-04-20
Metatag token overrides

joewhitsitt:
Minimal slice to get regents-option sites setup with a few metatag/favicon overrides. If we do this often enough we might want to capture some of this through a UI.

# How to test

Example override:

`ddev drush @sitesgis.dev cset uiowa_core.settings uiowa_core.org_name "Alternative Organization Name"`
`ddev drush @sitesgis.dev cset uiowa_core.settings uiowa_core.favicon_assets_path "/sites/gis.sites.uiowa.edu/files/favicons_alt/"`

Example config delete:

`ddev drush @sitesgis.dev cdel uiowa_core.settings uiowa_core.org_name`
`ddev drush @sitesgis.dev cdel uiowa_core.settings uiowa_core.favicon_assets_path`

Example rsync/permissions set:
`ddev drush rsync -v @sitesgis.local:%files/favicons_alt @sitesgis.dev:%files/`
`ddev drush @sitesgis.dev ssh`
`cd sites/gis.sites.uiowa.edu/files`
`chmod ug=rwx,o=rx favicons_alt` or `chmod 775 favicons_alt`
`cd favicons_alt`
confirm you are in the favicons_alt directory!
`find . -type f -print0 | xargs -0 chmod 0644`

Reload the homepage and inspect the source. You should see the overridden values. You should also be able to see the favicon replacement after a `cr`

Example assets:
[favicons_alt.zip](https://github.com/uiowa/uiowa/files/11290568/favicons_alt.zip)

Screenshot 2023-04-21 at 10 45 20 AM

PR #6378, opened 2023-04-25
Metatag overrides followup

joewhitsitt:
Cover the event feature too

follow up from https://github.com/uiowa/uiowa/pull/6367


cory-skeers (approved):
Nice catch! Reviewed, and it looks good. Also did a quick code search, and don't see any other instances where this would need updated.

PR #6415, opened 2023-05-04
now.uiowa.edu: moved article teaser label to inline instead of above to fix icon positioning

bspeare:
I just noticed the icon position being a little off on now.uiowa.edu.   This fixes the label settings to be like our default config implementation. 

Before:

Screenshot 2023-05-04 at 10 10 37 AM

After:

Screenshot 2023-05-04 at 10 10 46 AM


joewhitsitt (approved):
reviewed, not tested

PR #6416, opened 2023-05-04
Fix to get iswf the event status update

joewhitsitt:
# To Test

- Run a sync on oniowa.uiowa.edu and see that sitenow_events_9001 doesn't run.
- Run a sync on iowasummerwritingfestival.uiowa.edu and see that the update hook runs and that all nodes are updated to have a "scheduled" event status. No errors.

```
ddev blt ds --site=oniowa.uiowa.edu && ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli /events 
```

### DIFF
--- a/docroot/modules/custom/sitenow_events/sitenow_events.install
+++ b/docroot/modules/custom/sitenow_events/sitenow_events.install
@@ -5,7 +5,6 @@
  * Install hooks for sitenow_events.
  */
 
-use Drupal\Core\Config\FileStorage;
 use Drupal\node\Entity\Node;
 
 /**
@@ -46,6 +45,9 @@ function sitenow_events_update_8002() {
 
 /**
  * Update sites using the event feature to introduce a new field we set value.
+ *
+ * Updated to remove field config import as only ISWF hasn't run
+ * and already has the config in place.
  */
 function sitenow_events_update_9001(&$sandbox) {
   $event_split = \Drupal::entityTypeManager()->getStorage('config_split')->load('event');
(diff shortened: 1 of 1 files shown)


bspeare (approved):
This looks good.  Off-topic, should the "Scheduled" badge be green instead?  It seems like the badge color should change based on the selection.   Here are our badge colors: https://uids.brand.uiowa.edu/components/detail/badge.html.

bspeare:
Screenshot 2023-05-04 at 10 30 55 AM

Screenshot 2023-05-04 at 10 31 29 AM

briand44:
I'm not sure we even need a badge for scheduled but I agree if we are going to have one it should be green.

cory-skeers:
> This looks good. Off-topic, should the "Scheduled" badge be green instead? It seems like the badge color should change based on the selection. Here are our badge colors: https://uids.brand.uiowa.edu/components/detail/badge.html.

I'm not sure we created a followup issue, but in grooming we talked about keeping it simple with a single color for now, and probably update later with more discussion (eg should "rescheduled" be its own color, match "canceled" or "scheduled," etc).

bspeare:
I agree, it is weird to see the badge for a scheduled event :)

joewhitsitt:
For the first slice we decided to keep it one color. During an EoD, outputting scheduled was also approved for the first slice.

>Output badge for rescheduled and cancelled using a single color badge on teasers and entity

https://github.com/uiowa/uiowa/issues/5939

I will create a separate issue for these enhancements.

cory-skeers:
> I will create a separate issue for these enhancements.

Thanks @joewhitsitt!

joewhitsitt:
https://github.com/uiowa/uiowa/issues/6417

Issue #5659, opened 2022-09-06
Searchable list of FM managed buildings


briand44:
# Story

As a Facilities site owner, I would like to display a searchable list of FM managed buildings so that visitors can quickly find a FM managed building.

# Additional Information

Should be able to be placed on a layout builder page so that editors can add additional contextual content.
Should output building number, building name and building abbreviation.
Should be able to sort by and search each of the fields being output.
Building name should render as a link to a page that will eventually contain additional building information for that specific building.

Endpoint - https://bizhub.facilities.uiowa.edu/bizhub/ext/buildings
Existing code - /docroot/sites/www.facilities.uiowa.edu/modules/custom/facilities_buildings/includes/facilities_buildings.buildings.inc
/docroot/sites/www.facilities.uiowa.edu/modules/custom/facilities_buildings/facilities_buildings.module
/docroot/sites/www.facilities.uiowa.edu/modules/custom/facilities_core/facilities_core.module (API call)

Existing site - https://www.facilities.uiowa.edu/buildings

# Proposed Solution

- [x] Building content type (webmasters and below do not need access to add/edit/delete)
  - [x] Building Name
  - [x] Building Abbreviation
  - [x] Building Number
- [x] Create views block utility table to filter, sort on.
- [x] Periodically add/update/delete from https://bizhub.facilities.uiowa.edu/bizhub/ext/buildings (e.g. Feeds, a Facilities type MAUI API integration running on cron)

joewhitsitt:
Not tested but I saw this module when thinking about our approach here. It would appear to provide us with a views table that could be filtered on (using design patterns we already have). https://www.drupal.org/project/views_json_source

briand44:
Grooming conversation... Build out a structure for this to store the information needed for the view display and create the searchable/sortable display.

pyrello:
Just documenting the rationale for considering a content entity vs some other mechanism of locally caching the data. We're going to be getting the data to display a building page from an API. Its inefficient to have to call the API every time we display the page, so we'll be using some sort of local caching to store the API data for a little while. The data from the API will likely need some processing to get it into the format that we want to display it in, so we'll need some logic that identifies fields and transforms them before they are stored in cache. Finally, we'll need one or more templates that tell the building page how to look and we'll need those templates to use variables that come from the cached API data. So we are going to essentially be providing some sort of hard-coded mapping either in a theme preprocessing function or in the templates themselves which describes what field gets displayed where and how.

[...]

PR #6422, opened 2023-05-05
SPPA - Unpublished content access fix

### LINKED ISSUE #6377: Webmaster unable to view some unpublished content
# Problem
> I started a web article today titled, “Alumni Spotlight: Joe Wilensky” but I can’t find it now. I’m looking in Content and selected the Type (Article) and Published status of “any” and also “unpublished.” But it’s not coming up. I had this same problem last week with another “draft” document.
- Janet at SPPA

image

# Steps to reproduce
1. https://sppa.uiowa.edu/admin/content?status=2 and note the first two items.
2. https://sppa.uiowa.edu/admin/people?user=jabell and masquerade as `jabell`.
3. https://sppa.uiowa.edu/admin/content?status=2 and note the first two items are now missing.

# Proposed solution
1. Investigate to understand what the problem is.
2. TBD

joewhitsitt:
Resolves #6377 

By enabling private_content, the node_grants functionality in core is enabled which handles access checking in places like views listings. That means additional code is necessary to apply the same permissions we take for granted (pun intended).  This was recently "lost" for SPPA when the bypass content access control was [removed from all sites](https://github.com/uiowa/uiowa/issues/5659#issuecomment-1374111499). This node grants system is handled on node save though which means for SPPA to use this, all nodes (or a least the unpublished ones) need to be re-saved.

https://git.drupalcode.org/project/private_content#warning
https://api.drupal.org/api/drupal/core%21modules%21node%21node.module/group/node_access/8.8.x

# To Test

```
ddev blt ds --site=sppa.uiowa.edu && ddev drush @sppa.local uli --uid=6 "admin/content?&status=2"
```

Masquerade as other users with different roles and see how they are affected by the change.

Note: jabell can't see mccnnell's [unpublished page](https://sppa.uiowa.ddev.site/current-students) in the content overview. and mccnnell can't see jabell's unpublished articles as the `view any unpublished content` permission is not checked against in the node grants method.

### DIFF
--- a/docroot/sites/sppa.uiowa.edu/modules/sppa_core/sppa_core.install
+++ b/docroot/sites/sppa.uiowa.edu/modules/sppa_core/sppa_core.install
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * @file
+ * Update hooks for sppa_core.
+ */
+
+use Drupal\node\Entity\Node;
+
+/**
+ * Update all unpublished nodes to set grant access.
+ */
+function sppa_core_update_9001(&$sandbox) {
+
+  // Process all the unpublished nodes.
+  if (!isset($sandbox['total'])) {
+    $query = \Drupal::entityQuery('node')
+      ->accessCheck(FALSE)
+      ->condition('status', 0)
+      ->execute();
+
+    $sandbox['total'] = count($query);
+    $sandbox['current'] = 0;
+    $sandbox['query'] = $query;
+
+    if (empty($sandbox['total'])) {
+      $sandbox['#finished'] = 1;
+      return;
+    }
+  }
+  // Loop through the same nodes in batches.
+  $nodes_per_batch = 25;
+  $batch = array_slice($sandbox['query'], $sandbox['current'], $sandbox['current'] + $nodes_per_batch);
+  if (empty($batch)) {
+    $sandbox['#finished'] = 1;
+    return;
+  }
+
+  foreach ($batch as $nid) {
+    // Load the node, add a revision message, save.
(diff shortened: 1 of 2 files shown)


joewhitsitt:
Additional checks for who should be able to see unpublished "private" in the content overview and elsewhere should be checked as these are now applying together where they might not have been before 😬

pyrello:
Does this seem like functionality that belongs in the module? Should we turn this into a patch?

pyrello:
> Should we turn this into a patch?

[No commits](https://git.drupalcode.org/project/private_content/-/commits/8.x-2.x) on that module in about 2 years. Maybe we just need to see if there is a better way of handling this.

pyrello:
Pulled this down to test.

pyrello:
This seems to work as described. It seems like only a partial fix since the user in this case would have every expectation of being able to see unpublished content created by other users.

joewhitsitt:
Might be able to do something similar to https://www.drupal.org/project/drupal/issues/2971902#comment-13612843 to cover the same bases

Also saw https://www.drupal.org/project/nodeaccess for "more sophisticated access control"

joewhitsitt:
With https://github.com/uiowa/uiowa/pull/6422/commits/0ac67182260def491e099ad3d5daa50743a09a06, took the code from https://www.drupal.org/project/drupal/issues/2971902#comment-13612843 which seems to be a more complete solution based on their previous access.

pyrello (approved):
This appears to fix the issues that have been encountered on this site without exposing the private content to unauthorized users.

PR #6424, opened 2023-05-05
Badge visibility

### LINKED ISSUE #6417: Update event status visibility or badge color based on value
# Story

As a site visitor, if an event is scheduled, I am not sure I need to see that or if I should see it, that it should be green to indicate that it is open and continuing as planned.

# Proposed solution

- [ ] unset the field from render if the event status is scheduled
- [ ] show different colored badge than orange for rescheduled (gray?)

GaryRidgway:
This PR solves #6417. It hides event status if it is scheduled, makes it orange if cancelled, and green if rescheduled.

# How to test

1. Sync admissions and iswf.
2. go to the /events of both sites
3. edit some events to give them cancelled and rescheduled
4. ensure that the badges properly show as defined in the synopsis of this PR.
5. Also ensure that, if there is a location, the badge is above it.


e-marie-w:
Pulling down to test.

e-marie-w (approved):
Badges are correctly showing under the date and are red if cancelled, green if rescheduled, and not present for regular scheduled events. I tested admissions, iswf, and oniowa since they all have the split. I also turned the split on for default and tested it there as well to be extra thorough. Good work!

PR #6438, opened 2023-05-09
customizable event categories

### LINKED ISSUE #6364: SiteNow event categories
# Story

As an OnIowa site owner using the SiteNow events feature, I would like to be able to have customizable event categories.

# Additional Info

We discussed bringing Admissions and ISWF along for the ride.

- Existing categories for OnIowa
     Screen Shot 2023-01-24 at 10 59 10 AM

# Proposed Solution

GaryRidgway:
Resolves #6364 
~~- [x] Update hook to create the "Event categories" taxonomy if it does not exist.~~
- [x] Add the Event categories field to the Event content type and assign the taxonomy to it.
- [x] Decide on a place for the field to go. In the main area of the form? in the relationships dropdown? somewhere new?
- [x] Hook to move relevant terms to the event categories taxonomy (eg: The "event_type" terms from ISWF to "Event categories").
    ~~- [x] This also includes reworking how events reference genres I think?~~
- [x] Set perms for webmaster/publisher to have add/edit/delete permissions.
- [x] remove splits from ISWF and Admissions to consolidate implementations in the event module
- [x] Ensure that admissions and ISWF still function in the desired way after removal of split functionality.
~~- [x] Should we allow for the renaming of the event categories field? (eg: ISWF might want it to be called "Event type")~~
- [x] display the event categories at the bottom of events
    - [x] This means removing the complete splits for displays for admissions and iowasummerwritingfestival and adding it as a partial and ensuring that the pages still function the intended way.
- [x] Set Perms

After all this is done, new terms should be able to be added/edited by webmasters/publishers.

## To Test

Gremlin check before sync
```
ddev drush @iowasummerwritingfestival.local sql-drop -y && ddev drush @iowasummerwritingfestival.local cr && ddev blt ds --site=iowasummerwritingfestival.uiowa.edu
```

1. Pull main and observe that ISWF has an event type field and it is populated with many options.
2. Pull this branch and do all the setup stuff.
3. Sync ISWF and observe that a taxonomy has been created called 'event_categories' with all the options previously available in 'event_type'
4. Observe that events have the 'event_categories' field and that the 'event_type' data has been migrated in to it properly. You can use ISWF's live site to check what they were previously.
5. Observe that there are only slight visible changes to ISWF site. Event categories should be the only change, and the events display should remain roughly the same.
6. Sync Oniowa and Admissions
7. Observe that they do not change from their live version.
8. ???
9. Profit.


e-marie-w:
Got some clarity on this. We are not creating a new event category taxonomy, what we want to do instead is move the files for it from the admissions site split to the event feature split. We'll need to declare it as part of that split and resave it. 

We also need to keep ISWF as-is for now and make sure it isn't changed.

GaryRidgway:
From the ISWF meeting, it was decided that we were merging ISWF "Event type" in
to "Event categories", right? and then perhaps using a hook to migrate data?

e-marie-w:
> From the ISWF meeting, it was decided that we were merging ISWF "Event type" in to "Event categories", right? and then perhaps using a hook to migrate data?

Yes

GaryRidgway:
I started working on removing the full split from ISWF but I lost the genre display, the config will have to be adjusted.

bspeare:
@GaryRidgway is there a description for the 9002 update hook?  Or is that not needed here?   `iowasummerwritingfestival_core   9002`

bspeare:
@GaryRidgway I'm seeing a few errors after running `ddev blt validate:phpcs` - do you want to check that out?

GaryRidgway:
> @GaryRidgway is there a description for the 9002 update hook? Or is that not needed here? `iowasummerwritingfestival_core 9002`

That is a good catch, I will go through and add that.
> @GaryRidgway I'm seeing a few errors after running `ddev blt validate:phpcs` - do you want to check that out?

I will, thank you!

GaryRidgway:
@bspeare I have fixed the issues you outlined. Thanks again!

joewhitsitt:
pulling down to test

joewhitsitt:
There is some more code that could be cleaned up https://github.com/uiowa/uiowa/blob/62697327c7cd07aa052086aa70a5f12ec068b02d/docroot/sites/iowasummerwritingfestival.uiowa.edu/modules/iowasummerwritingfestival_core/iowasummerwritingfestival_core.module#L36-L89

Not blockers, but I noticed:

- we haven't been setting the revision author lately on these update hooks:

Screenshot 2023-05-24 at 2 30 36 PM

- should we remove the link on the node to these taxonomy terms or rabbit hole it all together? events will be sorted by created date regardless of their event date. add the default RSS to this and it could be pretty worthless to show these.
 
Screenshot 2023-05-24 at 2 31 23 PM

joewhitsitt (approved):
Looks good.

I think if event categories being public becomes a problem, we can unlink/rabbit hole them down the road.

GaryRidgway:
Thanks for all your work on testing this @joewhitsitt!

PR #6445, opened 2023-05-11
Removed research areas label config.

makurj:
Resolves "follow-up task to remove split config related to research areas on pharmacy"


e-marie-w (approved):
Code review looks good! I also checked that the site syncs successfully and that the recent label changes were still in effect.

Thanks for JDI-ing this!

PR #6448, opened 2023-05-12
[Iowa Now] Basic news type

### LINKED ISSUE #6423: Ability to add news that isn't one of the existing types
# Story

As an Iowa Now site owner, I would like to be able to add Employee News or potentially other news in the future that is not one of the existing article types so that we can manually tag those and have them show where we want.

# Additional Info

>Can we create a new article type called “Employee News” for adding content to show up exclusively in the “News you can use section”? Currently, we tried using the “News Feature” content type, and then using the “employee news” tag, but the content then shows up in BOTH the “featured headlines” section and the “news you can use” section, and we don’t know how to eliminate the appearance under “featured headlines.”

# Proposed Solution

-  add a new option for "basic" or "none" that doesn't auto add a tag
-  update help text

cory-skeers:
Resolves #6423 

# How to test

`git checkout now-basic-news && ddev blt ds --site=now.uiowa.edu && ddev drush @now.local uli`

- Create new articles and edit existing articles.
- Try out the "Basic" article type, as well as the existing article types.
- For "Basic": 
  - no additional fields (embargo info, original publication date, or points of contact) should be available
  - No tags should be auto-added
  - Any tags you manually add should remain untouched
- For other types:
  - News Feature/Achievements should get points of contact field
  - ui spotlight and in the news should get original publication date
  - News feature should get embargo info (and if it is filled, article should stay in draft state)
  - A corresponding tag related to the type should be added on save
- changing types and re-saving:
  - If a tag existed for the original type, it should be removed
  - If the new type has a corresponding tag (ie is not Basic) it should be added
  - If the new type is Basic, nothing should change in tags

### DIFF
--- a/config/sites/now.uiowa.edu/field.field.node.article.field_article_type.yml
+++ b/config/sites/now.uiowa.edu/field.field.node.article.field_article_type.yml
@@ -12,7 +12,7 @@ field_name: field_article_type
 entity_type: node
 bundle: article
 label: 'Article Type'
-description: 'A corresponding tag will be added/removed based on the article type selected.'
+description: 'A corresponding tag will be added/removed based on the article type selected. <em>Basic</em> article types can use any tags, and will not receive any auto tagging.'
 required: true
 translatable: false
 default_value:
--- a/config/sites/now.uiowa.edu/field.storage.node.field_article_type.yml
+++ b/config/sites/now.uiowa.edu/field.storage.node.field_article_type.yml
@@ -23,6 +23,9 @@ settings:
     -
       value: achievements
       label: Achievement
+    -
+      value: basic
+      label: Basic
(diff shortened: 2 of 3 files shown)


joewhitsitt:
pulling down to test

joewhitsitt:
I don't think this works quite right if you go from something like achievement > basic > spotlight. the achievement tag is retained alongside spotlight.

update: actually, from the test instructions, it seems to be how the functionality is setup, checking old to new. the switch from one to basic and then to a different one shouldn't happen to frequently and it will eventually clean up on further saves.

joewhitsitt (approved):
Works as described, but noting https://github.com/uiowa/uiowa/pull/6448#issuecomment-1548448001

PR #6449, opened 2023-05-12
[Facilities] Projects content type

### LINKED ISSUE #6398: Facilities content type for projects
# Story

As a web team, we would like to create a content type and fields to store project data that Facilities needs the ability to edit.

# Additional Info

Facilities will not be creating project nodes but they will need the ability to add additional information to existing projects.

We should follow our best practices for setting up a content type and we should set permissions accordingly.

**Info they need ability to add/edit**
Project Image - single image media field
What's New - text area currently filtered html
Project Description - text area currently filtered html
Webcam - link field - title + url - title defaults to "View webcam"

cory-skeers:
Resolves #6398 

# How to test

`ddev blt ds --site=facilities.uiowa.edu`
As administrator, create some Project content, test the fields function as they should

- Project Image - single image media field
- What's New - text area currently filtered html
- Project Description - text area currently filtered html
- Webcam - link field - title + url - title defaults to "View webcam"
- SEO Settings setup similar to other content types
- Schema.org coverage as https://schema.org/OrganizeAction which I thought suited this content type better than https://schema.org/Project

Masquerade as other roles. They should be able to edit existing project and change the above fields, but they should not be able to create new projects or delete existing projects

Check that the general bases have been covered for creating a new content type (eg https://wiki.uiowa.edu/display/drupaldev/Content+type+creation+notes)


joewhitsitt:
pulling to test/review

joewhitsitt:
Immediately noticing that metatag nor schema.org seem to be setup for this.

Cory said in slack:
>For the content type, I think everything is set up (permissions, fields, etc) except for potential Schema setup. Open to setting that up for current fields now, or when all fields are in place.

I think we should continue to cover with both with some basics. I'll see about adding that to this PR.

joewhitsitt:
Wasn't able to get to this today. #not-squatting

joewhitsitt (approved):
A bit biased, but I think this is a solid foundation for the upcoming work and since the permissions are set so webmasters and below can not create projects, there is little risk in this going to PROD

joewhitsitt:
Percy isn't happy...

```
Encountered an error uploading snapshot: About page with search input visible
Error: The name of each snapshot must be unique, and this name already exists in the build: 'About page with search input visible' -- You can fix this by passing a 'name' param when creating the snapshot. See the docs for more info on identifying snapshots for your specific client: https://percy.io/docs
```

@bspeare have you ever seen this before?

bspeare:
Maybe a momentary glitch?

PR #6450, opened 2023-05-15
Add click-a11y.js from UIDS4

### LINKED ISSUE #6138: clicka11y.js is not retriggered on AJAX calls
# Problem

Discovered in https://github.com/uiowa/uiowa/pull/6122, card and banner are using docroot/themes/custom/uids_base/uids3/assets/js/click-a11y.js to implement the click on card/faux button. This works for page load when the JS is being attached to the component through libraries, but because it isn't a drupal.behaviors it isn't getting re-triggered when an ajax views filter or page is fired.

# Steps to reproduce
https://sandbox.prod.drupal.uiowa.edu/people-directory on the first block, click the pager, the card is no longer clickable besides the card title.

# Proposed solution

- One possible solution is to wrap click-a11y.js in a function that can then be called through a drupal.behavior.

<details>
<summary>click-a11y.js</summary>

```js
function triggerClickA11y() {
  const items = document.querySelectorAll('.click-container:not([data-uids-no-link])');
  const link_elements = ['.click-target'];
  const RIGHT_BUTTON_CODE = 2;

  Array.prototype.forEach.call(items, item => {
    let up, down, link, i;
    // Loop through options and break on the first match.
    for (i = 0; i < link_elements.length; i++) {
      if (link = item.querySelector(link_elements[i])) {
        break;
      }
    }

    // If we have a match, attach behaviors.

    if (!link) {
      return;
    }
    item.style.cursor = 'pointer';
    item.onmousedown = (e) => {
      if (!ignoreClick(e)) {
        down = +new Date();
      }
    }
    item.onmouseup = (e) => {
      if (!ignoreClick(e)) {
        up = +new Date();
        // Trigger click event if the duration is short enough.
        if ((up - down) < 200) {
          link.click();
        }
      }
    }
  });

  /**
   * Ignore right button and <a> or <button> clicks.
   */
  function ignoreClick(e) {
    return e.button === RIGHT_BUTTON_CODE || ['A', 'BUTTON'].indexOf(e.target.nodeName) !== -1;
  }
}

triggerClickA11y()
```
</details>

Then in a separate file also attached to the component:

```js
(function ($, Drupal) {

    "use strict";

    Drupal.behaviors.clickA11y = {
            triggerClickA11y();
        }
    }
});
```

There might be multiple calls then that would have to be addressed.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6138

# How to test

- Review code
- Test functionality 
- Confirm unlinked cards/banners are not linked
- Confirm cards are still linked after using views exposed filters

```
ddev blt ds --site=housing.uiowa.edu
```
Go to /residence-halls and use the filters and verify that the cards are still clickable.  Compare with prod version. 
```
```
- Test click behavior on banners /https://sandbox.uiowa.ddev.site/banners-imagesvideos-w-black-and-white-gradients
- Test click behavior on cards https://sandbox.uiowa.ddev.site/timeline, https://sandbox.uiowa.ddev.site/cards


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Yay! good to get this fixed. also tested with some pagers. all seem to checkout

PR #6464, opened 2023-05-19
[Event feature] Attendance Mode field addition

### LINKED ISSUE #5940: As a site editor, I can designate the AttendanceMode of a local event as Online, Onsite, or Mixed/Blended.
# Story

As a site editor, I can designate the AttendanceMode of a local event as Online, Onsite, or Mixed/Blended to inform viewers of the attendance mode for the event.

[https://schema.org/eventAttendanceMode](https://schema.org/eventAttendanceMode) - This is not approved yet by schema.org but is available as a property in schema assignment and would be helpful to prepare for an anticipated approval. Useful for ISWF's mixed festivals (one Onsite, one Online).

Screen Shot 2022-11-10 at 11 27 25 PM

# Proposed solution

- Add a new field_event_attendance_mode field to the local event type at the default level as required.
- Attach field value to schema.org property "eventAttendanceMode".
- ~Does this need to carry over to all "child" or "subEvents" within the updated event???~ - this is out of scope for this issue
- Update hook to set this field for existing events.

### LINKED ISSUE #6463: Display virtual/address fields based on event type (in person, hybrid, virtual)

cory-skeers:
Resolves #5940 
Resolves #6463 

# Updated tests

Check that it has a clean sync and update

On any site with the event feature:
Edit existing or create new events
Set attendance mode to hybrid and add both a location and a virtual event info; save
Edit, and change attendance mode to in-person and save; see that the virtual event info has been removed
Repeat above, but instead change attendance mode to online and save; see that the location info has been removed

# How to test

`git checkout event-attendance-mode && git pull && ddev blt ds --site=default && ddev drush @default.local config-split:activate event && ddev drush @default.local uli`
Add an event using the new Attendance mode field
Check that if the attendance mode:

- is "in person" the location field shows but not the virtual
- is "online" the virtual shows but not the location
- is "hybrid," both virtual and location fields show

Check that the teasers and node display do not show the attendance mode
View the source, and check that the attendance mode is appearing in the meta markup

Test the sites that have the split activated, and that the update hook runs without errors
Check out existing events and see that they have an appropriate attendance mode set (see https://github.com/uiowa/uiowa/pull/6464#issuecomment-1557395672 for added context as of 05/22)
Check the same as on default--the values don't display directly, but update the meta markup
`ddev blt ds --site=admissions.uiowa.edu`
`ddev blt ds --site=oniowa.uiowa.edu`

Weigh in on https://github.com/uiowa/uiowa/pull/6464#issuecomment-1557542989 and if update hook should move to `sitenow_events` or stay in the profile.


cory-skeers:
Note: Display values for the event attendance mode changed slightly from the original issue description. It looks like in the course catalog, the equivalent options are "In-person," "Online," and "Hybrid," and unless there's a reason to use other labels, I think it's beneficial to keep it consistent. (https://myui.uiowa.edu/my-ui/courses/dashboard.page)

Machine names were generated straight from schema.org creating the field.

joewhitsitt:
careful with that update hook, iswf had more than one batch and it tried to reimport the field on the next run and said it already existed.

https://github.com/uiowa/uiowa/commit/29884e727ba1e7e79b99164dc42519bfb0040f12

because of it, we separated it from the sandbox here https://github.com/uiowa/uiowa/pull/6438/files#diff-05c542c605260bfcd5fcdd8c1b698930c94e5cbdc39d9e9826727b131adf19cf

also, should it move to sitenow_events?

cory-skeers:
Couple notes discovered while working on the update hook:

- Looks like ISWF doesn't have any virtual info filled out, so the update hook just gets the value set to in-person for all.
- Admissions does have one that has a virtual and location, and the update hook sets that event to hybrid, the rest to in-person.
- Oniowa doesn't have any events.

Should have our bases covered should the above change at all, but not much actually needed in the update hook beyond getting a value properly set.

cory-skeers:
> because of it, we separated it from the sandbox here https://github.com/uiowa/uiowa/pull/6438/files#diff-05c542c605260bfcd5fcdd8c1b698930c94e5cbdc39d9e9826727b131adf19cf

Ah, interesting. I'd planned on just adding a check if the config had already been added to skip over that on subsequent runs, but I like the separation. If the update hook gets moved, I'd want to change this (if it stays where it's at, I'd prefer to keep it contained to one for brevity, and sites that don't have the split will exit out early once).

Speaking of...
> also, should it move to sitenow_events?

I'm neutral on this. If things were split as `uiowa_events` and `sitenow_events`, where one dealt with the event feature solely, I'd want to move it for sure. Since the module is platform-wide as well, I figured we'd toss the update hook in with the others on the profile.

cory-skeers:
Set back to draft--double checking some config

makurj:
Pulling down to test.

makurj:
Noticed that if you switch from in-person with address to online, it retains the physical address info and displays it on the node, unless you manually remove all the address data. Same behavior if you switch from online to in-person, links are retained unless manually removed.

Huddle with Cory summary: schema markup for `eventAttendanceMode` is not supposed to be on the teaser when added through the list block. Approving, working as expected.

makurj (approved):
Tested, working as expected.

joewhitsitt:
>Noticed that if you switch from in-person with address to online, it retains the physical address info and displays it on the node, unless you manually remove all the address data. Same behavior if you switch from online to in-person, links are retained unless manually removed.

Did we want to do anything about this? Logic to only display content if the right selection is made? unset value on presave?

briand44:
We probably should do one of those. I would be fine if we handled it on the display end if that is easier but either option would be fine.

cory-skeers:
> > Noticed that if you switch from in-person with address to online, it retains the physical address info and displays it on the node, unless you manually remove all the address data. Same behavior if you switch from online to in-person, links are retained unless manually removed.
> 
> Did we want to do anything about this? Logic to only display content if the right selection is made? unset value on presave?

I'd lean toward presave; if data isn't going to be displayed, it'd be nice to be able to clean that up. And if someone switched it in a way that lost info by accident (or thought it was going to need changing, but needs to change back), there would still be the revisions record to use, up to a point.

cory-skeers:
Setting to draft to resolve conflicts and take a look at updating for https://github.com/uiowa/uiowa/pull/6464#issuecomment-1561861533

cory-skeers:
> careful with that update hook, iswf had more than one batch and it tried to reimport the field on the next run and said it already existed.
> 
> [29884e7](https://github.com/uiowa/uiowa/commit/29884e727ba1e7e79b99164dc42519bfb0040f12)
> 
> because of it, we separated it from the sandbox here https://github.com/uiowa/uiowa/pull/6438/files#diff-05c542c605260bfcd5fcdd8c1b698930c94e5cbdc39d9e9826727b131adf19cf

Ended up separating the update hook out for the opposite reason. No matter how I tried to set it up, was running into an issue where some nodes were being accessed prior to the config import, leading to errors. Splitting it into two update hooks resolved it.

bspeare:
Testing this out.

bspeare:
Tested this out and the presave works great.  No opinion on update hook location.

cory-skeers:
> Tested this out and the presave works great. No opinion on update hook location.

Thanks @bspeare!

PR #6471, opened 2023-05-23
Add Easy Breadcrumb to Periodical Split

### LINKED ISSUE #6013: Periodical feature - Breadcrumbs without menus
# Motivation

Our current system requires nodes to be in the menu to have a breadcrumb trail. The periodical feature was designed to work without using the menu system. Articles and issues would benefit from the context of breadcrumbs to help the user navigate through their structure.

# Proposed solution

Investigate using Easy Breadcrumbs or another solution to provide breadcrumbs without menu links. Ensure that the new solution doesn't conflict with how our system is currently set up.

joewhitsitt:
This PR introduces the easy_breadcrumb module to the periodical split (e.g. iwp91stmeridian) to provide faux breadcrumbs for non-menu based URL patterns.

Resolves https://github.com/uiowa/uiowa/issues/6013

# To Test

```
ddev composer install && ddev blt ds --site 91stmeridian.iwp.uiowa.edu && ddev drush @iwp91stmeridian.local pathauto:aliases-generate update canonical_entities:node && ddev drush @iwp91stmeridian.local cr && ddev drush @iwp91stmeridian.local uli issues/volume-6-issue-1-61-spring-2008/last-bird-burial-master
```

See that it has a breadcrumb trail based on the URL pattern where `issues` is unlinked and the `volume 6...` is linked.

Go to https://iwp91stmeridian.uiowa.ddev.site/node/46/edit
Provide a menu link that is nested under the about page. Save.
View the page and see that the path based on menu is as expected.

Go back and edit, override the generated URL alias with a complex fake nested path. Save.
View the page and see that it still respects the menu settings for the breadcrumb.

# Deployment Instructions

run `drush pathauto:aliases-generate update canonical_entities:node` on 91stmeridian that uses the feature


e-marie-w:
Pulling down to test

e-marie-w:
Getting error `The import failed due to the following reasons: Unable to install the <em class="placeholder">easy_breadcrumb</em> module since it does not exist. ` Which is weird to me because `easy_breadcrumb.settings` is listed as config to be created.

joewhitsitt:
I might have rushed this a bit. I'll take a look.

bspeare:
Try `ddev composer install` @e-marie-w

e-marie-w:
That worked!! My bad. Thanks @bspeare!

e-marie-w (dismissed):
Tested on iwp91stmeridian, works great! Articles get unlinked breadcrumb trails matching the URLs. Pages with a menu link have breadcrumbs that reflect the menu structure, regardless of any complex manually created URL path.

pyrello:
Only concern here is that the breadcrumb is unlinked and part of the reason for doing this was to allow the user a way to navigate inside the issue:
> Articles and issues would benefit from the context of breadcrumbs to help the user navigate through their structure.

joewhitsitt:
Yeah, that was one of my concerns https://github.com/uiowa/uiowa/issues/6013#issuecomment-1559837453 but assumed that was already known.

@pyrello Do you think we should pivot to provide content at these paths?

pyrello:
> Yeah, that was one of my concerns [#6013 (comment)](https://github.com/uiowa/uiowa/issues/6013#issuecomment-1559837453) but assumed that was already known.
> 
> @pyrello Do you think we should pivot to provide content at these paths?

My apologies that I missed this when you were demoing it. Am I understanding from the linked comment that if there is a valid upstream path that the fragments do get linked?

joewhitsitt:
Untested, but I imagine providing a contextual based view at those paths will cause the breadcrumbs to link.

joewhitsitt:
Looks like this could be solved by creating a page /issues (view page or basic page) and changing the publication issue pathauto pattern from "issue/..." to "issues/..."

A contextual based view doesn't make sense here since there is actually content for the issues path component, it just wasn't lined up.

Screenshot 2023-05-24 at 1 33 12 PM
Screenshot 2023-05-24 at 1 32 49 PM

joewhitsitt:
Re-requesting review after updating the periodical issue alias pattern to be plural. This creates a breadcrumb path back from the issue article. If an /issues page is created and assigned to the menu, the entire breadcrumb will be linked.

Updated the test instructions to include a bulk update of aliases.

bspeare:
Is this intended that the breadcrumbs on people now behave differently?   

Screenshot 2023-05-25 at 8 47 45 AM

Vs

Screenshot 2023-05-25 at 8 48 29 AM

joewhitsitt:
>Is this intended that the breadcrumbs on people now behave differently?

No I guess not. I can easily turn off the last segment option to be more consistent with our current offering.

Screenshot 2023-05-25 at 9 38 48 AM

bspeare (approved):
I tested out the scenarios again and they worked!

joewhitsitt:
aliases have been updated on PROD

PR #6478, opened 2023-05-25
[Bug] two column layouts

bspeare:
Since https://github.com/uiowa/uiowa/issues/3585 is still happening, I'm still curious about trying this out: https://github.com/uiowa/uiowa/issues/3585#issuecomment-941358866 to see if it makes a difference. 

# How to test

- Review code 
- Sync a site and verify that the two column layout is still an option and that there are no issues swapping the layout classes.   
- Adjust the layout definition to verify that it is using the class defined in `uiowa_core`


pyrello:
Pulling this down to test.

pyrello:
Code updates looked good in my code review. However, I'm unable to test at the moment because there is an issue preventing me from accessing a test site.

pyrello:
Actually pulling this down now to test out.

cory-skeers:
> Actually pulling this down now to test out.

I neglected to assign myself, and pulled it down.

Will hold off on approval, because I don't think the extra eyes/testing would hurt.

pyrello:
Only thing I'm not sure about is how to test whether a cache rebuild is necessary to pick up the new file. Doing a `blt ds` includes a cache rebuild. When I tried to just sync the DB, I got some errors that I'm not immediately clear on how to resolve. Otherwise, this seems to be good to go.

cory-skeers:
> Only thing I'm not sure about is how to test whether a cache rebuild is necessary to pick up the new file. Doing a `blt ds` includes a cache rebuild. When I tried to just sync the DB, I got some errors that I'm not immediately clear on how to resolve. Otherwise, this seems to be good to go.

I did a `ds` with main checked out, then pulled this branch without a cache rebuild, and it seemed to pick up the new files without an issue.

pyrello (approved):
This seems good and doesn't need a cache rebuild to pick up the change. Thanks, @cory-skeers for help figuring out how to test the CR part.

pyrello:
If this works and we decide to roll with it long term, is there any additional cleanup that needs to be done after this, @bspeare?

bspeare:
@pyrello if this works, we would expect to see the three column layout to be the next layout to disappear since that uses the core layout class.   The only additional cleanup task could be to change the label or icon so you can tell the difference between the core and `uids_base` layouts for developers.

joewhitsitt:
I haven't tested this, but think it is worth trying out in PROD. Should it be scheduled overnight just in case though?

bspeare:
@joewhitsitt Yes, I do like the idea of extra caution for an overnight deployment.

bspeare:
@joewhitsitt If Cory and Sean both think that a cache rebuild is not necessary - are there other things that you want to test to avoid an overnight deployment?

PR #6483, opened 2023-05-26
Facilities - maps for building pages

### LINKED ISSUE #6444: Facilities map on building pages
# Story

As a Facilities website owner, I would like our Facilities map to be displayed on each of our building pages since that map is more up to date than Concept3D map.

# Additional info

It sounds like we can pass in a lat/long (from bizhub api for each building) for a center point and render the map embed for each building. Joe put together a code example https://codepen.io/joewhitsitt/pen/yLRjgOJ.

# Proposed Solution

- [x] Programmatically construct iframe object from lat/long in Bizhub API
- [ ] Factor in zoom level
- [ ] Not sure if we need to use css in codepen
- [x] Place at bottom of building pages like in mockup

e-marie-w:
Resolves #6444 

# How to test
- `ddev blt frontend`
- `ddev blt ds --site=facilities.uiowa.edu `
- `ddev drush @facilities.local fm-buildings` (updates buildings to have lat/long coordinates, can take awhile to run)
- `ddev drush @facilities.local uli` 
- Check various buildings, see that map is rendering in iframe as expected. 
- Check iframe does not appear on buildings without lat/long values - example: https://facilities.uiowa.ddev.site/building/0013.


joewhitsitt:
will pull this down to test

joewhitsitt:
@e-marie-w would you be open to changing the map block machine name/label to be more FM specific in case we develop something else later on for the platform that would use C3D or whatever the UofI is using at the moment?

e-marie-w:
> @e-marie-w would you be open to changing the map block machine name/label to be more FM specific in case we develop something else later on for the platform that would use C3D or whatever the UofI is using at the moment?

Yes, I can change it, that's a good idea.

joewhitsitt:
Also, I'll put a vote in to change the zoom to 8 as a happy medium for desktop/mobile and the different size/shaped buildings such as kinnick stadium (0042), football facility (0395), westlawn (0003), lagoon shelter house (0060)

bspeare:
@e-marie-w I updated this to use our new media classes which I had forgotten about: https://github.com/uiowa/uids/blob/4.x/src/components/media/media.scss#L50.

joewhitsitt:
checking the latest changes

bspeare:
I added `ddev blt frontend` to the testing instructions.

e-marie-w:
Thank you @bspeare and @joewhitsitt!

joewhitsitt (approved):
Looks like the box got smaller on mobile which is good that the zoom changed. Good work!

I didn't pay a lot of attention to the layout builder stuff thinking that it will continue to be worked on.

joewhitsitt:
Just realizing after the fact that using geofield for the lat/lon fields would provide some validation and position us to do other stuff with these values later...

Screenshot 2023-06-02 at 2 09 13 PM

e-marie-w:
I didn't think of checking for existing modules, that looks neat.

PR #6499, opened 2023-06-01
Sprint 165 Provisions

joewhitsitt:
Only thing unusual here is `iihr.uiowa.edu` on `uiowa02` because we have coverage for `*.iihr.uiowa.edu` on that docroot.


cory-skeers (approved):

joewhitsitt:
Hey, https://github.com/uiowa/uiowa/pull/6499/commits/eefe44b124021317c08c36e08fe675e258c8744d means that blt umc is working again for the team? adding the signage group was enough for uihcsn01?

PR #6501, opened 2023-06-01
event attendance hotfix

joewhitsitt:
This _is_ an issue for all sites using the events feature. It renders once the field is "set" with either true/false. In cases where it wasn't showing yet, the node had not been resaved (no update hook was done) and the field was not set. This seems like an appropriate fix given that the field_block in preprocess_block is doing something similar.

https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_events/sitenow_events.module#L203-L205

Screenshot 2023-06-01 at 12 52 52 PM

# To Test

On `main` sync iswf and oniowa. save some events. some with the boolean checked some without. go to /events and see the no values.

Pull this PR down and do some `cr` see that the "no" values don't show.


cory-skeers (approved):
Thanks for the quick fix!

Pulled with iswf and confirmed it works.

Issue #6494, opened 2023-05-31
Probe reusable locations for OnIowa events


briand44:
# Story

As a web dev team, we would like to better understand options for our On Iowa partners for ways they can reuse event locations so that they are not recreating the same location many times for different events.

# Additional Info

https://github.com/uiowa/uiowa/issues/6402

They need lat/long coordinates and ability to add additional direction information.

A few options that have been discussed:
- Allow choosing an existing location from the event places API in Content Hub - https://github.com/uiowa/uiowa/issues/6402#issuecomment-1559972659
- Custom entity type for locations that are referenced from an event. (This is the requested solution by our partners and may make sense since they have many specific locations (e.g. East side steps).
- Use existing event location functionality on SiteNow events but add additional fields and introduce replicate functionality and add a location filter (not requested yet) so they can easily find an appropriate event to replicate. Could introduce extra effort of removing replicated info.

# Proposed Solution

bspeare:
I took a quick look at the locations that Oniowa is using.  I've pasted them below.  

<details>

*   41.659795000000, -91.535279000000
    
    University Capitol Center, Suite 3719
    
    University Capitol Center, Suite 3719
    
    200 South Capitol Street
    
*   41.658702850342, -91.551101684570
    
    Kinnick Stadium
    
    Kinnick Stadium
    
    Evashevski Drive
    
*   41.666050000000, -91.534919000000
    
    Green Room, Currier Hall
    
    Green Room, Currier Hall
    
    413 North Clinton St
    
*   41.665149000000, -91.537265000000
    
    Catlett Market Place
    
    Catlett Market Place
    
    350 North Madison Street
    
*   41.660295000000, -91.529453000000
    
    Zen Salon & Spa
    
    Zen Salon & Spa
    
    4 S Linn St
    
*   41.661000000000, -91.533000000000
    
    The Shop- Iowa City
    
    The Shop- Iowa City
    
    4 S Dubuque St
    
*   41.658900000000, -91.533900000000
    
    Revival
    
    Revival
    
    117 E College St
    
*   41.664010000000, -91.531880000000
    
    Field & Finch
    
    Field & Finch
    
    213 N Linn St
    
*   41.661080000000, -91.534032000000
    
    Beno's Flowers and Gifts
    
    Beno's Flowers and Gifts
    
    107 E Iowa Ave
    
*   41.656820000000, -91.535100000000
    
    Beadology
    
    Beadology
    
    355 S Clinton St
    
[...]

PR #6511, opened 2023-06-02
Event Locations Probe - Autofill edition

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/6494

# To do if it moves forward:

- Look into adding chosen or select2 to make dropdown selection easier
- Create a array map of api to field to autofill more efficiently
- Public view display of places on content hub with a webform to add/submit change requests
- Decide on whether dropdown should reset or store the location.
- Should the dropdown go disabled until the location fields are reset (js) to prevent overwriting?

# To Test

```
ddev composer install && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli /node/add/event
```

See the new autofill and geolocation field. Select a location from the list. See things populate and the select reset. Feel free to edit the result. Save. Replicate is available too. That is neat.

out

Video demo here: https://uicapture.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=6a0bf598-616b-41e9-9c5b-b01a00e3b961


joewhitsitt:
This would require modifying our sitenow/chosen library and add some styling but it is possible to add chosen to make selections even easier:

Screenshot 2023-06-07 at 10 13 37 AM

joewhitsitt:
Something I thought about while demonstrating it today. It should clear existing values in case previous location had more field values than the new one. Like if the previous address had lat/long but the new address didn't, autofill will fill values it has but the old lat/long would remain.

joewhitsitt:
Sounds like this solution won't be adopted in full at this time. Bits will be introduced so this PR is for reference for now.

joewhitsitt:
Looks like I messed the composer merge though...

PR #6512, opened 2023-06-02
Event Locations Probe - Custom entity edition 

e-marie-w:
Relates to #6494 

# How to test

- `ddev composer install`
- `ddev blt ds --site=oniowa.uiowa.edu`
- `ddev drush @oniowa.local en geofield`
- `ddev drush @oniowa.local config:import --source ../config/sites/oniowa.uiowa.edu --partial`
- `ddev drush @oniowa.local cim`
- `ddev drush @oniowa.local cr`
- `ddev drush @oniowa.local uli`
- try creating a location. then try creating an event and adding your location via the locations field.


e-marie-w:
This is what happens when I remove Geofield from default: `The import failed due to the following reasons: Configuration <em class="placeholder">core.entity_form_display.node.location.default</em> depends on the <em class="placeholder">Geofield</em> module that will not be installed after import.`

I have the module listed in the site split, am I missing a step here?  I started adding it to the event split only to realize that wouldn't help any, as I'm using geofield in the location content type found only in the site split.

e-marie-w:
Wait okay actually enabling the module helps.

joewhitsitt:
Yeah, i think you still need to remove from default/core.extension and update your test instructions to include the `drush en geofield` before the partial import. A bit different when establishing the site split.

e-marie-w:
Current problems: the locations field is missing from the form display after syncing. When I look at the content type it is still there however. Any attempts to export config to site split end up removing all config_split.patch files.

PR #6502, opened 2023-06-01
Untangling Events and Admissions config

pyrello:
This is hopefully the first step in undoing some of the fragmentation that had occurred between the Admissions site split and the event feature.

1. Removes the full split of `views.view.events` from admissions.
2. Removes layout split from admissions.
3. Auto-submit is turned on for the default /events view.
4. The subevents view on the full node page has the "No results" removed so that nothing will appear if there are no subevents. 
5. The /events/calendar view is disabled for the feature split but turned on for admissions only. 
6. The local tasks css was moved to a separate library and is loaded separate for the local tasks block so that it can be used for the "Tabs" that are unstyled at https://admissions.uiowa.edu/events. 

# How to test

`ddev blt frontend`

```
ddev blt ds && ddev drush @default.local config-split:activate event -y && ddev drush @default.local cim -y && ddev drush @default.local uli events
```
- Check default + event feature
```
ddev blt ds --site admissions.uiowa.edu && ddev drush @admissions.local uli events
```
- Check Admissions and compare local /events page with prod /events page. 
```
ddev blt ds --site iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli events
```
- Check ISWF and compare local /events page with prod /events page. 
```
ddev blt ds --site oniowa.uiowa.edu && ddev drush @oniowa.local uli events
```
- Check OnIowa! and compare local /events page with prod /events page. 

### Check Admissions
```
ddev blt ds --site admissions.uiowa.edu && ddev drush @admissions.local uli areas-of-study
```

- Test desktop/mobile breakpoints
- Test search and reset buttons
- Compare https://admissions.uiowa.ddev.site/areas-of-study against https://admissions.uiowa.edu/areas-of-study
- Compare https://admissions.uiowa.ddev.site/events against https://admissions.uiowa.edu/events (Upcoming and Calendar tabs are now styled for anonymous users)

### Check Housing
```
ddev blt ds --site housing.uiowa.edu && ddev drush @housing.local uli residence-halls
```
- Test desktop/mobile breakpoints
- Test search and reset buttons
- Compare https://housing.uiowa.ddev.site/residence-halls against https://housing.uiowa.edu/residence-halls

Issue #3171, opened 2021-03-03
SiteNow Events view page


joewhitsitt:
Split from #2396
Follow-up to #3170
Blocked by #3458

# Story

As a site owner with the local events content type split enabled, I would like to modify aspects of the view without asking a developer.

The problem right now with providing this solution is that being a feature split and not part of the default, the view needs to be imported when the feature is enabled but ignored when the configuration page tries to ignore it. This requires a dedicated module to install/uninstall. To do this right, it is my opinion that #3458 should be addressed first so that sitenow_events can be used.

# Proposed solution

- [ ] SiteNow Events configuration page similar to SiteNow articles that manipulates views.view.events with options to enable/disable view, custom title, custom header content.
- [ ] Config ignore views.view.events within the feature split.
- [ ] Install/Uninstall logic within sitenow_events that imports/removes the view configuration.

joewhitsitt:
There is a separate issue related to this that would tackle these differently going forward. https://github.com/uiowa/uiowa/issues/3151

IMO, I think we should keep with these dedicated view pages that can be disabled by the webmaster.

bspeare:
I don't know whether this is in the scope of this particular issue, but we might want to look into styling this view of events like the screenshot here: https://github.com/uiowa/uiowa/issues/2727#issuecomment-834403434.

joewhitsitt:
I think the custom path requirement should be removed. I had troubles with ignoring this path at the feature split level which is different than the default article/person paths. That and I think _if_ local events are used, they should have consistent paths. A developer implementing the feature split can split the path config if changes are needed.

joewhitsitt:
Based on the conversation at the Drupal meeting I think the view should be hardcoded as config within the feature split. Admissions should then site split the view for their needs.

We then need to prioritize https://github.com/uiowa/uiowa/issues/3458 so that the feature split can be served by a dedicated module that will handle the custom functionality better.

joewhitsitt:
https://github.com/uiowa/uiowa/issues/3151 is still something to consider but has other problems to solve.

joewhitsitt:
So even though #3513 was closed, I would like to hijack this issue and update it for when we can tackle and provide this functionality.

joewhitsitt:
To move forward with this for Admissions based on recent discussions, we should just create this view for admissions only without configuration. https://github.com/uiowa/uiowa/issues/3458#issuecomment-870836241

https://github.com/uiowa/uiowa/issues/3773

[...]

Issue #3151, opened 2021-03-01
Replace SiteNow People view page with content page


pyrello:
# Story

As a site owner I want manage my People listing page in the same way I manage other pages so that I can define a complex layout and use all the options that are built into the People list block.

# Proposed solution

- [ ] Write an update hook that creates a page node to serve as the People list page.
- [ ] Place the people list block on that page in the main content section.
- [ ] Set the alias of the page to whatever has been chosen as the SiteNow People path alias.
- [ ] Replace the path setting textfield on the SiteNow People settings page with a node entity reference field, similar to how the homepage is set in basic site settings. This new setting will be config ignored to be unique per site.
- [ ] Potentially add some validation that prevents deleting the page.
- [ ] When the site user choose to not have a people page, archive the selected page. If they turn it back on, publish the page again.

joewhitsitt:
SiteNow People view page can be disabled so I don't really understand why people can't just do this on their own without the complexities of deletion prevention, moderation, etc.? I would imagine that if this was considered, it should wait till v2 is gone. Then I wonder where SiteNow Articles' view page fits in to this with the attachment displays (monthly archive, feed). Right now, the view pages are something we can reliably target if we need to, but that gets out of the way if not needed. I think discussion about when and when not to use a view page is something we should define first.

PR #6530, opened 2023-06-09
[Event feature] Events view updates

### LINKED ISSUE #6365: OnIowa event filters and theming
# Story

As an owner of oniowa.uiowa.edu, I would like to have special filters for our event content and to group the events by day.

# Additional Info

**Display**
Would like filters for the following items:
- Attendance Required - Yes/No
- Category - Select List
- Event type - virtual/hybrid/in person
- select list of each day to filter by

They would like to group events by day on the display and output title and time similar to https://web.archive.org/web/20220815131834/https://oniowa.uiowa.edu/events

# Proposed Solution

- Determine if we want to make filters configurable by site webmasters or determine which should be in default vs custom to OnIowa
- See how we can handle the select list by day
- Add any additional theming
- Alter display for OnIowa to group by day

cory-skeers:
Resolves #6365 

# Todo

- [x] Move oniowa install to update hook and test
- [x] Determine a sane date range for when filter is not exposed.
- [ ] ~~Group results by date~~ https://github.com/uiowa/uiowa/pull/6530#issuecomment-1602856395
- [x] Move JS/Styles to work with all exposed filters?
- [ ] ~~Date select is not in secondary?~~ https://github.com/uiowa/uiowa/pull/6530#issuecomment-1602947319

# How to test

`ddev composer install && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli events`
Compare https://admissions.uiowa.ddev.site/events and https://admissions.uiowa.edu/events

- There should be no changes. (Note: there may be changes from #6502 present, if that has not yet deployed)
- Check https://admissions.uiowa.ddev.site/areas-of-study to make sure the exposed form selects get the color enhancement on change.

`ddev composer install &&  ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli events`
Compare https://iowasummerwritingfestival.uiowa.ddev.site/events and https://iowasummerwritingfestival.uiowa.edu/events

- There should be no changes.

`ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default && ddev drush @default.local config-split:activate event && ddev drush @default.local uli admin/config/sitenow/sitenow-events`

- Don't enable the date range filter. Create some events. You should see results.
- Filter changes should be reflected immediately.

`ddev composer install && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli events`
Check out https://oniowa.uiowa.ddev.site/events after creating a few events on different days.

- Date select should reflect the days from the results and work when a date is selected.
- Presenter filter should not be enabled.
- Check out the select text color changes when an option is selected.
- As a webmaster or other editor, you should see a linked message for the configuration page similar to articles and people.


cory-skeers:
There's a slight issue specifically with ISWF, related to https://www.drupal.org/project/drupal/issues/3073362.

The approach in this PR has been to add and expose all filters that would be used across each of the sites using events, and then configure which will be shown on the end page (which then un-exposes them on a views pre build). Specifically for the "Attendance Required" filter, nodes which have a null value for this (ie were created prior to the field being added and have not since been edited/saved to get a value) get excluded from the view.

A simple path forward would be to update hook the attendance required value for all events on iswf and set explicitly to false (or even a bulk save of all nodes should take care of this).

cory-skeers:
> A simple path forward would be to update hook the attendance required value for all events on iswf and set explicitly to false (or even a bulk save of all nodes should take care of this).

Oops, maybe not. The issue is with the attendance required filter, but setting values for them didn't fix it. Instead it is currently set to only show if attendance-required is true, though the filter settings should be defaulting to any true or false. Or I've been looking at it for too long.

bspeare:
I pushed up some work that adjusts the reset button position to make it more visible. 
- Moved the "apply" button away from the main search text field so that it no longer appears tied to the search field.
- Reset button now appears inline next to the "apply" to reset the form

Screenshot 2023-06-14 at 10 37 46 AM

Screenshot 2023-06-14 at 10 35 40 AM

bspeare:
Do we want to move the search text field inside the advanced options?  If we want to adjust the view for that, it would look like this after that change. 

Screenshot 2023-06-14 at 10 35 14 AM

cory-skeers:
> Do we want to move the search text field inside the advanced options?

Were you thinking this just for Oniowa? I think with their focus on per-day grouping and selection, that makes sense to me, and that they'd probably prefer that as well.

bspeare:
Yes, just for Oniowa if we are able to.

cory-skeers:
Adding some clarity to what I was seeing for https://github.com/uiowa/uiowa/pull/6530/commits/a0f271ec8479a5543e4f5fd9aea0af4c3f09ead5

Using the views UI to set the default value for the boolean attendance required filter sets the value to "All," which is supposed to translate to either True/False, but instead keeps getting converted to "True".

Setting the views config to set this value to `null` instead appears to work whether the filter is exposed or not (so, for both oniowa and admissions/iswf), but need to double check this.

cory-skeers:
Oniowa's events filters config (supposed to be set as part of enabling oniowa_core) isn't getting set.

Until this is fixed, manually setting the config filter displays you should include all but the 'presenter'

cory-skeers:
Current todos: 

- [x] Styling/labeling for additional field filters, which are needed for oniowa (attendance required, attendance mode, event categories), as well as any styling specific to oniowa (top or side, filter placement, etc).  
    - [x] Labeling
    - [x] Styling
- [x] See about grouping results on oniowa by day.
- [x] Config update for oniowa (currently resides in `oniowa_core.install`) needs fixing.
- [x] Caching for day select list filter may not be working? https://github.com/uiowa/uiowa/pull/6530#issuecomment-1594905789
- [x] Investigate alternative to `#process` and `#pre_render` array options.
- [x] Presenters filter is not showing up if there are no presenters, and it should.


<sub><sub>edited by gridgway, wed 06/21/2023 @ 11:04AM</sub></sub>

pyrello:
Going to start poking around.

pyrello:
This is what I'm currently seeing. It seems like there should be more filters?

Screenshot 2023-06-15 at 9 30 50 AM

pyrello:
> It seems like there should be more filters?

I think that maybe this is something that should be getting set in `oniowa_core_install`, but I don't think that runs when the module is enabled via config split. Not 100% about this though.

pyrello:
I adjusted the cli commands to enable the `oniowa_core` module before the config split is imported (didn't this used to be necessary?) and I'm seeing something different now:

Screenshot 2023-06-15 at 9 40 32 AM

Still not sure if this is what I'm supposed to be seeing, but seems like a step in the right direction.

pyrello:
Doing a code review and noticed this:

https://github.com/uiowa/uiowa/blob/1f1f09d6dfbd24ab18974aef7a22c3e33b63a79b/docroot/modules/custom/sitenow_events/sitenow_events.module#L106-L135

This seems similar to what we are doing in other cases where we are allow configurable alterations for views pages (e.g. SiteNow people sorting options). However this approach is different in that it reads the configurable values before the view is rendered and overwrites the config every time that happens. In our SiteNow people version of this, we actually update the stored configuration and have some mechanism for ignoring the difference.

I'm not commenting on the merits of either approach, but wondering about whether we want to diverge in our approach to solving this type of problem? Tagging @joewhitsitt because I think he may have some insight into the reasons why we chose the original approach.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
```diff
@@ -0,0 +1,83 @@
...
+      $data = \Drupal::cache()->get($cid);
+      if (!$data) {
+        $options = [];
+        $options[''] = t('Select a date');
+        $query = \Drupal::entityQuery('node');
+        $query->condition('type', 'event')
+          ->condition('status', 1)
+          ->condition('field_event_when.end_value', time(), '>')
+          ->sort('field_event_when', 'ASC');
+        $result = $query->execute();
```
pyrello: I think this query can be re-written to return the actual field value we are looking for so that we don't have to load all the nodes below.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
```diff
@@ -0,0 +1,83 @@
...
+            if ($date) {
+              $day = date('l, F j', $date);
+              if (!isset($options[$date])) {
+                $options[$date] = $day;
+              }
+            }
+          }
+        }
+        // Cache the options, but set them to be invalidated
+        // should any events be CUDed.
```
pyrello: Its not immediately clear what `CUD` might stand for, so we may want to spell this out.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
```diff
@@ -0,0 +1,83 @@
...
+      $form['field_event_when_end_value']['#access'] = FALSE;
+      array_unshift($form['#submit'], '_oniowa_core_date_select_list');
+      break;
+  }
+}
+
+/**
+ * Custom submit handler for upcoming events page.
+ */
+function _oniowa_core_date_select_list($form, &$form_state) {
```
pyrello: @joewhitsitt I seem to recall that we made some effort to move away from function names with an underscore prefix for reasons of incompatibility with future versions of PHPstan. Is this something we are trying to move away from?

pyrello:
> Oniowa's events filters config (supposed to be set as part of enabling oniowa_core) isn't getting set.
> 
> Until this is fixed, manually setting the config filter displays you should include all but the 'presenter'

Seeing this now and realized that it is related to my earlier comment:

> I think that maybe this is something that should be getting set in oniowa_core_install, but I don't think that runs when the module is enabled via config split. Not 100% about this though.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
joewhitsitt: CreatedUpdatedDeleted?

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
joewhitsitt: Not sure if we have made much effort, but that rule is on the horizon.

joewhitsitt:
>may have some insight into the reasons why we chose the original approach

It was just the knowledge at the time early on in the platform. Paired with config_ignore, this is how we delivered a default and modified the views based on the config. When it came time to update something it was a pain because we had to re-write the config as an update hook. It is also why we avoided providing an events view config in the first place because config ignore and it being a feature split. https://github.com/uiowa/uiowa/issues/3171

if this approach works, I think it might be worth trying to do the same for articles/people if we don't plan to eventually go this route: https://github.com/uiowa/uiowa/issues/3151

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
cory-skeers:
```suggestion
        // should any events be Created/updated/deleted.
```

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
cory-skeers: Good info! I've updated to remove the underscore.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
cory-skeers: Is there a way to use `entityQuery` and get the field values directly? There used to be `entityFieldQuery`, but I couldn't find an equivalent. For now, I've rewritten it using `database()` and just querying the field table directly.

pyrello: I was just trying to look that up. I think I might have been getting confused. The documentation for `entityQuery` is pretty light, though, so hard to know for sure. I think what you have looks good. Nice use of `tableExists` for some defensive checking!

pyrello:
Haven't looked into from a debugging perspective, but it seems like we may have a caching issue with the date filter. I would expect to see two dates in the filter, I think.

Screenshot 2023-06-16 at 10 27 56 AM

cory-skeers:
> Haven't looked into from a debugging perspective, but it seems like we may have a caching issue with the date filter. I would expect to see two dates in the filter, I think.

I think I'd broken this when switching it to a direct field db query instead of loading the entities, where it was only pulling in the first result. If that was the case, it should be working now in pulling in multiple options and caching appropriately.

cory-skeers:
It looks like enabling the module prior to enabling the site split is setting the configuration properly.

Currently, all the proper filters are displaying within the form with this configuration, but most of the filters (attendance required, attendance mode, and text search) are being laid overtop each other and end up hidden behind the date select filter (the category filter appears separate and within the Advanced options).

bspeare:
Regarding all the form fields being stacked on top of each other, all of the filters (besides the date select) need to be set inside the `bef` advanced options `<details class="bef--secondary js-form-wrapper form-wrapper">` or they will not appear as desired due to the CSS grid settings in place.

joewhitsitt:
Going to resolve merge conflicts

joewhitsitt:
Because of the merge from main, `ddev composer install` is now needed. Updated test instructions

joewhitsitt:
>all of the filters (besides the date select) need to be set inside the bef advanced options

I can push a fix for this if nobody else is doing it

pyrello:
> > all of the filters (besides the date select) need to be set inside the bef advanced options
> 
> I can push a fix for this if nobody else is doing it

@GaryRidgway was working on this, but I'm not sure if he was able to get a commit pushed or not.

GaryRidgway:
I have a fix for it and whatnot, but I was just about to investigate where to put the config the we changed before i did

joewhitsitt:
Based on the huddle with Alan, it looks like docroot/modules/custom/sitenow_events/src/Form/EventsSettingsForm.php could benefit from the same:

```
// Clear cache.
drupal_flush_all_caches();
```
as sitenow_people and sitenow_articles on submit

joewhitsitt:
I have also not been able to reproduce the presenters issue. My test:
```
ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default && ddev drush @default.local config-split:activate event && ddev drush @default.local uli admin/config/sitenow/sitenow-events
```
Add the filters. Create an event without a presenter. Go to /events. See the presenters filter.

Instead I think I found a separate issue that if date range filter is _not_ on, no results display.

bspeare:
This might be fixed already, but I just noticed that the reason the Categories select looks weird is because there are two `select` elements being rendered. 

Screenshot 2023-06-21 at 2 53 48 PM
Screenshot 2023-06-21 at 2 53 42 PM

joewhitsitt:
>Instead I think I found a separate issue that if date range filter is not on, no results display.

I am still encountering this and trying to figure it out. I feel like I encountered this way back when in trying to show hide filters for the people view. It is like it needs a value if not exposed and attempts to unset on the fly have not worked.

joewhitsitt:
If the date range exists but is not exposed, it seems to need a value. Thoughts on setting a range? 10+ year okay?:

```
diff --git a/docroot/modules/custom/sitenow_events/sitenow_events.module b/docroot/modules/custom/sitenow_events/sitenow_events.module
index e8b9a4de4..9f48da616 100755
--- a/docroot/modules/custom/sitenow_events/sitenow_events.module
+++ b/docroot/modules/custom/sitenow_events/sitenow_events.module
@@ -126,10 +126,13 @@ function sitenow_events_views_pre_build(ViewExecutable $view) {
     ] as $filter_option => $field_label) {
       if ($config->get("filter_display.{$filter_option}") == FALSE) {
         if ($filter_option === 'date_range') {
+          $timestamp = strtotime('now');
           $filter = $view->display_handler->getHandler('filter', "{$field_label}_value");
           $filter->options['exposed'] = FALSE;
+          $filter->value["value"] = date('m/d/Y', strtotime('+10 year', $timestamp));
           $filter = $view->display_handler->getHandler('filter', "{$field_label}_end_value");
           $filter->options['exposed'] = FALSE;
+          $filter->value["value"] = date('m/d/Y', $timestamp);
           continue;
         }
         $filter = $view->display_handler->getHandler('filter', $field_label);
```

joewhitsitt:
I think with the site split enabled in PROD, the setup instructions for oniowa need to change to enable the module run the install hook.

Probably just needs to move from install to update.

joewhitsitt:
>Group results by date

Of those surveyed, I think we want to see if we can move forward with just the date select and not the results grouping as this will cause the view to be split which we have avoided up till this point.

https://iowaweb.slack.com/archives/C0164FNN1PV/p1687446940962369

joewhitsitt:
>Date select is not in secondary?

I think we are okay here based on the discussion here: https://iowaweb.slack.com/archives/C0164FNN1PV/p1687447568051779

cory-skeers:
I did a code review of the updates since I'd left off, and I think things are in a really good state. 👍

joewhitsitt (approved):
Since Cory can't approve his own PR and I spent a good time testing while figuring out some of the remaining items, I am going to give an official approval. However, I still think another tester of the PR in it's current state would be neat.

bspeare:
I was looking at this yesterday to see if there was a way to further detangle the admissions split from this work. 

For instance, we could reduce the `config_split.patch.view.view.events.yml` to:

```
adding:
  display:
    default:
      display_options:
        pager:
          options:
            items_per_page: 20
            tags:
              next: ››
              previous: ‹‹
    page_calendar:
      display_options:
        enabled: true
removing:
  dependencies:
    module:
      - better_exposed_filters
  display:
    default:
      display_options:
        pager:
          options:
            items_per_page: 10
            tags:
              next: 'Next ›'
              previous: '‹ Previous'
    page_calendar:
      display_options:
        enabled: false
```

And the necessary overrides could happen in:

```
/**
 * Implements hook_views_pre_view().
 */
function admissions_core_views_pre_view(ViewExecutable $view, $display_id, array &$args) {

  if ($view->id() === 'events' && $display_id === 'page_upcoming') {
    // Add CSS classes.
    $view->display_handler->setOption('css_class', 'views-form--stacked grid--twocol--25-75');
    // Set the path for the page display.
    $view->display_handler->setOption('path', 'events/upcoming');
    // Set the menu options for the page display.
    $view->display_handler->setOption('menu', [
      'type' => 'default tab',
      'title' => 'Upcoming',
    ]);
  }
  ```
  
This would allow admissions to use the new configuration page, however, I was unable to alter the menu and path in the above `hook_views_pre_view()`.

joewhitsitt:
Does it work if they are in hook_views_pre_build() instead of hook_views_pre_view()?

I'd argue that Admissions doesn't need to be custom for per page count and next/prev button text reducing it further to just the calendar stuff.

GaryRidgway:
pulling to test
<details>
<summary>checkboxes</summary>    


- [x] ddev composer install && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli events
- [x] Compare https://admissions.uiowa.ddev.site/events and https://admissions.uiowa.edu/events

- [x] There should be no changes. (Note: there may be changes from https://github.com/uiowa/uiowa/pull/6502 present, if that has not yet deployed)
- [x] Check https://admissions.uiowa.ddev.site/areas-of-study to make sure the exposed form selects get the color enhancement on change.
- [x] ddev composer install &&  ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli events
- [x] Compare https://iowasummerwritingfestival.uiowa.ddev.site/events and https://iowasummerwritingfestival.uiowa.edu/events

- [x] There should be no changes.
- [x] ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default && ddev drush @default.local config-split:activate event && ddev drush @default.local uli admin/config/sitenow/sitenow-events

- [x] Don't enable the date range filter. Create some events. You should see results.
- [x] Filter changes should be reflected immediately.
- [x] ddev composer install && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli events
- [x] Check out https://oniowa.uiowa.ddev.site/events after creating a few events on different days.

- [x] Date select should reflect the days from the results and work when a date is selected.
- [x] Presenter filter should not be enabled.
- [x] Check out the select text color changes when an option is selected.
- [x] As a webmaster or other editor, you should see a linked message for the configuration page similar to articles and people.
</details>

Notes from review:
I'd love to see the Genre dropdown actually show a list of genres on the iowasummerwritingfestival site.

bspeare:
Thanks for the suggestion @joewhitsitt.  I couldn't get that working with `hook_views_pre_build()` but it would be nice if we could!

GaryRidgway (approved):
I have gone through and checked every box in the test instructions, and everything worked as described. Approved!

joewhitsitt:
Forgot about Ben's comments [here](https://github.com/uiowa/uiowa/pull/6530#issuecomment-1604375788). Should this be a separate slice?

GaryRidgway:
oh interesting, i think we should do it now in my opinion. The investment in familiarity seems better now, but I am open to being persuaded

joewhitsitt:
Discussed in standup, we are going to make it a follow-up issue. Created https://github.com/uiowa/uiowa/issues/6577

Issue #6381, opened 2023-04-26
June Dependency Updates


joewhitsitt:
We should try to review and merge in dependency updates on a monthly basis. Until this is automated more, here is a guide:

- Review the [outstanding dependency updates](https://github.com/uiowa/uiowa/labels/dependencies) and for ones that we plan to address in this epic, list them below. Some are PRs by @dependabot, some are issues.
- Dependencies related to our frontend build tools (not executed on the websites) are usually safe to merge as-is but might cause issue with other packages. Check travis build for messages related to `blt frontend`.
- Each @dependabot PR can be worked on separately. Each dependency PR you merge will cause merge conflicts with all the other ones. As long as you don't alter the branch (e.g. merge in `main`, push a commit), you can fix the merge conflicts by running `@dependabot rebase`.
- If a @dependabot PR does require some type of modification (e.g. removing a patch, updating incompatible code), it is okay to make those changes directly in relevant PR. However, it is a good idea to wait until you have merged all the other "easy" dependency PRs first because you will lose the ability to use `@dependabot rebase` to have the PR's merge conflicts resolved.
- Periodically check [locked dependencies](https://github.com/uiowa/uiowa/blob/main/README.md#locked-packages) for new versions. Make notes about why a dependency was/is locked in the repo Readme.
- Review the changelog for each version we are jumping to and the versions in-between. Make note of certain functionality in our codebase that is using the functionality for testing instructions later.
- It can also be helpful to review the git diff comparison between the old version and the new one to see how extensive the code changes are and to get an idea of what things you might need to test.
- Use https://github.com/uiowa/uiowa#updating-dependencies as a guide to test for update hooks, configuration changes and commit. This command is helpful if there are a lot of configuration changes across feature/site splits: https://github.com/uiowa/uiowa/blob/main/blt/src/Blt/Plugin/Commands/ConfigSplitCommands.php
- Dependency PR's with update hooks or requiring a cache rebuild may require off-hours deployment. If this is the case, add the appropriate label to the PR.
- Once this issue is completed, copy these notes to the next month's issue and introduce it in parking lot.

joewhitsitt:
As noted earlier, it would be good to focus on dependencies that move us towards https://github.com/uiowa/uiowa/issues/6061

joewhitsitt:
Focusing first on frontend build dependencies and locked packages (https://github.com/uiowa/uiowa/blob/main/README.md#locked-packages)

by way of

[...]

PR #6533, opened 2023-06-12
Upgrade node_revision_delete

### LINKED ISSUE #6347: remove revision module paypal link
# Problem
When deleting a revision I get to a page where I am asked to make a donation to the module owners paypal. Is there a way to remove this link? 

image

# Steps to Reproduce

1. Edit any page that has historical edits
2. click on Revisions
3. select Delete from the dropdown next to a revision

joewhitsitt:
Alternative to #6226
Resolves https://github.com/uiowa/uiowa/issues/6347
Relates to #6381 

_Note: some content types are configured with a different number of revisions to keep._

# To Test

- Pick a content type to test. They should all be updated and match their previous settings. Sync the site that has the content type. 
- There should be no issues with the database update or config import.
- Create some revisions of the content. Try manually deleting a revision.
- Create more than the limit and run cron to see what is cleaned up.
- There should still be a message that displays the current limit for the content type.
- Donation link of #6347 should be gone
- _Note the ability to delete prior revisions of the revision you are deleting is no longer a feature. Most of the drush commands have also been removed._


pyrello:
> Note the ability to delete prior revisions of the revision you are deleting is no longer a feature. Most of the drush commands have also been removed.

Did we decide we were cool with this functionality going away for the meantime? I can't remember.

joewhitsitt:
I haven't recorded any opinion from the team. I am personally okay with it. Most types are set at 10 revisions or less so I don't think the amount of revisions is too overwhelming. It was helpful when we were manually cleaning up coronavirus to switch from v2 to v3.

briand44:
I was also ok with it going away as long as the team wasn't concerned with the loss of drush commands

pyrello:
> as long as the team wasn't concerned with the loss of drush commands

I'm not sure I can remember ever using one of these commands.

e-marie-w:
Pulling down to test.

e-marie-w:
Just to double-check, I should be testing every custom content type we have?

joewhitsitt:
You can see from the files changed that the changes were fairly uniform across sites/content types, so I think testing the default types and maybe one or two others should be enough test coverage. If you see something that stands out, dive-deeper.

e-marie-w:
I'm noticing that we have some content types that don't use node revision delete, like thesis defense. Just wondering if it was purposeful, since I can still delete revisions for it.

pyrello:
I don't think it was purposeful.

e-marie-w (approved):
Tested article, page, person on default and sandbox
Tested area of study, event, and transfer tips on admissions
Tested building on facilities
Tested mentor on grad
Tested FAQ on hr
Tested opportunity on sppa
Tested book on uipress

All of them worked as expected. I could manually delete a revision, and running cron removed the oldest revisions so the number of revisions matched the limit. Looking at the changes, I can see the content types retained the correct limit number. 

Only thing of note is I often had to drop the local database before I could sync a site successfully, I guess I had a lot of gremlins. 

I can make another comment with a list of content types that don't use node revision delete, if that's something we're interested in knowing.

PR #6539, opened 2023-06-13
Update linkit to most recent compatible version

pyrello:
Relates to #6381 

# How to test

```
ddev composer install && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli
```
- Edit a page and check linkit functionality has no regressions when using the link button in CKEditor.
- Go to layout tab for a page and add/edit a card block. Check that link suggestions functionality works the same.
- Ensure that you can still link to webforms.
- Check that linking to media works the same.


pyrello:
@joewhitsitt I'm wondering if we might want to consider updating to this patch: https://www.drupal.org/project/linkit/issues/2946234#comment-14596374

joewhitsitt:
Sure. Surprised that is still out there. Because it is related to another contrib module, I wonder if it will ever get merged..

pyrello:
The patch referenced above (as well as the one we are currently using) are both targeting the 5.x branch. The patch applies cleanly, but causes the composer.json file for the module to be malformed, resulting in warning messages when running commands like `blt ds`.

joewhitsitt:
pulling to test

joewhitsitt (approved):
In the WYSIWYG this seems to work as usual. 

The webform matcher now excludes webforms that have this url path setting unchecked.

Screenshot 2023-06-13 at 4 35 05 PM

linkit_media_library still works as expected.

>Go to layout tab for a page and add/edit a card block. Check that link suggestions functionality works the same.

This isn't LinkIt, but yes, still autocompletes without context additions. Webform wasn't available for these fields. We could work on updating fields now..

Screenshot 2023-06-13 at 4 33 48 PM

Onward!

Issue #3458, opened 2021-05-05
Move UI Events functionality to uiowa_events


joewhitsitt:
## Story

As a sprint demo attending customer, I can expect speedy development of my [event](https://github.com/uiowa/uiowa/labels/uiowa_events) [feature](https://github.com/uiowa/uiowa/labels/sitenow_events) knowing the foundation and organization of the functionality has been we thought out and architected to accomodate the variety of event-like features on the SiteNow v2 and v3 platform.

## Issue

While developing the local event content type feature (https://github.com/uiowa/uiowa/issues/3170), I realized that I think our code for the dynamic UI events should be moved to uiowa_events and that the feature developed in #3170 should stay in sitenow_events.

>I think the current sitenow_events stuff (content/events.uiowa.edu) should be moved to a new module uiowa_events, and that the local event feature should take over sitenow_events. That is unless we are completely trying to get rid of sitenow_* prefixed modules. This all gets messy based on the existing feature splits for SiteNow (e.g. uiowa_areas_of_study) and the difference between local site events and University of Iowa Events. Downside to leaving this here now is that this code is always on even if the feature split isn't...

## Related

https://github.com/uiowa/uiowa/issues/2882

## Proposed Solution

- [ ] Move and rebrand SiteNow Events as UI Events (uiowa_events)
- [ ] Designate sitenow_events as the home of the local event content type similar to SiteNow Articles and SiteNow People.
- [ ] Move event feature split theme stuff out of uids_base if it is to stay turned off by default.
- [ ] We would need to capture current sitenow_events.settings config and import them into uiowa_events.settings before they are deleted from the db.
- [ ] Some things will still be called "event" or "events" and it will be important to document whether they belong to sitenow_events or uiowa_events.

pyrello:
This has been marked blocked for now pending a meeting to discuss how we want to move forward generally with events on the platform. There is the potential to use an event content type as a replacement for the event custom block type and to allow people to place listings of events on pages using a view block similar to the people block developed in #2702.

joewhitsitt:
Reducing the priority of this issue based on recent discussions. I will surface https://github.com/uiowa/uiowa/issues/3171 in parking lot with a custom to admissions only solution while work on pushing forward with  https://github.com/orgs/uiowa/projects/1#card-63850273 which could be done in conjunction with this issue.

pyrello:
[...]

PR #6553, opened 2023-06-15
Event Locations Enhancements

### LINKED ISSUE #6402: OnIowa event locations
# Story

As an OnIowa website owner, I would like to be able to reuse locations for events happening during the same year as well as from year to year so that we do not have to enter the same location information each time.

# Additional Information

They would like the ability to have a location title/name for places like Hubbard Park
They would like the ability to store lat/long coordinates for a location.
They would like the ability to provide additional direction information for a location.
The location information will need to be made available in the API consumed by the Hawktools app in a future slice.

# Proposed Solution

- Add replicate functionality to SiteNow events (local drupal events)
- Update company name label to location name
- Add lat/long geofields
- Add additional directions field
- Create an internally focused display that shows existing events on their site with the location where they can easily find and get to an appropriate event to replicate

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6402

Same as https://github.com/uiowa/uiowa/pull/6511 but without autofill 
stuff.

- New fields and replicate in the default event feature.
- Events backend overview for OnIowa only.

Screenshot 2023-06-19 at 9 34 50 AM

# Test

```
ddev composer install && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local config:import --source ../config/sites/oniowa.uiowa.edu --partial -y && ddev drush @oniowa.local config-split:activate site -y && ddev drush @oniowa.local uli node/add/event
```
- Create an event node with an address location including the ~company~ location name field 👀, geolocation and driving directions. geolocation should not display on full node.
- Go to admin/content. You should see a new "Events" link next to Overview and Moderated content. That view should have basic event information displayed that the customer can use to identify similar locations for use with replicate.
- Replicate? Yeah. Try it on the event you just created.
- Check admissions/iswf for clean imports. Clean imports should mean they have the new fields and are ready to go. Those sites should also have replicate available due to the update hook.
```
ddev blt ds --site=admissions.uiowa.edu && ddev blt ds --site=iowasummerwritingfestival.uiowa.edu
```


joewhitsitt:
Looks like I messed the composer merge though...

Update: fixed. Replaced the composer.* files with main and re-required geofield.

pyrello:
Going to test this.

docroot/profiles/custom/sitenow/sitenow.install
```diff
@@ -2459,3 +2459,16 @@ function sitenow_update_9062(&$sandbox) {
...
+ */
+function sitenow_update_9063() {
+  $config_factory = \Drupal::configFactory();
+  $config = $config_factory->getEditable('uiowa_core.settings');
+  $replicate_allowed = $config->get('uiowa_core.replicate_allowed');
+  $replicate_allowed[] = 'event';
+  $config
+    ->set('uiowa_core.replicate_allowed', $replicate_allowed)
+    ->save();
+}
```
pyrello, line 2474: Should this be limited to just sites that have the event feature turned on?

pyrello:
This looks good according to my testing. Just waiting for response on the outstanding question and the ready to approve.

docroot/profiles/custom/sitenow/sitenow.install
joewhitsitt: Good question. I originally went with the minimal slice for this thinking event feature might eventually be part of the default. Hopeful thinking I guess.

I think to properly do it the other way would entail:

- a config split check on the existing update hook
- an install hook for the event feature (ignored config)
- removing it from default config
- an uninstall hook for sites that turn off the feature.

All of which can be done, just wondering if it is worth it.

pyrello: Strictly from a standpoint of "Do we need to do this for it to work?" its not necessary. My concern is more that we are setting up config that doesn't match the actual options for some sites. This is probably very unlikely to cause any issues. But it is also a small detail that we are likely to miss when troubleshooting some future issue that comes down to the config option being overly broad. Maybe us having this conversation is enough.

pyrello (approved):
Works well.

docroot/profiles/custom/sitenow/sitenow.install
joewhitsitt: Sounds good. I appreciate you bringing this up.

oh the other gotcha is we don't have a dedicated module for the feature... https://github.com/uiowa/uiowa/issues/3458 the install hook would require a separate module to be turned on to perform this configuration change or we would have to add a drush config set to the feature instructions.

PR #6562, opened 2023-06-20
[Admissions] slideshow of areas of study using banner

### LINKED ISSUE #6523: Admissions visitor center AoS display
# Story

As a member of the Office of Admissions, we would like a fun visual display that cycles through Majors, minors and certificates showing Why Iowa and potentially other Area of Study information.

# Additional Info

Admissions has been in contact with the CSI app devs on this as well and would like a 2nd display showing a US map with pins representing visitors. This information comes from MAUI.

It seems like the display of MAUI data is outside of the scope of work our team should be doing. The Area of Study info on the other hand, could potentially be within our realm.

# Questions

Do we think we should take on the Area of Study display ourselves or is this something that seems like a better fit for the CSI app devs? If it seems like a better fit for the app devs, we would need to make the Area of Study data available for the app devs to consume.

# Proposed solution

- [x] They would like the banner text, heading, pre-title to fill the screen more
- [ ] They would like to see a few other animation transition options
- [ ] Look into making the animation more performant to handle a huge number of slides
- [ ] Dynamically generate the number of slides variable in the CSS?

bspeare:
- [x] Work on removing the scss variables from the file.
- [ ] Create other transition options???
- ~~[ ] Adjust picture spacing so it doesn't clip off the top of the screen.~~
- [x] Use JS to dynamically edit a root variable to change the slides.
- [x] perhaps use JS to dynamically create delays in animation as well?

Nice to have
- ~~[ ] Investigate Text and background color blend modes~~

## To test
```
ddev blt frontend && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli node/add/page
```
Add a slideshow block to the newly created page and adjust the container settings to "Edge-to-edge" and save. 

Notes:
- You may have to fiddle with some CSS to get the save button to be clickable (open dev tools and find the "shutter" div and hide it).
- Remember, this is not a review to be perfect, this is a review of the current slice before we deploy it to admissions.uiowa.edu for testing on their digital sign.


cory-skeers:
It looks like there aren't specific perms for the block, so anyone who has access to edit layouts can place it? For them, that should still only be webmasters, but that still leaves quite a few users who would be able to see it and place it. I'm mostly worried if someone not involved with the display comes across it working elsewhere on the site and places it out of curiosity, removing the block can be tricky.

Otherwise, this works great, and I think is in a good place to be pushed up for in-place testing with their displays.

bspeare:
Going to try to adjust the block permissions so we can get this merged.

bspeare:
I removed the slideshow block as an option from the `config/sites/admissions.uiowa.edu/core.entity_view_display.node.page.default.yml`.   Once this is deployed, we can temporarily override this and place the block on a page.

cory-skeers (approved):
I like the plan to temporarily override to place the block, and this way we can avoid any potential mishaps with accidental placements.

👍

PR #6571, opened 2023-06-21
[Proof of concept] hiding header and footer

### LINKED ISSUE #2665: Ability to hide the search, main navigation, and footer (but keep footer socket) per page
# Story

As a marketer crafting a landing page experience, I can hide the search, main navigation, and footer (but keep footer socket) per page.

# Proposed solution

- [ ] ?
- [ ] Accessibility: let's not lose or improperly hide a11y info that is still expected. For example, footer is still supposed to have contact information.

## Nice to have:

- [ ] Meta tag for nofollow tag
- [ ] Layer a form block over a background image
- [ ] A/B testing capabilities

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/2665. 

Like what we have set up with non-branded sites, I think this option would be useful for admins to use for customer requests like:

- Custom landing pages for social campaigns
- Custom one-off pages like admissions kiosk and slideshow
- Recreating a viewbook site
  - We might want to set more granular options like:
    - Hide main nav (useful for viewbooks)
    - Hide search
    - Hide header
    - Hide footer

# How to test
- Edit a page and click on the "Hide header and footer" checkbox
- Save page and verify that header and footer are not rendered in markup


joewhitsitt:
My concern with this stems from the functionality being available on an otherwise public site. You mentioned that the functionality would be admin-only which addresses some of my concerns.

I am not sure I totally understand unsetting the header and footer completely. Should we start with a "minimal" header/footer since the page is still supposed to be interactive. Aren't we supposed to have identifying elements according to the brand? Example 1 in #2665 still has some kind of header. Even powerpoint slides are supposed to have some information https://brand.uiowa.edu/powerpoint-templates

I don't think it has been enforced but since this will be user-interactive, a contact link is supposed to be available. https://github.com/uiowa/uiowa/issues/2665#issuecomment-745520064

bspeare:
I'm going to close this and hopefully we can revisit this in the future.

PR #6574, opened 2023-06-23
[pharmacy.uiowa.edu] Territory updates

### LINKED ISSUE #6485: Pharmacy person updates
# Story

As a Pharmacy editor, I don't want to see the international recruitment option (copied from Admissions) in the person state field dropdown.

# Proposed solution

- [ ] check for any content that uses the value and switch to another option.
- [ ] remove the option https://github.com/uiowa/uiowa/blob/main/config/sites/pharmacy.uiowa.edu/field.storage.node.field_person_territory.yml#L14-L16

cory-skeers:
Resolves #6485 

# How to test

`ddev blt ds --site=pharmacy.uiowa.edu && ddev drush @pharmacy.local uli`

Update hook should complete cleanly, followed by config import with no errors.

Edit a few existing Persons with the "Palliative Care Graduate" person type. 
Compare with their prod counterparts and check that they still have the same state(s) selected.
Check that there is no "International" option

Add a new Person with the "Palliative Care Graduate" person type.
Check that there is no "International" option

Check out https://pharmacy.uiowa.ddev.site/pharmd/dual_cert/palliative/certificate_graduates and see that there is no "International" option (as opposed to https://pharmacy.uiowa.edu/pharmd/dual_cert/palliative/certificate_graduates)

Extra credit testing option:
`git checkout main && ddev blt ds --site=pharmacy.uiowa.edu && ddev drush @pharmacy.local uli`
Edit a person and select the International option
`git checkout pharmacy-territory-updates && ddev drush @pharmacy.local updatedb -y && ddev drush @pharmacy.local cim -y`
Check that the update hook runs cleanly. Edit the person again, and see that there are no errors, and the international option is no longer selected (nor available)


cory-skeers:
From https://github.com/uiowa/uiowa/issues/6485#issuecomment-1604363321 there are currently no International instances, there are only 20 rows in that table so it doesn't seem to be frequently changing, and they would like the International option removed anyway, so it seems unlikely that an international value will be added prior to this running.

That said, if one happens to get added, it may lead to issues later. I added a quick check as values are getting re-added to the database to completely skip over those rows if the international option is found, as it seemed the lightest lift to cover our bases in the unlikely event it does get added.

joewhitsitt:
pulling down

joewhitsitt (approved):
Nice fast work. I tried to trip up the extra test by leaving creating a new node but leaving it in a draft state. Thanks for implementing using uiowa_core_state_options

PR #6579, opened 2023-06-26
Alert content type

### LINKED ISSUE #6561: Alert content type
# Story

As a Facilities website owner, we would like to have the ability for Construction Managers to create alert/closure content so they can be made visible to our users.

# Additional Info

Likely a new role is required here, possibly collaborator if we want to try to share naming when possible? (Joe suggestion)

Moderation not required

Ability to schedule needed but not part of this slice
Theming can be a separate slice as well.

**Fields**
Same as on their current site but location should be required.

# Proposed Solution

- Create content type according to our best practices docs
- Create role and assign perms (should only have ability to create alert/closure content)

GaryRidgway:
Creating the Alert content type for #6561 

# Todo

- [x] Add collaborator role (see brand.uiowa.edu) to feature split.
  - [x] should this just be facilities or feature?  
    - Facilities only for now. Can move to alerts later if desired.
- [x] Create an "Add Alert" link for collaborators to use since they don't have access to the content overview. This means a module is born. see docroot/sites/brand.uiowa.edu/modules/brand_core/brand_core.links.menu.yml
  - [x] should there be a link and if so, should everyone see it? - _No and thus no_
- [x] Permissions in the feature for content type and taxonomy. Make sure permissions merge with site split.
  - [ ] Should the permissions be as they are? Current facilities has content overview and whatnot for people who can create alerts.
- [x] Edit the content type settings to remove main menu, set revisions, include in sitemap, disable promote to front, etc. similar to other content types. rabbit hole isn't needed yet for this ctype right?
- [x] Setup basic metatag/schema.org
- [x] ~~Placement of category taxonomy in references sidebar similar to other content types (event)?~~
- [x] Consider default value of "now" for date field?
- [x] Date field doesn't allow a date range of multiple days with start and end times.

# How to test
### Default
- Do the following from any branch while not inside ddev.

```
git pull && git checkout alert-content-type && git pull && ddev composer install && ddev blt ds --site=default && ddev drush @default.local config-split:activate sitenow_alerts -y && ddev drush @default.local uli admin/structure/types
```
- observe that the alert content type exists and it meets the necessary criteria defined in #6561 (expected fields, taxonomy and basic content type setup is in place)

### Facilities
- Do the following from any branch while not inside ddev.

```
git pull && git checkout alert-content-type && git pull && ddev composer install && ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli admin/structure/types
```
- observe that the alert content type exists and the differences from default make sense. Again check #6561 for reference, but also logging in to the live facilities site and checking the fields there is also a good idea.
- create a user assigned to the collaborator role, masquerade and you should be able to access the content overview, and navigate to add an alert. Collaborators should not have the ability to edit any other content but their own alerts.


joewhitsitt:
just a heads up, the facilities part of this should only extend the split. you will need only specify unique facilities items in the complete split instead of using the wildcards. everything else should be added as a partial so you don't have a complete node.type.alert, etc. in both the feature and the site.

GaryRidgway:
> just a heads up, the facilities part of this should only extend the split. you will need only specify unique facilities items in the complete split instead of using the wildcards. everything else should be added as a partial so you don't have a complete node.type.alert, etc. in both the feature and the site.

This should now be fixed! Thanks for the assist @joewhitsitt!

joewhitsitt:
Before it is too late, can we think of a situation where these values would change in which the field would be better served by a allowed options function. That way we don't have repeat of https://github.com/uiowa/uiowa/issues/6485 down the road?

https://github.com/uiowa/uiowa/pull/6579/files#diff-1811d5cc46896dd2446994248cc0e077d19d37cddb6ba953b590e892be6f01b7R13-R23

joewhitsitt:
making it official that i am reviewing after doing it all day ;) pulling down.

joewhitsitt:
I am not seeing the collaborator role from the proposed solution, content type/taxonomy permissions and it looks like we need to run through some more of the wiki https://wiki.uiowa.edu/display/drupaldev/Content+type+creation+notes (menu, sitemap, revisions)

joewhitsitt:
Updating the PR IS with todo items

joewhitsitt:
our wiki for content types mentions metatag and schema.org 

https://schema.org/SpecialAnnouncement exists and actually has a couple fields that match up.

joewhitsitt:
I reviewed this quite a bit initially and while I was working on the additional items. Willing to co-approve this PR with someone. The important part was establishing the feature with site specific fields with clean import/setup which this does. This won't be the last time we touch this content type before they go live.

e-marie-w:
I can test it quick.

e-marie-w:
When masquerading as a collaborator I can still see the content overview.

joewhitsitt:
Sorry, we forgot to update the test intstructions...

We decided that they should be able to similar to their current site. Brian didn't have any concern and thought it would be easier for webmasters to train collaborators through the same interface. Collaborators should not have the ability to edit any other content but their own alerts.

e-marie-w (approved):
Works as expected on default and facilities, code looks good as well. Normally I'd spend more time testing but based on the fact others have tested various parts as well I feel comfortable approving now.

cory-skeers:
I'd pulled this down to poke at a bit as well, and two questions:

1. ~Should there be content moderation?~ Nevermind. Saw the note re: this in the issue after.
2. Should the category allow for multiple values? Facilities' current closure category field is currently a multi-select checkboxes, and usually our term reference fields allow for multiple values. 

Other than those ~two~ one questions, testing went well and this is looking really good!

cory-skeers:
> I'd pulled this down to poke at a bit as well, and two questions:
> 
> 1. ~Should there be content moderation?~ Nevermind. Saw the note re: this in the issue after.
> 2. Should the category allow for multiple values? Facilities' current closure category field is currently a multi-select checkboxes, and usually our term reference fields allow for multiple values.
> 
> Other than those ~two~ one questions, testing went well and this is looking really good!

Some follow-up here after a convo with @briand44

1. Re content moderation. It is not needed for facilities, but could be included as long as collaborators had the ability to publish their alerts as well. But the question remains if it should be included in the default? My thoughts are that for a content type that is designed to be published temporarily and then unpublished, the need for drafts/archive is less important. Maybe could be useful for review state, but unsure.

For now, good to move forward without it, and it can be layered on later if needed.

2. Re: multi-value category. Facilities needs multiple values, as they currently have alerts w/multiple category values (eg Road Closure and Parking Lot Closure on a single alert). ITS currently has similar usage but which currently allow for only one value (and limiting to only one value may be desirable for training/usage ease on site users).

I've updated it to allow for multiple values in the default here to fit Facilities needs, thinking that in the future if it's needed to be limited in number of values, that can be the path toward a split (or depending on the field, a separate field from the default alert category ref field). Another option would be to make it single-value in default, and split for facilities to allow for multiple.

@GaryRidgway or @joewhitsitt, any thoughts or concerns on that change?

cory-skeers:
Set back to draft for now for feedback on https://github.com/uiowa/uiowa/pull/6579#issuecomment-1618885785. If no concerns, feel free to mark as ready again and move forward with merging.

GaryRidgway:
This is fine for me, looks good cory!

joewhitsitt:
Same. Fine with it with the hope that most folks don't turn it into tags 2.

PR #6620, opened 2023-07-07
[OnIowa Events] Map link and directions

### LINKED ISSUE #6608: Event location display updates
# Story

As an OnIowa site owner, I would like to output the event location information in a way that makes it easier for attendees to find the location.

# Additional Info

- Can we output a map of the location or at a minimum link the location to a map?
- Can the additional directions be output near the rest of the location information?

# Proposed solution

- Output map link if lat/long filled out that links to Google Maps.
- Additional directions in details toggle near other location info.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6608

# How to test

- Code review
- Test maps link
- Test directions details toggle
- Test user story "As an OnIowa site owner, I would like to output the event location information in a way that makes it easier for attendees to find the location."
```
ddev blt frontend && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli  event/111/block-i-class-photo
```

Test on other sites using events:
```
ddev blt ds --site=admissions.uiowa.edu && ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @admissions.local uli events && ddev drush @iowasummerwritingfestival.local uli events
```

Test on IWSF split layout
- Verify that fields work and appear in the same order as OnIowa


GaryRidgway:
pulling to review

GaryRidgway:
I got all the way through the testing until I got to testing admissions, and then I was unable to test because I could not get the events to display in the view, and i think that may be broken right now. Everything before that looked okay though.

I do agree there could be some styling done for the details, but i think that could be done in another slice as well. 

I am going to unassign myself for now.

pyrello:
It sounds like this PR does not create the regression on Admissions. While we should definitely fix that, I don't know that this needs to be held up until that bug is fixed.

bspeare:
Here's an updated look at the details toggle. 

https://github.com/uiowa/uiowa/assets/1036433/323e07a7-7e57-41a0-8bcb-bfd1cd62376e

GaryRidgway:
> It sounds like this PR does not create the regression on Admissions. While we should definitely fix that, I don't know that this needs to be held up until that bug is fixed.

If this is correct, then I have been able to test everything, and this is all good to go!

GaryRidgway (approved):

cory-skeers:
> > It sounds like this PR does not create the regression on Admissions. While we should definitely fix that, I don't know that this needs to be held up until that bug is fixed.
> 
> If this is correct, then I have been able to test everything, and this is all good to go!

Can confirm--I started poking around with it off of main and the issue is there, so unrelated to this PR.

Issue #6073, opened 2023-01-05
Use an alternate date format for published date on some articles


briand44:
# Story

As a site owner for Iowa Law Review, I would like our print only news articles to only output month and year for the published date rather than day, month and year.

# Additional Info

>We were wondering if it would be possible to add articles to the ILR website without inserting specific dates, as some of them are only listed with a month/year. We aren't able to publish articles without inserting a month, day, and year.

(e.g. November 2022 rather than Tuesday, November 15, 2022) https://lawilr.prod.drupal.uiowa.edu/print/volume-108-issue-1/presidential-motive

# Proposed Solution

pyrello:
With regard to the comment from the customer, the answer is likely no. I believe that the D8+ date ecosystem requires a day be entered. This is separate from how it is displayed.

briand44:
Thanks @pyrello. I told them that we likely wouldn't change how the dates are entered and they should continue using a placeholder date (currently 15th of each month I believe). I said we would discuss displaying only month/year and follow up with them once we have discussed as a team.

cory-skeers:
Related feedback from Iowa Now:

> Can the publish date conform to our style? (the way we abbreviate months?) For example: Tuesday, Feb. 14, 2023, etc. (Abbreviate Jan., Feb., Aug., Sept., Oct., Nov., Dec.; spell out March, April, May, June, July)

This matches AP style guide, which I believe is the fallback university standard for anything not explicitly listed in our brand style guide, so we may want to update the default date display platform-wide as part of this.

Relates to #5974

PR #6626, opened 2023-07-10
OnIowa - events grouped by date

### LINKED ISSUE #6607: Group events by date
# Story

As an owner of the OnIowa website, we would like our events page to display the events grouped by date so that users can quickly glance to see where one days events end and another begins.

# Additional Info

Previously displayed like https://web.archive.org/web/20220815131834/https://oniowa.uiowa.edu/events

Current display https://oniowa.uiowa.edu/events

# Proposed solution

- Override default events view for OnIowa to group by date?
- Updates to event teaser display (show time rather than date and time)

### LINKED ISSUE #6646: Only show upcoming events for OnIowa
# Story

As an OnIowa site owner, I would like to only see events occurring today or future days so that users don't have to sift through events that have already taken place.

# Additional Info

Currently past events are showing https://oniowa.uiowa.edu/events

# Proposed Solution

- Limit events display to show only current day and upcoming
- possibly limit cache lifetime of event day filter

e-marie-w:
Resolves #6607 
Resolves #6646 

# How to test
```
ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli events
```
- See events are grouped together under date headings.
- Check that the time only - no date - is displaying for each individual event under a date. Uunless it is a multiday event, in which case the entire start date and end date should display instead.
- Click on an event and make sure the entire date and time still displays on its page. Add a presenter to an event and check the entire date and time still displays on their page too.
- Date format should now resemble Monday, July 17, 2023 for the date headings and the teasers - including on an individual event's page and a presenter's page.

## Date option/form caching
- Set a breakpoint at line 22 in `oniowa_core.module`.
- Turn on "Start Listening for PHP Debug Connections"
- Change line 42 to:
```php
$expire = strtotime('now') + 60;
```
```
ddev drush @oniowa.local cr && ddev xdebug
```
- https://oniowa.uiowa.ddev.site/events
- Note that breakpoint on line 22 is triggered.
- Set a timer for 60 seconds.
- Immediately refresh the page. Breakpoint should not be triggered.
- After the timer goes off, refresh the page again. Breakpoint should be triggered again.


bspeare:
I added a date range for one event.  Do we want a better fallback date display for this?

Screenshot 2023-07-13 at 10 39 53 AM

e-marie-w:
> I added a date range for one event. Do we want a better fallback date display for this?
> 
> Screenshot 2023-07-13 at 10 39 53 AM

I think it was mentioned that OnIowa doesn't have multi-day events. But I can work on a fallback just in case.

bspeare:
Tested this with `devel generate` and it looks like the date formatting resets once the pager is used. 

Screenshot 2023-07-13 at 11 05 27 AM

bspeare:
Do we want the date format heading to match what we are using in our date teasers? `Tuesday, July 11 2023` vs `Tuesday, July 12, 2023`?


Screenshot 2023-07-13 at 11 13 36 AM

e-marie-w:
> Tested this with `devel generate` and it looks like the date formatting resets once the pager is used.
> 
Good catch! Their old site had no pager so I completely forgot about it. 

I have noticed that to trigger xdebug I have to run cache rebuild first, even if I haven't made any code changes. So now I'm wondering if node_view only runs once, and maybe that's why the formatting resets and I need to run `cr` constantly when debugging. 

> Do we want the date format heading to match what we are using in our date teasers? Tuesday, July 11 2023 vs Tuesday, July 12, 2023?

This I can change, but I thought the extra comma was the more proper way? It's one of the default date options available. I can add a custom format if we want the consistency though.

joewhitsitt:
Bummer. I think we are going to have to approach this differently. I am going to poke a bit at the date format issue.

pyrello:
I'm taking a look at this.

pyrello:
I think I'd like to see being able to hide the relevant date parts in `oniowa_core_preprocess_views_view_unformatted`. At that point, we don't have access to any of the fields yet. One way to handle this would be to refactor our `#hide_fields` logic to be able to allow us to point to nested elements. The core `FormState` class does something like this using values in the form of `field_event_when][0][start][#text][date][value][markup`, which then gets exploded on `][` and parsed down.

pyrello:
If we don't include the date modifications (removing the date part) from the teaser display, does this work as expected otherwise? Seems like it might make sense to let it go without that and take the time to do the refactor indicated in my last comment without holding up this update.

joewhitsitt:
>Unknown "smart_date_format" filter. 

Ugh. Travis. Worked locally..

Don't suppose anyone knows how to get validate:twig to be aware of contrib smart_date

pyrello:
We probably need to display the year in the drop-down too, unless we are going to try to limit it to only display events in the same calendar year.

Screenshot 2023-07-14 at 3 28 22 PM

cory-skeers:
> This I can change, but I thought the extra comma was the more proper way? It's one of the default date options available. I can add a custom format if we want the consistency though.

```
Use a comma after the year in a complete date that appears in continuing text.
  He received his appointment June 10, 1998, and was promoted a year later.
Do not use a comma before or after the year when only a month and year are used.
  He announced his team’s discovery at the August 2007 meeting of the UFO Search Society.
```

The UI editorial guide doesn't address it directly, but in the examples, it looks like it should have the comma between day and year like `June 10, 1998`

pyrello:
> I can add a custom format if we want the consistency though.

Can we just update the existing date formatter to use the correct format? I think that would cause the date to be updated to the correct display format anywhere that formatter is used.

cory-skeers:
> The UI editorial guide doesn't address it directly, but in the examples, it looks like it should have the comma between day and year like June 10, 1998

And quick followup: If not specified in the UI editorial style guide, the fallback is AP style, and while I don't have an account or copy of the book anymore to look up directly, this appears to be the current ruling on that:

> Commas are not necessary if only a year and month are given, but commas should be used to set off a year if the date, month and year are given. Use the letter s but not an apostrophe after the figures when expressing decades or centuries. Do, however, use an apostrophe before figures expressing a decade if numerals are left out. Examples: Classes begin Aug. 25. Purdue University was founded May 6, 1869. The semester begins in January. The 1800s. The ’90s.


All that to say, I think we should update whichever end to have the `Day, Month Date, Year` format.

joewhitsitt:
So there are core default formats and smart date formats. /admin/config/regional/smart-date

doing a search in config `date_format` and `type: smartdate` we have quite a mix.

e-marie-w:
I'm wondering if the fact the teaser has the month abbreviated makes a difference. Teaser display doesn't have a Monday, Jul 17, 2023 option. But the view doesn't have a Monday, July 17 2023 option either. With the view I can set a custom format though.

pyrello:
This is the format we want, right? https://github.com/uiowa/uiowa/blob/main/config/default/core.date_format.medium.yml

pyrello:
If we are using smart date formats, then it looks like we need to update https://github.com/uiowa/uiowa/blob/main/config/default/smart_date.smart_date_format.default.yml and https://github.com/uiowa/uiowa/blob/main/config/default/smart_date.smart_date_format.date_only.yml

cory-skeers:
> This is the format we want, right? https://github.com/uiowa/uiowa/blob/main/config/default/core.date_format.medium.yml

That looks correct to me.

There's may be some additional followup work we could do platform-wide for abbreviated months in places following AP styleguide, like in https://github.com/uiowa/uiowa/issues/6073, but I think outside of that, this is the closest match.

e-marie-w:
I updated the date formats through the UI, but when I `drush cex` the changes are added as config split patches to the event feature. That's not where the changes should go, correct?

joewhitsitt:
>That's not where the changes should go, correct?

Can you provide a git diff?

e-marie-w:
> > That's not where the changes should go, correct?
> 
> Can you provide a git diff?

It creates the files
`config/features/event/config_split.patch.smart_date.smart_date_format.date_only.yml
config/features/event/config_split.patch.smart_date.smart_date_format.default.yml`
and for both it is
```
adding:
  date_format: 'l, F j, Y'
removing:
  date_format: 'D, M j Y'
```

joewhitsitt:
yeah, these shouldn't be patch files. it should be changing the existing config in default. you will need to do a `cex` instead of the split-export.

Here is PROD currently:
Screenshot 2023-07-17 at 10 44 26 AM

`l, F j, Y` matches admin/config/regional/date-time/formats/manage/medium

e-marie-w:
Running`ddev drush @oniowa.local cex` is what's giving me the patch files. The complete list I get is this:
```
Changes not staged for commit:
        deleted:    config/default/core.entity_view_display.node.person.default.yml
        modified:   config/default/core.extension.yml

Untracked files:
        config/features/event/config_split.patch.smart_date.smart_date_format.date_only.yml
        config/features/event/config_split.patch.smart_date.smart_date_format.default.yml
        config/features/event/config_split.patch.views.view.events_overview.yml
```
The core.extension change is removing the smart date recurring module, and the events_overview is adding the field_event_geolocation field. 

I'll keep trying things.

joewhitsitt:
Hmm.

Make sure you are removing the patch files before trying cex. Yes, there is some unintended config that comes with cex, but you will discard anything that was not intended.

e-marie-w:
I switched to changing the files directly, and so far it seems to be working.

cory-skeers:
Tested the display and the updates are looking great! Tried to throw a few curveballs like a 2 hour event starting at 11pm (displays full dates and times and works well), and an in-progress event that started prior to but extends after today (appears in the display, which in this context I think is accurate), and past events (do not display).

Also appreciate the extra work in updating the date formats. Those are looking good 👍

GaryRidgway:
Reviewed the caching portion, and it is looking good! I would say that with Cory's review and mine, we can probably call this approved.

GaryRidgway (approved):

e-marie-w:
Thanks for testing @cory-skeers and @GaryRidgway!

PR #6642, opened 2023-07-12
Uiowa profiles - setttings page php fix

joewhitsitt:
# To Test

On `main`, sync down the default site
config-split:activate profiles
Go to admin/config/sitenow/uiowa-profiles and click the new bucket button. see error.

pull down `hotfix-profiles-settings` and `cr`
initial fieldset should show and be fillable.
evaluate if anything in the validate function is needed on that first run.


pyrello:
Pulling to test.

pyrello (approved):
There are a couple things I'm not totally sure about in the validation code, but this fix works as expected.

PR #6653, opened 2023-07-14
[Classrooms] Scheduled Regions updates

### LINKED ISSUE #6623: Ability to designate level one rooms
# Story

As a Classrooms website owner, I would like the ability to differentiate between University Classrooms and Level 1 University Classrooms because scheduling policies and procedures are different between the two room categories.

# Additional Info

They currently have setup Level 1 as a scheduling region. This will work for them if we allow multiple scheduling regions to be assigned to a single room.

Currently the field allows multiple but I believe our code is only assigning one value from the API.
https://api.maui.uiowa.edu/maui/api/pub/registrar/courses/AstraRoomData/Abw/112 - The region list contains 1) University Classrooms - Level 1 and West but if you look at https://classrooms.prod.drupal.uiowa.edu/abw-112, it is only assigned Level 1 and not West.

Currently we link scheduling regions to a single basic page. Ideally, Classrooms doesn't want Level 1 to be linked to that page but could live with it if it needs to.

Outputting multiple regions as a comma separated list is fine.

# Proposed Solution

- pull in multiple values that match term mappings
- output multiple regions on a room as comma separated
- don't link level 1

cory-skeers:
Resolves #6623 

# How to test

```
ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local classrooms-rooms && ddev drush @classrooms.local uli
```

Check for rooms that have both "Level 1" and at least another scheduling region
Check that on the specific room display, "Level 1" should not be linked, bot other scheduling regions should link to the scheduling regions page

### DIFF
--- a/config/sites/classrooms.uiowa.edu/core.entity_view_display.node.room.default.yml
+++ b/config/sites/classrooms.uiowa.edu/core.entity_view_display.node.room.default.yml
@@ -484,7 +484,7 @@ third_party_settings:
                   link: true
                 third_party_settings:
                   field_delimiter:
-                    delimiter: ''
+                    delimiter: ','
             weight: 2
             additional: {  }
             third_party_settings: {  }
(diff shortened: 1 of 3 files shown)


bspeare:
Pulling to test.

bspeare (approved):
Works great!

PR #6656, opened 2023-07-14
July 14 support provisions

GaryRidgway:
# How to test


docroot/sites/bnelson.lab.uiowa.edu/blt.yml
```diff
@@ -0,0 +1,18 @@
...
+    local: self
+    remote: labbnelson.prod
+drupal:
+  db:
+    database: bnelson_lab_uiowa_edu
+uiowa:
+  stage_file_proxy:
+    origin: 'https://labbnelson.prod.drupal.uiowa.edu'
+  requester: bnelson3
+  site-name: bnelson3
```
joewhitsitt, line 18: did the provision script mess this up or was this actually the value for site name?

cory-skeers (approved):
I kind of wish they'd asked for choirauditions.org.uiowa.edu instead of uiowachoirauditions.org.uiowa.edu, but I don't think the full one is a problem outside of being long with an extra uiowa.

docroot/sites/bnelson.lab.uiowa.edu/blt.yml
cory-skeers: It's in the request form as that. Not a good site name to launch with, but the provision script functioned as expected with it.

GaryRidgway: i think that is the correct value, as the hawkid is `bnelson`

joewhitsitt:
wait, we are okay with the double iowa?

GaryRidgway:
I'm pinging Michael now, Hostmaster didn't flag it, but I can undo if necessary before release

bspeare:
It's our job to flag the double uiowa :)

GaryRidgway:
MJ said it is good on an org site, so we are good to go

PR #6657, opened 2023-07-14
[Admissions] Added subtitle field to pdf template 

### LINKED ISSUE #5858: Add subtitle to Area of Study PDFs
# Story

As an Admissions website owner, I would like to display the Area of Study subtitle on the PDFs.

# Additional Info

>We would like the “subtitle” areas of study field to be added to the PDF version if there is content. We don’t use this a lot now, but will be bringing that into more in the near future. It can show up under the main heading in larger type than the body copy.

Since they will be using the subtitle for the AoS display at the visitor center, they will be using this field more and increases the desire to display on printable pdfs.

We have flexibility to figure out what looks best here.

Here is an example of an Area of Study that contains subtitle info.
https://admissions.uiowa.edu/node/2716/edit
https://admissions.uiowa.edu/print/pdf/node/2716

# Proposed solution

- Add subtitle to printable pdf, which could cause multiple pages (need to print in twig file)
- Share with Scott so they can be aware they may need to adjust content to keep within single page

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5858. 

# How to test

```
ddev yarn workspace admissions_core build && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli print/pdf/node/2786
```

Edit subtitle field on https://admissions.uiowa.ddev.site/node/2786/edit and try italic, underline, and a link and verify they are styled in the PDF. 

It should look like:

[1019 (1).pdf](https://github.com/uiowa/uiowa/files/12071177/1019.1.pdf)

### DIFF
--- a/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/pdf.scss
+++ b/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/pdf.scss
@@ -468,6 +468,19 @@ h6 {
     }
   }
 
+  .field--name-field-area-of-study-subtitle {
+    margin-bottom: 10px;
+    p, strong, em {
+      @extend .element--light-intro;
+      line-height: 1;
+      font-size: 1rem!important;
+      margin-bottom: 10px;
+    }
+    em {
+      font-weight: bold;
+    }
+  }
+
   .field--name-field-area-of-study-competitive,
   .field--name-field-area-of-study-four-year,
   .field--name-field-area-of-study-honors,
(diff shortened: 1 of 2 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
Testing instructions checkout. I assume they are okay with bold not having an effect on the PDF version? superscript also works

Screenshot 2023-07-17 at 11 29 49 AM


Screenshot 2023-07-17 at 11 27 21 AM

PR #6662, opened 2023-07-17
[Alerts] Ability to schedule publishing/unpublishing 

### LINKED ISSUE #6637: Ability to schedule publishing/unpublishing alert content
# Story

As an editor using the SiteNow alert feature, we would like to be able to schedule alerts to be published/unpublished at a period of time in the future.

# Additional Info

Summary from probe https://github.com/uiowa/uiowa/issues/6585#issuecomment-1631098306

Scoping to alert feature not just Facilities.

Alert feature does not currently have content moderation so we don't currently need https://www.drupal.org/project/scheduler_content_moderation_integration.

We would also want to remove the field that is part of the Facilities split that is related to publishing date. 

# Proposed solution

- Use https://www.drupal.org/project/scheduler to allow scheduling when to publish/unpublish alerts for the alert feature
- Verify publishing happens within an hour of chosen time
- Verify we are not seeing any performance concerns
- Remove the field on the Facilities split that is related to when to publish
- Communicate with Facilities that it will publish/unpublish within an hour and that if we run into issues of performance down the road, they may need to manually publish/unpublish alerts while we chart a new path.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6637

# How to test

1. Schedule an alert to publish/unpublish.   
2. Run cron and verify that alert has been published or unpublished. 

## Test on default
```
ddev composer install && ddev blt ds --site=default && ddev drush @default.local config-split:activate sitenow_alerts -y && ddev drush @default.local uli node/add/alert
```
- Verify that editor, publisher, webmaster can all schedule alerts. 

## Test on facilities
```
ddev composer install && ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli node/add/alert
```
- Verify that editor, publisher, webmaster can all schedule alerts + collaborator.
- Verify that custom publish date field `field_alert_publish_on` has been removed from facilities site split.


joewhitsitt:
Pulling to test

joewhitsitt:
Wondering if editor should have the ability to publish/schedule alerts? Normally their role does not allow publishing of content. They can't see the published checkbox (in place of moderation) but upon save and a blank date, they are able to publish the content immediately.

Also wonder if webmaster and publisher should have the `Override Alert published option` so they can create an alert without publishing and can un-publish without scheduling.

bspeare:
Great questions! Thanks for pointing those out.

joewhitsitt:
In my mind, I think the content type default should be published (unchecked), publisher and webmaster have the scheduling capability and the override publish option. editors can still create alerts but cannot publish. And/or similar disabled/help text as Facilities currently has to point out the unique pattern here compared to other content types (default/custom)

Screenshot 2023-07-18 at 3 41 03 PM

joewhitsitt:
Looks like collaborator already has the override permission (default as published checked)

Screenshot 2023-07-18 at 3 45 35 PM

bspeare:
@joewhitsitt this should be ready to test again with your suggested changes.

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -654,6 +654,16 @@ function uiowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id)
...
       break;
 
+    // Update help text for scheduler on alert content type.
+    case 'node_alert_edit_form':
+    case 'node_alert_form':
+      $additional_published_help_text = t('Note: this alert will be published within a one-hour window of the selected date and time.');
+      $additional_unpublished_help_text = t('Note: this alert will be unpublished within a one-hour window of the selected date and time.');
+      $form['publish_on']['widget'][0]['value']['#description'] .= '<p>' . $additional_published_help_text . '</p>';
+      $form['unpublish_on']['widget'][0]['value']['#description'] .= '<p>' . $additional_unpublished_help_text . '</p>';
+      $form["status"]["widget"]["value"]["#description"] = t('Use <em>Scheduling options</em> to publish this content');
```
joewhitsitt:
```suggestion
      $form["status"]["widget"]["value"]["#description"] = t('Optional: Use <em><a href="#edit-scheduler-settings">Scheduling options</a></em> to publish this content');
```

Screenshot 2023-07-19 at 2 21 12 PM

joewhitsitt:
Ack. Maybe editor should just not be allowed to use this content type? Once saved (unpublished), without moderation, it is unclear the next step should be for the editor.

bspeare:
@joewhitsitt removing access for editors sounds like a good solution for the time being.  Maybe can revisit if we have a bunch of editors needing to create alerts?

joewhitsitt:
I am okay with that. So just delete config_split.patch.user.role.editor.yml? They can stack collaborator on facilities for those editors that need to add alert content. I think there is more unknown when it comes to the default feature.

Everything else in mind mind looks good

joewhitsitt (approved):
Thanks for adjusting. Looks good to me

bspeare:
I moved the scheduling fieldset to sit above the published checkbox if someone wants to pull this down and verify that it looks like https://github.com/uiowa/uiowa/issues/6637#issuecomment-1642768464.

pyrello:
Pulling to check the form move.

pyrello:
Approving the form changes. Nice work on this!

bspeare:
Thanks for confirming!

PR #6671, opened 2023-07-19
[Alerts] Current alerts block

### LINKED ISSUE #6658: Current/Active alerts block
# Story

As a SiteNow site owner using the alerts feature, I would like to be able to place a block that shows current/active alerts so that users can be aware of these.

# Additional Info

Facilities existing example https://www.facilities.uiowa.edu/alerts

They said they can be a little flexible here with how things are output, possibly teaser with title, dates, categories, that links to full node

Alerts/closures should no longer display once the end date has passed

These should stay published though, just not displayed

# Proposed Solution

cory-skeers:
Resolves https://github.com/uiowa/uiowa/issues/6658

# How to test

`composer install` if you haven't recently

Default
```
ddev blt ds --site=default && ddev drush @default.local config-split:activate sitenow_alerts -y && ddev drush @default.local uli
```

Facilities
```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli
```

- Add some events. Past, current, future. Categorize.
- Confirm custom help text from previous work is still displayed with the scheduling options and the published checkbox.
- As a webmaster, add the Alerts list block
- Note that only currently active, published alerts are displayed sorted by start time, earliest to latest (after sticky)
- Try filtering by alert categories.
- More link should work with custom path, otherwise defaults to `<front>`
- Try hiding fields.
- Review teaser styling.


joewhitsitt:
>It's sorted by alert start time, earliest to latest (after sticky). This, or alert end time, or something else?

I think until we discover more this is a good place to start (earliest to latest (after sticky)). The facilities site is currently sorted by posted date which just seems out of order to me without severity or something else first.

ITS separates active, scheduled (7-days) and service announcements to make this more clear to users but also provides a calendar view listing all together (color-coded, labelled). The active/service announcements are based on severity, then post date. scheduled is by start date (earliest).

joewhitsitt:
pulling down to look it over

pyrello:
Pulling to test

pyrello:
Just noting that the "Scheduling options" link in the help text under the "Published" checkbox seems less useful now that the scheduling options are displayed right above them. It kind of seems like nothing happens when you click it.

bspeare:
@pyrello maybe it would scroll you up on mobile?

pyrello:
> @pyrello maybe it would scroll you up on mobile?

On mobile, it may actually look like something happens and that the link isn't broken. My feedback is about the experience on Desktop.

When you make an item sticky, there is nothing to indicate it's stickiness to the user, so it just looks like the items are in an order that doesn't match what was chosen.

joewhitsitt:
I can remove the link if you think we should.

As for sticky, I wonder if a separate issue to revisit our sticky representation would be better as we have implemented the functionality on other content types previously and haven't provided any kind of indication either.

pyrello:
> As for sticky, I wonder if a separate issue to revisit our sticky representation would be better as we have implemented the functionality on other content types previously and haven't provided any kind of indication either.

This is in line with my thinking.

> I can remove the link if you think we should.

Would it be crazy to only have the link render below a certain breakpoint?

joewhitsitt:
>Would it be crazy to only have the link render below a certain breakpoint?

I'd rather rip it out. I originally suggested because of the disconnect between the published checkbox and the sidebar. I think sites that use this content type are going to be training users so it will lose effect quickly.

joewhitsitt:
Created https://github.com/uiowa/uiowa/issues/6695. Feel free add/adjust

pyrello (approved):
This seems to work as expected. Nice work!

PR #6703, opened 2023-07-26
Add a unique ID to sections

### LINKED ISSUE #6680: P2LB - Add ID's to sections
To account for the fact that sections were able to add unique HTML ID's in V2, we need to add that functionality in V3, so we have somewhere to put that data.

# Proposed solution
- Alter the section form to add the unique ID field
- Process the form submission to store the unique ID in the section's third party settings
- Preprocess layouts to add the ID to the HTML attributes
- Update layout templates if the above item doesn't just cause the ID to be rendered in HTML automatically

pyrello:
Resolves #6680 

# How to test

## V3
```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli node/1221/layout
```
- Click an "Add section" link
- Add a unique ID and Save
- Inspect HTML to check that the ID is being added to the layout container div
- Hover over a different existing section to show the controls and click "Configure..."
- Add a unique ID and click "Update"
- Inspect HTML to check that the ID is being added to the layout container div
- Save the layout
- Inspect the HTML to check that the ID is being added to both layout container divs

## P2LB conversion
```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local uli admin/content/sitenow-converter
```
- Find the "Messaging" page, click the checkbox next to it and run the conversion
- Visit the layout tab of the newly converted node: https://brand.uiowa.ddev.site/node/56/layout
- Consult https://brand.uiowa.edu/node/56/edit to see which sections have unique HTML ID's
- Check if the unique ID was set for which it was supposed to be set and not set for the rest. Some V2 (paragraph) sections get split into multiple (layout) sections in V3. In that case, the unique ID will only be set for the first instance.


GaryRidgway (approved):
Both Cory and I have huddled to review this, and we both agree that this seems to function as  intended. Nice!

PR #6708, opened 2023-07-28
P2LB reporting tool

### LINKED ISSUE #6735: P2LB - Add reporting tool to allow users to see potential degredation or lost data
# Motivation

We have identified some cases where we don't believe that it is worth the effort to find a solution for a mismatch between V2 and V3 components. As such, we have decided to allow some content to be converted into a state that is degraded or data is lost. To help customers understand this, we would like to provide some type of reporting tool. A proof of concept was already developed in #6708.

# Proposed solution
- Provide mechanism for analyzing a node and providing feedback about possible issues
- Implement a tab for viewing the report
- Provide warnings on the node form about possible issues

cory-skeers:
Resolves #6735 

# How to test

```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local cim -y && ddev drush @brand.local uli node/4406/edit
```
- Observe the warning.
- Click the "V3 Status" tab and observe output. It should match up to https://github.com/uiowa/uiowa/pull/6708#issuecomment-1662863485
- Convert the page.
- See the status is updated on the "V3 Status" tab.
- See that the warning no longer shows on the "Edit" tab.
- The idea with this is not that it is complete, but that the experience of using it is good enough to merge it so we can add more too it later.

### DIFF
--- a/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.links.task.yml
+++ b/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.links.task.yml
@@ -8,3 +8,8 @@ sitenow_p2lb.content_converted:
   title: "Converted"
   base_route: sitenow_p2lb.content_converter
   weight: 100
+
+entity.node.conversion_status:
+  route_name: entity.node.p2lb_conversion_status
+  base_route: entity.node.canonical
+  title: 'V3 Status'
--- a/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
+++ b/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
@@ -1207,6 +1207,7 @@ function _sitenow_p2lb_block_styles($block_type, $paragraph = NULL) {
         'card_image_medium',
         'list_format_list',
         'block_grid_threecol_33_34_33',
+        'no_border',
       ];
 
     case 'uiowa_quote':
--- a/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
+++ b/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
@@ -13,6 +13,7 @@ use Drupal\Core\Url;
 use Drupal\layout_builder\Field\LayoutSectionItemList;
 use Drupal\layout_builder\InlineBlockUsage;
 use Drupal\layout_builder\Section;
+use Drupal\sitenow_p2lb\P2LbHelper;
 
 /**
  * Implements hook_form_alter().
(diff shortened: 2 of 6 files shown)


pyrello:
I'm thinking that maybe we should turn this PR into the reporting tool PR instead of an all purpose PR. I think having it this way is confusing to people.

joewhitsitt:
pulling to test

joewhitsitt:
Here are some thoughts:

- Can the purpose of this page be expanded to show not only issues, but whether the page has already been converted (link to revision), "This page has been converted to v3"? In that case, can the title be "Conversion status" or "Conversion report"?
- Seeing the same messages appear multiple times can be daunting. Without identifiers, is there much value in showing them repeatedly or could they be consolidated with a count and displayed using the messenger functionality?
- Can there be an explainer at the top? "Conversion issue(s) identified with this page's content. They don't prevent the conversion of the page but..."
- Can the message on the edit form link to the tab?
- Can we link out to a page on SiteNow as describe here if there are identified issues: https://github.com/uiowa/uiowa/issues/6735#issuecomment-1662618157

bspeare:
Great feedback from @joewhitsitt.  I agree that consolidating with a count would be the best approach.

pyrello:
Ready for review again. For realzies this time.

joewhitsitt:
>Ready for review again. For realzies this time.

I will pull it down and take another look. EoD was looking good.

bspeare:
Is there a way in the "V3 Status" tab to indicate that the page has been converted and needs to be moved from draft to published in the "Latest Version" tab?
Screenshot 2023-08-03 at 8 59 42 AM

pyrello:
> Is there a way in the "V3 Status" tab to indicate that the page has been converted and needs to be moved from draft to published in the "Latest Version" tab?

It looks like we'll need to some logic to determine whether the most recent revision is the current revision and if not, to base the status on the most recent revision instead.

joewhitsitt:
> > Is there a way in the "V3 Status" tab to indicate that the page has been converted and needs to be moved from draft to published in the "Latest Version" tab?
> 
> It looks like we'll need to some logic to determine whether the most recent revision is the current revision and if not, to base the status on the most recent revision instead.

Beyond this ^^, I think generally this PR is offering something we can build upon further. Do we want to try to get this to work or can I approve the PR as-is?

pyrello:
> Beyond this ^^, I think generally this PR is offering something we can build upon further. Do we want to try to get this to work or can I approve the PR as-is?

Let's move ahead as is. We need to add some additional warnings, but I think that can happen in future slices as we are refining final product.

joewhitsitt (approved):

PR #6709, opened 2023-07-28
[hotfix] Smart trim hotfix

cory-skeers:
Smart trim update broke config syncing. Hotfix to update and prepare for deployment.

# How to test

Sync a v2 site, don't run into config errors.

### DIFF
--- a/config/features/sitenow_v2/config_split.patch.core.entity_form_display.node.page.default.yml
+++ b/config/features/sitenow_v2/config_split.patch.core.entity_form_display.node.page.default.yml
@@ -51,3 +51,8 @@ removing:
       weight: 2
     moderation_state:
       weight: 16
+    simple_sitemap:
+      weight: 10
+      region: content
+      settings: {  }
+      third_party_settings: {  }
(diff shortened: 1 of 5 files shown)


GaryRidgway (approved):
Observed the sync not having an issue

config/features/sitenow_v2/config_split.patch.core.entity_form_display.node.page.default.yml
```diff
@@ -51,3 +51,8 @@ removing:
       weight: 2
     moderation_state:
       weight: 16
+    simple_sitemap:
+      weight: 10
+      region: content
+      settings: {  }
+      third_party_settings: {  }
```
pyrello: This seems like it should be part of the parent config that is being patched.

Edit: Seeing now this is in the `removing` section. I think we should just remove this so that its not different.

cory-skeers:
```suggestion
```

pyrello:
```suggestion
```

PR #6712, opened 2023-07-28
Turn off revision deletion in P2LB

### LINKED ISSUE #5633: P2LB - Prevent deletion of rollback revision
# Motivation
When the conversion of the page node happens, the revision ID for the revision that existed prior to the conversion is stored in a special field on the node so that it can be used to rollback converted nodes. We need to prevent the deletion of these revisions in the system until we have completed transitioning all V2 sites.

# Proposed solution
- [ ] Implement code similar to our front page protection that prevents the deletion of the revision manually.
- [ ] The oldest revisions greater than the last 10 automatically get pruned by the system when cron runs. We need to figure out what hook we need to implement or event we need to subscribe to to prevent this revision ID from being deleted automatically.

GaryRidgway:
Remove revision limit for pages where p2lb is enabled

# How to test

`ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local uli
`
Check the page content type
ensure that in publishing options, revisions are not limited

### DIFF
--- a/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
+++ b/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
@@ -17,84 +17,96 @@ use Drupal\layout_builder\Section;
 /**
  * Implements hook_form_alter().
  */
-function sitenow_p2lb_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  switch ($form_id) {
-    case 'node_page_form':
-    case 'node_page_edit_form':
-      $entity = $form_state->getFormObject()->getEntity();
+function sitenow_p2lb_form_alter(&$form, FormStateInterface $form_state, $form_id)
+{
+  if (in_array($form_id, [
+    'node_page_form',
+    'node_page_edit_form',
+  ])) {
+    $entity = $form_state->getFormObject()->getEntity();
 
-      if ($form_id === 'node_page_form') {
-        // Disable display of the Paragraph sections for new content.
-        $form['field_page_content_block']['#access'] = FALSE;
-      }
-      elseif ($form_id === 'node_page_edit_form') {
-        // Check if field_v3_conversion_revision_id is set,
-        // and if the user hasn't yet interacted with the form
-        // (so that we don't duplicate our messages).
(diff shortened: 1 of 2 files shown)


pyrello:
> Implement code similar to our front page protection that prevents the deletion of the revision manually.

This PR does not satisfy this requirement from the linked issue.

cory-skeers:
> > Implement code similar to our front page protection that prevents the deletion of the revision manually.
> 
> This PR does not satisfy this requirement from the linked issue.

For reference here's the code that handles preventing deleting the homepage:
https://github.com/uiowa/uiowa/blob/3558b3161fcc40f3b7aa45c00bb96a70e6fae4b8/docroot/profiles/custom/sitenow/sitenow.profile#L231-L259

And a similar example with Iowa Now preventing deletion of their article category taxonomies:
https://github.com/uiowa/uiowa/blob/3558b3161fcc40f3b7aa45c00bb96a70e6fae4b8/docroot/sites/now.uiowa.edu/modules/now_core/now_core.module#L233-L271

Might be a little different preventing specific revision deletions, though.

docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
```diff
@@ -18,40 +18,87 @@ use Drupal\layout_builder\Section;
...
+      if(!$most_recent_revision_id) {
+        break;
+      }
+      $most_recent_revision = $node_storage?->loadRevision($most_recent_revision_id);
+      if(!$most_recent_revision) {
+        break;
+      }
+
+      // Get protected revision Id from that revision's values.
+      $protected_revision_id = $most_recent_revision?->get('field_v3_conversion_revision_id')?->getValue()[0]['value'] ?? null;
```
pyrello:
```suggestion
      $protected_revision_id = $most_recent_revision->field_v3_conversion_revision_id?->value;
```
This is equivalent.

e-marie-w (approved):
Tested three different pages, can successfully save more than ten revisions, cannot delete last revision before conversion and get expected error message when I try.

Issue #6706, opened 2023-07-28
Get P2LB ready for customer testing


pyrello:
We are going to get P2LB ready for customer testing by sites who are interested in early adoption.

## User stories
- As a v2 site owner, I would like to be able to migrate a single page to its v3 version so that I can see how it works and get comfortable with it
- As a v2 site owner, I would like to be able to migrate multiple/all of my pages to their v3 versions so I can check to make sure that they are ready
- As a v2 site owner, I would like to be able to revert a single page from its v3 to its v2 version so that I can easily undo all changes to a page since it was migrated.
- As a v2 site owner, I would like to be able to revert multiple/all of my pages in case I need to start over.
- As a v2 site owner, I would like to be unable to delete the revision of my page that allows me to revert it to its v2 version.

```[tasklist]
### Tasks
- [ ] #6680
- [x] Update block styles - @bspeare #6716
- [ ] #5633
- [ ] https://github.com/uiowa/uiowa/issues/6714
- [ ] https://github.com/uiowa/uiowa/issues/6733
- [x] - ListBlock (Articles and People): Update `sitenow_p2lb_process_paragraph` to set tags in new block based on tags in paragraph settings. Added todo comments in #5613. (@cory-skeers)
- [ ] #6681
- [ ] https://github.com/uiowa/uiowa/issues/6735
- [ ] https://github.com/uiowa/uiowa/issues/6707
- [x] Ensure that paragraphs are properly removed every time when a page is converted - Brand's "Web" page shows both blocks and paragraphs after conversion - handled in #6718
- [x] Clear tempstore on conversion and reversion (https://github.com/uiowa/uiowa/issues/6706#issuecomment-1660934479)
- [ ] https://github.com/uiowa/uiowa/issues/6859
- [ ] https://github.com/uiowa/uiowa/issues/7049
- [ ] https://github.com/uiowa/uiowa/issues/6879
```

### Previously identified issues
- ~~Cards previously did not have buttons.  Are we ok with all P2LB cards having buttons? - Ben~~ Added task above. 
- Background images/banners: Do we need to convert these to image blocks if there is no text area in use?  The default gradient gets in the way. - Sean
  -  We could do https://github.com/uiowa/uiowa/issues/6368 or image block? - Ben
- Sidebar menus: All pages need the `Page w/ left sidebar` layout with the `Main navigation` block set to `Vertical` that sidebar menus will work for converted pages. 
[...]

PR #6713, opened 2023-07-28
P2LB - V2 Card Cases

### LINKED ISSUE #6714: P2LB - Handle all V2 card cases
# Motivation

As a SiteNow v2 site owner, all the potential ways that I could have used the V2 card component will be covered by the migration process so that I don't have to worry about losing data and/or I know what data I might have to lose.

```[tasklist]
### Cases
- [ ] If there's a link + link text, and a title - link the title only and delete the link text?
- [ ] If there's a link + link text, and no title - display circle button?
- [ ] V2 card body format is less restrictive than V3 card excerpt format https://github.com/uiowa/uiowa/issues/6706#issuecomment-1656234510
```

pyrello:
Resolves #6714 

# How to test
```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local cim -y && ddev drush @brand.local uli admin/content/sitenow-converter
```
- Find existing examples or create examples of the cases below and validate what happens during the conversion.
- Optionally provide feedback about the result.

## Cases
- A card has no title and only an image or an image and a link => V3 image block
- A card has no title => V3 text block
- A card has a body with content that is incompatible with the Minimal + format => V3 text block
- Otherwise => V3 card block

### DIFF
--- a/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
+++ b/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
@@ -592,7 +592,7 @@ function sitenow_p2lb_process_paragraph($paragraph, $node) {
           $type = 'uiowa_text_area';
         }
       }
-      elseif (!P2LbHelper::formattedTextIsEquivalent($excerpt['value'], 'filtered_html', 'minimal_plus')) {
+      elseif (!P2LbHelper::formattedTextIsSame($excerpt, 'filtered_html', 'minimal_plus')) {
         // Converting to a card would result in data loss. Convert to a text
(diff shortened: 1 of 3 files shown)


joewhitsitt:
https://github.com/uiowa/uiowa/issues/6706#issuecomment-1658729205

Changing the code back seems to handle the warning.

```php
$link = ($paragraph->get('field_card_link')->getValue()) ?
        $paragraph->get('field_card_link')->getValue()[0] : '';
```

e-marie-w (approved):
Tested the various cases, all converted as expected.

PR #6718, opened 2023-08-01
[ P2LB ] V2 Revert action

### LINKED ISSUE #6707: P2LB - Create action to revert to V2 version of page
# Motivation

When we migrate a page from its v2 to v3 version, we capture the revision ID of the node before its migration so we can fall back to it. The migration is accomplished via an action that can be applied to one or many pages via the SiteNow Converter tool. This task is to create a complimentary action that reverts the migration.

## Question

If the customer migrates a piece of content and then makes changes to it, there will be multiple revision IDs after the one that we are reverting back to. Should this process simply be a nice wrapper for the option to revert to a previous revision in the revision system? Or do we need to remove any revisions after the revert point revision ID so that the user can start fresh?

# Proposed solution
- Create an action ([reference](https://github.com/uiowa/uiowa/tree/main/docroot/modules/custom/sitenow_p2lb/src/Plugin/Action)) which:
- Gets the fallback revision ID from the node (the last revision before it was converted)
- Make the fallback revision the current revision

GaryRidgway:
Tracks https://github.com/uiowa/uiowa/issues/6707

Would love to have this only show on the converted form

# How to test

```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local cim -y && ddev drush @brand.local uli admin/content/sitenow-converter
```
- Convert the "Web" page
- Check revisions to make sure that the correct revision was created with a note about the conversion
- View the latest revision tab and make sure that the header is not duplicated
- View the layout tab and ensure that the paragraph sections do not display
- Revert the page
- Make sure that everything is back to normal
- Review code changes and spot check other P2LB functionality to make sure the refactor didn't break anything


pyrello:
Working on resolving conflicts.

pyrello:
Just tested this. The workflow for this to work is:
- Convert the node
- Publish the draft
- Revert

If you don't do the publish step, the revert action seems to have no effect.

docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module
```diff
@@ -1499,3 +1463,57 @@ function sitenow_p2lb_block_definition($type, array $fields) {
...
+ * @param string $revision_id
+ *   The revision id of the protected revision of a node.
+ *
+ * @return \Drupal\Core\Entity\RevisionableInterface|null
+ *   The Entity array for the protected revision or null if not found.
+ *
+ * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+ * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+ */
+function sitenow_p2lb_get_protected_revision(string $revision_id): ?RevisionableInterface {
```
cory-skeers: From the function name/comment, I was expecting this to take an `$nid` (or `$node`) and return the loaded protected revision (or `NULL` if there wasn't one), and not a specified revision.

docroot/modules/custom/sitenow_p2lb/src/Plugin/Action/RevertToParagraphs.php
```diff
@@ -0,0 +1,91 @@
...
+    $original_revision_timestamp = date('m/d/Y - h:i A', $protected_revision->getRevisionCreationTime());
+
+    /** @var \Drupal\Core\Entity\RevisionableStorageInterface $node_storage */
+    $node_storage = \Drupal::entityTypeManager()->getStorage('node');
+
+    // Create a new revision from node storage.
+    $new_revision = $node_storage->createRevision($protected_revision);
+
+    // Set moderation state.
+    $protected_revision->set('moderation_state', 'published');
```
cory-skeers: I think this might be leftover from previous iterations? `$protected_revision` doesn't end up getting saved, and the `$new_revision` created has the same moderation state as the protected revision, leading it to match whether that had previously been published or draft. At first I thought this should set the new revision to published, but that doesn't work if the page had previously been unpublished.

I'm wondering what the desired moderation state should really be, though. In the current setup, I tried:

1. Convert an unpublished page to v3
2. Publish the page
3. Revert to v2

In this case, the reverted v2 page has a moderation state matching the previous protected revision (draft). But the v3 version remains published. We could instead grab the moderation state of the latest revision to determine if it should be draft or published, but that feels odd for the editor experience.

Right now I'm thinking this is likely the best way to set the moderation state (where reverting to the last v2 version also reverts to that moderation state). If a page were published as v3 and the editor needed to revert to v2, the option to immediately publish that version is always available.

cory-skeers:
```suggestion
```

cory-skeers: In the above example, this can get kind of funky if you convert/revert multiple times, and have a combo of 1) starting with an unpublished v2 page and 2) publish a v3 page. After you've converted and reverted, the following occurs:

1. Draft v2 page, published v3 page
2. Convert -- draft (newly converted) v3 page
3. Revert -- draft v2 page, published v3 page, page remains in the 'converted' view
4. Revert again -- draft v2 page, published v3 page, page no longer in the 'converted' view.

It still might only be an issue with trying to convert/revert over multiple times, in which case, some oddity is fine, and the option to just publish the reverted page remains available. But I do want to take a more structured approach to testing the different draft/published options that editors may actually use.

cory-skeers (approved):
There is one particular instance where it gets weird.

1. convert an unpublished page
2. publish the v3 version of the page
3. revert
4. do NOT publish the v2 version of the page
5. convert
6. revert

This goofs with the order of revisions, and the previously published v3 version of the page will remain in the 'converted' view after the unpublished v3 version is properly removed. Technically it's all accurate, because that _is_ a converted version of the page, and it still exists in the active revisions as the currently published page. It just looks funny, because you hit the 'revert' button only to see the page remain with a different revision id (and you can then ALSO revert this revision).

But at any point, if the v2 version of the page gets published, this behavior disappears.

So I think it's technically accurate functionality, just kinda weird, the fix is easy, and the situation to create it is kind of convoluted and not likely to be a typical editor pattern to encounter.

As such, approving.

Really nice work with some really confusing moderation states!

PR #6732, opened 2023-08-01
Collection Item Unique ID with p2lb support

joewhitsitt:
# To Do

- ~~Make sure descriptive list and tabs are covered?~~
- ~~Streamline the ajax callback so it isn't so spelled out and specific?~~

# To Test

```
ddev blt ds --site brand.uiowa.edu && ddev drush @brand.local config-split:activate p2lb -y && ddev drush @brand.local uli admin/content/sitenow-converter
```
- Edit a page and add an accordion item with a unique id, save.
- Convert
- See the id in the new collection item field.
- edit the value with some spaces, tab or enter off the field. the ajax should update the value.
- edit the collection's unique id field. should be unaffected.
- change to description list and the id should still output if provided.
- edit a sections unique id field. should be unaffected.

### DIFF
--- a/config/default/core.entity_form_display.paragraph.uiowa_collection_item.default.yml
+++ b/config/default/core.entity_form_display.paragraph.uiowa_collection_item.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.paragraph.uiowa_collection_item.field_automatically_open
     - field.field.paragraph.uiowa_collection_item.field_collection_body
     - field.field.paragraph.uiowa_collection_item.field_collection_headline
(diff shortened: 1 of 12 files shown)


pyrello:
I'm pulling this to check it out.

pyrello (approved):
Tested and this looks good.

joewhitsitt:
Just tested @pyrello work in https://github.com/uiowa/uiowa/pull/6732/commits/cb8ee77fdf1d297f182d6cfcb9b8560c25ce4bca

Streamlines the path and still works for the collection item, the block and section id fields. Nice work!

PR #6736, opened 2023-08-02
Post-deploy config fixes

joewhitsitt:
# To Test

- Sync Grad, SPPA, UIPress before and after this branch is checked out.
- For Google Analytics error on Education, there isn't an ID which makes it seem like they gutted it in favor of Google Tag. Maybe, uninstalling the module and re-enabling it will solve that issue if it persists on the next deployment.

# For Reference
https://github.com/orgs/uiowa/projects/20?pane=issue&itemId=34781616


cory-skeers (approved):
Reviewed, and looks like mostly updates related to the smart_trim dependency update, similar to what was added for v2. Tested, and got clean syncs after.

PR #6737, opened 2023-08-02
Hotfix - Turning off v4 shim for font awesome for a temporary fix for "add component" button

bspeare:
`ddev blt ds --site=uiowa.edu` and verify that brand fonts are loading in the footer.  Verify in the source that only the `all.min.js` is being loaded.   This will turn off font awesome 4 support which was replaced by version 5 in 2017. 

## future work

- Look to change the order of how lb_direct_add js is loaded so that something like siteimprove doesn’t slow load ahead of it


joewhitsitt (approved):
to the future!

PR #6741, opened 2023-08-03
Hotfix - Added lb_direct_add patch for popover

bspeare:
```
ddev composer install && ddev blt ds --site=education.uiowa.edu && ddev drush @education.local uli /private-sector-partnerships-strengthen-higher-education
```

Test that the popover still closes when you click outside of the button.  That is our one concern here before pushing this out.


joewhitsitt:
pulling down to test

joewhitsitt (approved):
>Test that the popover still closes when you click outside of the button.

That it does.

PR #6744, opened 2023-08-03
More Fontawesome: adjusted font awesome library usage

bspeare:
After seeing how much JavaScript we are loading in https://github.com/uiowa/uiowa/pull/6737, this pr reduces the amount of libraries being loaded for Fontawesome. 

Since we are fixed at 5.14 in https://github.com/uiowa/uiowa/blob/main/composer.json#L72, we can revisit these settings when we decide to move to 6. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /icons
```
- Compare prod with local and verify that the icons remain the same.


bspeare:
The svg with js settings don't seem to work without the giant [/libraries/fontawesome/js/all.min.js?v=6.1.0](/libraries/fontawesome/js/all.min.js?v=6.1.0)"

PR #6749, opened 2023-08-04
[Events] Update events paths

### LINKED ISSUE #6746: Request to fix event breadcrumbs from OnIowa!
We can do this by changing the event path alias pattern to `/events/[node:title]`

Scope to events feature?

cory-skeers:
Resolves #6746 

# How to test

`ddev blt ds --site=oniowa.uiowa.edu`
Visit https://oniowa.uiowa.ddev.site/events, check out some events
See that the path is `events/id/node-title`, and that as a result, breadcrumbs include a link to upcoming events

Look at a few events, note their URLs as `event/id/node-title`
Save an event
See that the path is the new `events/id/node-title` path
Visit the old path (eg `event/` instead of `events/`, and get redirected to the new path (eg the old path is still valid)
Create a new event, and see that it gets the new `events/id/node-title` path


joewhitsitt:
pulling to test

joewhitsitt (approved):
Looks good.

PR #6750, opened 2023-08-04
Alert Categories Rabbit Hole

### LINKED ISSUE #6677: Rabbit hole alert category term pages
# Story

As a web team, we would like to rabbit hole the alert feature category term pages so that folks aren't going to those pages and seeing published alerts where the alert end date is passed.

# Additional Info

If there is a use case for going to these term pages we can re-evaluate in the future.

# Proposed Solution

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6677

# To Test

```
ddev blt di --site=default && ddev drush @default.local config-split:activate sitenow_alerts
```

Go to admin/structure/taxonomy/manage/alert_categories/add and add "Test" then try and go to https://default.uiowa.ddev.site/alert-categories/test as an anonymous user.

```
ddev blt ds --site=facilities.uiowa.edu
```
Confirm there are no sync issues and that the functionality works there as well.

Disregard the config updates for smart trim and scheduler. They are harmless.

### DIFF
--- a/config/default/config_split.config_split.sitenow_alerts.yml
+++ b/config/default/config_split.config_split.sitenow_alerts.yml
@@ -10,6 +10,7 @@ stackable: true
 storage: folder
 folder: ../config/features/sitenow_alerts
 module:
+  rh_taxonomy: 0
   scheduler: 0
   sitenow_alerts: 0
 theme: {  }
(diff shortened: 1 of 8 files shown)


cory-skeers (approved):
Added `rh_taxonomy` to the split, as it wasn't enabled in default. After that, all worked well.

joewhitsitt:
Thanks for catching that. In moving swift, I missed the details...

PR #6762, opened 2023-08-09
Hotfix Linkit `<nolink>`

joewhitsitt:
- add patch for nolink support
- add matchers for front page and nolink
- remove linkit menu link support as it doesn't cover nolink


joewhitsitt:
patch seems to work for the block types. added front page matcher as well to cover that. not sure if we should worry about <button>. patch does not work for menu link, so i am removing coverage for that and will reopen the issue. 

**should we roll it all back or continue on with linkit support for block types?**

joewhitsitt:
I am going to work to roll things back as this was rogue work to begin with and it is time sensitive with provisions.

PR #6772, opened 2023-08-11
Updated Oniowa to use table instead of list

### LINKED ISSUE #6770: OnIowa events table display with attendance required
# Story

As an OnIowa site owner, we would like to see our events displayed in a table so it is easier for our users to look through all 200+ events.

# Additional Info

Should keep grouping by day

still designate which events require attendance

output time only unless event spans multiple days (I don't think any of their events span multiple days).

Example without showing attendance required and without displaying time only https://github.com/uiowa/uiowa/pull/6759#issuecomment-1670261933.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6770

# How to test

```
ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli /events
```

1. Review code changes
2. Compare against https://web.archive.org/web/20220815131834/https://oniowa.uiowa.edu/events and https://oniowa.uiowa.edu/events 
3. Try to think of ways that this display might break
4. Running `ddev drush @oniowa.local config-split:export site` seems to export a reversion to the patch - is this a concern?

### DIFF
--- a/config/sites/oniowa.uiowa.edu/config_split.patch.views.view.events.yml
+++ b/config/sites/oniowa.uiowa.edu/config_split.patch.views.view.events.yml
@@ -2,6 +2,7 @@ adding:
   dependencies:
     config:
       - field.storage.node.field_event_attendance
+      - field.storage.node.field_event_status
   display:
     default:
       display_options:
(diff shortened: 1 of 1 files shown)


cory-skeers:
Time display gets a little weird if it goes over multiple days, since it's only beginning time and ending time w/out dates (so can have something like "1:00 PM - 11:00 AM", or for All Day multi-day events, "12:00 AM - 11:59 PM" with no indication of the multi-day aspect.

But between the old example and events on the site currently, they don't seem to have any events for which this is a problem, so I think it's safe for On Iowa.

config/sites/oniowa.uiowa.edu/config_split.patch.views.view.events.yml
```diff
@@ -15,6 +19,125 @@ adding:
...
+            element_label_colon: false
+            element_wrapper_type: ''
+            element_wrapper_class: ''
+            element_default_classes: true
+            empty: ''
+            hide_empty: false
+            empty_zero: false
+            hide_alter_empty: true
+            date_format: custom
+            custom_date_format: 'g:i A'
```
cory-skeers: There's some leeway, but the UI editorial guide usually has lowercase

>Time: a.m. and p.m.
>The abbreviations “a.m.” and “p.m.” are lowercase with periods and no space between letters. There should
be a single space between “a.m.” or “p.m.” and the numeral that precedes them.
>6 a.m.
>10 p.m.
>In display type, tabular material, or lists, “a.m.”/“p.m.” style may vary depending on graphic design. For
example, small caps may be used, or the periods may be deleted. [UI]

```suggestion
            custom_date_format: 'g:i a'
```

config/sites/oniowa.uiowa.edu/config_split.patch.views.view.events.yml
```diff
@@ -67,10 +190,235 @@ adding:
...
```
cory-skeers:
```suggestion
            custom_date_format: 'g:i a'
```

cory-skeers:
> 4\. Running `ddev drush @oniowa.local config-split:export site` seems to export a reversion to the patch - is this a concern?

I'm seeing this too, and not sure what's going on with it... Ideally we'd want to figure this out so we don't risk undoing the changes by accident in future work, but with the time sensitivity of this, I think we could move forward for now and plan looking into this more later.

config/sites/oniowa.uiowa.edu/config_split.patch.views.view.events.yml
```diff
@@ -67,10 +73,352 @@ adding:
...
+            id: area
+            table: views
+            field: area
+            relationship: none
+            group_type: group
+            admin_label: ''
+            plugin_id: text
+            empty: false
+            content:
+              value: '<p><br /><strong>*</strong> Attendence Required</p>'
```
cory-skeers:
```suggestion
              value: '<p><br /><strong>*</strong> Attendance Required</p>'
```

cory-skeers (approved):
Getting a clean `config-split:export` with it now, and the changes look good. The only ways I could sort of "break" it were with multi-day events, which on iowa doesn't seem to have, and even so, doesn't fully break--it just gets difficult to fully differentiate the times without going to the specific node.

bspeare:
Thanks @cory-skeers!

PR #6785, opened 2023-08-15
Website reporting tool

pyrello:
Provides a BLT command that generates a report of website analytics data.

Adds drush aliases for D7 applications.

# How to test
```
blt ura --export
```
- Observe that the file gets created.
- And gets updated with new records as domains are checked.
- No need to run the entire thing.


joewhitsitt:
Is there an issue or epic this is related to?

pyrello:
> Is there an issue or epic this is related to?

No. This was unplanned work based on an emergent need to deliver a report with all the GA/GTM information we have for all "live" sites. I don't think it ever got turned into an issue.

joewhitsitt:
Okay. Just looking to understand the "compliance" bit to this. I guess I was out when this work emerged and was discussed.

>Generates a report with compliance details for sites.

Is used for analytics and lastlogin. Compliance with what? All sites need GA or GTM?

joewhitsitt:
I was provided a bit more information about this work, enough to proceed with reviewing the PR as-is.

joewhitsitt:
Do we know what causes the following?

```
Processing domains for uiclas...
Processing art.uiowa.edu
 [Acquia\Blt\Robo\Tasks\DrushTask]   
 [Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1 
Updated /var/www/html/SiteNow-Report-Analytics-20230821-192125.csv
```

There is still a result in the file, so not a huge worry.

blt/src/Blt/Plugin/Commands/ReportCommands.php
```diff
@@ -0,0 +1,404 @@
...
+/**
+ * Global multisite commands.
+ */
+class ReportCommands extends BltTasks {
+
+  use AcquiaCloudApiTrait;
+  use InspectorTrait;
+
+  /**
+   * Generates a report with compliance details for sites.
```
joewhitsitt:
```suggestion
   * Generates a report of websites with their Google Analytics and Google Tag Manager ids.
```

pyrello:
> Do we know what causes the following?
> 
> ```
> Processing domains for uiclas...
> Processing art.uiowa.edu
>  [Acquia\Blt\Robo\Tasks\DrushTask]   
>  [Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1 
> Updated /var/www/html/SiteNow-Report-Analytics-20230821-192125.csv
> ```
> 
> There is still a result in the file, so not a huge worry.

I'm not sure what is causing it to have that output. Without the `--debug` flag, it should be suppressing all output, but it doesn't. There are quite a few things that cause that type of output to display, although they are all part of the normal functioning of the tool. In D7 sites, drush responds with an error when you query for a variable that doesn't exist and the GTM variable doesn't exist on all sites. So maybe that's it?

joewhitsitt:
>In D7 sites, drush responds with an error when you query for a variable that doesn't exist and the GTM variable doesn't exist on all sites.

Ah okay. That seems to align with the data I get back.

blt/src/Blt/Plugin/Commands/ReportCommands.php
```diff
@@ -0,0 +1,404 @@
...
+                ->ansi(FALSE)
+                ->drush('config:get')
+                ->args(['config_split.config_split.sitenow_v2', 'status'])
+                ->option('uri', $domain)
+                ->run();
+
+              $site['version'] = str_contains(trim($result->getMessage()), ': false') ? 'V3' : 'V2';
+            }
+
+            // Run 'drush config:get google_analytics.settings account'.
```
joewhitsitt:
```suggestion
            // Get the last login.
```

joewhitsitt:
https://github.com/uiowa/uiowa/pull/6785/files#diff-e58379d737967dfd847c0da29f5198c12e242c198a048fe22c46c9ea35bf93e7R207-R208

These should be removed/replaced.

Perhaps it could become name, date and the report gets the `name` from the same table?
https://github.com/uiowa/uiowa/pull/6785/files#diff-e58379d737967dfd847c0da29f5198c12e242c198a048fe22c46c9ea35bf93e7R284

blt/src/Blt/Plugin/Commands/ReportCommands.php
```diff
@@ -0,0 +1,404 @@
...
+  public function lastLogin($options = ['export' => FALSE, 'debug' => FALSE]) {
+    $site_data = [];
+
+    $headers = [
+      'Application',
+      'URL',
+      'Version',
+      'Status',
+      'GA Property IDs',
+      'GTM Container IDs',
```
pyrello:
```suggestion
      'Last Login,
```

pyrello:
> These should be removed/replaced.

Done.

joewhitsitt:
Final thing is whether it is worth it or not to include icgov in this? This app has less needs less campus oversight compared to our other sites, no sites in production currently and this app is eventually going to be wiped.

joewhitsitt (approved):
Nice work. Thanks for indicating that the last login command was a proof of concept. D7 sites already have an drush/inactive.php script that accounts for user 1 and administrator roles. Depending on the use-case here, it will provide a different result.

I think based on internal discussion the icgov stuff doesn't need to be included but if it is left in, we will need to make sure we remove it when that app is repurposed.

PR #6794, opened 2023-08-18
Fixes featured content headline bug

### LINKED ISSUE #6792: Featured content items don't respect block's headline settings
# To reproduce

- https://sandbox.prod.drupal.uiowa.edu/featured-content
- Inspect HTML and observe that the featured content items use an H2 tag despite their block headline using H3 and H4.

# Proposed solution
TBD

pyrello:
Resolves #6792 

# How to test
```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli featured-content
```
- Confirm that existing featured content block items use the appropriate h tag (one level below the headline)
- Add a new block, update existing blocks, anything else you can think of to ensure that this works across the board

## Removal of logic from #3715
- Review #3715 and https://github.com/uiowa/uiowa/issues/3558#issuecomment-868700956
- Test by creating/editing multiple pages that use article lists, people lists, and featured content that include the same nodes with settings that cause the headline to be a different size.
- Confirm that all node item headlines in each type of list display the correct size headline in all instances and across updates and additions of new blocks.

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
+++ b/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
@@ -1189,18 +1189,10 @@ function layout_builder_custom_preprocess_block__inline_block__featured_content(
     if ($target) {
       $entity = $target->getEntity();
 
-      // Associate the block's cache tags with each entity so that the
-      // appropriate render cache ID is cleared when the block changes.
-      $cache_tags = $block->getCacheTags();
-      $entity->addCacheTags($cache_tags);
-      // Grab a fresh copy of the node.
-      $id = $entity->id();
-      Cache::invalidateTags(['node:' . $id]);
-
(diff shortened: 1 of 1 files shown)


bspeare:
What does the caching removal do here?  What was it doing before?

bspeare (approved):
I've tested this on all the blocks on multiple pages, tried visually hiding the title, emptying the title field out and adjusting block styles and the heading still stays the correct level.  Nice fix!

pyrello:
> What does the caching removal do here? What was it doing before?

That's what the section about removing the logic from #3715 is about. At some point we added the logic to prevent issues with the headings being incorrectly sized when there were multiple instances of the same node displaying in multiple blocks. I think we implemented a different and more generalized fix for that issue in #6054. So I don't believe the cache invalidation is necessary anymore.

PR #6799, opened 2023-08-18
P2lb Conversion Status column

GaryRidgway:
I wanted a way to know if a page had issues with conversion really quickly instead of having to go to every node to check, so I added a column to the converter page that checks for issues.

Screenshot 2023-08-21 at 11 27 54 AM

# How to test

1. Sync a V2 site
2. Enable P2lb
3. Check the converter page
4. Investigate the links and whatnot and if it is evaluating correctly.


joewhitsitt:
If you are looking for different wording input, I'd change the table header to "V3 Compatibility" and I would have text for both statuses. "Fully compatible" and "Partially compatible - See details".

For performance reasons, I'd also suggest reducing the pager to 25 for both this check and to reduce the number of conversions that can be triggered at one time. It seems that there is some caching here which is good.

GaryRidgway:
> If you are looking for different wording input, I'd change the table header to "V3 Compatibility" and I would have text for both statuses. "Fully compatible" and "Partially compatible - See details".
> 
> For performance reasons, I'd also suggest reducing the pager to 25 for both this check and to reduce the number of conversions that can be triggered at one time. It seems that there is some caching here which is good.

Do you think we can keep the Emojis? I am really liking that there is an "At a glance" feeling to knowing what is fully or partially compatable. Perhaps there is a better emoji for partially compatible rather than an "X"?

GaryRidgway:
After getting feedback from @joewhitsitt, I have updated the wording and changed the image in the PR description to match its new feel.

joewhitsitt:
>Do you think we can keep the Emojis? I am really liking that there is an "At a glance" feeling to knowing what is fully or partially compatable. Perhaps there is a better emoji for partially compatible rather than an "X"?

I think your selections are fine for this backend view. I was just wanting a11y text alternatives.

cory-skeers (approved):
I think this is a great addition to incorporate the analyzer with bulk conversion options, and from testing, everything worked well.

I kind of wish the compatibility status was centered in the cell instead of offset slightly by the text bumping up to be inline with the emoji (so that it would be vertically aligned with the user/moderation columns in the example here), but that doesn't detract from the usefulness of the tool.

image

PR #6801, opened 2023-08-21
ITS - Site Split and Service Content Type

### LINKED ISSUE #6666: ITS Services content type
# Story

As an ITS webmaster, I would like to to be able to create service content so that services can be dynamically displayed to our users.

# Additional Info

**Existing mockups - based on initial feedback**
All services page: https://sandbox.prod.drupal.uiowa.edu/all-services-and-software-copy
Results page: https://sandbox.prod.drupal.uiowa.edu/results-page
Services page: https://sandbox.prod.drupal.uiowa.edu/its-service-office-365-mockup

**Fields we are fairly confident ITS needs**
Title
Also known as (alias/synonym/alternate name)
Description/Body/teaser
Need a way to feature
Way to designate audience
Service category
Related services

Fees?
Website?
Quick Links?

# Proposed Solution

- [x] Setup site split
- [x] Create content type and path alias according to https://wiki.uiowa.edu/display/drupaldev/Content+type+creation+notes
- [x] layout builder

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6666

# To Test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local config:import --source ../config/sites/its.uiowa.edu --partial && ddev drush @its.local cim && ddev drush @its.local cr && ddev drush @its.local uli admin/structure/types/manage/service/display/default/layout
```

- Upon `uli`, you should see the moderation control block within the layout builder display settings without errors.
- As an editor/publisher/webmaster, you can add/edit any service.
- As a publisher/webmaster, you can delete a service. As an editor, you can delete your own service.
- Basic metatag and schema.org ([Service](https://schema.org/Service)) output should be present in the DOM.
- Alias should match the /services/[node:title] pattern as discussed here: https://github.com/uiowa/uiowa/pull/6801#issuecomment-1686991745

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -42,6 +42,8 @@ Deploy the code changes to each environment per the normal process and import
 the configuration from the split manually.
 ```drush @mysite.dev config:import --source ../config/sites/mysite.uiowa.edu --partial```
 
+You may have to run `cim` and `cr` after the partial import for certain config like a moderation_control block to register.
+
 ## Caveats
 **DO NOT** split core.extension for your site. This will only lead to problems.
 You can complete-split individual modules that your site needs and Config
--- a/config/sites/its.uiowa.edu/.htaccess
+++ b/config/sites/its.uiowa.edu/.htaccess
@@ -1,0 +1,27 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php7.c>
+  php_flag engine off
+</IfModule>
+<IfModule mod_php.c>
+  php_flag engine off
+</IfModule>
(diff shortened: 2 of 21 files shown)


pyrello:
Pulling to test

pyrello:
I think the path alias for these should be `services/[node:title]`. I know that is different than the old site, but I think its better information organization.

pyrello:
> As an editor/publisher/webmaster, you can add/edit a service.

As an editor, I can add the service, but I'm limited to the states that I can set it to: Draft and Review. Is that correct?

pyrello:
> As a publisher/webmaster, you can delete a service.

As a publisher, I can delete my own service. Is that what is meant?

pyrello:
> As a publisher/webmaster, you can delete a service.

As an editor, I can delete my own service. Is that what is meant?

joewhitsitt:
>I think the path alias for these should be services/[node:title]. I know that is different than the old site, but I think its better information organization.

Maybe we should discuss this with them. It would also require a redirects migration. Their current site is using https://its.uiowa.edu/all-services-and-software as the landing page so /services would be a 404 if folks deconstructed the URL path in the hopes of "going back." They also currently have a breadcrumb redirect that adds to the mess.

>As an editor, I can add the service, but I'm limited to the states that I can set it to: Draft and Review. Is that correct?

Correct.

>As an editor, I can delete my own service. Is that what is meant?

"As a publisher/webmaster, you can delete *any* service." is what I meant, but "As an editor, I can delete my own service" should also be true based on the permissions config.

pyrello:
> > I think the path alias for these should be services/[node:title]. I know that is different than the old site, but I think its better information organization.
>
> Maybe we should discuss this with them. It would also require a redirects migration. Their current site is using https://its.uiowa.edu/all-services-and-software as the landing page so /services would be a 404 if folks deconstructed the URL path in the hopes of "going back." They also currently have a breadcrumb redirect that adds to the mess.

It would be my intention to change the services landing page to just `/services`. If we don't establish a pattern like that, then we are stuck having to try to engineer a breadcrumb solution. I think we should recommend these updates to them.

A redirect migration is probably necessary no matter what because I'm not sure that we can be 100% sure that all the path aliases will get generated the same. I have encountered examples where the slug is different in D7 vs D9.

joewhitsitt:
Fair enough. I'll make that change and we can shop it with them as we develop the content type further.

pyrello (approved):

pyrello:
Works as advertised. Thanks for the adjustments!

PR #6802, opened 2023-08-21
[Image block] Added stretch to fit style

### LINKED ISSUE #6803: Add "cover" or "stretch" layout builder style to the image block
# Story

As an editor, I can place an image block and text area in a 50/50 column and have the image fill the vertical space. 

# Proposed solution

- Add an image style to the image block that sets the image to vertically fill the container space. 

Before

Screenshot 2023-08-21 at 3 30 52 PM

After

Screenshot 2023-08-21 at 3 26 11 PM

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6803

# How to test

```
ddev yarn workspace uids_base build && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/1776/layout
```

Add "Stretch image to vertically fill container" style to the image block on the page and verify that the image fills the entire container.


pyrello:
Pulling to test

pyrello:
Adding the cover option as a separate style that can be layered on top of the media shape is a bit confusing in its current form. If I choose widescreen or ultrawide and select the "Stretch image to vertically fill container" option at the same time, I miss getting to see what the image would have looked like without the cover setting. Maybe I'll just understand that this was the shape of the actual image file that I working with that is then being turned into a different shape. Or maybe I won't.  The actual effect of combining the two options is that the media shape acts as kind of a zoom on the image. But its unclear how much zoom is being applied or even why.

If it is actually desirable to combine these two options together, I think that we will need to potentially redefine the experience of how people select the image shape. Or we might need to change the labeling of one or both of the options to make them make sense together.

My alternative suggestion is to add a "Fit height" option to Media shape which uses the un-cropped image behind the scenes and applies the cover style to it to achieve the desired effect.

bspeare:
Thanks @pyrello, what about only showing the style option if the "No crop" media shape is selected?

bspeare:
This is no longer a "JDI" type of issue.  I'm going to close this and link to the comments here in the original issue.

pyrello:
> Thanks @pyrello, what about only showing the style option if the "No crop" media shape is selected?

I'm curious why this would be better than adding a "shape" option?

bspeare:
@pyrello I'm not sure it is better, but might help users understand that the "No crop" is being stretched to fill the space? I'd be happy to have either solution.

PR #6804, opened 2023-08-22
Classrooms - Fix classrooms view sort

### LINKED ISSUE #6767: Classrooms room order
>We did just notice something a little strange. When we go to the classroom search tool and filter to Lindquist Center, room N105 is out of order from the rest of the rooms. Instead of coming after N104, it’s last on the page.

It looks like Room N105 has the "Room Name" of "TILE-Flex (BYOD)". I'm guessing it's ordered by that and not necessarily Title?

^ Maybe not. Leonard Feldt Lecture Room is still in the correct order by room title.

I'm also seeing inconsistencies in the order prior to filtering vs post filter applied (AJB W240 on initial load vs filtering by building).

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6767

# To Test

```
ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local uli spaces/find-university-classroom
```

- Update hook should fire, with no issues.
- Though there is a combined id field, it is not visible in the form or any of the displays.
- See that ABW rooms are showing up first before AJB.
- See that AJB W240 is the last AJB building
- Go to page 3 unfiltered and see LC N105 is right after LC N104
- Filter by Lindquist Center and see that LC N105 is right after LC N104
- See that the view A-Z sort options are the same as PROD and work as expected based on the card titles.


bspeare:
Pulling to test

bspeare:
I've synced this twice, once with a `sql-drop` and I see a different sort order every time.  Maybe local gremlins?  I'm going to unassign myself to see if others have the same result.

joewhitsitt:
taking a look at this again. from my early childhood studies i can say that AJB does not come before ABW. the card title is not the title and so this is still not the expected sort...

joewhitsitt:
Different approach. Updated test instructions. Ready for review again.

bspeare:
All the testing steps check out.   The only difference that I am noticing is the `Low to High` sort is Displaying Van first instead of College of Pharmacy.

joewhitsitt:
>The only difference that I am noticing is the Low to High sort is Displaying Van first instead of College of Pharmacy.

My understanding is that by switching the sort, the other sort no longer applies, particularly because we are combining the options into one with BEF. This also occurs in PROD.

I tried to add a tertiary (not exposed) sort for the combined id (asc) to apply before and after the exposed ones. When placed after, it worked but then the Z-A sort stopped working. When placed before, the occupancy filters didn't work. Additionally, I recall this view already being SQL query heavy and we might not want to add to the fire.

I think with occupancy applied, it is then sorted by nid and then by title when it doesn't randomly glitch. 🤷 

Not sure what more can be done without going back to basics.

joewhitsitt:
To comment on the update hooks, I don't think it requires a scheduled release. The view will work as-is until after the update hooks run. The first update hook imports some config and the second just populates those fields. Then the view and the rest of the config changes happen. It takes a bit to save the 400+ nodes but it is one site compared to the featured content update hook already merged in main.

briand44:
I did a little digging to see if I could find that sort logic that I thought we had done previously for Registrar site but now that I am looking into it I think that was actually for the old Diversity site and not the Registrar site. It appears the Registrar site has sorting issues when sorting by max occupancy as well https://registrar.uiowa.edu/ucr-sps?field_gac_building_tid=All&field_gac_room_value=&field_gac_expanded_capacity_value=&field_gac_expanded_capacity_value_1=&field_gac_room_type_tid=All&field_gac_seating_type_tid=All&field_gac_type_tid=All&field_gac_cls_manages_assignment_value=All&sort_bef_combine=field_gac_expanded_capacity_value+ASC&items_per_page=50.

If they've lived with similar sorting issues for the Registrar site for years maybe not a big deal on the Classrooms site either?

bspeare (approved):
Approved based on testing from yesterday.

joewhitsitt:
>Downloading https://nodejs.org/dist/v16.20.2/node-v16.20.2-linux-x64.tar.xz...
>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Ugh..

PR #6812, opened 2023-08-23
Hotfix: Added a check for `class` member of attributes array to be an array

pyrello:
# How to test

```
ddev blt ds --site usg.uiowa.edu && ddev drush @usg.local uli branches/legislative-branch/constituency-senators
```

### DIFF
--- a/docroot/modules/custom/layout_builder_custom/src/LayoutBuilderStylesHelper.php
+++ b/docroot/modules/custom/layout_builder_custom/src/LayoutBuilderStylesHelper.php
@@ -62,14 +62,16 @@ class LayoutBuilderStylesHelper {
   public static function removeStylesFromAttributes(array &$attributes, array $style_map) {
     // Filter class list to only elements didn't match a style from the style
     // map.
-    $attributes['class'] = array_filter($attributes['class'], function ($class) use ($style_map) {
-      foreach ($style_map as $style) {
-        if (str_contains($style, $class)) {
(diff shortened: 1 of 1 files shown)


joewhitsitt:
The deployment finished and seems to have resolved the underlying issue. Is this still something worth adding in?

pyrello:
> The deployment finished and seems to have resolved the underlying issue. Is this still something worth adding in?

On the one hand, this seems like a pretty obvious way to prevent issues in the future. On the other hand, this failing would seem to indicate a problem that we need to look into and so implementing this fix means that we won't get that feedback.

cory-skeers:
> > The deployment finished and seems to have resolved the underlying issue. Is this still something worth adding in?
> 
> On the one hand, this seems like a pretty obvious way to prevent issues in the future. On the other hand, this failing would seem to indicate a problem that we need to look into and so implementing this fix means that we won't get that feedback.

My initial thought is to log a message in the event it's getting skipped over, so that we can avoid the wsod but still get that feedback. But I worry that'll just end up being noise.

pyrello:
I think it was a mistake to design this method so that it gets passed the `$attributes` array when all we care about is updating the `$attributes['class']`. Ideally, it should be rewritten so that it gets passed `$attributes['class']` instead.

pyrello:
Closing this in favor of a potential rewrite of the affected functionality down the road.

PR #6821, opened 2023-08-25
[ ITS ] Service content type fields

### LINKED ISSUE #6806: Create fields for Services content type
**Existing mockups**
All services page: https://sandbox.prod.drupal.uiowa.edu/all-services-and-software-copy
Results page: https://sandbox.prod.drupal.uiowa.edu/results-page
Services page: https://sandbox.prod.drupal.uiowa.edu/its-service-office-365-mockup

**Fields we are aware of based on mockups**
Title
Also known as (alias/synonym/alternate name)
Description/Body/teaser
Need a way to feature
Way to designate audience
Service category
Related services
Fees
Website
Quick Links
Contact Info/Getting Help?

# Proposed Solution

- [x] Create fields
- [x] theme full node
- [x] theme teaser

cory-skeers:
Resolves #6806 

# How to test

Currently importing the full site split leads to an error, but removing `config/sites/its.uiowa.edu/schemadotorg.schemadotorg_mapping.node.service.yml` prior to import gets around this. After the site split has been imported, you can restore this and then import it by itself.

One liner:

```
rm config/sites/its.uiowa.edu/schemadotorg.schemadotorg_mapping.node.service.yml && ddev blt ds --site=its.uiowa.edu && ddev drush @its.local config:import --source ../config/sites/its.uiowa.edu --partial -y && ddev drush @its.local cim -y && git restore config/ && ddev drush @its.local cim -y && ddev drush @its.local cr && ddev drush @its.local uli /node/add/service
```

Add a service or two (or three--go wild!)
Check out the new fields, and try to recreate some services https://its.uiowa.edu/all-services-and-software (note: Support and Training and News sections will come later). See that the field-types and formats make sense and fit for the type of data needed
(Note: You'll need to create Audience and Service Category taxonomy terms for those fields to appear and function)
Compare the full node display with the finalized mockup: https://sandbox.prod.drupal.uiowa.edu/its-service-office-365-mockup
Check the functionality for the Fees and Fees info fields, and that it matches the description at https://github.com/uiowa/uiowa/pull/6821#issuecomment-1707241922

On a Page, add a Featured Content block in Layout Builder
Include a service (or services)
Check the teaser functions with title summary (or trimmed body if no summary)


config/sites/its.uiowa.edu/core.entity_view_display.node.service.default.yml
```diff
@@ -64,38 +75,248 @@ third_party_settings:
...
+              context_mapping: {  }
+              block_revision_id: 2243
+              block_serialized: null
+              type: uiowa_spacer_separator
+              uuid: ac628a5b-420d-4c66-bb9c-ecd8db8197ee
+            weight: 0
+            additional:
+              layout_builder_styles_style:
+                block_margin_default_removed: block_margin_default_removed
+            third_party_settings: {  }
```
pyrello: I think we are trying to move away from using the component. @bspeare your thoughts?

cory-skeers:
After wiring things up, the links in the card sidebar are different colors depending on if it's the single website link or one of the multi-field quick links:
image
I'm not sure here if they should be consistently using the brand-secondary, or the blue.
image
It looks like the blue, and the website field should be updated to accommodate?

bspeare:
@cory-skeers I think we just need to remove the`bg--white` from the card in the layout.  Maybe we can set up the card layout to pass that value in at https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/layouts/layout--twocol--card-sidebar.html.twig#L43?

cory-skeers:
> Maybe we can set up the card layout to pass that value in at https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/layouts/layout--twocol--card-sidebar.html.twig#L43?

Unless we want to allow for adjusting the card's background to be separate from the rest of the section background, I think removing the `bg--white` is enough to get it working as expected. It and its contents pick up the background from the parent containers, and it works for both the `- None -` option as well as the others.

cory-skeers:
> Unless we want to allow for adjusting the card's background to be separate from the rest of the section background, I think removing the `bg--white` is enough to get it working as expected. It and its contents pick up the background from the parent containers, and it works for both the `- None -` option as well as the others.

Ope. Also changes the `tel:` link colors on Person content types (and other fields in the card).

image

vs prod

image

bspeare:
@cory-skeers yes, I was thinking we could either preprocess the class out for this layout or add it as a configurable property in the card.   Would you like me to take a look at it?

cory-skeers:
> @cory-skeers yes, I was thinking we could either preprocess the class out for this layout or add it as a configurable property in the card. Would you like me to take a look at it?

I would appreciate that 😄

bspeare:
Maybe it has been mentioned already, but by deleting `schemadotorg.schemadotorg_mapping.node.service.yml` from the config and split I was able to sync the site without errors.

cory-skeers:
> Maybe it has been mentioned already, but by deleting `schemadotorg.schemadotorg_mapping.node.service.yml` from the config I was able to sync the site without errors.

Thank you! I've been going down all sorts of different rabbit holes that were not getting anywhere.

joewhitsitt:
What is the effect of that? Did the recent update to the module move that information elsewhere? Is it needed in PROD for it to output the schema info? Is there just a bit of bad config that needs updating inside it?

cory-skeers:
> What is the effect of that? Did the recent update to the module move that information elsewhere? Is it needed in PROD for it to output the schema info? Is there just a bit of bad config that needs updating inside it?

Unsure of that yet. I've removed it from the branch for now for ease, going to try re-setting things up and re-exporting and see if it can get working again.

cory-skeers:
> > What is the effect of that? Did the recent update to the module move that information elsewhere? Is it needed in PROD for it to output the schema info? Is there just a bit of bad config that needs updating inside it?
> 
> Unsure of that yet. I've removed it from the branch for now for ease, going to try re-setting things up and re-exporting and see if it can get working again.

I don't think it's the config itself, but a change to processes in `schemadotorg` when config is imported (changes themselves, or maybe a change that affected timing where it happened to not be an issue before). Current working process is to remove `schemadotorg.schemadotorg_mapping.node.service.yml`, import the site split, re-add `schemadotorg.schemadotorg_mapping.node.service.yml`, and import. 

Seems related to https://drupal.stackexchange.com/questions/253307/how-can-i-fix-the-deleted-and-replaced-configuration-entity-error-during-confi, and browsing `schemadotorg` code, looks like they have some entity inserts that don't check if config is currently syncing.

Plan on opening an issue, and see if we can get a patch in place, otherwise this might come up again in prod.

cory-skeers:
> Seems related to https://drupal.stackexchange.com/questions/253307/how-can-i-fix-the-deleted-and-replaced-configuration-entity-error-during-confi, and browsing `schemadotorg` code, looks like they have some entity inserts that don't check if config is currently syncing.
> 
> Plan on opening an issue, and see if we can get a patch in place, otherwise this might come up again in prod.

Maybe not. Manually added checks in a few spots I'd thought might be related, but still running into the issues. I think this is on the right track, but not quite there yet. Might be another module's hook that's triggering off `schemadotorg`'s config import? The issue definitely seems to be caused at the point of its import, and only if it's being imported alongside other config.

cory-skeers:
@bspeare I set some things up in [fd2a5a1](https://github.com/uiowa/uiowa/pull/6821/commits/fd2a5a1b4a783569a4df228301b1470a83c20bc4) and [685a00c](https://github.com/uiowa/uiowa/pull/6821/commits/685a00c71cb719cd6472e1f8821d44ad8862e01a). For the ITS Services, it's passing a specified `bg--none` into `uids_base.theme` to handle it. In cases where it's not specified (either with a specific `bg--black`, etc or with the `bg--none`) it defaults to `bg--white`. Since the theme hooks fire before the module hooks, it has the bit of oddness in passing it into the theme rather than being able to just unset it at the site module level.

This feels a bit sloppy, though, so would welcome any thoughts on it.

I'm also struggling a bit with the best way to handle this:
image
The top padding gets removed from the container directly below the container which contains the title (quite the mouthful). I could preprocess out the `layout--title` or something like that, but I have the feeling my go-to there is more hacky than we should be doing...

bspeare:
@cory-skeers I would go with https://github.com/uiowa/uiowa/pull/6821/commits/19924b1b57246aed49177b17ddb5bf869c329784 to fix the header.   Some of the preprocess stuff (page title highlight and card bg) is not showing up with my layout, do you think that is because of the errors during the install?

cory-skeers:
> @cory-skeers I would go with [19924b1](https://github.com/uiowa/uiowa/commit/19924b1b57246aed49177b17ddb5bf869c329784) to fix the header. Some of the preprocess stuff (page title highlight and card bg) is not showing up with my layout, do you think that is because of the errors during the install?

Thanks--the layout update looks great!

For the preprocess stuff not showing up, hrm...I never ran into an issue with the module getting enabled and functioning properly from the install errors, but it's possible. If you directly `ddev drush @its.local en its_core`, does it show up as enabled (or get enabled if not already)?

bspeare:
@cory-skeers I removed the schema files again locally and installed and all of the preprocess work is showing now.

pyrello:
Pulling to test

pyrello:
In the old Info Center content type, the Service Category field is a single value field, which is not how it is implemented in this PR.
Screenshot 2023-09-01 at 2 08 46 PM

On the old site, it is represented in the breadcrumb:
Screenshot 2023-09-01 at 2 07 28 PM

We do not appear to have it represented in our service mockup currently. If we want to continue to be able to use it as a breadcrumb, then we need to maintain Service Category as a single value field. If the customer is interested in this being a multiple value field (and I don't think we should change it if they are not requesting this) then we'll need to adjust the way that it is represented on the page. Either way, we should probably update the mockup to include this important element.

pyrello:
On the old site, the Fees field is broken down into two fields:
- A select with "No charge," and "Other"
- A text area for Additional Fee Information

The current approach just uses a text area. Is the idea with this that if there is no charge, the customer will just enter "No charge" in the text area?

pyrello:
Just checking: Are **Quick Links** a replacement for **Common Actions**?

pyrello:
> We do not appear to have it represented in our service mockup currently. If we want to continue to be able to use it as a breadcrumb, then we need to maintain Service Category as a single value field. If the customer is interested in this being a multiple value field (and I don't think we should change it if they are not requesting this) then we'll need to adjust the way that it is represented on the page. Either way, we should probably update the mockup to include this important element.

After discussing this with @briand44 a bit, I think we should encourage ITS to move away from using service categories as a breadcrumb parent for the service. The rationale for this is: If they are going to be part of the breadcrumb, I would argue that we should include the service category in the path alias for the service (e.g. `services/office-365` => `services/email-calendar/office-365`) to have the path structure reflect the informational hierarchy. However, if we go this route, changing the service categories becomes more of a maintenance problem because changing a service's category would require updating the path alias. If we instead represent the categor(y/ies) in the content of the the service similar to how we represent tags, then there is still a path to the category landing page. If the categories need to change over time to keep pace with our understanding of how we group software/services together, then this can happen with no additional work to update the service unless a new category is being added.

cory-skeers:
> On the old site, the Fees field is broken down into two fields:
> 
> * A select with "No charge," and "Other"
> * A text area for Additional Fee Information
> 
> The current approach just uses a text area. Is the idea with this that if there is no charge, the customer will just enter "No charge" in the text area?

Good question.

I'd looked at some of the additional fee info and based it off that, but missed that there was a select for no charge or other.

In my mind it makes sense to just not display Fees if there is no charge, but I could understand there being a desire to explicitly show "No charge" on the page. I think in that case, though, we'd want to display that directly rather than within the "Fees" details.

The two options I'd want to see put in place would either be

1. Leave it as is as a single text field, and have them leave it blank going forward and not explicitly display "No charge"
2. Include a select list with two options
  a. "No charge": On the form, no textarea is visible. On the page, it displays a simple bit of text as "No charge"
  b. "Other": On the form, the textarea becomes available. On the page, it displays a details listed as "Fees" that expands to show the output from the textarea

bspeare:
@cory-skeers I like the idea of nothing rendering if there are no fees.  It seems like adding text explaining no fees is visual clutter that we don't need.  However, will the editor still type "No fees" in the fees text area?  Maybe I would vote for option 2 where if 2a was selected, nothing would render on the page.

pyrello:
> @cory-skeers I like the idea of nothing rendering if there are no fees. It seems like adding text explaining no fees is visual clutter that we don't need. However, will the editor still type "No fees" in the fees text area? Maybe I would vote for option 2 where if 2a was selected, nothing would render on the page.

The downside of having nothing render if there are no fees is that you lose the convention of how to find that information. If you can look in a standard position on the page and see at a glance that there are no fees, than you can quickly find that information. However if there is nothing listed about fees, you might spend a lot more time searching for it and then being unsure about it when you never find it.

bspeare:
@pyrello good point about the downside!

cory-skeers:
I've updated it to include the fees boolean and textfield, with the boolean controlling visibility of the textfield.

With the current setup, it either displays the fee information or nothing since the display is tied only to the textfield. But that is a fair point about the use of displaying no charge. I'm fairly even on which way to go, so maybe we go ahead with matching their current functionality and display the "No Charge" in that case.

cory-skeers:
> With the current setup, it either displays the fee information or nothing since the display is tied only to the textfield. But that is a fair point about the use of displaying no charge. I'm fairly even on which way to go, so maybe we go ahead with matching their current functionality and display the "No Charge" in that case.

This has been updated.

New functionality:

1. "No charge" option is selected, then the textarea is not visible on the form, and the text from their current site is displayed: "There is no charge for use of this service."
2. "Other" option is selected, then a textarea for additional fees info is visible on the form, and this text will be used within a details display on the page

cory-skeers:
> Just checking: Are **Quick Links** a replacement for **Common Actions**?

I believe so, yes, just a generalized term for that to use as a display label as well as the field itself.

bspeare:
Should we adjust the Audience and Category fields to be checkboxes instead of autocomplete?  It looks like they have 25 Info Center Categories on https://its.uiowa.edu/admin/structure/taxonomy/info_center_categories and 8 items for Audiences on https://its.uiowa.edu/admin/structure/taxonomy/audience.

bspeare:
All of the text formats are giving me the option to switch.  Can anyone else confirm this? Screenshot 2023-09-05 at 3 48 01 PM

briand44:
@bspeare 

>Should we adjust the Audience and Category fields to be checkboxes?

I thought we were only going to allow referencing one category per service?https://github.com/uiowa/uiowa/pull/6821#issuecomment-1703245392

bspeare:
If there is no fee, should we display the field label to make the card easier to scan?

Screenshot 2023-09-05 at 3 50 59 PM

cory-skeers:
> If there is no fee, should we display the field label to make the card easier to scan?

Looks like a good change!

cory-skeers:
I believe this is ready for review again.

Recent updates:
Changed Audience to checkboxes
Changed Category to single- rather than multi-select; changed to select list
Limited Contact and Fees Info textareas to minimal+ text format
When "No Charge" fee is selected, a field label is displayed. When "Other" is selected, the label is used to construct a details element

bspeare:
Maybe this is the config issue, but I when I create a new service and "No charge" is selected for fees the text will not show on the saved page.  It will only show up if I switch to "Other" save and then switch back to "No charge" on the page.

cory-skeers:
> Maybe this is the config issue, but I when I create a new service and "No charge" is selected for fees the text will not show on the saved page. It will only show up if I switch to "Other" save and then switch back to "No charge" on the page.

I think the preprocess field isn't firing because in those instances the `service_fee_info` field is null, and the logic needs to be switched around.

Taking a closer look.

cory-skeers:
> Maybe this is the config issue, but I when I create a new service and "No charge" is selected for fees the text will not show on the saved page. It will only show up if I switch to "Other" save and then switch back to "No charge" on the page.

This should now be fixed to work as expected with or without content in the additional fee info field.

I also updated the fees textarea label to Additional Fee Information, as the select list is now what gets used on the node display, and is labeled the simple Fees.

GaryRidgway:
I will pull this to test

GaryRidgway (approved):
After looking over this, everything looks good!
I saw that summary fields are not expanding by default to show all their content:
Screenshot 2023-09-07 at 9 58 32 AM

and I know I would like that to change, but I think that is probably more to do with the summary field itself than with this issue, and is thus out of scope.

Approved! yay!

PR #6795, opened 2023-08-18
[Facilities] Alerts on building pages

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6689. 

# How to test

```
ddev composer install && ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli /building/0374
```
- Code review
- Verify that the alert appears on the Carver building page
- Verify that the alert will disappear if end date is before today
- Verify that alert only appears on reference building page
- Review View settings and check for any settings that might be missing
- The styles will look messed up, further style updates will need to be made in https://github.com/uiowa/uiowa/issues/6790

## Additional cache testing

- Temporarily change the view from `today` to `now` for the date end filter
- Change the alert to expire within a few minutes. In an anonymous browser, load the carver building page and get it to cache.
- After time expires, refresh the anonymous window and see that the alert is still there.
- Run cron, the alert within the anonymous window on refresh should disappear. If not, you may need to wait for a rolling full hour to run cron again to have the alert disappear. Honestly don't know if the state timer is triggered on `blt ds` or is carried from PROD which sets the hour long count.

## Schemadotorg testing

- `ddev composer install` and try syncing known schemadotorg sites (facilities, classrooms, 91st meridian, iswf, sandbox (person)) to see if there are any import errors and to see if json+ld output is still working as expected.

PR #6827, opened 2023-08-28
Re-exported schemadotorg config

pyrello:
I noticed a config status warning when syncing a site. It led me to re-exporting config, which is what this update includes.

# How to test

```
ddev blt ds
```
- Ensure no config status errors.


pyrello:
Nevermind... I hadn't run `composer install` since the updates in #6795

PR #6842, opened 2023-08-31
[Image block] Added circle and no crop media shapes

### LINKED ISSUE #6368: As a SiteNow editor I can use a Banner without a gradient
# Story

As a SiteNow editor I can add a Banner to a page without it have a gradient. Currently you have to option to select a gradient option as Dark (default) or Light. There are certain scenarios where I need to use a Banner without a Title so that I can have a flexible height image which would then not need a gradient.

One example. This section with image and content side by side looks fine of large screens because the image is larger than the content.
image

but on smaller screens the image and content do not have equal heights.
image

Using a Banner solves this issue since it is a background image with 100% height, but now I have a gradient I don't need.
image

# Proposed solution

Make the gradient required if there is a Title filled out.

### LINKED ISSUE #6803: Add "cover" or "stretch" layout builder style to the image block

bspeare:
Trying again on https://github.com/uiowa/uiowa/issues/6803. 

Resolves https://github.com/uiowa/uiowa/issues/6368
Resolves https://github.com/uiowa/uiowa/issues/6803

# How to test

```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/1776/layout
```

- Add `No crop - fit height` style to the image block on the page and verify that the image fills the entire container. 
- Add `Circle` style to the image block and very that the image is set to a square format with a circle around it


pyrello:
I've been wondering if we should change the label on this style from "Media shape" to "Media format"?

bspeare:
@pyrello I like that idea!

joewhitsitt:
pulling to test

joewhitsitt:
Would it be possible to scope this CSS a bit to prevent the contextual links from going full-height as well?

Screenshot 2023-09-08 at 9 47 08 AM

bspeare:
@joewhitsitt yes, thanks for the reminder on that one.

joewhitsitt (approved):
Might not be the best solution, but https://github.com/uiowa/uiowa/pull/6842/commits/22ee79e55836b4430c2ee13744f6d1d4a8ac2323 works for the contextual links and this new style. Everything else in the PR seems to work as expected

joewhitsitt:
Can we capture the feedback that was given and the left-todo here?

bspeare:
I am investigating to see if I can set the cover style without overriding the height from `media--type-image.scss`  https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/media/media--type-image.scss#L38.

bspeare:
Since we are already setting the `img` width and height at https://github.com/uiowa/uids/blob/4.x/src/assets/scss/reset.scss#L233, I am going to remove the duplicate declaration here: https://github.com/uiowa/uiowa/pull/6842/commits/60c9e3f42b71b784544fa057bbd27b2858713194.   It looks like we added it 3 years ago.  We can always add it back in if we find visual regressions.

bspeare:
@joewhitsitt would you like to review the changes that I recently pushed up or should I just merge?

joewhitsitt:
reviewed, not tested. i am fine with merging for now.

PR #6843, opened 2023-08-31
Sprint 178 Provisions

joewhitsitt:


briand44:
I'm curious about putting intranet.shl.uiowa.edu on uiowa05. Are we doing that because we don't want to add any more sites to uiowa01? There are currently other *.shl.uiowa.edu sites on uiowa01 and the cert covers the wildcard already. I'll get on board with whatever the team decides here just thought I would check before moving forward with #6844

bspeare:
It looks like only `shlintranet.site.yml` needed updating.   Let me know if anyone sees other files that need an update.

cory-skeers (approved):
Things look good to me 👍

PR #6871, opened 2023-09-11
[ JDI ] Fix P2LB status column styles.

GaryRidgway:
JDI to fix up styles for status column


bspeare (approved):
Tested and this looks better!

PR #6878, opened 2023-09-13
[Bug fix] Card block: provide default value for field_uiowa_card_button_display

### LINKED ISSUE #6815: Card arrows not displaying when using an external link
# Problem
The Card arrows are not displaying when the card is using an external link. 

Example page:
https://imu.uiowa.edu/dining-and-retail

All cards are set to use the Page default which is to display arrows linking to pages. Hawk Shop is an internal link and that has an arrow but the other cards are external and the arrows are not displaying.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6815. 

# How to test

```
ddev blt ds --site=imu.uiowa.edu && ddev drush @imu.local uli /dining-and-retail
```
Code review

1. Change a card url to `https://example.com` and confirm that the card arrow displays if "Use site default" is selected.
2. Change settings on https://imu.uiowa.ddev.site/admin/config/sitenow/uiowa-pages and confirm that the cards display arrows based on that setting. 
3. Try all other options on the "Show button" and confirm that they still work


bspeare:
This pr fixes the bug outlined in https://github.com/uiowa/uiowa/issues/6815, but I'm not sure if it answers the question of why this doesn't work if an external link is used in the link field.

GaryRidgway (approved):
Based on testing instructions, this works

GaryRidgway:
So with some pretty extensive testing, I have come to the conclusion that when no button text is set, it is Always `null` on set and `null` on load.

My tests were `https://dining.uiowa.edu`, `<front>`, `/node/158`, and an autofilled `Dining and Retail (158)`.

I think this means that the solution is good, because it is not simply a workaround for one possibly incorrect issue, it is setting a default for all possible options. I think this also may be interacting with the fact that it is the default option to a non-required field.

joewhitsitt:
Thanks for putting this together @bspeare and @GaryRidgway for the in-depth testing. I am not surprised that we may have another instance where a new _none/"not set" type block setting has cropped up. Best to cover both until we can confirm one or the other completely taken over

PR #6884, opened 2023-09-14
lb_direct_add config fix

joewhitsitt:
# To Test

```
ddev blt ds --site=admissions.uiowa.edu
```

```
ddev blt ds --site=iowajpec.org
```

Both should be successful with config imports. On `main`, iowajpec.org has a config issue with lb_direct_add


GaryRidgway (approved):
When I tested the config imports on Main, I did receive the errors stated, but when I tested on this branch, I did not receive those errors. Approved.

Issue #5405, opened 2022-06-28
As a content creator, I would like to be able to schedule publishing a section and/or block for a certain period of time


makurj:
# Story

As a content creator, I would like to be able to schedule publishing a section and/or block for a certain period of time. For example, Section 2 or card X of the page should go live on Wednesday, July 6, 2022 through Friday, July 8, 2022. It would automatically get unpublished on Friday, July 8, 2022. This will be used for deadlines, event-related content, etc.

cory-skeers:
Progress on https://www.drupal.org/project/drupal/issues/3208766 would be good if this is based on section(s). As is, sections are only tracked by a delta, so if it were set to un/publish Section 2 on X date, any changes to the ordering or additions to sections at or before Section 2 would change which section is to be un/published.

Blocks already have a UUID, so that could be feasible.

Otherwise, there could be a path forward for node revisions to be published on a certain date, and maybe there are some contrib modules that already do this.

joewhitsitt:
Something to keep in mind is that scheduling content in general has been an up-hill battle so far based on our stack https://github.com/uiowa/uiowa/issues/723

cory-skeers:
Another thought I wanted to get down:

Would the overall page need to be editable while going through this scheduled un/publishing, or would it be possible to "lock" the content (aside from the block/section that need to go up and down)? It could be dramatically easier to handle if there can be a content freeze during it, including being able to use section deltas instead of waiting on UUIDs.

pyrello:
At first I was thinking about this in terms of trying to get scheduled workflows working for individual blocks, which seems pretty hard (see my original comments below). But it occurs to me that if we could limit the parameters on this to one set of changes to all blocks on a page that would be published on a date and revert back to another version on another date, that we don't have to get into all that nonsense. Instead, we would need a way to stage multiple versions of a piece of content and essentially manage them separately. While no cakewalk, this is likely a lot easier than what I was originally thinking about.

If this does need to be block-centric, then feel free to read what I originally wrote below.

<details>
<summary>Original comment</summary>

I was curious if there was anything I could find on the web that described something like this and I was not successful. So, now I'm trying to think through what would need to happen to make this possible:
- Introduce the concept of a block having a workflow (draft, published, unpublished)
- Resolve how we handle #723 
- Figure out how to apply the solution in the above bullet point to blocks
[...]

PR #6886, opened 2023-09-14
[Prof Dev] Small prof dev signage idea I had.

GaryRidgway:
This is a proof of concept for a signage solution build in to sitenow pages. simply go to any SiteNow page and add the `?signage` variable to the URL and the page will be converted in to a signage display of that page, using each section as a slide. An example can be found in the Gif below.

**NOTE**: This is just a proof of concept so it is pretty slapdash, and you have to be a bit finnickey with setting up some sections, but hey, its BUILT IN TO SITENOW.

<details><summary>Example here</summary>
<p>

proofOfConcept

</p>
</details> 

## To test.
1. Run :
```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /signage-view
```
2. Observe the page and that It looks fine without any URL variables.
3. Add the `?signage` argument to the URL.
4. Observe S I G N A G E


joewhitsitt:
i thought this was pretty neat when you first put it together. a couple points to consider though:

- not sure if the multi-widget on a single screen is still desired (weather, schedules, slideshow, marquee) or if we can get away with a header and one slide at a time.
- scheduled content is a big feature of the current signage. we still haven't gotten there on SiteNow: https://github.com/uiowa/uiowa/issues/5405
- acquia is billing for page views now and signage is largely uncached and triggering reloads to show scheduled content. also seeing signage go down temporarily more due to DoS type hits so i'm not sure we can afford to host the sites we used to (signage, content hub). i'd say the emergency website falls under this as well, but it is only one path and small output.

GaryRidgway:
> i thought this was pretty neat when you first put it together. a couple points to consider though:
> 
> * not sure if the multi-widget on a single screen is still desired (weather, schedules, slideshow, marquee) or if we can get away with a header and one slide at a time.
> * scheduled content is a big feature of the current signage. we still haven't gotten there on SiteNow: [As a content creator, I would like to be able to schedule publishing a section and/or block for a certain period of time #5405](https://github.com/uiowa/uiowa/issues/5405)
> * acquia is billing for page views now and signage is largely uncached and triggering reloads to show scheduled content. also seeing signage go down temporarily more due to DoS type hits so i'm not sure we can afford to host the sites we used to (signage, content hub). i'd say the emergency website falls under this as well, but it is only one path and small output.

Some awesome thoughts here, thank you Joe. I'm thinking it may not even be correct to use drupal on signage? or we really need to figure out a way to lower page views on signage.

PR #6924, opened 2023-09-22
[Hotfix] Fix config issues

cory-skeers:
# How to test

Sync ilr.law.uiowa.edu or jgrj.law.uiowa.edu on `main`, see a config diff error.
Check out this branch, re-sync, see no config diff errors.


joewhitsitt (approved):
makes sense based on recent merges. reviewed, not tested

PR #6892, opened 2023-09-15
[D10] Custom module compatibility updates

cory-skeers:
Relates to #6856 

# To Do

- [x] uiowa_profiles
- [x] docroot/sites
- [x] docroot/themes/custom/uids_base/assets/js/carousel.js

# How to test

```
ddev composer install
```

## Module info updates
Check for `core_version_requirement:` listings in any modules' `.info.yml` in our custom or site modules that does not include `^10`. They should all include the 10 compatibility.

## Explicit access checks
https://www.drupal.org/node/3201242
Code review should be all that's necessary. Entity queries that didn't specify their access check defaulted to `TRUE`. The only changes here are including an `->accessCheck(TRUE)`, which does not affect behavior.

## drupal-off-canvas updates
https://www.drupal.org/node/3305664
Edit a section/block in layout builder, and compare the off-canvas tray with prod. There should be no changes in style.

## javascript updates
https://www.drupal.org/node/3158256
https://www.drupal.org/node/3104677

Check the following still function (some it may be helpful to test in dev):

- sitenow_events: `single-event.js`: Behavior for handling events with repeated dates
- uiowa_alerts: `uiowa_alerts.es6`: Fetches and attaches hawk alerts
- uiowa_search: `cse.js`: Attaches the search results behavior for the site-wide search
- uids_base: `video-cooke.js` Reads from and sets a cookie to control the video autoplay
... and many more..
- uiowa_profiles: wasn't sure how to do core/once on a document each? so i grabbed what I thought would work. https://github.com/uiowa/uiowa/pull/6892/commits/92f839389474fe51448b9c984026e6c389ab19fc

## PHPCS updates

Also included a lot of phpcs warning cleanups.
```
ddev phpcs
```
Witness the lack (or close to) of warnings!

For full testing, can test the dependency injection updates:

- Replicate a node, and check the revision created time of the newly created node, and see that the time is accurate
- Check the Events config on a site without the Events feature split, and see that the Site Events config doesn't show up. Enable the split (`ddev drush @site.local config-split:activate event`) and see that the options now appear.
- Sync an intranet site or enable the intranet split on another site. Check that anonymous users don't have access
- Run the P2LB cleanup command and check that messages are still logged and orphan paragraphs are cleaned up

PR #6928, opened 2023-09-26
Update uiowa_auth and dependencies

joewhitsitt:
# To Test

As yourself, not `uli`, log in to education.dev.drupal.uiowa.edu using the hawkid form.


joewhitsitt:
Moved this into https://github.com/uiowa/uiowa/pull/6892 so this is only needed for testing in development and can be deleted after the other merges.

joewhitsitt:
This PR is no longer needed.

Issue #6798, opened 2023-08-18
Pair while provisioning a CCOM site on uiowa06


briand44:
# Story

As a UIHC MarCom developer, I would like to be able to submit and process a site request according to normal processes so we can contribute to CCOM provisions and not be dependent on others to process all CCOM provisions.

# Additional Info

   - Submit a site request using the UIHC site type
   - Process the site request according normal processes (Wednesday provision deployment)
   - Request DNS/SSL certs, etc. if necessary

briand44:
@ethan-fry has the uihc-web inbox been setup yet?

briand44:
SSL cert should be good since that was handled as part of setting up the application

briand44:
I got confirmation from Max that the existing request for NCBC can be processed with the other few initial CCOM sites.

bspeare:
Most of this is done.  What remains:
 - [ ] Request DNS/SSL certs, etc. if necessary
   - [ ] Request domains to be pointed to acquia-dlb.its.uiowa.edu with a Cherwell - HCIS-TO-Hostmaster with a “Hostmaster Request” 
 - [ ] Figure out what to do with SiteNow form submission

briand44:
The UIHC MarCom team plans to handle provisioning for the CCOM sites so we can likely archive those site requests after a quick peak if there is no concerns.

The UHC MarCom team also plans to follow our process of adding notes to the webform submissions and starring/locking when appropriate.

As far as DNS, we have a meeting next week to discuss with UIHC Hostmaster and Luke Maloney, but at this point it sounds like those requests will be submitted through Cherwell by the UIHC MarCom team and we don't need to worry about those.

The one question that came up is if the CCOM provisions should be part of our sprint provision PR or should that be a separate provision PR?

briand44:
I will grant the UIHC MarCom devs access to the Sharepoint Docs so they have access to those.

briand44:
I want to keep this open until:

- ~Access is granted to Sharepoint docs~
- ~Existing CCOM site request submissions have been updated with notes, etc.~
- ~We determine if the CCOM provisions are part of the normal sprint provision PR or separate~

briand44:
I've granted access to Sharepoint docs. @ethan-fry @seandui @bzalasky-uiowa can you confirm access?

joewhitsitt:
>The one question that came up is if the CCOM provisions should be part of our sprint provision PR or should that be a separate provision PR?

[...]

PR #6930, opened 2023-09-26
Sprint 181 Provisions

ethan-fry:
# How to test


joewhitsitt:
@ethan-fry 

Are you okay with waiting till tomorrow to merge this and align with our weekly cadence for these?

Based on the discussion here: https://github.com/uiowa/uiowa/issues/6798 I think we would only want one provisions/deprovisions PR per sprint

ethan-fry:
@joewhitsitt totally fine with this, I just wanted to make sure we had the provisions up for the 8 sites remaining.  I think it makes sense to do them all on the same day.

joewhitsitt:
I think @pyrello might have a few to add to this yet.

joewhitsitt (approved):

pyrello:
I have one more that I'm going to squeeze in because the site launch is time-sensitive.

Issue #6880, opened 2023-09-13
Facilities alert improvements


briand44:
- Update Dispatch template to use branded template https://github.com/uiowa/uiowa/issues/6880#issuecomment-1724070014
- Add project number at the bottom of the template
- Don't output labels in the template for the following fields:
    - Don't show Traffic Control if value is No
    - Don't show Affected Bus Routes if no routes checked
    - If no building selected don't show in template
    - We should check to see if this is consistent with how we are outputting on an alert node
- Require the category field on the Facilities alert content type split
- Spell check feature is a very nice to have feature
    - Chrome seemed to spell check text fields but maybe other browsers only work on text area fields?

<details>

> The email portion is more straight-forward than our email (text wise) and thinking we can pare down our language and use a different font if that is possible.
> 
> > Here's the P&T template per my other email. I like a couple of things about this, but not an exact match to what we need. We can discuss in our meeting. Thanks.
> 
> ![Image](https://user-images.githubusercontent.com/1017022/264413113-8cb7a55f-da92-4bec-9bf9-ccfa63325131.png)

> While inspecting what is wired up vs what is sent, I don't see the project number (AlertProjectNumber) included in the email. That is a new field correct? Should it be included?
> 
> Update: Brian suggested this be part of a separate slice: [#6823 (comment)](https://github.com/uiowa/uiowa/issues/6823#issuecomment-1699513676)

</details>

```[tasklist]
### Tasks
- [x] Update Dispatch template to use branded template Facilities alert improvements #6880
- [x] Add project number to bottom of template
- [x] Require category field for Facilities alerts
- [x] Don't output labels in the template for: If no building selected don't show in template https://apps.its.uiowa.edu/dispatch/help/freemarker#ifelse
- [x] Don't output labels in the template for: Don't show Traffic Control if value is No https://apps.its.uiowa.edu/dispatch/help/freemarker#ifelse
- [x] Don't output labels in the template for: Don't show Affected Bus Routes if no routes checked https://apps.its.uiowa.edu/dispatch/help/freemarker#ifelse
- [x] Check to see if this is consistent with how we are outputting on an alert node
- [x] Fix the node title config that should be hidden on the form
- [ ] Spell check for both text fields and text areas
```

bspeare:
Here's what the current template looks like:

[...]

PR #6934, opened 2023-09-28
Facilities Dispatch Conditionals Fix

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/6880

Blank but not empty values (`render()` doing what it does) were being sent over which dispatch considered a value to print. 

Screenshot 2023-09-28 at 12 00 30 PM

`array_filter()` in php 7.4+ handles `!is_null($value) && $value !== ''`

https://github.com/uiowa/uiowa/issues/6880#issuecomment-1739524380

Screenshot 2023-09-28 at 12 18 52 PM

# To Test

```
ddev composer install && ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli 
```

Add the dispatch key from facilities.uiowa.edu. Try scheduling a message. I will receive it unless you change the communication details: https://apps.its.uiowa.edu/dispatch/communications/1238605612/edit

Only values you have supplied will be included in the message.

### DIFF
--- a/docroot/sites/facilities.uiowa.edu/modules/facilities_core/src/Form/NodeAlertDispatchForm.php
+++ b/docroot/sites/facilities.uiowa.edu/modules/facilities_core/src/Form/NodeAlertDispatchForm.php
@@ -307,6 +307,9 @@ class NodeAlertDispatchForm extends FormBase {
       }
     }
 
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Nice improvement, @joewhitsitt!

Issue #6901, opened 2023-09-19
Update contrib to d10 compatible releases


briand44:
# Story

As a web team, we need to update contrib to D10 compatible releases so that we can get off of D9 by Nov 1.

# Additional Info

>Updating contrib could be a good next slice, but it will be a biggie. To make it smaller, I separated out core modules that have moved to contrib out as a separate slice. See issue summary: https://github.com/uiowa/uiowa/issues/6856) - @joewhitsitt

Remove unused modules? https://github.com/uiowa/uiowa/issues/5312?

# Proposed Solution

- uiowa_auth https://github.com/uiowa/uiowa/pull/6928
  - [x] assigned
  - [x] done
- drupal/acquia_purge                  1.2.0         requires          drupal/core (^8.9 || ^9)
  - [x] assigned
  - [x] done                                 
- drupal/administerusersbyrole         3.0.0         requires          drupal/core (^8 || ^9)  
  - [x] assigned
  - [x] done                                           
- drupal/better_normalizers            1.0.0-beta5   requires          drupal/core (^8.7.7 || ^9)        
**On first glance we might not need this for d10. Need to confirm that core normalizers for default_content is all we need**    
  - [x] assigned
  - [x] done                       
- drupal/big_pipe_sessionless          2.0.0         requires          drupal/core (^9)        
  - [x] assigned
  - [x] done                                 
- drupal/components                    2.4.0         requires          drupal/core (^8.7.7 || ^9)
**This looks murky, seems like there are some issues with d10 for it.**                
  - [x] assigned
  - [x] done            
- drupal/config_filter                 2.2.0         requires          drupal/core (^8.8 || ^9)        
  - [x] assigned
  - [x] done               
- drupal/config_ignore                 3.0.0-beta1   requires          drupal/core (^8.8 || ^9)    
  - [x] assigned
  - [x] done                                                          
- drupal/crop                          2.2.0         requires          drupal/core (^8.8 || ^9)    
  - [x] assigned
  - [x] done                             
- drupal/cshs                          3.6.0         requires          drupal/core (^8 || ^9)    https://github.com/uiowa/uiowa/pull/6727  
  - [x] assigned
  - [x] done                             
- drupal/devel                         4.1.5         requires          drupal/core (^8.8 || ^9)  #6943 
  - [x] assigned
  - [x] done                          
- drupal/draggableviews                2.0.1         requires          drupal/core (^8 || ^9)    https://github.com/uiowa/uiowa/pull/6729                               
[...]

PR #6953, opened 2023-10-02
Uninstalls the `components` module

pyrello:
Related to #6901 

I believe this module was originally introduced to allow us to setup a namespace to directly consume the templates from UIDS. However, we moved away from this practice a while ago and as far as I can tell, there are no instances of `'@uids/` in use in the code base. Which means we can remove it.

# How to test
```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli /
```
- Check out the homepage and validate there are no issues.
- Click around the frontend of the site a little bit.


bspeare:
Testing

bspeare (approved):
Tested on a few sites and everything is still working.

PR #6881, opened 2023-09-13
[Gallery block] Image gallery block enhancements

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6650
Resolves https://github.com/uiowa/uiowa/issues/5386
Resolves https://github.com/uiowa/uiowa/issues/5566

This **could** solve https://github.com/uiowa/uiowa/issues/4795 in that it offers a "No crop" option where, if you uploaded faculty photos of all the same size, the focal point would be respected.  Perhaps if this pr gets merged we can discuss the next steps in that issue.

```
ddev blt frontend && ddev blt ds --site=karandikar.lab.uiowa.edu && ddev drush @labkarandikar.local uli /lab-photos/group-photos-over-years
```

```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /multiple-image-gallery-layout-builder-blocks
```
Also test https://sandbox.uiowa.ddev.site/news/2022/09/alan-makes-gallery

## Test Housing and Classrooms grids

```
ddev blt frontend && ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local uli /abw-110
```

```
ddev blt frontend && ddev blt ds --site=housing.uiowa.edu && ddev drush @housing.local uli /residence-halls/catlett
```

## Test P2LB conversion

```
ddev blt sync --site hawkeyemarchingband.uiowa.edu --define drush.aliases.remote=hawkeyemarchingband.test && ddev drush @hawkeyemarchingband.local config-split:activate p2lb -y && ddev drush @hawkeyemarchingband.local uli admin/content/sitenow-converter
```

Convert the "Test page" and scroll all the way to the bottom and confirm that masonry was converted to the masonry style.

PR #6962, opened 2023-10-04
[Hotfix] Added fix for card styles option missing

bspeare:
In https://github.com/uiowa/uiowa/pull/6881/files, I put my `#states` code in the wrong place.  This issue fixes that. 

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/21/layout
```

- Edit/add a card and confirm that "Styles" option exists and "Align button" option exists
- Edit/add an Image gallery and confirm that "Styles" option exists if "No crop" option is selected
- Edit/add a text area block and confirm that text area padding option has `None` as the default option


cory-skeers:
Pulling down to test

cory-skeers (approved):
Reviewed and tested. Looks good and works well!

Got lost for a bit trying to keep track of breaks and fallthroughs when reviewing, but I think this looks good.

Issue #6868, opened 2023-09-08
Probe migrating media assets from v1 to v3 for CCOM sites


briand44:
# Story

As a CCOM site owner transitioning to SiteNow v3, I would like to have my media assets from my existing site easily available to me so I don't have to try and find the asset on my existing site and try to figure out how to download the original file in order to upload it to my new site.

# Questions

Would it be possible to migrate files to the new sites after provisioning?

Copy existing files directory to a more accessible location?

Other thoughts/ideas?

# Proposed solution

- [x] Probe using our existing file migration for CCOM sites
- [x] Any updates needed to existing migration for this to work
- [x] Can we connect or do we need to sync locally
- [x] Outline what the process would look like
- [x] Other options that could make transitioning media easier

cory-skeers:
It hasn't been used in a long time, but we have a default file migration as part of `sitenow_migrate`; https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/source/Files.php

Because files are fairly standard, this should more-or-less just involve enabling the `sitenow_migrate` feature split, setting up the config, and checking for the `d7_file` migration registered correctly.

The other migrations we've been doing are based off of https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/CreateMediaTrait.php instead, so part of this we'd probably want to look into the trait and see if there are updates that should be made to the Files if we went that route.

Alternatively, there's a process plugin https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_migrate/src/Plugin/migrate/process/CreateMediaFromFile.php that could probably be used with a basic migration definition YAML using some of the core migration stuff.

The original process we synced the d7 and d9 sites and files locally, did the migration, then synced the d9 sites/files back up, and could be a backup option if there's any difficulties connecting between applications.

briand44:
Grooming conversation... downside is that we may be storing a bunch of files that never get used or are not optimized. Files migrated would not be cleaned up by core processes.

multi-file upload?

If we did a file migration, would need to enable the feature split, configure the db info and run the migration and then turn it off. We may run into issues connecting to the db which would require us to sync the files locally first.

If articles were to be migrated, files associated with the articles would come over as part of that migration.

joewhitsitt:
[...]

PR #6964, opened 2023-10-04
[Migration] Public files path option

cory-skeers:
Relates to #6868 

# How to test

Sync a CCOM site locally using DevDesktop. (if you have just gotten access to the CCOM application, in DevDesktop you may need to click More > Refresh Acquia Cloud Sites in order to get it to appear)

```
ddev blt ds --site=default && ddev drush @default.local config-split:activate sitenow_migrate -y && ddev drush @default.local uli /admin/config/sitenow/sitenow-migrate 
```

Complete the config page using the help text.
For the new "Public Files Path", it should be something like "sites/**medicine.uiowa.edu.internalmedicine**/files" where the bolded should match the multisite designation in devdesktop. (You can also log into the site and visit `/admin/config/media/file-system` to check)

```
ddev drush @default.local ms
```
Check that migrations, particularly the `d7_file` migration, are registered.

```
ddev drush @default.local mim d7_file --limit=10
```
Feel free to adjust the `limit`, or remove it to import all files.

Check for no errors
Check the https://default.uiowa.ddev.site/admin/content/media, and see that media entities have been created for the imported files

Try syncing a uiowa70x site and hooking up, running the files migration _without_ filling in the public files path, and see that it works (a code review may also be sufficient, and that the logic should in that case match `initializeIterator` in https://api.drupal.org/api/drupal/core%21modules%21file%21src%21Plugin%21migrate%21source%21d7%21File.php/class/File/8.9.x)

### DIFF
--- a/docroot/modules/custom/sitenow_migrate/src/Form/MigrateSettingsForm.php
+++ b/docroot/modules/custom/sitenow_migrate/src/Form/MigrateSettingsForm.php
@@ -161,6 +161,13 @@ class MigrateSettingsForm extends ConfigFormBase {
       '#default_value' => $config->get('shared_configuration.source.constants.source_base_path'),
     ];
 
+    $form['constants']['public_file_path'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Public Files Path'),
+      '#description' => $this->t('The public files system path, if not set as <em>file_public_path</em> in the database <em>variables</em> table. Eg. <em>sites/default/files</em>'),
+      '#default_value' => $config->get('shared_configuration.source.constants.public_file_path'),
+    ];
+
     return $form;
   }
 
@@ -173,6 +180,7 @@ class MigrateSettingsForm extends ConfigFormBase {
         'key' => 'drupal_7',
         'constants' => [
           'source_base_path' => $form_state->getValue('source_base_path'),
+          'public_file_path' => $form_state->getValue('public_file_path'),
           'drupal_file_directory' => 'public://' . date('Y-m'),
         ],
         'database' => [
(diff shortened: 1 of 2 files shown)


bspeare:
@cory-skeers does the person testing this need CCOM access to their Acquia application?

cory-skeers:
> @cory-skeers does the person testing this need CCOM access to their Acquia application?

Yes, the tester will need to be able to access their Acquia application in order to pull it down to DevDesktop. The Signage group in AC was added, so people who had not had access previously may now have access.

joewhitsitt:
pulling to test

joewhitsitt:
Not sure if this is related or not, but in grabbing the CCOM assets, they don't have a file extension. There is not thumbnail, but i can can place the media item. I get an error trying to resave the entity.

Screenshot 2023-10-06 at 1 29 09 PM

With the 703 migration, thumbnails for most appear. I can re-save, etc.

I am going to test with another site but need to step away again for a bit.

cory-skeers:
> Not sure if this is related or not, but in grabbing the CCOM assets, they don't have a file extension.
> I am going to test with another site but need to step away again for a bit.

Thanks @joewhitsitt!

I'd originally tested with anesthesia and internalmedicine and they worked fine from what I'd checked, if you want to try with one separate from those.

What site were you running into this with? I'd be curious to take a look, and try and determine if it's something we want to try and account for, whether by appending extensions where there aren't ones, or skipping if there's something wrong with them in the source.

In the past we've run into issues where files had the wrong extensions, but in those cases, the file copy operation is smart enough to account for and correct it, we just end up with extra noise in the message logs. In this case, if we can at least tell that it's an image, we may even be able to just slap a `.jpg` at the end of it if we can't easily tell the specifics and let it handle things from there.

joewhitsitt:
Still away from my desk but I originally picked immunology but realized it now redirects.. Switched to dermatology and got some headshot files migrated over. Yes, I will test more sites in a bit with a fresh setup

joewhitsitt:
Perhaps it was just something weird with dermatology. Much better results with alumni

Screenshot 2023-10-06 at 2 47 29 PM

joewhitsitt (approved):
The scope of the PR works. Nice find and nice solution!

PR #6970, opened 2023-10-04
Update `google_tag` to 8.x-1.6

### LINKED ISSUE #6900: Update default GTM config
# Story

As a site owner using Google Tag Manager, I would like my scripts to fire on appropriate pages so that we can get the expected data for those pages.

# Additional Info

The default GTM config is excluding `/admin*` which prevents GTM scripts from firing on pages such as `/administration`.

Should exclude `/admin` and `/admin/*`

Check for `/admin*` and similar `/batch*` etc. Compare with Google Analytics setup.

And may be an upstream issue with GTM

# Proposed Solution

- [ ] Update GTM /admin* to /admin and /admin/*
- [ ] Update GTM /batch* to /batch and /batch/*

pyrello:
Resolves #6900 and updates `google_tag` to a D10 compatible version without going to the most recent update.

# How to test

Code review is sufficient. I have tested the update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -128,7 +128,7 @@
         "drupal/fullcalendar_view": "^5.0",
         "drupal/geofield": "^1.53",
         "drupal/google_analytics": "^4.0",
-        "drupal/google_tag": "^1.3",
+        "drupal/google_tag": "^1.6",
         "drupal/heading": "^1.4",
         "drupal/honeypot": "^2.1",
         "drupal/iframe_title_filter": "^2.0",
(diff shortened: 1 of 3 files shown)


pyrello:
BTW, it looks like the change to the 2.x version of Google Tag is going to be much more involved because a lot is changing: https://git.drupalcode.org/project/google_tag/-/compare/8.x-1.6...2.0.2?from_project_id=57970&straight=true. 

We'll likely need to make it its own task. After D10 :)

joewhitsitt (approved):
looks correct to me.

PR #6975, opened 2023-10-06
Uninstall `oembed_providers` module and associated code/config changes

### LINKED ISSUE #4533: Remove Kaltura support
after 12/28? - Confirm with OTLT

pyrello:
Uninstalls the `oembed_providers` module and removes the associated code/config. This change means that we are reverting back to the core functionality for this.

Resolves #4533 

# How to test

```
```
- https://sandbox.uiowa.ddev.site/media-kaltura/kalturamediaspace-small-captions - See that nothing renders except the caption where the kKaltura media entities have been placed
```
ddev drush @sandbox.local uli
```
- Test editing existing media embeds (changing style) for YouTube and Vimeo in a WYSIWYG field (like body)
- Test deleting existing media embeds for YouTube and Vimeo in a WYSIWYG field (like body)
- Test adding new media embeds for YouTube and Vimeo in the body field
- https://sandbox.uiowa.ddev.site/media/remotevideo/121 - Delete the Kaltura video
- https://sandbox.uiowa.ddev.site/media-kaltura/kalturamediaspace-small-captions - Confirm that the missing media entity widget is rendering for the embeds referencing the media entity you just deleted


bspeare:
Pulling to test

pyrello:
Not sure how I was seeing that this was working before, but its definitely not.

pyrello:
Still not sure what exactly is happening, but when I look at the database table where the video URL gets stored, I see that somehow all the existing field values are marked as deleted. Not sure what sequence of things causes that to happen.

pyrello:
An issue that we have to work around in being able to successfully uninstall `oembed_providers` is that it becomes a dependency of the `media.type.remote_video` config. When we uninstall the module, it deletes the config for the media type. It then reinstalls the media type, but when it gets removed all the media field data for the field get deleted. To work around this issue, I manually removed the dependency on the `oembed_providers` module from the config. However, the issue is that the config in the database (which still includes the dependency) is used at first. I think I fixed it by adding a small update hook to pre-load the updated version of the `media.type.remote_video` config so that the deletion of the media type is never triggered.

bspeare (approved):
Testing steps all worked.   Youtube and Vimeo videos are no longer being deleted.

PR #6997, opened 2023-10-12
Move aggregator module to contrib

### LINKED ISSUE #5064: Add contrib Aggregator module when switching to Drupal 10.x
Per https://www.drupal.org/node/3267461, the aggregator module is being removed from Drupal core in 10+ and its functionality is being moved into a contrib module: https://www.drupal.org/project/aggregator

pyrello:
Resolves #5064 

# How to test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,6 +90,7 @@
         "drupal/acquia_purge": "^1.0",
         "drupal/address": "^1.8",
         "drupal/administerusersbyrole": "^3.0",
+        "drupal/aggregator": "^1.1",
         "drupal/allowed_formats": "^1.1",
         "drupal/antibot": "^2.0",
         "drupal/better_exposed_filters": "^6.0",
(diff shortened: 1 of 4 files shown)


pyrello:
Here is the process I used to update the config:
```
git checkout main && ddev composer install && ddev blt ds && git checkout contrib_aggregator && ddev drush @sandbox.local updb -y && ddev drush @sandbox.local cex
```
I had previously created the `contrib_aggregator` branch which included the composer addition of `drupal/aggregator`.

I think I've been doing it wrong in other dependency updates that includes config updates.

cory-skeers (approved):
Tested, and everything looks good, not noticing any regressions in functionality.

I was curious about the config changes dropping the `teaser_length`, but looks like it was removing a non-functional config setting https://www.drupal.org/project/aggregator/issues/2283877

PR #7011, opened 2023-10-16
[ITS] Update taxonomy perms

cory-skeers:
# How to test

`ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli`
Masquerade as a webmaster
Check that you can create, edit, and delete Audience and Service Category taxonomy terms
Masquerade as an editor/publisher
Check that you cannot create, edit, or delete Audience and Service Category taxonomy terms

### DIFF
--- a/config/sites/its.uiowa.edu/config_split.patch.user.role.webmaster.yml
+++ b/config/sites/its.uiowa.edu/config_split.patch.user.role.webmaster.yml
@@ -2,13 +2,21 @@ adding:
   dependencies:
     config:
       - node.type.service
+      - taxonomy.vocabulary.audience
+      - taxonomy.vocabulary.service_category
   permissions:
     - 'create service content'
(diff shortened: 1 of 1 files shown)


cory-skeers:
@briand44 I added perms for both the Service Category and Audience taxonomies, but limited it to the webmaster. The standard taxonomy (Tags and Research Areas) perms also include Editor and Publisher in being able to create, edit, and delete terms. 

I was guessing that this may be similar to Service content that we limited to only webmasters and would be desired to limit the number of users that could edit, but would appreciate your thoughts on that.

briand44:
I think that makes sense. If they don't want editors and publishers creating service content, I doubt they want them creating service categories.

joewhitsitt:
pulling to test

joewhitsitt (approved):
checks out

PR #7032, opened 2023-10-23
[Facilities] Update Dispatch to use custom fields in subscriber signup

cory-skeers:
# How to test

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli admin/config/sitenow/sitenow-dispatch
```
Go to [facilities.uiowa.edu](https://www.facilities.uiowa.edu/admin/config/services/facilities-alerts-emails) and get the dispatch API key and enter it here https://facilities.uiowa.ddev.site/admin/config/sitenow/sitenow-dispatch

Edit the Dispatch base path locally to pull from the test endpoint: https://test.its.uiowa.edu/dispatch/api/v1/
https://github.com/uiowa/uiowa/blob/4918096632bfc29d74dcf7d6e3ec562023a97102/docroot/modules/custom/sitenow_dispatch/src/DispatchApiClient.php#L44-L46

Check out the Alerts sign up form at https://facilities.uiowa.ddev.site/alerts

Configure the signup form in the [Dispatch test environment](https://test.its.uiowa.edu/dispatch/populations/574308527/configure) and add new fields and toggle the display of the phone number field

See that the changes are reflected in the SiteNow form.

Turn on xdebug and submit the form, catching it prior to https://github.com/uiowa/uiowa/blob/4918096632bfc29d74dcf7d6e3ec562023a97102/docroot/modules/custom/sitenow_dispatch/src/Form/SubscribeForm.php#L93C1-L99C8

Check that each parameter matches the keys for the Dispatch side of things, and with strings containing the values entered into the SiteNow form.

## Followup

- [x] Communicate with Michael A if this all looks good for things to go into prod on their end


cory-skeers:
In EoD, found the 'required' option does not appear to be working, and the select (dropdown) is adding an additional blank option. Looking into these.

joewhitsitt:
pulling to test

cory-skeers:
One thing I hadn't thought until just now--I didn't specifically test this with the prod endpoint prior to changes going into place, and we may want to double-check proper checks are in place in case that happens.

joewhitsitt:
I'm not saying this is an "us" problem but I was able to sneak some trickery past on the dispatch end

Options:
```
blue
red,Red
```
>Deprecated function: explode(): Passing null to parameter #2 ($string) of type string is deprecated in Drupal\sitenow_dispatch\Form\SubscribeForm->processCustomField() (line 235 of modules/custom/sitenow_dispatch/src/Form/SubscribeForm.php).

joewhitsitt:
Also I wonder if we should add help text to the block form to suggest re-saving the block after each time the population side is configured to flush caches

joewhitsitt:
>One thing I hadn't thought until just now--I didn't specifically test this with the prod endpoint prior to changes going into place, and we may want to double-check proper checks are in place in case that happens.

Can we just hold on merging until the dispatch side is deployed?

cory-skeers:
> > One thing I hadn't thought until just now--I didn't specifically test this with the prod endpoint prior to changes going into place, and we may want to double-check proper checks are in place in case that happens.
> 
> Can we just hold on merging until the dispatch side is deployed?

I think that's fine. I'd started thinking that maybe it would be good to add extra checks in case anything changes on the Dispatch side, but 1) I don't think there are frequent changes and 2) we'd still want to log those failures to make them visible, so maybe it's alright as it is.

joewhitsitt:
>Warning: Undefined property: stdClass::$subscriptionList in Drupal\sitenow_dispatch\Form\SubscribeForm-&gt;buildForm() (line 79 of modules/custom/sitenow_dispatch/src/Form/SubscribeForm.php)

This is what happens when I switch the endpoint and the functionality isn't there.

docroot/modules/custom/sitenow_dispatch/src/Form/SubscribeForm.php
```diff
@@ -123,4 +154,87 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+      }
+      $form['custom_fields'][$custom_field->key]['#options'] = $options;
+
+      // Checkboxes take an array as their default value,
+      // unlike the other field options.
+      // Split on commas and replace the single string
+      // that we added prior.
+      if ($field_type === 'checkboxes') {
+        $form['custom_fields'][$custom_field->key]['#default_value'] = explode(',', $custom_field->defaultValue);
+      }
```
cory-skeers:
```suggestion
      if ($field_type === 'checkboxes' && !empty($custom_field->defaultValue)) {
        $form['custom_fields'][$custom_field->key]['#default_value'] = explode(',', $custom_field->defaultValue);
      }
```

joewhitsitt (approved):
Nice work! Very clean and well documented

cory-skeers:
Switching to draft until the updates are available in prod on the Dispatch side of things.

cory-skeers:
Changes are live on the API-side of things, and this is good to be merged now.

PR #6994, opened 2023-10-11
Upgrade to Drupal 10.0

pyrello:
Resolves #6959 
Resolves #6537 
Resolves #5738
Resolves #5454
Resolves #6566

# ToDo

- [x] Update tippie split once https://github.com/uiowa/uiowa/pull/7033 merges with the robots.txt changes
- [x] Fix layout builder settings tray regression
- [x] Resolve CCOM split issue - _https://github.com/uiowa/uiowa/pull/6994#issuecomment-1777826089_

# Explanation of some of the configuration differences

- Some sites had rabbit hole before default had it, so this PR cleans those instances up.
- Admissions, the events categories taxonomy was moved to the feature so the site split no longer needs those permissions.
- Robots.txt was updated based on a core change to hide oembed routes.
- ighn.international permission clean up. v2 permissions related to blocks take precedence.
- Some modules updated and provide a new setting or replace an existing

# How to test
- Check to see if update to typhonius/acquia-php-sdk-v2 can close https://github.com/uiowa/uiowa/issues/5738
  - `./vendor/bin/blt uiowa:multisite:create foo.uiowa.edu --no-db --no-commit` - See no early warnings in the command. Ctrl+C to exit.
- Check to see if update to drupal/redirect solves https://github.com/uiowa/uiowa/issues/1072 (I know the linked upstream issue is still out there, but I can't reproduce the issue)
- A patch, "Support third party settings for components within a section," was updated related to https://www.drupal.org/project/drupal/issues/3015152
  - This was introduced to make our unique id field on blocks a reality and we have run with it since. To test, make sure you are still able to see/add/edit unique ids on our blocks. 
- A patch, "Changing an existing embedded media's alignment or alt data attributes does not get saved with CKEditor," was removed in favor of the now contrib CKEditor module's version https://www.drupal.org/project/drupal/issues/3102249 -> https://www.drupal.org/project/ckeditor/issues/3330723
  - To test, go to https://sandbox.prod.drupal.uiowa.edu/node/251/edit and make sure that when you change alignment for the media item, it retains your selection after save.
- A patch, "Hidden buttons in off-canvas dialog are not being hidden," was removed related to https://www.drupal.org/project/drupal/issues/3070375
  - A CSS rule was added to fix the button showing up.
  - To test, with the media library widget, there is a button that is supposed to be hidden labeled "Update widget," and it gosh darn better stay that way.
[...]

Issue #6835, opened 2023-08-31
Block ChatGPT for Tippie paths


briand44:
# Story

As a SiteNow webmaster, I would like to block ChatGPT from some of my content but allow search engines to still index those pages in order to protect our information.

# Additional Information

Seems like currently this would need to be handled by administrators manually updating robots.txt?

Looks like you can target a specific User-agent `User-agent: GPTBot`

# Questions

Would we want to allow webmasters ability to edit robots.txt?

Is there anything available or in the works as part of metatag?

I would assume dev and stage environments would be handled already since we are disallowing all?

# Proposed Solution

joewhitsitt:
https://thehackernews.com/2023/08/how-to-prevent-chatgpt-from-stealing.html

This article mentions user-agent: CCBot

briand44:
interesting... I've seen a couple articles mention GPTBot but they were published in early August.

cory-skeers:
> https://thehackernews.com/2023/08/how-to-prevent-chatgpt-from-stealing.html
> 
> This article mentions user-agent: CCBot

ccbot is used by https://commoncrawl.org/, which gets used for a lot of tools/research and was probably used by gpt originally. It's my understanding that for future models, gpt will be using only its own crawlers (largely for tracking/legal reasons), but it's still in flux, so may or may not stay the case.

It does raise the question though--what is the purpose in blocking GPT? If it's to prevent OpenAI specifically, then I think gptbot should cover it. But if it's protecting content from being crawled and used in tools like language models, we should also cover ccbot, and will likely need to expand it even more.

joewhitsitt:
starting to sound like a rabbit hole of sorts... chatgpt is not the last one that is going to make folks uneasy over this.

mjoneill:
OpenAI's documentation on robots exclusion for Plugins:

`User-agent: ChatGPT-User`

https://platform.openai.com/docs/plugins/bot

mjoneill:
OpenAI's documentation for exclusion on core bot:

`User-agent: GPTBot`

https://platform.openai.com/docs/gptbot

JKurtzlebenUI:
I meant to add our business reason for this earlier, and it fell off my list.

[...]

PR #7033, opened 2023-10-23
Tippie - Split robots.txt

joewhitsitt:
Has the robots.txt adjustment from https://github.com/uiowa/uiowa/issues/6835

Screenshot 2023-10-23 at 5 10 59 PM

Unfortunately this can't be partially split as it is all one line. We will have to update with the d10 updates once this is merged https://github.com/uiowa/uiowa/pull/6994/commits/b57ac7aa5603a416260ebea32f7d94ff35f691b5


pyrello:
I wondering about, instead of doing this for Tippie, should we just make this the default for the platform? Opt out of being crawled by AI crawlers (that observe the directive) until we can better understand it and/or someone asks for their content to be available to be crawled?

joewhitsitt:
So you suggest disallow / for the user agent GPTBot for all sites? Tippie is only doing the /advisor path currently.

>The rest of the Tippie we[b] properties can certainly be open to generative app crawlers. For The Advisor, I’ll continue to look for every way I can find to block generative app crawlers.

cory-skeers:
> I wondering about, instead of doing this for Tippie, should we just make this the default for the platform? Opt out of being crawled by AI crawlers (that observe the directive) until we can better understand it and/or someone asks for their content to be available to be crawled?

As silly as it may be, I worry that ChatGPT and its ilk may be used (or be used in the future) for things like prospective students looking for information on UI, admissions, etc. The tool absolutely should not be expected to be accurate in that realm, but the use of digital tools often doesn't match up with "should" (thinking of people using Instagram search features over Google's). And in that case, even if it's a tiny niche of the billions of parameters that get slightly tuned toward accuracy by crawling our actual content, I would lean toward allowing the tools that are going to be used to crawl (the majority) of our sites.

I think we absolutely need to keep the conversation going, but in the interim, I'd be in favor of an opt-out approach like this until we come up with a policy on what we block and how we determine what gets blocked (block any for-profit tools? any generative tools regardless of for-profit or non-profit, etc).

pyrello:
After a conversation about this in Daily Scrum, I see there is more complexity to this and differing opinions that will require more time and information to sort out. As such I think we should proceed with this change for now.

pyrello (approved):
Reviewed, not tested.

PR #7044, opened 2023-10-25
Article Scheduling

### LINKED ISSUE #723: Scheduling article Content
I think this is definitely blocked until we get away from poor man's cron, but Wordpress offers it out of the box and several clients have asked for it in the past.

This would need to tie into our moderation workflow if implemented.

**Update (bdickey)**
Created a new IS at https://github.com/uiowa/uiowa/issues/723#issuecomment-1758037258 and scoped to focus on scheduling publishing of articles for the service and scheduled unpublishing only for ITS for now. Scheduled upublishing for articles and adding scheduling functionality to Pages or other entity types can come later in separate issues. Scheduling blocks/sections is discussed in #5405

### LINKED ISSUE #3066: Remove temporary claro node form CSS
# Motivation

#3065 introduced a temporary CSS override to "fix" an issue that was introduced in 9.1 with the node form in Claro. There are open issues to address this issue:
* https://www.drupal.org/project/drupal/issues/3184667
* https://www.drupal.org/project/drupal/issues/3181964
* https://www.drupal.org/project/drupal/issues/3076820

# Proposed solution

Whenever a workable fix drops via the above issues and we update to that code, we can remove the changes introduced in #3065.

joewhitsitt:
Resolves #723 
Resolves https://github.com/uiowa/uiowa/issues/3066

See modified issue summary: https://github.com/uiowa/uiowa/issues/723#issuecomment-1758037258

# To do

- [x] Fix article.minimal config diff with 91st Meridian
- [x] Fix config diff with SPPA

# Scheduler with moderation/revisions

Screenshot 2023-10-25 at 12 11 49 PM

# Iowa Now embargo protection

Screenshot 2023-10-26 at 9 24 25 AM

Screenshot 2023-10-26 at 9 23 14 AM

>jwhitsit@ITS-02342 uiowa % ddev drush @now.local cron    
 [warning] Publishing of "test" is prevented until after the content embargo has been lifted.
 [notice] Message: The content will only be published after the content embargo has been lifted.

# Odd publish/publish conflict protection
Screenshot 2023-11-03 at 10 04 04 AM

# To Test

- `ddev composer install`
- No import errors on any custom or feature split site as you sync and test. Inspect modified config. No surprises.

## Default

- Starting on the default site, `ddev blt di --site=default`, as a webmaster/publisher, you should now see "Scheduling options" for publish on within the node article form. Editors should still be able to set for article for review but cannot see the scheduling options fieldset.
- Note that when scheduling a date and time, the seconds input is not longer involved. You can compare this to facilities.prod.drupal.uiowa.edu/node/add/alert
- Custom help text has been added to the the publish on field to explain that despite the time selected, the article will be published within one hour.
- Though there is a new "publish_state" field, at no point in the form editing process is that field visible to the publisher/webmaster.
- Select a date and time to publish the article (a minute or so) and keep the article in a draft or needs review state.
- On save, note that the content is still unpublished logged in and as an anonymous user. As authenticated, you should see messages that the content has been scheduled and with moderation involved, additional text about Editors being restricted until after the scheduled date.
- Since you are local, manually trigger cron after reaching the time set within scheduling options. `ddev drush @default.local cron`
- You should see a message in the terminal saying that the content is being published and upon refreshing the browser you should see that it is published.
- Create a new article and this time in addition to setting a publish on date/time, try to set the moderation state to published and save the node. You should see that scheduling is still set but that the moderation state has been moved back to draft. Go to /admin/content and try to bulk publish the same article. Again you should see a message saying that the two settings conflict and that the moderation state has been set back to the previous state.

## Now

- Feel free to repeat similar steps from default on the Iowa Now website `ddev blt ds --site=now.uiowa.edu`
- If there is an embargo message on the article within Iowa Now, you will still be able to schedule a publish on date, but each time you run cron after the date, it will not be allowed to proceed and the core status/moderation state will stay as it was.

## Facilities

- Confirm there are no alert split regressions as it relates to the alert/site split. `ddev blt ds --site=facilities.uiowa.edu` See https://github.com/uiowa/uiowa/pull/6662 for previous testing detail.
- For alerts, the messaging should not include mentions about editor moderation restrictions.

## ITS

- Feel free to repeat similar steps from default on the ITS website `ddev blt ds --site=its.uiowa.edu`
- ITS has both publish on and unpublish (archive) on functionality for articles.
- Same as with publish on, archive on's unpublish_state field is hidden from webmasters/publishers. Editors again cannot interact with the scheduling options fieldset.
- In testing the archive on scheduling option, there is some validation already to prevent invalid transitions. The downside to this is that there are validation messages that link to the hidden unpublish_state field.
- From a published article, edit and set a archive on date/time and save. Go to admin/content and try to bulk archive the latest revision. If the operation occurs before the archive on date/time, you should see a similar message saying the two settings conflict and that it was switched back to the previous state. the archive on date is still set and will archive the article on the first cron after the date/time has passed.

- See the latest docs draft: https://github.com/uiowa/uiowa/issues/723#issuecomment-1780127928


joewhitsitt:
With touching so much config, I am going to run a big `dsa` to check for import errors.

joewhitsitt:
Oye, it might be worth pausing until the `d10` branch merges in (https://github.com/uiowa/uiowa/pull/6994). I am getting a lot of simple sitemap diffs so far which I think got pretty good coverage there.

```
-  simple_sitemap:
-    weight: 10
-    region: content
-    settings: {  }
-    third_party_settings: {  }
```

update: apparently the d10 branch wasn't hit by this. ugh.

config/features/event/core.entity_form_display.taxonomy_term.event_categories.default.yml
```diff
@@ -46,4 +46,6 @@ content:
     settings:
       display_label: true
     third_party_settings: {  }
-hidden: {  }
+hidden:
+  publish_on: true
+  unpublish_on: true
```
pyrello, line 51: I'm curious why this is different than other config updates.

joewhitsitt: taxonomy doesn't have moderation so the publish_state and unpublish_state don't apply

joewhitsitt:
I think i found all of the `simple_sitemap` instances related to core.entity_form_display. need to follow-up monday with a ds on custom sites lower in the list than ITS and then go back to 91st meridian and add `core.entity_form_display.node.article.*` to the partial split and export similar to https://github.com/uiowa/uiowa/pull/7044/commits/10a0d20ca0d932b629ea3a2c356f4345c33baf74.

pyrello:
Pulling to test.

config/sites/hr.uiowa.edu/core.entity_form_display.taxonomy_term.units.default.yml
```diff
@@ -88,4 +88,18 @@ content:
...
+    region: content
+    settings: {  }
+    third_party_settings: {  }
+  status:
+    type: boolean_checkbox
+    weight: 100
+    region: content
+    settings:
+      display_label: true
+    third_party_settings: {  }
```
pyrello, line 102: Just double-checking that this should be here.

config/sites/now.uiowa.edu/core.entity_form_display.node.article.minimal.yml
```diff
@@ -0,0 +1,124 @@
+uuid: 5736330e-e26f-4d7d-8979-7fe6e218747a
```
pyrello: Is there a reason this needs to be a complete split?

pyrello:
Screenshot 2023-10-30 at 4 27 45 PM

There are a couple new tabs being added to the user's page. I'm wondering if we want to leave these here? Or should we disable them? My thinking is that it is a little confusing. Maybe not a bad idea, but not the thing I would choose to add for users if we were going to add something there.

config/sites/now.uiowa.edu/core.entity_form_display.node.article.minimal.yml
joewhitsitt: i imagine that was just exported that way. i can try to switch it to a partial. i wonder why caused this one to do that when it is defined in the partial split section for this site.

joewhitsitt:
I believe those views require permissions and webmaster/publisher/editor don't have them. I'd be okay with disabling the views (especially the media one) until the use-case comes along.

config/sites/hr.uiowa.edu/core.entity_form_display.taxonomy_term.units.default.yml
joewhitsitt: Hmm. Good eye. I think this is just a really old piece of config that hasn't been touched in a while. It looks like publish and simple sitemap is visible on other taxonomies (e.g. admissions). I'll try to adjust this one as well

joewhitsitt: wait, i misread this diff. this is correct and aligns with other taxonomies.

config/sites/now.uiowa.edu/core.entity_form_display.node.article.minimal.yml
joewhitsitt: This was fixed in https://github.com/uiowa/uiowa/pull/7044/commits/530e4461c5b9d5518d6b8217d6024566a688d836

pyrello:
Screenshot 2023-10-31 at 12 58 09 PM

I'm testing this as a publisher right now. I'm wondering about having the `Publish state` field showing even though it can't be changed.

joewhitsitt:
>I'm wondering about having the Publish state field showing even though it can't be changed.

Yeah this module doesn't align perfectly with our editor workflow. I didn't want to invest too much to change how it works. The patch makes it the default option and the disable prevents them clicking the dropdown only to see no other option. 

We have other instances where the disabled field pattern is used (using #access=false can affect saving the value). We talked about this in grooming a bit. Revealing the state it would be transitioning to seemed valuable especially for unpublished where we don't actually have an "unpublished" state, rather an "archived" state.

Open to ideas about how to approach this. Element hidden with CSS?

joewhitsitt:
EoD discussion

leave it for now and see if it becomes a problem. small benefit of providing a bit more context. Updating help text and labels could still be a step forward before we look into hiding. Because of the input tab index, if we were to hide later, we probably want to use `display: none;`

We want to resist harvesting just the parts we want out of this module and going it alone.

pyrello:
Not a 100% this still works, but I was thinking something like: 
```php
function hook_form_FORM_ID_alter() {
    $form['field_fieldname']['#disabled'] = TRUE;
    $form['field_fieldname']['#type'] = 'hidden';
}
```

Not clear on what the value is to the user to know what the future state will be. The field widget uses the terminology publish/unpublish, which I think is pretty clear even if it is a little inconsistent with our "Archived" state. If we concerned about that, maybe just change the field label to "Archive" in the case of scheduling un-publishing.

pyrello:
> Help text about the hour window are provided to both publish and unpublish.

I can see from looking at the code that the unpublish case applies and can be tested on ITS, but the instruction is a bit unclear that you have to test it that way.

pyrello:
I tested the Now embargo + scheduling functionality:
- Created an article with content in the embargo field and a scheduled publish date/time in the near future.
- After the publish date/time had arrived, ran cron using `ddev drush @now.local cron`.
- I saw the following:
```
[warning] Publishing of "ddasfasf" is prevented until after the content embargo has been lifted.
[notice] Message: The content will only be published after the content embargo has been lifted.
```
- Removed the embargo field content (and updated the publish date/time to be able to re-save).
- After the new publish date/time had arrived, ran cron again.
- I saw the following:
```
[warning] Publishing failed for ddasfasf. scheduler_content_moderation_integration_scheduler_publish_process returned a failure code.
[notice] Message: /ddasfasf/ is scheduled to be published Wednesday, November 1, 2023 - 8:42 
AM.
```
- Article was not published.
- Ran cron via the UI, but the article was still not published.

joewhitsitt:
>scheduler_content_moderation_integration_scheduler_publish_process returned a failure code

Hmm. I had not seen that earlier. I will take a closer look.

joewhitsitt:
So far it appears that changing the field to disabled and/or hidden is affecting the underlying code. I guess I didn't test this piece after adding the disabled bit.

Leaving it as an exposed drop down with only one option seems to be the way to get this to work in all conditions currently. hiding with CSS might be back in play?

There are some similar upstream issues - https://www.drupal.org/project/scheduler_content_moderation_integration/issues/3390085, https://www.drupal.org/project/scheduler_content_moderation_integration/issues/3312463

joewhitsitt:
I used the same js + js-hide combo as what is in our docroot/themes/custom/uids_base/scss/global.scss to override the inline style of display: block;

joewhitsitt:
Screenshot 2023-11-01 at 1 19 01 PM

pyrello:
I re-tested the scheduling + embargo and the above commits appear to have resolved the issue I was encountering before.

Overall, the changes here look good, especially with the tweaks to cleanup terminology and hide fields from displaying.

Last issue I'm seeing is what happens if a person selects a date/time to schedule publishing and then also sets the article to published:
Screenshot 2023-11-01 at 2 02 46 PM

I looked at this with @joewhitsitt and we tested if you can access the page as an anonymous user and you cannot. Not sure how this is happening exactly, since the content moderation state appears to be correct, but nonetheless, there it is.

joewhitsitt:
Not a solution, but clearing the date from scheduler allows full moderation transition.

joewhitsitt:
The scheduler_content_moderation_integration module doesn't seem to provide the right kind of protection from https://git.drupalcode.org/project/scheduler/-/blob/2.x/scheduler.module?ref_type=heads#L968 happening as it relates to content moderation.

joewhitsitt:
Adding a check to docroot/modules/contrib/scheduler_content_moderation_integration/src/Plugin/Validation/Constraint/PublishStateConstraintValidator.php

```
if ($from_state === $to_state) {
      return FALSE;
    }
```

prevents saving, but leads to an unhelpful error message for a now invisible field

Screenshot 2023-11-01 at 3 34 39 PM

Starting to wonder if we should do some presave or submit validation of our own similar to how we validate embargo/moderation_state.

joewhitsitt:
I just pushed up a validation constraint and it is working for node save. The documentation I've read seems to say this will also work for API driven type operations. However this doesn't seem to be a silver bullet.

What doesn't seem to work is when it is triggered by VBO, or more specifically `moderated_content_bulk_publish`. Validators within core's `content_moderation` don't seem to trigger either. Same behavior seems to occur using layout builder's interface to change states. This is revealing a bigger gap in what I previously thought was solid.

At the moment, I am not sure how we can cover all of these different situations.

joewhitsitt:
This morning I tested the previous presave validation for the embargo message on iowa now. It is triggered in all situations above and moves the moderation back safely to the previous state. This morning I am going to see if I can shift validation over to presave and cover all of our bases. I don't love this error message, but the end result is what we want.

Screenshot 2023-11-02 at 8 56 40 AM

joewhitsitt:
what is interesting is there seems to be partial coverage (node save) for unpublish but not publish docroot/modules/contrib/scheduler_content_moderation_integration/src/Plugin/Validation/Constraint/UnPublishStateConstraintValidator.php

what happens though is if bulk archived, the unpublish_on is still set on the node and cron reports that it failed to unpublish because it is already archived. not ideal.

it is very specifically the operation that scheduler is doing that is causing all the hubbub.

joewhitsitt:
Planning to update testing instructions after standup

joewhitsitt:
Test instructions have been updated

bspeare:
Testing this out.

bspeare:
I'm seeing a few differences on the Alert and Article nodes and I'm not sure this is an issue, but noting it here:
 
1. Set alert/article node to published
2. Edit alert/article node and set to published on date and save

- Alert will show pink draft background, but when you edit the node it says **"Current state: Published"** - is this the same as https://github.com/uiowa/uiowa/pull/7044#issuecomment-1789528713?  
  - Access denied when I try to view this "published" page as anonymous user. 
- Article will continue to be published with alerts of **"Content scheduling criteria for "Article title" conflicts with the set "published" moderation state. Switching back to "published."** while also it says **"Article title is scheduled to be published Tuesday, November 7, 2023 - 12:00 AM.**"

bspeare:
> There is some validation already to prevent invalid transitions. The downside to this is that there are validation messages that link to the hidden unpublish_state field.

Can you provide an example of an invalid transition to test?

joewhitsitt:
>I'm seeing a few differences on the Alert and Article nodes

I can pull down to get better idea of what you are seeing. I just remembered that ITS Alerts have been customized to use moderation and I haven't accounted for that. The logic at this point should be moved and scoped for multiple content types. Facilities alerts will be different. I'll set back to draft.

> > There is some validation already to prevent invalid transitions. The downside to this is that there are validation messages that link to the hidden unpublish_state field.
> 
> Can you provide an example of an invalid transition to test?

To trigger that one, I think if you set an unpublished date and mark as archived on the edit form, it will prevent save.

e-marie-w:
Pulling down to test, starting with default.

bspeare:
Tested https://github.com/uiowa/uiowa/pull/7044#issuecomment-1795691620 again on ITS and it is working as expected now.

e-marie-w:
I've tested default, Now, Facilities, and ITS successfully.

Only weird thing: after an article is scheduled to be published or unpublished, someone with an editor role cannot edit it. Editors can edit normal article drafts still and even the scheduled article once the date passes/cron runs. 

Just to check my understanding: articles on ITS are the only piece of content with the ability to schedule unpublishing at this time, correct? I got a little confused looking at the Facilities Alert PR.

joewhitsitt:
>Only weird thing: after an article is scheduled to be published or unpublished, someone with an editor role cannot edit it. Editors can edit normal article drafts still and even the scheduled article once the date passes/cron runs.

Oh interesting. I am pulling down to take a look at that.

>Just to check my understanding: articles on ITS are the only piece of content with the ability to schedule unpublishing at this time, correct? I got a little confused looking at the Facilities Alert PR.

That is correct.

joewhitsitt:
Yes that appears to be true and there is an issue for it. https://www.drupal.org/project/scheduler_content_moderation_integration/issues/3336702

Trying to determine how I feel about it. I see both sides, but I do lean towards the, don't hijack and try to implement new access controls, just do what the module is supposed to do.

joewhitsitt:
Switching back to draft to work on providing help text and messaging surrounding the additional access check functionality the module implements

joewhitsitt:
Screenshot 2023-11-09 at 10 05 35 AM
Screenshot 2023-11-09 at 10 06 41 AM

Screenshot 2023-11-09 at 10 09 31 AM

e-marie-w:
I am running into internet issues. I was able to sync most sites touched by this PR successfully. I have left:
- ~~classrooms.uiowa.edu~~
- ~~ighn.international.uiowa.edu~~
- ~~iisc.uiowa.edu~~
- ~~uiowa.edu~~
- ~~uipress.uiowa.edu~~

Connections were intermittently timing out when trying to sync, right now it's every time. I am going to see if I can pull the latest changes and see them on an already synced site. I did view the documentation change, that looks great. 

Update: it fixed itself I have successfully synced these sites now.

joewhitsitt:
Commented on the upstream issue as this is starting to be a pain to provide reliable messaging for

joewhitsitt:
Doing a combination of form_alter messaging and preprocess_node messaging. Provides consistency where possible whether or not the content type has moderation.

Screenshot 2023-11-09 at 11 53 05 AM
Screenshot 2023-11-09 at 11 53 25 AM
Screenshot 2023-11-09 at 11 53 34 AM
Screenshot 2023-11-09 at 11 53 49 AM

joewhitsitt:
In preparing for demo, I noticed there are similar but different messages related to scheduler rendering at the same time depending on the operation. Going to see if I can reduce that.

joewhitsitt:
The Drupal Messenger service doesn't allow altering messages already in the queue. Examples I have seen have all been, delete existing and re-add with the modified message.

joewhitsitt:
Oye. The PR shows the added bloat to achieve this but here we go:

_Most recent focus was to unset and modify messages._

Screenshot 2023-11-09 at 4 20 05 PM

_Keep this form_alter one around for unpublished (no moderation) and published (moderation) instances._

Screenshot 2023-11-09 at 4 20 16 PM

_Then the lovely attempt to wrangle the bulk publish situation._

Screenshot 2023-11-09 at 4 21 31 PM

joewhitsitt:
Realized that ITS alert is still not in a good spot. Might push a change here shortly.

joewhitsitt:
Follow-up from https://github.com/uiowa/uiowa/pull/7044#issuecomment-1804780040, to reduce the bloat, I have modified it to just state the new thing, add it to the queue and output them all.

Screenshot 2023-11-27 at 2 49 11 PM
Screenshot 2023-11-27 at 2 49 22 PM

https://github.com/uiowa/uiowa/pull/7044/commits/7fd00d0711e859444ed3293b7da842617a0317c3

joewhitsitt:
Putting this back up for review. No big changes since it was delivered at demo. Fixed the issue raised in https://github.com/uiowa/uiowa/pull/7044#issuecomment-1802674353 with https://github.com/uiowa/uiowa/pull/7044#issuecomment-1828584859 and the form_alter portion of https://github.com/uiowa/uiowa/pull/7044#issuecomment-1804297882

Website doc page was updated and is ready for publish. https://github.com/uiowa/uiowa/issues/723#issuecomment-1804001591

docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -241,6 +287,27 @@ function uiowa_core_entity_view(array &$build, EntityInterface $entity, EntityVi
...
+  if (!$admin_context->isAdminRoute()) {
+    // Get the current user.
+    $user = \Drupal::currentUser();
+    // Showing a message to viewers and up about how scheduling locks
+    // editors from editing the content for content with moderation.
+    if ($user->hasPermission('view any unpublished content') === TRUE) {
+      if (isset($variables['publish_on']) && $variables['node']->moderation_state?->value) {
+        if ($variables['view_mode'] === 'full') {
+          \Drupal::messenger()
+            ->addStatus(t('Due to scheduling, editors are restricted from editing until published.'));
```
pyrello:
```suggestion
            ->addStatus(t('This content has been scheduled for publishing and editors are restricted from editing until published.'));
```

joewhitsitt: Some redundancy with the "`node:title` is scheduled to be published `node:publish_on`" message that may also appear at the same time but I am fine with this.

joewhitsitt:
Switching to draft to resolve conflicts. Just realized docroot/modules/custom/uiowa_core/css/claro-node-form.css was removed in the 10.1 branch... I was counting on that.

joewhitsitt:
Still need to run through ITS and Facilities site splits to confirm they aren't splitting anything additional not already covered by default and sitenow_alerts.

pyrello:
> Switching to draft to resolve conflicts. Just realized docroot/modules/custom/uiowa_core/css/claro-node-form.css was removed in the 10.1 branch... I was counting on that.

Looking at the style rule that gets added, this seems generic enough that maybe it could just get added to a global stylesheet instead of this? Was it just using the node form CSS because it was already set up load at the correct time?

joewhitsitt:
Is global.scss attached to non-layout builder admin pages?

Update: `uiowa_edu_core` calls global-styling in preprocess_page but the same call isn't done for all sites. AFAIK, the claro form css is the only thing we have custom that is loading for these forms.

joewhitsitt:
I think config has been cleaned up since the 10.1 merge.

Didn't do anything with the claro form css for the moment based on https://github.com/uiowa/uiowa/pull/7044#issuecomment-1838827587

bspeare:
I am going to start working on testing this.

bspeare:
@joewhitsitt getting config error on ` core.entity_form_display.node.alert.default   Different` for facilities.

joewhitsitt:
> @joewhitsitt getting config error on ` core.entity_form_display.node.alert.default Different` for facilities.

I'll take a look tomorrow morning. Perhaps this is why the config was setup the way it was. Weight error better than config mismatch.

bspeare:
Is it worth setting the moderation state as a default on alerts so that we have a consistent way of testing the behavior of scheduling in the future?  The main difference is:

**Articles (already published, moderation)**

- If you try to set a "publish on" date, you get the message "_conflicts with the set "published" moderation state. Switching back to "published._"

**Alerts (already published, no moderation)**

- If you try to set a "publish on" date, the alert is unpublished and scheduled for future publication. 

If we aren't concerned about keeping track of these two sets of testing criteria, then I'm happy to approve this.

joewhitsitt:
>Is it worth setting the moderation state as a default on alerts so that we have a consistent way of testing the behavior of scheduling in the future

Not sure what config hiccups that will cause by changing the default feature and getting facilities to accept it based on the recent issues, but that would be a small slice towards https://github.com/uiowa/uiowa/issues/7127

joewhitsitt:
Maybe I read that wrong. As of right now, I don't think we have customer approval to add moderation to facilities alerts so there would still be a separate set of test criteria.

briand44:
Should I reach out to Facilities and see if they would be open to moderation being on for Facilities alerts as long as Construction managers had the ability to publish without approval?

joewhitsitt:
>as long as Construction managers had the ability to publish without approval?

That would be good to know for #7127, but would add some scope to this PR because they would then have publish transition permissions across the site as the permissions are based on the workflow, not the content type. We would have to create a facilities specific workflow for this content type.

Screenshot 2023-12-05 at 11 53 01 AM

joewhitsitt:
I think https://github.com/uiowa/uiowa/pull/7044#issuecomment-1841133843 is resolved for default sitenow_alerts, ITS with https://github.com/uiowa/uiowa/pull/7044/commits/062e0f7e0032487e4c9e58421511c406c3663280. Facilities is still different.

bspeare:
Just tested on default for alerts and I'm getting the "moderation block is broken or missing. You may be missing content or you might need to enable the original module" and I'm also not seeing the moderation option when creating an alert.   Should I be seeing the moderation option for alerts on default?

joewhitsitt:
i think it is good now though but with this config stuff, hopefully it isn't a case of "fixed for me, not for thee"

pyrello:
I'm going to test this.

pyrello:
> Go to /admin/content and try to bulk publish the same article. Again you should see a message saying that the two settings conflict and that the moderation state has been set back to the previous state.

@joewhitsitt is this error expected?

Screenshot 2023-12-07 at 12 38 03 PM

I see it in this comment, but that seems to be implicating a different context: https://github.com/uiowa/uiowa/pull/7044#issuecomment-1790799883

joewhitsitt:
Yeah, it is expected. Not supplied by our code and not very lovely, but it signals that they are trying to do something they shouldn't be. This time it properly sets the moderation state back whereas when you tested it earlier the moderation state was different than the core status which scheduler set back to 0.

pyrello:
While testing facilities on Friday, I encountered this error while running cron:
```
Publishing failed for sdfasasd - asdfasfdda. scheduler_content_moderation_integration_scheduler_publish_process returned a failure code.
```
Going to try resetting up my local instance and running through it again. If I encounter it again, I'll debug.

pyrello:
When @joewhitsitt and I were comparing notes, we figured out that the content moderation widget was still displaying on the  alert content type form on facilities, even though it should not be. When I disabled it, the regular status checkbox stopped showing up. I tried a few things to figure out why, but was not able to figure it out.

I also pushed a few minor changes that remove the content moderation field from displaying on the alert teaser, since we pretty much never want that to display on a teaser. I also re-exported ITS and Facilities after that.

joewhitsitt:
@pyrello can you test facilities again or post the steps to reproduce https://github.com/uiowa/uiowa/pull/7044#issuecomment-1850255460 ?

pyrello:
> @pyrello can you test facilities again or post the steps to reproduce [#7044 (comment)](https://github.com/uiowa/uiowa/pull/7044#issuecomment-1850255460) ?

Seems to work now. Not sure what the deal was before. It looks like you just undid the change that I made. So not sure what actually change between when I was testing before and now.

joewhitsitt:
I think the key was config/sites/facilities.uiowa.edu/config_split.patch.workflows.workflow.editorial.yml. It hasn't always exported correctly. Also, had to maintain the weird fix for alert title on facilities which I broke in a previous update. https://github.com/uiowa/uiowa/pull/7044/commits/c501eda3ddef65f4771840fb47e5b6209fc994ef

pyrello:
So, I'm ready to approve this, but I've been tinkering a little bit to get the last few things to work and it would make me feel better to have someone double check my work. Mostly concerned with alert functionality, not articles.

joewhitsitt:
I can check that here shortly. Also might look at moving the js-hide to the profile admin-overrides.css per https://github.com/uiowa/uiowa/issues/3066

joewhitsitt:
I reviewed the two latest commits by @pyrello and I think they look fine. tested default, its, facilities. mainly wanted to make sure the facilities alert title didn't come back out from hiding. ✅

pyrello (approved):
Approved based on my own testing and the testing of others.

joewhitsitt:
Holy guacamole! Thanks @pyrello and everyone else!

PR #7053, opened 2023-10-30
[ITS] Service Alerts

### LINKED ISSUE #7035: Turn on Alert feature and add additional ITS specific fields
# Story

As an ITS website owner, we would like to be able to create alert content so we can notify students, faculty and staff when there is a problem/change with a service so that they can be aware of any impacts to them.

# Additional Info

Help Desk full time staff and webmasters should have the ability to publish alerts.
Other site editors can create alerts but not publish so moderation is needed.

**Fields**

- Title - generated based on name of service + alert type
    - If additional title details is not empty use additional title details + alert type
    - need clarity on title when multiple services selected and no additional title details?
    - need clarity on if building populates title?
- Description (filtered html)
- Date - require start date, end date optional
- Closed - boolean, need some clarity on how this functions.
- Alert types - Outage, Service Announcement, Maintenance, Degredation of Service (Degredation of Service may have a name change)
- Ability to specify multiple services that are affected
- Ability to specify multiple buildings that are affected
    - Currently they manually maintain a building taxonomy - need clarity on if they continue with that or integrate with Facilities Buildings
- Additional title details

# Proposed Solution

- Enable alert feature
- Add moderation functionality
- Add additional ITS specific fields
- Roles and perms?

cory-skeers:
Resolves #7035 

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /admin/structure/taxonomy/manage/alert_categories/add
```

- Add some Alert Categories, such as `Outage, Service Announcement, Maintenance, Degradation of Service`
- Visit `/node/add/alert` and add an alert
- Try adding different versions of known duration alerts (eg Tuesday from 9am - 10pm) or open-ended alerts (eg started at 9am, assumed to be unresolved at this time), and see if the smartdate form makes sense/is intuitive for these.
- Test perms w/content moderation: editors/publishers/webmasters should be able to create and edit alerts, but only publishers/webmasters should be able to publish

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli /node/add/alert
```
Check for a clean sync and test the form with an eye toward the change to Dates affected, considering Facilities more often than not works with a specified end time for their maintenance alerts, and compare w/the experience on ITS
https://iowaweb.slack.com/archives/C0164FNN1PV/p1698764954582949


cory-skeers:
I tried poking at larger adjustments to the smartdate form, but it became a rabbit hole pretty quickly, as it starts to involve both PHP field elements and JS.

Instead, I took an aim at adjusting the a) duration option to include an "ongoing" label for the zero duration, which in `smartdate` parlance is a single point in time rather than a date range, storing both start and end dates as equivalent, and b) using different defaults for ITS vs Default and Facilities. I think this makes the form slightly more user-friendly for the two different sites' most likely scenarios (planned closures with alerts posted in advance vs emergent posted at the time of the incident and w/out a necessarily defined endpoint).

There is also slight adjustment of the 'dates affected' label in the case of no endpoint, but I think this should be looked at closer as we get into more of the styling and adjustments as additional fields get added, teaser views get looked at, etc.

GaryRidgway:
pulling to review

GaryRidgway (approved):
Everything seems to be working for me, I like the inline more than the default which is nice because it doesn't make the form jump when toggling the all day. Thanks for making the changes Cory!

PR #7071, opened 2023-11-01
Add label for reason for closure

### LINKED ISSUE #7066: Facilities Alert Feedback
They want to add an inline label, similar to the rest of the fields, for the Reasons for Closure field and the email template if it is not there either.

https://facilities.prod.drupal.uiowa.edu/dental-science-building-and-carver-hawkeye-arena-steamcondensate-and-forced-sewer

GaryRidgway:
Tracks https://github.com/uiowa/uiowa/issues/7066

## To test
1.  sync site
2. Observe that an alert now has a label on the "Reason for closure" field.
3. Did I export the right config?

### DIFF
--- a/config/sites/facilities.uiowa.edu/core.entity_view_display.node.alert.default.yml
+++ b/config/sites/facilities.uiowa.edu/core.entity_view_display.node.alert.default.yml
@@ -258,12 +258,15 @@ third_party_settings:
             region: content
             configuration:
               id: 'field_block:node:alert:body'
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Looks good!

I think at least for now, keeping this limited to the Facilities alert specifically sounds good. They made the request, but I think there's still a case for general alerts (and then also ITS alerts) to not show the label for the Body field.

PR #7087, opened 2023-11-02
Hotfixes: Listblock pager offset values and CSHS once regression

### LINKED ISSUE #6899: Listblock offset allows negative values that break the page
# Problem
Page displays generic 'There was a problem' error when clicking view or layout tabs for a page with a negative offset listblock. Locally, error specifies problem in the database with sql syntax.

This came into the support inbox, I was able to fix it for that site by deleting the node tempstore and any revisions with the negative value.

# Steps to reproduce
Add or edit a listblock - I've only tested Articles and People - and give it a negative offset. Add/save block, then close configure block tray (since it won't on its own). Save layout. You will now see the error. Go to /admin/content. Click edit for the page with your listblock. Try the view and layout tabs, they will give the error. 

# Proposed solution
Limit offset values to anything equal or greater than zero.

pyrello:
Resolves https://github.com/uiowa/uiowa/issues/6899

# How to test

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,12 @@ You can run `composer update package/name` to update additional dependencies. Th
 ### Locked Packages
 The packages below are locked at specific SHAs and will not update using the method described above. They should be periodically checked for new stable releases and updated, if viable.
 
-| Package                      | Reason                                                                                                                         |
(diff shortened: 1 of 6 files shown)


joewhitsitt (approved):
I tested the guts of this locally and reviewed the subsequent commits. Looks good. Nice speedy work

PR #7088, opened 2023-11-02
Config fixes for ITS/Facilities Alerts

cory-skeers:
# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev blt ds --site=facilities.uiowa.edu && ddev blt ds --site=default && ddev drush @default.local config-split:activate sitenow_alerts -y
```
Check that all syncs are clean and that default is able to enable the `sitenow_alerts` feature split


GaryRidgway (approved):
Ran through all testing instructions, worked as expected, clean syncs and default can enable `sitenow_alerts`. APPROVED! or else...

PR #7109, opened 2023-11-10
re-export contact form ignored config after several webform updates

joewhitsitt:
While looking into an issue with a customer's webform I realized the contact form is installed by default but then is ignored but there have been several core and webform module updates since we last exported it. Should mostly be ordering, etc.

# To Test

```
ddev blt di --site=default
```

Take a look around the form and see that it still operates similar to https://default.prod.drupal.uiowa.edu/form/contact


joewhitsitt:
of course Travis hates me though...

bspeare (approved):
Clicked around and examined the form and everything looks pretty much the same.

joewhitsitt:
Thanks @bspeare

Issue #1157, opened 2020-03-09
Modal save dismisses underlaying Media modal without finishing task.


joewhitsitt:
## Problem
When adding images to an image gallery content block, there is an issue when adding a link to the global caption where both the link configuration modal and the image upload modal are dismissed without finishing the image upload.

Focal point preview within media library modal is also affected by this.

_Note: Documented as a known issue on sitenow.uiowa.edu with workaround. https://sitenow.uiowa.edu/known-issues#uiowa-issue-1157_

joewhitsitt:
https://www.drupal.org/project/drupal/issues/2984865
https://www.drupal.org/project/drupal/issues/2741877

joewhitsitt:
[untested] Someone came up with a "terrible hack" as a temporary workaround: https://www.drupal.org/project/drupal/issues/2741877#comment-13524964

joewhitsitt:
Patch available for testing: https://www.drupal.org/files/issues/2020-04-20/2741877-48.patch

joewhitsitt:
reading up on this, this might not be an issue with ckeditor5 as it doesn't use dialogs like these.

joewhitsitt:
I witnessed Alan encountering this issue with CKEditor5.

PR #7098, opened 2023-11-07
Upgrade to Drupal 10.1

pyrello:
Resolves #6538 
Resolves #7094

# How to test
- Read through all the instructions before starting.
- Review code changes.

## Config updates
- Review config changes.
```
ddev blt di
```
- Ensure no config status warnings.

#### Feature Splits
- Review feature split config changes and determined what, if any, functionality you would like to review for each split based on changes. Feel free to ping @pyrello if you are unsure.
- To check that all feature splits run without issue:
```
ddev blt uiowa:usplits:feature --skip-export
```
- For any feature splits that you would like to test functionality to ensure no changes (using `areas_of_study` as an example):
```
ddev blt uiowa:usplits:feature --skip-export true --split areas_of_study && ddev drush @default.local uli
```

#### Site Splits
```
ddev blt uiowa:usplits:site --skip-export
```
- Ensure that all sites sync/update without issues.
- Review all site split config changes and determine what, if any, functionality you would like to review for each split based on changes. Feel free to ping @pyrello if you are unsure.
- After the above command runs, each site should be locally synced and able to be checked (using `admissions.uiowa.edu` as an example):
```
ddev drush @admissions.local uli
```

## Allowed formats
- On the [files changed page](https://github.com/uiowa/uiowa/pull/7098/files), search (Cmd-F) for `allowed_formats:` to see instances where text area fields were updated from the contrib `allowed_values` format to the core one. I think spot checking that these are still working is sufficient.

## Block content permissions
- As an editor/webmaster, I can edit the footer contact info block.
- As an editor/webmaster, I am not able to use the Hero, Page title hero, or Spacer separator blocks when creating/updating a page with Layout Builder.
- As an editor/webmaster, I can use all the other blocks when creating/updating a page with Layout Builder.

## Lazy loaded images
- Test that lazy loaded images are still working and that there doesn't appear to be a conflict between the core behavior and the contrib module we still have installed.
- Check that lazy loaded images appropriately get their height and width output per https://www.drupal.org/node/3279032 after removal of the associated patch.

## "Views Block Display skips preBlockBuild() call on ajax rebuild"
- Test list blocks with non-default LB style options and/or hidden fields and filters AND a pager to check that second "page" results match the expected block settings.

## "Support third party settings for components within a section"
- Confirm that unique ID functionality for inline blocks is still working as expected.

## "Node form layout looks squeezed"
- Review changes to node form layout and confirm that no unexpected regressions have been introduced.

[...]

PR #7132, opened 2023-11-20
Switch to CKEditor 5

### LINKED ISSUE #5667: Move to CKEditor 5
# Motivation

CKEditor 4 is being moved to a contrib module as part of Drupal 10.x and CKEditor 5 will be a core stable module starting in Drupal 9.5.x. I believe it has a semi-automated upgrade path. See details here: https://www.drupal.org/node/3308362

If we move to CKEditor 5 prior to moving to D10, then we won't have to temporarily include the CKEditor 4 contrib module. 

```[tasklist]
# Proposed solution
- [ ] https://github.com/uiowa/uiowa/issues/6061
- [x] Test out semi-automatic upgrade path
- [ ] https://github.com/uiowa/uiowa/issues/7123
- [x] Compare functionality and document any possible regressions
- [x] Re-write any CKEditor JS customizations
- [x] Re-write/replace `FilterIframe` class to be compatible or find a new solution for embedding iframes
- [x] Update `responsive_table_filter` module to 2.0.x
- [x] `iframe_title_filter` module still functioning
- [x] `<pre>` tag
- [x] Test oembed
- [x] Reset formatting not quite working as expected
- [x] `<hr class>` issue
- [x] Transition or move any styles (e.g table styles)
- [x] Crazy markup testing, including developer placed embeds/iframes
- [x] confirm that allowed attributes aren't being stripped out of config (span role, iframe sandbox)
- [x] check that the code block languages work as expected, otherwise remove certain language options (html, javascript, php)
- [ ] Dismiss linkit balloon from image gallery media add form
- [x] Update documentation
- [ ] Communication ahead of deployment pointing out things like `hr` `ul` `ol`
```
- Create follow-up task to remove CKEditor 4 contrib module

pyrello:
Resolves https://github.com/uiowa/uiowa/issues/5667

# How to test

```
ddev composer install && ddev blt frontend
```

## `<hr>` update hook

- Note that https://sandbox.uiowa.ddev.site/kitchen-sink had `hr` elements in between the headings and that they have been converted to `div` elements while retaining classes and styling. Classless `hr` elements should still be `hr` elements.

## General WYSIWYG 

https://sandbox.uiowa.ddev.site/kitchen-sink

Cool feature - Checkout partial markdown support (headings, bold, italic, bulleted lists, blockquotes)!

`pre` is now within the code block as plain text.

Most uids_base styling is provided.

Tables now have caption rendering in the editor as figcaption outside the table markup so some extra styles had to be written to sort of account for it and the positioning.

## Linkit

- Webform matcher still works
- Editor advanced link fields still work, render (aria-label, id, target blank, etc.)

### LinkIt Media Library

- Test whether you can add new/existing media items to a linkit dialog and have it work as expected.
- Note that this is an issue: https://github.com/uiowa/uiowa/issues/5667#issuecomment-1866676436

## Testing oEmbed:

For Facebook/Instagram, you need the credentials local: https://iowa.sharepoint.com/sites/drupal-developer-documentation/SitePages/Miscellaneous/Facebook_Instagram%20Graph%20API%20for%20oEmbed/Facebook_Instagram%20Graph%20API%20for%20oEmbed.aspx

Make sure your browser isn't blocking the source.

There shouldn't be issues related to oEmbed creation/update/deletion, but make sure oEmbed items can be selected and placed within the WYSIWYG. Existing pages to check oEmbeds on would be https://sitenow.uiowa.ddev.site/documentation/media/social-media-embed, https://sandbox.uiowa.ddev.site/inline-video-caption-alignment, https://sandbox.uiowa.ddev.site/media-vimeo

Twitter now has a placeholder instead of being empty.

## Serif font

- Toggle serif on sandbox to see the editor use and render the serif font.

## Saving while on source doesn't result in data loss.

Make edits (e.g. `<p>this is my source text</p>`) in the source view and then save. Edits should be retained if not stripped for other reasons.

## Callouts

https://now.uiowa.ddev.site/news/2023/10/reminder-familiarize-yourself-uis-extreme-weather-plan

## Developer placed embeds/iframes

https://github.com/uiowa/uiowa/issues/5667#issuecomment-1864729696

## Fontawesome

- icons render in frontend/backend (wysiwyg, social media menu)
- wysiwyg icons get `role: presentation` attribute.

## title attribute for iframes

- add an iframe without a title attribute and see that on render, the iframe auto gets one.

## allowed iframes

- in filtered_html try to paste a google maps iframe in the source window. save. see it rendered.

# Follow-up

- Publish draft https://sitenow.uiowa.edu/node/161/latest
  - Add note about preformatted text moving on the toolbar.
  - Add note about WCAG guidance for opening links in new windows. 
- Draft communication and schedule release.
  - Communication should include note about `<hr class="*">` to `<div class="*">` conversion, possibility of close lists being joined together, a beta version of our linkit solution.
- Create known issue for https://github.com/uiowa/uiowa/issues/5667#issuecomment-1866676436
- Remove CKEditor 4 contrib module
- Look into documented upstream issues:
  - styles dropdown in off-canvas - https://www.drupal.org/project/drupal/issues/3409561
  - link console errors - https://ckeditor.com/docs/ckeditor5/latest/support/error-codes.html#error-matcher-pattern-deprecated-attributes-class-key
  - stripped empty block elements (br, hr) https://www.drupal.org/project/drupal/issues/3398223
  - UX change in applying styles: https://www.drupal.org/project/drupal/issues/3362451
  - Too fast to save: https://www.drupal.org/project/drupal/issues/3396742


pyrello:
FYI, this is branched from the #7098 to avoid having to deal with annoying merge conflicts once that PR is merged.

joewhitsitt:
>Does this resolve https://github.com/uiowa/uiowa/issues/1157?

Yes. Unfortunately, I will be documenting a separate issue with LinkIt where I can get the balloon window to go away.

joewhitsitt:
I think as folks twiddle thumbs, this is ready for more testing. Ultimately I'd say we are waiting to see progress on https://github.com/uiowa/uiowa/issues/5667#issuecomment-1866915708 before we merge in case we need to update core to 10.2 for any reason. It sounds like we might be partially protected with this managed by config. Already had to restore some things that were lost during the update: https://github.com/uiowa/uiowa/pull/7132/commits/d4199fd5ecc945ffd67b0006e14dd83473641d25

bspeare:
I'm going to take some time this afternoon and test this out.  I'll add observations and screenshots here in a details element. 

- [x] Image size and image caption is not rendering as expected on https://sandbox.uiowa.ddev.site/node/1986/layout. 

<details>
Screenshot 2023-12-22 at 2 17 31 PM
Screenshot 2023-12-22 at 2 17 15 PM

</details>

- [x] Tables not displaying as expected on https://sandbox.uiowa.ddev.site/node/706/layout and https://admissions.uiowa.ddev.site/node/66/layout.

<details>
Screenshot 2023-12-22 at 2 33 26 PM

</details>

- [x] List styling is very indented compared to prod.  View accordion on https://sandbox.uiowa.ddev.site/node/706/layout. 

<details>
Screenshot 2023-12-22 at 2 35 15 PM

</details>

- [ ]  Accordion text area is majorly overflowing on https://sandbox.uiowa.ddev.site/node/706/layout and https://admissions.uiowa.ddev.site/node/66/layout

<details>
Screenshot 2023-12-22 at 2 36 51 PM

</details>

- [x] Text area tools are getting `text-transform: uppercase` compared to the body WYSIWYG:

<details>
Screenshot 2023-12-22 at 2 53 51 PM
Screenshot 2023-12-22 at 2 51 59 PM
</details>

- [x] Need the `reversed` attribute https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol added. 
- [ ] If two or more `<ul>` or `<ol>` lists are directly next to each other, they will be joined together.  See https://sandbox.prod.drupal.uiowa.edu/ol-li-test as an example. 
- [ ] New table widget will need some documentation to show how to set a header row, caption
- [x] Table caption renders at bottom of the table

bspeare:
It looks like font awesome icons are no longer rendering as svg, @joewhitsitt did you notice that when you were testing the icons?

joewhitsitt:
@bspeare after changing the config, I only looked to see that the icon was rendered and in my spot checking that seemed to be true. https://github.com/uiowa/uiowa/issues/5667#issuecomment-1863473788

i can pull down in a bit to poke more.

bspeare:
It looks like https://www.drupal.org/project/drupal/issues/3332416 is going to be an issue with our paragraphs.   At certain viewport widths, v2 sites users will not be able to edit their content.

joewhitsitt:
Yes, I can confirm that it no longer renders using the SVG method. With the setting checked though, we are getting unresponsive WYSIWYG areas. Found this upstream issue https://www.drupal.org/project/fontawesome/issues/3355253 which then points to an issue that says using svg and pseudo elements are not recommended.

So if we want to continue using SVG, we will have to update uids/uids_base to use the @font-face method? or replace all of the css pseudo with replacement markup. https://fontawesome.com/docs/web/setup/upgrade/pseudo-elements

bspeare:
Thanks for looking into it @joewhitsitt, I'm fine with ditching the svgs but we will need to adjust our uids_base theme styles so that we can get a passing Percy build.   I can start working on those adjustments.

joewhitsitt:
I don't recall the original rationale for the svg/js version, but if that is the new way of doing things, are we just kicking the can down the road then by not using it?

joewhitsitt:
I tried recording the page performance with load assets enabled for fontawesome. all.min.js seemed to continually looping based on frame changes to the point it became unresponsive. when i resaved with `allow_pseudo_elements: true` it seemed to load fine but of course now all of our pseudo icon implementations (superfish, etc.) had a duplicate icon unstyled and out of place. it seems the DOM gets processed and isn't looped on again because it can't find any more unconverted instances. it seems that despite allow_pseudo_elements being false initially, it is still triggering that search stuff.

https://fontawesome.com/docs/web/add-icons/pseudo-elements#css-pseudo-elements-with-our-svg-js-framework makes it seem like we can't have the best of both worlds and that we either need to go full SVG (pseudo elements replaced with svgs). Additionally, unless we are going to use the WYSIWYG plugin, we could be implementing fontawesome ourselves instead of using this module.

joewhitsitt:
Getting back _into_ (hehe) this and I was wondering what you ran into @bspeare that changed the library load from admin route to isloggedin

bspeare:
@joewhitsitt the admin route wasn't working for me and the library was still loading on all front-end logged in pages, but maybe it was local gremlins?  My understanding is that we only want the `all.js` to load for anonymous users/site to keep the better accessibility with `svg` icons.    While logged in, all icons render as `<span>` to avoid the ckeditor from crashing.   Feel free to adjust if you can get that working!  I still need to file the bug report over at `drupal/fontawesome`.

joewhitsitt:
Ah I think the issue is that `$admin_context->isAdminRoute()` is false on node/123/layout

Wondering how we could scope this more because we do have things like the viewer role that is authenticated but doesn't have formatted text abilities.

bspeare:
@joewhitsitt yes, that was the issue!  It would be great if we could figure out scoping it further.

joewhitsitt:
I think it works with aggregation on

joewhitsitt:
>Accordion text area is majorly overflowing on https://sandbox.uiowa.ddev.site/node/706/layout and https://admissions.uiowa.ddev.site/node/66/layout

@bspeare this was solved with the patch right?

bspeare:
@joewhitsitt the CSS that I added does fix it, but I'm wondering if the JavaScript in https://git.drupalcode.org/project/drupal/-/merge_requests/5728/diffs would be better?

joewhitsitt:
Oh, I thought you had added the patch. My bad. I can try it and compare to what you have in place.

joewhitsitt:
I think https://github.com/uiowa/uiowa/pull/7132/commits/17d8787e6d412f56e8cc0e64d244543b23312746 is better for now. This is what the patch currently is like:

https://github.com/uiowa/uiowa/assets/4663676/fd9c3203-43e1-47fd-b72c-346a86ad57af

joewhitsitt:
>Update documentation

>New table widget will need some documentation to show how to set a header row, caption

Took a run at updating the WYSIWYG documentation: https://sitenow.uiowa.edu/node/161/latest

joewhitsitt:
Famous last words, but I _think_ this is ready for review again. Definitely not perfect. Still testing to see what can be done to improve the linkit dialog in those very specific modal/modal situations.

GaryRidgway:
~~I am seeing another issue here which involves a different z index for the context balloon when it is in the off canvas tray. I tried resetting the z index to what it is when not in the off-canvas tray, and that seems to work just fine, but I am wondering If I should be adding those type of styles in this issue as well?~~

My mistake, I forgot we already added styles.

joewhitsitt:
Looks like Alan encountered https://github.com/uiowa/uiowa/issues/1157 again in CKEditor5. I have removed that from being resolved with this PR.

bspeare:
Pulling to test again.

bspeare:
Noticing this code formatting option on `filtered_html` that I didn't notice before - do we want this?
Screenshot 2024-01-16 at 10 30 26 AM

bspeare:
Looks like we are losing the guidance on opening links in new windows:

Screenshot 2024-01-16 at 10 40 22 AM
Screenshot 2024-01-16 at 10 41 02 AM

joewhitsitt:
> Noticing this code formatting option on `filtered_html` that I didn't notice before - do we want this?

[We need it to keep preformatted text](https://github.com/uiowa/uiowa/issues/5667#issuecomment-1858193939). The other languages don't [hurt](https://github.com/uiowa/uiowa/issues/5667#issuecomment-1867030802).

joewhitsitt:
> Looks like we are losing the guidance on opening links in new windows.

I doubt people actually click that in the moment. Would be a CK plugin to alter now. I suggest we add that to documentation/training.

bspeare:
I can't get the link dialog to work while adding a link within a media item's global caption.  Has this already been discussed?

 Screenshot 2024-01-16 at 10 57 33 AM

bspeare:
If I hit the "Source" button, I can see it in the upper left corner. 

Screenshot 2024-01-16 at 11 08 18 AM

joewhitsitt:
>I can't get the link dialog to work while adding a link within a media item's global caption. Has this already been discussed?

Yes, the recent work solved similar issues with single layer layout builder, but [this modal on top modal issue](https://github.com/uiowa/uiowa/issues/5667#issuecomment-1866676436) still occurs. We were going to document it as a known-issue. Even if this was fixed, the https://github.com/uiowa/uiowa/issues/1157 issue still causes issues.

bspeare:
In a text area block, If table captions have formatting, they will not render white:

https://admissions.uiowa.ddev.site/node/141 - Has a `<strong>` tag on table header. 


Screenshot 2024-01-16 at 11 16 31 AM

bspeare:
@joewhitsitt I also can't get the link dialog working on a webform text area - is this the same issue?

joewhitsitt:
>link dialog working on a webform text area

interesting. haven't tested that.

bspeare:
I'm done testing for today.   I think I fixed the table caption issue.  What do folks think about the next step for getting linkit to work with webform?  "Disable off-canvas system tray" fixes the issue, but we lose the off canvas dialog. 
Screenshot 2024-01-16 at 3 45 29 PM

joewhitsitt:
Planning on taking a look at the link dialog/webform issue tomorrow morning if Alan can't.

Is this the only open item? Thanks for making that table css change and increasing scope on the js-hide @bspeare .

bspeare:
Yes, @joewhitsitt I think the webform is hopefully the last issue.

joewhitsitt:
Screenshot 2024-01-18 at 9 16 04 AM

Success! Using the latest version of the MR and upping the z-index from 999 to 1260...

`--ck-z-modal: calc(var(--ck-z-default) + 1260);`

I can look at adjusting this upstream a bit later but I am not currently setup to do so.

Update: Alan raised a good point about why it was 999 and not more (linkit balloon on top media library modal on top of off-canvas). Alan then pointed out the inherit option which worked. Adjusted the MR upstream to use that. Updated the PR with the latest patch

bspeare (approved):
Webform changes look good.  I even tested this in Browserstack and it appears that the link dialog works better for mobile than the current editor.

joewhitsitt:
Wow! Thanks for going the extra mile with the browser stack testing @bspeare

joewhitsitt:
Going to switch to draft until we can settle on communication/scheduling

PR #7137, opened 2023-11-28
[POC] SiteNow Migrate optional updates

cory-skeers:
# How to test


joewhitsitt:
This has been hanging out in PRsAhoy for a while. If it isn't too much of a lift, I say we work on merging in these improvements even if we don't have an immediate use-case.

cory-skeers:
> This has been hanging out in PRsAhoy for a while. If it isn't too much of a lift, I say we work on merging in these improvements even if we don't have an immediate use-case.

Re-looking at this, in addition to the definition-setting it has a few updates/fixes to the Page and Article migrations that'd be good to get in prior to CCOM's possible interest in a Page migration. And it's pretty lightweight in what it touches; nothing outside of migration-related stuff, and even that it's limited to the pre-defined migrations and the migration settings form.

joewhitsitt:
pulling to test

joewhitsitt (approved):
well, it took some wrangling to test this and i didn't necessarily pick the best site to use but I tried ukchoralcomposers with their non-standard content types `work` `text` etc and was able to use the d7_page, d7_article to migrate them based on the mapping added to this PR. I think the other changes are straight forward and are inline with our current practices `create_media_from_file_field` etc. I don't know if I ran into the unmapped field logic though.

Screenshot 2024-10-31 at 4 38 37 PM

PR #7148, opened 2023-11-29
Update drush info for uiowa07

joewhitsitt:
# To Test

- git pull
- `ddev drush @uiowa07.test ssh`

Assuming you have permissions (you should), you should be taken in to `uiowa07.stage@sshd-685758d998-q2xrz` or something weird like that. If you were to try this on `main` currently, you would not be able to get in compared to `@uiowa07.dev`


bspeare (approved):
Worked for me!

PR #7149, opened 2023-11-29
Update drush info for uiowa07 part deux

joewhitsitt:
Essentially the same as https://github.com/uiowa/uiowa/pull/7148 but for PROD

# To Test
git pull
`ddev drush @uiowa07.prod ssh`
Assuming you have permissions (you should), you should be taken in to uiowa07.prod@sshd-56db667b45-th9ql or something weird like that. If you were to try this on main currently, you would not be able to get in compared to @uiowa07.dev


bspeare (approved):
It worked!

PR #7153, opened 2023-11-30
Initialize sandboxnext.sites.uiowa.edu multisite on uiowa07

joewhitsitt:
It looks like the database creation was successful on all three environments.


cory-skeers (approved):
🚀

PR #7177, opened 2023-12-05
Aggregator Extended

### LINKED ISSUE #5615: Purge aggregator items no longer in feed source.
Currently we are retaining aggregator items for 1 year. Would it make more sense for our use cases to only retain items that exist in the feed source?

This would allow folks to have archived feeds like https://apps.its.uiowa.edu/dispatch/archives/past/--iowa-law---september-25-lxj144j without requiring us to keep items for 1 year.

It could also help prevent issues with high volume feeds like we have seen on careers.uiowa.edu.

It sounds like there is conversation around this in the issue queue but is not currently an option for the aggregator module.

### LINKED ISSUE #7068: Probe aggregator to remove items no longer in source
# Story

As a web team, we would like to probe removing items that are no longer in the source feed as part of the existing aggregator functionality prior to building a new component so that users can easily display staff and faculty jobs on their website.

# Additional Info

Related issues:
https://github.com/uiowa/uiowa/issues/6631
https://github.com/uiowa/uiowa/issues/5615

https://www.drupal.org/project/aggregator/issues/1294472

# Proposed solution

- Determine if we can/should enhance the existing aggregator functionality to allow removing items that are no longer in the source feed or if we should build a new component.

joewhitsitt:
Resolves #7068
Resolves #5615

# To Test

```
ddev composer install && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli aggregator/sources/31/configure
```

See the purge items checkbox and check it. Change the interval to 15 min. Save the source.
Go to https://sandbox.uiowa.ddev.site/admin/config/services/aggregator and note the number of items in the campuslabs table. Notice the number in a few other tables (e.g. HR liveWell, USG events, etc.)

`ddev drush @sandbox.local cr && ddev drush @sandbox.local cron` _(ignore the htaccess error)_

Note the dramatic reduction in campuslabs items.

Manually "refresh items" should also work.

To test the upstream feed better, you can use https://sandbox.prod.drupal.uiowa.edu and add/remove article content. Keep in mind the current 1 year limit. Even though it might be in the feed, it is no longer imported given how old it is. (Sandbox only had 4 items when I tested)

# Deep Dive

With xdebug on, and breakpoints set at docroot/modules/contrib/aggregator/src/ItemsImporter.phpL105 and docroot/modules/custom/uiowa_aggregator/src/ItemsImporterOverride.phpL45 you will notice that on cron run or the update items operation it steps into ItemsImporterOverride refresh first and then proceeds to do the rest of the refresh method in ItemsImporter meaning there is no functionality lost by hijacking with the service decorator.


joewhitsitt:
Based on https://github.com/uiowa/uiowa/issues/7068#issuecomment-1841298848, I think I will work on removing purge items being added programmatically and instead add it through the field ui.

pyrello:
Pulling to test.

bspeare:
Tested and it works!
1. Synced site
2. Set purge checkbox and update interval to 15 minutes
3. Ran `dev drush @sandbox.local cr && ddev drush @sandbox.local cron`
4. 3262 was reduced to 66

Also tested on careers and it worked there as well.

pyrello (changes requested):
Getting an error on this page after following test instructions: https://sandbox.uiowa.ddev.site/aggregator/sources

```
Error: Only arrays and Traversables can be unpacked in uiowa_aggregator_aggregator_item_view() (line 74 of modules/custom/uiowa_aggregator/uiowa_aggregator.module).
```

joewhitsitt:
My test instructions have not been updated with a `ddev composer install`. Could that be the error you are receiving?

joewhitsitt:
Hmm. This must be related to the update or recent core updates:

Screenshot 2023-12-07 at 10 59 49 AM

joewhitsitt:
Currently errors on PROD too: https://sandbox.prod.drupal.uiowa.edu/aggregator/sources

joewhitsitt:
It was discussed in planning that the bug https://github.com/uiowa/uiowa/pull/7177#pullrequestreview-1770608022 is not related to this PR and something that should not increase the scope of this PR. I will create a separate issue for it.

pyrello (approved):
@joewhitsitt created an issue for #7190, which we believe is not related to this PR. Works as expected.

joewhitsitt:
As suggested, I posted a comment on the upstream issue so that others can possibly benefit: https://www.drupal.org/project/aggregator/issues/1294472#comment-15350316

Also documented the new field/functionality here: https://sitenow.uiowa.edu/documentation/layout-builder/rss-feeds

PR #7179, opened 2023-12-05
[Admisisons] Extended student profile years to 2030

### LINKED ISSUE #7130: Admissions student profile select list updates
> We need to update the dates listed in the “Anticipated graduation date” drop-down in the Admissions student profile content type.
> Remove: Spring 21, Fall 21, Spring 22, Fall 22, Spring 23, Fall 23.
> Add: Spring 27, Fall 27, Spring 28, Fall 28, Spring 29, Fall 29, Spring 30, Fall 30.

I think we painted ourselves into a corner on this one by not using the in-code allowed options function. Not sure we can add/remove options currently. config/sites/admissions.uiowa.edu/field.storage.node.field_student_profile_grad_date.yml - @joewhitsitt 

# Proposed Solution

- [ ] Add: Spring 2027, Fall 2027, Spring 2028, Fall 2028, Spring 2029, Fall 2029, Spring 2030, Fall 2030 through config

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7130. 

```
ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli /node/add/student_profile
```
Confirm that new dates are listed to select on node/add and in the code.

### DIFF
--- a/config/sites/admissions.uiowa.edu/field.storage.node.field_student_profile_grad_date.yml
+++ b/config/sites/admissions.uiowa.edu/field.storage.node.field_student_profile_grad_date.yml
@@ -12,65 +12,65 @@ type: list_string
 settings:
   allowed_values:
     -
-      value: 'Spring 2021'
-      label: 'Spring 2021'
-    -
-      value: 'Fall 2021'
-      label: 'Fall 2021'
-    -
-      value: 'Spring 2022'
-      label: 'Spring 2022'
-    -
-      value: 'Fall 2022'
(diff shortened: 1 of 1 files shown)


pyrello:
Only thought here is that it might be more useful to reverse the order of the years so that the latest years are listed first. Also possibly set the default to something that they think is likely to be useful.

pyrello (approved):
Works as expected, no import errors.

PR #7185, opened 2023-12-06
[Bug] Fix for responsive image width/height values are not used from fallback image style

### LINKED ISSUE #3614: Image editing functionality in CKEditor fix

### LINKED ISSUE #6263: Responsive images bug
# Problem
As @bewoldt noted in https://github.com/uiowa/uiowa/issues/3614, "The "Small - Square" and "Large - Square" images options end up being the same size."

# Proposed solution

- Verify that lazy loading is currently working.
- Investigate removing width and height patch from `composer.json`, set CSS aspect ratio for image styles like `.widsecreen img {aspect-ratio: 16/9;} ` to avoid cumulative layout shift.
- Investigate switching to using `<picture>` element so width/height work with these patches: https://www.drupal.org/project/drupal/issues/3192234#comment-14787534. 
- Test out either solution across a range of SiteNow sites and figure out if we need so set any update hooks 
    - For instance, if someone is using a `small- square crop` in a layout section set "edge-to-edge" and has "remove column spacing" selected.

### LINKED ISSUE #7184: Responsive images bug
# Problem
The 10.1 update included a responsive images width and height attribute fix (https://www.drupal.org/project/drupal/issues/3192234#comment-14787534) that now properly sets width and height properties for the "Small - no crop" image style".  Each "no crop" image style was previously not getting these attributes set. 

Before:

Screenshot 2023-12-05 at 3 00 28 PM

After:

Screenshot 2023-12-05 at 3 00 19 PM

# Steps to reproduce
Remove image attributes on https://whitmanweb.iwp.uiowa.edu/song-myself/turkce-turkish/bolum-17-kendimin-sarkisi as an example. 

# Proposed solution
- [x] Apply patch from https://www.drupal.org/project/drupal/issues/3377420, update fallback small image styles.
- [x] Develop SiteNow article announcing the change using bits from https://apps.its.uiowa.edu/dispatch/archive/an-update-to-inline-element-sizing-Q76x0l0. 
  - [x] Show before and after examples
  - [x] Outline image sizes in SiteNow documentation that we can link to at https://sitenow.uiowa.edu/documentation/media/images-and-files
  - [x] Possibly explain difference between Drupal core image styles and SiteNow image styles
  - [x] Suggest looking at GA for high value pages to inspect
  - [x] Suggest using Find Text module for `data-view-mode="small__no_crop"` and `data-view-mode="large__no_crop"` searches

```[tasklist]
### Remaining tasks
- [ ] Create Dispatch campagin from https://github.com/uiowa/uiowa/issues/7184#issuecomment-1847320410
- [ ] Publish https://sitenow.uiowa.edu/node/326/latest
- [ ] Publish https://sitenow.uiowa.edu/news/2023/12/important-update-sitenow-image-display-styles
```

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7184
Resolves https://github.com/uiowa/uiowa/issues/3614
Resolves https://github.com/uiowa/uiowa/issues/6263

```
ddev composer install && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /small-responsive-images
```

Confirm that the local version is actually loading the pixels sizes listed below on https://sandbox.prod.drupal.uiowa.edu/small-and-medium-responsive-images and https://sandbox.prod.drupal.uiowa.edu/large-and-full-responsive-images
  - Small 384 px
  - Medium 768 px
  - Large 1024 px
  - Full 1312 px

# Also

- [ ] Review any other sites for image regressions


joewhitsitt:
Did you see bkosborne's comments on that issue? Should we be setting up responsive images differently?

https://www.drupal.org/project/drupal/issues/3377420

bspeare:
He seems to be suggesting that the order is set by "The list is sorted alphanumerically by machine name of the image style and processed in that order." and that a fix could come from updating how that list is sorted so show the smallest image style last. 

If you look at https://web.dev/patterns/web-vitals-patterns/images/responsive-images/, their example says **For responsive images that use width descriptors**:

- Set the width and height attributes to match the dimensions of the fallback image.

Which this patch seems to be doing. So it seems like the best practice here is to take the attributes from the fallback image setting.  Is that your take?

joewhitsitt:
Thanks for deep dive @bspeare. The opinion bkosborne and the help text have of the fallback image only being of use when there isn't browser support doesn't seem to match that web.dev link you shared. Still scratching my head how Max 1300x1300, Max 2600x2600, Max 325x325, Max 650x650 is ordered in that list? Am I having a brain fart or should Max 325x325 follow Max 650x650 in that list?

There was a reference to srcset vs picture implementation. Didn't know where that intersects.

GaryRidgway:
To clarify Joe, I think the order would be looked at as `Max 1*`, `Max 2*`, `Max 3*`, `Max 6*`. So I think that is the correct AlphaNumeric order.

joewhitsitt:
Thanks Alan! It wasn't connecting with me. Perhaps I should call it for the day?

GaryRidgway:
No harm in getting some extra rest!

joewhitsitt:
Pulling down for sanity check before this is merged/deployed based on the comms that have already gone out. 😅

joewhitsitt (approved):
Confirmed that this does indeed work as Ben has previously demonstrated. Let's rip the bandaid off.

Should we switch to draft until Thurs merge?

joewhitsitt:
trying to fix the conflict since i caused it

PR #7189, opened 2023-12-07
Search toggle Hotfix

GaryRidgway:
I think I didn't scope the styles for the search button properly.

## To test
`ddev blt frontend` Test with all variations of the menu, toggle, horizontal, mega, with and without top links.


joewhitsitt (approved):
On International I tried all of the header/nav/sticky combos and then did it again after removing top links.

In all cases I was able to toggle the search button with my mouse.

Screenshot 2023-12-07 at 12 09 50 PM

Update: adding that it seems to resolve the writinguniversity issue as well:

Screenshot 2023-12-07 at 12 57 30 PM

PR #7193, opened 2023-12-08
[ITS] Alert fields round 2

### LINKED ISSUE #7058: ITS specific fields part 2
Follow up to https://github.com/uiowa/uiowa/issues/7035.

- Building field 
    - manually controlled taxonomy 
        - term name
        - description
        - abbreviation? - leftover from old Facilities Sims integration, probably not needed anymore.
        - pull these over as they are? - csv export/import? taxonomy migration?
        - rabbit hole
- Title logic - sounds like we can use same logic as they currently have
  - The current mockup adjusts the title construction a bit to put the type first and to consolidate multiple services/buildings when possible: `[alert type] - Multiple Services, Buildings - [additional details]`, `[alert type] - [service], [building] - [additional details]`, `[alert type] - Multiple Services, [building] - [additional details]`, `[alert type] - [service], Multiple Buildings - [additional details]` See https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7 for a visual. Announcements don't have the alert type prepended. One could argue that these should be articles that are included in the alert digest.
- Closed field - [Background](https://github.com/uiowa/uiowa/issues/5952#issuecomment-1802511979)
  - End dates greater than start date that passes, should 'close' an alert (e.g. planned maintenance)
  - Manually closing an alert, should set the end date automagically, but this time around, the date should be editable for back dating/correction purposes.
  - Not to be confused with "unpublished" or "archived" as node still needs to be publicly viewable.

We should make sure building is output on the alert node, not linked
Can we output last updated date on the alert node?

https://github.com/uiowa/uiowa/issues/7058#issuecomment-1802708717

# Proposed Solution

```[tasklist]
### Tasks
- [x] Add building vocabulary
- [x] Building export/import
- [x] Rabbithole building terms
- [x] Add ER field to alert node for building
- [x] Add building name field to alert full node output (LB) displaying as the name, not linked
- [x] Add logic to presave to handle saving title according to the logic above
- [x] Output last updated date on alert node
- [x] Add an `isClosed` method to the Alert entity bundle class. The method returns true if the end date is different than the start date and has passed.
- [x] Make Alert Category field single-value and required in the ITS split
```

cory-skeers:
Resolves #7058 

# How to test

```
git pull && ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /node/add/alert
```

Check that the update hook completes successfully

Title should be hidden, and updated on save based on alert settings.
New "Additional title details" field that affects the alert's title
Category field should be required and allow only 1 value to be selected
Services and Buildings affected fields are select lists allowing multiple values, and the buildings should be prepopulated (there is only one service in prod, but adding more services should allow for more options here as well)

Upon save, the alert title should be set automatically.
Try various combos of including/not including building(s), service(s), and additional details, and see that an accurate title matches each

Test the `isClosed()` method on the Alert Bundle class
`xdebug` works, or you can do something like updating Lines 61+ in `its_core.module.php` with
```
      $output = implode(' - ', array_filter([
        ($entity->isClosed()) ? 'Closed' : 'Open',
        $category,
        $services_buildings,
        $additional_details,
      ]));
```
^ Then create or edit alerts and change the "dates affected" field and see the results in the updated title.
Essentially, if an alert has a definitive end time and that time has passed, then it should show up as "closed," but if it is ongoing, or if the end hasn't passed (whether or not the beginning has passed), it should be considered unclosed.


bspeare:
**All testing steps seem to work well:**
- Update hook imports buildings and they are rh'ed
- Title is hidden
- New "Additional title details" will appear in the title
- Category is required and is only value needed
- Multiple services and building location values work and appear if more than one is selected
- Title set automatically on save
- Various combos all seemed to work well

**Close functionality:**
I'm not completely sure how to test this.  @cory-skeers could you demo the close function for me tomorrow?

cory-skeers:
Setting back to draft for a few updates

cory-skeers:
@bspeare I updated two parts, if you wouldn't mind double checking my latest commits.

1) The "last updated" should appear on the node (based on the `changed` field with an altered label), which I'd missed from the original issue
2) The `isClosed` is declared in the `ClosableInterface`. This change doesn't affect anything in this work and so probably doesn't need specific testing, but will allow us to check that the entity has the `isClosed` method as defined prior to calling it in future work.

bspeare (approved):
Tested the additional fields and everything is working as described.  Nice work!

PR #7207, opened 2023-12-13
Allow the sitenow request form login to pass query string parameters

pyrello:
Adjusts the logic that adds a login link to the SiteNow request form to allow and preserve variables that are passed to the page.

# How to test

```
ddev blt ds --site sitenow.uiowa.edu
```
- Visit https://sitenow.uiowa.ddev.site/request-site?domain=test.uiowa.edu&requestor=sadamshiett
- Confirm that when you hover over the login link, you see the query string parameters are transferred to it
- Confirm that adding, removing, and changing the parameters adjusts the login link
- Please note that I was not able to figure out how to avoid overwriting the `destination` param with one if it's provided.


joewhitsitt:
Won't this cause the whole page to cache expire?

pyrello:
> Won't this cause the whole page to cache expire?

I'm not 100% on how the caching bubbles up, but I think it is specific to the block. We have similar logic in the footer of every page, so I wouldn't expect this to be an issue.

joewhitsitt:
Going to take a look at this with a focus on the caching bit.

joewhitsitt:
Going to send this out to DEV? So far what I am seeing is that a unique URL is uncached and if I just hit refresh, it is cached. Problem is that the login link constructs like this `https://sitenow.uiowa.ddev.site/saml/login?domain=test.uiowa.edu&requestor=jwhitsit&destination=/node/26` which makes me think that it won't be preserved after redirect from the SAML stuff.

joewhitsitt:
From my testing, caching doesn't seem to be an issue (getting cache hits for varnish/drupal), but the destination is not properly constructed. I get redirected to https://sitenow.dev.drupal.uiowa.edu/request-site?check_logged_in=1

Also, already logged in and passing the URL with parameters to the browser, it does not fill my details (domain, hawkid). We already have code for this form to fill in the current user.

joewhitsitt:
Went ahead and updated the form to prepopulate:

Screenshot 2023-12-13 at 1 37 32 PM

~~Going to resync the prod db down to dev~~ deployment is happening. going to just duplicate the setting in dev.

joewhitsitt:
This works now `request_type=Existing&uri=test.uiowa.edu&hawkid=sadamshiett` but again, the hawkid is probably not needed.

joewhitsitt:
Wonder if `urlencode` could be used to append to the $path in the destination.

joewhitsitt:
Looks like that worked. Result after SAML auth, fills the form:

`https://sitenow.dev.drupal.uiowa.edu/request-site?check_logged_in=1&request_type=Existing&uri=test.uiowa.edu`

joewhitsitt:
If there are caching issues I didn't notice while testing in DEV, my mind is a bit eased now knowing that this block is not the footer login block that is on every unauthenticated page.

docroot/sites/sitenow.uiowa.edu/modules/sitenow_core/src/UserLoginBlockPreRender.php
```diff
@@ -37,8 +37,13 @@ public static function preRender(array $build) {
       /** @var \Drupal\Core\Url $url */
       $url = $build['content']['hawkid']['link']['#url'];
 
+      $qs = \Drupal::request()->getQueryString();
       unset($build['content']['hawkid']['link']);
```
joewhitsitt: Can this be reordered to not break up the `$qs` logic?

joewhitsitt (approved):

PR #7214, opened 2023-12-14
Provisions Sprint 192

e-marie-w:
Provisioned iip.lab.uiowa.edu on uiowa03


joewhitsitt:
sorry if i missed discussion around this. don't we try to avoid the year in domains so that it can be used multiple years?

joewhitsitt (approved):
"Let’s solve for tomorrow, not yesterday" - Michael

Issue #7101, opened 2023-11-07
Move SiteNow Articles/People views back to tracked config with prebuild options


joewhitsitt:
From https://github.com/uiowa/uiowa/issues/6832, we started to consider a future where we moved away from the brittle functionality behind SiteNow Articles and People view pages but thought it was too big of a slice at the time.

The current functionality is re-writing ignored views config and is potentially growing out of date with views core defaults.

# Proposed solution

- [ ] Move SiteNow People/Articles view pages back to tracked config with prebuild functionality to alter the view based on options (view header, view title, filters, sorts, etc.) set by the configuration page. Any settings would extend to any additional view displays.
- [ ] Update hook to start saving the config to fields on the settings form and move away from writing directly to the views config.
- [ ] Revisit some one-off custom views that could be moved back as a separate view display of the main view and then just partially split.Then instead of a separate view for Pharmacy, adding a new display to the tracked default view and partially splitting that. Any settings set would extend to this partially split display.

# Related Issues

- https://github.com/uiowa/uiowa/issues/3151
- https://github.com/uiowa/uiowa/issues/3804
- https://github.com/uiowa/uiowa/issues/7030

PR #7215, opened 2023-12-14
Update SiteNow Articles Views config

joewhitsitt:
# ToDo

- [ ] Write update hook to modify config for existing sites as this config is ignored.
- [x] Re-export view config with body/summary as feed description value instead of field_teaser.


joewhitsitt:
Based on this git diff, might be easier to do https://github.com/uiowa/uiowa/issues/7101 instead

Issue #7217, opened 2023-12-15
Service Announcement component on alert landing page


briand44:
# Story

As an ITS website owner, I would like to be able to display active service announcements on our alerts landing page so that users can easily find information about these important announcements and how it affects them.

# Additional Info

Mockup - https://sandbox.prod.drupal.uiowa.edu/it-service-alerts-copy

View block that shows published and active (not closed) service announcement alerts.

Order by post date

Theme according to mockup

No results text

# Question

Do we want to continue pushing on these being articles vs alerts?

# Proposed solution

briand44:
Grooming conversation.... ITS wants to keep these as alerts. Can use cache tag from previous work if needed, may want to rename to be more generic. Doesn't need to match the mockup if we run into any hiccups. Can make it look more like an alert if we want since it is an alert.

pyrello:
I'm interested in working on this. I feel like I missed the whole conversation about this being alerts vs. articles. Looking at the mockup and the type of information being presented here, it seems obvious that these should be articles. Can someone provide some context as to why the customer wants these to be alerts?

briand44:
@pyrello This was discussed in the last meeting with them. I didn't take great notes but they mentioned they see service announcements separate from a news article and that these service announcements need to be displayed on the alert calendar, in alert emails, on the alert status history page, etc. so they would prefer to keep these as alerts like they are now.

I believe we intentionally mocked them up to be more like an article because that was what we were suggesting at that time. We mentioned while grooming this that we may need to make these look more like alerts rather than how they are currently mocked up since they didn't want to use articles here.

GaryRidgway:
In the mockups, we don't have the badges for service announcements on the teaser view for them. Additionally, the teaser view doesn't have the arrow circle either. Should I be making a new display, working with some sort of pre-process, or leaving them how they are for consistency?

<details><summary>This is an example of the current treatment</summary>
<img src="https://github.com/uiowa/uiowa/assets/23122142/5e62233e-1a39-4e85-bd0a-5c54712c6152">
</details> 

<details><summary>And this is how the original mockup looked</summary>
<img src="https://github.com/uiowa/uiowa/assets/23122142/eeb55cbe-be8a-4d4d-ac47-6a8ff55d93a3">
</details>

joewhitsitt:
[...]

Issue #7122, opened 2023-11-15
Follow Brand editorial style guide for outputting months and time


briand44:
# Story

As a web team, we want to follow the Brand editorial style guide.

# Additional Info

It sounds like we should be abbreviating names of months when accompanied by day number

> Months
> In text, write out names of months not accompanied by a day number.
> The family goes camping each year in October.
> The building will be completed in November 2002.
> In text, abbreviate names of months accompanied by day number; some months (March, April, May, June,
> July) are not abbreviated. [AP]
> Jan. 7 Feb. 28 March 21 April 24
> May 9 June 6 July 4 Aug. 31
> Sept. 17 Oct. 25 Nov. 11 Dec. 31
> Other abbreviation systems for month names may be used in certain formats (e.g., calendars, display type,
> lists). For exceptional cases, CMS 15.42 provides three abbreviation systems for month names; choose one
> system and use it consistently (for calendars, display type, lists, etc.) throughout a document or
> publication.

It doesn't seem like we are doing this on articles.
https://tippie.uiowa.edu/news/2023/11/seat-pentagon

a.m. and p.m. was also brought up but no examples were shared.

Here is the excerpt from the editorial style guide.

> Time: a.m. and p.m.
> The abbreviations “a.m.” and “p.m.” are lowercase with periods and no space between letters. There should
> be a single space between “a.m.” or “p.m.” and the numeral that precedes them.
> 6 a.m.
> 10 p.m.
> In display type, tabular material, or lists, “a.m.”/“p.m.” style may vary depending on graphic design. For
> example, small caps may be used, or the periods may be deleted. [UI]

https://brand.uiowa.edu/sites/brand.uiowa.edu/files/2023-10/IOWA-Editorial-Style-Guide-10-30-23.pdf

# Proposed solution

cory-skeers:
Related to #5974 and possibly #6073

GaryRidgway:
While [working on this PR](https://github.com/uiowa/uiowa/pull/7219), I did some investigation on this issue but we decided not to follow through with larger changes, so I will note here things that I found.

I was able to produce a date string that looked like the following example:

### Jan. 19, 2024 | 9:07 - 10:07 am

Which is probably as close to what the PHP date formats could get to to match the formats listed [in the PDF above](https://brand.uiowa.edu/sites/brand.uiowa.edu/files/2023-10/IOWA-Editorial-Style-Guide-10-30-23.pdf).


[...]

PR #7219, opened 2023-12-18
[ITS] Outage and Degradation component

### LINKED ISSUE #7197: Current outages/degredation of service landing page component
# Story

As an ITS site owner, I would like to be able to display current outages and degradation of services on our services landing page so users can quickly find these and get more information about them.

# Additional Details

Mockup - https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7

View block that shows active (not closed) outages and degredation of service alert types.

Order - currently term weight (outage first), then post date

May need some cache busting since they are date/time based.

"View more" needs accessible link titles similar to card component.

Outages and Degradation of Services should have different style. 

No results text

# Proposed Solution

- [ ] Use additional details as a build spec.

### LINKED ISSUE #7216: Planned Maintenance component for Alert landing page
# Story

As an ITS site owner, I would like to be able to display planned maintenance alerts that will be happening over the next 7 days on our alerts landing page so that users can easily find information about upcoming maintenance and how it will affect them.

# Additional Info

Mockup - https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7

View block that shows active (not closed) planned maintenance alerts where the start date is less than or equal to now + 7 days.

Order by start date - asc

Theme according to mockup

No results text

Cache busting since date/time based?

# Proposed solution

- [ ]  Use additional details and https://github.com/uiowa/uiowa/issues/7216#issuecomment-1861518009 as a build spec.

cory-skeers:
Resolves #7197 
Resolves #7216 

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli && ddev drush @its.local uli
```
^ Note the added cache clear, which is necessary to get it to recognize the `ItsAlert` over the standard `Alert`.

Visit https://its.uiowa.ddev.site/admin/structure/taxonomy/manage/alert_categories/overview and try to delete any of the four category terms, and see that the delete button is disabled and a suitable message is displayed.
Add a new category taxonomy term and check that you are able to delete it. (Go ahead and delete it to keep things cleaner when creating alerts later.)

Create/edit a page and add the "Outages and Degradations", "Planned Maintenance", and "Service announcements" list blocks (feel free to test in whatever layout you'd like, but https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7 can be used as a reference mockup to recreate).

Check the no results messages appear (may have to delete the existing "Test Alert" outage).

Add some alerts with the various categories, and various states of "open" or "closed" (eg an "open" alert is one that is ongoing, or has a defined end date later than right now). Check out the views and see that they include any "open" alerts, but not any "closed" ones.

Add a new alert (outage/degradation, maintenance, or service announcement category) and refresh the page--see that the views refresh.
Delete an existing alert and refresh the page--see that the views refresh.

Create an alert with an end time close to the current time. Load the page with the views blocks and see that it appears in the appropriate view.

Test any other instances you can think of where an alert should appear or drop off of the views, and check that the caching handles it appropriately (if an alert is not created/edited, cron will need to be used)

```
ddev drush @its.local cron`
```

After cron runs, reload the page and see that the now closed alert no longer appears in the view.

## Iowa Now bug fix

Delete confirmation got moved to a modal, and so the message that previously got added as a warning isn't visible until after canceling (and depending on your actions, sometimes not viewed until moving to a different page, at which point all warnings appear). This moves it from a warning and instead replaces the modal's description with our own.

```
ddev blt ds --site=now.uiowa.edu && ddev drush @now.local uli /admin/structure/taxonomy/manage/tags/overview
```
Try to delete 'achievements' and see that you can't, and are given an accurate message that you cannot
Try to delete another term, and see that you can


cory-skeers:
Shouldn't be an issue with the further planned caching work, but noting:

How it is today, the view caches is not invalidated on adding a new node, but rather by updating any nodes included in the view and/or the custom cache invalidation. Need to ensure that is included that in any further updates.

cory-skeers:
Note to self: update the cache tag labeling, as it may be used for other alert views.

cory-skeers:
I wanted to get a quick check on my thoughts here: Is it wrong to render these alerts utilizing the card?

image

What I'm doing here is using the bundle class to wrap the teaser card for these specific categories to create the separate alerts. It has two primary consequences:

1. These are essentially cards within an alert, so they're a bit structurally different than an ordinary alert
2. It's altering how the teaser for these alerts are displayed, so if the teaser is used elsewhere (eg when it likely gets added to render on affected services), it will be given this same treatment

cory-skeers:
General dump of how the caching works for this as of now:

There is a custom cache tag "its-core-alerts" that is being added to the outage/degradation view (and could be used for other views), which by default is indefinite (any custom cache tag added through views is indefinite; the assumption is that if it is being added, custom cache invalidation must also be added).

There is a "its_core_alerts_next_clear" state that keeps track of a time, and if that time has passed at the time of cron running, the "its-core-alerts" tag is invalidated (and acquia purge triggered). When this happens, the next earliest clear time is queried and set for the state. This earliest clear time corresponds to the next time an alert would be considered to be closing. If there are no more alerts with defined close times in the future, then the cache tag won't be invalidated until something changes.

During an alert presave, if the alert is already closed, unpublished, or not considered one of the alerts we're concerned with (at this point, only 'outage' and 'service degradation' are tied to views, but this would expand later), nothing further happens. If it is published, not closed, and a category we're concerned with, it checks whether the "its_core_alerts_next_clear" date should be updated based on this new/changed alert, and invalidates the "its-core-alerts" tag.

I think this generally covers desired cache clears, so any new alerts and/or changes to existing alerts will be reflected in the view, while allowing for extended caches when there isn't much "action" going on. This will be lessened somewhat as more views are added (eg a page with Degradations/Outages, Planned Maintenance, and Announcements will end up losing cache when more or less any alert is updated).

The other little bit is that an alert will naturally fall off _around_ its close time, because that will be dependent upon cron running. I think this is okay, as it will be displayed with the end time, so if it appears cached for a while (less than an hour) while it is technically closed, the time that it ended would still be visible. If that's a no-go, I think we have to switch to a time-based cache on this and similar views and limit it to something like a 1 or 5 minute cache.

pyrello:
> Is it wrong to render these alerts utilizing the card?

I'm not sure that it is wrong, but potentially it is problematic for some of the reasons you have outlined. We haven't done much (any?) testing overlapping those two components.

IMO, these should be cards. I'm not a big fan of making them all use the alert styling (despite the content type being called an Alert) because I think when that styling is used too much it becomes kind of meaningless. If we have to use the alert styling, I guess we just add that to the card, see what happens and try to fix anything that looks broken?

bspeare:
@cory-skeers that looks like a creative solution!  If we are going that route, I wonder if we need to build the color coding and icons into cards instead?  There was some previous discussion on this at https://github.com/uiowa/uids/issues/770#issuecomment-1802659931.  @pyrello what do you think?

bspeare:
I agree with @pyrello: "_when that styling is used too much it becomes kind of meaningless_" - I'd keep them as cards are reserve the alert styling for when really necessary.

pyrello:
Possibly we could use a badge on the card with an icon to signify severity level?

joewhitsitt:
I think the caching strategy is worth trying for a while. I was trying to think of a situation where it wouldn't clear right, but like you said, it is ultimately dependent on how often cron runs. I think we might have to have a ITS-specific cron run setup like we currently do. The only thing I could think of is any change to the date that doesn't go through a presave, but that should be rare if at all.

bspeare:
It looks like this is still adding `alert` classes to the card.  I think we either need to use the alert template for these or card template with a badge with a `danger` icon?  Perhaps we can huddle about these when everyone is back tomorrow? 

Here are some examples where we could use a card with a badge inline in the title or under: https://sandbox.prod.drupal.uiowa.edu/it-service-alerts-copy

cory-skeers:
Before getting back to this, I could use a bit of clarification.

We're not going with the https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7 mockups with these, and are doing cards with either a badge or icon to help indicate severity?

pyrello:
> Before getting back to this, I could use a bit of clarification.
> 
> We're not going with the https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7 mockups with these, and are doing cards with either a badge or icon to help indicate severity?

That is correct. We would like to use this as the way forward. The UIDS alert component wasn't designed to be as flexible as the UIDS card component and based on the work that has gone into this, it seems like the card component would be better for what we actually want to be able to display. When it comes time to explain it to the customer, I volunteer as tribute.

cory-skeers:
> > Before getting back to this, I could use a bit of clarification.
> > We're not going with the https://sandbox.prod.drupal.uiowa.edu/mockup-its-alerts-landing-7 mockups with these, and are doing cards with either a badge or icon to help indicate severity?
> 
> That is correct. We would like to use this as the way forward. The UIDS alert component wasn't designed to be as flexible as the UIDS card component and based on the work that has gone into this, it seems like the card component would be better for what we actually want to be able to display. When it comes time to explain it to the customer, I volunteer as tribute.

Thanks for the clarification (and volunteer) @pyrello!

GaryRidgway:
Because it seems like this has become more similar to this issue, I will use this PR to work on [this issue as well](https://github.com/uiowa/uiowa/issues/7217).

GaryRidgway:
Hey, we are noticing some odd caching behavior with service announcements. Not sure if this is because of the new strategy or not, but thought I would say something.

cory-skeers:
> Hey, we are noticing some odd caching behavior with service announcements. Not sure if this is because of the new strategy or not, but thought I would say something.

I pushed up a quick addition in https://github.com/uiowa/uiowa/pull/7219/commits/ce0dea1fe7bb904777562f225d9a56d8d3f1d1ae that I think should cover it. The presave logic to check whether the cache tag should be cleared or not was/is limited to specified alert categories since the taxonomy could potentially be expanded beyond those used in here. I add the service announcements to it.

We could take out that condition and just leave it to check all alerts regardless of category, while still retaining the attempt at optimization where it checks if the alert is closed or not.

e-marie-w:
Pulling down to test.

bspeare:
It looks like the "Service announcements" block might need some no results text?  Screenshot 2024-01-17 at 11 21 37 AM

bspeare:
The no results alerts markup in the view can be updated to include:

- `alert--vertically-centered`
-  `<div>There are currently no alerts</div>` (without `<p>` tag)

GaryRidgway:
> It looks like the "Service announcements" block might need some no results text?

I think @cory-skeers and I decided that we didn't want service announcements to display if there weren't any, but I can't specifically remember why that was the decision right now. Do we want to add that in? I could go either way

GaryRidgway:
> The no results alerts markup in the view can be updated to include:
> 
> * `alert--vertically-centered`
> * `<div>There are currently no alerts</div>` (without `<p>` tag)

I can get on this, thanks Ben!

GaryRidgway:
After updating some of the alert "No results text", I'm feeling like maybe we should be styling the no results the same for each category? @bspeare do you think the green success alert style would be too overwhelming if we used it for every list?

bspeare:
A few more questions/comments after testing this:

- Should we disable the "items to display" view block option?
- Should we change the pager to Full as that is our default SiteNow pager default?   I'm testing with devel_generate and the pager is not appearing. 
- Should Outage and Service Degradation both have icons or different colors? See details below. 

<details>
Screenshot 2024-01-17 at 11 41 49 AM
</details>

- Do we want to implement any of these style guide time fixes from https://github.com/uiowa/uiowa/issues/7122#issue-1995086341? 

- Would it be useful to move the building into `card_meta` with an icon?  It might create a more cohesive visual grouping of the information being presented.  See details below. 
<details>
Screenshot 2024-01-17 at 11 53 35 AM
</details>

bspeare:
@GaryRidgway I think "Outages and Degradations" was the only category that they wanted that type of major alert notification for.  I think it makes sense to not adjust the rest.

GaryRidgway:
> * Should we disable the "items to display" view block option?

I could maybe see them using smaller numbers of items, but it does seem like an outlier? I don't think there is a lot of harm leaving it though.

> * Should we change the pager to Full as that is our default SiteNow pager default?   I'm testing with devel_generate and the pager is not appearing.

If that is the default, then sure! I think that is a good detail to keep consistent.

> * Should Outage and Service Degradation both have icons or different colors? See details below.

When we were in the meeting they told us "Triangle" for outage, nothing for Service degradation. I would be open to another color for the service degradation though, as it would be good for it to visually be different from the outages so that the outages are more eye catching. Maybe a yellow or info blue? I think yellow feels more in line with what i think of with "Service degradation", but since it is our primary, perhaps blue would be more visually catching.

> * Do we want to implement any of these style guide time fixes from [Follow Brand editorial style guide for outputting months and time #7122 (comment)](https://github.com/uiowa/uiowa/issues/7122#issue-1995086341)?

I can look in to this. I think we should.

> * Would it be useful to move the building into `card_meta` with an icon?  It might create a more cohesive visual grouping of the information being presented. 

I think this would look nice! I think the services are bundled in the provided picture though, should we separate those with an icon as well?

GaryRidgway:
How does this look for all the "no results text"?
Screenshot 2024-01-17 at 1 34 26 PM

bspeare:
>  services are bundled in the provided picture though

Good point, maybe just the info `<span role="presentation" class="fas fa-info-circle fa-li"></span>` circle then?

GaryRidgway:
> > services are bundled in the provided picture though
> 
> Good point, maybe just the info `<span role="presentation" class="fas fa-info-circle fa-li"></span>` circle then?

Sounds good to me!

bspeare:
Screenshot 2024-01-18 at 1 12 13 PM

I adjusted the alert teaser display to use the card_meta to display the fields.  The one difference here as you can see on the screenshot is that it will display the service if there is one item where as we were previously only displaying services if there were multiple.   Maybe it is worth displaying the service and building here in the info slot no matter the number of entries?  This might make the alerts easier to scan to know that the service will be consistently present within that info field.

joewhitsitt:
I'll be pulling this down to test. Anything that can be glossed over as it was already reviewed?

bspeare:
@joewhitsitt no, after the changes in the past few days I think this needs a full review of alert teaser styles, alert views blocks, and open/closed caching test.

docroot/sites/its.uiowa.edu/modules/its_core/its_core.module
```diff
@@ -65,11 +70,67 @@ function its_core_entity_presave(EntityInterface $entity) {
...
       break;
 
   }
 }
 
+/**
+ * Implements hook_preprocess_fieldset().
+ */
+function its_core_preprocess_fieldset(&$variables) {
+  if (
```
joewhitsitt: Wondering if we can avoid the regex, etc. and handle this similar to https://github.com/uiowa/uiowa/blob/main/docroot/sites/facilities.uiowa.edu/modules/facilities_core/facilities_core.module#L436

```
  if (isset($variables['element']['#type']) &&
    $variables['element']['#type'] === 'radios') {
    $admin_context = \Drupal::service('router.admin_context');
    if ($admin_context->isAdminRoute()) {
      $node_type_data = $current_route?->getParameter('node_type');
      $node_type = $node_type_data?->get('type');

      if (isset($node_type) && $node_type === 'alert') {
        $variables['description_display'] = 'before';
      }
    }
  }
```

joewhitsitt:
Added follow up issue to address the front page prevention stuff: https://github.com/uiowa/uiowa/issues/7281

config/sites/its.uiowa.edu/config_split.config_split.site.yml
```diff
@@ -10,7 +10,7 @@ stackable: true
 storage: folder
 folder: ../config/sites/its.uiowa.edu
 module:
-  its_core: 0
+  its_core: 10
```
joewhitsitt: just curious why this had to be bumped up.

cory-skeers: I think for ordering of field preprocessing, but there have been a lot of changes in what's changed and how, so it may not be necessary any more...I wish I had a better recollection of precisely what needed to happen in order.

config/sites/its.uiowa.edu/config_split.patch.core.entity_view_display.node.page.default.yml
```diff
@@ -0,0 +1,12 @@
+adding:
+  third_party_settings:
+    layout_builder_restrictions:
+      entity_view_mode_restriction:
+        allowlisted_blocks:
+          'Lists (Views)':
+            - 'views_block:alerts_list_block-outages_degradations'
```
joewhitsitt: Am I reading this right? Can we avoid this now that it is tagged with the Site Custom views category?

Screenshot 2024-01-19 at 11 43 49 AM

joewhitsitt:
Got some questions out here, but overall this seems to be working as tested. I don't love how often we are writing those tids into the code. We do anymore with follow-up work (emails, etc.) I think we should keep it in one place and reference it so we don't have to remember all the places these are if they change, etc.

I'll approve after the questions are answered.

config/sites/its.uiowa.edu/config_split.config_split.site.yml
joewhitsitt: thanks, i am fine with keeping it. hopefully we don't need to revisit it again.

docroot/sites/its.uiowa.edu/modules/its_core/its_core.module
joewhitsitt: Shared a DM with Alan about this. Initially there was some concern of the reduced specificity of not singling out this one field, "Category". I think that we have scoped to the site, content type, field type and that should be enough. If another radios field is added to the form, it will be consistent between the two. Scoping on a field label is a bit shaky as a label change would stop this from functioning.

config/sites/its.uiowa.edu/config_split.patch.core.entity_view_display.node.page.default.yml
GaryRidgway: Messaged Joe personally, he will re-export the block options to fall in line with our desired strategy.

joewhitsitt:
Ack, final question. Do we need a diff between weights on alert teaser between the feature default and ITS?

https://github.com/uiowa/uiowa/pull/7219/files#diff-3b89de1592fb19ad0fe3db2dcb1b375b497c24efd97f1c6678d50d5ab7bd1aed

bspeare:
@joewhitsitt no, I think they should be the same weight.

GaryRidgway:
I think we do so that the date appears over top of the badge. The intent is to have the badge last i think?

GaryRidgway:
> @joewhitsitt no, I think they should be the same weight.

I AM OUTVOTED

bspeare:
I can double check in a few minutes.

config/sites/its.uiowa.edu/config_split.patch.core.entity_view_display.node.page.default.yml
joewhitsitt: locally enabled Devel was also getting added to this.

joewhitsitt (approved):
Thanks for working on this y'all!

joewhitsitt:
One more thing I noticed unrelated to this PR is that the alerts list block default seems kind of restrictive on what is shown. I couldn't get an alert that was ongoing to show. If another site uses this feature beyond ITS and Facilities, we may need to revisit this filtering logic.

Screenshot 2024-01-19 at 3 19 46 PM

Issue #7156, opened 2023-11-30
Develop communication plan for SiteNow v1 standard site owners


briand44:
# Story

As a webteam, we would like to have a communication that we can send to SiteNow v1 site owners letting them know the timeline and what is required of them to move to SiteNow v3 and how to respond with questions. 

# Additional Info

Identify timeline?

Should it contain something about what to do if you no longer need your site?

Are we migrating anything?

# Proposed solution

```[tasklist]
### Tasks
- [x] Create a page/article for communicating details about the end of service for SiteNow V1
- [x] Develop a timeline that includes relevant dates and deadlines
- [x] Identify the audiences we will be targeting for communication
- [x] Identify the communication channels that will be used for each message
- [x] Articulate the intent of the messages that we would like to send to each audience
- [x] Create a schedule for when the messages should go out
- [x] Finalize and publish an initial message on V1 sites
- [x] Draft and finalize a first communication to be sent
```

briand44:
Grooming conversation... should this be more about the communication plan vs the actual communication?

Who to send to?
Opt in vs opt out?
timelines?
rough idea of what we want to say in first communication

bspeare:
A few thoughts:

**Communication:**

- Reduce as much technical jargon as possible and keep the communication extremely short with a distinct call to action (button)
- Create a "Keep my website" or "Yes" link that is pre-populated with webform fields
  - If we were able to do that, could we auto-submit the form as well?
- Allow users to respond to the email if they want?

**Plan:**

- Create a buffer before eol, where we could turn off the domain in Acquia Cloud and easily turn it back on if the user notices their site is down so they could have enough time to request a new V3 site?

pyrello:
I started a draft timeline for a V1 standard service shutdown. Ping me for access.

https://iowa-my.sharepoint.com/:w:/r/personal/sadamshiett_uiowa_edu/Documents/Draft%20timeline%20for%20SiteNow%20V1%20standard%20service%20shutdown.docx?d=w5a9b208f418243f694574b643cd709a4&csf=1&web=1&e=12d8Cp

bspeare:
Wondering if we might want to do an unbounce popup instead of or in addition to an email communication? Maybe the email communication is only for inactive sites and we rely upon an unbounce popup for active sites?

pyrello:
[...]

PR #7220, opened 2023-12-19
D7 comms report command

pyrello:
Related to #7156 

# How to test

This was already used to develop the initial list of emails/websites that we used in #7156. I think code review is sufficient.


joewhitsitt:
I told Sean I'd take a look at this, particularly the issue where some emails aren't returned. pulling down

pyrello:
We'll be using this again for #7248

joewhitsitt:
Did we want to include any safeguard from running this on a remote instead of locally?

pyrello:
> Did we want to include any safeguard from running this on a remote instead of locally?

We didn't previously, but I can add something.

pyrello:
> > Did we want to include any safeguard from running this on a remote instead of locally?
> 
> We didn't previously, but I can add something.

I thought I would find an example of this, but it seems like we don't actually implement this type of guard with most of our BLT commands. If you have an example, please surface it!

joewhitsitt:
My concern is the CSV export. I think only this ReportCommands has them and I didn't think of this before when we first shipped it out. I don't have an example to surface, but I wonder `isDDev()` from blt/src/Blt/Plugin/Commands/ValidateCommands.php or `$env = EnvironmentDetector::getAhEnv() ?: 'local';` would be enough to prevent a CSV from being saved on one of the remotes in a directory where it shouldn't be.

pyrello:
> My concern is the CSV export. I think only this ReportCommands has them and I didn't think of this before when we first shipped it out. I don't have an example to surface, but I wonder `isDDev()` from blt/src/Blt/Plugin/Commands/ValidateCommands.php or `$env = EnvironmentDetector::getAhEnv() ?: 'local';` would be enough to prevent a CSV from being saved on one of the remotes in a directory where it shouldn't be.

I'm not sure about this, but I don't think permissions would allow a file to be written in the project root on prod. I can look into adding this.

GaryRidgway:
Would someone be able to review this code for me? I think it works but I need someone else to try it remotely and see if it terminates before trying to create the CSV. (Reason: My ddev is giving me grief.)

joewhitsitt:
Sorry, my thought from https://github.com/uiowa/uiowa/pull/7220#issuecomment-1883199328 is that we just don't want to export a csv to a remote environment, not prevent the command from running at all. 

https://github.com/uiowa/uiowa/pull/7220/files#diff-e58379d737967dfd847c0da29f5198c12e242c198a048fe22c46c9ea35bf93e7R417

joewhitsitt:
in picking up https://github.com/uiowa/uiowa/issues/7248, I am pulling this down and will try to articulate my thought process from https://github.com/uiowa/uiowa/pull/7220#issuecomment-1898984856

joewhitsitt (approved):
My original concerns has been addressed. Previously reviewed and have been using this command locally. I approve, but since I was the last to touch, another thumbs up would be neat.

Issue #7221, opened 2023-12-19
Card media sizes are being displayed for all formats


bspeare:
# Problem
Card media sizes are being displayed for all formats

# Steps to reproduce
Edit https://sandbox.prod.drupal.uiowa.edu/node/21/layout and remove an existing image and confirm that the "Media size" option is now showing for all media. 

# Proposed solution
Revisit code in https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/js/card-block-form.js

bspeare:
I'm not sure that I'm still seeing this bug so I am going to close this.

PR #7226, opened 2023-12-21
Card-block-form.js machine name updates

bspeare:
Related to https://github.com/uiowa/uiowa/issues/7221 but does not resolve that issue. 

In https://github.com/uiowa/uiowa/commit/e5c62aadc2df3df19c717df3b20bc9571c73a9d7, we updated the media size style and missed changing the machine name in a few spots. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/1221/layout
```
- Inspect source on the layout builder styles and verify that the machine name change updates are correct
- Set media position to stacked and circle small, then attach a video and save the card
- Confirm in the html source that `media--large` class is added
- Run the same steps on prod and see that the `media--small` class is added


e-marie-w:
I'm really struggling with testing this for some reason, not confident if it's a me thing or environment thing so I'm unassigning myself.

joewhitsitt:
pulling to test

joewhitsitt (approved):
This fix checks out. Overall, still funky and could be improved in https://github.com/uiowa/uiowa/issues/7221

PR #7231, opened 2023-12-27
OnIowa Event Map

### LINKED ISSUE #7052: Dynamically embedded map on event node
# Story

As an OnIowa site owner, I would like to have an embedded map appear on each event node dynamically so that a site visitor can interact with the map directly on the site and not have to link out to Google Maps.

# Additional Info

Trying to get clarity on using lat/long vs address.

Trying to understand if they have a preference on placement.

# Questions

How would we want to implement this? Google Maps? This would require an API key and potentially costs, correct? Other preferences on implementation?

# Proposed solution

e-marie-w:
Resolves #7052 

# How to test
- `ddev composer install && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli`
- Check an existing event from https://oniowa.uiowa.ddev.site/events, confirm that the sidebar now renders a map view from OpenStreetMap instead of a link to Google Maps.
- Create an event, confirm that the geofield is still set to take in lat and long values
- Check an event with no lat/long values to confirm no map appears
- Test other sites using the event feature (admissions, iowasummerwritingfestival) also work as expected.

Currently the map is set to not appear if no lat/long, and to lazy load if there is. Open to any feedback about those choices, the Zoom level, and anything else.

- `ddev blt ds --site=pharmacy.uiowa.edu && ddev blt ds --site=admissions.uiowa.edu` and check that the https://pharmacy.uiowa.ddev.site/pharmd/dual_cert/palliative/certificate_graduates and https://admissions.uiowa.ddev.site/counselors still render the map and that the map assets load from the module instead of /libraries


e-marie-w:
Just realized this is should be in the event feature split, not the site split.

joewhitsitt:
I am interested in testing this. One thing I would like to look at is whether we can remove our direct call to the leaflet/leaflet library within our composer.json file and just rely on this module to manage it for us, and if updating leaflet affects our existing usage.

```
        "leaflet": {
            "type": "package",
            "package": {
                "name": "leaflet/leaflet",
                "version": "1.7.1",
                "type": "drupal-library",
                "dist": {
                    "url": "https://github.com/Leaflet/Leaflet/archive/v1.7.1.zip",
                    "type": "zip"
                }
            }
        },
```

joewhitsitt:
pulling to test

joewhitsitt:
I was able to reduce our leaflet footprint and allow the module to handle the dependency.

joewhitsitt (approved):
Looks good to me. There is decent left and right margin around the map on mobile, otherwise we might need to implement that gesture support.

In the future we can do more UI Branding treatment if desired.

joewhitsitt:
@e-marie-w sorry for the scope creep. thanks for reviewing the addition. let me know if you would like help with the composer.lock

PR #7232, opened 2023-12-27
[Bug] Fix component order in replicated pages

### LINKED ISSUE #7102: Replicate for layout builder is getting order incorrect
# Problem
Using replicate for layout builder based pages, the order of items is not always correct.

**Before**

before

**After**
after

# Steps to reproduce
- Try replicating a complex layout builder page.

# Proposed solution
- [ ] The replicate module recently updated as part of the D10 upgrade and we still have some custom code. Perhaps we need to revisit and determine if code needs to be adjusted or removed. (https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/ReplicateSubscriber.php can likely be removed completely)
- [ ] Test patch mentioned in https://github.com/uiowa/uiowa/issues/7102#issuecomment-1811235901

cory-skeers:
Resolves #7102 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /replicate-tester
```

Replicate the page and see that the left column has the correct order, but the right column and bottom section do not

```
ddev composer install
```

Replicate the page again, and see that the cards in both columns and bottom section are in the correct order

Test with any others. The culprit comes into play after moving a component. Prior to the patch, components are essentially replicated in the order they were _created_, rather than their actual placement, which works if components are not rearranged, but breaks once they have been moved.


cory-skeers:
Needs some cleanup, but basically, because of the way `->appendComponent(...)` works and because our list blocks get skipped in the layoutBuilder replicate event subscriber, replicated list blocks are always moved to the top of a section in which they're placed after replication.

Our custom subscriber fires after the replicate subscriber and (along with adding a custom revision message) was already replacing the duplicated list block with one that has a new uuid, so I did a bit of exploring in how we might also put it into the correct spot. The original weights are lost in the first subscriber (part of the `appendComponent()` workings), so it looking back at the original entity again, unfortunately.

cory-skeers:
Also removing the 'needs cache rebuild' label. That should only have been necessary if we were going to remove our additional event subscriber, but I think we want to keep it for the uuid and revision additions, as well as maybe this list block reorder. And in that case, we're not removing a service, so no cr needed.

composer.json
```diff
@@ -282,6 +282,9 @@
             "drupal/redirect": {
                 "Match redirect": "https://www.drupal.org/files/issues/2020-04-13/redirect-n2831605-31.patch"
             },
+            "drupal/replicate": {
+                "Replicate components in the proper order.": "https://git.drupalcode.org/project/replicate/-/merge_requests/8.diff"
```
joewhitsitt: This should be brought locally and stored in the patches directory so that if it changes upstream we won't automatically accept the changes.

joewhitsitt:
Meant to review this yesterday. pulling down and i will fix the patch bit.

joewhitsitt (approved):
Glad you were able to figure this out. So far in my testing, nothing has tripped it up. Thank you very much for the detailed comments!

PR #7244, opened 2024-01-03
DDEV update changes `.ddev/providers/acquia.yaml`

GaryRidgway:
This is a file that changed for ddev 1.22.6

## To test
- Update ddev version v1.22.6 (`brew upgrade ddev`)
- Look at the file, is this going to break anything?


.ddev/providers/acquia.yaml
```diff
@@ -1,75 +1,71 @@
...
-#     acli remote:ssh -n ${project_id} -- "cd /tmp && gunzip db.${TIMESTAMP}.sql.gz"
-#     acli remote:drush -n ${project_id} -- "sql-cli </tmp/db.${TIMESTAMP}.sql"
-#     acli remote:drush -n ${project_id} -- cr
-#     acli remote:ssh -n ${project_id} -- "rm /tmp/db.${TIMESTAMP}.*"
+db_push_command:
+  command: |
+    set -eu -o pipefail
+    export ACLI_DB_HOST=db ACLI_DB_NAME=db ACLI_DB_USER=db ACLI_DB_PASSWORD=db
+    set -x   # You can enable bash debugging output by uncommenting
+    acli push:db ${ACQUIA_ENVIRONMENT_ID} --no-interaction
```
pyrello, line 64: Noting the comment here. Probably we need to understand what this actually is before we make this change.

.ddev/providers/acquia.yaml
```diff
@@ -1,75 +1,71 @@
...
-#   command: |
-#     # set -x   # You can enable bash debugging output by uncommenting
-#     set -eu -o pipefail
-#     ls ${DDEV_FILES_DIR} >/dev/null # This just refreshes stale NFS if possible
-#     drush rsync -y --alias-path=~/.drush @self:%files @${project_id}:%files
+files_push_command:
+  command: |
+    # set -x   # You can enable bash debugging output by uncommenting
+    set -eu -o pipefail
+    acli push:files ${ACQUIA_ENVIRONMENT_ID} --no-interaction
```
pyrello, line 71: Same here.

joewhitsitt:
Yeah, I think even though the update re-wrote these, we don't want to accidentally trigger them. https://github.com/uiowa/uiowa/pull/4458/commits/30d78a4c771f8d68e6e885ef5ec6e09ab6d73dfa

GaryRidgway:
Hmm, I'm confused then, if it is automatically generated, how is it that we persisted those commented out sections before? And how might we do that again if funning `ddev` regenerates it?

pyrello:
> Hmm, I'm confused then, if it is automatically generated, how is it that we persisted those commented out sections before? And how might we do that again if funning `ddev` regenerates it?

No idea. But it seems clear from reviewing the changes that this represents a change as-is from how we previously had it set up. If that is an upstream change, it might be worth understanding more. Or it might be easier to just comment those sections out since it seems to be functionality we don't currently use. I could go either way.

joewhitsitt:
Looked upstream at DDEV's release notes and related issues/pulls. The shift is that they don't think this file should be edited anymore and instead the web_environment info should be provided in the config.yaml file if needed. Those doc comments say we can delete those commands.

I am going to pull down and see if I can get this in a good spot.

joewhitsitt:
actually, i am not sure we use this integration. this might have just been a remnant from adoption testing. for the most part i think we are using blt to interact with acquia.

I think we can try to fly without editing or providing the web environment variable and see if any of our command call these and error out. these commands require the ACQUIA_ENVIRONMENT_ID `uiowa.dev` to be passed in.

joewhitsitt (approved):
Approving after reviewing the upstream changes and testing some blt commands (e.g. `ds` `ugc`).

>I think we can try to fly without editing or providing the web environment variable and see if any of our command call these and error out. these commands require the ACQUIA_ENVIRONMENT_ID uiowa.dev to be passed in.

If something like provisions fails, we should be able to just add the id to to the config and proceed.

joewhitsitt:
Also, while I think it would be good to have everyone upgrade soon `brew upgrade ddev` and `ddev poweroff && ddev auth ssh && ddev composer install`, I don't think your local envs will be affected in the short term.

PR #7254, opened 2024-01-05
Alert category descriptions.

### LINKED ISSUE #7225: info popup on alert edit form that displays category term descriptions
# Story

As an ITS website owner, I would like alert category descriptions to be easily visible when adding/editing alerts so that alert creators do not miscategorize alerts.

# Additional Info

Comment from Joe - https://github.com/uiowa/uiowa/issues/5952#issuecomment-1863168989
>I think we can use the existing select widget but change the reference method to call an entity reference view. The entity reference view could have a title | description (strip tags) result for each row. If the descriptions are long maybe the help text could link to an article. If we run into trouble with formatting a select list dropdown, we could switch to radio list since it is single value, required.

Screenshot 2023-12-22 at 2 36 10 PM

https://drupal.stackexchange.com/questions/297321/what-does-the-views-filter-by-an-entity-reference-view-reference-method-do

# Proposed Solution

- [x] Try the entity reference view that Joe suggested ^^
    - We can use descriptions from https://its.uiowa.edu/support/article/107561 for testing if they haven't populated the term descriptions yet.

GaryRidgway:
Adds alert category descriptions to the alert category field on the alert content type.

## How to test
Log in to the site after syncing it and look at the alert categories on the alert content type. Do they feel natural to you? Do you think this solves the need of the linked issue?


cory-skeers (approved):
Reviewed and tested, and I think this looks great as a way to give ITS webmasters some control over adjusting help text (ie term descriptions) if needed, without it requiring a code push.

With the way the descriptions are displayed, they can allow for quite a bit of length, so I think there's a good chance this will fit their needs. The only issue I see is that the term descriptions are filtered html and the button help text strips html, and this means the "If the descriptions are long maybe the help text could link to an article" isn't an option outside of placing the full URL in there and having users copy+paste it. (It isn't just that the view strips html--even if html is allowed, links in particular get stripped from it.) Similarly, if are expecting to be able to do something like a bulleted list, that's out.
image

But from the last meeting, I think they are open to how it's implemented and really are looking for _some way_ to provide a little extra context, and that as long as they're aware that the formatting needs to remain simple, this fits what they're looking for.

Issue #6960, opened 2023-10-04
Feature request: Allow users who are “Blocked” to be referenced in the “Authored by” field


briand44:
Requested by Genevieve Johnson

Without details, not sure if this is related: https://github.com/uiowa/uiowa/issues/4761 @joewhitsitt

joewhitsitt:
@briand44, I am confused by this issue. Is this saying a blocked users can't be added to an authored by field or is there an error when saving an existing piece of content with a user that has been blocked?

joewhitsitt:
If "blocked users can't be added to an authored by field" i am not sure i understand the use case.

joewhitsitt:
>is there an error when saving an existing piece of content with a user that has been blocked?

Bummer, looks to be this.

<img width="340" alt="Screenshot 2024-01-10 at 10 01 58 AM" src="https://github.com/uiowa/uiowa/assets/4663676/e9dd3f05-f4dc-40e6-a612-9167da2403cf">

On next save of the content, a new author has to be assigned.

joewhitsitt:
This seems to go against #4760

Issue #4761, opened 2022-02-04
Media owned by canceled account is not assigned anonymous user ID


richardbporter:
# Problem
Deleting a user and making its content belong to anonymous does not work properly for media. That results in the media UID being null and throwing a WSOD when saving it.

# Steps to reproduce
Cancel an account and make its content belong to anonymous user. Masquerade as webmaster. Try editing and saving a media item that was owned by the canceled account. See WSOD.

# Upstream

https://www.drupal.org/project/drupal/issues/3057004
https://www.drupal.org/project/drupal/issues/3043725

# Proposed solution
TBD.

joewhitsitt:
Should the cancel operation even be available to webmasters if "blocked" is our recommended workflow?

pyrello:
> Should the cancel operation even be available to webmasters if "blocked" is our recommended workflow?

Is "blocked" our recommended workflow because we actually think that is the best option? Or because we are afraid of what will happen if they cancel and an account? If it is the former, then I think we should resolve the issues that make cancelling an account not work as expected.

joewhitsitt:
Blocked is what I have recommended because it has been the easiest to explain/same effect of our most ideal cancel option "Disable the account and keep its content."

So far the concern of data loss/workflow issues have outweighed the user table count/GDPR privacy requests.

I think if we keep the cancel functionality, we should review the options to make it less scary, but then ultimately we have this issue (WSOD).

pyrello:
Should we consider changing the default view for users to exclude blocked users? The site owner could still filter to blocked accounts.

joewhitsitt:
With https://github.com/uiowa/uiowa/pull/7267 this should be deprioritized as accounts going forward can't be cancelled by webmasters. Existing cancelled accounts and any privacy based cancellations performed by the team could still run into this issue.

PR #7267, opened 2024-01-10
remove permissions related to user cancellation

joewhitsitt:
Due to issues with cancelling accounts (#4761 and [moderation revisions](https://www.drupal.org/project/drupal/issues/3089747)), we are removing the ability for webmasters to cancel accounts. Instead we recommend blocking the account.

In the future, we may be able to make this available again.

Content authored by a blocked user must be reassigned for successful save. See https://github.com/uiowa/uiowa/issues/6960

# To Test

```
```

As webmaster, you no longer see the ability to cancel a user account. The option is still in the bulk operations, but you get an access denied message on execute.

Screenshot 2024-01-10 at 10 00 41 AM


joewhitsitt:
Updated docs:

>We have temporarily disabled the ability to cancel accounts. We recommend changing the user status to blocked to prevent the user from logging into the website. Please contact us if the user must be completely removed from the website due to privacy reasons.

https://sitenow.uiowa.edu/node/181/latest

bspeare (approved):
Works as expected.  Draft copy change look good as well. Thanks!

joewhitsitt:
Published the documentation change as well.

PR #7275, opened 2024-01-18
Facebook Media - switch to using embedded posts feature instead of oembed_read api

joewhitsitt:
media_entity_facebook added support for https://developers.facebook.com/docs/plugins/embedded-posts/ which bypasses the API requirement. This does not solve our issue with Instagram, but cuts the issue down in half or so.

# Test

```
ddev blt ds --site=sitenow.uiowa.edu
```

Go to https://sitenow.uiowa.ddev.site/documentation/media/social-media-embed and see the facebook embed render assuming your browser (e.g. Brave) is not blocking third-party assets. Note: Instagram embed lower in the page is still not rendering.

Screenshot 2024-01-18 at 2 00 15 PM


joewhitsitt:
based on how the app review is going, I think this is a good temp workaround or permanent solution.

GaryRidgway:
pulling to test

GaryRidgway (approved):
This works as stated, though it does say that the facebook credentials are not properly configured at the top, which may confuse users? either way, it works!

joewhitsitt:
>though it does say that the facebook credentials are not properly configured at the top

Sorry, that is a local issue if you don't have the credentials in your local settings file. I keep forgetting that part of the instructions and assume folks have it setup.

joewhitsitt:
Updated our docs here about Facebook no longer relying on the API: https://iowa.sharepoint.com/sites/drupal-developer-documentation/SitePages/Miscellaneous/Facebook_Instagram%20Graph%20API%20for%20oEmbed/Facebook_Instagram%20Graph%20API%20for%20oEmbed.aspx

PR #7279, opened 2024-01-19
[CCOM] Analytics split

### LINKED ISSUE #7091: CCOM Specific Analytics / Tag Manager
As the development/web strategy team supporting CCOM websites, we want to adjust the tag manager (and analytic) environments used by our CCOM sites for compliance, and strategy reasons.

What this means is:
1. We would like to explore a way to exclude CCOM sites from including the global SiteNow Google Tag Manager.
2. We would like to include our own tag manager environment.
3. The tag manager we likely will be using (still in contracting) is Piwik Pro.

This is being done due to our legal and compliance team's interpretation of what Google products mean for health care properties. While CCOM sites are not direct patient care sites, they are adjacent and still a part of our single covered entity footprint.

For this reason we believe longterm, it is best to remove 3rd party Google products from the web experience.

The current "plan" is to leverage Piwik Pro, https://www.drupal.org/project/piwik_pro

# Proposed Solution

- Identify what is all in campus wide GTM.
- Stop campus wide GTM from running on CCOM sites
- Enable and allow admins to configure https://www.drupal.org/project/piwik_pro for CCOM sites
- Possibly turn off piwik in dev and stage environments (are we doing this with google analytics?)
- Turn off google analytics functionality for CCOM sites if easy - nice to have but not requirement
- Hopefully this can be controlled by the existing CCOM feature split

cory-skeers:
Resolves #7091 

# How to test

```
git pull && ddev composer install && ddev blt ds --site=community.medicine.uiowa.edu && ddev drush @medicinecommunity.local uli
```

Check that the piwik pro module is on, and that you can adjust the settings as an admin (the settings should be savable and ignored to be configurable by each site, but not having a valid id for it, this is difficult to actually test)

Check that the campus-wide gtm is not added on page load

```
ddev blt ds --site=default && ddev drush @default.local uli
```
Check that the piwik pro module is not on, and that campus-wide gtm is added on page load

Assuming the above works as expected, the config value is properly working on setting (or not) the campus-wide gtm container.

```
ddev drush @default.local cget uiowa_core.settings
```
See that `uiowa_core:campus_gtm` is `1`

```
ddev drush @default.local en ccom_core && ddev drush @default.local cim -y && ddev drush @default.local cget uiowa_core.settings
```
Check that enabling the module and config import goes smoothly, and that `uiowa_core:campus_gtm` is now `0`

```
ddev blt di && ddev drush @default.local cget uiowa_core.settings
```
See that `uiowa_core:campus_gtm` is `1` on a new install.


GaryRidgway:
I will pull to test this

GaryRidgway:
Most things I could test for here seem to be working, though I am hesitant to say that everything works "As intended" as we cant really test this until it's on a production site, and we have a site ID. In that case, I would say that I am okay with approving this in order to get it to the next stage of testing, as long as it is okay with everyone else. Perhaps we bring it up in EOD?

joewhitsitt:
>get it to the next stage of testing, as long as it is okay with everyone else

Reviewed, not tested. My only suggestion would be to create another setting that uiowa_core looks at before applying the script. something that ccom_core can set before uiowa_core fires. That way other modules can use it without doing the same module weight stuff.

```
if ((int) $env === 1) {
if (!$prevent_gtag) {
      $page['#attached']['library'][] = 'uiowa_core/gtag';
}
    }
```

cory-skeers:
Setting back to draft for 
> create another setting that uiowa_core looks at before applying the script. something that ccom_core can set before uiowa_core fires. That way other modules can use it without doing the same module weight stuff

and composer conflicts

cory-skeers:
Changes to this were to add a new config in `uiowa_core.settings` which is used to check if the campus gtm container should be added or not. There's an update hook to set it to `0` for sites with the CCOM split enabled and `1` for all others.

The last remaining work is to automate changing it to `1` when the CCOM split is enabled. I'd initially put this as a module install hook for `ccom_core`, but this doesn't work because the module is already installed, just not enabled. Once that gets moved to a better spot, re-testing on a new site install, standard site sync, standard site that enables the ccom split, and current ccom site to ensure the config is right for each.

cory-skeers:
This should be ready to review again.

Updated testing instructions.

The check is now on the side of adding the campus GTM container, rather than removing it for CCOM. This is based on a new config value in `uiowa_core`, so in addition to checking that that actually happens as expected, new testing needs to be around checking that

- Existing non-CCOM sites have that new config set to 1 by the update hook
- New site installs get that config set to 1 by default
- Existing CCOM sites have that config set to 0 by the update hook
- New sites enabling the CCOM feature (will now be enabled by enabling the module and config import, rather than activating the site split) will then be updated to have that config set to 0

> New sites enabling the CCOM feature (will now be enabled by enabling the module and config import, rather than activating the site split) will then be updated to have that config set to 0

The readme has been updated, but this is a change in process and we'll want to touch base on this

bspeare:
Pulling to test this.

bspeare (approved):
Testing worked as expected.

PR #7285, opened 2024-01-22
[Facilities] Project data

### LINKED ISSUE #6447: Facilities fields for project data coming from API
# Story

As a webteam, we would like to have structured fields for the project data coming from the Facilities APIs so that we can store this data similar to how we are storing the building information.

# Additional Info

We currently have a project content type with fields that will be manually populated.

Current code example on 702 `/docroot/sites/www.facilities.uiowa.edu/modules/custom/facilities_projects/includes/facilities_projects.single_project.inc`

single project endpoint - https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=0897601
projects by building endpoint - https://buildui.facilities.uiowa.edu/buildui/ext/projects?bldgnumber=0025
capital projects endpoint - https://buildui.facilities.uiowa.edu/buildui/ext/capitalprojects
featured projects endpoint - https://buildui.facilities.uiowa.edu/buildui/ext/featuredprojects

**Info coming from API - don't need add/edit ability**
- [ ] Title - `buildingName . ' ' . projectTitle` is currently being used on the project page. Building page outputs `buiProjectId - projectTitle`. Capital and Featured projects pages outputs `buildingName . ' - ' . projectTitle`
- [x] Project Number - looks like these are always a number - `buiProjectId`
- [x] Project Scope - plain text text area - `projectScope`
- [x] Project Status - text? I've seen examples of "Construction" and "Design", there could be others `projectStatus`
- [x] Square Footage - number - `number_format($d, 0, '.', ',')` - `grossSqFeet`
- [x] Estimated Cost of Construction - needs to support both numbers and text (e.g. "TBD"). if `is_numeric` we want to output `'$' . number_format($d, 2, '.', ',')`
- [x] Construction start date - data comes over as unix timestamp output as `format_date(($d / 1000), 'custom', 'F Y')` comes from `constructionStartDate` - it can be `null`
- [x] Substantial Completion Date - similar as above - comes from `substantialCompletionDate`
- [x] Pre Bid Date - similar to other date fields  - `preBidDate` - looks like only output if is greater than today
- [x] Pre Bid Location - assume text? - `preBidLocation` - looks like only output if pre bid date is greater than today
- [x] Bid Date - similar to other date fields - `bidOpeningDate` - looks like only output if pre bid date is greater than today
- [x] Awarded To - text - `vendorName`
- [x] Architect(s) - text - `primaryConsultant`
- [x] Building - reference field
- [x] Building alt - text field
https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=0660601 is an example where there is a building related to this project but the building doesn't exist currently on the Facilities website.

# Proposed Solution

- Create fields listed above.
- Build upon docroot/modules/custom/uiowa_facilities/src/FacilitiesAPI.php to establish calls to project endpoints
- Extend EntityProcessorBase in uiowa_core to move data into fields
- Disabling fields populated from API so it can't be manually added/edited

### LINKED ISSUE #7278: Theme Facilities Projects
# Story

As a Facilities site owner, I would like our projects to look nice.

# Additional Info

The majority of projects only contain info from the API so we should try to theme them in a way where they look good with just that info but also when an image and additional supporting info is added manually.

Current example with image and webcam:
https://www.facilities.uiowa.edu/project/0614902

Current example with just info from API:
https://www.facilities.uiowa.edu/project/0847001

# Proposed solution

- Mockup not needed
- Setup template in layout builder
- Style similar to existing site
- Make sure looks good if not all manual info added

**Field instructions:**
- [x] Pre Bid Date - only output if is greater than today.  
- [x] Pre Bid Location - only output if pre bid date is greater than today 
- [x] bidOpeningDate - looks like only output if pre bid date is greater than today 
- [x] Estimated Cost of Construction, if this field is empty output "TBD" text. 
- [ ] Title - `buildingName . ' ' . projectTitle` is currently being used on the project page on https://www.facilities.uiowa.edu/project/1062101 creating duplicate titles like "Parking Lots and Ramps Parking Lots and Ramps - 2024 Surface Lot Maintenance"
- [ ] Building page view outputs `buiProjectId - projectTitle`. 
- [ ] Capital and Featured views projects pages outputs `buildingName . ' - ' . projectTitle`

### LINKED ISSUE #7280: Output projects on building pages
# Story

As a Facilities site owner, I would like to display projects related to a building on that particular building page so that users can easily see what projects are happening with that building.

# Additional Info

Example of current building listing projects:
https://www.facilities.uiowa.edu/building/0204

Currently outputs `buiProjectId - projectTitle` as a link to the individual project

# Proposed solution

- views block with appropriate title info outputting as link to project node
- no paging necessary at this time (Fixed heading, display all items, hide block if empty)
- no additional theming necessary

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6447. 
Resolves https://github.com/uiowa/uiowa/issues/7278. 
Resolves #7280 

# How to test
```
 ddev composer install && ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local facilities_core:projects_import && ddev drush @facilities.local uli /all-projects
```
1. Code review
2. Should take around 3:30 to import the projects. 
3. Run `ddev drush @facilities.local facilities_core:projects_import` again and should take around 2:40. 
4. Confirm dates are being imported correctly, compare https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=0874301 to https://facilities.uiowa.ddev.site/project/john-pappajohn-business-building-upgrade-hvac-controls-system using https://www.epochconverter.com/. 
5. Confirm API call is bringing together all projects for import based on the list below. 
6. On `/all-projects` there should be 91 capital projects and 16 featured projects
7. Test a few buildings on `/all-projects` and compare against https://www.facilities.uiowa.edu/projects.  

1.   Grab all projects and data based on existing node ids from https://buildui.facilities.uiowa.edu/buildui/ext/projects?bldgnumber=0001 for example. 
2.   Grab all capital projects https://buildui.facilities.uiowa.edu/buildui/ext/capitalprojects
3.   Grab all featured projects https://buildui.facilities.uiowa.edu/buildui/ext/featuredprojects
4.   Finally, grab all additional project data from https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=0897601

## Outstanding issues?
- ~Projects without an entity reference for a building are not importing the dates for some reason. `Parking Lots and Ramps - 2024 Surface Lot Maintenance` is an example: https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=1062101.~
- https://facilities.uiowa.ddev.site/project/biomedical-research-support-facility-expand-facility imports even though https://buildui.facilities.uiowa.edu/buildui/ext/projectinfo?projnumber=0941301 is empty. 
- Do we need to find the total count of projects from the endpoint and compare that against what is imported?
- What determines the updated nodes when `ddev drush @facilities.local facilities_core:projects_import` is run again?  `Facilities projects content sync completed. 233 projects were created, 0 updated, 0 deleted, 0 skipped. That is neat.` is the first time message and ` 0 projects were created, 101 updated, 0 deleted, 132 skipped. That is neat.` the second time.

## Test themed page
- Project number 1062101 is the only one with a pre bid date greater than today
- Project number 0941301 estimated cost should output as "TBD" since the value is `0`.
- Add an image, webcam link, description, and what's new to an existing project. 
- Confirm two column layout has fields listed in the expected order https://www.facilities.uiowa.edu/project/0852601

## Test projects on building pages
- Confirm projects associated with a building appear on that building page.
- Example: https://facilities.uiowa.ddev.site/building/0204 should list several projects
- Check that the link to a given project works correctly.
- Confirm a building with no active projects does not list any. The Projects headline should also not display on these pages.
- Example: https://facilities.uiowa.ddev.site/building/0005


bspeare:
Moving this back to draft to fix a few issues.

bspeare:
This is ready for review again.  The main issue that I was working through has been resolved.  There is a minor "Outstanding issues" list in the testing instructions that the reviewer can comment on.

joewhitsitt:
planning to pull this down for testing.

joewhitsitt:
>there should be 91 capital projects

maybe the freshness but i got 92.

>Test exposed filter for buildings

Ideally this would be alphabetical

>What determines the updated nodes

`docroot/modules/custom/uiowa_core/src/EntityItemProcessorBase.php` and from stepping through a bit, it looks like project type, estimated cost are tripped up several times. perhaps we are processing that too early? i know we have to group results here, but classrooms for example has their altering logic in `processRecord` with the comment "Process the record before comparison." continuing to poke

joewhitsitt:
Maybe for project type since it is designating the content based on the endpoint instead of a value within the record we could pass an optional array of ignored properties. Downside is if a project was created and then reclassified (showing on a different endpoint) then it wouldn't get updated unless something else changed.

Screenshot 2024-01-26 at 3 21 01 PM

Update: I should test more but that alone gets the updated down to like 14.

joewhitsitt:
For estimated cost, even though the drupal field is a float and the record is cast as a float, they don't match up on comparison. Changing the field to a 15/2 decimal field seems to work better. Hopefully there aren't projects more than 9 trillion? 9,999,999,999,999.99

```
ddev drush @facilities.local facilities_core:projects_import                            
 [notice] Starting the facilities projects sync. This may take a little time if the information isn't cached.
 [notice] Facilities projects content sync completed. 232 projects were created, 0 updated, 0 deleted, 0 skipped. That is neat.
ddev drush @facilities.local facilities_core:projects_import
 [notice] Starting the facilities projects sync. This may take a little time if the information isn't cached.
 [notice] Facilities projects content sync completed. 0 projects were created, 0 updated, 0 deleted, 232 skipped. That is neat.
```

joewhitsitt:
Revisiting the projectType, it just seems the record format comparison to the entity value is different so it is always updated. Not sure how to get the record to look identical at this point in the process. Wonder if this has to do with keys and multiple select checkboxes

joewhitsitt:
might need a better set of eyes. from what i can tell it is the 100 project types that are single value (109 total in the db) that always trigger as different. i feel like i am missing something simple. `resolveFieldValuePropName` to me is right in that it is fetching value, but from what i can tell this is the first time we are handling an allowed values multiple values field.

joewhitsitt:
It seems to be related to changing projectType to a multivalue array. Writing a single string temporarily prevents a count of updated right after initial import. @briand44 had suggested yesterday to me that it might be easier to capture these as separate booleans (isFeatured, isCapital) and then not worry about single/multi/null. @bspeare would you be open to adjusting the content type and other pieces to that? I guess the filtering on the view would change from a single dropdown.

bspeare:
@joewhitsitt I'm open to it!

joewhitsitt:
>Do we need to find the total count of projects from the endpoint and compare that against what is imported?

204 from buildings (didn't actually look into this number)
16 featured
93 capital
313 projects grouped (contains duplicates from the multiple endpoints)
232 projects by id (unique) UPDATE: after fetching the latest commits till https://github.com/uiowa/uiowa/pull/7285/commits/09d305e4ce5f5e6a7b7aa8236d6b233cabefb2d7, there is 234 projects by id.

>What determines the updated nodes when

The recent changes seems to have fixed this and made the import quite a bit quicker. That one remaining one seems to be our blank project 0941301 `null != 0.00`

joewhitsitt:
**Suggestion**: Change project url alias to match current site? (e.g. /project/{projectid})

>Project number 1062101 is the only one with a pre bid date greater than today

Ugh. Should have tested this earlier (1/31). It isn't showing on the node... as expected, so thumbs up?

> https://github.com/uiowa/uiowa/issues/6447#issuecomment-1915030062

Confirmed by manually injecting a project that as long as it has a project number it is cleaned up on next import if it doesn't exist in the feed.

>Confirm two column layout has fields listed in the expected order https://www.facilities.uiowa.edu/project/0852601

Description displays under project number unlike their current site.

>Test projects on building pages

This all was as expected. Can't wait to see the finished building page with all of the things.

joewhitsitt:
Ben and I noticed the projects by building is a separate view config from projects. @e-marie-w would you be willing to add this to the projects view as a separate display?

e-marie-w:
> Ben and I noticed the projects by building is a separate view config from projects. @e-marie-w would you be willing to add this to the projects view as a separate display?

Sure!

joewhitsitt (approved):
After some breakthroughs this afternoon with quick shifts from @bspeare, I am liking what I see here. Would eventually like to sort out how our processor can properly handle multi-select fields.

Approving now since I am out the rest of the week, but I think https://github.com/uiowa/uiowa/pull/7285#issuecomment-1920030176 is the only thing left.

>Description displays under project number unlike their current site.

Not sure we this is desired or not

PR #7289, opened 2024-01-24
Facilities - deployment config error fix

joewhitsitt:
# Test

```
ddev drush @facilities.local sql-drop -y && ddev drush @facilities.local cr && ddev blt ds --site=facilities.uiowa.edu
```

No config import errors.


cory-skeers (approved):
Reviewed and tested; looks good!

PR #7293, opened 2024-01-24
[ITS] Alert status view

### LINKED ISSUE #7147: ITS Alert Status
# Story

As an ITS site owner, I would like to be able to display a status history of services over the last 5 days (current and four previous).

# Additional Info

Current page - https://its.uiowa.edu/alerts/status-history

Some notes - https://github.com/uiowa/uiowa/issues/7147#issuecomment-1832309215

- Display of all services
- Ability to see if service was operational, or if there was an outage, planned maintenance, service announcement, service degradation or multiple alert types over the last 5 days.
- Ability to filter/sort?
- Ability to view alert(s) details
- Legend depending on display?

# Proposed Solution

GaryRidgway:
This adds a view of all alerts in descending date order that has exposed filters. This should solve for a few different things that ITS desires, including alert history, finding alerts for updating them, and more.

Additionally, we identified that since we are using better exposed filters more and more, this PR adds better exposed filters to core.

# How to test

1. Pull, sync, and uli in to ITS
2. Create some alerts (using devel generate for this is helpful to create a lot of them)
3. Create a page and drop the "Alert status" view block on it.
4. Try to break it?
5. Does it meet the desires of the issue?


bspeare:
@GaryRidgway do you want to turn on "Enable auto-submit" for this?  I think that makes for a better search experience.

e-marie-w:
I'm having some issues with the building filter, when I hit reset it returns to the first building I set it to, instead of clearing it completely. 

I haven't been very involved with this work so I have some general understanding questions. If I understand correctly, this block will display every alert ever, in descending order of when the alert starts, correct?

e-marie-w:
> I'm having some issues with the building filter, when I hit reset it returns to the first building I set it to, instead of clearing it completely.

Good news this is now working correctly for me. Hopefully it was just a random local thing.

GaryRidgway:
> @GaryRidgway do you want to turn on "Enable auto-submit" for this? I think that makes for a better search experience.

Good idea Ben! 

Marie has offered to enable it, so it should be pushed up soon

e-marie-w:
Note about the auto-submit: It only works for the date filters, I think because they are single-select while the other filters allow for multiple selections. Please correct me if I'm wrong.

GaryRidgway:
[I created a follow up issue to remove BEF from all other splits](https://github.com/uiowa/uiowa/issues/7315).

e-marie-w (approved):
Successfully added block and can filter as expected. Sidenote: devel generate is my new best friend.

Issue #6578, opened 2023-06-26
Page list block with exposed filters


briand44:
# Story

As a web team, we would like to develop a page list block with exposed filters so that site editors have the ability to dynamically display a set of pages and can choose to allow site visitors to search by title and/or filter by tag.

# Additional Information

Epic - #7385 

Page list block similar to other existing list blocks.

Ability to expose a filter to search by title.

Ability to expose a filter to filter by tag? (Would like thoughts from the team here. Are there use cases where exposing the tag adds value? For Well Being, I see them using tags to tag pages as "Well Being Resources" but they likely wouldn't expose this, they would just expose the additional taxonomy references that we will add in the future.)

Display page title as a link to node and brief description in a list.

Display results in a grid would be nice as well but doesn't have to be part of this slice.

# Proposed Solution

PR #5951, opened 2022-11-15
Update ctools, introduce exposed filters to list blocks

pyrello:
Replaces #4479 and resolves #3804.

# How to test
```
 ddev composer install && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli people-directory
```
## Article and people list blocks
* View some existing article and people lists block and compare the settings with the production sandbox site to make sure that the update hook worked as expected.
* Place blocks on a page and try changing the following settings, saving the block, saving the layout, and viewing the page.
  * Test setting headline or not
  * Test changing items to display
  * Test changing offset
  * Test setting pager
  * Confirm that the 3 settings above work predictably with one another
  * Test changing sort options
  * Test hiding fields.
  * Test adding a more link, setting a custom path, and setting custom text
  * Test setting filter values
  * Test exposing filters without setting a value
  * Test exposing filters with a value set
  * Test setting multiple filters at the same time
  * Test setting multiple filters to different variations of exposing the filter and not
  * Test that exposed filter and pager for one block do not affect any other views blocks on the same page of either the same type or a different types
  * Confirm that "Items per page" and "Offset" do not show up in exposed filters

### People list block
* When setting the "Person type" filter, it should limit the options display in "Person type status" drop-down to the same option.

## Event list block
```
ddev drush @sandbox.local config-split:activate event -y && ddev drush @sandbox.local cim
```
* Create some local events.
* Run through the above tests.

PR #7304, opened 2024-01-29
Custom list block proof of concept

GaryRidgway:
This is a proof of concept that shows how we could offer options for custom views made in the off canvas tray that allows users to have extreme flexibility. It is build on listblock, and kicks butt.

Of course, the code is messy and probably not at all nice to look at, but I wanted to spike hard and polish little.

<details><summary>Here is a gif example of it working:</summary>
<p>

</p>
</details>


joewhitsitt:
I am curious to know the differences between this and the approach here: https://github.com/uiowa/uiowa/pull/5951

joewhitsitt:
Is this worth keeping around since https://github.com/uiowa/uiowa/issues/6578 ?

PR #7330, opened 2024-02-02
Fixed use strict in lazy load.

GaryRidgway:
use strict was being globally scoped, breaking other JS

## To test
1. before checking out this branch, Sync engineering
2. Go to `/me`
3. observe no secondary menu dropdown arrows
4. open console
5. observe js error for 'loadjs not declared'
6. checkout this branch
7. clear cache
8. observe the above is not true.


bspeare (approved):
Nice fix!

joewhitsitt:
I am interested in getting this out as it might solve other issues that were recently reported

PR #7331, opened 2024-02-02
[ITS] Alert updates

### LINKED ISSUE #7283: Ability to add updates to ITS alerts
# Story

As an ITS alert creator, I need the ability to provide updates to existing alerts so that those affected by the alert can stay up to date on any changes.

# Additional Info

- Need the ability to add one or more updates.
- Updates should have ability to choose a date and time (default to now) and provide filtered html through a text area.
- Update date/time and description should be displayed on alert node and themed.
- Ideally there would be revision history for alert updates.

# Proposed Solution

- Use paragraphs to create the date/time and Update text area. 
- Output update information in horizontal card format on alert node towards the top of alert node (above body). 
Our partners are open to what we come up with for displaying this information.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7283

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /outage-gartner-becker-communication-studies-building-testing-additional-title-details
```

- Add an alert update
- Test multiple updates
- Test repositioning updates
- Test date and time output


joewhitsitt:
going to pull and test

joewhitsitt (approved):
I do think disabling sort/weight and listing these chronologically would be a future slice. I did a quick glance at the issue queue but nothing seemed to turn up. Either I am searching for the wrong thing or nobody else seems to have our use-case.

Given that by default, newer items are added to the bottom and the latest update is then rendered at the bottom currently, to me it feels like the original alert description should then be above this in the node so that the node reads oldest to newest.

Screenshot 2024-02-09 at 9 53 51 AM

Otherwise I think this is working as expected.

PR #7337, opened 2024-02-06
[Pharmacy] State bug fix

### LINKED ISSUE #7336: State field not appearing when adding people on Pharmacy Alumni Directory
# Problem
The state field appears when adding a person of type Palliative Care Graduate but not when adding a person without a person type or with any other person type.

Pharmacy recently expanded from just a palliative care alumni directory (https://pharmacy.uiowa.edu/pharmd/dual_cert/palliative/certificate_graduates) to having an all alumni directory (https://pharmacy.uiowa.edu/alumni/alumni-directory).

It seems like we just need to remove the conditional logic for showing only when the palliative care type is chosen so the state field always is available. They only add people that are alumni because they are using Directory Profiles for all other person directories.

# Steps to reproduce

# Proposed solution

- remove the conditional logic for showing state field only when the palliative care type is chosen so the state field always is available.

bspeare:
Resolves  https://github.com/uiowa/uiowa/issues/7336

# How to test

```
ddev blt ds --site=pharmacy.uiowa.edu && ddev drush @pharmacy.local uli /node/5981/edit
```
- Confirm that State field is always available

### DIFF
--- a/config/sites/pharmacy.uiowa.edu/sitenow_people.person_type.palliative_grad.yml
+++ b/config/sites/pharmacy.uiowa.edu/sitenow_people.person_type.palliative_grad.yml
@@ -6,5 +6,4 @@ id: palliative_grad
 label: 'Palliative Care Graduate'
 description: ''
 allow_former: false
-allowed_fields:
-  - field_person_hometown
+allowed_fields: {  }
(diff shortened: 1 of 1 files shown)


joewhitsitt:
@briand44 should we make hometown visible to all types as well?

briand44:
>should we make hometown visible to all types as well?

Good question. I'm not sure if they plan to use that field for all alumni or not. It probably doesn't hurt to make it available otherwise we could reach out to Barb and Kayla and find out from them.

bspeare:
The hometown field is disabled on the form.  It has never been in use on Pharmacy.   I can remove it from allowed fields to avoid future confusion.

joewhitsitt (approved):
Sounds like potentially more cleanup for pharmacy hometown field down the road if they aren't using it. Reviewed, not tested.

PR #7359, opened 2024-02-13
[ITS] Alerts Calendar

### LINKED ISSUE #7234: calendar view of alerts
# Story

As an ITS site owner, I would like to have a calendar display of alerts so that users can quickly navigate the calendar to find alerts affecting particular dates.

# Additional Info

- Current calendar view of Alerts - https://its.uiowa.edu/alerts/calendar
- Displays all types of alerts
- Differentiate by category
- Include legend here
- Ability to view future and past alerts on calendar
- Would like improvements on mobile view of calendar

# Proposed Solution

cory-skeers:
Resolves #7234 
Relates to #7360 

# How to test

```
```

Edit an existing alert and/or add a new alert
Check that the Title field is not available

Edit a page and add the `Alerts calendar` block
Add some alerts of different types, check them out on the calendar, and see how it works!
Check both the `Month` and `List` displays
Load the page on a smaller screen, and see that it defaults to `List`
Load the page on a bigger screen, and see that it defaults to `Month`

Any suggestions on the 'Legend' part of this would be great as well

## Updates

The style adjustments have moved to uids_base, so a `blt frontend` will be needed

In addition to the calendar stuff specifically, check https://its.uiowa.ddev.site/alerts and see that the badges for the 4 types of alerts match the colors in the calendar (ie planned maintenance and service announcement badge colors have been updated)


joewhitsitt:
Arm-chair tweaker here but I wonder if we can just force override this style to be `normal` instead of `nowrap`.
Screenshot 2024-02-14 at 4 26 51 PM

Screenshot 2024-02-14 at 4 25 18 PM

cory-skeers:
> Arm-chair tweaker here but I wonder if we can just force override this style to be `normal` instead of `nowrap`.

That's the hope! I'm worried how it'll play with long titles of overlapping alerts, but going to check that out.

cory-skeers:
I'd gone with adding the calendar as a placeable block, with the goal of allowing the overall page it gets placed on still be editable as opposed to a full views page.

Just stumbled upon this--because the calendar is controlled by JS, if the block is placed by itself in a section without a heading it will be treated as an empty section and will not render. Looking into if there is a nice workaround for this.

joewhitsitt:
Perhaps add a header or footer to the view? We still need to include a legend somewhere correct?

cory-skeers:
> Arm-chair tweaker here but I wonder if we can just force override this style to be normal instead of nowrap.

Got this in place and tested with a few combinations of overlapping events, maxing out the number of events displayed on a day, etc., and everything seemed to work well. Moving forward with this small change.

> Perhaps add a header or footer to the view? We still need to include a legend somewhere correct?

Good idea. I had been starting to wonder if a legend was still necessary with the alert type in the title for every one, but maybe this can push that over into an absolutely yes, feeding two birds with one scone type of add.

cory-skeers:
I re-adjusted the colors used for this a bit, and would welcome feedback:
image

One issue I was running into is fullcalendar adds the background color directly as a style through JS, and so it bypasses some styling rules. In particular, all text ends up getting the white `color` treatment irrespective of the background it applies, so lighter colors (eg `primary`) don't work out-of-the-box. These colors seemed like a good mix of ones that are all workable with white text.

joewhitsitt:
pulling to test

joewhitsitt:
>Check that the Title field is not available

Any change to the entity form display brings it back. I guess we just need to be careful about this going forward.

Semi-related. Sorry, but I am pushing a small tweak to the alert updates to prevent a blank with date paragraph item being saved/created with each new alert. Now the user needs to press the add update button.

Screenshot 2024-02-15 at 12 02 21 PM

joewhitsitt:
For all of the displays here https://its.uiowa.ddev.site/admin/structure/views/view/alerts_list_block some are using the Views (List) category, some are using Site Custom. I think this calendar should be Site Custom so it is found with the other ones in layout builder.

joewhitsitt:
Not a biggie, but since they have the ability to change the alert categories, the legend would be one of the things hardcoded and requiring a config change to update. I glanced at the available tokens and didn't see a quick solution.

joewhitsitt:
I'll defer to others on the legend colors. They look fine to me.

cory-skeers:
> Any change to the entity form display brings it back. I guess we just need to be careful about this going forward.

Dang. I was able to make changes immediately after without it being brought back and thought we were good to go, but I guess not.

docroot/sites/its.uiowa.edu/modules/its_core/css/alerts-calendar.css
bspeare, line 1: Would this be useful to put in https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/views/view-calendar.scss instead?

cory-skeers: I probably should have looked for a place like this 😬 

That would be great. I'll work on shifting things around.

bspeare:
Maybe we could add a little more padding to the calendar?
```
.fc-day-grid-event {padding: .5rem!important;}
.fc-day-header {padding: .5rem!important;}
.fc-day-number {padding: .5rem!important;}
```

bspeare:
The colors look good to me.  Should we update the badges to match the calendar legend for Service Announcement and Planned Maintenance? Service Outage and Service Degradation currently match. 

Screenshot 2024-02-15 at 1 35 59 PM
Screenshot 2024-02-15 at 1 35 32 PM

briand44:
>The planned maintenance and service announcements look very similar. Outage and Service Degradation stand out because of their coloring and the icon. Can we change the look for either planned maintenance or service announcement so they differ?

https://github.com/uiowa/uiowa/issues/7360

cory-skeers:
I've updated the badges to match the calendar/legend:

image

The other possible addition could be icons for the two in addition to the color differentiation.

joewhitsitt (approved):
Thanks for making these changes

PR #7362, opened 2024-02-15
Remove ckeditor4 from composer

joewhitsitt:
We have switched to ckeditor5 in core, uninstalled the module from core.extension config, so this is the last step!

Passing build should be enough review.


GaryRidgway (approved):
IT'S BUILT DIFFERENT

PR #7366, opened 2024-02-16
Fix for buildings import

joewhitsitt:
Currently only building hours are being imported. If there are any new updates to other fields, those are being skipped over.

# To Test

If you haven't in a while, `ddev composer install && ddev blt ds --site=facilities.uiowa.edu`

On `main` set a breakpoint on \Drupal\uiowa_core\EntityItemProcessorBase::process and \Drupal\facilities_core\BuildingItemProcessor::process

Enable xdebug and run `ddev drush @facilities.local fm-buildings`

Note that only \Drupal\facilities_core\BuildingItemProcessor::process is triggered and hours are added to building nodes. Feel free to kill the import a couple nodes in.

Now checkout `fix_building_hours_import` and run fm-buildings again. Note that both run.

Referenced \Drupal\classrooms_core\RoomItemProcessor::process for this fix.


cory-skeers (approved):
Reviewed and tested, and both are firing properly now.

PR #7374, opened 2024-02-19
Added memory limit consistent with production

pyrello:
Sets the PHP memory limit for DDEV web server to 128, which is what our production web servers use.

# How to test

TBD


joewhitsitt (approved):
had something similar this AM for testing. makes sense to me

PR #7375, opened 2024-02-19
[ITS] Alert updates

### LINKED ISSUE #7360: ITS Feedback on alerts
**Alert node add/edit form**

- Can we remove the title field? It's a mandatory field, but it appears that it is overwritten with type of alert, service impacted, and building. If it's going to be overwritten, do we need to include it AND make it mandatory? title vs. additional title details. We don't currently have a title field on the alert template, only additional title details. If nothing is entered into additional title details, the title of the alert will programmatically be created based on type of alert, service and building selected. 
- Scheduling options will be new for editors, and we'll need to focus on training here to make sure folks know when to use it and how it works.  Can we modify the explainer text? Don't have it drafted yet, but I can put something together. Is that something I'll have access to do?
- Can we remove the Summary box since we're asking people to leave it blank? 

**Alert Status page** - https://its.prod.drupal.uiowa.edu/testing-new-alert-thing

- Can we move the Apply and Clear buttons below filters...both James and I thought it only applied to the search box
- Jamie commented that the search box was huge, but I don't hate it. 
- The planned maintenance and service announcements look very similar. Outage and Service Degradation stand out because of their coloring and the icon. Can we change the look for either planned maintenance or service announcement so they differ? 

**Alert Update display**

- Need the word "Update" on the display of alert updates somewhere...the clock icon doesn't scream "update"
- Can we get the list of updates to display in chronological order where most recent update is at the top? 

# Proposed Solution

- [x] title has been removed, just be careful to not reintroduce (config should be updated manually)
- [x] #7384
- [x] Make summary collapsed
- [x] offer sidebar option but we can't move the buttons with this design - similar to https://classrooms.uiowa.edu/spaces/find-university-classroom
- [x] Add "Update" heading to updates
- [x] get the list of updates to display in chronological order where most recent update is at the top - either through code or views

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7360. 

# How to test

```
ddev blt frontend && ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /testing-new-alert-thing
```
- Confirm sidebar filters are working
- Confirm summary is hidden on https://its.uiowa.ddev.site/node/add/alert
- Add some updates to an alert
  - Confirm that you are unable to drag to reorder the alerts
  - Confirm that they are sorted chronologically upon save, with the most recent on top on page display
  - Confirm that when re-editing the alert, the updates appear ordered


docroot/sites/its.uiowa.edu/modules/its_core/its_core.module
```diff
@@ -107,6 +116,23 @@ function its_core_entity_presave(EntityInterface $entity) {
...
+    // by the node presave.
+    // Disable the draghandles by disallowing the reference changes,
+    // and also remove the dragdrop mode from the dropdown actions.
+    $element['widget']['#allow_reference_changes'] = FALSE;
+    if (isset($element['widget']['header_actions']['dropdown_actions']['dragdrop_mode'])) {
+      unset($element['widget']['header_actions']['dropdown_actions']['dragdrop_mode']);
+    }
+  }
+}
+
```
cory-skeers: Would love another set of eyes/thoughts on this. It _works_ but is a bit hacky. It's based on https://www.drupal.org/project/paragraphs/issues/3036020#comment-12989072.

In the future (like if that issue moves forward), there may be an actual setting to disallow dragging instead.

joewhitsitt: Nice find. Seems like there are others with similar use-cases. I haven't pulled down and tested this, but if this does what it says, it doesn't seem too hacky at all and will be easy to switch to a config setting later on.

cory-skeers: Thanks! I added a `@todo` comment to point to the issue for possibly removing later assuming testing doesn't reveal any reason to avoid this.

cory-skeers:
May need to update "update help text on scheduling once they provide it and address weight issue" from #7360 in a followup issue, but otherwise I believe this is ready for review.

joewhitsitt:
pulling to test

joewhitsitt (approved):
As-is, this PR checks out, however, any ideas on how to prevent refocus to the date filter on submit? Having the date picker popup each time is kind of annoying as it covers the other filters.

PR #7382, opened 2024-02-20
Updated status command hooks implementations for Drush 12

### LINKED ISSUE #7380: BLT/Drush Commands - deprovisions, etc.
# Motivation

The deprovision and transfer commands currently call getApplicationFromDrushRemote which fetches a modified `status` drush command that includes the application name. There is code related to this customization here: \Drush\Commands\UiowaCommands::alterStatus and \Drush\Commands\UiowaCommands::initStatus

The application name is no longer being returned and the commands fail early on in the process.

Seems to be a unrelated issue for the reporting commands: https://github.com/uiowa/uiowa/issues/7354

# Proposed Solution

pyrello:
Resolves #7380 

# How to test

## First, test that this works as expected under normal local conditions:
```
ddev drush @default.local status
```
Output should not include a line with `Application      : `.
```
ddev drush @default.local status --fields application
```
There should be no output.
```
ddev drush @default.local status --field application
```
Should output a blank line.

## Next, test this as if we were in production:
Add the following snippet in `.ddev/config.yaml` after the `disable_settings_management: true` line:
```yaml
web_environment:
  - AH_SITE_GROUP=uiowa20
```
And run `ddev restart`
```
ddev drush @default.local status
```
Output should include `Application      : uiowa20`.
```
ddev drush @default.local status --fields application
```
Output should be
```
Application : uiowa20 
```
```
ddev drush @default.local status --field application
```
Output should be
```
uiowa20
```

## After you're done testing
```
git reset --hard && ddev restart
```
should restore your local environment to not reporting an application.

### DIFF
--- a/drush/Commands/UiowaCommands.php
+++ b/drush/Commands/UiowaCommands.php
@@ -55,16 +55,13 @@ class UiowaCommands extends DrushCommands implements SiteAliasManagerAwareInterf
    * @hook init core:status
    */
   public function initStatus(InputInterface $input, AnnotationData $annotationData) {
-    $fields = explode(',', $input->getOption('fields'));
-    $defaults = $annotationData->getList('default-fields');
+    $field_labels = $annotationData->getList('field-labels');
+    $field_labels['application'] = 'Application';
+    $annotationData->set('field-labels', $field_labels);
 
-    // If no specific fields were requested, add ours to the defaults.
-    if ($fields == $defaults) {
-      $annotationData->append('field-labels', "\n application: Application");
-      array_unshift($defaults, 'application');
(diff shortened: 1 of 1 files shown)


joewhitsitt:
Pulling to test. Thank you so much for taking a look at this.

joewhitsitt:
Oh shoot, I was so excited to realize that this might not yet be up for review. So far so good though. THanks!

pyrello:
> Oh shoot, I was so excited to realize that this might not yet be up for review. So far so good though. THanks!

It's ready for review. I was looking into some additional fixes for the report command, but that is not an easy fix and possibly not worth the effort.

joewhitsitt (approved):

PR #7389, opened 2024-02-21
Add warning to media delete with usage

### LINKED ISSUE #4810: Stop users from being able to delete media items that are in use
Users can currently delete media items that are referenced in other blocks (e.g Timeline referencing a deleted image). We should find a way to remove this as a possibility.

joewhitsitt:
Resolves #4810

Screenshot 2024-02-22 at 11 00 15 AM

# To Test

```
```

Go to https://sandbox.uiowa.ddev.site/media/406/delete and see the message and disabled delete button.

Go to https://sandbox.uiowa.ddev.site/media/536/delete and see that you can delete the media entity without warning.


joewhitsitt:
To ensure proper deletion we probably want to prevent deletion if there is usage, but in some cases it would require quite a bit of work of the editor to then be able to delete something especially if the file was heavily referenced. (creating new revisions, deleting past revisions). On the flip side, the file won't be released and deleted from the server until after it is naturally not referenced anymore (node revision delete).

So the one benefit of letting them delete with usage would be to rid it from the UI immediately and prevent further references from being created.

joewhitsitt:
After sleeping on it, given the support burden it creates when not properly deleted (file_managed, file_usage, tracking down revisions, ssh'n in and deleting the file), disabling the button is probably worth the customer inconvenience.

cory-skeers (approved):
Reviewed and tested, and looks good 👍 

Agreed that the inconvenience upfront is probably worth it overall.

joewhitsitt:
`drush entity:delete media {media id}` is a way to circumvent this feature if needed for stubborn/corrupt usage counts. If you go this route, you will want to confirm that the file_usage for the `fid` is lifted and the file in file_managed is marked for deletion (status=0).

Issue #7384, opened 2024-02-21
ITS Feedback on alerts part 2


cory-skeers:
- ~remove/hide/change perms for scheduling because they don't plan to use it.~
- ~Update alert url path to match existing site `/alert/[nid]`~
- ~Move category down on alert node display - they mentioned it seemed redundant since it was listed in title and the services affected should be listed higher.~

These ^^ have already been addressed in https://github.com/uiowa/uiowa/commit/a49862344a3de281a0f9c00c2102a6cae19e3222

- On the alert status page, they would like the alert categories filter to be a select list rather than autocomplete.
- any ideas on how to prevent refocus to the date filter on submit? Having the date picker popup each time is kind of annoying as it covers the other filters (sidebar filter on /testing-new-alert-thing)

# Proposed Solution
- Change the widget for the alert categories to a select list
    - There was some worry that this might not work, but until it doesn't, DO THE ABOVE

briand44:
Blocked until we receive help text for scheduling.

briand44:
No longer blocked because they said they don't need the ability to schedule alerts.

briand44:
Updated the issue summary here based on feedback from the meeting with ITS on 2/27/24

joewhitsitt:
I have some of these changes in https://github.com/uiowa/uiowa/pull/7397/commits/a49862344a3de281a0f9c00c2102a6cae19e3222

briand44:
Grooming conversation... Alan suggested that the order of the filters could be a factor in the date popup focus. Lowering them in the order also prevents them from covering other filters. Also could do something with JS. See if any related issues on D.O.

bspeare:
I created a separate issue for the focus issues at https://github.com/uiowa/uiowa/issues/7447.

PR #7397, opened 2024-02-23
ITS Alert Emails

### LINKED ISSUE #7350: Ability to send alert notifications via email on alert creation and update
# Story 

As an ITS alert creator, I would like to be able to send an email communication about an alert so that folks subscribed to our email list can be notified right away.

# Additional Info

**Current email example**

Image

**Email info**
From: IT Service Alerts (IT-Service-Alerts@uiowa.edu)
To: IT-Service-Alerts-Members@iowa.uiowa.edu
BCC: e7199078.iowa.onmicrosoft.com@amer.teams.ms
Subject: [Alert Title]

Alert Start Date (https://github.com/uiowa/uiowa/issues/7122)
Alert title linked to alert node
Alert Description

Related Links:
IT Service Alerts page - https://its.uiowa.edu/alerts
Calendar view of alerts - https://its.uiowa.edu/alerts/calendar

**Current example with update that is closed**

Screenshot 2024-02-14 at 4 34 03 PM

**Email info with update that is closed**

From: IT Service Alerts (IT-Service-Alerts@uiowa.edu)
To: IT-Service-Alerts-Members@iowa.uiowa.edu
BCC: e7199078.iowa.onmicrosoft.com@amer.teams.ms
Subject: "[Closed]" [Alert Title]

Alert Start Date (https://github.com/uiowa/uiowa/issues/7122)
Alert title linked to alert node

Update Info - latest first

Original Notice

Related Links:
IT Service Alerts page - https://its.uiowa.edu/alerts
Calendar view of alerts - https://its.uiowa.edu/alerts/calendar

**Future**

Daily digest email - Future Slice
Email logs?
Email notification to help desk?

# Proposed Solution

- Establish a boolean checkbox, "Send email on save" that is acted upon during node presave. The boolean is reset, but a email is triggered through the website with the alert node content similar to how it worked on the D7 site. See https://github.com/uiowa/uiowa/issues/7350#issuecomment-1942034243 for a more detailed description of the functionality.
- The email should be formatted similar, paying attention to the changes noted in https://github.com/uiowa/uiowa/issues/7350#issuecomment-1942129625
  - First send email and alert update email can be the same but with conditional logic to include closed information, updates/labelling the original alert descriptions.
- Keep in mind the future email digest slice as this functionality is developed.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7350
Relates to https://github.com/uiowa/uiowa/issues/7122
Relates to https://github.com/uiowa/uiowa/issues/7384

Screenshot 2024-02-26 at 12 06 43 PM

# Todo

- [x] Alert update dates also need to be "on brand"?
- [ ] Format revision log?
- [ ] Production/Debug refinement?
- [x] General cleanup
- [x] Restrict to role instead of moderation permission?
- [x] ~~Test multiple BCC addresses~~

# How to test

## Emails

- As a webmaster, see that the new field "Send alert email on save" is visible to you on alert add/edit.
- Create a new alert, click the checkbox and mark as published. Save.
- Run `ddev launch -m` which will bring up the local email client. See that an alert email is there.
- As an editor, see that the new field is not visible to you.
- Create an initial alert as an editor, save as draft. Note that no email is sent and you get an empty revision message.
- As a webmaster, edit the draft alert, add a revision message like, "approved" and check the send email checkbox with the moderation state set to published. Save.
- Run `ddev launch -m` which will bring up the local email client. See that an alert email is there.
- Inspect the email for all of the specifics here, https://github.com/uiowa/uiowa/issues/7350 , paying attention to the message details, related links, date format and that the original alert body does _not_ have a heading.
- Now go back into the alert and add one or more updates to the node. Again, check the box and save (assumes published). The resulting email should have those updates and the body should now have an original notice heading.
- Try testing with bulk publish. Alert nodes with the field checked should run through the same presave and properly send the email and reset the checkbox.
- Regarding the change to the compact date format. This should just be a slight visual change to users. I highly doubt we are doing any kind of functionality that would be brittle enough to trip up on the difference. Smart date was introduced to the default with the local events feature, but it is not clear that it is used. It is being used for alerts.

## Alert url path
- For the feature default, ITS, facilities the url pattern has been switched to /alert/[node:nid] as there is potential for duplicate titles.

## ITS Alert Category

- This has been moved down because it was duplicative of the node title that is constructed for ITS. Services and Buildings are more important.

## Disable ITS Alert Scheduling

- Removed scheduling from ITS alerts because they didn't think they would use it and didn't want to confuse users on limited use-cases.


joewhitsitt:
Based on https://github.com/uiowa/uiowa/issues/7122 and this work I talked to @briand44. I mentioned that string replace to achieve periods with the meridan appears to be the only way since PHP datetime does not support it. Additional the work to handle months besides abbreviating all seems like too much at this point.

For consistency as it relates to this content, I've removed the string replace merdian formatting and I am committing a ITS patched config version of the `Compact` format that adds a period after the abbreviated month and space between the time and the meridian. Baby steps towards the ideal output.

joewhitsitt:
Going to be looking for date feedback at EoD or before, otherwise, I think this is ready for review.

Update: based on EoD going to move the date format change to default and link up to #7122 as a minor step towards full implementation

joewhitsitt:
Hit a snag with link generation. It doesn't have the nid in time.. Taking a look.

joewhitsitt:
Also realized I need to output the services/buildings if multiple.

joewhitsitt:
Going to need to refactor a bit to move this to a helper function and call it during entity_insert and entity_update as the node id doesn't exist at the time the node runs through presave. This might have been why the button was not on the node form in d7. Bummer to realize this just now.

joewhitsitt:
@pyrello mentioned it might be worth trying to represent the checkbox using Form API instead of Field API since the field value is never retained. Going to look at that a bit. On the surface, the form api version doesn't seem to be retained for use in presave so some logic might have to move then to form submit for this to work.

joewhitsitt:
Immediately ran into an issue where an alert is saved as a draft but the email field is checked. This captures a use-case/scenario where alerts are bulk published or scheduled publish and the intent was to send an email at that time. Reverted the otherwise working experiment.

pyrello:
Pulling to test.

pyrello:
> Note that no email is sent and you get an empty revision message.

What is meant by "you get an empty revision message"?

pyrello:
> Try testing scheduling an alert or a bulk publish. Alert nodes with the field checked should run through the same presave and properly send the email and reset the checkbox.

Testing scheduling seems to conflict with this:

> Removed scheduling from ITS alerts because they didn't think they would use it and didn't want to confuse users on limited use-cases.

joewhitsitt:
> > that no email is sent and you get an empty revision message.
> 
> What is meant by "you get an empty revision message"?

If an email is not sent and no revision message is provided, the revisions tab won't display a message for that revision.

If an email is sent and/or a revision message is provided, the revisions tab will display a message for that revision.

joewhitsitt:
> > Try testing scheduling an alert or a bulk publish. Alert nodes with the field checked should run through the same presave and properly send the email and reset the checkbox.
> 
> Testing scheduling seems to conflict with this:
> 
> > Removed scheduling from ITS alerts because they didn't think they would use it and didn't want to confuse users on limited use-cases.

Yep. Forgot to update that test instruction. Testing with bulk publish should still apply.

pyrello (approved):
Works per test instructions. Doesn't send an email when alert is in draft. We are saved from a bunch of additional testing condition combinations by not having to worry about scheduled publishing.

PR #7400, opened 2024-02-23
Disabling security check for now until dompdf

bspeare:
# How to test


pyrello (approved):
Approved based on passing build.

PR #7402, opened 2024-02-23
Hcaptcha work

### LINKED ISSUE #7394: Implement CAPTCHA spam solution for forms on partner site
# Proposed solution

Build a solution adding a CAPTCHA on forms on a partner website who is experiencing issues with spam.
Work with TM/ISPO to resolve accessibility/security concerns.
Potentially transition the solution from a site split to a feature split.

e-marie-w:
Resolves #7394 

# How to test

- For testing locally, DM Marie for necessary settings file changes.
- `ddev composer install && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli /admin/structure/webform`
- You can create a new webform or click 'Build' on an existing one.
- Click Add Element, then find CAPTCHA under Advanced Elements and add it. 
- If current role is admin, check the Admin mode box to see the captcha as an admin and save. 
- View the webform, the hCaptcha should be appearing now. Trying submitting without filling it out, you should receive an error message. Then see if you can pass it. (Bonus question: Does the challenge level seem right or should it be adjusted?) 
- Confirm non-logged in users can see the hCaptcha on the page.
- Masquerade as a webmaster and confirm webmasters can also add hCaptchas to webforms. (Looks like webmaster is the only other role that can create/edit webforms at this time.)

# Next steps
- Enable this on the uiowa02 dev environment
- Send a link to TM for accessibility testing and ISPO for security testing


config/sites/admissions.uiowa.edu/captcha.captcha_point.contact_message_personal_form.yml
joewhitsitt, line 1: Are these leftover from the global "all forms" setting? This is part of what worries me with making this available to the platform. Can this be specified and split in config.ignore?

e-marie-w: Besides the two search-specific ones, the rest were created automatically when I first exported config for the site I think. I can try deleting them and seeing what happens. Also, I was going to briefly pull down a different site and turn it on locally there for some multi-form testing, I can double-check there when exactly those files are created.

e-marie-w: Just enabled the modules on a standard V3 site. I did not touch the global settings but 7 captcha points were created automatically, all disabled by default.

joewhitsitt: Then I think we should consider ignoring this config pattern similar to our config splits so they can be enabled/disabled in the active db and so we don't need to do additional config exports when a new form (e.g. content type) is created

joewhitsitt:
I think ignoring those in the default vs partial makes sense since we will most likely expand reach or pull it out completely

joewhitsitt:
pulling to test

joewhitsitt:
@e-marie-w did you ever encounter an issue where it was falling back to the math captcha? even with admin mode toggled, i am not seeing hcaptcha (logged in/anonymous)

e-marie-w:
> @e-marie-w did you ever encounter an issue where it was falling back to the math captcha? even with admin mode toggled, i am not seeing hcaptcha (logged in/anonymous)

Not until today when I was trying to set it up on sandbox. If it's happening on admissions too I guess it's not a default vs custom site thing, shoot. Maybe the recent main merge did something somehow. 

I'll put back into draft until it's figured out, and just won't resync admissions ahead of demo. 

Edit: nevermind the recent main merge, it was literally only site provisions.

joewhitsitt:
Figured it out. Assumed the settings file format. Correct format is:
```
$config['hcaptcha.settings']['site_key']
$config['hcaptcha.settings']['secret_key']
```

You DM'd me them as sitekey/secret

e-marie-w:
I completely forgot about that part, great catch!!!

joewhitsitt (approved):
Pushed a small change to track our usage as we determine if this is preventing spam but still allowing use.

I noticed that hCaptcha only has the visual challenge compared to reCaptcha which has an audio alternative. This puts more importance on the accessibility cookie.

I could see us opening it up to allow the math challenge as well, putting the decision more on the site owner about how restrictive they want to make their forms.

e-marie-w:
Putting back into draft so it isn't merged accidentally.

joewhitsitt:
>Putting back into draft so it isn't merged accidentally.

Wait, why don't we want to merge this?

PR #7404, opened 2024-02-26
Sprint 201 provisions

cory-skeers:
- Initialize tan.lab.uiowa.edu multisite on uiowa07


joewhitsitt:
Can you create a SSL issue for testiowa?

joewhitsitt (approved):

joewhitsitt:
Wait, was there any conversation on this `stuorg-muhs.org.uiowa.edu` just being `muhs.org.uiowa.edu`

cory-skeers:
> Wait, was there any conversation on this `stuorg-muhs.org.uiowa.edu` just being `muhs.org.uiowa.edu`

Good catch--I think my brain was so used to lab sites, I'd read it as something like 'stuart muhs.' Reached out to double check, and will wait until we can get that sorted/confirmed.

PR #7416, opened 2024-02-29
fix the weight warning on ITS alerts

joewhitsitt:
Solves the following warning on ITS alerts
>Undefined array key "weight" in Drupal\Core\Entity\Entity\EntityFormDisplay->processForm()

Without revealing the title field!

# Test

- On `main` `ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli node/add/alert` - see the warning message.
- On `its_alert_config_fix` `ddev drush @its.local cim && ddev drush @its.local cr` - don't see the warning message.
- Run `ddev drush @its.local config-split:export site` - there shouldn't be a config diff.

`url_redirects` was getting booted from weight 11 by simple sitemap to the point where it no longer had a weight defined. drupal core doesn't like this currently: https://www.drupal.org/project/drupal/issues/2662388 https://www.drupal.org/project/drupal/issues/2427201

weight 12 wasn't being used by anyone so instead we are kicking sitemap to 12 and everyone is happy.


GaryRidgway:
I can test this

GaryRidgway:
Is there a picture of what the warning would look like? I am not getting anything when i test this on main.

GaryRidgway (approved):
I didnt see the error on the add page, but i did on the edit page, and this PR makes it go away. Neat!

joewhitsitt:
Thanks @GaryRidgway

PR #7420, opened 2024-03-01
[ITS] alerts updates

### LINKED ISSUE #7384: ITS Feedback on alerts part 2

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7384

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /alerts/status-history
```


GaryRidgway:
closed in favor of https://github.com/uiowa/uiowa/pull/7429

PR #7422, opened 2024-03-01
ITS Alerts Digest Emails

### LINKED ISSUE #7373: Daily digest emails
# Story

As an ITS site owner, I would like an email digest of alerts to be sent to our subscribers daily so subscribers can be notified about any active alerts.

# Additional Info

Current example

Image

**Email info**

Sends daily at 7 a.m.

Subject, From and To can stay the same.

Format date and times according to #7122 

Output Planned Maintenance below the other alert types similar to how it is currently being handled and Ongoing Maintenance last.

Keep related links

# Proposed Solution 

- Construct a drush command similar to brand_core that triggers an email that can then be scheduled for 7 AM daily in PROD.
- Building off of #7350 by trying to use the same EmailBuilderBase but with a different subtype, the email details and output should largely be the same. Since alert node titles now combine services, make sure services and buildings are included if multiple.
- Make sure the drush command or the email writes to drupal watchdog so that we can trace whether problems are with Drupal or Teams.
- Account for the different buckets of alerts: active outages, service degradation and service announcements. Planned Maintenance - upcoming planned maintenance in next 7 days. Ongoing Maintenance - Planned maintenance that has already started but is not yet closed.
- Prevent sending to teams outside of PROD by using an environment check.

joewhitsitt:
Resolves #7373

Screenshot 2024-03-04 at 2 48 30 PM

# Summary

This PR established a alerts digest drush command that will be scheduled and triggered in PROD. The command reuses the code put together for https://github.com/uiowa/uiowa/pull/7397 so some of that has been adjusted to handle both types of emails.

# To Test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local its-digest && ddev launch -m
```

- You should see terminal output that the Alert Digest was sent.
- You should see an email with the from information set correctly. The to value should be using the site email address since the env is not PROD.
- Subject line should match the current website's email: `IT Service Alerts Daily Digest`
- You should then see an output of alerts grouped by type (Alerts (Outages and Degradations), Planned Maintenance, Service Announcements, Ongoing Maintenance) depending on if content for those types exist.
- The email should not be sent if there are no alerts that meet the above types' criteria.
- The alerts should contain the date, linked title, updates if they exist and the description.
- At the bottom of the email should be related links.
- Enable `dblog` and trigger the drush command again. You should see a message related to the command here: https://its.uiowa.ddev.site/admin/reports/dblog
- Make sure the individual alert email is still working by triggering one of the alerts to send an email on save. The output should match.
- Special notes about dates - "On the hour" time representations should have a space between the value and the meridian (am/pm). Also, the email should include the end date if not ongoing (e.g. planned maintenance, closed alerts)


joewhitsitt:
Wondering if @cory-skeers can take a look at this ongoing filtering to see if I got it right based on the new way of handling closed.

V3:
Screenshot 2024-03-04 at 10 00 48 AM

V1:
Screenshot 2024-03-04 at 10 02 09 AM

cory-skeers:
> Wondering if @cory-skeers can take a look at this ongoing filtering to see if I got it right based on the new way of handling closed.

That looks right to me. The start date is prior to now, so the maintenance has or had started, and then either there is an end date and it is in the future (so the planned end time has not passed) or the duration is 0 so it's considered ongoing, and should show up until and end date gets added (when it would then fall back to the end date being before or after today logic).

^ Thought it would be easier for me to just write through the logic again, and so even more set that this is correct.

joewhitsitt:
I appreciate you taking a look. My Monday brain is off to a slow start.

pyrello:
Pulling to test.

pyrello (changes requested):
I'm not sure what the exact issue is, but there seems to be a disconnect (possibly caching) between the output of the email and what is shown via the Outages and Degradations block. Every digest email I received looked exactly like the one in the screenshot in the IS. Below is the current version of the `/alerts` page which contains my new alert.

Screenshot 2024-03-04 at 3 59 37 PM

joewhitsitt:
Oh I see. I think I might be overwriting the same alert per group where it is only grabbing the last? I'll pull down and check it.

joewhitsitt:
I believe https://github.com/uiowa/uiowa/pull/7422#pullrequestreview-1915440535 has been fixed by https://github.com/uiowa/uiowa/pull/7422/commits/ff0b2e9cd32efb2822fc118c00d6ff7fd4818435

pyrello (approved):
This appears to be working as advertised now. I was able to make changes to alerts and see those changes reflected in the digest. Existing email functionality continues to work. Nice job!

PR #7429, opened 2024-03-04
[ ITS ] Alerts filter focus

### LINKED ISSUE #7447: Better exposed filters focus
# Problem
Certain filter combos cause the focus order to mess up, see https://iowaweb.slack.com/archives/C0164FNN1PV/p1709566223488189 for more discussion. 

# Proposed Solution

- Investigate if the ordering of the filters changes what is focused when you submit the filters. Based on that, we could proceed in one of the following ways:
    - If the ordering change works fine enough, neato.
    - Otherwise, potentially look up issues on drupal.org for things that might be tied to this.
    - Worst case scenario, we can always manually change the focus on submit with Javascript.

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/7447

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /alerts/status-history
```


GaryRidgway:
closing in favor of https://github.com/uiowa/uiowa/pull/7465

PR #7201, opened 2023-12-12
Adding block_margin_default_removed style to all blocks

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/4917. 
Resolves https://github.com/uiowa/uiowa/issues/6341. 
Resolves https://github.com/uiowa/uiowa/issues/7245. 

# How to test

## How to test adding block_margin_default_removed style to all blocks

```
ddev blt ds --site=facilities.uiowa.edu  && ddev drush @facilities.local uli /node/3636
```
- Confirm that margin is removed for nationality and birth year
- Edit a content type and confirm that this style is available on all blocks
- Edit a page and confirm that this style is available on all blocks

## How to test the allowed list on layout_builder_custom_preprocess_block 'base_plugin_id'

-  Test these views blocks and confirm that you can set a background color style and grid style on the cards. 
   - 'views_block:article_list_block-list_article',
   - 'views_block:events-card_list',
   - 'views_block:events_list_block-card_list',
   - 'views_block:people_list_block-list_card',
     - Can you think of any other list blocks that use this functionality?  I did not find any.

PR #7431, opened 2024-03-04
Page listblock

### LINKED ISSUE #6578: Page list block with exposed filters

pyrello:
Resolves #6578 

# How to test

```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli
```
- Add a new page or use an existing page. Add a page list blocks (Add block > More > Pages) to the layout.
- Run through the test scenarios below with one page list block.
- Add a second page list block and run through the scenarios again with two list blocks on the page, ensuring that neither block is affected by other one.
- Add one of the existing list blocks (article or people) to the page and test again, ensuring that the existing list block is not affected by the page list blocks.
- Independently test the article or people list block on a separate page to ensure that there are no regressions.

## Test scenarios
- Setting one filter value.
- Exposing one filter with no value set.
- Exposing one filter with a value set.
- Exposing one filter with no value set and setting the value for the other filter but not exposing it.
- Exposing both filters with no values set.
- Exposing both filters and one filter has a value set.
- Exposing both filter with values set.
- Combine the above settings with a pager.
- Combine the above settings with different sort settings.
- Test different combinations of LB styles.


joewhitsitt:
Would it be possible to rename the machine name of this views display?

Background/reasoning: https://github.com/uiowa/uiowa/pull/7201#pullrequestreview-1804656602

more specifically avoiding `block_1` and instead start moving to something like `list_{type}`

pyrello:
> Would it be possible to rename the machine name of this views display?
> 
> Background/reasoning: [#7201 (review)](https://github.com/uiowa/uiowa/pull/7201#pullrequestreview-1804656602)
> 
> more specifically avoiding `block_1` and instead start moving to something like `list_{type}`

Not opposed to this, but I guess I'm not sure I'm following the reasoning behind it.

joewhitsitt:
For possible reasons in the future, moving towards a more consistent naming structure for like-things will be targetable so we don't have to do so much allowlist type stuff.

bspeare:
I'm getting a duplicate page title after I add the block to the page and save.  Do you get this?

Screenshot 2024-03-04 at 4 27 27 PM

pyrello:
> I'm getting a duplicate page title after I add the block to the page and save. Do you get this?
> 
> Screenshot 2024-03-04 at 4 27 27 PM

Hmm... I saw that, but didn't pay attention to it. I wonder if it is an artifact of adding the Page entity bundle class? Not sure what else would be causing it.

pyrello:
> I'm getting a duplicate page title after I add the block to the page and save. Do you get this?

This seems to be related to the page that the list block is being placed on showing up in the result. I was able to recreate it in the `main` branch by modifying the article list block to allow pages. Then I added a new tag with no content in it and limited the list block to that tag. The extra title disappeared. I added one other piece of content (not the testing page) to the tag and refreshed the testing page and there was still no extra title. Finally, I added the testing page to the tag and then refreshed it and the duplicate title showed up.

Ideally, I think we would always (?) want to exclude the current page from being listed in a block on itself. Let's come out against page-ception ;)

pyrello:
> Ideally, I think we would always (?) want to exclude the current page from being listed in a block on itself. Let's come out against page-ception ;)

I think I fixed this with a contextual filter that pulls the current node ID from the URL and excludes it from the result.

joewhitsitt:
Ran into a multiple tags causing duplicate results. Pushing a fix similar to https://github.com/uiowa/uiowa/pull/4406

I think tag is all that needs it for now.

joewhitsitt:
Should we adjust the placeholder text explaining that we can't preview the result similar to articles/people?
Screenshot 2024-03-05 at 11 38 38 AM

pyrello:
> Should we adjust the placeholder text explaining that we can't preview the result similar to articles/people?

Do you mean something like this? https://github.com/uiowa/uiowa/blob/1a4c4941e9b2f1b563d65292ca93d5838bd6cc87/docroot/modules/custom/layout_builder_custom/src/LayoutBuilderPreRender.php#L57-L65

joewhitsitt:
Yeah, just thinking about avoiding a support ticket, "Why is articles list block showing a preview and page list block isn't?"

Is this just because of the exposed filters or did we miss a setting somewhere? Will this be the same if the other blocks are converted?

joewhitsitt (approved):
I think this is looking good. I didn't experience any major bumps. The contextual filter helps! I do think the preview in this case could be more clear compared to the others.

pyrello:
> Is this just because of the exposed filters or did we miss a setting somewhere? Will this be the same if the other blocks are converted?

I believe this is happening because of the contextual filter. It is a bit of a mystery to me why some blocks show a placeholder vs not. We encountered this with the menu block where sometimes it would show and sometimes not.

pyrello:
This is ready for a quick re-test. Only thing that needs to be tested is that the default of 10 items displayed is set for all 3 list block types.

cory-skeers:
> This is ready for a quick re-test. Only thing that needs to be tested is that the default of 10 items displayed is set for all 3 list block types.

I pulled this and can confirm that the default of 10 is set for each of them. 👍

PR #7448, opened 2024-03-07
[ITS] Support article list

### LINKED ISSUE #7418: display support articles on services and theme
# Story

As an ITS site owner, I would like support articles about services to be displayed on the related service page so that visitors can quickly get support information related to the service they are interested in.

# Additional Info

Mockup - https://sandbox.prod.drupal.uiowa.edu/its-serviceinfocenter-mockup-final-copy

Group support articles by category

Display getting started category first

May have some services with a lot of support articles

# Proposed Solution

- Using taxonomy term weight set by webmasters (e.g. getting started first), create a views block that is contextually filtered by the service it is placed on that groups support articles with reference to the service based on the support category they are designated with then chronological for now.
- Display all in a multi-column output similar to the mockup.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7418

# How to test

```
ddev blt frontend && ddev blt ds --site=its.uiowa.edu && ddev drush @its.local en devel_generate && ddev drush @its.local devel-generate:content 20 --bundles support_article && ddev drush @its.local uli /services/gartner
```
- Confirm articles appear in columns like https://mentalhealth.uiowa.edu/help
- Confirm articles only appear on service pages that are tagged from support article
- Confirm taxonomy weight adjusts order of article headings

This uses our `uids_base/related` library, so any changes will need to go there and be tested against the `topic_page` feature.


cory-skeers:
I don't think it's an issue, but wanted to note it because I came across it:

Support articles can end up with duplicate services and/or categories listed, and on the service page it will appear once per connection combination (eg two instances of the service and two instances of a category will end up with 4 results. Two instances of the service, but no duplicated category will end up with 2 results, etc.)

image

I don't think anything needs to be done, but especially when content is put into place initially, I could see this happening by accident and leading to some confusion. The fix is easy enough.

cory-skeers (approved):
Looks great, and the arrangement by taxonomy weight works really well.

👍

PR #7455, opened 2024-03-08
[Wellbeing] Site split and vocabularies

### LINKED ISSUE #7442: Well Being taxonomies and filters for page content type
# Story

As an owner of the well being website, we would like the ability to categorize and filter pages by audience, type and topic so that our users can quickly find the well being resource appropriate for them.

# Additional Info

https://github.com/uiowa/uiowa/issues/7385#issuecomment-1981688974

Discussed creating a site split for this work.

Need the following vocabularies associated with the page content type in addition to the existing tags taxonomy:

- Audience - multiple
- Type - multiple
- Topic - multiple

Webmaster should be able to add/edit terms to these taxonomies.

They need the ability to exposed filters for audience, type and topic.

# Proposed solution
```[tasklist]
### Tasks
- [x] Add site split for Well-Being site.
- [x] Add vocabularies defined above. Vocabularies will be completely split.
- [x] Add entity reference fields to the page content type referencing each vocabulary based on specification above. Fields and page content type will be completely split.
- [x] Update the page list block view to include the new ER fields as exposed filters. View will be added to the split as partially split.
- [x] permissions for taxonomies
- [x] don't output taxonomy terms on pages, rabbit hole and wait for feedback
- [x] determine url pattern for these term pages in case we ever link to them
```

cory-skeers:
Resolves #7442 

# How to test

```
ddev blt ds --site=wellbeing.uiowa.edu && ddev drush @wellbeing.local config:import --source ../config/sites/wellbeing.uiowa.edu --partial && ddev drush @wellbeing.local cim && ddev drush @wellbeing.local cr && ddev drush @wellbeing.local uli
```

As a webmaster, add some taxonomy terms to Audience, Topic, and Type vocabularies; other roles should not be able to add/remove these terms. (https://wellbeing.uiowa.ddev.site/help can be used as a reference to what they've done with Tags so far, and a good place to get example pages to work with)
Try visiting a term page while logged in and logged out; logged out, you should get an access denied
Edit some Pages and use some of the newly added terms
On a page, add a Pages List Block, and expose a combo of filters. The new Audience, Topic, and Type vocabularies should be available as filters
Exposed filters for the new vocabs should work

Check #7442 tasklist and that everything is covered


cory-skeers:
Bit of feedback: Can we move the fields into the Relationships section alongside Tags?

e-marie-w:
One question: because of the rabbithole redirect, even webmasters should not be able to view a taxonomy term, correct? Because right now I see the access denied message when logged out and when logged in as any role but admin.

Assuming that's correct everything is working great and as expected. I'm going to work on moving the fields into the Relationships section now.

cory-skeers:
> One question: because of the rabbithole redirect, even webmasters should not be able to view a taxonomy term, correct? Because right now I see the access denied message when logged out and when logged in as any role but admin.

Good catch! Webmasters can bypass the content rabbit hole, and it might be useful to be able to use as an administrative tool, since they're allowed/intended to manage the taxonomies.

I've updated webmaster permissions, but other roles and anonymous should still be rabbit holed.

> Assuming that's correct everything is working great and as expected. I'm going to work on moving the fields into the Relationships section now.

Thanks for tackling that part!

e-marie-w:
Okay, so the Relationships sidebar section is actually set in sitenow.profile, and as far as I can tell there isn't a way to add to that via the UI. I believe we could do it in code via a hook_form_alter, but at this point we do not have a wellbeing_core module set up. I can set it up, if we want to do this. Another option with a smaller footprint would be adding a new fieldgroup to the sidebar and labelling it 'Additional Relationships' or something. 

I don't have strong feelings at this time so I'm looking for some feedback before I go too far into one direction. Setting up the module makes sense if further custom work is planned, but I'm not seeing any other open issues at this time. Adding a second fieldgroup is a little redundant but would only alter existing files.

joewhitsitt:
I'd vote to follow our existing pattern and creating wellbeing_core for this. If this was to move to a feature split or default in the future, this would be how it is handled.

docroot/sites/wellbeing.uiowa.edu/modules/wellbeing_core/wellbeing_core.module
```diff
@@ -0,0 +1,20 @@
...
+ * Primary module hooks for wellbeing_core module.
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function wellbeing_core_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if ($form_id === 'node_page_edit_form') {
```
joewhitsitt: Is it because you are only targeting the edit form and not the add form for page? `node_page_form`

e-marie-w: Turns out it was actually because I forgot to enable the wellbeing_core module and add it to the site split. But this change was necessary too, thanks for catching it!

e-marie-w:
New taxonomies successfully grouped with existing now: 
Screenshot 2024-03-15 at 10 43 39 AM
Edit: I'm going to make an issue for the tags expanding outside of the container, I checked and it's happening on prod sites too.

e-marie-w (approved):
Everything is working as expected. Taxonomy terms can only be viewed and added/deleted by webmasters and exposed on the pages list block. Terms can be easily added to pages and used to filter by on the pages list block. I think the default URL pattern for the taxonomies is fine. 

Great job @cory-skeers!

PR #7457, opened 2024-03-11
[ITS] Allow alerts to be replicated

### LINKED ISSUE #7433: replicate functionality for Alerts
# Story

As an ITS site owner, I would like to have the ability to replicate alerts so it is easier for service owners and HD staff who have regular planned maintenance alerts to create.

# Proposed Solution

Add alerts content type to the replicate allowed list in default config: https://github.com/uiowa/uiowa/blob/main/config/default/uiowa_core.settings.yml
Make sure things work!

cory-skeers:
Resolves #7433 

# How to test

```
```

Replicate an alert, see that it works

Try syncing another site and make sure it syncs cleanly. If you sync facilities, you should be able to replicate alerts


joewhitsitt:
testing quick

joewhitsitt (approved):
The one thing I noticed is I suspect that with the ITS alert title construction, it doesn't get the ` (Copy)` at the end. Also, replicating Facilities' alerts makes it published right away.

Do we care about either of these things? If so, I think they can be follow-ups.

Also tested new install that enables the split and existing with split activated. Both worked due to install/update config.

cory-skeers:
> The one thing I noticed is I suspect that with the ITS alert title construction, it doesn't get the ` (Copy)` at the end. Also, replicating Facilities' alerts makes it published right away.
> 
> Do we care about either of these things? If so, I think they can be follow-ups.

For ITS, maybe we can disable (but still show) that field when replicating, keeping in line with how they're unable to edit the title on the node. I think it not getting the ` (Copy)` is probably how it should work, but showing that when it won't actually be the title, and allowing the title to be edited only to be overwritten could be a better experience.

For Facilities, odd...Not sure what the interaction is there, but with their having scheduling in place, I think there's a good chance they'd want to replicate into an unpublished state, so that should probably be looked into.

Thanks @joewhitsitt! I'll create a followup issue for these.

Issue #6961, opened 2023-10-04
Exposed filter views with autosubmit delay is causing input issues for users.


joewhitsitt:
# Story

As a user of the classrooms room finder, when I type in the https://classrooms.uiowa.edu/spaces/find-university-classroom search box, the view autosubmits before I finish typing.

Reported by Makur Jain and Anne Schimke

https://github.com/uiowa/uiowa/assets/4663676/563697a8-9347-4643-8d77-67e1cde9599d

# Proposed solution

- [ ] Increase delay on autosubmit?
- [ ] Add some more logic to determine if user has finished typing. I think Alan did some cool stuff with the brand lockup generator.
- [ ] Disable autosubmit for mobile?
- [ ] Disable autosubmit entirely?

PR #7465, opened 2024-03-13
[ITS] Filter focus fix #3

### LINKED ISSUE #3: Update core patch
The patch we are using to set site configuration values from the install profile no longer applies against 8.5.x. A re-roll of the patch is available at https://www.drupal.org/project/drupal/issues/2849074#comment-12546690.

### LINKED ISSUE #7447: Better exposed filters focus

GaryRidgway:
Resolves https://github.com/uiowa/uiowa/issues/7447

# How to test

```
ddev composer install && ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /alerts/status-history
```

1. Check to see if the form works okay after autosubmit is turned off.


joewhitsitt:
pulling to test

joewhitsitt:
I think this mix of autosubmit and non-autosubmit with this patch is now causing problems. I am sure we have other views like this as well. As a keyboard user I get booted back to the first filter.

I entered a start date that didn't return any results. Cleared it out and tabbed to the end date. It autosubmits with the end date picker displayed but then changes my input focus back to the title filter.

https://github.com/uiowa/uiowa/assets/4663676/14eed5c5-b0eb-4fb2-9bdc-3bc0d74c3bfc

joewhitsitt:
If this patch was just for this view, I'd say move forward but with autosubmit off, but since we have autosubmit almost everywhere now, I can see how this can cause issues. [We already have issues with autosubmit](https://github.com/uiowa/uiowa/issues/6961).

bspeare:
I tested this out and I do like that this change removes the focus from the date picker after a category is selected!   Fixing that is a pretty big improvement for usability. 

There is a quirk, that moves the focus to the building or service field based on a few interactions:
- If a building is selected and at least one category is selected and you use the search field, the focus goes down to the building field.  This can be an issue if you are slow typing out your search terms and it will automatically move you to the building focus (see details below for screen capture). 
- If a building is selected, and you select two categories, the focus goes up to the building field. 
- If a service and building is selected, and you select one category, the focus goes up to the service field. 


<details>

https://github.com/uiowa/uiowa/assets/1036433/6af6bd2c-28f4-4b41-8e5b-e2fda78e5230

</details>

GaryRidgway:
> 
> There is a quirk...
> 

Ugh. This is seriously annoying. There is not any support for those type of fields currently, so in addition to the datefields technically not being supported, any focus that those autocompletes get is a total wash right now too. Perhaps this issue transitions in to a future upgrade for BEF and we just disable autosubmit for now? It feels odd to try and completely refactor and fix all the issues with BEF right now just to solve this one datepicker popup issue when we really didn't scope for more than that.

Thanks for the comments Y'all. I think I will bring this back up to the team in PL.

pyrello:
Pulled to test.

pyrello (approved):
This isn't the improvement we were hoping for, but it'll do for now. Thanks for your work on this @GaryRidgway!

PR #7469, opened 2024-03-14
remove apr from codebase

### LINKED ISSUE #6216: Audit and Remove APR module if no usage
# Story

As a developer, I would like to remove the APR module for D9 since we don't believe it is in use anymore so that we can reduce our technical debt.

# Additional Info

I believe Pharmacy was the only site using this for D9 and I believe they have since transitioned to Directory Profiles.

# Proposed solution

- [x] Audit to see if there is any usage
- [x] Remove if there is no longer any usage

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6216

PO gave JDI approval on this issue a while back as this module is no longer used and is technical debt otherwise.

# To Test
- Inspect that this removal is largely the reverse of https://github.com/uiowa/uiowa/pull/2725/files
- Confirm no import errors `ddev blt ds --site=pharmacy.uiowa.edu` which was the only known website to use the module.


pyrello (approved):
Nice! Thanks for the DX improvement!

PR #7488, opened 2024-03-20
Add authored by filter to content admin page

### LINKED ISSUE #7474: Filter by author on content overview
# Story

As a SiteNow user with access to the content overview, I would like the ability to filter by author so that I can quickly find a particular piece of content I authored or was authored by a particular person.

# Additional Info

Requested by ITS but this seems like something that could be useful for the service.

# Proposed Solution

- take Seans work and see if we can address display possibly with patch to Claro

pyrello:
Resolves #7474 

# How to test

```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli admin/content
```
- Ensure that filter works as expected.
- Ensure that other filters work without regression.
- Test the combination of the new filter and existing filters.
- Confirm tags and author fields are shorter in length and match appearance on https://github.com/uiowa/uiowa/issues/7474#issuecomment-2015871637.


cory-skeers:
For sites that have site-specific or feature-specific content types, it's possible that the spacing gets thrown off a bit.

Wellbeing.uiowa.edu has the topic collection feature, so its "content type" filter is larger to accommodate the "Topic Collection" content type, and the Filter and Reset buttons get pushed down below after a filter selection is made (when it's only a Filter button, it is still on the same line).

image

I don't think it's a problem, but @bspeare have any ideas on elegant fix for this case? (admissions is another good example to test with)

bspeare:
@cory-skeers everything will gradually move to the next line as the browser collapses.   I don't see an elegant solution beyond shortening the sizes of the fields.  Maybe we can run it by everyone at EoD?

Maybe moving to `size=15` would tighten up the space a bit more?

cory-skeers (approved):
The recent updates are working great! They were some small details, but a really nice UX improvement. Thanks!

bspeare:
Thanks @cory-skeers!

PR #7492, opened 2024-03-21
Remove core project message dependency

pyrello:
This dependency is just for displaying a message to users when they first setup Drupal. We don't need that.

# How to test
```
ddev composer install
```
- No composer install issues.
- Code review.


pyrello:
Follow up issue: Remove the line related to this dependency in composer.json that was left in to avoid composer complaining.

joewhitsitt:
checking this out quick

joewhitsitt (approved):
Shaving milliseconds off our build times!

PR #7499, opened 2024-03-22
ITS Permissions Update

### LINKED ISSUE #7456: Roles/Permissions for ITS
# Motivation

ITS has said they are ok with moving forward with the existing roles and would utilize the editor role for folks like service owners and help desk students.

They would also like editors to have the ability to create and publish support articles.

My understanding is that in order for editors to publish support articles we either need to remove the default moderation workflow from support articles or create a new workflow.

Adrienne said we can remove moderation from support articles because it would become a huge bottleneck.

# Proposed Solution

- [x] Remove the default workflow moderation from support articles
- [x] Update permissions for support articles so editors can create/edit/publish
- [x] Provide summary of what roles on existing site would transition to which default roles on the new site

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7456

# To Test

```
```

- Create and masquerade as an editor on the website.
- You should be able to create/edit/delete support articles and toggle the published state (unpublished by default).
- Because categories was switched to radios, we don't want a N/A choice so I made it required.
- Role mapping recommendations doc: https://github.com/uiowa/uiowa/issues/7456#issuecomment-2015891103


bspeare:
I can test this.

bspeare (approved):
Roles work as described. Nice work!

Issue #4452, opened 2021-11-17
Add config split support for Layout Builder templates


pyrello:
# Motivation

As part of #4091 we had to settle for something less than completely moving away from keeping complete splits of overridden config, specifically with regard to `core.entity_view_display.node.*.default` config where Layout Builder is turned on and used for templating. Any split that includes alterations to the Layout Builder template for a content type (such as adding a field or other block) must split that config completely since there is no way to tell sections apart per https://www.drupal.org/project/drupal/issues/3208766

https://www.drupal.org/project/config_split/issues/3238855 was added to config split to allow more intelligent handling of objects based on UUID.

# Proposed solution

* [ ] Test the patch/MR at https://www.drupal.org/project/drupal/issues/3208766. Specifically pay attention to:
  * [ ] Ensure that it works with config split and we can use it to replace our complete splits of `core.entity_view_display.node.*.default` with partial splits.
  * [ ] Ensure that the update hook included with the patch catches all existing layout sections in the database and updates them to include UUIDs. 
* [ ] Update the patch if testing shows its necessary.
* [ ] Update/re-export all relevant config.

pyrello:
Related: #7119

PR #7518, opened 2024-03-28
[Wellbeing] Remove display of new taxonomy terms

cory-skeers:
The new taxonomy terms (Audience, Topic, Type) ended up getting included in the page display.

It's been confirmed that these are not to be displayed, and should be for back-end use only.

# How to test

```
ddev blt ds --site=wellbeing.uiowa.edu && ddev drush @wellbeing.local uli node/add/page
```

Check for a clean sync and config import.
Add some terms (audience, topic, and type) and save the page
See that they are not displayed on the page


joewhitsitt:
pulling to test

joewhitsitt:
Aren't we supposed to complete split `core.entity_view_display.node.*.default` or has there been progress there in terms of handling config diffs? Ping @pyrello

https://iowaweb.slack.com/archives/C0164FNN1PV/p1642689940252600 , https://github.com/uiowa/uiowa/issues/4452

pyrello:
> Aren't we supposed to complete split `core.entity_view_display.node.*.default` or has there been progress there in terms of handling config diffs? Ping @pyrello
> 
> https://iowaweb.slack.com/archives/C0164FNN1PV/p1642689940252600 , #4452

If the difference doesn't affect the LB template, then its okay to partially split this config. If the LB template is being altered at all, then it has to be completely split.

cory-skeers:
> Aren't we supposed to complete split `core.entity_view_display.node.*.default` or has there been progress there in terms of handling config diffs? Ping @pyrello
> 
> https://iowaweb.slack.com/archives/C0164FNN1PV/p1642689940252600 , #4452

In this case in particular, I think we can partial split. The actual layout display should be the same as core, but it needs the extra `hidden: field: true` for each of the 3 additional fields.

joewhitsitt (approved):
Thanks for chiming in. Based on that and pulling locally, this works as expected. Tried to re-export to trip things up and it was identical. As an editor I could rely on the default permissions and still edit in layout builder

PR #7527, opened 2024-03-29
Fix Moderation Filter

### LINKED ISSUE #7511: moderation overview doesn't allow filtering by "review" state
Not sure if this is a bug or feature request?

On the content moderation overview (https://sandbox.prod.drupal.uiowa.edu/admin/content/moderated), the moderation stage filter only has values of "draft" and "archive" to filter by. Seems like an option to filter by "Review" would be the most useful option.

Image

# Proposed Solution

joewhitsitt:
Resolves #7511

Go to `/admin/content/moderated` and see that the "Review" filter is an available option in the moderation view and that if there is content with the status that the results reflect that.


bspeare (approved):
Works as described!

PR #7227, opened 2023-12-21
[Card] Added linkit to card field link

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7228./

# How to test
- Test with P2LB
```
ddev blt ds --site=brand.uiowa.edu && ddev drush @brand.local uli /node/761/revisions
```

- Revert to the pre-V3 page
- Edit a card paragraph
   - Add an image 
   - Add a title
   - Remove image from WYSIWYG
- Convert the page to V3
  - Verify that the `/media/` url is converted into the card

## Test overall card functionality
```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /cards
```
Test out card link functionality with Linkit
Test with `<nolink>`, `<front>`, and any others you can think of

### Styling
- Do we need extra styling here from https://github.com/uiowa/uiowa/pull/6742
### Circle button
- Circle button logic needs to be reworked

Issue #6343, opened 2023-04-11
As a content creator, I would like to be able to link to draft pages with auto-complete


makurj:
# Problem
As a content creator, I would like to be able to link to draft pages with auto-complete whether the page I'm on is draft or published, because I create many draft pages at one time that go live later, sometimes at the same time. It works if I grab the relative URL manually and paste it.

# Steps to reproduce

https://user-images.githubusercontent.com/52135878/231269950-c44db6f5-391d-4e7b-876d-f95025e8a94d.mp4

# Proposed solution

joewhitsitt:
This currently works as designed. Are you suggesting an enhancement instead? https://github.com/uiowa/uiowa/blob/main/config/default/linkit.linkit_profile.default.yml#L23

makurj:
Good to know, thanks.

Yes, this being addressed as an enhancement is fine. I'll remove the bug label.

joewhitsitt:
Can you adjust it then to be a user story? Is there a use-case for linking to unpublished pages?

makurj:
Done, and yes. Use case in issue summary.

joewhitsitt:
I this this should be done for all as default. Change the setting https://github.com/uiowa/uiowa/blob/main/config/default/linkit.linkit_profile.default.yml#L23 and update the matcher to include published status here https://github.com/uiowa/uiowa/blob/main/config/default/linkit.linkit_profile.default.yml#L18

pyrello:
When you turn this setting on the behavior is that you can see the link when logged out, but when you click it, it takes you to a 403 page (which now conveniently includes a login link). The autocomplete results do not indicate that the node is unpublished, which I think would be desirable.

joewhitsitt:
>The autocomplete results do not indicate that the node is unpublished, which I think would be desirable.

That is what I meant by "update the matcher to include published status here https://github.com/uiowa/uiowa/blob/main/config/default/linkit.linkit_profile.default.yml#L18"

joewhitsitt:
actually, i don't know if a token for node:status exists and a node moderation state token is not available yet.

https://www.drupal.org/project/drupal/issues/3225851

PR #6742, opened 2023-08-03
Extend Linkit to other link fields

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/4235

Menu link content support:

258191224-a9b8c6aa-5b25-4f63-8856-c1caa200abdf

Layout builder block and paragraphs support:

Screenshot 2023-08-04 at 2 00 45 PM

# To Test

As an editor, I can get autocomplete content context (linkit) as I start typing for menu links administered through /admin/structure/menu as well as all of the default sitenow v3 block content types (e.g. card, banner, cta) and their nested paragraphs (e.g. timeline item) similar to when using the WYSIWYG.

The following link fields are supported as part of this PR:

menu_link_content link

field_collection_more_path (ui events)
field_uiowa_banner_link
field_uiowa_button_link
field_uiowa_card_link
field_uiowa_cta_link
field_uiowa_event_link
field_uiowa_image_link
field_uiowa_quote_citation (blockquote)
field_timeline_link
field_uiowa_slide_link

PR #7529, opened 2024-03-29
[Linkit] Linkit for block components

### LINKED ISSUE #4235: Add link context to structure menu link search
# Story

As an editor, I would like to be able to link content on block types as well as the menu management interface using the autocomplete functionality found in the WYSIWYG.

# Problem
Within a Structure > menu page, when you use the search to choose where to link the page to, it does not provide context on whether it is a page, article, or person.

Screen Shot 2021-09-30 at 1 21 33 PM

# Steps to reproduce
Go into  Structure > Menus and edit/add a page and use the search menu to choose a link destination.

bspeare:
Trying https://github.com/uiowa/uiowa/pull/7227 again. 
Resolves https://github.com/uiowa/uiowa/issues/4235
More support of https://github.com/uiowa/uiowa/issues/6343

Linkit doesn't separate the url and title field separately.   If a link is in the url field, it automatically populates the title field on save where previously the title field was empty.   In order to get the card button, circle button, show/hide to work we need to unset that title field value. 

# How to test

For all blocks `core.entity_view_display.block_content.uiowa_*.default.yml` must be set to linkit or the media link will not render to the public path url. 

## Card

```
ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /cards
```
1. Add internal media file or node path, the title field should auto-populated from the entity, but if you delete the text the circle button should render. 
2. Add an external link, confirm the link title field is unset unless custom link text is added. 
  i. If someone wants to display "https://example.com" as their button link text or currently has a "https://..." in their card text, it would currently be unset. Since this is a decorative button, maybe we just add some help text linking to documentation on button text best practices?
3.  Add a `<button>` and `<nolink>` and confirm these do not render any links on the card. 
4. Add `<front>` links to the home page of the site and confirm these render as links to the home page.  
5. See published/unpublished context for content matching and a pink bg

## Banner and Button 

1. Add internal media file or node path, the url should be populated with something like `/media/416` the title field should auto-populated from the entity, save the block and the public file path should render.   Edit the block again and the url should have something like `/media/file/416` in it.  Save the block again and confirm the public file path is rendering. 

## Menu

1. Does not support media matching/substitution but should handle internal paths, external, <nolink>, <front>, etc.
2. As of right now, core supports both formats for front and nolink so while the database is not consistent, they are both handled appropriately by LinkIt.
3. See published/unpublished context for content matching and a pink bg

## Timeline

Similar to Card in how the title value is handled. Has a link indicator if URL is empty.

## UI Events

Invoking the field api formatter to construct link and place it where we want in the template if more link is checked. Falls back to https://events.uiowa.edu if checked but no LinkIt value is provided.

## List Blocks, Academic Dates block

This should still be using our franken-code entity reference autocomplete form api element and our public copy of `getUriAsDisplayableString()`. Basically just test that this is working as it currently is in PROD.


joewhitsitt:
Haven't pulled and tested yet but should handle button, front, nolink like this patch? https://www.drupal.org/project/linkit/issues/3223781

joewhitsitt:
looks like there is frontpage matcher in linkit and we just need to add it to our linkit profile: 
Screenshot 2024-12-11 at 4 28 29 PM

bspeare:
@joewhitsitt thanks for the frontpage matcher and patch suggestion.  Those are great to have but I think we still need custom processing in Card.php to handle how the link text is set within the card button.  Since linkit auto-populates the empty link title field with the link url, there needs to be some processing so that people can use the circle button if they leave the link text field blank.  Otherwise, this switching to linkit would remove our ability to use the circle button on cards.   This is the only component where we would need to do this type of processing.

joewhitsitt:
Got it. Yeah that title/url processing still makes sense to me. Thanks for investigating these matchers.

bspeare:
I've added checkboxes for the remaining work on this in the pr description area.

joewhitsitt:
Looking into the media processing a bit. In our Linkit profile, we have substitution set to use the direct media file but that isn't getting triggered in the current custom render code. There was an issue with the formatter not getting this, but updating the module fixes this. https://www.drupal.org/project/linkit/issues/3354873

I think we just need to get the benefits of the field formatter and put the result back in place in our custom code. That way we don't do our own media matcher processing.

joewhitsitt:
Made some tweaks to the layout builder form styles based on platform changes since https://github.com/uiowa/uiowa/pull/6742/files

Screenshot 2024-12-27 at 2 31 02 PM

joewhitsitt:
>Menu Link: Media file links are rendering as internal uri

This is going to be hard to do. I wonder if instead we should create a separate linkit profile for menu links so that the media matcher is not part of it. Menu links can be called all different ways (superfish, menu_block, etc.) and it isn't using a traditional field formatter which would run through the substitution override.

joewhitsitt:
>Do we need an update hook since the matchers are converting `<front>` to `/` and `<nolink>` to `route:<nolink>`?

I don't think we _need_ to if the existing links don't break. It would get our data consistent across the board, but from what I have spot checked, they seem to both work.

joewhitsitt:
Are the default list blocks (articles, people, pages) in scope? If not, maybe we should hold off on adding linkit to the field_collection_more_path (e.g. ui events)

joewhitsitt:
Extending the unpublished info to the content matcher's context output and applying the pink bg to layout builder offcanvas.

Screenshot 2024-12-31 at 11 59 14 AM

bspeare:
@joewhitsitt your changes look great! The only issue I found was with the academic dates block.  I am getting this error when using the more button: `InvalidArgumentException: The URI '/node/231' is invalid. You must use a valid URI scheme. in Drupal\Core\Url::fromUri() (line 293 of core/lib/Drupal/Core/Url.php).`

joewhitsitt:
Okay, thanks @bspeare. I can try to revisit that here soon. I feel like I switched it to run through that based on a different error. Was that an existing block with a link or a new one?

bspeare:
@joewhitsitt that was with a new block.

joewhitsitt:
Bummer. I think this means extending this to Form API situations like list blocks or the academic calendar might be out of reach or at least not worth the squeeze at this time. A lot of redundant logic will have to be implemented beyond media matching to handle these links (route:<nolink>, /node/123, https://example.com). Would require checking the type of link based on the string and then using one of the core utilities to form it into the right kind of Url Object (fromRoute, fromUri, fromUserInput, etc.).

From my search around the codebase, there just isn't a good core utility method or linkit helper that handles all links like the `l()` function seemed to do back in Drupal 7.

I will try to push up a change in a bit to revert list block and the academic date block back to using that entity reference autocomplete field and the old LinkHelper function. In fact the old function was copied line for line from the LinkWidget because that one is protected and we needed it to be public...

bspeare:
@joewhitsitt I was only getting the error on the academic dates block.  The page, articles, and people list blocks seem to work just fine with your solution.   Can we just revert to https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php and move forward?

joewhitsitt:
@bspeare I think I was also encountering issues with `route:<nolink>` on the list block, but my mind could be going in circles. The academic dates block is expecting a Url Object for the link render array. The listblock is using a string value and then handling it in the template.

If listblock is working, we could continue on and just change how the academic date renders the link. (e.g. type markup).

bspeare:
I see `<nolink>` renders as `route%3A` is that the issue you saw?

joewhitsitt:
Yeah. I think so. For the list block it doesn't properly converting those wacky type "links" which linkit is properly doing through the field formatter. works for block_content where it exists but isn't part of the process here. I might see how menu_link_content handles it, otherwise I think will end up with a big if/elseif/else type statement in addition to the media matching where we need to handle (<front>, route:<nolink>, internal:, entity:, base: etc.). Though /node/123 doesn't error, it doesn't get the alias conversion like the others.

joewhitsitt:
With https://github.com/uiowa/uiowa/pull/7529/commits/e26eae415f1fed3aa81a8b715cda0e38e0828644, this reduces the scope of this PR. Perhaps we can follow-up and continue trying to extend support to list block, but I want to avoid recreating linkit for form api/config values.

Should probably still test those to make sure I didn't miss something, but they should work as they currently do. I'll update the test instructions.

This leaves the new stuff in LinkHelper so that we can pick up where we left off. processLinkIt would need some expansion to handle other link types and Academic dates would have to most likely have to change from a link render array to a markup render array as the link will be a string instead of an Url object.

joewhitsitt:
Looks like Percy is failing on the footer copyright updating to 2025. I am going to approve the build.

james917 (approved):
Looks good

joewhitsitt:
Thanks @james917. Your review plus @bspeare and my intersecting work on this is finally getting this off the ground!

pyrello:
This was a big project! Nice work!

PR #7541, opened 2024-04-04
[Emergency site] Rave integration for alerts

### LINKED ISSUE #7502: Rave integration
# Story

As an Emergency site owner, I would like the Emergency site to consume alerts from Rave since that is where Hawk Alerts are initiated.

# Additional Info

Establish test and prod connection with rave

Rave data should populate title and body field on Emergency. Both fields need to be editable in case there is a typo. Discussion https://github.com/uiowa/uiowa/issues/7487#issuecomment-2013664430 and https://github.com/uiowa/uiowa/issues/7487#issuecomment-2013675035

**Rave endpoints**
- stage - https://content.getrave.com/cap/uiowatest/channel1
- prod - https://content.getrave.com/cap/uiowa/channel1

# Proposed Solution
```[tasklist]
### Tasks
- [x] Fetch alert info from Rave endpoints, similar to Classrooms and Facilities API calls
- [x] Use Rave info to create a Hawk Alert and populate the title and body fields. Fields need to remain manually editable. Theming/display not in scope for this slice.
- [x] Set up scheduled job to check Rave for new alerts - D7 job checks every minute. No need to check for updates to existing alerts as all updates will happen manually on the emergency site.
```

e-marie-w:
Closes #7502 

# How to test
- Code review. I believe I removed all code that would update or delete existing hawk alerts but a second set of eyes would be appreciated.
- `ddev blt ds --site=emergency.uiowa.edu && ddev drush @emergency.local config:import --source ../config/sites/emergency.uiowa.edu --partial -y && ddev drush @emergency.local cim -y && ddev drush @emergency.local cr && ddev drush @emergency.local uli`
- Import command: `ddev drush @emergency.local rave-alerts`
- Output should say 1 alert created, go to the content overview page to see it. Title, body, and Hawk Alert Identifier should match the headline, description, and identifier at the test endpoint - https://content.getrave.com/cap/uiowatest/channel1. Identifier should only appear on edit form and not be editable.
- Change the text of the title and/or body. 
- Run the import command again, output should now say 1 skipped.

Before merging I will change line 70 of EmergencyAPI.php to use BASE_URL_2 instead of BASE_URL_1. You can also make this change locally to test the prod endpoint instead of stage if you'd like.


docroot/sites/emergency.uiowa.edu/modules/emergency_core/emergency_core.module
```diff
@@ -0,0 +1,55 @@
...
+  $logger = \Drupal::logger('emergency_core');
+  $logger->notice('Starting the hawk alert content sync.');
+
+  $sync_service = new AlertsProcessor();
+  $sync_service->init();
+  $sync_service->process();
+
+  $arguments = [
+    '@created' => $sync_service->getCreated(),
+    '@updated' => $sync_service->getUpdated(),
```
joewhitsitt: These don't get updated/deleted right? Just created?

e-marie-w: Correct! Only created and skipped ever change. I originally left the others in to test I had completely removed everything. But perhaps I should remove them now to avoid confusion.

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/EmergencyAPI.php
```diff
@@ -0,0 +1,126 @@
...
+      if (isset($response)) {
+        $contents = $response->getBody()->getContents();
+        $alert = simplexml_load_string($contents);
+        $json = json_encode($alert);
+
+        /** @var object $data */
+        $data = json_decode($json, TRUE);
+
+        // Cache for 1 minute.
+        $this->cache->set($cid, $data, time() + 60);
```
joewhitsitt: if we need to cache at all, i'd argue we want this to be less than a minute so there isn't a chance that the scheduled job doesn't call it and serve a cached version, waiting a whole minute more to get the latest data. i think the simple fix would be to change it to 30. if there is no use-case to cache though, i think we could just remove all of the caching since this isn't visitor-triggered

joewhitsitt:
interested in testing this tomorrow

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/EmergencyAPI.php
e-marie-w: Great catch! I don't think there is a use-case to cache, I was just following previous work that updates far less frequently. I will remove it.

e-marie-w:
Hmm for some reason the `[notice] Hawk alerts content sync completed. 0 alerts were created and 1 skipped. That is neat.` message is now outputting twice.

docroot/sites/emergency.uiowa.edu/modules/emergency_core/emergency_core.services.yml
joewhitsitt, line 1: I see you went with ripping out cache stuff. This file should be adjusted then too. Sorry I haven't had a chance to pull down yet

joewhitsitt:
pulling to test

docroot/sites/emergency.uiowa.edu/modules/emergency_core/emergency_core.services.yml
e-marie-w: Thanks!! I found a few other places referencing it as well and so far everything still seems to work fine after removing.

joewhitsitt (approved):
Looks good. Thanks for making those changes!

I assume you are still planning on this:

>Before merging I will change line 70 of EmergencyAPI.php to use BASE_URL_2 instead of BASE_URL_1

PR #7563, opened 2024-04-12
Commencement event type

### LINKED ISSUE #7559: Ceremony content type
# Stories

As a commencement site owner, I want to be able to display ceremony information on the commencement website that comes from Content Hub.
As a commencement site owner, I want to be able to add additional information to the ceremony events that are pulled into my site.

```[tasklist]
### Tasks
- [x] Work with customer to determine necessary fields
- [x] Create commencement.uiowa.edu site split
- [x] Create event content type
- [x] Add fields necessary for API import
- [x] Add remaining fields
- [x] Content type best practices: content moderation?, path alias pattern, permissions, metatag, etc.
```

### LINKED ISSUE #7562: Probe program functionality, come up with plan, execute?
# Stories

As a commencement site owner, I can attach program PDFs to one or multiple ceremonies.

```[tasklist]
### Tasks
- [x] Work with customer to understand relationship between programs and events
- [x] Develop a plan for how programs will work going forward
- [x] Build a prototype or the actual thing if that makes sense
```

e-marie-w:
Resolves #7559 
Resolves #7562 

# How to test
```
```


pyrello (approved):
Noting here that the use of `uiowa_collection` as the paragraph type for ceremony information as currently implemented is creating a situation where there is a paragraph already open and there is a required headline field which prevents saving the form without filling it in. Going to approve this for now to keep things moving and make a note that we need to fix this later.

PR #7564, opened 2024-04-12
[Aggregator] Decode raw html in title

cory-skeers:
Raw HTML from the source is being displayed

image

# How to test

```
ddev blt ds --site=provost.uiowa.edu
```
Check https://provost.uiowa.ddev.site/test-rss-feed and see that the HTML gets encoded


joewhitsitt (approved):
Sorry for the delay! Works as expected

PR #7568, opened 2024-04-12
Scheduled hawk alert bash script

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7502

# Test

`ddev drush @uiowa07.dev ssh`
`cd /shared/logs && tail -f drush-cron.log`
Observe status messages each minute.

Back to filtered_html capability on body and situation updates so they can include images if needed.

Webmaster has the ability to override the import date (created) if need be.

```
/shared/logs$ tail -f drush-cron.log
 [notice] Checking for alerts.
 [notice] Hawk alerts content sync completed. 1 alerts were created and 0 skipped. That is neat.
[HAWK ALERT IMPORT] Mon Apr 15 18:32:49 UTC 2024
 [notice] Starting the hawk alert content sync from drush. This may take a little time.
 [notice] Checking for alerts.
 [notice] Hawk alerts content sync completed. 0 alerts were created and 1 skipped. That is neat.
[HAWK ALERT IMPORT] Mon Apr 15 18:33:49 UTC 2024
 [notice] Starting the hawk alert content sync from drush. This may take a little time.
 [notice] Checking for alerts.
 [notice] Hawk alerts content sync completed. 0 alerts were created and 1 skipped. That is neat.
```

# Follow-up

- Move scheduled job to PROD once script is deployed.
- Remove scheduled job from DEV.
- Empty drush-cron.log

PR #7587, opened 2024-04-17
Emergency website - Hawk Alert set default value for complete field

joewhitsitt:
# To Test

- `ddev blt ds --site=emergency.uiowa.edu`
- Add the active view to a page. Open that view in an anonymous browser, caching it.
- Run the `ddev drush @emergency.local rave-alerts` command and see one item is imported.
- Refresh the anonymous browser to see the alert.


joewhitsitt:
Turns out it wasn't a [caching issue](https://github.com/uiowa/uiowa/pull/7568#issuecomment-2059958104) but the filter on the views not working properly because the field was not getting set on node creation. Having a default value of false makes it work.

bspeare:
Pulling to test

bspeare (approved):
Works as the testing instructions described.

PR #7603, opened 2024-04-19
[Commencement] Display map on venue

e-marie-w:
# How to test

- go to https://commencement.uiowa.ddev.site/admin/content and check a few venues, the lat/long values should now be rendered as a map. ~~Note: some values are off and need updating, the important thing is that a map is there. Carver-Hawkeye Arena is accurate.~~ Lat/long values are updated, the IMU is no longer showing as in the Iowa River. :)
- Do we like the current Zoom level?


pyrello (approved):
Update looks good. We'll check zoom level with the customer when we have access to them again.

Issue #7611, opened 2024-04-22
SSL - ned.uiowa.edu on uiowa07


joewhitsitt:
existing site pointing to itsnt2709.iowa.uiowa.edu

joewhitsitt:
Adding www. coverage as well since the site request mentioned it.

joewhitsitt:
The approval emails seem to be delayed with this one.

joewhitsitt:
This is done.

PR #7612, opened 2024-04-22
Sprint 209 provisions

joewhitsitt:
SSL for ned: https://github.com/uiowa/uiowa/issues/7611
SSL for cgrer: https://github.com/uiowa/uiowa/issues/7614


docroot/sites/miller.lab.uiowa.edu/settings/default.local.settings.php
```diff
@@ -149,4 +149,4 @@
  */
 $settings['trusted_host_patterns'] = [
   '^.+$',
-];
\ No newline at end of file
+];
```
pyrello, line 152: ??

joewhitsitt: I think the provision process did that. not sure how it got that way in the first place.

pyrello (approved):
2 * 7 + 1 + 1 = 16

PR #7629, opened 2024-04-26
Classrooms - loop properly over all bundles

joewhitsitt:
Recent adjustments to the Processor caused a regression where only room features were processed, skipping over the additional taxonomy bundles associated with the record data.

Because they were skipping over, previous features remain unchanged.

# To Test

- On `main` sync down and resave https://classrooms.uiowa.ddev.site/node/2166/edit (CPH 155). Note that the accessibility feature `Table - ADA Student Individual Height Adjustable` is not mapped.
- On `classrooms_import_hotfix` resave the node again and see that accessibility feature is now mapped.


cory-skeers (approved):
Reviewed and tested, and all checks out 👍

pyrello:
Thanks for the quick fix, @joewhitsitt!

PR #7631, opened 2024-04-26
Allow for webform exports

### LINKED ISSUE #7551: Incomplete webform submission results download
# Problem
Downloading submissions from webforms with lots of results is inconsistent.  It took me 4-6 downloads to get a file that contained all of the submissions. 

# Steps to reproduce
- https://uiowa.edu/admin/structure/webform/manage/landing_page_introduce_yourself_/results/download

# Proposed solution
- Look at the suggestions in https://www.drupal.org/project/webform/issues/3284953?

```[tasklist]
### Tasks
- [ ] Try the webform specific settings.php temp dir solution here https://docs.acquia.com/acquia-cloud-platform/manage-apps/files/broken
- [ ] Deploy to DEV for testing, using the webform:generate functionality to make a boat load of submissions to try downloading
```

cory-skeers:
Resolves #7551 

# How to test

Code review.

Can test exporting any webform results for a site synced down in dev. The homepage on `uiowa01` has webform results, possibly others on the same or other applications.

Go to a webform Results:Download tab. Try different settings, and see that you are able to download an export.

The original trouble webform was https://home.dev.drupal.uiowa.edu/admin/structure/webform/manage/landing_page_introduce_yourself_/results/download


cory-skeers:
I'm interested in some feedback on
```
  $config['webform.settings']['export']['temp_directory'] = "/mnt/gfs/{$site_name}/tmp";
```
This is the suggested path from acquia for the webform-specific issue (or `"/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp";` if setting it for ALL tmp files), but does have the complication that we need to either preemptively create this directory for every site in our ecosystem or check for its existence and create it dynamically the first time someone tries to export a webform.

One alternative could be to do this for uiowa.edu specifically, as it was the original case where this problem was reached and (I think) the only currently known one. I was able to download all of sitenow.uiowa.edu's webform data despite them being quite large. Tippie has a couple webforms in the low 2-3k that could go either way (haven't tested). Every site has the potential to run into this issue, but the submissions tables need to get relatively large before it ever becomes a concern. I'd expect over enough time we'd start to see the issue more, but it's also dependent on people wanting/needing to export full results, which may not come up very often.

All this to say--this seems like the only approach to fully fix the issue, but creating directories for all sites for this one unique case (export webform results tables) felt awkward, and wanted to get opinions on how wide or narrow we may want to scope this.

joewhitsitt:
So my understanding is that the current tmp for each app is `/mnt/tmp/{$_ENV['AH_SITE_GROUP']}` which ends up like `/mnt/tmp/uiowa` and is problematic because of the load balancers.

The path `/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp` ends up like `/mnt/gfs/uiowa.prod/tmp` which would place it down the path similar to sites' files directories without being website specific.

If we wanted to make it more specific for webforms to avoid potential collision with other websites, wouldn't it be something like `$config['webform.settings']['export']['temp_directory'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/sites/{$site_name}/tmp";` which would be like `/mnt/gfs/uiowa.prod/sites/hr.uiowa.edu/tmp`

I am curious if it needs the specific /tmp directory to be created beforehand or if it is smart enough to create it if it doesn't exist. If it needs to be created beforehand, I'd probably opt to try the application.env path first since we would only need to do that for the existing applications and during the setup of future ones.

cory-skeers:
> The path `/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp` ends up like `/mnt/gfs/uiowa.prod/tmp` which would place it down the path similar to sites' files directories without being website specific.

Re-looking at the suggested updates, this is actually what they suggest for the non-webform-specific solution, and I don't know any reason why we couldn't still scope it to only the webform export config. Their webform-specific path is based on sitename, but I don't see any reasoning on that.

> If we wanted to make it more specific for webforms to avoid potential collision with other websites, wouldn't it be something like `$config['webform.settings']['export']['temp_directory'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/sites/{$site_name}/tmp";` which would be like `/mnt/gfs/uiowa.prod/sites/hr.uiowa.edu/tmp`
> 
> I am curious if it needs the specific /tmp directory to be created beforehand or if it is smart enough to create it if it doesn't exist. If it needs to be created beforehand, I'd probably opt to try the application.env path first since we would only need to do that for the existing applications and during the setup of future ones.

The docs say `The temporary paths referenced in these methods must be manually created before use. Setting variables does not create the folders, and causes uploads to fail until you create the directory`, so let's go with the application.env path and test it out in dev.

docroot/sites/settings/global.settings.php
```diff
@@ -84,3 +84,10 @@
 if (isset($_SERVER['REQUEST_URI']) && strpos($_SERVER['REQUEST_URI'], 'admin/structure/menu') !== false ) {
   ini_set('memory_limit', '256M');
 }
+
+// Under Acquia Cloud, this will override the temporary folder used by
+// the Webform module and might resolve issues around exporting.
+// See https://www.drupal.org/project/webform/issues/2980276 for more information.
+if ($ah_env !== 'local') {
+  $config['webform.settings']['export']['temp_directory'] = "/mnt/gfs/{$site_name}/tmp";
```
cory-skeers:
```suggestion
  $config['webform.settings']['export']['temp_directory'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp";
```

joewhitsitt:
I wish i had a time machine to understand the current tmp path which I assume Rick set. I wonder if we should be using that path for the global default.

I agree to start testing the webform version and see how that goes first.

joewhitsitt:
Sorry for the umpteenth reminder: a deployment out to DEV will require adjusting the its.dev website again https://github.com/orgs/uiowa/projects/20/views/1?pane=issue&itemId=22614984

pyrello:
This has been added to the `develop` branch and deployed to the DEV environment.

cory-skeers:
Tested in dev, and after the `group.env/tmp` directory was created, I was able to download the full webform results without issue.

On subsequent requests to download, it does looks like the original export is not held in the tmp directory and things are re-exported, so I don't think there's concern of different websites on the same application running into issues. _Maybe_ if two websites on the same application have requests to export webforms with identical machine names at the same time, but in that case I'd expect the file system to handle the concurrent requests.

joewhitsitt (approved):
Hmm. I wish the tmp file wasn't so plainly named (`landing_page_introduce_yourself_.csv`). Given that several sites have the same contact form, the chance for coincidence is higher. Could downloader B potentially get downloader A's CSV?

Maybe longer term we should be looking to make these site specific by creating the folders as part of the provision process.

So the follow up before deploy would be to create this directory with the same chmod 750 on each env of every app?

cory-skeers:
> Hmm. I wish the tmp file wasn't so plainly named (`landing_page_introduce_yourself_.csv`). Given that several sites have the same contact form, the chance for coincidence is higher. Could downloader B potentially get downloader A's CSV?
> 
> Maybe longer term we should be looking to make these site specific by creating the folders as part of the provision process.

The settings update would be pretty trivial for this, but yeah...pretty much the only limitation there is needing to have all those directories made prior. I'm hopeful that one site actively writing to one when another tries to create would be managed by the system to not overlap, and that the time that they're kept is short enough to limit collisions, but I'm making some assumptions there for sure.

> So the follow up before deploy would be to create this directory with the same chmod 750 on each env of every app?

Yup. Gonna make a PL card for this to help keep track.

PR #7632, opened 2024-04-29
updated config to support ddev 1.23.0

joewhitsitt:
# How to test

- `git checkout ddev_upgrade`
- `brew upgrade ddev`
- `ddev delete images --all`

# Few notes:

- I think the python stuff is necessary for the Debian web server.
- default port makes it easier for database apps to connect like phpstorm
- switch to mysql which aligns more with Acquia
- ~php 8.2 in preparation for our future~

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -7,55 +7,59 @@ router_http_port: "80"
 router_https_port: "443"
 xdebug_enabled: false
 additional_hostnames:
-  - "*.uiowa"
+    - '*.uiowa'
 additional_fqdns: []
-mariadb_version: ""
-mysql_version: "5.7"
-nfs_mount_enabled: false
-mutagen_enabled: true
+database:
+    type: mysql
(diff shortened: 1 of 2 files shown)


joewhitsitt:
WARNING: Because of the database change, this will wipe out your local dbs and start fresh.

pyrello:
> php 8.2 in preparation for our future

I'm not seeing that the PHP version is being changed in this except in a commented out line which is essentially documentation on settings we could be using.

pyrello:
Honestly, I'd be good with entirely removing all the commented out stuff at the bottom of the config.yaml file. It makes it harder to see at a glance what is actually modified.

joewhitsitt:
Since this is auto generated each time `ddev config --update` is run, I think we should leave the commented stuff in to better understand new options, etc.

pyrello (approved):
Reviewed code changes and ran through instructions. Not sure how to test it, honestly.

PR #7634, opened 2024-04-29
[ITS] Alert email settings

### LINKED ISSUE #7617: ITS email safeguards
# Story

As an ITS site owner, I would like some safeguards in place for our alert emails that are sent so we can avoid damage control in the case an email is accidentally sent to subscriber lists.

# Additional Info 

There has already been one alert email that was sent to the production population accidentally while testing.

# Proposed solution

```[tasklist]
### Tasks
- [ ] Establish a site specific configuration form
- [ ] Create a separate input for alert email, digest email that can take multiple values, comma separated
- [ ] Prevent errors if inputs are blank
- [ ] Update the email logic to call the config values, remove the hardcoded emails and the Env protection
- [ ] Wait to hear whether a separate input for the alert creation email to helpdesk staff is needed. If so, figure out a way to override the workbench_email template email
```

cory-skeers:
Resolves #7617 

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /admin/config/sitenow/its-settings && ddev launch -m
```
Email settings should be empty
Go to an alert and send an alert notification; no email should be sent to mailpit
`ddev drush @its.local its-digest`
You should see a relevant message that no email was sent; no email should be sent to mailpit

Masquerade as a non-webmaster, and see that you can't edit the settings at /admin/config/sitenow/its-settings
Masquerade as a webmaster and visit the /admin/config/sitenow/its-settings
Set email(s) for the various options
Re-test sending notification and digest emails; emails should be sent to mailpit with their respective to/bcc addresses


e-marie-w:
The daily digest is now working for me but the alert notification emails are not still.

cory-skeers:
Setting back to draft while we troubleshoot the single alert notification not building properly.

Other notes from eod: check that a measure gets displayed to user if they attempt to send an alert email and it does not send because an email has not been set, and add if not currently displayed. Also update the revision message in this scenario. Also update to config ignore the settings so they don't get wiped on deployments.

cory-skeers:
Capturing in writing what's otherwise been discussed:

Specifically, calls to `$emailFactory->sendTypedEmail(...)` appear to be working, but `$emailFactory->newTypedEmail(...)` do not. While the former are essentially a combo of the latter and `$email->send()`, the latter is not hitting the custom email builder as it should. Both methods had (I thought) been working previously, but I'm also unclear what could have changed to cause the change in behavior. So...troubleshooting.

joewhitsitt:
I think it is your check in between the newTypeEmail and the ->send() that is exiting early as those aren't fetched/set yet.

cory-skeers:
> I think it is your check in between the newTypeEmail and the ->send() that is exiting early as those aren't fetched/set yet.

Yup, that did it. Misunderstanding on my part on when it actually gets "built." Thanks!

With that settled, the checks and error handling here need some work then.

joewhitsitt:
pulling to test

joewhitsitt:
Can I suggest assigning `$this->helper()->config()->get('its_core.settings')` to a variable to make it more "dry"

joewhitsitt (approved):
Looking good! I tried my best to break it by testing the partial config ignore, only including a bbc, etc. messenger and revision message was as expected. For a moment I had a space after my comma and it said the email was invalid. Removing the space, it saved fine.

PR #7645, opened 2024-05-02
Classrooms - reduce revisions related to imported content

joewhitsitt:
Fresh off https://github.com/uiowa/uiowa/issues/7623, this is a small change that could greatly reduce the size of Classrooms' database based on the number of related taxonomy terms per node and the amount of entity usage tracking that happens with revisions.

# How to test

Compare with other import-based content types:

https://github.com/uiowa/uiowa/blob/main/config/sites/facilities.uiowa.edu/node.type.building.yml#L16
https://github.com/uiowa/uiowa/blob/main/config/sites/facilities.uiowa.edu/node.type.project.yml#L16

Scheduler stuff is just added default config. Not enabled.

### DIFF
--- a/config/sites/classrooms.uiowa.edu/node.type.room.yml
+++ b/config/sites/classrooms.uiowa.edu/node.type.room.yml
@@ -5,6 +5,7 @@ dependencies:
   module:
     - menu_ui
     - node_revision_delete
+    - scheduler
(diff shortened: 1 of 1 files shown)


bspeare (approved):
Looks good!

PR #7660, opened 2024-05-07
Static Map URL validation updates

### LINKED ISSUE #7593: Concept3D has changed the share URLs for Static Map media
# Problem
https://maps.uiowa.edu/?id=1890#!m/590647

has become:

https://maps.uiowa.edu/?id=1890#!ce/58422?m/590647?mc/41.6629197739891,-91.54511300000001?z/16?lvl/0

And our code is only looking for the fragment and then stripping the `!m/` so this new URL fails validation and media creation. It looks like the old URL is still supported and those URLs resolve on the website.

# Steps to reproduce
- Go to maps.uiowa.edu and get a marker share URL.
- Try creating a static map image on a SiteNow website. You will get a validation error if the URL does not contain the `#!m/590647` type pattern.

# Proposed solution
- Update validation and processing logic to handle this new marker pattern while maintaining support for the old pattern.
- Update the media content overview page to include this media type as a filterable selection.

# Known issue

https://sitenow.uiowa.edu/known-issues#uiowa-issue-7593

e-marie-w:
Closes #7593 

# How to test

- `ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /admin/content/media/`
- Check that you can now filter by Type 'static map image' 
- Copy the link to a location marker from https://maps.uiowa.edu/?id=1890#!ct/51247, or use this one https://maps.uiowa.edu/?id=1890#!ct/51247,57310,57311?m/579297?mc/41.66227490890242,-91.53613999999999?z/17.856476238675306?lvl/0
- Click Add Media->Static map image, paste the location link in the URL field. Add alt text and save. Hooray, the new validation pattern works!
- Now test the old validation pattern still works. You can use the same map link, but this time strip out everything after the ! and before the first m. Leave the m and everything after it until the next question mark. Remove this question mark and everything after it. Or, use this one: https://maps.uiowa.edu/?id=1890#!m/579297.
- Try adding another map but this time remove the ?m/ or !m/. Validation should fail with an error message.
- Bonus points: test the validation also works when adding via the media library from a wysiwyg. 

# Additional testing
- Remove ` static_map_regex: '%[!?]m\/([^?]*)%'` from `sitenow_media_wyiwyg.settings.yml` (found in `config/default`)
- Now try adding a static map, it should not work and an error message should appear.


joewhitsitt:
pulling to test

joewhitsitt:
Nitpicking here, but I see the same regex is used multiple times. I am wondering if it would be worth it to store this as module config that can be called instead of copy/pasta'ed.

joewhitsitt:
Follow-up feature ideas: 

- capture the Z (zoom) value from the URL if it exists and set it to the form field we have for it.
- WYSIWYG preview on insert is currently empty. After save it renders.

joewhitsitt (approved):
Works as expected. Thanks for working on this.

e-marie-w:
> Nitpicking here, but I see the same regex is used multiple times. I am wondering if it would be worth it to store this as module config that can be called instead of copy/pasta'ed.

I really like this idea so I'm going to spend some time seeing if I can do it and if it is worth it at this time.

e-marie-w:
Ready for review again! Please let me know if there's a better or simpler way to do the settings checks.

joewhitsitt:
pulling to test again

joewhitsitt (approved):
The checks look good. Just swapped them to assume were are good first and to proceed versus assume that we aren't. Added a message if there is a failure to process and render the map

PR #7666, opened 2024-05-09
[CKEditor] Added CKEditor lists patch

### LINKED ISSUE #6283: More options for styling lists
# Problem
When editing a page of policies I need more options for styling lists other than numerical and bullet lists. Options for alpha and maybe roman would be helpful. 

These additional options would also be necessary for when https://opsmanual.uiowa.edu converts to SiteNow.

```
While the patch in https://github.com/uiowa/uiowa/pull/7666 works great, the issue is that it allows users to add <ul type> which is deprecated: https://caniuse.com/mdn-html_elements_ul_type. The Drupal issue https://www.drupal.org/project/drupal/issues/3274635 continues to blocked until CKEditor can provide a way to only allow <ol> type https://caniuse.com/mdn-html_elements_ol_type.
```

# Proposed Solution

- [ ] Use the 10.4 compatible patch from https://www.drupal.org/project/drupal/issues/3274635
- [ ] See about limiting `ul type` in filtered_html config, otherwise override the CSS to display as expected until the type attribute can be removed: https://codepen.io/bspeare/pen/EaVxarz

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6283. 

# How to test

```
ddev composer install && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /ol-li-test
```

Test adjusting the `<ol>` then test converting it to `<ul>` and test out all the styles on the entire `<ul>` and nested `<ul>`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -238,7 +238,8 @@
                 "Layout builder off-canvas positioning problem when resizing browser ": "https://www.drupal.org/files/issues/2023-07-19/3359465-8.patch",
                 "Responsive image width/height values are not used from fallback image style": "https://www.drupal.org/files/issues/2023-08-04/3377420-responsive-image-dimension-fallback-style-9.patch",
                 "Simple decimals fail to pass validation": "https://www.drupal.org/files/issues/2023-12-21/2230909-309.patch",
-                "[3163299] Ajax exposed filters not working for multiple instances of the same Views block placed on one page": "patches/core-3163299.patch"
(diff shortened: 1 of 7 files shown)


briand44:
Here are a few comments from the D.O. issue that might be good to understand if we are concerned about this:
https://www.drupal.org/project/drupal/issues/3274635#comment-15456312
https://www.drupal.org/project/drupal/issues/3274635#comment-15457719

bspeare:
Closing because of https://github.com/uiowa/uiowa/issues/6283#issuecomment-2108521408.

PR #7668, opened 2024-05-10
IWP - Writer Bio Updates

### LINKED ISSUE #7653: [IWP] Writers Bio updates
# Proposed Solution

- [x] On the map interview - sounded like they were fine with just a multi-valued link field with text and url components.  This will allow us to migrate old data and be flexible for them to link internally or externally. They plan to use this for podcasts, videos and other types of interviews with folks. It won't be specific to On the Map Interviews. Label "Media".
- [x] Use square photo for Writers Bio rather than circle
- [x] Session/Status label should just be Session. This should be updated everywhere displayed (filter, node display, when creating content, taxonomy, etc.)
- [x] They would like the session filter to default to a session of their choosing. Possibly a default session boolean field on the session taxonomy or config form to set?
- [x] Asked about pdf icon for the writing samples or some better way to represent those - put in list item first. 
- [x] Adjust writing sample link text to be `[Person Name] Writing Sample` and `[Person Name] Writing Sample in Original Language` for accessibility

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7653

# How to test

```
ddev blt ds --site=iwp.uiowa.edu
```

- On https://iwp.uiowa.ddev.site/node/add/writer_bio see that the Session label no longer has /Status and that there is a new multiple title/link field at the bottom of the form. Add a couple links.
- See on the full node render that the Media field is rendered similar to other fields below the writing samples. Also see the same label treatment is extended for Session.
- Go to https://iwp.uiowa.ddev.site/admin/structure/taxonomy/manage/writer_bio_session_status/overview and edit a term or pick a new term. See the default session checkbox and check it. Notice except for the machine name, all mentions of Session have the same label treatment.
- Go to https://iwp.uiowa.ddev.site/residency/participants and see that the term you selected is loaded by default and that the filter label is "Session." Open this page in an anonymous window. In your authenticated window, go back and choose a different term to serve as default. Check the box. Save. Inspect the old term and see that it is no longer checked. Leave without saving. Refresh the anonymous window and see that the view updates with your new default term.
- If you have stage_file_proxy set for iwp.prod.drupal.uiowa.edu, you should see square images for the full node and teaser.
- On the teasers you should see a bulleted list of samples (name prepended) and media links together.
- `ddev drush @iwp.local cim` does not remove your default session config.

### DIFF
--- a/config/sites/iwp.uiowa.edu/config_split.config_split.site.yml
+++ b/config/sites/iwp.uiowa.edu/config_split.config_split.site.yml
@@ -16,6 +16,7 @@ theme: {  }
 complete_list:
   - config_split.config_split.site
   - core.entity_form_display.node.writer_bio.default
+  - iwp_core.settings
   - metatag.metatag_defaults.node__writer_bio
   - node.type.writer_bio
   - pathauto.pattern.writer_bio
@@ -25,6 +26,7 @@ complete_list:
   - 'field.storage.node.field_writer_bio_*'
   - 'taxonomy.vocabulary.writer_bio_*'
 partial_list:
+  - config_ignore.settings
   - metatag.settings
   - workflows.workflow.editorial
   - 'user.role.*'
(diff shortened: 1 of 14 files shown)


cory-skeers (approved):
I thought I broke the default session at one point, but I think it was having multiple tabs open for the terms and my confusing which I'd saved or left unsaved. Trying to recreate it, everything worked as expected.

Everything looks good! 👍

joewhitsitt:
Thanks @cory-skeers

PR #7671, opened 2024-05-13
Move CAPTCHA to default, clean up Admissions

### LINKED ISSUE #7657: Ability for webform creators to add CAPTCHA Math Challenge to their webforms
# Story

As a SiteNow v3 webform creator, I would like the ability to add additional spam prevention so we can better filter out spam submissions.

# Additional Info

Epic #6470 

Provide ability for webform creators to add a visible Math CAPTCHA to their webforms in addition to the default Honeypot and Antibot spam prevention measures.

A future slice could be to add https://www.drupal.org/project/riddler as another option.

What do we do with hCAPTCHA that is on Admissions?

# Proposed Solution
```[tasklist]
### Tasks
- [x] Leaving the hCAPTCHA bits in the Admissions split, move the other CAPTCHA config out to the default and changing the default challenge type.
- [x] Make sure webmasters the ability to select and implement the math challenge
- [x] Remove the following restriction from Admissions https://github.com/uiowa/uiowa/blob/main/docroot/sites/admissions.uiowa.edu/modules/admissions_core/admissions_core.module#L260-L265
- [x] Since Spam comes up a lot in support, establish some docs to point to on how to place/opt-in to this functionality
```

joewhitsitt:
Resolves #7657

Screenshot 2024-05-13 at 11 38 57 AM

# Test

- As a webmaster, I can add a math challenge captcha to my webform and have it show up for an anonymous session.
- As an admissions webmaster, I can add a math or hcaptcha challenge to my webform and that my default challenge type is set to hcaptcha (anonymous session) assuming I have the api keys necessary for it to work. DM Joe for keys or grab from the secrets file off of one of the remotes.

# Review

https://sitenow.uiowa.edu/documentation/webforms/handling-webform-spam-submissions

### DIFF
--- a/config/default/captcha.settings.yml
+++ b/config/default/captcha.settings.yml
@@ -2,7 +2,7 @@ _core:
   default_config_hash: QDFjOXYIYVwCPQYHY4wAx4DUqOEkNaZokIx6DGApR9I
 enable_globally: 0
 enable_globally_on_admin_routes: false
-default_challenge: hcaptcha/hCaptcha
+default_challenge: captcha/Math
 description: ''
 title: ''
 administration_mode: false
--- a/config/default/core.extension.yml
+++ b/config/default/core.extension.yml
@@ -15,6 +15,7 @@ module:
   block_content: 0
   block_content_template: 0
   breakpoint: 0
+  captcha: 0
   ckeditor5: 0
   config: 0
   config_ignore: 0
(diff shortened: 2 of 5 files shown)


e-marie-w (commented):
Works great! I tested adding the math captcha to both new and existing webforms on sandbox as a webmaster, and then tested the same on admissions. Also on admissions I tested that the default captcha remained hcaptcha instead of math captcha like on the other site. Also on both sites I checked the captchas could be seen by anonymous visitors and that they actually submitted/errored based on correct/incorrect answers. I also reviewed the draft doc, I really like it and it'll be great to be able to refer people to it.

e-marie-w (approved):

joewhitsitt:
Thanks @e-marie-w

PR #7693, opened 2024-05-16
[IWP} Added quick fixes to how writing sample is displayed

bspeare:
This pr adjusts writing sample link text to be `[Person Name] Writing Sample` and `[Person Name] Writing Sample in Original Language` to remove apostrophe and added `in` to the text. 

# How to test

Code review


joewhitsitt (approved):
oops. that was an assumption on my part. if that is the preferred output, this looks correct.

PR #7710, opened 2024-05-22
[Menu block] Horizontal menu with centered alignment

### LINKED ISSUE #7703: Feature request - option to center horizontal navigation menu
# Story
Request from Makur for IWP and Whitman Web. They are using the horizontal navigation menu and do not like that it's off-kilter compared to the rest of the page. They are asking for the ability to center the menu.

Example pages:
https://whitmanweb.iwp.uiowa.edu/civil-war-writings
https://iwp.prod.drupal.uiowa.edu/current-programs/digital-learning

# Proposed solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7703

# How to test

```
ddev blt frontend && ddev blt ds --site=tippie.uiowa.edu  && ddev drush @tippie.local uli  /faculty-research
```

- Edit layout and edit the menu block
- Select "Center" from "Aligment" layout builder style at bottom of menu block edit
- Confirm that the alignment option only appears if "Horizontal" is selected
- Save page and confirm that only top level menu is centered and dropdowns are not centered
- Check mobile view and confirm everything is still left aligned
- Test with "Vertical" alignment and confirm that everything looks as expected
- **Question:** Should we unset the "Center" value if you switch back to "Vertical" - we do this in other blocks, but in this case having the center alignment class on the vertical navigation does not change the appearance of the menu.  Reference: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/js/image-block-form.js#L14


pyrello:
A few thoughts:
- The alignment option shouldn't be multiple select.
- The horizontal and vertical options should have the same presentation for mobile device widths. That is, both should present the hamburger menu and small visual footprint. I know this isn't strictly related to this work, just making a note of it.
- In this case, I don't think we should worry about removing the setting if it has no effect.

The first bullet should be changed for approval. Everything looks good except that.

pyrello (approved):
Looks good. Nice update.

PR #7727, opened 2024-05-28
Remove config_filter dependency

pyrello:
Removes the `drupal/config_filter` dependency from composer.json

# How to test

- Confirm that `config_filter` is not an installed module: https://github.com/uiowa/uiowa/blob/main/config/default/core.extension.yml
- Code review


joewhitsitt (approved):
Thanks for checking into this

PR #7698, opened 2024-05-20
CORS Handling for paths and json file extensions

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7667 and https://github.com/uiowa/uiowa/issues/7686

This opens things up a bit with specifying file types and paths instead of handling a list of allowed origins and opening everything up using the Drupal Core CORS functionality in the services.yml file.

# How to test

## JSON:API endpoints:

- Using https://signage-dev.uiowa.edu/admin/config/services/hawkalert set the endpoint to https://emergency.stage.drupal.uiowa.edu/api/active
- Go to https://signage-dev.uiowa.edu/admin/config/development/performance and clear caches
- Visit https://signage-dev.uiowa.edu/college-education/coe-main in an anonymous browser session and open up the console log. See that there are CORS related errors about every min.
- Go back to https://signage-dev.uiowa.edu/admin/config/services/hawkalert and set the endpoint to https://emergency.dev.drupal.uiowa.edu/api/active
- Go to https://signage-dev.uiowa.edu/admin/config/development/performance and clear caches
- In the anonymous browser session, do a hard refresh. CORS issues should no longer be outputting.
- Go to https://emergency.dev.drupal.uiowa.edu/node/51/edit and change the authored date to now.
- Wait for the anonymous browser session on signage-dev to update. The hawk alert should display.

## Workflow Form

_Waiting on workflow form to test against_

## JSON file extensions:

This should work:
https://cors-test.codehappy.dev/?url=https%3A%2F%2Fgrad.dev.drupal.uiowa.edu%2Fsites%2Fgrad.uiowa.edu%2Ffiles%2F2024-05%2FJSON.json&origin=https%3A%2F%2Fworkflow.uiowa.edu&method=get

This should not:
https://cors-test.codehappy.dev/?url=https%3A%2F%2Fgrad.uiowa.edu%2Fsites%2Fgrad.uiowa.edu%2Ffiles%2F2024-05%2FJSON.json&origin=https%3A%2F%2Fworkflow.uiowa.edu&method=get

For more in-depth testing, I think a workflow form is necessary.

## Alternative testing

With a postman account you can run these tests and observe the access-control headers responses for DEV tests.
https://www.postman.com/joewhitsitt/workspace/sitenow-postman/collection/19389417-833749e4-ed0e-4250-9042-cd0041555cf2?action=share&creator=19389417

Screenshot 2024-05-21 at 4 00 52 PM

PR #7733, opened 2024-05-29
Update uiowa_alerts to handle both versions of emergency website endpoint

### LINKED ISSUE #7687: Update SiteNow v2/v3 so it can consume Hawk Alerts from new Emergency endpoint
# Story

As a SiteNow v2/v3 service owner, we would like the SiteNow v2/v3 services to be able to consume Hawk Alerts from the new Emergency endpoint so that our visitors can be notified of any active Hawk Alerts.

# Additional Info

https://github.com/uiowa/uiowa/issues/7550#issuecomment-2098621468

> Removed the /api/active.xml endpoint option.
> We couldn't determine significant usage for this endpoint. Consumers are encouraged to switch to the JSON format.
> Changed the /api/active.json endpoint to /api/active
> It is still JSON but without the file extension. The endpoint won't resolve with .json appended.
> 
> The other changes are in order to follow the JSON 1.0 specification:
> The outside wrapper uihphawkalert becomes data.
> The inside wrapper hawkalert becomes attributes.
> date changes from UNIX timestamp to ISO-8601 YYYY-MM-DDThh:mm:ssTZD
> more_info_link has been updated to output the website URL with https:// prefix.
> Public Safety would like to deprecate use of e.uiowa.edu going forward.
> 

We don't want to deploy this until launch of the new endpoint.

Related to v2/v3, Sean mentioned maybe there is a way we can make the endpoint configurable so the new code could be deployed in advance of the Emergency site launch. At launch we could do some sort of `blt ume cset`.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7687

Screenshot 2024-05-29 at 11 58 59 AM

# How to test

```
ddev blt frontend && ddev blt ds --site=default && ddev drush @default.local cset uiowa_alerts.settings hawk_alert.source https://emergency.dev.drupal.uiowa.edu/api/active.json
```
See the old test alerts from emergency.dev.drupal.uiowa.edu show. This has remained unchanged but there is now a bit of code to look for the .json file extension on the endpoint. If it doesn't exist, it will load the new library.

```
ddev drush @default.local cset uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active && ddev drush @default.local cr
```
When the time comes we can update hook and set this new endpoint, clearing necessary caches and updating the default config for the source. We can also remove the older library and the code that handles the switch out.

See the test alert from emergency.stage.drupal.uiowa.edu show with the correct CT date time and a proper link to the website.

# Follow up

- Update hook to set source on existing, flush cached block.
- Default config change for new sites.
- Remove library switch logic, old library entry, old library js.


GaryRidgway:
Pulling to test

GaryRidgway:
>When the time comes we can update hook and set this new endpoint, clearing necessary caches and updating the default config for the source. We can also remove the older library and the code that handles the switch out.
>
>See the test alert from emergency.stage.drupal.uiowa.edu show with the correct CT date time and a proper link to the website.

I got to this portion and got confused. I ran the second command and then got a `cors` error and didnt see anything. Could you explain the latter half of these instructions/expectations again?

Edit: [Brian provided me this comment](https://github.com/uiowa/uiowa/pull/7698#issuecomment-2140680665)

GaryRidgway (approved):
After finding the comment about changing something on the stage site, I have observed this working.

Issue #7728, opened 2024-05-28
Update sidebar link styling, update link decoration


joewhitsitt:
# Problem

Sidebar links default as plain dark text and are hard to distinguish from other text in the sidebar.

# Background

              >     * Wasn't there EoD talk about making the links in the sidebar look like links? There is no underline, color change, or hover effect.

The default behavior of the card links are black due to the how the card handles background colors: https://sandbox.prod.drupal.uiowa.edu/people/sean-adams-hiett.    We can either process in the `bg--none` class like on https://github.com/uiowa/uiowa/blob/main/docroot/sites/its.uiowa.edu/modules/its_core/its_core.module#L522 or shift our default for this layout to `bg--none` but that would require some greater refactoring and testing.

_Originally posted by @bspeare in https://github.com/uiowa/uiowa/issues/7695#issuecomment-2135698232_

# Proposed solution

- [ ] Shift our default for the layout to `bg--none` and check against existing splits for a consistent, easy to interpret interface across the platform.
- [ ] Remove one-off adjustments.

joewhitsitt:
This design pattern also surfaces on the contact block.

https://intradent.dentistry.uiowa.edu/tms

PR #7734, opened 2024-05-29
[Commencement] jump link

GaryRidgway:
Josh explicitly asked for jump links in the top area. Apparently it has been requested of him many times and he seemed pretty adamant.

>they have had a lot of feedback that visitors didn't see the Livestream section on their current site and Josh is worried that will continue to be the case so he was requesting a jump link so folks would easily know there is a Livestream available. He also thought it would be nice for a jump link to the Ceremony Information at the bottom.

# How to test
- Check to see if there are jump links on the ceremony events.


joewhitsitt:
sorry, i wasn't in the meeting. was this a regression or a requested enhancement?

bspeare:
Seems a little weird to add jump links here - seems a bit overboard? :)

GaryRidgway:
Sorry for not providing enough context, I'll update it.

bspeare:
Instead of having a lot of buttons on the page:

Screenshot 2024-05-29 at 3 25 04 PM

Should we just add the anchors in a `<ul>` with links?

anchors

joewhitsitt:
with a long body, regardless of buttons or ul, this will be "below the fold"

bspeare:
Then we should probably just use the horizontal menu displaying below the page title?

briand44:
>with a long body, regardless of buttons or ul, this will be "below the fold"

I mentioned this to Josh but he didn't seem that concerned with long body causing issues here.

joewhitsitt:
>Then we should probably just use the horizontal menu displaying below the page title?

If that is the pattern we are establishing on this website for this (e.g. Venue), I would think so.

bspeare:
Here's what the horizontal menu could look like:

Screenshot 2024-05-29 at 3 36 14 PM

Screenshot 2024-05-29 at 3 38 06 PM

joewhitsitt:
hmm. it seems to get lost in the pattern. black is quite bold here. when i think of horizontal nav, i think of https://tippie.uiowa.edu/iowa-mba

venue and tippie benefit from the header image...

and now that i've seen it in the [sidebar](https://github.com/uiowa/uiowa/issues/7728), links that don't look like links are starting to get to me. Tippie has the drop arrows for most which helps.

bspeare:
@joewhitsitt I agree, maybe something like this. 

Screenshot 2024-05-29 at 3 48 19 PM

briand44:
Would ^^ always render automagically or would Josh need to add menu links similar to Venues to get these to render? Also, if this is what we are suggesting, we probably want to run it by Josh before moving forward. He signed off on Alan's approach but think it would be good to get feedback if we are planning to go a different direction.

briand44:
Also, minor nitpick. "Watch Live" is the preferred text.

bspeare:
> Would ^^ always render automagically or would Josh need to add menu links similar to Venues to get these to render? Also, if this is what we are suggesting, we probably want to run it by Josh before moving forward. He signed off on Alan's approach but think it would be good to get feedback if we are planning to go a different direction.

This could be built as a custom block or using the menu block.

briand44:
I think in this case he would expect them to always show (Custom Block?) since the headers for those sections always show. He said he could add text that there is no livestream if that was the case.

bspeare:
> I think in this case he would expect them to always show (Custom Block?) since the headers for those sections always show. He said he could add text that there is no livestream if that was the case.

I'd probably vote to have the links hidden under a menu item on the live site so we aren't trying to recreate the menu block in a custom block.

bspeare:
Next steps on this:
- Update the template to look like https://github.com/uiowa/uiowa/pull/7734#issuecomment-2138238704 and use the menu block set to use hidden links in the main navigation.

GaryRidgway:
> Next steps on this:
> 
> * Update the template to look like [[Commencement] jump link #7734 (comment)](https://github.com/uiowa/uiowa/pull/7734#issuecomment-2138238704) and use the menu block set to use hidden links in the main navigation.


Sounds good! I do think that they wanted the 'Watch live' first and the second one to explicitly say "Ceremony info". any issues with that?

GaryRidgway:
Now that these most recent changes have been made,

<details><summary>The Ceremony pages with the jump links looks like this:</summary>
<p>

Ceremony

</p>
</details> 



<details><summary>And the Venue pages look like this, except the links are centered on the page (Sorry about not running `blt frontend` before recording)</summary>
<p>

Venue

</p>
</details>

briand44:
Did we decide not to center the Venue links anymore?

Screenshot 2024-05-30 at 10 44 35 AM

bspeare:
@briand44 they are probably still centered, perhaps Alan hasn't run `blt frontend` recently?

GaryRidgway:
> Did we decide not to center the Venue links anymore?
> 

That is quite odd, as I have them selected as centered in my local. I only changed the color. Let me poke at this.

GaryRidgway:
> @briand44 they are probably still centered, perhaps Alan hasn't run `blt frontend` recently?

I think this is what's happening

GaryRidgway:
After running `blt frontend` things went back to being centered! so just imagine my above gifs have them centered lol

bspeare:
Added in the background pattern from https://github.com/uiowa/uiowa/pull/7734#issuecomment-2138238704.

bspeare (approved):
The changes look good!

PR #7745, opened 2024-05-31
[Classrooms] Entity usage stuff

### LINKED ISSUE #7675: Investigate why entity_usage table on classrooms keeps growing
https://iowaweb.slack.com/archives/C0164FNN1PV/p1715283733005369

2+ GB
```[tasklist]
### Tasks
- [ ] Inspect the entity_usage table for clues on how it differs from other websites
- [ ] Step through room import process to see if unecessary ineffieciencies exist
- [ ] https://github.com/uiowa/uiowa/issues/7768
```

cory-skeers:
Resolves #7675 

# How to test

```
ddev blt ds --site=classrooms.uiowa.edu
```
Go grab some lunch, a coffee, read a book...It'll take awhile.

```
ddev drush @classrooms.local pmu entity_usage && ddev drush @classrooms.local en entity_usage && ddev drush @classrooms.local uli /admin/config/entity-usage/batch-update
```

Bulk update the entity usage table
Use your preferred db tool and check out the the `entity_usage` table, or use
```
ddev drush @classrooms.local sqlq "select count(*) from node_revision" && ddev drush @classrooms.local sqlq "select count(*) from entity_usage"
```
Both should be more reasonable (in the 100s to 1000s for node revisions and under 100k for entity usage)

Run the classrooms rooms import
```
ddev drush @classrooms.local classrooms-rooms
```
No (or very few) updates should be made from it. Check the node revisions and entity usage tables again, and they should be the same size as before.

Delete and re-add a room or edit a room and change the imported values
An example of this could be
```
ddev drush @classrooms.local sqlq "delete from node__field_room_technology_features where entity_id=1066"
```
which would remove all tech features from node 1066, which if the importer works correctly, should be seen as a mismatch with the source and that node (and only that node) should get updated. If there are fewer than 3 revisions for that node, the `node_revision` count should not increase as a result.

Note: The room needs to be updated like this or you need to comment out the code in the room presave, otherwise it will update upon saving and you will still have 0 records which need updating.

## Followup for merging

I added the 'needs scheduled deployment' label. I don't think it's strictly necessary, but the update does take quite a bit of time, and it may be good to avoid possible overlap with site edits.

After deployment, the prod `entity_usage` table should be truncated and regenerated. Probably the easiest way to do this is like the testing instructions here, and to just uninstall and reinstall the module, then log in and bulk update the table.


cory-skeers:
I think this is working without updating unnecessarily (which is also limiting the expansion of the entity_usage table).

Yet to do--determine how we want to clean up the already large table, and investigate why the num updated from the process isn't correctly reporting.

joewhitsitt:
Nice detective work. 

>determine how we want to clean up the already large table

Technically I think we can uninstall the module in PROD and re-enable, batch generate after the module cleans up the footprint. Could try locally first. I don't think there are any blocking dependencies on the module.

cory-skeers:
So what _should_ work after reducing the number of allowed revisions for a content type is running `drush node-revision-delete:queue` to queue up content revisions and the next scheduled/manual cron run clears them out.

Something is wonky with rooms where the revision count is set to 3, but older revisions are not properly getting queued for removal. I'm trying to see if there's a right recipe of queuing and clearing to get the automated processes to clear it up, but a backup option could be to just throw an update hook at the problem.

docroot/sites/classrooms.uiowa.edu/modules/classrooms_core/classrooms_core.install
```diff
@@ -90,3 +90,65 @@ function classrooms_core_update_9002(&$sandbox) {
...
+      ->execute();
+    // For each revision, delete it. If it is the default revision,
+    // it will be protected by the system and the delete process
+    // will error. Which is great! That is the revision that should be kept.
+    foreach ($results as $result) {
+      try {
+        $service->deleteRevision($result->vid);
+      }
+      catch (\Exception $e) {
+        continue;
```
pyrello, line 138: We should maybe add some type of output here so when this runs if there are any revisions that don't get deleted we can trace them down.

cory-skeers: As it is, we'll have an exception for every node for the default revision not getting deleted. But I think if we add a join with the `node` table when getting the revision ids, we can add an extra condition to not grab the default revision id, and add a message here without the extra noise.

I'll take a look

pyrello: Ah.. not a big deal if we don't do this.

e-marie-w (approved):
Yesterday this wasn't working for me but today it is perfectly so I'm blaming local gremlins. Table results are lower as expected, and running the room importer does not increase them unless an update is actually made. Awesome work!!

joewhitsitt:
Going to merge in preparation of the scheduled deploy for tomorrow morning.

PR #7782, opened 2024-06-07
[Bug] Button `aria-hidden`

GaryRidgway:
Explicitly sets the value for `aria-hidden` to a string value.

What happens when you set a value in an attributes to `TRUE` is that you are telling Drupal that you want that attribute, but that is all. you are not _setting_ the value. This is relevant for other boolean values that are on html elements like `hidden`.

# How to test

Pull a site, look at a card, observe the GLORY OF ARIA HIDDEN on the button.


bspeare:
pulling to test

bspeare (approved):
Great fix! Thank you!

Issue #7049, opened 2023-10-26
P2LB - Disable v2-isms for a page after v3 revision has been published


joewhitsitt:
# Story

As an editor after I have converted a page to v3, there is nothing stopping me from creating new v2 content within the v3 revision.

# Background

- I had to fix a site before turning off the functionality because the editor had gone back to old ways and added v2 content again to the now v3 content.
- Larger sites that will need more time converting (HR 1100+ pages) will allow them to convert without creating a mess again.

# Proposed solution

- [ ] Figure out the best way to identify if the current revision is v3 or v2. Account for the possibility of the page being converted to v3 and then reverted back to v2.
- [ ] If current published revision is v3, disable the paragraphs functionality on the edit form. The add form shouldn't be necessary as the body is available for new pages.

pyrello:
This is probably still worth doing.

briand44:
Grooming conversation... What about new content? Would that have paragraphs functionality enabled? Would we want that? Would the body field be available during this transition phase? Joe confirmed on brand.uiowa.edu that the body field is available and that new content doesn't have paragraph functionality.

How can we make sure we are only targeting v3 revisions with this? Any revisions newer than the protected revision or are there v3isms we can look for? May not be able to go off of newer revisions than the protected revision because they could revert to the protected revision.

Once we identify what revisions we are targeting, it should just be disabling the paragraphs functionality.

joewhitsitt:
This might be a non-issue now. `sitenow_p2lb_form_node_page_edit_form_alter()` seems to be doing enough here. Going to poke more to make sure there isn't a gap somewhere. Perhaps I had misidentified what went wrong:

>I had to fix a site before turning off the functionality because the editor had gone back to old ways and added v2 content again to the now v3 content.

joewhitsitt:
Unless someone else recalls how this might have been exposed, I am marking this up as a non-issue.

<img width="1142" alt="Screenshot 2024-03-22 at 11 30 19 AM" src="https://github.com/uiowa/uiowa/assets/4663676/5c4b58b1-0b6a-43c9-ac31-705b540e55dc">


https://github.com/uiowa/uiowa/blob/aeb2ea6372286a7161d68f140c989e1909df1c1c/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module#L40-L55

joewhitsitt:
[...]

Issue #6859, opened 2023-09-07
Revisit p2lb protected node revision to ensure proper cleanup


joewhitsitt:
# Problem

With the current protected revision, there is a potential for v2 draft revisions to exist that are "newer" (high id increment). We are currently targeting the protected node revision id and anything _less than_ which leaves these newer revisions around and prevents complete clean up of v2 content.

>Do we want to delete all revisions except for the "current" one? I've run into an issue on brand where things get messy and there are v2 revisions newer than the protected revision. This results in the orphan purger not cleaning things up because the current code only deletes the protected revision and anything that is "less than"

<img width="1154" alt="Screenshot 2023-09-06 at 2 46 11 PM" src="https://github.com/uiowa/uiowa/assets/4663676/232ac891-64be-4969-90e7-2f372e633e27">

_Originally posted by @joewhitsitt in https://github.com/uiowa/uiowa/issues/6830#issuecomment-1708993147_

# Proposed Solution
- [ ] Look for a ways to make sure the protected revision is the newest revision before conversion or change our criteria for cleaning up v2 content after conversion is complete.

# Workaround
- Manually delete those leftover revisions and trigger the `delete orphans` entity_reference_revisions functionality an additional three times to handle paragraph nesting.

cory-skeers:
Couple options could be to 

- On conversion, create a new revision as a copy of the current published one, to set that as protected
- On conversion, get the "true" latest (not latest published) and mark that one as the protected revision
- Leave logic the same as now, but on cleanup, just remove all revisions prior to the latest revision (which should be V3)

cory-skeers:
> On conversion, create a new revision as a copy of the current published one, to set that as protected

After discussing in grooming, I think I'm leaning toward this option, as it gives the possibility of keeping some V3 revisions (up to the 10 limit), which while isn't a necessity, could be nice.

It does bring up the question of which content should be converted: the currently published page or the drafted content? I think we just need to make a choice on which happens here and be consistent, and possibly add some messaging as such.

pyrello:
My hunch is that this is related to the content moderation module, although I'm not totally clear on exactly how yet. I did some debugging to compare what happens when a new revision is created vs when we convert a node and create a new revision based on that. 

https://git.drupalcode.org/project/drupal/-/blob/9.5.10/core/modules/content_moderation/src/EntityOperations.php?ref_type=tags#L173-182

[...]

PR #7786, opened 2024-06-07
P2LB  - Prevent access to paragraphs widget in new nodes

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7049

# Problem

New page content that is created with the p2lb converter on thinks it is a v2 page the next time it is edited because a v3 revision id is never set for the node.

This PR does not address nodes that got created since P2LB got turned on. Instead it is focused on limiting the scope of the problem to prevent new instances where the problem described above can happen.

# To test
- Code review
```
ddev blt ds --site law.uiowa.edu && ddev drush @law.local uli node/add/page
```
- Add a page and name it something like "After P2LB" so you can easily identify it later.
- Click the "Edit" tab and confirm that you are not able to see the v2 Section Blocks field.
- Click the "V3 Status" tab and confirm that the message you see says "This is a v3 page that was created after the SiteNow v2 to v3 converter was enabled."
- Confirm that you can see/click the "Layout" tab and add blocks through LB.
- Visit https://law.uiowa.ddev.site/admin/content/sitenow-converter and confirm that the new page does not show up in the list.
- Confirm the number of results matches https://law.uiowa.edu/admin/content/sitenow-converter
- Visit https://law.uiowa.ddev.site/admin/content/sitenow-converter/converted and confirm that the new page does not show up in the list.
- Confirm the number of results matches https://law.uiowa.edu/admin/content/sitenow-converter/converted. The quick way of doing this is to find the last page of results, make sure it matches and then count just the results on the last page.
- Presumably, you've created some revisions by this point to the page you created. If not, edit the page to add a revision.
- Confirm that you can delete all previous revisions.
- Confirm that you can delete the node you created.

## Cleanup
**Note: Running the cleanup script is currently not working on HMB for reasons that I'm not entirely clear about. I think we could potentially cover the code change related to that with a code review since it seems like we'll be investing some more time in fixing it regardless.**
```
ddev blt ds --site hawkeyemarchingband.uiowa.edu && ddev drush @hawkeyemarchingband.local config-split:activate p2lb && ddev drush @hawkeyemarchingband.local cim -y && ddev drush @hawkeyemarchingband.local uli node/add/page
```
- Add a new page. Edit it a few times to create some revisions.
- https://hawkeyemarchingband.uiowa.ddev.site/admin/content/sitenow-converter
- Convert all nodes
- Publish all converted nodes
```
ddev drush @hawkeyemarchingband.local sitenow_p2lb:cleanup
```
- Confirm that the page you created has not had its revisions removed.

<details>
<summary>Original problem/testing instructions</summary>

# Problem

New page content that is created with the p2lb converter on thinks it is a v2 page the next time it is edited because a v3 revision id is never set for the node.

By setting a numeric value for this field for new content, the existing logic works. The v3 page will show in the converted section but reverting it will have no effect. This field is removed when the converter is removed so no bad data will persist after conversion.

For existing nodes, It looks like nodes that are _not_ empty when it comes to layout_builder__layout but _are_ empty when it comes to field_v3_conversion_revision_id. Setting it to the current revision id allows p2lb:clean up to successfully delete any v2 paragraphs that got caught in the mix.

# To Test

On `main` sync law `ddev blt ds --site=law.uiowa.edu` and create a new page. You see the body field when saving the first time, but then go and edit it after saving. Because the field was never set, you get the v2 paragraphs fields and it thinks the page has not yet been converted.

Create a temp view that has the following filters:

Content: Content type (= Page)
Content: Conversion revision ID (empty)
Content: Layout (not empty)

See the results if any. If Law, there should be a couple.

On `hotfix_p2lb_new_page_default` run `ddev drush @law.local updb && ddev drush @law.local cim`

First check content overview to see if any page nodes were updated. They should have "automated save" revision entry by user 1. The temp view, if it had any results, should now be empty.
Second create a new page again. Editing this new page should keep the body field and not show you the v2 paragraphs fields. Find the page in the converted section and try to revert back to v2. It silently fails without error to the user.

</details>


config/features/p2lb/field.field.node.page.field_v3_conversion_revision_id.yml
```diff
@@ -13,7 +13,9 @@ label: 'Conversion revision ID'
 description: ''
 required: false
 translatable: false
-default_value: {  }
+default_value:
+  -
+    value: '1'
```
pyrello: I don't think we need this. I think we just need to hook into the entity save process and check if the node is being created for the first time and if it is store the newly created revision ID as the value.

joewhitsitt: This seemed like a smaller footprint than custom code

joewhitsitt: node_presave doesn't have a revision id yet. entity_insert has the revision id but modifying the p2lb revision field would count as another save. we recently ran into this with ITS alerts with needing a temp field value and the nid at the same time.

by using the same dummy revision id it also allows us to track and identify which nodes started as v3 while p2lb is enabled. this data will be deleted as part of cleanup.

bspeare:
When creating a new page on law after running the update hook and config import, I get this warning at the top.  Any concerns or is this just an artifact that will stick around?

Screenshot 2024-06-10 at 3 24 37 PM

joewhitsitt:
It is just an artifact. We could expand our checks to differentiate and have different behavior here, but my main goal with this is to prevent v2/v3 data loss.

https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module#L54

bspeare:
Thanks, final question.  With the test pages that I created before the update hook, are those expected to still have the paragraphs after the hook and config update?

joewhitsitt:
>With the test pages that I created before the update hook, are those expected to still have the paragraphs after the hook and config update.

I appreciate you raising this question. I hadn't thought that much about this. There could be potential orphans caused by this.

By default, new node pages have the content blocks #access = FALSE which means that on save those won't get values. It is when the node is edited and saved without the revision id field set that paragraphs could be introduced. The update hook creates a new revision and I am not 100% sure what happens paragraphs-wise.

joewhitsitt:
Pushed a change to unset the field basically. There seems to be 300+ orphaned paragraphs entities that don't get cleaned up on Law and and cleanup fails. Not sure if they are related to the nodes in question or not..

joewhitsitt:
I was able to finish law's p2lb cleanup by setting existing with the loaded revision id.

pyrello:
> > With the test pages that I created before the update hook, are those expected to still have the paragraphs after the hook and config update.
> 
> I appreciate you raising this question. I hadn't thought that much about this. There could be potential orphans caused by this.
> 
> By default, new node pages have the content blocks #access = FALSE which means that on save those won't get values. It is when the node is edited and saved without the revision id field set that paragraphs could be introduced. The update hook creates a new revision and I am not 100% sure what happens paragraphs-wise.

Essentially, if anyone has created any paragraph content for a page, I think that we should just treat it as V2 content. It gets added to the list of content that can be converted and goes through the conversion process. If we're not converting it, then I think our only other option is just to delete it, which would result in data loss for those pages. I don't think we should convert it via update hook because that violates the whole premise of allowing the customer to control that process. So, in any case where a page has paragraph content, it should not be affected by the update hook.

pyrello:
So, I like the approach of setting a default value for `field_v3_conversion_revision_id` and going back through with an update hook to retroactively set that value where it should be set (see my previous comment for caveats). I do not think the default should a value that could be an actual node revision because of code like the following, which would potentially return the wrong result (rather than `NULL`):

https://github.com/uiowa/uiowa/blob/912abdf87afbcb1deaa35950c553c90d194f11a1/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module#L1350-L1374

My idea for this is to use a non-numeric value as the value that gets set for new content (I'm tentatively using `v3_new`). This gives us a way to differentiate content that has been converted from content that was created since P2LB was turned on. This will allow us, for example, to exclude these items from showing in either the converter or converted lists.

joewhitsitt:
> > > With the test pages that I created before the update hook, are those expected to still have the paragraphs after the hook and config update.
> > 
> > 
> > I appreciate you raising this question. I hadn't thought that much about this. There could be potential orphans caused by this.
> > By default, new node pages have the content blocks #access = FALSE which means that on save those won't get values. It is when the node is edited and saved without the revision id field set that paragraphs could be introduced. The update hook creates a new revision and I am not 100% sure what happens paragraphs-wise.
> 
> Essentially, if anyone has created any paragraph content for a page, I think that we should just treat it as V2 content. It gets added to the list of content that can be converted and goes through the conversion process. If we're not converting it, then I think our only other option is just to delete it, which would result in data loss for those pages. I don't think we should convert it via update hook because that violates the whole premise of allowing the customer to control that process. So, in any case where a page has paragraph content, it should not be affected by the update hook.

The problem that led to this issue is that they created a v3 page, edited it later on to find the paragraph fields visble, saved it which saved those empty v2 items, then saw the "you should convert this" which dumps all of the v3 stuff they had in layout builder in favor of empty v2 converted stuff. Then they start reverting/converting and getting into a mess.

The commit https://github.com/uiowa/uiowa/pull/7786/commits/2c85e1ce0e20c963021ad53559fa68035c904577 handles the issue above that was raised/described.

joewhitsitt:
> So, I like the approach of setting a default value for `field_v3_conversion_revision_id` and going back through with an update hook to retroactively set that value where it should be set (see my previous comment for caveats). I do not think the default should a value that could be an actual node revision because of code like the following, which would potentially return the wrong result (rather than `NULL`):
> 
> https://github.com/uiowa/uiowa/blob/912abdf87afbcb1deaa35950c553c90d194f11a1/docroot/modules/custom/sitenow_p2lb/sitenow_p2lb.module#L1350-L1374
> 
> My idea for this is to use a non-numeric value as the value that gets set for new content (I'm tentatively using `v3_new`). This gives us a way to differentiate content that has been converted from content that was created since P2LB was turned on. This will allow us, for example, to exclude these items from showing in either the converter or converted lists.

If we want to approach it that way, i am okay with it, but there will be more code to touch then as there is at least one numeric check elsewhere in p2lb related to the message on the form.

pyrello:
>If we want to approach it that way, i am okay with it, but there will be more code to touch then as there is at least one numeric check elsewhere in p2lb related to the message on the form.

I think we're going to have to touch more code no matter what. The hotfix as currently formulated has a couple downsides and still the potential for data loss:
- It doesn't prevent new nodes (created since P2LB was turned on) from appearing in the converter or converted lists.
- This means that a person could attempt to rollback a new node.
- Setting the protected revision ID to 1 means that the code snippet I included above could return the node revision for `vid` 1.
- The combination of the previous two bullet points means that someone could "revert" a new page to a revision that is a different page (probably the first revision of the home page that gets populated when the site is installed).
- It creates two different states for "new" nodes: Any created before the update hook will be populated with an actual revision ID and treated as converted content going forward. Any created after the update hook will continue to have 1 as their value.

pyrello:
> The combination of the previous two bullet points means that someone could "revert" a new page to a revision that is a different page (probably the first revision of the home page that gets populated when the site is installed).

I tested this and my assumption was incorrect. The page just fails to revert.

>The commit https://github.com/uiowa/uiowa/commit/2c85e1ce0e20c963021ad53559fa68035c904577 handles the issue above that was raised/described.

When I revert a page that has a revision ID set as in the commit above, it goes back to an "unconverted" state, which means it could be "converted"

> which dumps all of the v3 stuff they had in layout builder in favor of empty v2 converted stuff

pyrello:
I was trying to understand better whether the query that @joewhitsitt was using in the update hook was actually a foolproof way to get the list of nodes that we want to affect. This led to me getting the most recent revision of all nodes and checking their properties to better understand the group. Here's what I came up with:
```
page_node_count = 234
no_fallback_revision_id = 24
fallback_revision_id = 210
no_layout = 21
layout = 213
draft = 172
published = 62
```
Based on this, it seems that there are 3 nodes sitting at the intersection of no fallback revision ID and having a layout. 1 is a test page that I created locally. Here are the other two nodes:
- https://law.uiowa.edu/faculty-and-scholarship - this is an older page, first created in 2020. It appears to have a layout but doesn't appear to have a revision where it was converted. This is a head-scratcher. No idea how this is possible.
- https://law.uiowa.edu/experiential-learning/citizen-lawyer-program - this seems to have been created more recently. It is one of the pages that has been the source of trouble with this site and has multiple revisions that indicate that it was converted.

I was hoping to find conditions that would specifically allow me to target new nodes that had been created since P2LB was turned on. But the above results seem to indicate that I have not found it yet.

joewhitsitt:
On Monday, this was the list of nodes that returned in my views query:

https://law.uiowa.edu/node/2886
https://law.uiowa.edu/node/3036
https://law.uiowa.edu/node/10721
https://law.uiowa.edu/node/10756
https://law.uiowa.edu/node/10751
https://law.uiowa.edu/node/10771

Since then some of those got retouched/converted/messed up/etc. Sorry if this causes more confusion than clarity.

pyrello:
Added this note to the test instructions: _Running the cleanup script is currently not working on HMB for reasons that I'm not entirely clear about. I think we could potentially cover the code change related to that with a code review since it seems like we'll be investing some more time in fixing it regardless._

joewhitsitt:
Haven't pulled down HMB yet to test, but I recall this: https://github.com/uiowa/uiowa/issues/6859#issuecomment-1783475036

joewhitsitt:
Here is the Canvas thread: https://iowaweb.slack.com/archives/C05NDUQ7VKN/p1717774914698279

joewhitsitt:
pulling to test

joewhitsitt:
i was able to cleanup HMB. noticed with law that it was actually the new pages that are preventing cleanup because page content section blocks do get saved even though #access is false. going to push a change to adjust this.

this could be why Sean wasn't able to cleanup HMB (created a new page on the branch) and Ben was (didn't create a new page, just tested against existing)?

using engagement.uiowa.edu to test and law to confirm.

joewhitsitt:
This has my ✅ approval. Need a non-author though.

cory-skeers (approved):
Reviewed code and looks good. Others have tested and given ✅

PR #7799, opened 2024-06-13
Classrooms - import iatl building

joewhitsitt:
- classrooms_core:buildings
- config-split:export site


briand44:
FYI, rolling this out as part of our normal release schedule is fine. It is not needed sooner than that.

bspeare (approved):
Looks good!

PR #7801, opened 2024-06-13
Entity Usage Delete

### LINKED ISSUE #7644: Probe - Entity usage seems to still be holding on to media usage
# Problem
Three know situations now where after cleaning up referenced revisions, the media still shows usage to the node(s).

- Setting up ned.uiowa.edu behind authentication and deleting default files
- Education (single file two pages)
- Rhetoric (multiple files one page)

# Steps to reproduce
Try delete files as described by our documentation (https://sitenow.uiowa.edu/documentation/media#delete-media)

# Workaround

I've been pulling the site down locally, finding the corresponding `fid` for each media entity id using the `file_usage` and `file_managed` tables.

Locally, I try using `drush entity:delete media {comma-separated media ids}` and then expecting the files in `file_managed` to make sure their status is set to `0`

# More info

It doesn't seem to show up in find text but I have had some results in finding the media in various tables `paragraph__field_collection_body` like `media/123` but then it is kind of a pain to track back from there.

# Proposed solution

- [ ] Look for upstream issues with entity_usage, linkit_media_library or Drupal 10.2 as this wasn't an issue when we released https://github.com/uiowa/uiowa/pull/7389
- [ ] ?

### LINKED ISSUE #7768: Revisit entity_usage configuration
We haven't looked at `entity_usage` configuration in what is being tracked in quite some time, and it may be time to revisit and see if reducing what is tracked (e.g. crop, aliases) makes an impact in db sizes without detriment in what is no longer being tracked.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7768
Resolves https://github.com/uiowa/uiowa/issues/7644

Sorry, not sorry I've added a bit of scope-creep to this solution.

With 7768 we blindly accepted some entities to be tracked with entity_usage which is adding unnecessary rows to the table but is otherwise no visible to users. This also starts to cover fragments/layout builder which were previously unaccounted for.

Updated node_revision_delete because we had to upgrade to their rewrite for d10 support? in which some of the previous functionality was lost. Moving from an alpha to an RC brings back the "delete prior revisions" functionality which makes it easier for folks to clean up revisions before trying to delete media.

# How to test

- On main, `ddev blt di --site=default` and try to delete media associated with layout builder. Ya can't. Try to do the same for a v2 site in which the media is nested within some paragraphs. Ya can't.
- On `entity_usage_delete` try the delete operations again after doing the normal expected workflow. https://sitenow.uiowa.edu/documentation/media#delete-media, you should be able to. Also note the ease in which you can delete prior revisions all in one go versus having to delete each revision one by one.
- Go create a prefooter textarea with a media item added to it. On the media item you should see that there is entity usage coverage there now for the block_content, but that you can't delete the media until it is removed from the fragment. Fragments have revisions ui but only one revision is tracked in the ui.
- For each media that you are able to delete. The corresponding file_managed row should get a status of 0 which means it will be cleaned up in about six hours on cron. file_usage should also clear of the `fid` as media is no longer tied to it. You can expedite the process by changing the modified timestamp and running cron manually.
- To understand better the changes to config, I found it helpful to look at the entity_usage table on the default site. There you will see what it means by source/target entities.


joewhitsitt:
Need to dig deeper to understand entities other than node should trigger the disabled button.

joewhitsitt:
Taxonomy will be revision-able in D11 https://www.drupal.org/project/drupal/issues/2936995 but without a module like node_revision_delete I doubt we would use it and probably block access to it like we did for Media. https://github.com/uiowa/uiowa/pull/4719

bspeare:
Pulling to test.

bspeare:
For both fragments and blocks, I'm noticing the scenario listed below:

**Delete**
- Add media to text area block, delete block, delete page revisions, go to `/media/*/delete` and can delete as expected. 

**Can't delete**
- Add media to text area block, delete image from text area and replace with text update, delete page revisions, go to `/media/*/delete` and can't delete as expected. 

Is this expected that the block needs to be deleted in order to delete the media?

joewhitsitt:
>Is this expected that the block needs to be deleted in order to delete the media?

Not ideal either, but it looks like the relationship to the block by way of the node is similar as a published block.

Until Drupal fixes the block content issue upstream, I'm not sure we can fully cover this without doing more intensive queries.

Pulling down to see if it there is an easy extension of this to implement

joewhitsitt:
Yeah, to go further, we would then have to query the block_content table to get the revision id and/or uuid and trace it back to whether the node has a revision that can be reverted.

Reducing coverage of entity usage could be something to consider, but I think it provides enough value in other ways to keep it.

We could update the known issue (if we ever created one) or I can update the documentation around deleting media to mention this use-case.

Open to suggestions.

bspeare:
I think updating the documentation to specify that the block must be deleted makes sense.

bspeare:
Is it possible to unset the warning message when the recorded usages are empty?
Screenshot 2024-06-20 at 10 45 04 AM

joewhitsitt:
>Is it possible to unset the warning message when the recorded usages are empty?

Well, technically if that message is showing there are usages but I get what you are saying.

I recall doing some messenger altering with Scheduler and will try to unset it.

And I will draft an additional snippet covering the issue with block content

joewhitsitt:
easier than i thought since they didn't use the messenger system...

here is the draft: https://sitenow.uiowa.edu/node/171/latest

bspeare (approved):
Great bug fix and love the new revision delete options!

joewhitsitt:
published the draft

PR #7784, opened 2024-06-07
[Registrar] CeCredential Validation module

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7724. 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local en cevalidationsr -y && ddev drush @registrar.local config:import --source ../config/sites/registrar.uiowa.edu --partial -y && ddev drush @registrar.local cim -y && ddev drush @registrar.local uli /node/316/layout
```
1.  Check phpcs warnings. 
2. Add a "More.." block to /node/316/layout.
3. Add "CeCredential Validation Scholar Record Form block" and save page. 
4. Test form validation:
   - Submit empty form
   - Submit half-filled form
   - Submit CeDiD = `1628-A38C-J8E8`
   - Submit bad CeDiD = `1628-A38C-J8E9`
   - Confirm id is rendered with dashes as you type.

PR #7804, opened 2024-06-13
[Registrar] Added status message library so we can dismiss error alert on form

bspeare:
Noticed that I couldn't dismiss the form status message as an anonymous user.  This adds the library to the form. 

# How to test
- Code review or follow instructions on https://github.com/uiowa/uiowa/pull/7784.


cory-skeers (approved):
Nice catch! Looks good, testing worked as expected.

PR #7813, opened 2024-06-17
[IWP] Up items per page for writer bios

### LINKED ISSUE #7795: Increase number of items per page for writers bios
# Story

As an IWP site owner, I would like to increase the number of items per page so that visitors can see all residency participants for a given session without having to page.

# Additional Info

https://iwp.dev.drupal.uiowa.edu/residency/participants

They want to be able to see all participants for a given session without having to page. Currently there is the ability to search by `- Any -` Session so we can't just show all items or we would potentially be trying to display over 1600 items on a single page. They mentioned increasing the number of items per page as a potential solution here. They thought 40 would cover most every session. 50 they think would cover every case.

cory-skeers:
Resolves #7795 

# How to test

Code review, and/or pull down and check out /residency/participants (if so, you'll likely want to sync from dev for the migration)

### DIFF
--- a/config/sites/iwp.uiowa.edu/views.view.writer_bio.yml
+++ b/config/sites/iwp.uiowa.edu/views.view.writer_bio.yml
@@ -78,7 +78,7 @@ display:
         type: full
         options:
           offset: 0
-          items_per_page: 10
+          items_per_page: 50
           total_pages: null
           id: 0
           tags:


GaryRidgway:
Pulling to test

GaryRidgway (approved):
Change observed and tested after syncing from dev.

Issue #7683, opened 2024-05-15
Investigate performance improvements for article list blocks


briand44:
256 MB memory limit on uiowa03 didn't solve now.uiowa.edu/campus listblock

https://iowaweb.slack.com/archives/CSQA18BK4/p1715450560643829

briand44:
Grooming conversation.... The Iowa Now example is paging and still running into memory limits. We should investigate and see if we can understand what is going on here. Not committing to solving as part of this.

GaryRidgway:
Re-posting from slack for posterity:

So taking a look at the cache tags for the view for listblock (Which is what is overflowing), it seems like the media tags are what is growing exponentially.

This is being called at `/mnt/www/html/uiowa03/docroot/core/lib/Drupal/Core/Cache/Cache.php`
[Take a look here for what I found.](https://codepen.io/purple_salmon/pen/YzbrXep)

GaryRidgway:
That the error is bubbling up through a few different functions, so the top level functions that we are seeing in new relic are probably not the actual issue, just where the app is running out of memory.

From some more digging it seems as though the error is coming up through the `fieldBlock()` call in `docroot/modules/contrib/ctools/modules/ctools_views/src/Plugin/Display/Block.php`. Which could mean that there is something going on with the exposed filters?

GaryRidgway:
Removing the filters from the back end views UI did not let the views UI load the preview, so perhaps it is something else.

cory-skeers:
Adding the little bit of exploration I've been able to do using xhprof memory profiling, as best I have been able to piece together.

The primary memory usage seems to build up from the database get and getMulti when pulling in results check for cached versions prior to going to the database. Fetching this cache version does a lot of `unserialize` calls, and these results get held in memory as the next items get called. Changing what gets displayed from the view doesn't appear to make a difference (eg changing the view to display only a single field rather than rendering the content teaser display results in essentially the same memory usage).

There also doesn't appear to be any difference between whether AJAX is enabled or not. The precise functions that are called in the stack vary slightly, but both end up in similar 'check the cache and unserialize or check the database' buildups of memory.

[...]

PR #7819, opened 2024-06-18
Performance cleanup

pyrello:
Related to #7683

# How to test
- For each of the standard list block types, ensure that limit, limit per page, and paging is still working as expected and hasn't changed for existing blocks.
- Ensure that breadcrumb and caption are still displaying for featured images when they are displayed on a full node.

## Query logging
```
ddev general_log
```
Visit a page locally or run a command (in a different terminal window) and observe the overwhelming amount of query logs. Cmd-c to stop logging.


joewhitsitt:
pulling to test. should we extend this to the events listblock?

joewhitsitt:
Looks like we are setting a default in code for items per page rather than config

https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L301

>// This case should fall through to the next
>// to set the default items_per_page.

I don't think the fall through switch statement setup here is working.

config/default/views.view.people_list_block.yml
```diff
@@ -687,19 +687,19 @@ display:
...
           total_pages: null
           id: 0
           tags:
             next: 'Next ›'
             previous: '‹ Previous'
           expose:
-            items_per_page: true
+            items_per_page: false
             items_per_page_label: 'Items per page'
             items_per_page_options: '9'
```
joewhitsitt:
```suggestion
            items_per_page_options: '10'
```

joewhitsitt (approved):
I'd be curious to know what can be done with the general_log outputting like that. My terminal was overwhelmed and put a limit, not letting me scroll back to the beginning of the command. That was just one page load (sandbox home).

Though I think the config is good to keep similar here, it is the defaults set in code that control the items_per_page.

Article caption/breadcrumb/featured image still works.

joewhitsitt:
deleted caches and restarted the build

pyrello:
> I'd be curious to know what can be done with the general_log outputting like that. My terminal was overwhelmed and put a limit, not letting me scroll back to the beginning of the command. That was just one page load (sandbox home).

I had to switch to Apple's terminal to be able to see all the output from the command. PhpStorm's terminal has a maximum output that gets truncated after its exceeded. My thought is that it would be even better to use the Console app or some other log viewer that could allow you to filter results more easily.

> Though I think the config is good to keep similar here, it is the defaults set in code that control the items_per_page.

The views config changes help with the potential confusion of visiting the Views UI and seeing different results there than the default block settings. It will never actually have an effect on the block output.

PR #7828, opened 2024-06-19
[Bug] Menu toggle fix for multiple menu blocks on a page

### LINKED ISSUE #7780: [Bug] Multiple menu blocks on mobile all display collapsed next to each other regardless of position on the page
# Problem
If multiple menu blocks are placed on the page, when viewing on mobile the collapsed version of the menus all display in the same spot on the page regardless of where they actually sit in the content. All collapsed menus render at the location of the first menu block on the page.

Noting that it is not a common practice to use multiple menu blocks on a page, so this is likely a very low priority bug.

# Steps to reproduce
View https://sandbox.prod.drupal.uiowa.edu/about-1/menu-variants in a mobile device width to see the behavior.

# Proposed solution
TBD

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7780

# How to test

```
ddev blt frontend && ddev blt ds --site=clas.uiowa.edu && ddev drush @clas.local uli /node/726/layout
```
1. Code review - are there custom unique ids for blocks?  I couldn't find one an opted for a counter [variable](https://github.com/uiowa/uiowa/blob/8b156f1435519d2a88742aee8864f1c396e8a002/docroot/themes/custom/uids_base/uids_base.theme#L858).
1. Add an additional horizontal menu block further down the page.  Set the block title to something custom. 
2. Save the page and collapse for mobile and see if the mobile menu's are positioned where they are placed in the layout and that they are functioning.

### DIFF
--- a/docroot/themes/custom/uids_base/assets/js/accessible-menu.init.js
+++ b/docroot/themes/custom/uids_base/assets/js/accessible-menu.init.js
@@ -8,7 +8,7 @@
       if (this.executed) {
         return;
       }
-      
+
       const menuWrappers = context.querySelectorAll('.menu-wrapper--horizontal');
 
       // Bail early if no menu wrappers are found.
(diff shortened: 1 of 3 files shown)


cory-skeers (approved):
Reviewed and tested, and everything checks out. Nice fix!

> Code review - are there custom unique ids for blocks? I couldn't find one an opted for a counter [variable](https://github.com/uiowa/uiowa/blob/8b156f1435519d2a88742aee8864f1c396e8a002/docroot/themes/custom/uids_base/uids_base.theme#L858).

I believe these are all dynamic blocks and don't receive any unique identifier unless provided one by something like this approach.

In the case where an ID is added by the editor, we could use that, but that ID still gets used as expected with this approach, and this way we cover both those and blocks that haven't been given one. I don't know that there'd be anything to gain by looking for that first if we need a counter anyway.

bspeare:
Thanks @cory-skeers!

PR #7841, opened 2024-06-24
remove path access on media for non-admins, resolves #4613

### LINKED ISSUE #4613: Disable URL Aliases for media
## Story

As an editor, I won't be tempted to change a media entity alias and cause trouble by not being able to access the media entity edit screen.

## Problem

Changing the media url alias prevents webmaster/etc from accessing the media edit form.

I can't think of a use-case where the media alias should be changed.

## Proposed Solution

- [x] Prevent webmasters and below from modifying the alias. Allow admins to access it so they can fix media paths that have been modified.

joewhitsitt:
# How to test

As a webmaster, you don't see

- Media path alias inputs/container, you can still save/edit a media entity without issue.
- Theme settings (logo, favicon, layout)
- Webform wrapper, element, label attributes
- ~~Footer Contact Information headline field~~ this was a test. no users should be able to see the headline field!
- Footer Contact Information text formats besides minimal _and lower_

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.profile
+++ b/docroot/profiles/custom/sitenow/sitenow.profile
@@ -502,6 +502,11 @@ function _sitenow_node_form_defaults(&$form, $form_state) {
 function sitenow_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   $form_object = $form_state->getFormObject();
 
+  /** @var Drupal\uiowa_core\Access\UiowaCoreAccess $check */
+  $check = \Drupal::service('uiowa_core.access_checker');
+
+  $access = $check->access(\Drupal::currentUser()->getAccount());
+
   if (is_a($form_object, ContentEntityForm::class)) {
     /** @var \Drupal\Core\Entity\ContentEntityForm $form_object */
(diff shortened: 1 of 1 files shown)


pyrello:
Pulling to test.

pyrello:
> Footer Contact Information headline field

I'm not finding this even as an admin... where should I be looking?

> Footer Contact Information text formats besides minimal

Screenshot 2024-06-25 at 2 31 44 PM

joewhitsitt:
Sorry, the original help text was covering a whole section of code.

```
        // For Footer Contact Information, limit non-admins
        // to minimal and remove headline field.
```
The text format in that screenshot is as expected and the headline field is removed for all users.

pyrello (approved):
Looks good. Thanks for the update, @joewhitsitt!

Idea for a future improvement: An access procedural function that takes a form array fragment as its argument and encapsulates getting the access checker service and running the check before setting that array fragment's `'#access' => FALSE`. Then we could just call that helper function and reduce some of the code duplication.

joewhitsitt:
yeah, main goal was the bug fix. reducing a few lines seemed overdue. further refinement sounds nice but for another day.

Issue #7822, opened 2024-06-19
commencement import


briand44:
Josh said he has created some new commencement events on content hub and they are not showing on the new website.

Example of content hub event https://content.uiowa.edu/143441/college-pharmacy-doctor-pharmacy-commencement-ceremony

Verified the event is approved and has the unit "Commencement" (7266) and the type "Commencement" (355).

briand44:
Grooming conversation... Not sure if we ever had a successful import in the production environment so sizing a little bigger based on that.

e-marie-w:
The scheduled jobs were using the old alias name for the event import. I've updated them to use the new. We should be able to see if it worked later this afternoon.

e-marie-w:
Update: event importer now running fine and the migrated duplicates were removed. I noticed that paragraph breaks and link tags were being lost, however, so fixing in #7849.

PR #7849, opened 2024-06-25
[Commencement] event importer fixes

e-marie-w:
Sort of relates to #7822 

# How to test

- ~~Code review - is this the best way to do this? I tried doing this a different way (see commit number 4) but couldn't get that way to work correctly.~~
- Note how events like https://commencement.uiowa.ddev.site/ceremony/spring-2025/college-liberal-arts-and-sciences-9-am-commencement-ceremony have everything in one paragraph and no links, vs. how it is on the [events calendar](https://events.uiowa.edu/70820)
- `ddev drush @commencement.local import-events`
- Spot check a few different events. Paragraph breaks and links should now appear correctly in the details field.

### DIFF
--- a/docroot/sites/commencement.uiowa.edu/modules/commencement_core/src/EventItemProcessor.php
+++ b/docroot/sites/commencement.uiowa.edu/modules/commencement_core/src/EventItemProcessor.php
@@ -14,6 +14,8 @@ class EventItemProcessor extends EntityItemProcessorBase {
    */
   protected static $fieldMap = [
     'title' => 'title',
+    'body' => 'description',
+    'body:format' => 'format',
     'field_event_contact' => 'contact_name',
     'field_event_contact_email' => 'contact_email',
     'field_event_contact_phone' => 'contact_phone',
@@ -26,23 +28,5 @@ class EventItemProcessor extends EntityItemProcessorBase {
     'field_event_when:end_value' => 'end',
     'field_event_when:duration' => 'duration',
   ];
-
-  /**
-   * Process the body field.
-   */
-  public static function process($entity, $record): bool {
-    $updated = parent::process($entity, $record);
-
-    if (isset($record->description)) {
-      // Set both value and format for the body field.
-      $entity->set('body', [
-        'value' => $record->description,
-        'format' => 'filtered_html',
-      ]);
-      $updated = TRUE;
-    }
-
-    return $updated;
-  }
 
 }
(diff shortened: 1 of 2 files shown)


joewhitsitt:
pulling to test

joewhitsitt:
>is this the best way to do this

Looks like this is how facilities' building combined hours was done. Future work could be done to process different field types beyond value only.

joewhitsitt:
>Commencement event content sync completed. 0 events were created, 58 updated, 0 deleted, 0 skipped. That is neat.

Concerned there is always updated nodes. Looks like the same thing on `main` though so unrelated to this change. Has this been mentioned before?

Update: The room number field is always updated when record is "" and entity is NULL.

e-marie-w:
> > Commencement event content sync completed. 0 events were created, 58 updated, 0 deleted, 0 skipped. That is neat.
> 
> Concerned there is always updated nodes. Looks like the same thing on `main` though so unrelated to this change. Has this been mentioned before?

That also worries me. I feel like this issue came up recently but I can't recall if it was about commencement or another site. Either way, I can poke at this some and see if there's an easy fix.

joewhitsitt:
Pushed a fix for body always being updated = true. The room number probably needs to be processed similarly and not included in the field map. Either that or the EntityItemProcessorBase needs to be updated to handle that blank/null non-value.

https://github.com/uiowa/uiowa/blob/3ced9f3cb2b92db0f9b4d7aa16fd876e28f43711/docroot/modules/custom/uiowa_core/src/EntityItemProcessorBase.php#L52

joewhitsitt:
@e-marie-w was this ready for review again or are you still tweaking it?

e-marie-w:
> @e-marie-w was this ready for review again or are you still tweaking it?

It's ready for review again!

joewhitsitt:
Works fine as-is, but a with a little nitpicking, this will simplify the comparison. We don't care that it `!is_null()`, only that it exists and isn't the value we want to later work with.

```php
if (property_exists($record, 'room_number') && $record->room_number === '') {
    $record->room_number = NULL;
}
```

As for Percy, hopefully just a momentary hiccup:

```
[percy] asset-discovery - Articles - 44.639s
[percy] Encountered an error taking snapshot: Articles
[percy] Error: Timed out waiting for network requests to idle.
[percy] asset-discovery - Article - 45.643s
[percy] Encountered an error taking snapshot: Article
[percy] Error: Timed out waiting for network requests to idle.
```

joewhitsitt (approved):

e-marie-w:
> Works fine as-is, but a with a little nitpicking, this will simplify the comparison. We don't care that it `!is_null()`, only that it exists and isn't the value we want to later work with.
> 

Oh that is way better, thank you! Change committed.

PR #7869, opened 2024-07-01
Added validation for offset and items per page.

### LINKED ISSUE #7848: Add validation to listblock items_per_page to prevent blank/non-values
# Problem

With https://github.com/uiowa/uiowa/pull/7819 and https://github.com/uiowa/uiowa/pull/7847 we discovered that a blank/non-value can be saved for the items_per_page override on listblocks. Previously this was a loophole for displaying all items. Now it defaults to 10 items.

The offset number field has some kind of validation in place to prevent save, but items_per_page does not. Offset does not use the normal validation messages however.

# Proposed Solution

- Prevent blank items_per_page values

# Possible additional slice

- Update hook to set a value for existing blocks

pyrello:
Resolves #7848 

# How to test

- Code review. Note: I removed our custom code to set offset in the block config because it gets set in the CTools block class we inherit from.
```
git checkout main && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli node/add/page
```
- Add a page.
- Add a list block to the layout and set it's items per page setting to ` ` (nothing). 
- Save the block **BUT DON'T SAVE THE LAYOUT.** Just leave the page open.
```
git checkout valid_items_per_page && ddev drush @sandbox.local updb
```
- Click "configure" on the block you created and check it's items per page. It should now be set to 50.
- Try putting in an empty value and confirm that you cannot save the form and it prints an error message (you may have to scroll back up).
- Trying putting in a numeric value outside the 1-50 range and confirm that you cannot save the form and it prints an error message.
- Trying putting in an empty value for the "offset" field and confirm that you cannot save the form.


joewhitsitt:
pulling to test

joewhitsitt (approved):
Nice solution and I agree with moving forward with the update hook.

Scheduled deployment given how much content this could potentially touch?

pyrello:
> Scheduled deployment given how much content this could potentially touch?

Sounds good.

Thanks for testing!

pyrello:
Going to merge this when it's ready and create a release.

PR #7878, opened 2024-07-02
[Commencement] Venue theming updates

### LINKED ISSUE #7877: Venue theme changes
https://commencement.prod.drupal.uiowa.edu/venue/hancher-auditorium

>We're wondering if we could reduce the margins between the sections a little bit? Especially between the first "ceremony info" section and the first collection header? Just seems like quite a bit of blank space.

>For the collection section titles/headers we're wondering if those could also be the underline style? I don’t know if it’s possible/easier to hide those titles but we could use the headers/style options in the text editor instead?

image

pyrello:
Resolves #7877 

# How to test

```
```
- Review https://commencement.uiowa.ddev.site/venue/hancher-auditorium, #7877, and https://commencement.prod.drupal.uiowa.edu/venue/hancher-auditorium to confirm that the visual changes fulfill what was requested.

### DIFF
--- a/config/sites/commencement.uiowa.edu/core.entity_view_display.node.venue.default.yml
+++ b/config/sites/commencement.uiowa.edu/core.entity_view_display.node.venue.default.yml
@@ -225,7 +225,8 @@ third_party_settings:
                     delimiter: ''
             weight: 1
             additional:
-              layout_builder_styles_style: {  }
+              layout_builder_styles_style:
+                block_margin_default_removed: block_margin_default_removed
             third_party_settings: {  }
         third_party_settings:
           layout_builder_lock:
--- a/docroot/sites/commencement.uiowa.edu/modules/commencement_core/commencement_core.module
+++ b/docroot/sites/commencement.uiowa.edu/modules/commencement_core/commencement_core.module
@@ -283,6 +283,7 @@ function commencement_core_preprocess_field(&$variables) {
       uiowa_core_set_attributes($variables, [
         'headline',
         'headline--serif',
+        'headline--underline',
         'h3',
       ]);
       break;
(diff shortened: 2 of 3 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
Checks out. Loses some of the hierarchal visual effect though. Noticed that they don't have the anchors setup for this page but that is a content issue right?

pyrello:
> Loses some of the hierarchal visual effect though.

What visual effects are lost?

bspeare:
When we use the `headline--underline` modifier class on a field it creates more space between the underline and the heading.   Ideally, we would unset the field classes on this `headline headline--serif headline--underline h3 field field--name-field-coll-section-title field--type-string field--label-hidden field__item` to accurately reflect the spacing used on all `headline--underline` headings.

Screenshot 2024-07-03 at 10 58 37 AM

joewhitsitt:
> > Loses some of the hierarchal visual effect though.
> 
> What visual effects are lost?

Just that before it looked more different between the h2 and h3. Now it just looks like a couple of pixels worth of spacing.

pyrello:
> When we use the `headline--underline` modifier class on a field it creates more space between the underline and the heading. Ideally, we would unset the field classes on this `headline headline--serif headline--underline h3 field field--name-field-coll-section-title field--type-string field--label-hidden field__item` to accurately reflect the spacing used on all `headline--underline` headings.
> 
> Screenshot 2024-07-03 at 10 58 37 AM

As far as I can tell, we don't have access to any of the `field` or `field-*` classes during preprocess. These classes get added in the template (in this case its the generic `field.html.twig` template). This means that to "remove" them, we'll need an overriding template for this field. Longer term, we should probably just fix the `field` classes so that they play nicely with the `headline` classes so we don't need to add extra template to unset them. Not sure how much work that is. Going to pause for additional feedback.

pyrello:
It doesn't look like the change in https://github.com/uiowa/uiowa/pull/7878/commits/a3fa1de2b4ff6a1931d6b320ff809b7cd407f647 made any difference that I could see with regard to the spacing for the h3's. There is still more space between the h3's and their underline than the h2 and its underline.

pyrello:
@bspeare confirmed that my change fixed the issue we were seeing before.

PR #7881, opened 2024-07-03
Hotfix - Views blocks check for exposed input

joewhitsitt:
# Background

https://iowaweb.slack.com/archives/C0164FNN1PV/p1720017748392929

# Test

```
```

Try placing a site custom views block like outages and degradations that doesn't have items_per_page or offset exposed to the editor.

Try placing a default listblock like articles that does have the fields exposed and try saving with a blank or invalid value (e.g. items_per_page = -2, 51).


pyrello:
Pulling to test.

pyrello (approved):
Nice, quick fix. Thanks, @joewhitsitt!

PR #7882, opened 2024-07-03
Admissions - Fix student profile grid (photo only) paths

joewhitsitt:
# How to test

```
ddev blt ds --site=admissions.uiowa.edu --sync-public-files
```

Try hitting these profiles via the random student profile grid

https://admissions.uiowa.ddev.site/student-profiles/katarina-okulich
https://admissions.uiowa.edu/student-profiles/anna-watts
https://admissions.uiowa.ddev.site/student-profiles/duong-jun-le


bspeare (approved):
Thanks for the quick fix!

joewhitsitt:
looks like the config is still there for the default/master display, but it is hidden from the UI and not used.

PR #7890, opened 2024-07-05
Academic calendar part deux

### LINKED ISSUE #7854: frontend work for academic calendar list
# Additional Info

Frontend work to create block display similar to prototype in https://github.com/uiowa/uiowa/issues/7851#issue-2373477052.

```[tasklist]
### Frontend tasks
- [x] display dates themed like prototype using recent UIDS date card stuff
- [x] implement search/filtering similar to prototype
- [x] Add loading spinner so user knows data is being loaded `<span class="fa-solid fa-spinner fa-spin"></span>`
- [x] Subsessions display with "group by month" unchecked is displaying Summer subessions after Fall
- [x] Swap out jQuery for standard JavaScript?
- [x] Lazyload in items to reduce page load time?
- [x] Sessions filter also seems buggy - Try filtering for "Fall 2024" with group by month checked
```

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7854

# How to test

- Sync registrar.uiowa.edu
- Place the Academic Calendar block on https://registrar.uiowa.ddev.site/academic-calendar
- See that the date cards are using the new UIDS component definition/styles.
- See that on initial load there is a spinner that is replaced by results.
- The group by month can be configured to show/hide as well as be selected as default. Show/hide previous works the same even if the group by month is not available.
- All of the filters trigger the necessary fetch and filter.
- If you happen to have VoiceOver on, even though it isn't in-scope, there should be announcements letting the user know that the results are changing along with basic details about what to expect.
- The jquery dependency has been removed, travis is passing.
- The subsessions are showing in the correct order
- The sorting is happening in PHP and is being cached.
- The date is now using a gold border instead of a full gold background.
- Items after the first day in a month do not have a date display circle, grouping them together.
- There is a day of the week display in the subtitle.
- Sessions in the `sessions` dropdown are sorted
- There are no errors when changing the settings of an Academic calendar block after it has already been placed on a page.


pyrello:
I'm pulling this to take a closer look at the JS. I was trying to review the changes since I last looked at it on Monday and its hard to get a sense of the whole thing just reviewing the code changes.

docroot/sites/registrar.uiowa.edu/modules/registrar_core/js/academic_calendar.js
```diff
@@ -9,59 +9,75 @@
         let uniqueSessions = new Set();
 
         // Cache objects for frequently used elements.
-        const $form = $('#academic-calendar-filter-form', context);
-        const $groupByMonthCheckbox = $('#group-by-month', context);
-        const $showPreviousEventsCheckbox = $('#show-previous-events', context);
+        const form = context.querySelector('#academic-calendar-filter-form');
+        const showPreviousEventsCheckbox = context.querySelector('#show-previous-events');
+        const categoryChosen = context.getElementById('edit_category_chosen');
```
pyrello: Curious about the use of underscores instead of dashes (like in the next item). I understand this is finding an ID that is defined somewhere else, but maybe we should change that value to use dashes as well?

joewhitsitt: Chosen adds that element I believe, and the others were pre defined. I am open to the change if it is desired.

joewhitsitt:
Keep running into this phpcs error `#lazy_builder callback 'registrar_core.lazy…' at key '0' is not callable.`

I tried moving stuff around so that array is not empty. I've see empty examples elsewhere. Passing more things runs into a "scalar" issue.

Might roll back https://github.com/uiowa/uiowa/pull/7890/commits/2ee98b8b357da2f1c0515835f1dd2d81aac13d59 since it didn't solve anything.

✅ https://github.com/uiowa/uiowa/pull/7890/commits/b5b1e14b666035fdf601f1feec4ee6b4db57b773

pyrello:
Noting my comments in https://github.com/uiowa/uiowa/issues/7854#issuecomment-2221681173 and making it known here also that @GaryRidgway and I are planning on spending some time tomorrow looking at some additional possible refactors.

GaryRidgway:
I am reverting this to draft because I am noticing that some of the filters are not working. Investigating now.

cory-skeers:
~The session sorting looks accurate for the dates, but in the select list, it goes summer > winter > fall > spring, where fall should come before winter.~

~For the block options for session(s) to display, it seems like "Current session" and "Current session, plus next session" are the same, both with the current and next session (eg selecting either today shows both the current Summer and next Fall, where I would expect "Current session" to only show summer and "Current session, plus next session" to include both).~

In the Layout Builder preview, the card display is slightly different with centered above date? (The question mark is because I could have sworn this wasn't the case earlier but am seeing it now...)
image

cory-skeers (approved):
> In the Layout Builder preview, the card display is slightly different with centered above date? (The question mark is because I could have sworn this wasn't the case earlier but am seeing it now...)

I think this difference was always there and I just didn't notice it right away. It'd be nice to have the card layouts match, but it's a minor detail that doesn't effect the end user's experience.

Percy build will need to be approved with the new UIDS hash. It's a very minor visual difference, but I didn't feel great about hitting the approval without knowing exactly where/why the difference is in place.

But...given those to caveats, I've tested this quite a bit and it is working great! 💯

Issue #7891, opened 2024-07-08
SSL *.uihealthcare.org on uiowa06


briand44:
> So on Wednesday I submitted to the UIHC host master requests for two uihealthcare.orgsubdomains they completed the task and they are pointing correctly:
> 
> https://respiratorycareseminar.uihealthcare.org/
> https://cpss.uihealthcare.org/
> 
> So I am not sure if you would like a formal request to add *.uihealthcare.org to the SSL on uiowa06 to cover the domains
> 
> the CNAMEs are pointing to the acquia-dlb.its.uiowa.edu

briand44:
Tried adding the wildcard domain but it won't work. Guessing at a minimum that wildcard domain needs to be added to our bucket. Max is scheduling some time with appropriate folks to discuss. Blocked until that conversation happens.

briand44:
No longer blocked. ISPO got approval from Healthcare leadership and allocated the domains to our bucket.

briand44:
Activated a new cert on uiowa06. This is done.

PR #7894, opened 2024-07-09
Provisions 07 09 24

bzalasky-uiowa:


joewhitsitt (approved):
Just connecting the dots here: https://github.com/uiowa/uiowa/issues/7891

medicine.uiowa.edu is already covered.

Looks good.

PR #7889, opened 2024-07-05
Use loggerChannelTrait again

cory-skeers:
Resolves #7597 

# How to test

The following sitenow and site drush commands have been updated:

- Uiowa Core: `uiowa_core:toggle-gtag`, `uiowa_core:custom_role_mappings`
- Sitenow p2lb: `sitenow_p2lb:cleanup`
- its.uiowa.edu: `its_core:alerts-digest`
- brand.uiowa.edu: `brand_core:lockup-digest`
- classrooms.uiowa.edu: `classrooms_core:buildings`, `classrooms_core:rooms_import`
- facilities.uiowa.edu: `facilities_core:buildings_import`, `facilities_core:projects_import`

For each, test that on a fresh site sync 

1. the command functions and you see terminal output
2. Enable dblog and re-run the command, and check that
   1. you see terminal output 
   2. watchdog has caught the log (eg `ddev drush @site.local watchdog-show --count=10` you see a line item for what you just ran, or at /admin/reports/dblog)
3. Clear cache on the site you're working with, re-run again, and check that 
   1. you see terminal output
   2. watchdog has caught the log

If the command ran properly in step 1, for steps 2 and 3 you only need to let it run far enough to get output and/or logging (for instance, facilities building import, you can cancel after the initial "starting" message)

PR #7901, opened 2024-07-11
add if statement to check if logger is null

bzalasky-uiowa:
Article migration was throwing an error `Error: Call to a member function notice() on null in Drupal\ccom_migrate\Plugin\migrate\source\CcomArticle->prepareRow()` on the metatag section of CcomArticle.php:

```
$this->logger->notice('Metatag @item: @value not migrated for node: @nid', [
            '@item' => $item,
            '@value' => $value,
            '@nid' => $row->getSourceProperty('nid'),
          ]);
```

I added an if statement to check if $this->logger was null around that block of code to fix the issue.

The error was thrown on this article: https://medicinepsychiatry.uiowa.ddev.site/news/2021/02/new-inpatient-psychiatric-unit-now-open

Have not been able to figure out what about this article has caused the issue.

### DIFF
--- a/docroot/modules/custom/ccom_migrate/src/Plugin/migrate/source/CcomArticle.php
+++ b/docroot/modules/custom/ccom_migrate/src/Plugin/migrate/source/CcomArticle.php
@@ -104,11 +104,13 @@ class CcomArticle extends BaseNodeSource {
         $this->migration
           ->getIdMap()
           ->saveMessage(['nid' => $row->getSourceProperty('nid')], $message);
-        $this->logger->notice('Metatag @item: @value not migrated for node: @nid', [
-          '@item' => $item,
-          '@value' => $value,
-          '@nid' => $row->getSourceProperty('nid'),
-        ]);
+        if ($this->logger != null) {
+          $this->logger->notice('Metatag @item: @value not migrated for node: @nid', [
+            '@item' => $item,
+            '@value' => $value,
+            '@nid' => $row->getSourceProperty('nid'),
+          ]);
+        }
       }
     }
     // If a date is specified in field_date,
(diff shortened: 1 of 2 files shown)


pyrello:
As best as I can tell from looking at this, the section of code that resulted in the error is just providing feedback about an article's metatag properties that were not migrated for whatever reason. There is probably an earlier section of code that is generating this list. Presumably the article in question has a metatag value that doesn't map to anything in the system and this code is supposed to print out those values so you can check them and take appropriate action, if necessary.

The reason the `logger` property is null is that the base class that the CcomArticle class is inheriting from has a confused implementation. We are using the `LoggerChannelTrait` but we also have a `$logger` property that never gets set to anything. Probably at one point, we were setting the `$logger` property.

To understand how this should implement the logging, I looked at #7889 and ultimately the `LoggerChannelTrait` to see what guidance it offers on its own usage. It contains this help text:

```
This utility trait should only be used in application-level code, such as
classes that would implement ContainerInjectionInterface. Services registered
in the Container should not use this trait but inject the appropriate service
directly for easier testing.
 ```

Based on this, I think we may be incorrectly using that trait for our `BaseNodeSource` class and should instead be injecting the service. There are also some instances I saw in #7889 where we are using both the trait and injecting the logger service dependency, so we may want to clean those up too at some point.

I pushed up a fix that should prevent this issue in the future.

pyrello:
@bzalasky-uiowa Any chance you are able to re-run the migration you were working on to see if my changes fix the issue?

bzalasky-uiowa:
@pyrello How can I reset it if I already have the articles migrated?

bzalasky-uiowa:
[warning] Array to string conversion CcomArticle.php:103
 [notice] Metatag robots: @value not migrated for node: 5531

PR #7905, opened 2024-07-11
[Housing] Generalize text on residence halls view page

e-marie-w:
The text on https://housing.uiowa.edu/residence-halls mentions a specific number of residence halls. That number does not match how many there will be this school year. Since this number will most likely continue to fluctuate this changes removes the number mention entirely. 

# How to test

Probably just code review is enough? If you wanna test though:

- `ddev blt ds --site=housing.uiowa.edu && ddev drush @housing.local uli /residence-halls`


joewhitsitt (approved):
I know the copy comes from the customer, but `located` in here sounds wrong now.

"we offer residence halls ~~located~~ on both the east and west side~~s~~ of campus"

PR #7910, opened 2024-07-12
[INRC] Inrc feedback

### LINKED ISSUE #7895: Feedback on Grants
# Stories

We would like the ability to easily select all counties when creating a grant so that we don't have to choose each county individually when the grant applies to all.

For grants that apply to all counties, we would like a nice way to display the counties so it doesn't just display a huge long list.

We no longer need to associate specific cities with grants, that field can be removed.

We would like the Grants search by name to also search the purpose/mission text area so that users can more likely find the appropriate grants if they can't remember the name.

# Additional Info

They started not selecting all counties because of how those are currently output. Instead they put text to say available to all counties. The problem with this is those grants won't display to end users filtering by a specific county. They would like to choose all counties when applicable and have them show up when filtering by an individual county.
```[tasklist]
### Tasks
- [x] Ability to easily select all counties
- [x] Make sure counties display nicely if all selected
- [ ] remove cities from grants
- [x] the search by name should also search the purpose/mission text area
```

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/7895. 

# How to test

- Code review 

```
ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli /iowa-grants-guide
```

- Confirm you can search for Grantor Name or body content in the Grantor name field. 

```
ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli /node/126/edit
```
- Confirm you can select all and deselect all Counties
- Confirm value is saved and edit the node again and the value is still shown as selected
- Confirm "All counties" selected outputs on the page as "All counties"
- Confirm all other county selections render as expected


cory-skeers (approved):
Reviewed and tested with locally migrated grants, and the updates work great!

bspeare:
Thanks @cory-skeers!

PR #7928, opened 2024-07-22
[Registrar] Course deadlines

### LINKED ISSUE #7885: Course Deadlines API calls and Filters
# Story

As a Registrar site owner, I would like the ability for users to view course deadlines and additional course information after choosing the particular  session, department, course and section of their course.

# Additional Info

Happy with their current functionality - https://registrar.uiowa.edu/course-deadlines
Would like the ability to make content edits that live on that page

There are off-cycle courses - we should display a message showing this course is an off-cycle course and we believe we need to add the business logic to figure out the last day to drop etc. based on the start date of that particular course's session. Logic should be the same as it exists on the D7 site.

Currently there is a data.its call - Joe's comment "course deadlines is calling a static list of building abbr. to building numbers for the table result output. we can do this outside of data.its"

**Filters** - filters update based on previous filter selection
- session
- department
- course
- section

**Info to Display**

- Course Name/Title
- Course Number -  e.g. ACCT:2100:0AAA
- Begin Date
- End Date
- Deadlines for this course - e.g. last day to add/drop
- Collegiate Office Contact
- Time and Locations - link to building
- Instructors

cory-skeers:
Resolves #7885 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli
```

- Edit a page's layout and add a Course Deadlines block, save
- Use the deadlines block and compare the (unthemed) results with https://registrar.uiowa.edu/course-deadlines
- Check for an off-cycle course like 'Summer 2024, ABRD, 3016, 0001' or others, and see that it has an "this is an off-cycle course" label
- Try different combinations of filling out the fields then changing one of the previous fields--results and fields below the changed one should reset.
- When choosing a course with only one available section, that section should autopopulate the section field and the deadlines info block should appear


pyrello:
Pulling to test.

pyrello:
This is looking really good and I'm ready to approve as is. However, I was wondering if it was feasible to have the section get "auto-selected" when there is only one option. I started tinkering a bit to see if I could figure it out. I mostly have it working but the default value isn't getting set for the section. Going to spend a little bit more time on it and if I can't figure it out, I'll move on.

cory-skeers:
> This is looking really good and I'm ready to approve as is. However, I was wondering if it was feasible to have the section get "auto-selected" when there is only one option. I started tinkering a bit to see if I could figure it out. I mostly have it working but the default value isn't getting set for the section. Going to spend a little bit more time on it and if I can't figure it out, I'll move on.

I'm really liking the update, but after the change I'm running into an error with the JS that is breaking the form reset functionality when changing a previous selection (eg changing department after results have been fetched leaves course, section, and the course deadlines for the previous selections)
image
image
I went back to a previous commit to make sure the error wasn't present previously, but I don't see anything from the code changes that I would've suspected would have effected it. I'm going to do a little more digging.

cory-skeers:
This is ready for review again. It's been updated to drop the additional JS, so full testing of the testing instructions and that the filters properly dis/enable upon selection would be appreciated!

joewhitsitt:
pulling to test

docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Plugin/Block/CourseDeadlinesBlock.php
```diff
@@ -0,0 +1,570 @@
...
+          $location = $this->t('@building @street @city @country', [
+            '@building' => $value->offsiteBuilding ?? '',
+            '@street' => $value->offsiteStreet ?? '',
+            '@city' => $value->offsiteCity ?? '',
+            '@country' => $value->offsiteCountry ?? '',
+          ]);
+        }
+        else {
+          if ($value->building) {
+            $lookup = json_decode(file_get_contents('https://data.its.uiowa.edu/maps/number-lookup'));
```
joewhitsitt, line 527: I think this is okay to launch with but we should revisit this as the list on data.its as it is static and hasn't been maintained.

joewhitsitt:
With all of the work we have put into the gpa_calculator, academic calendar to make it more accessible I want to pause to think about this block form a bit. With each selection we should announce the changes to the user to tell them to keep going or consider switching focus to the next required dropdown. At the end, tell that result data has been provided.

cory-skeers:
After starting to look into updates in accessibility for this, I'm starting to think it might make sense to wrap that into #7932. In particular, I'd worked some on getting the announcing to work for the final results block, and some of that started to depend on how `<div>`s were nested, which could once we start styling. 

E.g. When switching course sections, results were not being announced I _think_ because content wasn't being seen as having changed despite it being within an `aria-live` region.

docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Plugin/Block/CourseDeadlinesBlock.php
cory-skeers: I added a `@todo` in the code for this, and added a quick check so that, if it is out of date and cannot find a building code, it will still output the location sans link.

e-marie-w:
This is looking and working great! I have just a couple of questions/clarifications.

> Try different combinations of filling out the fields then changing one of the previous fields--results and fields below the changed one should reset.

This is _mostly_ working, but for me the the department is not being reset when the session is changed. 

> When choosing a course with only one available section, that section should autopopulate the section field and the deadlines info block should appear

The deadlines info block just means the result of the filters, correct? If so, it's working great! 

There's that 'See Note Below' text, I assume they'll just add a text area to the page for that? Like, it's not something we need to handle as part of the block.

cory-skeers:
> This is looking and working great! I have just a couple of questions/clarifications.
> 
> > Try different combinations of filling out the fields then changing one of the previous fields--results and fields below the changed one should reset.
> 
> This is _mostly_ working, but for me the the department is not being reset when the session is changed.

Yup! Lack of clarity in the instructions, but the departments aren't dependent on the session and can stay static even if the session changes.

> > When choosing a course with only one available section, that section should autopopulate the section field and the deadlines info block should appear
> 
> The deadlines info block just means the result of the filters, correct? If so, it's working great!

Mmhmm!

> There's that 'See Note Below' text, I assume they'll just add a text area to the page for that? Like, it's not something we need to handle as part of the block.

Uh, I kind of read over that and didn't really think about it. But it's only really applicable when there are results, so maybe it should be part of the block as well. I'll take another look at it.

e-marie-w:
> Yup! Lack of clarity in the instructions, but the departments aren't dependent on the session and can stay static even if the session changes.

Ahh okay! But should it still work? Right now if I switch the session and leave the department alone the course filter is unclickable. If I switch to a different department and then switch back to the original I will be able to click and see courses, so it's not a situation where no courses are available. (Which does happen, especially when looking at future sessions, but even then the course filter is clickable, it just has no options.)

cory-skeers:
> > Yup! Lack of clarity in the instructions, but the departments aren't dependent on the session and can stay static even if the session changes.
> 
> Ahh okay! But should it still work? Right now if I switch the session and leave the department alone the course filter is unclickable. If I switch to a different department and then switch back to the original I will be able to click and see courses, so it's not a situation where no courses are available. (Which does happen, especially when looking at future sessions, but even then the course filter is clickable, it just has no options.)

Ahhh...Yes, that should work, and needs an update then.

cory-skeers:
> > There's that 'See Note Below' text, I assume they'll just add a text area to the page for that? Like, it's not something we need to handle as part of the block.
> 
> Uh, I kind of read over that and didn't really think about it. But it's only really applicable when there are results, so maybe it should be part of the block as well. I'll take another look at it.

After looking at this again, I think your original thought makes more sense, so they can edit/update the text as needed, so I think this can stay as is for now. There may be some adjustment of this when it gets into their hands, but that can wait 'til then.


> > Ahh okay! But should it still work? Right now if I switch the session and leave the department alone the course filter is unclickable. If I switch to a different department and then switch back to the original I will be able to click and see courses, so it's not a situation where no courses are available. (Which does happen, especially when looking at future sessions, but even then the course filter is clickable, it just has no options.)
> 
> Ahhh...Yes, that should work, and needs an update then.

I've pushed an update which I think mostly covers this, but there is an odd case when the same course is available in multiple sessions (eg Acct 2100 is offered most sessions). In this case, the course remains selected, but Section is stuck as Disabled. Still poking a bit at this.

cory-skeers:
> > > There's that 'See Note Below' text, I assume they'll just add a text area to the page for that? Like, it's not something we need to handle as part of the block.
> > 
> > 
> > Uh, I kind of read over that and didn't really think about it. But it's only really applicable when there are results, so maybe it should be part of the block as well. I'll take another look at it.
> 
> After looking at this again, I think your original thought makes more sense, so they can edit/update the text as needed, so I think this can stay as is for now. There may be some adjustment of this when it gets into their hands, but that can wait 'til then.
> 
> > > Ahh okay! But should it still work? Right now if I switch the session and leave the department alone the course filter is unclickable. If I switch to a different department and then switch back to the original I will be able to click and see courses, so it's not a situation where no courses are available. (Which does happen, especially when looking at future sessions, but even then the course filter is clickable, it just has no options.)
> > 
> > 
> > Ahhh...Yes, that should work, and needs an update then.
> 
> I've pushed an update which I think mostly covers this, but there is an odd case when the same course is available in multiple sessions (eg Acct 2100 is offered most sessions). In this case, the course remains selected, but Section is stuck as Disabled. Still poking a bit at this.

@e-marie-w I think I've gotten this working in a way that the department can still stay selected when switching sessions, but the course and section reset and properly disable/enable (course stays enabled, but section returns to disabled until a course is chosen).

e-marie-w (approved):
Filters are working correctly and display all the desired information. Also, the results match the results of the old site. Awesome job!

PR #7935, opened 2024-07-25
[IWP] Media gallery migration

### LINKED ISSUE #7920: Create migration for IWP media gallery
# Story

As an IWP website owner, we would like to have our media gallery items migrated from our current site to our new site so we do not have to manually move all that content.

# Additional Info

Migrate all published YouTube content (~ 300. We are not migrating vimeo or its_hosted_video)

create redirects as part of the migration

We should bring over genre taxonomy terms during migration. Countries should already exist.

Shouldn't need to migrate into summary or link.

**Mappings - v3 -> v1**

- Title - title
- Featured Image - field_media_image - image
- Body - field_video_description - text area - I don't think there was a summary on D7
- Type - These should all be set as video
- Media - field_youtube_id - can we create a media entity for remote video with just a  youTube id?
- Country - field_media_country - Countries should already exist so we should just be mapping based on name
- Genre - field_media_genre - create terms through migration
- Year - field_media_year
- Authored on -

cory-skeers:
Resolves #7920

# How to test

```
ddev blt ds --site=iwp.uiowa.edu && ddev drush @iwp.local en iwp_migrate -y && ddev drush @iwp.local uli /admin/config/sitenow/sitenow-migrate
```

- Configure SiteNow Migrate following instructions for DevDesktop. 
  - Database: `iwp`.
  - User: `root`
  - Host: `host.docker.internal`
  - Port: `33067`
  - Base URL: `https://iwp.uiowa.edu`
- Check the migration status: `ddev drush @iwp.local ms`
- Run all migrations `ddev drush @iwp.local mim --all  `  
- or
- Import the media gallery: `ddev drush @iwp.local mim iwp_gallery`
- Import the redirects: `ddev drush @iwp.local mim iwp_gallery_redirects`
- Rollback `ddev drush @iwp.local mr iwp_gallery`

Go to https://iwp.uiowa.ddev.site/media-gallery to view all. 
Check against live site to make sure things migrated over correctly


pyrello:
Pulling to test.

briand44:
Makur mentioned that when she saw this in demo, it looked like the images were being cut off and not displayed correctly. She was wondering if there is anything we can do about those or if they are going to manually need to set focal point or update new images after the migration?

pyrello:
> Makur mentioned that when she saw this in demo, it looked like the images were being cut off and not displayed correctly. She was wondering if there is anything we can do about those or if they are going to manually need to set focal point or update new images after the migration?

Just based on reviewing what is in this PR, the logic for what image style derivative is being used is baked into code we wrote previously because I don't see anything in this PR that touches that. I don't recall if the old site was also using focal point to set the cropping on images, but we don't and have never brought that type of data over. No idea if the data from the old site would be directly usable or what we would need to do to get it, but it seems likely that it wouldn't be worth the effort. So much hunch is that the most straight-forward route involves the images having their focal points set on the new site.

pyrello:
I'm running into a duplicate entry error at about the 30% point in the gallery redirects migration. Maybe I missed something?

cory-skeers:
> I'm running into a duplicate entry error at about the 30% point in the gallery redirects migration. Maybe I missed something?

It looks like there are 3 resources on prod (eg https://iwp.prod.drupal.uiowa.edu/archives/media/map-2018-usman-ali-pakistan). If these get deleted prior to running the migration, I think the redirects migration should run cleanly.

I think we've run into this before and added a workaround so existing ones didn't need to be deleted, but that should only be necessary if we think we'll want to / need to keep non-migrated content along with the migrated.

pyrello:
> > I'm running into a duplicate entry error at about the 30% point in the gallery redirects migration. Maybe I missed something?
> 
> It looks like there are 3 resources on prod (eg https://iwp.prod.drupal.uiowa.edu/archives/media/map-2018-usman-ali-pakistan). If these get deleted prior to running the migration, I think the redirects migration should run cleanly.
> 
> I think we've run into this before and added a workaround so existing ones didn't need to be deleted, but that should only be necessary if we think we'll want to / need to keep non-migrated content along with the migrated.

I'm assuming this is going to be an issue running the migration in dev too? Should this content just be deleted now on the prod site?

cory-skeers:
> > > I'm running into a duplicate entry error at about the 30% point in the gallery redirects migration. Maybe I missed something?
> > 
> > 
> > It looks like there are 3 resources on prod (eg https://iwp.prod.drupal.uiowa.edu/archives/media/map-2018-usman-ali-pakistan). If these get deleted prior to running the migration, I think the redirects migration should run cleanly.
> > I think we've run into this before and added a workaround so existing ones didn't need to be deleted, but that should only be necessary if we think we'll want to / need to keep non-migrated content along with the migrated.
> 
> I'm assuming this is going to be an issue running the migration in dev too? Should this content just be deleted now on the prod site?

Yup, they'll either need to be deleted in each environment post-sync and pre-migrate, or just deleted in prod now to handle it for all.

pyrello:
It appears that there is no tag term field on the new Resource content type. The old content type had a tags field, which was populated. Just double-checking that this omission is intentional.

briand44:
Yeah, I know they didn't find the tags filter very valuable so they didn't need to keep that filter. Not 100% sure if they were ok with tags going away altogether but since they didn't bring it up when we demoed this work to them, I would move forward without accounting for tags. Thanks for checking!

pyrello (approved):
This all looks good. Nice work!

PR #7941, opened 2024-07-29
[Commencement] Collection item bug fix and headline style updates

### LINKED ISSUE #7936: Ceremony info collection item bug and header style changes
- >it looks like the Ceremony Info collection items are displaying as a descriptive list instead of accordions? I could've swore at one time those were displaying as accordions?
  - https://commencement.prod.drupal.uiowa.edu/ceremony/fall-2018/college-education-commencement-and-teacher-education-program-recognition
- >And I'm sorry I missed this, but would it be possible to change the Livestream and Ceremony Header style to the underline like we did on the venue pages?

e-marie-w:
Fixes #7936.

# How to test
- Visit an event page with ceremony information - like https://commencement.uiowa.ddev.site/ceremony/fall-2018/college-education-commencement-and-teacher-education-program-recognition#arriving-at-hancher-auditorium
- The previous Ceremony information heading should now be Additional Info for Graduates and Guests.
- Check that the style (serif, underlined) of the Livestream and Additional Info headings match the heading style on a venue page like https://commencement.prod.drupal.uiowa.edu/venue/carver-hawkeye-arena. 
- Check that any accordions under Additional Info now render like our other accordions


joewhitsitt:
Something isn't quite right with the accordion markup. Missing the id on the accordion wrapper that the data-parent info of the individual accordion items doesn't align.

Screenshot 2024-07-31 at 2 02 50 PM

joewhitsitt:
@e-marie-w I think since this is a field implementation, it needs something like docroot/sites/housing.uiowa.edu/modules/housing_core/templates/field--node--field-residence-hall-floor-plans--residence-hall.html.twig where it sets the data parent both for the wrapper and the items.

joewhitsitt:
wait nevermind. sorry. that example has a bunch of processing going on in addition to the template.

joewhitsitt:
Just pulled. Though this fixes the accordion wrapper id to match the accordion item data-parent attribute value, the accordion item id is the same for both Apparel and Arriving... `accordion-item-141-0` The delta is not incremented. The h2's above do have the friendly label-based ids.

e-marie-w:
Capturing some thoughts: When looking at other accordions on the site, I can see the venue (so, the collection section paragraph) accordion items properly iterate the IDs. The collection block accordions do not, and neither do the event accordions with the current code. When using the housing template above or similar templates another layer of accordions surround the accordion items. These extra layer accordions do have different IDs as they should based on the code, but they are unnecessary and the accordions we want still do not (probably because each one _is_ the 0 index item of its surrounding accordion). 

Now, there is a commencement_core_preprocess_paragraph__uiowa_collection_item function that is making changes for the collection section paragraph. If I remove this function the collection section accordions lose their unique IDs, but the collection block accordions gain unique IDs. Unfortunately the event accordions still do not, whether with the current code or code based on the housing template example.

So, two things to fix now!

Update: fixed it so collection block and collection section accordions both have unique item IDs. Still trying to figure out how to fix that for the event accordions.

joewhitsitt:
@e-marie-w I wanna say that this is commit is getting us closer. I haven't had time to figure out this bit though `'accordion_item_content': item.content|without('field_collection_headline'),` so it is still display all of the item.content in the accordion body.

Screenshot 2024-08-05 at 12 29 10 PM

Screenshot 2024-08-05 at 12 29 42 PM

joewhitsitt:
I am not sure if I understand everything from https://github.com/uiowa/uiowa/pull/7941#issuecomment-2265706378 so please take a look and let me know if there are still outstanding issues.

e-marie-w:
> I am not sure if I understand everything from [#7941 (comment)](https://github.com/uiowa/uiowa/pull/7941#issuecomment-2265706378) so please take a look and let me know if there are still outstanding issues.

IT WORKS!! (and the existing accordions still work as expected!) Thank you for taking a look at this, I was running into a wall.

joewhitsitt (approved):
Approving based on previous review and Marie's confirmation that the recent commits fixed the remaining issues.

PR #7893, opened 2024-07-08
v2 articles and people column width bug

GaryRidgway:
resolves https://github.com/uiowa/uiowa/issues/7665

1. On `main`, sync `pentacrestmuseums.uiowa.edu`
2. Run `ddev drush @pentacrestmuseums.local sqlc`
3. In the MySQL terminal, run the following commands, and copy their output to a scratch text file:
    1.  Get all people with the `paragraph__field_uip_colwidth` field set
    ```
    select * from paragraph__field_uip_colwidth where bundle="people";
    ```
    2.  Get all articles with the `paragraph__field_uip_colwidth` field set
    ```
    select * from paragraph__field_uip_colwidth where bundle="articles";
    ```
    3.  Get all people revisions with the `paragraph__field_uip_colwidth` field set
    ```
    select * from paragraph_revision__field_uip_colwidth where bundle="people";
    ```
    4.  Get all articles revisions with the `paragraph__field_uip_colwidth` field set
    ```
    select * from paragraph_revision__field_uip_colwidth where bundle="articles";
    ```
4. Checkout this branch, and sync the site again
5. Check that the update hook ran smoothly, and didn't error out.
6. Repeat step 3
7. Note that there are now more entries in the queried data, showing that we found entries without a default set, and we set them, so now they are discoverable by the SQL queries.
8. Sync another v2 site like `hawkeyemarchingband.uiowa.edu` and make sure that the update hook runs with no issues.
9. Code review?

<details><summary>Example output gathered during initial testing.</summary>

```

--BEFORE--

MySQL [pentacrestmuseums_uiowa_edu]> select * from paragraph__field_uip_colwidth where bundle="people";
+--------+---------+-----------+-------------+----------+-------+--------------------------+
| bundle | deleted | entity_id | revision_id | langcode | delta | field_uip_colwidth_value |
+--------+---------+-----------+-------------+----------+-------+--------------------------+
| people |       0 |       671 |      122726 | en       |     0 | col-md                   |
+--------+---------+-----------+-------------+----------+-------+--------------------------+
1 row in set (0.000 sec)

MySQL [pentacrestmuseums_uiowa_edu]> select * from paragraph_revision__field_uip_colwidth where bundle="people";
+--------+---------+-----------+-------------+----------+-------+--------------------------+
| bundle | deleted | entity_id | revision_id | langcode | delta | field_uip_colwidth_value |
+--------+---------+-----------+-------------+----------+-------+--------------------------+
[...]

PR #7943, opened 2024-07-29
V2 col width update fix

### LINKED ISSUE #7937: Failed Deployments
As far as I can tell, there was an issue with the update hook that we rolled out as a fix for v2 sites. I'm not sure it failed on all sites, but one of the queries results in a condition that doesn't allow the update hook to finish for some sites. I saw it on at least 3.

Every time we do a deployment now, it tries to run that update hook for any sites it previously failed on again.

I'm not sure if it might have succeeded on some sites and not others.
```[tasklist]
### Tasks
- [x] Sync one of the sites that failed locally, run the update hook, and use xdebug to see what's going wrong
- [x] Update the existing update hook to fix the issue
```

joewhitsitt:
Resolves #7937

# To Test

Reference https://github.com/uiowa/uiowa/pull/7893

`ddev blt ds --site=accreditation.uiowa.edu` on `main` and see the update hook fail.

Checkout `v2_col_width_update_fix`, run `ddev blt ds --site=accreditation.uiowa.edu` and see it succeed.

```
ddev drush @accreditation.local sql-drop -y && ddev drush @accreditation.local cr && ddev drush sql-sync @accreditation.prod @accreditation.local && ddev drush @accreditation.local uli
```

Add an entry that doesn't have the column width set for people and/or articles. Run `ddev drush @accreditation.local updb` and see it succeeds and that the node entry has been updated to Full.

### DIFF
--- a/docroot/modules/custom/sitenow_paragraphs/sitenow_paragraphs.install
+++ b/docroot/modules/custom/sitenow_paragraphs/sitenow_paragraphs.install
@@ -20,12 +20,21 @@ function sitenow_paragraphs_update_9001() {
       ->condition('cw.bundle', ['people', 'articles'], 'IN')
       ->execute()->fetchCol();
 
-  $people_and_articles_without_column_entries =
-    $database->select('paragraphs_item', 'pi')
-      ->fields('pi', ['id', 'revision_id', 'type'])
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Works as described. Nice fix!

PR #7948, opened 2024-07-31
Delete michaelson-lab.uiowa.edu multisite on uiowa07

joewhitsitt:
michaelson.lab.uiowa.edu still exists on uiowa


pyrello (approved):
I think this looks all good.

Issue #7689, opened 2024-05-15
Launch/Post-launch tasks for Emergency v3 site


briand44:
Launch planned for July 31st, 2024 in the am. There will be a production test Hawk Alert on Aug 1st at 10am.

```[tasklist]
### Tasks
- [x] Need to have hostmaster update DNS for e.uiowa.edu to point to Vulture
- [x] Need to deploy Signage code `update_signage_hawkalert` for both signage and healthcare signage to work with new endpoint - Update Signage/HC Signage to consume Hawk Alerts from new Emergency endpoint #7686
- [x] Need to deploy SiteNow v1 code `update_uiowa_alerts` to uiowa701-3 to work with new endpoint https://github.com/uiowa/uiowa/issues/7688
- [ ] https://github.com/uiowa/uiowa/issues/7772
- [x] Verify everything is working during the test Hawk Alert on Aug 1st
- [x] If test goes flawlessly, ask about removing the D7 site
- [ ] https://github.com/uiowa/uiowa/issues/7778
```

briand44:
blocked until 7/31/24

joewhitsitt:
With https://github.com/uiowa/uiowa/pull/7949 deployed this afternoon, I am no longer seeing CORS console errors as SiteNow and Signage try to ping emergency.uiowa.edu

Based on successful testing previously with each of the development pieces I believe this should resolve any remaining issues related to this work.

Reached out in #redesign-police to pitch the idea of another test soon to confirm.

briand44:
Another test hawk alert happened and the hawk alert did appear on SiteNow sites and Signage. We should make a backup of the D7 site but then can remove. I'll add this to the site removal issue. Closing this as we can address #7778 separately.

PR #7949, opened 2024-08-01
CORS tweaks

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7689

So, the previous _very cool_ approach to CORS was to add the necessary headers for specific paths, keeping the wildwest at bay, however, there was early issues that hinted at cached responses and those responses not always triggering and providing those headers.

Backing away from that solution for now, we are now using Drupal core functionality that is maintained at the services.yml level and is actually included but not turned on in the default site (default.services.yml). When it comes to Drupal multisite, the idea is to copy that file and only include the bits that you need for a specific site. In this case, I copied, gutted and modified the services file to enable the cors functionality.

Knowing that most requests to the emergency website could come from any domain and request the application/json content type header that is how I defined the configuration. Other example configurations go above and beyond and allow more things but each additional thing starts to expose us to potential XSS attacks if already vulnerable and performance hits.

content-type - Indicates the media type of the request body (e.g., application/json for JSON data)
allowedOrigins as * means that any domain like (testiowa.com, uiowa.edu, brand.uiowa.edu) can make a request.
allowedMethods as GET and OPTIONS keeps it restricted from doing things like POST which we don't need at this time.
other things are false by default so until we have a use-case, let's keep those off.

# How to test

- Use a cors tester like https://cors-test.codehappy.dev/ to set the URL to https://emergency.dev.drupal.uiowa.edu/api/active and the Origin to something like https://brand.uiowa.edu or https://gis.sites.uiowa.edu or https://signage.uiowa.edu with the method as GET.
- Observe that https://admissions.stage.drupal.uiowa.edu/ is showing a hawk alert that comes from emergency.dev.drupal.uiowa.edu and that there are no CORS related console errors.
- Authentication issue in DEV seems to be unrelated but was an early concern. Worse case, we can `drush uli` for Public Safety to make changes as we remove the changes.


joewhitsitt:
Obviously `allowedOrigins` being open with `*` is not ideal, but it doesn't support wildcard subdomain support (e.g. *.uiowa.edu) and we have other domains that need coverage.

joewhitsitt:
So I think this works, but it is scoped too much maybe that now authentication doesn't work?

Screenshot 2024-08-01 at 12 10 10 PM

Screenshot 2024-08-01 at 12 11 19 PM

I am able to login to another DEV site its.dev

GaryRidgway (approved):
Tested as described, saw the hawkalert, looks good to me

joewhitsitt:
Oye. Just saw this:

> Configure requests allowed from origins, matching against regex patterns.
>    allowedOriginsPatterns: []

But I think we are scope fairly well for now. We have more non-uiowa.edu domains than I want to have to keep track of here. Also, I saw somewhere else that client support for this is mixed.

PR #7975, opened 2024-08-12
[Registrar] Dates by Session

### LINKED ISSUE #7843: Dates by session
# Story

~As a Registrar site owner, we would like the ability to configure the academic calendar to group by session so that we can maintain similar functionality as our existing dates by session.~ - This should already be addressed.

As a Registrar site owner, we would like the ability to show calendar dates going back two years and forward 2 years since that is how far out the academic calendar committee plans the calendar dates.

# Additional Info

https://registrar.uiowa.edu/dates-by-session

They liked what we demoed them for the Academic Calendar when we had that grouped by session. Will likely need its own page still and have more sessions available than the academic calendar.

They mentioned 2 years back and 2 years forward but it looks like they may currently be showing more than that. Possibly current session +/- 10?

Filter by category - same as academic calendar categories
is web display - similar to academic calendar
is reviewed - similar to academic calendar
more used internally by departments so they don't have to scroll through each month

GaryRidgway:
# How to test
1. Pull this branch
2. Go to https://registrar.uiowa.ddev.site/calendars-and-deadlines/academic-calendar
3. Edit the layout of the page and set the `Session(s) to display` to `Current session, plus next ten sessions` and also check the `Include past sessions` checkbox.
4. Notice that we are now displaying 10 sessions before the current session, and 10 sessions after, and all the subsessions in between.
5. Note as well that you can emulate the old [dates by session](https://registrar.uiowa.edu/dates-by-session) page and that this shows all the same information.
6. Check that subsessions show by default now, and that the toggle no longer shows on the user facing form.
7. Check that the session dropdown now has all subsessions indented to show that they are under the main seasonal sessions.


GaryRidgway:
TODO: split the skeleton loader in to another branch.

GaryRidgway:
[Skeleton load moved to here.](https://github.com/uiowa/uiowa/pull/7981)

GaryRidgway:
@joewhitsitt I am looking in to the JS code right now for the session dropdown being filtered one whether the "show subsession" checkbox is checked. I am not seeing anything in the code that is a vestige of that once occurring. I also think that it probably is best if we don't change the dropdown every time we toggle that, as it will reset the user's choice, or we will have to come up with some logic to change their choice based on what it was before a filtered option that they selected was removed.

Perhaps we can do some other type of separation here? Maybe we can offset/tab in the subsessions in the filter? I think I've seen that before

GaryRidgway:
> Perhaps we can do some other type of separation here? Maybe we can offset/tab in the subsessions in the filter? I think I've seen that before

<details><summary>I made a quick mockup of what that might look like here</summary>

Screenshot 2024-08-14 at 11 05 15 AM

</details>

joewhitsitt:
>I am not seeing anything in the code that is a vestige of that once occurring.

Yeah I don't have a working version to confirm, but previously the js was very fond of fetching events and `fetchAndDisplayEvents‎` was triggered any number of ways all through out the JS. Inside included `populateSessionFilter` so I think it was constantly updating the dropdown based on the events returned from the fetch.

This might all be moot based on what Ben said during EoD of removing the subsession toggle?

https://github.com/uiowa/uiowa/blob/eae8d962a58af2cf3b55d57fc36c9463d16a7328/docroot/sites/registrar.uiowa.edu/modules/registrar_core/js/academic_calendar.js#L32

joewhitsitt:
pulling to test

joewhitsitt:
I think because this block was already placed, we have a not set issue with the new block config:

`Warning: Undefined array key "include_past_sessions" in Drupal\registrar_core\Plugin\Block\AcademicCalendarBlock->build() (line 237 of sites/registrar.uiowa.edu/modules/registrar_core/src/Plugin/Block/AcademicCalendarBlock.php)`

joewhitsitt:
I might need to huddle with you @GaryRidgway as I can't seem to get results to show for anything but All Sessions and checking past dates doesn't help.

Update: Nevermind see https://github.com/uiowa/uiowa/pull/7975#issuecomment-2294204044

joewhitsitt:
Actually it might seem related to the category dropdown. It doesn't require a value, but without one it doesn't list all. Add to that, selecting an option does not re-trigger the fetch. Reproduced in PROD.

GaryRidgway:
> Actually it might seem related to the category dropdown. It doesn't require a value, but without one it doesn't list all. Add to that, selecting an option does not re-trigger the fetch. Reproduced in PROD.

Argh, good to know. you said you weren't able to get some of it to work at all?
yeah we may need to huddle quick for me to see
Good to know about the array index though, That looks like I just wasn't safe enough with the code

joewhitsitt:
I am noticing some events like Early Registration are not returning in the new version, but I feel like they have evaluated how we are filtering events previously and will let us know if something is supposed to be returning.

Screenshot 2024-08-16 at 3 45 51 PM

Screenshot 2024-08-16 at 3 48 40 PM

GaryRidgway:
> I am noticing some events like Early Registration are not returning in the new version, but I feel like they have evaluated how we are filtering events previously and will let us know if something is supposed to be returning.
> 
> Screenshot 2024-08-16 at 3 45 51 PM Screenshot 2024-08-16 at 3 48 40 PM

I am confused that in this picture it is saying `August 9` in `Winter 2021`...

joewhitsitt:
I was once confused like you, but for example, as a student i want to see what courses are available for the winter session well before it starts so i can plan and register for them.

joewhitsitt (approved):
I think this is a clean addition and accurate enough to get in front of them for review/feedback. It seems to deliver similar results as what the site currently does.

I assume the issues with the form might already be addressed in the other form with the removal of Chosen.

I could see the need for show previous dates to be available when group by month is unchecked.

TIL of `&emsp;` and plan to use it more than being a one-trick pony with `&nbsp;`

PR #7981, opened 2024-08-14
[Registrar] Academic calendar skeleton load

GaryRidgway:
This PR makes a new skeleton load block that can be programmatically loaded to show users and screen readers alike that the page is busy loading things.

# How to test
Check the academic calendar to see if when it loads, there is a skeleton before it populates data that shows the user it is loading. For best results, set the block to grab 10 sessions after, and check the "Show previous events" checkbox so that it takes a while to load.


docroot/modules/custom/uiowa_core/src/Plugin/Block/SkeletonLoadBlock.php
```diff
@@ -0,0 +1,160 @@
...
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides an 'Skeleton Load' block.
+ *
+ * @Block(
+ *   id = "skeleton_load_block",
+ *   admin_label = @Translation("Skeleton Load"),
+ *   category = @Translation("Site custom")
```
joewhitsitt: We don't want this block to be placed in Layout Builder directly right? Shouldn't this be `Restricted`?

```suggestion
 *   category = @Translation("Restricted")
```

joewhitsitt:
pulling to test

joewhitsitt:
I think this is a clever solution. The aria-hidden additions are nice, but I wonder if the values should be adjusted to be more boring, "placeholder headline" type stuff since users could have their own stylesheet overrides where color: transparent is no longer respected.

GaryRidgway:
> I think this is a clever solution. The aria-hidden additions are nice, but I wonder if the values should be adjusted to be more boring, "placeholder headline" type stuff since users could have their own stylesheet overrides where color: transparent is no longer respected.

I think this is completely reasonable, however I guess since we are the users it might not matter too much. But I will adjust them to ensure some potential fallback support.

joewhitsitt (approved):
thanks for the tweaks. looks cool. works well.

PR #7983, opened 2024-08-14
[Facilities] Include both current and upcoming alerts

cory-skeers:
Request from Facilities

> Before we migrated facilities.uiowa.edu to SiteNow v3, we had a previous iteration of this custom Alerts app that allowed us to publish construction alerts ahead of time, especially for high impact access alerts like this road closure on Madison Street for new student move-in. In our training for the system, we advise our construction managers to give as much advance notice as possible for access alerts so those commuting to or around campus are aware of closures ahead of time and can plan accordingly.  

> Can we adjust this custom app to allow us to include alerts in the Alerts block as soon as they are marked “published?”

> We would prefer to have the items sort by most recently published with the newest items at the top, and keep that “sticky” option if we were to have a super high-impact project that we want to display first.

# How to test

`ddev blt ds --site=facilities.uiowa.edu`

Check for a clean sync
Visit https://facilities.uiowa.ddev.site/alerts and see that the "S. Madison Street south of Washington Street between the Main Library and Seamans Center - Library and EPB Steam Service" alert appears at the top of the list, and that further alerts follow in descending order by start date.

Code review and or sync ITS and check that its alert lists block still only contains current (not upcoming) alerts.


cory-skeers:
Adjusted the sort order to go by alert Start Date but desc, so the most recent come to the top. I think this fits their needs as well as avoids some of the (larger) complications with sorting by publish date.

joewhitsitt:
going to go ahead and review in case Facilities responds and this needs to go out pronto. pulling

joewhitsitt:
Wondering if you had tried syncing again @cory-skeers? I am getting a views config issue with the removed handler.

Screenshot 2024-08-15 at 10 27 35 AM
Screenshot 2024-08-15 at 10 27 42 AM

joewhitsitt:
I think it is caused by

```
        filter_groups:
          operator: AND
          groups:
            1: AND
```

Was that intentional or an auto-add?

cory-skeers:
> I think it is caused by
> 
> ```
>         filter_groups:
>           operator: AND
>           groups:
>             1: AND
> ```
> 
> Was that intentional or an auto-add?

Auto-add, and not sure I tried a fresh sync myself with this.

joewhitsitt:
something is still funky and saying the alert date filter is missing. poking around

joewhitsitt:
ugh. removing the broken handler doesn't change the export, but then trying to run a cim again it wants to add this:

```
diff --git a/tmp/drush_tmp_1723739866_66be2eda874be/views.view.alerts_list_block.yml b/tmp/drush_tmp_1723739866_66be2edac4b16/views.view.alerts_list_block.yml
index 0536a79f9..a18c04bc2 100644
--- a/tmp/drush_tmp_1723739866_66be2eda874be/views.view.alerts_list_block.yml
+++ b/tmp/drush_tmp_1723739866_66be2edac4b16/views.view.alerts_list_block.yml
@@ -547,6 +547,14 @@ display:
           hierarchy_depth: 0
           required_depth: 0
           none_label: '- Please select -'
+        field_alert_date_value:
+          value: {  }
+          expose:
+            operator_list: {  }
+            remember_roles: {  }
+          group_info:
+            default_group_multiple: {  }
+            group_items: {  }
       filter_groups:
         operator: AND
         groups:
```

joewhitsitt:
Worse case I think this could be switched to a full split as I think there is something going on with the partial.

cory-skeers:
> Worse case I think this could be switched to a full split as I think there is something going on with the partial.

Ugh, yeah, I tried going through from a fresh sync off main and end up in the same position, and had a similar experience trying to remove the broken handler and re-export after.

I'll push up an attempt with a full split.

cory-skeers:
Full split seems to work as expected.

joewhitsitt (approved):
checked the full split. working. not loving it, but I saw the email saying they want this change.

cory-skeers:
> checked the full split. working. not loving it, but I saw the email saying they want this change.

Thanks @joewhitsitt! 🤞 we can get a cleaner solution as a follow-up later.

PR #7995, opened 2024-08-19
Parallel deployment updates processing

### LINKED ISSUE #7800: test running parallel tasks to shorten length of time deployments take
In order to upgrade all applications to ACN, we need to find a way around our long deployment times. We have discussed adding more applications and distributing sites more evenly across those applications so we don't have such a large number of sites on a given application.

Acquia has also recommended we try to run tasks in parallel.

https://acquia.my.site.com/s/case/500Pb00000AdNySIAV/cloud-hooks-troubleshooting

**This comment from the ticket was on 4/24/24**

> You may be able to actually parallelize tasks via the use of the GNU Parallel utility. It's important to note that GNU Parallel is only installed on ACN, not Cloud Classic.
> 
> Here's a simple proof of concept showing how you might use it in something like a post code deploy script.

```
#!/bin/bash
 
# Define the function that carries out the operations against each site
drush_tasks_for_site() {
    local site="$1"
    echo "$site"
    # Do Drush tasks here for each site: updb, cim, etc
 
}
 
# Export the function so it's available to parallel
export -f drush_tasks_for_site
 
# Array of sites
sites=(
    'site1'
    'site2'
    'site3'
    'site4'
    'site5'
    'site6'
    'site7'
    'site8'
    'site9'
)
 
# Check if GNU parallel is installed. Another option could be checking for environment name, for envs we know are on ACN
if command -v parallel >/dev/null 2>&1; then
    echo "GNU parallel is installed, running tasks in parallel."
    # Run 3 tasks at a time using GNU parallel
    parallel -j 3 drush_tasks_for_site ::: "${sites[@]}"
else
    echo "GNU parallel is not installed, running tasks sequentially."
    # Run tasks sequentially
    for site in "${sites[@]}"; do
        drush_tasks_for_site "$site"
    done
fi
```

> A few things to note:
> Please be aware that the output from the tasks might be a little intermingled since the tasks will be occurring simultaneously. There might be ways of directing individual output to task-specific log files
> It may be tempting to run more than three items in parallel just to get the hooks to run quicker, but we'd recommend being conservative because we don't know what other bottlenecks would emerge from running too many at once e.g. command pod resources, or max DB connections
> Your team should ensure the tasks that will be run in parallel are discrete and self-contained. If some tasks end up having dependencies on the output of other tasks, that could cause issues
> This overall idea is untested, so we're interested in seeing how if/how you can make it work and would really appreciate any feedback you can provide in terms of timing
> If you do make use of this idea, please let us know if it speeds up the overall execution of your hooks.

---

**Email from Acquia 4/23/24**

> Arden and I met with the Product team and shared the deployment time concerns. They replied saying as this is a blocker for upgrading to Acquia Cloud Next that until we are able to support or solve for at this time, an upgrade to ACN will not be required by the Acquia Cloud Classic EOL. The page in our docs also essentially says the same under "Important": [Software EOL Schedule](https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.acquia.com%2Fservice-offerings%2Feol&data=05%7C02%7Cbrian-dickey%40uiowa.edu%7C743a595a645d4ee9ebc008dc63a8c4d6%7C1bc445959aba4fc3b8ec7b94a5586fdc%7C1%7C0%7C638494823713801221%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=68RoadKLVUCeMH6O1HJ4wdiEFzro29ANLOW4JG%2B4%2Bpo%3D&reserved=0). 
> 
> The team also shared that there are some code deployment time enhancements coming up in Q2 so there may be updates around this in the near future. 
> 
> If your team finds that you otherwise have concerns about deployment times, moving to ACN or engaging with our Professional Services team, please let us know.
```[tasklist]
### Proposed Solution
```

pyrello:
Resolves #7800 

# How to test
- To avoid running updates against all sites locally, make sure you have a `multisites` entry set up in `local.blt.yml`.
- Test first without parallelization:
```
```
- Set up parallelization:
```
ddev ssh
```
- Inside DDEV:
```
sudo apt install parallel && exit
```
Note: This will be removed with a ddev update/restart.
- Test with parallelization:
```
```
- Mangle some config: make an edit to any config file in the default config set so that the YAML is invalid. This will cause site deployment updates to fail.
```
```
- After all output from parallel deployments, you should see `[error]  Error deploying updates. Check the log output for more information. `


blt/src/Blt/Plugin/Commands/ReplaceCommands.php
```diff
@@ -42,47 +42,27 @@ public function replaceDrupalUpdateAll() {
...
-            $this->logger->error("Failed deploying updates to {$multisite}.");
-            $multisite_exception = TRUE;
-          }
-        }
+    // Check if the parallel command exists.
+    if (trim($parallel_installed->getMessage())) {
+      $this->say('Running multisite updates in parallel.');
+      // Run site updates in parallel.
+      $parallel_output = $this->taskExec('parallel -j 3 blt uiowa:site:update ::: ' . implode(' ', array_map('escapeshellarg', $multisites)))
+        ->printOutput(false)
```
joewhitsitt:
```suggestion
        ->printOutput(FALSE)
```

joewhitsitt (approved):
i've been testing along the way. let's see how it does in stage/prod!

PR #8000, opened 2024-08-21
[ITS] Remove editor perms for Page content type

### LINKED ISSUE #7987: ITS feedback re: editor role
From James to its-web:

Hey,

Adrienne noticed something during our first training session for the ITS website today and it will need to be addressed soon.

I was using the editor role to train people (not stacked with any other roles). As an editor, I also had the ability to edit the alerts pages using Layout Builder. This is not okay for an editor to do.

We would like to restrict the editor role from using Layout Builder; we think in doing that we should also remove the ability to create/edit 'pages' or 'basic pages' from the editor role. 

If we want people to be able to modify (basic) pages then we can give them the publisher role stacked.

Open to discussing this and open to your thoughts and ideas on how to best address this issue. Sorry that we didn't catch it earlier.
```[tasklist]
### Tasks
- [ ] Remove the ability to create and edit basic pages for the editor role
- [ ] Test to make sure there are not any undesired consequences from the permissions changes
```

cory-skeers:
Resolves #7987 

# How to test

```
```

Masquerade as a user with only the Editor role
See that you cannot create Pages or edit Pages (such as `/alerts`)

Masquerade as a user with only the Publisher role
See that you can create and manage publishing workflow of Pages, but cannot edit their LB layout (such as `/alerts`)

Masquerade as a user with BOTH the Editor and Publisher role
See that you can do it all, including create new Pages, manage their publishing workflow, and edit new/existing Pages including editing their LB layout

Any other perms testing/considerations you can think of


joewhitsitt:
pulling to test

joewhitsitt:
We will have to remember this if we ever address the [permissions stacking stuff in the default](https://github.com/uiowa/uiowa/issues/4294)

joewhitsitt (approved):
I don't think the perms i added can directly be used, but in terms of housekeeping i thought it would be good to add. It otherwise checks out.

cory-skeers:
> I don't think the perms i added can directly be used, but in terms of housekeeping i thought it would be good to add. It otherwise checks out.

Looks good--thanks for the added tidying!

PR #8009, opened 2024-08-22
ITS Updates Sprint 227

joewhitsitt:
# How to test

```
```

Might be moderation control warnings for service, support articles, but that is nothing new.

- As an editor/publisher/webmaster I don't see the node title field on the Alert content type form. Additional details is lower in the form.
- As an editor, I can update a published alert and move it directly to "submit" (needs review). I also see field description text that explains I need to save as submit in order for the helpdesk to publish my alert.
- As a publisher/webmaster I can update a published alert and go directly from published to published and don't see the field description about the submit state.
- As a visitor, I can see the calendar render at /alerts/calendar
- As an editor, I can add/edit a service/support article and save as a draft or directly publish.
  - Saving as a draft, I can see the moderation control functionality when I view the unpublished node and switch it to published.
- As a publisher/webmaster I can archive service/support article content.
- Sync another website to confirm that there are no import errors associated with the new workflow and that you still see "Needs Review" type states for the article content type.

### DIFF
--- a/config/default/workflows.workflow.draft.yml
+++ b/config/default/workflows.workflow.draft.yml
@@ -1,0 +1,56 @@
+uuid: d237e481-7313-4ec4-83a8-75b3d13ca8de
+langcode: en
+status: true
+dependencies:
+  module:
+    - content_moderation
+id: draft
+label: Draft
+type: content_moderation
+type_settings:
+  states:
+    archived:
+      label: Archived
+      weight: 1
+      published: false
+      default_revision: true
+    draft:
+      label: Draft
+      weight: -1
+      published: false
+      default_revision: false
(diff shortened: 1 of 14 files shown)


e-marie-w:
Right now I cannot add a service as an editor (or publisher for that matter). I can edit existing services just fine and use the draft or directly publish options as expected, though. Everything else on the testing list is working for me.

joewhitsitt:
@e-marie-w that is the expected behavior. only webmasters can add/delete services. ITS wants to prevent new services from springing up and accidental deletion

e-marie-w (approved):
I went through the testing instructions, everything is looking and working as expected, awesome job!

PR #8012, opened 2024-08-23
[Registrar] five year calendar

### LINKED ISSUE #4943: Five-year academic calendar
As a Registrar site owner, I would like the ability to display the five-year calendar.

# Additional Info

Existing D7 implementation https://registrar.uiowa.edu/five-year-calendar.

API should have some sort of "Five Year Date" flag in MAUI (also should be "reviewed = true" and "printDate = true")

Dates should be grouped by academic year/session rather than by month.

Ability to filter by academic year https://github.com/uiowa/uiowa/issues/4943#issuecomment-2195448532. current academic year +/- 5.

Ability to show sub-sessions.

# Proposed Solution
```[tasklist]
### Tasks
- [ ] Add a new route in `registrar_core.routing.yml` that points to a new method in the controller we created for the Academic Calendar
- [ ] The new route method can share whatever logic sets the API call to return results where `reviewed = true` and `printDate = true` but also needs to add `fiveYearDate = true`
- [ ] Add a new JS file/library to encapsulate a new Drupal behavior that can be applied for the 5-year calendar
- [ ] Build the JS logic to fetch results from the controller route and display the events grouped by year and then by session
- [ ] Figure out how to toggle showing sub-sessions
```

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/4943. 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /calendars-and-deadlines/five-year-calendar
```

- Add the "Five Year Calendar" block to the page.
- Confirm that it defaults to current academic year.
- Confirm that each year is listed in this order:
  -  Fall
  -  Winter
  -  Spring
  -  Summer
    -  4wk
    -  6wk I
    -  6wk II
    -  8wk 
    -  12wk
- Compare events against https://registrar.uiowa.edu/five-year-calendar and confirm same events are being shown.
- Confirm url hash is updated each time the Academic Year select changes. 
- Confirm reset button resets to default state.
- Look at Controller refactor and confirm that https://registrar.uiowa.ddev.site/calendars-and-deadlines/academic-calendar is still working as expected. 
- Code review 
  - Please feel free to change or suggest changes to make the code better!


GaryRidgway:
Pulled for review

GaryRidgway (approved):
This looks good to me! I've modified things a little bit just to ensure things are working correctly.

GaryRidgway:
Some thoughts:

- I think there was probably a way to make this block just a variant of the academic calendar and keep all the code in one place. However, I think there is also a case for splitting something so flexible in to multiple files anyways. I think we are on the precipice of that decision right now, but probably don't have to make it currently
- For example, we could make the academic calendar class its own thing, and then have JS extend that class and allow it to gain all the benefits of the base class without having to redefine it in another JS file.

joewhitsitt:
>Please feel free to change or suggest changes to make the code better!

I'd like to look at this a bit

joewhitsitt:
Is the idea that I can take a URL like https://registrar.uiowa.ddev.site/calendars-and-deadlines/five-year-calendar#2026-27 and link to it from somewhere else, that on load it changes the filter and thus the result?

bspeare:
@joewhitsitt that was the idea, but I see now that it is resetting to the default.   I can look into this.

joewhitsitt:
Nothing appears to be a no-go with me, but I see some areas where things can be done differently:

- Looks like we broke the mold with `getDateCategories()` but I don't love non-api calls like `getFallSession()` and `getTerm()` which are more like post-call helpers/processors in the MauiApi.php file.
- i am late to the review party on the academic calendar suite but i noticed we aren't doing the same kind of drupal.announce work we did recently (and for the same site) for gpa and course deadlines. i know that isn't part of our normal views work either, but might be worth sprinkling in later on especially as we do more autosubmit
- my brain doesn't want to step through JS right now, so i am deferring to Alan's review/approval there.
- The separate `fetchAndProcessFiveYearCalendarData` function could be eliminated and `fetchAndProcessCalendarData` could and anything that calls it could just be updated right? `fetchAndProcessCalendarData($start, $end, $categories, $steps = 0, $includePastSessions = FALSE, $isFiveYearCalendar = FALSE) {`
- Untested, but i think the empty validate and submit can be removed from the block plugin.

joewhitsitt:
ugh. my bad on the drupal.announce stuff. i was using devel_a11y but masquerading as a webmaster which doesn't have the necessary permissions to see the console activity...

bspeare:
Thanks @GaryRidgway & @joewhitsitt!

PR #8013, opened 2024-08-26
BLT UMD - Application check added to database operation

### LINKED ISSUE #7996: Make `umd` command more foolproof
# Story

As a web team, we would like to make our `umd` command more foolproof so that we don't have any more incidents where the wrong database is deleted.

# Additional Info

https://github.com/uiowa/uiowa/issues/7958#issuecomment-2287150540

>We could potentially solve the problem that we encountered before by adjusting the deletion command to use the information that is already known about the application the site lives on to ensure that the database is deleted from that application specifically.

>Add a warning for when a domain would produce a duplicate database name. A developer would still need to decide what to do in that situation. Lightweight, but more potential for developer error.

https://github.com/uiowa/uiowa/issues/7958#issuecomment-2287192372

>I'd lean towards a slice to make the umd more foolproof to ensure the site that was selected in the command is what is deleted. If that means the umd command needs to go beyond the human name and confirm the application and/or the database machine name by fetching more info about the site before deleting I think that would be good. We have worked hard to make deprovisions a JDI type task and I'd like to be confident we can continue treat them as such once all of the administrative checking is out of the way.

> An optional additional slice would be to make sure the provision command would stop if the $databases->create() failed.
> 
> Though we have some validation on the request form our provisioning process still requires some human checking for issues. Like you said their could be a use-case for a particular near duplicate domain name and that we should just continue to take more than a glance at what we are moving forward with.
```[tasklist]
### Tasks
- [x] Use application/machine name to target deletion of database
- [ ] Warning prior to creating duplicate database name?
```

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7996

The main change with this is that it gets the $app variable early in the process from the drush alias info and uses that for the files deletion. When the file deletion code was added, the database portion of the command was not updated to leverage this info.

The command is set to return early if the app does not exist.

So with the $app variable, instead of looping through all of the apps, it targets that specific app and returns back just the databases that reside there. Since there can't be duplicate database names on the same app the rest of the command works as expected.

# How to test

- See that https://github.com/uiowa/uiowa/pull/8013/commits/8262348cdd038768185f79550d231a5b471b0a16 was successful in deleting the gel_sites_uiowa_edu db from uiowa05
- Provision two test sites foo.lab.uiowa.edu (e.g. uiowa) and foo-lab.uiowa.edu (e.g. uiowa07). Run the provision command with the --no-commit flag to create the databases on the remotes. Now run the umd command, selecting foo-lab.uiowa.edu and making sure that the uiowa07 databases is removed and not the uiowa databases. Follow up with the deprovision of foo.lab and see that it is also successful. reset and discard git commits from the deprovisions.


joewhitsitt:
```
PHP Warning:  Undefined variable $name in /Users/jwhitsit/acquia/uiowa/blt/src/Blt/Plugin/Commands/MultisiteCommands.php on line 365

Warning: Undefined variable $name in /Users/jwhitsit/acquia/uiowa/blt/src/Blt/Plugin/Commands/MultisiteCommands.php on line 365
```

The commit, [c43352f](https://github.com/uiowa/uiowa/pull/8013/commits/c43352f48254bf1f92d87a6db3e48708175fb7c9), fixes this undefined var warning which resulted in the incomplete commit message for [8262348](https://github.com/uiowa/uiowa/pull/8013/commits/8262348cdd038768185f79550d231a5b471b0a16)

joewhitsitt:
I'd be happy to walk through the foo.lab/foo-lab test instructions to ensure we don't make a mess of things.

joewhitsitt:
Switching back to draft to look at the provision part of this issue.

joewhitsitt:
Putting back up as-is based on my probe into enhancing the provision command: https://github.com/uiowa/uiowa/issues/7996#issuecomment-2313055965

GaryRidgway:
Pulling to review

blt/src/Blt/Plugin/Commands/MultisiteCommands.php
```diff
@@ -288,34 +288,40 @@ public function delete(array $options = ['simulate' => FALSE]) {
...
 
-          // Find the application that hosts the database.
-          foreach ($databases->getAll($uuid) as $database) {
-            if ($database->name == $db) {
-              $databases->delete($uuid, $db);
-              $this->say("Deleted <comment>{$db}</comment> cloud database on <comment>{$name}</comment> application.");
+        foreach ($databases->getAll($uuid) as $database) {
+          if ($database->name === $db) {
+            $databases->delete($uuid, $db);
+            $this->say("Deleted <comment>{$db}</comment> cloud database on <comment>{$name}</comment> application.");
```
GaryRidgway:
```suggestion
            $this->say("Deleted <comment>{$db}</comment> cloud database on <comment>{$app}</comment> application.");
```

joewhitsitt: Nice find!

GaryRidgway (approved):
I think this looks like a nice improvement! APPROVED

joewhitsitt:
To expand on Alan's approval here. We tried to run through the test instructions but we ran up against a separate issue with the UMD command where it doesn't work if the site directory hasn't been pushed all the way up through PROD. I think we have run into this issue a couple times lately by provisioning a mistake (e.g. typo) and then trying to correct it in the same PR. So besides seeing the successful deprovision in this PR based on the new code and Alan reviewing the code changes, we weren't able to fully test this without sending out two test sites out to PROD.

PR #8017, opened 2024-08-26
[Registrar] Correspondence block

### LINKED ISSUE #4965: Display Registrar Correspondence in table
# Story

As a Registrar site owner, I would like to publicly display some Dispatch email correspondence so that our students, faculty and staff can catch up on communications that have been sent.

# Additional Info

D7 example https://registrar.uiowa.edu/correspondence

Current code in `uiowa703/docroot/sites/registrar.uiowa.edu/modules/custom/registrar_core/registrar_core.module`

Their current display takes a very long time to load. Hoping limiting to 2 years will help with performance.

They would like to display 2 years worth of correspondence that have the public archive option set, have a registrar tag and have an audience of student or faculty & staff.

Ability to filter by audience. (All, Students, Faculty & Staff)

Would like the ability to pass an audience filter in through a query parameter so they can have separate links to student correspondence and faculty & staff correspondence.

Would also like the ability to filter by unit/department tags.

They are ok with how things currently display but would like to replace the Intended Population with Audience. The email column should link to the archive view on Dispatch.

Screenshot 2024-08-20 at 3 29 37 PM

cory-skeers:
Resolves #4965 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /admin/config/sitenow/sitenow-dispatch
```
Update hook should run cleanly
Enter a Dispatch API key and save
```
ddev drush @registrar.local cron
```
Add a page and include the `Correspondence` block
See that there are results!
Check out the filter and see that it properly filters things.
Check the results themselves, that the URL works, etc.
Check that there appear to be 2 years' worth of results

Note: There will be a slight difference in the results for the `Student` and `Faculty/Staff` selections when compared to the existing site. The existing site has extra results not associated with Registrar that should not be appearing. See https://github.com/uiowa/uiowa/issues/4965#issuecomment-2316149199 for details.

Try adding a query parameter to the URL (eg `?audience=student`) and see that it properly sets the filter default on page load

Additional notes:
There will be a follow-up issue for accessibility review/updates and the additional filter mentioned on the issue.


e-marie-w:
When I ran `ddev drush @registrar.local cron` I got the following:
```
 [notice] Fetching correspondence archives.
 [notice] 331 archives upserted.
 [notice] Pruning 0 records from the correspondence archives.
 [error]  Security warning: Couldn't write .htaccess file. Create a .htaccess file in your private:// directory which contains the following lines: 
```
Followed by the code that does exist in the .htaccess file. I continued with the testing instructions and everything works fine, is this error message expected/safe to ignore then?

cory-skeers:
> When I ran `ddev drush @registrar.local cron` I got the following:
> 
> ```
>  [notice] Fetching correspondence archives.
>  [notice] 331 archives upserted.
>  [notice] Pruning 0 records from the correspondence archives.
>  [error]  Security warning: Couldn't write .htaccess file. Create a .htaccess file in your private:// directory which contains the following lines: 
> ```
> 
> Followed by the code that does exist in the .htaccess file. I continued with the testing instructions and everything works fine, is this error message expected/safe to ignore then?

That looks fine, and like it's a local setup issue. The notices are the custom cron job notices when it runs properly, and the security warning is separate. I remember having a similar issue the last time I worked on a cron-related task, but I don't remember the fix...I think it involved setting up a matching private-files directory and adjusting its permissions.

e-marie-w (approved):
Everything looks/works great! Update hook ran cleanly and I could successfully add a block. Results show up as expected (no non-Registrar correspondence) and only goes back two years as desired. Filters work, and clicking a result brings you to the Dispatch page for it. URL query parameters work - I couldn't figure out the faculty/staff one, but ?audience=student and ?audience=all worked great. The error message I got after running cron is apparently a local setup issue, so I feel comfortable approving now.

PR #8020, opened 2024-08-28
BLT UMC - Additional Checks

### LINKED ISSUE #7996: Make `umd` command more foolproof

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7996

References:

- The database creation trigger/response: https://cloudapi-docs.acquia.com/#/Applications/postApplicationDatabaseCreate
- How our database names are generated: https://github.com/uiowa/uiowa/blob/474e37a60c17283774fc7954f4679766332110d0/blt/src/Multisite.php#L28
- Optimized str_replace: https://www.phptutorial.net/php-tutorial/php-str_replace/

# How to test

- Try a command like with a similar site you know will have a `.`/ `-` type conflict:
```
./vendor/bin/blt uiowa:multisite:create gis-sites
.uiowa.edu --requester=jwhitsit --site-name="https://gis-sites.uiowa.edu" --no-commit --no-db
```
- Code previously existed to check for an existing site that matches
- You can try the try/catch (leave the --no-commit, but remove the --no-db flag) but it will create the database on the remotes which will have to be cleaned up through the UI as the deprovision script won't work until the provision is pushed all the way up to PROD.

# Additional

Screenshot 2024-08-28 at 10 32 53 AM

### DIFF
--- a/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
+++ b/blt/src/Blt/Plugin/Commands/MultisiteCommands.php
@@ -398,9 +398,42 @@ EOD
       return new CommandError('Cannot parse URI for validation.');
     }
 
-    // RMI does this but we run it with no interaction.
-    if (file_exists("{$root}/docroot/sites/{$host}")) {
+    $sites_dir = "{$root}/docroot/sites/";
+
+    // RMI does this, but we run it with no interaction.
+    if (file_exists("{$sites_dir}{$host}")) {
       return new CommandError("Site {$host} already exists.");
+    }
+
+    // Normalize the host for conflict checking.
+    $normalized_host = str_replace(['.', '-'], '_', $host);
+
+    // Scan the sites directory for potential conflicts with normalized names.
+    $directories = scandir($sites_dir);
+    $potential_conflicts = [];
+
+    foreach ($directories as $dir) {
+      // Skip '.' and '..' directories and non-directories.
+      if ($dir === '.' || $dir === '..' || !is_dir($sites_dir . $dir)) {
+        continue;
+      }
+
+      // Normalize the directory name same as Multisite::getDatabaseName().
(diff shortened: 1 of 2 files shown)


joewhitsitt:
trying the sites directory approach first instead of fetching all of the applications' databases and looping

GaryRidgway:
Pulling to test

GaryRidgway (approved):
Tested the site collision checks and they both worked, super cool!

Additionally I checked that the try/catch worked, and saw the correct output.

I'd say this is ready to go!

PR #8029, opened 2024-08-30
Deyployment site filtering

### LINKED ISSUE #8015: Filter list of sites during deployment to sites on that application
# Story

As a web team, we would like to filter the list of sites during a deployment to only includes sites on the application running so that we can increase performance.

# Additional Info

From Sean:

>Our current process doesn't pre-filter the list of sites to just include sites from the application it is being run on. Instead it checks if a config file exists for each site and then only continues execution for sites where it finds one. We have 1259 v2/v3 sites as of this writing, but the largest application has only 223 sites. That means we are doing at least >1000 checks for a non-existent file. Even with parallel processing, that means that each of those sites that doesn't exist on that application gets to temporarily occupy one of the 3 processing slots. We could create a mechanism to filter the list before the site updates are initiated and it would likely save us some time vs checking sites that aren't there as part of the update process. This would also help us clean up the log output and make it easier to read the logs because there wouldn't be a lot of noise related to 1000+ sites being "skipped".

Grooming conversation:
- https://github.com/uiowa/uiowa/issues/8004#issuecomment-2310826574
- https://github.com/uiowa/uiowa/issues/8004#issuecomment-2310834335
- https://github.com/uiowa/uiowa/issues/8004#issuecomment-2310850017

GaryRidgway:
Resolves #8015 

# How to test

## Deployment
```
```
- By default the above will test deployment against any sites you have in your `local.blt.yml` file under the `multisites` key so that still works as expected for local testing. If you have no `multisites` key, then it will test against every site. For any site that you don't have a local DB setup, it should print out a message that the site is being skipped.
- To test that it appropriately filters by application, in line 28 of `Uiowa\Blt\Plugin\Commands\ReplaceCommands`, change `local` to the ID of a specific application, such as `uiowa` or `uiowa05`. Then run the command above again. You should see the list of sites that it attempts to run updates for is limited to the sites listed under that application in the manifest.

## Create a site
```
blt umc foo.uiowa.edu --no-db --no-commit
```

## Delete a site
```
blt umd --simulate
```


GaryRidgway:
After this is done it would be cool if some of the code here was abstracted in to a trait so that we could make an easier transition off of blt when it is necessary

GaryRidgway:
Things todo:

1. Remove limit that was made for testing on the update command
2. pass yaml by reference for the utilsTrait
3. Check if the `replaceDrupalUpdateAll()` works with the new manifest.
4. Ensure that local doesn't break with `replaceDrupalUpdateAll()` because `local` isn't an app.
5. Add site to manifest on create command.
6. Remove site from manifest on delete command
7. Re-run the manifest creation before final merge to ensure that all sites that have been provisioned or deprovisioned over the duration of this PR being open are accounted for.

joewhitsitt:
Is cleaning up blt/src/Multisite.php in scope? Duplicates between this and MultisiteTrait

joewhitsitt:
i'll be ready to pair test later today. so far things seem good. curious about https://github.com/uiowa/uiowa/pull/8029#issuecomment-2341519216

pyrello:
@joewhitsitt Go ahead and approve at your leisure. I have marked this as draft because we need to wait until #8054 is merged so we can re-generate the manifest with those updates.

joewhitsitt (approved):

PR #8050, opened 2024-09-06
[Events] Exposed filters for local event list block 

### LINKED ISSUE #8041: Probe exposed filters for local event list block
# Story

As a web team, we would like to understand if exposing filters for the local event list block will be straight forward or if it is going to get complex so that we can determine the best way to solve for AFR event needs.

# Additional Info

Trying to decide between making enhancements to the page or enhancements to the list block. List block may add more value long term but could also be the more complex slice of work here.

Filters AFR would like:

- Search
- Category filter
- Date Range?

Scheduled a meeting with AFR to better understand how they intend to use and if the date range filter is necessary or nice to have here.

Team didn't think this would require much investment JDI/small type of task.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8041. 

# How to test

- Code review
  - [What's the best way to set the block date to today](https://github.com/uiowa/uiowa/blob/f2bec976846c65aae7904516e6d5067db1826a86/docroot/modules/custom/sitenow_events/sitenow_events.module#L128)? 

```
ddev composer install && ddev blt frontend && ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli  /page-list-block-test
```
1. Edit the block
2. Scroll to the filters section and check each filter to expose them to users
3. Save block and test the filters
4. Confirm that the block date is set to today
5. Confirm that exposed filters work as expected
6. Test `/events` page and confirm start and end date fields are working

## Test other views

```
ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli  /events
```
1. Test OnIowa date filter
2. Submit and compare against prod

```
ddev blt ds --site=iisc.uiowa.edu && ddev drush @oniowa.local uli  /projects
```

```
ddev blt ds --site=classrooms.uiowa.edu && ddev drush @oniowa.local uli  /spaces/find-university-classroom
```
1. Also test `/spaces/programmed-classrooms`

```
ddev blt ds --site=commencement.uiowa.edu && ddev drush @oniowa.local uli  /ceremonies
```
```
ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli  /iowa-grants-guide
```
```
ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default && ddev drush @default.local config-split:activate event && ddev drush @default.local uli admin/config/sitenow/sitenow-events
```


joewhitsitt:
@bspeare had you tried `hook_views_pre_view` instead of `sitenow_events_form_views_exposed_form_alter` to set the default date? just thinking if that is possible it won't need a rebuild.

I see sorts and classes are set in `classrooms_core_views_pre_view` and I think something similar could be done with `getExposedInput` and `setExposedInput`

bspeare:
Thanks @joewhitsitt, I'll try that!

e-marie-w:
Pulling down to test.

e-marie-w:
All the existing views retain the same functionality as before and so work great. But I have a question about the https://iowasummerwritingfestival.uiowa.ddev.site/events-list-block-test. The block date is set to today, as desired, but it cannot be changed - is that also expected/desired? Like I can select a different date but when I hit apply it changes it back to today's date.

bspeare:
@e-marie-w that is not the intent so thanks for testing that.  I will see if I can fix it!

bspeare:
@e-marie-w I added a fix if you can test it out again.

e-marie-w (approved):
New exposed filters work as expected, and existing event listblock views still work too. Great job! Down the line we may want to add a reset button to resemble the other views, but it's not necessary now.

bspeare:
Thanks @e-marie-w! I left the reset button out because if you add it to the view it always appears.  Even if the form hasn't been submitted.  Perhaps a future issue to investigate.

PR #8051, opened 2024-09-06
inrc.law - export expired non-profit organizations

### LINKED ISSUE #7964: Report of expired NPO's
# Story

As a LAWINRC website owner, we would like to be able to view/export a report or nonprofit organizations that are expired so that we can communicate with them appropriate information needed.

# Additional Information

There is an expiration date field on the nonprofit organization content type. They would like a report that displays all of the organizations where the expiration date is in the past with the ability to export to a csv file.

The report should have at minimum the following fields:

- Organization Name
- Contact Name
- Email
- Expiration Date
- Link to node would likely be helpful

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/7964
# To Test

```
git checkout inrc_npos_export && git pull && ddev composer install && ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli /admin/content
```
- Try the task link in the header to download expired npos.
- Get a complete csv file that has today's date in the file name with the columns suggested by #7964.
- Note that all dates are in the past.
- Try to visit the path as an anonymous user. You should get an access denied. https://lawinrc.uiowa.ddev.site/admin/content/expired-nonprofits

# Glamour shots
Screenshot 2024-09-06 at 11 51 53 AM
Screenshot 2024-09-06 at 11 55 43 AM

# How to test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -190,6 +190,7 @@
         "drupal/tvi": "^2.0@alpha",
         "drupal/views_bulk_operations": "^4.1",
         "drupal/views_custom_cache_tag": "^1.3",
+        "drupal/views_data_export": "^1.4",
         "drupal/views_tree": "^2.0@alpha",
         "drupal/webform": "^6.2",
         "drupal/weight": "^3.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "54446df7abc201d0e37c8d2130b6a071",
+    "content-hash": "d0b463a4da428889014ee680180aa613",
     "packages": [
         {
             "name": "acquia/blt",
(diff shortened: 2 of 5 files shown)


cory-skeers:
It looks like anyone with access to `/admin/content` has access to download, including editors/publishers and not just webmasters. I think that's okay, and that the main concern was it not being publicly available which checks out.

If it's alright that it's available to editors/publishers, then everything else looks good and works great.

joewhitsitt:
I think the main concern is to keep it from being triggered by anonymous visitors. They only have four non-admin users on their current site (2 webmasters, 2 editors).

I think all the information itself is available on the public facing nodes.

cory-skeers (approved):
> I think the main concern is to keep it from being triggered by anonymous visitors. They only have four non-admin users on their current site (2 webmasters, 2 editors).

Great, then I think this is ready to go as is.

joewhitsitt:
Thanks @cory-skeers

PR #8056, opened 2024-09-09
[Registrar] Correspondence archives follow-up

### LINKED ISSUE #8039: Correspondence additional filter and accessibility review/enhancements
# Stories

As a Registrar site owner, we would like to be able to filter correspondence by additional tags so that users can more easily find the correspondence they are searching for.

As a web team, we would like to review the correspondence feature for accessibility and add any enhancements so we can make sure we meet WCAG 2.1 aa standards.

# Additional Info

Registrar is planning on providing the tags that they want to include as part of the additional filter and any mappings needed.

cory-skeers:
Resolves #8039 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu
```
Visit https://registrar.uiowa.ddev.site/correspondence
Test the new Topic filter
Test visiting the page with one or two query parameters in the url (eg `?topic=commencement&audience=student`)
Check the accessibility of the form and its results

Added by @gridgway:
- ensure that when you get to the page, [the issue that appears here](https://github.com/uiowa/uiowa/pull/8056#issuecomment-2363927684) no longer persists.


cory-skeers:
Ugh...Might need to move the reset button to a JS addition. It at least can follow our pattern elsewhere then.

Building it as a form api submit button and using `#states` seems to conflict with the way the query parameters work, and one or the other (parameter or `#states` settings) end up getting cached on the form and 'stuck'. I think one way around that would be to move the query parameter setting to something more dynamic, but I think at that point, should go the route of matching the reset button to other applications and leave the query parameter handling as is.

joewhitsitt:
Hey Cory, I am seeing you struggle with the reset so I am wondering if it is needed. Nice-to-have for sure, but maybe not required to ship if they didn't request it.

cory-skeers:
> Hey Cory, I am seeing you struggle with the reset so I am wondering if it is needed. Nice-to-have for sure, but maybe not required to ship if they didn't request it.

Yeah, I think it's not needed, and was something I'd wanted to include alongside the submit button, but I think at this point we should move forward without it.

GaryRidgway:
pulling to review

GaryRidgway:
~~Unfortunately I don't think that the query parameters are working for me. I tried multiple combinations and the one suggested in the testing instructions and none seemed to work for me. I am going to look through the code to see if there is something needed to make them work.~~

EDIT: Never mind, I think I just needed a cache clear quick. Not sure why that was necessary when I was authed and I JUST synced the site though?

GaryRidgway:
There is an issue that I am finding that if a filter is set in the url params, and then you swap the topic filter manually to `- Any -`, it doesn't remove the topic from the results message because it still sees the topic in the params. Poking at this now.

GaryRidgway:
<details><summary>Just wanted to make sure I was showing what was going on here that I was trying to fix.</summary>

RegistrarCorrespondence

</details>

GaryRidgway:
What I am finding is that `All` and `- Any -` are treated very differently here. Ultimately, `- Any -` doesn't count as a value, but `All` does, meaning that it doesn't pick it up. Thus, the topic is always going to be set to the url parameter if one is set.

joewhitsitt:
Do we need to set the URL params each time the filter is triggered to get these to align?

GaryRidgway:
> Do we need to set the URL params each time the filter is triggered to get these to align?

I have been thinking about that, and I think we would. However, thats another portion of work we didn't plan for. Should we replan or continue on with adding them?

GaryRidgway (approved):
Alright! Now that I've added a few changes to fix the filters, This is good to go!

PR #8059, opened 2024-09-10
inrc - export all npos

joewhitsitt:
Quick update to include all npos, not just expired. renaming things. Includes categories (piped-separated) for additional filtering capabilities in Excel.

# How to test

```
ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli admin/content
```

The link there (available to all who can visit the route), downloads an updated file that contains all npos regardless of their published state or expiration date. It also includes categories, pipe-separated to deal with comma separated term names.


e-marie-w (approved):
This works great! Expired and unpublished npos are in the export as desired. The pipe-separated categories are a nice addition. I masqueraded as different roles to confirm they all (besides viewer) could download the file as expected.

PR #8061, opened 2024-09-11
Updated to Drupal core 10.2.8.

pyrello:
# How to test

It looks like the only real change was to the version of twig. So, I'd assume this is pretty innocuous change. Whatever level of scrutiny helps you feel confident that it's good to go.


GaryRidgway (approved):
pulled and composer installed and it ran with no errors. No testing beyond that.

pyrello:
Converting to draft to avoid this accidentally being merged before it is fully vetted.

PR #8062, opened 2024-09-11
[Registrar] File archive block

bspeare:
# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /media/add/file
```
## Test session
1. Add a file and choose "End of Session Class Counts"
  - General Catalog, Past Miscellaneous, Tuition and fees should show the `date range` field
  - Census Report, Commencement Program, End of Session Class Counts, Iowa College and University Enrollment Report, Official Census Date Class Counts, Profile of Students Enrolled should show the `session` field. 
2. Add "File archive" block to `node/426/layout` and select "End of Session Class Counts" in the filter config. 
3. Save page and confirm that you can see and download the file listed in a table with the title "Session - File Archive Type" pattern. 
4. Test sorting - do they display newest date first?
5. Test editing the files you created and confirm the fields show/hide as expected. 

## Test date range
1. Add a file and choose "General Catalog"
  - General Catalog, Past Miscellaneous, Tuition and fees should show the `date range` field
  - Census Report, Commencement Program, End of Session Class Counts, Iowa College and University Enrollment Report, Official Census Date Class Counts, Profile of Students Enrolled should show the `session` field
2. Add "File archive" block to `/node/431/layout` and select "General Catalog" in the filter config
3. Save page and confirm that you can see and download the file listed in a table with the title "Date Range - File Archive Type" pattern
4. Test sorting - do they display newest date first?
5. Test editing the files you created and confirm the fields show/hide as expected. 

 ### Code review and questions
- Is setting up `SessionOptionsTrait` here the best way to create the allowed_values function for the session field?
  - How many previous sessions should we pull in?
   - As this updates each year older sessions will not be visible in the allowed values, will there be data loss?
- Do values need to be unset on the form if they are hidden?


bspeare:
This is ready for review, but there are some outstanding questions that I have in the testing instructions that could lead to some further development.

joewhitsitt:
Pulling to familiarize and see if I can help answer questions.

joewhitsitt:
>best way to create the allowed_values function for the session field

It looks like the current site has file archives going back to 2000? https://registrar.uiowa.edu/tuition-fees

Previously I was focused on the exam schedule and suggested getSessionsBounded there, but I am starting to realize the value of the taxonomy vocabulary for sessions now. This will allow them full range of sessions for maintaining an archive as well as prevent future data loss if the entity is re-saved. This could then be used to supply the exam schedule block the legacyCode values.

>Do values need to be unset on the form if they are hidden?

It looks like the current site has validation to prevent this. So, either that needs to happen again or we need to figure out how to unset if hidden.

joewhitsitt:
https://iowaweb.slack.com/archives/C0164FNN1PV/p1726238422450229

Josh said he will have to ask higher up and can have an answer early next week. It looks like Ben upped it to 60 previous sessions to get back to fall 2009.

To move this forward for now, I'd be fine leaving as-is, or giving it a comfortable 90 or something higher. I think we need to cache the allowed_values result though because I think it fetches again during submit.

joewhitsitt:
@bspeare, was making this exposed-capable intended? It seems to work, but just checking if it is needed.

Screenshot 2024-09-13 at 1 50 15 PM

bspeare:
@joewhitsitt it just needs to be exposed in the block config.  We don't want it exposed to users.

joewhitsitt (approved):
>it just needs to be exposed in the block config. We don't want it exposed to users.

Easier said than done. It needs to be exposed to to be in the block config. Seems like the new norm now unless I am missing some cool trick. Instead of one-off code to hide it, I think we can just say what it does and that they probably don't want to use it.

Leaving the past sessions at 60 for now with some caching. Waiting to hear back from Josh about whether we need to pivot to a more static list.

Unsets added to fix issues with naming convention.

bspeare:
Thanks @joewhitsitt!

PR #8069, opened 2024-09-13
Write parallel deploy to log and grep for errors

### LINKED ISSUE #8016: Write parallel command output to log file and search it for errors instead of variable
# Story

As a web team, we would like to write parallel command output to log file rather than variable and search it for errors so that we can have these logs around longer and possibly speed up the deployment.

# Additional Info

> 1. At the point where we invoke the parallel processing command, we could pipe that output to a log file just to have it since the AC task log doesn't stick around long enough to be useful a lot of the time. This could be done to provide a more robust feedback system regardless of whether we use those logs for further processing.
> 2. We could experiment with swapping out the method we are using for error-checking in parallel processing to use log file output instead of assigning the output to a variable. This might be both more performant and less susceptible to failure based on trying to assign to much data to a variable.

Grooming conversation - https://github.com/uiowa/uiowa/issues/8004#issuecomment-2310826574

joewhitsitt:
Resolves #8016

Lifted and adjusted https://github.com/uiowa/uiowa/issues/7800#issuecomment-2299134874 to use `tee` to output the result both to the terminal and to a longer term log file.

# To Test

- Set up parallelization locally:
```
ddev ssh
```
- Inside DDEV:
```
sudo apt install parallel && exit
```
Note: This will be removed with a ddev update/restart.
- Test with parallelization:
```
```

Observe that the terminal outputs the site specific deployment info, but also mentions writing to a log file with a datetime suffix.

- Mangle some config: make an edit to any config file in the default config set so that the YAML is invalid. This will cause site deployment updates to fail.
```
```
- After all output from parallel deployments, you should see `[error]  Error deploying updates. Check the log output for more information. `

You can review the log file by ssh'n back into the ddev container, `cd /tmp` and `cat` the latest file prepended with `parallel_deploy_log_`

We have log rotation setup on each prod app that we can make sure is also added to all stage environments (dev/stage have a shared /shared/logs directory). It calls the logrotate.conf file which has been updated to handle `drush-cron.log` separately as we are potentially dealing with multiple different log files each week for `parallel_deploy_log_` still keeping the latest three. so if more than three deployments happen in a week only the latest ones will be kept for a bit longer. We will need to push this out and watch it for a while to determine if it is actually working as expected.

# Follow-up

- Ensure log rotate jobs are setup in for all ACN apps' stage environments.
- Watch for log writing and rotation as time goes by.

### DIFF
--- a/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
+++ b/blt/src/Blt/Plugin/Commands/ReplaceCommands.php
@@ -32,6 +32,7 @@ class ReplaceCommands extends BltTasks {
     // get the list of sites from the manifest.
     if ($app === 'local') {
       $multisites = $this->getConfigValue('multisites');
+      $log_dir = '/tmp/';
     }
     else {
       // Load the manifest.
@@ -41,6 +42,7 @@ class ReplaceCommands extends BltTasks {
         $this->logger->warning('No multisites found in manifest for application: ' . $app);
       }
       $multisites = $manifest[$app] ?: [];
+      $log_dir = '/shared/logs/';
     }
 
     // Unshift sites to the beginning to run first.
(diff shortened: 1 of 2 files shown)


GaryRidgway (approved):
Looks good to me! Worked with Joe to ensure this worked well. SHIP IT

PR #8075, opened 2024-09-17
ITS - change replyto from site email to node author email

joewhitsitt:
Screenshot 2024-09-17 at 12 56 10 PM

instead of

image

# How to test

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli admin/node/add/alert
```
Create an alert and change the moderation state to submit.

```
ddev mailpit
```

See the email now has a replyto pointing to the user's email.


bspeare (approved):
Thanks @joewhitsitt!

PR #8076, opened 2024-09-17
[Registrar] Final exam block plugin

### LINKED ISSUE #4967: Final Exam Schedule Block Plugin
# Story

As a Registrar site owner, I would like to be able to display final exam schedules so students can stay up to date about times and locations of exams.

# Additional Info

They seem happy with current functionality explained below https://github.com/uiowa/uiowa/issues/4967#issuecomment-2313262121 - https://registrar.uiowa.edu/final-exam-schedules

Need the ability to search the final exam schedule.

Don't need the ability to download/export to CSV.

Don't need the preview functionality.

Supporting content hasn't needed to change but would be nice if they had the ability to control that content if the need arises.

Could possibly simplify this.

# Proposed solution

Implement first slice mentioned in https://github.com/uiowa/uiowa/issues/4967#issuecomment-2338602035

```[tasklist]
### Tasks
```

```[tasklist]
### Tasks
```

e-marie-w:
Closes #4967 

# How to test
- `ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli`
- add/edit a page and add the Final Exam Schedule block via Layout Builder
- try to save as-is and see that you cannot save unless a session and last updated date is set. 
- Save again with a session and date chosen. 
- Block should now display the selected session, a legacyCode, and the date last updated. You can check https://api.maui.uiowa.edu/maui/api/pub/registrar/sessions/bounded to see that the output legacyCode matches the selected session (aka shortDescription).

### DIFF
--- a/docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Plugin/Block/FinalExamScheduleBlock.php
+++ b/docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Plugin/Block/FinalExamScheduleBlock.php
@@ -27,7 +27,7 @@ class FinalExamScheduleBlock extends BlockBase implements ContainerFactoryPlugin
   protected $maui;
 
   /**
-   * Constructs a new AcademicCalendarBlock instance.
+   * Constructs a new FinalExamScheduleBlock instance.
    *
    * @param array $configuration
    *   The plugin configuration.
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
I think this is a solid foundation. I wonder if a default date of today could be set for new blocks.

PR #8080, opened 2024-09-18
[ Bug ] Fullcalendar: only rebuild if AJAX is from views.

### LINKED ISSUE #8074: Full calendar resets to default view around 10-20 seconds after navigating to a different month
# Problem
Our full calendar resets to the default view around 10-20 seconds after navigating to a different month.

# Steps to reproduce
Try navigating to a different month on one of the sites below and notice how it will reset after around 10 seconds. 

- https://grad.uiowa.edu/academics/thesis-defenses
- https://its.uiowa.edu/alerts/calendar

# Proposed solution
- ?

GaryRidgway:
[Copying my comment from D.O:](https://www.drupal.org/project/fullcalendar_view/issues/3323399#comment-15778682)
> "I made a change in the fork here to try and ensure that the only ajax that changes the calendar view comes from views ajax. I don't think it covers the entirety of the need here considering other views on the page sending ajax requests will still reload the calendar, but it's a start!"

# How to test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -263,6 +263,9 @@
             "drupal/fragments": {
                 "remove_views_data_type": "https://www.drupal.org/files/issues/2022-08-24/3305851-remove_views_data_type-6.patch"
             },
+            "drupal/fullcalendar_view": {
+                "only_rebuild_on_view_change": "patches/fullcalendar_only_rebuild_on_view_change.patch"
+            },
             "drupal/google_analytics": {
                 "[3373921] Cannot install from existing config": "patches/3373921.patch"
             },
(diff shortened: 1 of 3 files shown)


joewhitsitt (approved):
>I forgot to REQUEST review, but yes, it can be approved - @GaryRidgway 

i pulled down and it looks good to me. hawk alerts fired and there was no update. At EoD yesterday I think I got confused by the >= 0 and < 0 between the two indexOf patterns [when thinking they should be separate if else/ORs].

tested adding a filter, making the calendar editable (drag update), turning on modal viewing. Those all are still respected.

PR #8081, opened 2024-09-18
[Grad] Removed permissions to create fellowship content

bspeare:
This is a support request from Makur. 

# How to test

```
ddev blt ds --site=grad.uiowa.edu  && ddev drush @grad.local /admin/people
```
- Code review or Masquerade as editor, publisher, webmaster and confirm you can no longer `/node/add/scholarship`


joewhitsitt:
>Should this be done in https://github.com/uiowa/uiowa/tree/main/config/features/scholarship instead?

I don't think so because they aren't the only ones to use the feature. It was originally built for Admissions.

joewhitsitt:
pulling to check

joewhitsitt (approved):
checks out

bspeare:
Thanks @joewhitsitt!

PR #8088, opened 2024-09-20
[Wellbeing] Filter adjustments

### LINKED ISSUE #8053: Update filters to display "- Any -" rather than "- Select -" and remove chosen
# Story

As a site owner for the Well Being website, we would like to update the placeholder texts for the filters so that there is less confusion from site visitors.

# Additional Info

https://wellbeing.uiowa.edu/well-being-resources

>A few people struggled with the [resource directory](https://wellbeing.uiowa.edu/well-being-resources). It seemed like they felt it was necessary to use all three filter categories. We did add language to the top but wondered about changing the text on the filter tool.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8053. 

# How to test

```
ddev blt frontend && ddev blt ds --site=wellbeing.uiowa.edu && ddev drush @wellbeing.local uli /well-being-resources
```
1. Confirm chosen is removed from filter and it is switched to using BEF collapsible details elements 
2. Confirm filters are working


joewhitsitt:
pulling to test

joewhitsitt:
Just curious if the auto-submit is needed for this view. Keyboard nav wise it is kind of frustrating to get booted out of the dropdown after each checkbox selection.

bspeare:
@joewhitsitt it's probably not needed and I can turn it off.  I just know we have it turned on for all other bef exposed forms.

joewhitsitt:
Is this the only one with multi selects though because I feel like it is pretty noticeable here. Then there is the string search delay issue that has been raised before: https://github.com/uiowa/uiowa/issues/6961

I don't think it is as big of an issue with single drop downs

joewhitsitt (approved):

bspeare:
Thanks @joewhitsitt.  Classrooms has a lot of them: https://classrooms.uiowa.edu/spaces/find-university-classroom.

joewhitsitt:
i wouldn't mind if we updated that too 😁, but this is the work currently in front of us and I think something we should think about for future instances.

PR #8094, opened 2024-09-23
[Facilities] Update alert field labels 

e-marie-w:
its-web request from Lauren C. to update labels so that alerts can also be used for high-impact events that aren't closures.
> Change “Dates of closure” to “Dates of closure/impact”
> Change “Reason and details for closure” to “Reason and details for closure/impact”

# How to test
Code review is probably fine, if you want you can sync down facilities.uiowa.edu and check the labels are displaying correctly on an alerts page.


joewhitsitt:
pulling to check quick.

joewhitsitt (approved):
i was slightly concerned about the special character in the label, but it seems to work.

Issue #8060, opened 2024-09-11
Setup uiowa09 Cloud Next application


briand44:
# Story

As a web team, we would like to setup a new application so that we can provision sites on it.

# Additional Info

Docs on setting up a new application - https://docs.prod.drupal.uiowa.edu/acquia-cloud

https://github.com/uiowa/uiowa/issues/7929

Teams have been assigned permissions to this application.

```[tasklist]
### Tasks
- [x] Setup uiowa09 similar to other applications
- [x] work with ISPO on log forwarding for uiowa08 (likely have EIPs to allowlist)
- [x] Setup uiowa09.stage and follow the docs
```

briand44:
We probably want to wait until we are notified by Acquia that this application is ready.

briand44:
Grooming conversation...  Production EIPs will need allowlisted by ISPO for log forwarding (uiowa08 and uiowa09). Cert stuff will also need added in Acquia Cloud for log forwarding (not sure if uiowa08 is done or not).

bspeare:
`uiowa09.stage` is ready to setup.  I have added a new checkbox to the tasklist and moving this out of blocked/waiting and back to top of backlog.

joewhitsitt:
I see https://github.com/uiowa/uiowa/pull/8154 was merged. Looks like there are some other items (auth, cloud UI, etc.) to do.

joewhitsitt:
https://git.uiowa.edu/its-web/data.its.uiowa.edu/-/merge_requests/130

joewhitsitt:
It looked like some config/vars/stuff was retained or set up by someone else. Added the SSL cert, installed the default site, set up secrets file, updated auth. I think this is done and ready to go.

PR #7952, opened 2024-08-02
setup uiowa08

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7929
Following step 2 of the Code section: https://docs.prod.drupal.uiowa.edu/acquia-cloud

# How to test

- Compare to previous app setup PR: https://github.com/uiowa/uiowa/pull/5898
- Note that https://github.com/uiowa/uiowa/pull/7952/files#diff-0ba8f676e05070f29dc5fb883fb82eba43544fd76f7c3ca1efb68d3514880b50R22 is the same SVN as what is listed in https://github.com/uiowa/uiowa/blob/main/.travis.yml#L21 so no addition to that file is necessary.
- Compare drush file to current main version of ACN uiowa07 https://github.com/uiowa/uiowa/blob/main/drush/sites/uiowa07.site.yml

PR #8095, opened 2024-09-24
Setup uiowa09

bspeare:
Relates to https://github.com/uiowa/uiowa/issues/8060

# How to test

- Compare to previous app setup pr: https://github.com/uiowa/uiowa/pull/7952
- Note that the SVN is the same SVN as what is listed in https://github.com/uiowa/uiowa/blob/main/.travis.yml#L21 so no addition to that file is necessary.
- Compare drush file to current main version of ACN uiowa08: https://github.com/uiowa/uiowa/blob/main/drush/sites/uiowa08.site.yml


joewhitsitt (approved):
looks correct to me

Issue #8064, opened 2024-09-12
Investigate file structure change


briand44:
While grooming https://github.com/uiowa/uiowa/issues/7966, we discussed a probe could be helpful here.

Discussed a probe to look into https://www.drupal.org/project/media_alias_display, reach out to Acquia to see if they have any guidance or limitations to how many files can live in a given directory. Goal of probe to determine if we should move forward with a change here and if so, are we changing the location of where the files are stored or only how they are accessed?

https://github.com/uiowa/uiowa/issues/7966#issuecomment-2344444870

joewhitsitt:
Going to start by opening a ticket with Acquia

joewhitsitt:
Starting to look at the module and here are a couple observations so far:

- The current pattern set is media/[media:bundle:target_id]/[media:mid] which could change, but we ultimately probably want to avoid using the uploaded filename because of the media replace functionality already in place.
- Had to check the "standalone media URL" option under Media > Security config. This makes easier for spammers, etc. to guess and hit file paths by hitting media/file/117, media/file/118, etc.
- When referenced via media field/linkit_media_library the output URL is still the date based URL
- There seems to be a disconnect between existing files and new files. New files, hitting media/file/117 for example will trigger the save as of the browser. Old files will take me to the media entity route where the linked generic file is output.
- Re-saving an existing file causes something like media/file/117-0 to happen and still doesn't trigger the download properly.
- The save file as filename is the alias (e.g. 117, or a custom "my-file" instead of the filename stored on the server)
- There is a sub-module that "adds a media field to override media alias display on specific entities" which I think means for new files takes you to the media entity page instead of triggering download. Seems to have no effect on existing.
- We would have to remove our protection/fix in place surrounding the media alias (editors previously had the ability to change the alias which prevented them from getting to the edit route of the media entity) https://github.com/uiowa/uiowa/blob/3526f8284d4c486e684590c165e668dda846173b/docroot/profiles/custom/sitenow/sitenow.profile#L521

As it relates to afr.fo, this will give them non-date paths to access files directly, but won't prevent date-based paths from showing when referenced and by default won't match the URLs from the v1 site unless they set custom aliases (which as mentioned earlier could cause confusion/conflict with the replace functionality).

joewhitsitt:
Didn't find anything in d8+ that made it seem like an editor could override a default directory setting and upload it elsewhere. It looks like file_paths might be the closest but at a glance I wasn't able to get it to work as I would expect it to.

[...]

PR #8098, opened 2024-09-24
POC - media_alias_display

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/8064

To update existing WYSIWYG links, I think we would have to do a find and replace for `data-entity-substitution="media"` `data-entity-substitution="canonical"`

Additionally, linkit_media_library is hardcoded to use `media` right now: https://git.drupalcode.org/project/linkit_media_library/-/blob/1.0.3/src/LinkitMediaLibraryEditorOpener.php#L76

There is currently an path alias set for all existing media entities - /media/file/101 => /media/101/edit that needs to be bulk deleted/regenerated to prevent the -0 suffix on next media entity save, however, the media overview page's edit operation doesn't point to `media/mlid/edit` until the media entity has been saved again.

# How to test

```
ddev composer install && ddev drush @foafr.local sql-drop -y && ddev drush @foafr.local cr && ddev blt ds --site=afr.fo.uiowa.edu --sync-public-files && ddev drush @foafr.local uli
```


joewhitsitt:
ack. why is this happening:

`Unable to install the <em class="placeholder">Better Exposed Filters</em> module since it requires the <em class="placeholder">jQuery UI, jQuery UI Slider, jQuery UI Touch Punch, jQuery UI Datepicker</em> modules.</li>`

joewhitsitt:
Not sure how those modules got removed from core.extension. Will push a fix.

PR #7246, opened 2024-01-03
Deprovision sandboxnext.sites.uiowa.edu

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7017
Resolves https://github.com/uiowa/uiowa/issues/7202
Resolves https://github.com/uiowa/uiowa/issues/7258

# Test

- Mainly looking for code review as this updated script has successfully deleted files for a site on Acquia Cloud Next (sandboxnext.sites) as well as two sites on Acquia Cloud Classic. The parent site folder will still exist at mnt/gfs/... but according to the acquia ticket, that is the way it is now based on server permissions.

```
/mnt/gfs/uiowa07.stage/sites/sandboxnext.sites.uiowa.edu/files$ ls -la
total 8
drwxr-xr-x. 2 uiowa07.stage uiowa07    6144 Nov 30 16:47 .
drwxr-xr-x. 4 root          cloudadmin 6144 Nov 30 16:47 ..
```

```
/mnt/gfs/uiowa.test/sites/covidresponse.research.uiowa.edu/files$ ls -la
total 0
drwxr-x--- 2 uiowa uiowa  6 Jan 17 18:34 .
drwxr-x--- 4 uiowa uiowa 38 Apr 30  2020 ..
```

PR #8102, opened 2024-09-25
[uiowa09] test instead of stage for drush alias

bspeare:
# How to test

```
ddev drush @uiowa09.test ssh
```

### DIFF
--- a/drush/sites/uiowa09.site.yml
+++ b/drush/sites/uiowa09.site.yml
@@ -16,9 +16,9 @@ prod:
   root: /var/www/html/uiowa09.prod/docroot
   user: uiowa09.prod
 test:
-  uri: uiowa09stage.prod.acquia-sites.com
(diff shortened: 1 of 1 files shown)


joewhitsitt:
I thought originally that this was an ACN difference. Their inconsistency is going to cause problems. I wonder if we should raise a stink about this as we already had to adjust our `umd` command once https://github.com/uiowa/uiowa/pull/7246#issuecomment-1889664092 . I don't know if the transfer command still works or not.

uiowa.test@uiowastg
uiowa02.test@uiowa02stg
uiowa07.stage@uiowa07stage
uiowa08.stage@uiowa08stage
uiowa09.test@uiowa09test

joewhitsitt (approved):
I am fine with moving forward with this to get it to done, but i'd hold off on provisioning sites on this app until we decide on whether it needs to be fixed.

PR #8104, opened 2024-09-25
[Registrar] Event display bugs

### LINKED ISSUE #8090: Some Academic Calendar dates not showing
> I'm wondering if you all could take a look at the academic calendar and see if I'm missing something. It doesn't look like the "five year dates - breaks, close of classes, commencement" are showing up in the list?
> 
> https://registrar.prod.drupal.uiowa.edu/calendars-and-deadlines/academic-calendar
> 
> So fall break should run through 11/24 - 12/1, close of classes should be 12/13, commencement 12/19 - 12/22. Do I have something toggled I shouldn't? Any ideas what's going on?

I confirmed Fall Break is displaying on their current academic calendar - https://registrar.uiowa.edu/academic-calendar

@bspeare I think this is an issue with "reviewed = true". 

----

Josh said reviewed was set to true for these. He is curious if we are excluding five year dates here?
```[tasklist]
### Tasks
- [x] Figure out why some of these dates that were mentioned (Fall Break) are not showing and see if we can get them to show
- [ ] Figure out why some dates are showing in the default "All Student Dates" category when they are not showing in that category on the current live site
```

GaryRidgway:
closes #8090

## To Test

1. Check that some of the dates that were mentioned (Fall Break) are now showing.

2. Check that some dates are no longer showing in the default "All Student Dates" category when they should not have been
3. Check that the dates by session display behaves in the correct way with these changes


cory-skeers (approved):
Checked it out, and the updates look good to me! Nice bonus formatting update too.

PR #8110, opened 2024-09-27
[Commencement] Simplify Commencement program archive

### LINKED ISSUE #8093: Program Archive Updates
# Story

As a commencement site owner, I would like our program archive to display more similar to the Registrar File archive.

# Additional Info

https://commencement.uiowa.edu/programs#Archive

They would like this to display in a single column that would output something like "Spring 2024 Commencement Program" rather than having three separate columns. They would want it ordered with the most recent first and don't need the ability to sort by column header.

GaryRidgway:
# How to test
1. Sync the commencement site and then navigate to [the programs archive page](https://commencement.uiowa.ddev.site/programs#Archive)
2. Observe that each session is linked to a program and that its only one column.
3. Note that you no longer have the ability to sort.


bspeare:
Should we add a `<th>` here for the title?  AI inspector is noting that this table is in violation because "Data cells must have row/column headers."

bspeare:
Does the text need to read as "Spring 2024 commencement program" as noted in https://github.com/uiowa/uiowa/issues/8093?

GaryRidgway:
> Does the text need to read as "Spring 2024 commencement program" as noted in #8093?

For this, in the original comment Brian said something "Like" that, so I took it as being close to it. I figured that there was already enough context to know that the program was from commencement. I am okay with changing it again as well though! I could go either way

GaryRidgway:
> Should we add a `<th>` here for the title? AI inspector is noting that this table is in violation because "Data cells must have row/column headers."

Definitely! I can do that

bspeare (approved):
Works and looks good!

PR #8116, opened 2024-10-01
[Registrar] Final exams

### LINKED ISSUE #8077: Display Final Exam Schedule
# Story

As a Registrar site owner, after placing the final exam schedule block for a give session, I would like to be able to click the link that is output and view that sessions final exam schedule.

# Additional Information

In #4967, we are creating the block plugin. Now we would like to grab and display the Final Exam Schedule based on the legacy code for the session chosen while placing the block.

Use views xml backend or custom api call to pass in legacy code in render exam schedule.

cory-skeers:
Resolves #8077 

# How to test

```
git checkout registrar-final-exams && git pull && ddev composer install && ddev blt frontend && ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /test-exam-schedule
```

Check out the Exam schedule, test out the search box, etc., and compare it to the existing site's: https://registrar.uiowa.edu/final-exam-schedules/fall-2024

Note: There are 2 exam blocks on this test page. In prod, we wouldn't expect needing the second, so overlap between their functionality is probably okay. Feel free to delete one.

Edit the layout and change the session, see that the page doesn't break, and has an appropriate "no results" message.

Check that other MAUI-based services still work correctly, including the session dates at the bottom of this page, the classrooms buildings importer, Academic Dates calendar (https://registrar.uiowa.ddev.site/calendars-and-deadlines/academic-calendar), and others.

(A search for something like `$this->maui` in the codebase can help reveal others to test.)


cory-skeers:
Currently, I have things set up to call the API and parse the results on the fly (this does get cached) to output to a table.

It seems not so bad, or that it _could_ be not so bad if some things were shifted around and cleaned up. Right now it's a bit hacky.

For starters, I was hoping someone could review some of the previous discussion in #4967 re: using `views_xml_backend`, and provide some thought on how this approach might compare to that, now that we have a partially functional example of rolling our own API call/parse.

cory-skeers:
Quick note: came up in a meeting that classrooms may want to enable the same or similar widget to display final exams schedule, so we may want to keep that in mind while creating this.

bspeare:
Pulling to test.

cory-skeers:
Noting here that we demoed this with the registrar folks, and they are going to double-check with a few people that column sorting is not a must-have as long as the search is working. It sounds like that is the case, though, and sorting can be a nice addition at a later date.

bspeare (approved):
Testing instructions all work as expected.   Nice work!

bspeare:
Should we merge this and plan the sortable tables for future work if needed?

cory-skeers:
> Should we merge this and plan the sortable tables for future work if needed?

I think that sounds like a good plan. It sounded like a good chance they'll be fine without it for now, but even if it is higher priority to get done, it can be handled in a separate slice.

PR #8123, opened 2024-10-02
update drupal/diff to 1.8.0

joewhitsitt:
# How to test

Travis ain't complaining.

You could test for a successful install and sync

Compare a revision comparison with the same revision comparison in prod and see that there is no _diff_. Pun intended.

https://git.drupalcode.org/project/diff/-/compare/8.x-1.0...8.x-1.8?from_project_id=369&straight=false

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -104,7 +104,7 @@
         "drupal/date_popup": "^2.0",
         "drupal/default_content": "^2.0@alpha",
         "drupal/devel": "^5.1",
-        "drupal/diff": "^1.1",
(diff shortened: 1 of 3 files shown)


joewhitsitt:
the diff settings update sets the view mode for the visually inline comparison but we have that option disabled and not available for folks to use.

bspeare (approved):

bspeare:
Still shows diffs. 
Screenshot 2024-10-02 at 2 37 23 PM

PR #8155, opened 2024-10-21
Prospector webform handler

### LINKED ISSUE #8138: Allow webform integration with Prospector app to be more widely distributed
# Story

As a web team, we would like to avoid having to touch each webform that needs to integrate with the Prospector app so that we can use our resources elsewhere.

# Additional Info

In #8103 we tested integrating a webform with the prospector app middleware. Since it sounds like there will be potentially a lot of these forms on the Education site and likely other sites as well, we would like to make this easier to distribute. 

Make it more widely available or document or make it easier for webmasters to set this up without our team having to be involved with each form.

pyrello:
Resolves #8138 

# How to test
```
ddev blt ds --site education.uiowa.edu && ddev drush @education.local config-split:activate advanced_webform && ddev drush @education.local uli
```
- Test that error messages are working by adding the Prospector handler to a webform. It should display messages about auth credentials, site UUID. If you want to test the error message related to endpoint URL, you'll need to figure out the best way to temporarily unset the config value for it, which should be set when the site syncs.
- Add the following to the `docroot/sites/education.uiowa.edu/settings/local.settings.php` and replacing the `username` and `password` with the ones stored in esaccess.
```php
$config['sitenow_advanced_webform.settings']['prospector']['auth']['user'] = 'username';
$config['sitenow_advanced_webform.settings']['prospector']['auth']['pass'] = 'password';
```
- Visit https://education.uiowa.ddev.site/admin/config/sitenow/advanced-webform. There is a pinned email in the web team inbox that contains the `siteInteractionUuid` for education.uiowa.edu. This should be entered into the settings form.
- Add the Prospector handler to a webform that ideally already has `first_name`, `last_name`, `email`, and `phone`. Note that the error messages from before are no longer present.
- Submit the form and observe that there is a message indicating that the data was successfully transferred to the application.
- To test error logging, enable dblog locally and comment out line 141 in `Drupal\sitenow_advanced_webform\Plugin\WebformHandler\ProspectorRemotePostWebformHandler`: `$options['json'] = $data;` and then resubmit the form. Visit the logs and see the error printed there.


docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,133 @@
...
+      if ($value = $webform_submission->getElementData($element)) {
+        $data[$element] = $value;
+      }
+    }
+    $data['siteInteractionUuid'] = $site_uuid;
+    $options['json'] = $data;
+
+    // Send http request.
+    try {
+      // @todo Should the URL be stored in config or as a setting?
```
joewhitsitt: i feel like it should be a non-UI config for now similar to the sitenow_events endpoint as I imagine it will be changed soon and need to be updated across sites. ignored so we could do some site-specific testing if needed

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function postSave(WebformSubmissionInterface $webform_submission, $update = TRUE) {
+    // Load configuration.
+    $config = $this->configFactory->get('sitenow_advanced_webform.settings');
+    $site_uuid = $config->get('prospector.site_uuid');
+    // We need a site UUID to proceed.
+    if (!$site_uuid) {
+      // @todo Log this?
```
joewhitsitt: >// @todo Log this?

I'd say so. "Prospector Site UUID is missing."

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+    $options = [
+      'auth' => array_values($auth),
+    ];
+
+    // Add data from first_name, last_name, email, and phone fields.
+    $elements = ['first_name', 'last_name', 'email', 'phone'];
+    $data = [];
+    foreach ($elements as $element) {
+      if ($value = $webform_submission->getElementData($element)) {
```
joewhitsitt, line 136: does this return nicely if the field doesn't exist? with the field api we have to do a lot of that hasField type checking and i am less familiar with the webform elements.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+    if (!$site_uuid) {
+      // @todo Log this?
+      return;
+    }
+
+    // Load basic auth credentials from config.
+    $auth = $config->get('prospector.auth');
+    // We need auth credentials to proceed.
+    if (!isset($auth['user']) || !isset($auth['pass'])) {
+      // @todo I think this needs to be moved similar to the site_uuid check.
```
joewhitsitt: Are you saying moving it to the prepareForm? If so, I think that would be good.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+      // @todo Should the URL be stored in config or as a setting?
+      $response = $this->httpClient->request('POST', 'https://test.its.uiowa.edu/prospect-api/api/prospect/submit', $options);
+      // Prints the message received from the API.
+      $this->messenger()->addStatus($this->t('@response', [
+        '@response' => $response->getBody()->getContents(),
+      ]));
+    }
+    catch (GuzzleException $e) {
+      // Log the exception.
+      $this->getLogger()->error('An error occurred while posting the webform submission to Prospector. Error: @error', [
```
joewhitsitt, line 150: should this also contain `$response->getBody()->getContents()`

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: It does work if the element doesn't exist. I know this because I was testing against a form that used `email_address` and `phone_number`, which is another issue.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function prepareForm(
+    WebformSubmissionInterface $webform_submission,
+    $operation,
+    FormStateInterface $form_state,
+  ) {
+    // @todo This doesn't seem quite right because it might show to an end user.
```
joewhitsitt: Should this be warning messages in `buildConfigurationForm` and be validated in `submitConfigurationForm`?

pyrello: Yeah, I think that makes sense for the message. I'm not sure about the validation. Like would we make it so they couldn't save the handler if there wasn't a UUID? It might just need to be informational.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: I'm not 100% sure, but I think this might be the same output as`$e->getMessage()` because of the type of exception.

pyrello:
Going to work on some changes based on feedback.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: I don't think it is too bold to require values for those before the handler is set and saved. We do it for the thank you block:
Screenshot 2024-10-31 at 10 29 47 AM

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: would a bad uuid be a way to test/confirm this?

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: To be honest, I really don't like the UX of the above. Someone can add the block only to find out that they can't use it. Preventing it from showing in the list of blocks and providing a help message somewhere else seem like a better pattern.

I think allowing the handler to be added and then adding a message either on the configuration form and/or the handlers page that it won't submit to the app until that extra step is taken is an okay user experience. Its annoying to be at the point where you are trying to submit a form, you have to leave to go do something (add the UUID to a different config form) and then come back to add the handler again.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: I think this would go in the same place(s) we put the messages about the UUID.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: I think so.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: i get what you are saying. i think it is a balance of working within/leverage what already exists/is available.

joewhitsitt:
Do the authentication instructions need to be updated? I am having troubles with the current authentication checks again.

pyrello:
> Do the authentication instructions need to be updated? I am having troubles with the current authentication checks again.

Whoops! Yes, I have updated them.

joewhitsitt:
Can a menu link be added for the settings form similar to docroot/modules/custom/uiowa_core/uiowa_core.links.menu.yml?

pyrello:
> Can a menu link be added for the settings form similar to docroot/modules/custom/uiowa_core/uiowa_core.links.menu.yml?

I added this but it didn't get added to git. It is fixed in the last commit.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: a bad uuid seems to still return a 200 OK and not trigger this

pyrello: Try commenting out line 113: `$options['json'] = $data;`

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
pyrello: `buildConfigurationForm` doesn't seem to work or least I'm not sure how to add a message that gets displayed on the form since it is loaded into the sidebar via AJAX. The message about the Site UUID gets displayed after the page refreshes. Probably there is a way to set this correctly via the `$form` or `$formState`. Either way, it should possibly also be displayed on the handler page if the handler is enabled.

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: That worked and does include the response.

Screenshot 2024-10-31 at 12 00 52 PM

docroot/modules/custom/sitenow_advanced_webform/src/Plugin/WebformHandler/ProspectorRemotePostWebformHandler.php
joewhitsitt: ah. i think we encountered that with pardot as well and just did a form api markup element: https://github.com/uiowa/uiowa/blob/316d7964ab5bf8724e02cd3f5702fba67b048dce/docroot/sites/tippie.uiowa.edu/modules/tippie_core/src/Plugin/WebformHandler/PardotRemotePostWebformHandler.php#L42

joewhitsitt:
```
if (!$site_uuid) {
        $form['missing_uuid'] = [
            '#markup' => $this->t('<strong>Warning:</strong> The site UUID is missing. Please configure it...
...
```

pyrello: I added something but I'm not figuring out how to position it at the top of the form. I wondering if this is enough now and the message can be improved in future work?

pyrello:
An open question I have not resolved yet: The form I was testing against used `email_address` and `phone_number` as the field names. However, the original round of investigation and testing used `email` and `phone`. Any thoughts on how this should be defined?

joewhitsitt:
With Tippie, the webform conformed to the upstream, so my gut would say that we should design it how the upstream expects it and the forms would either be adjusted or re-created. It sounded like AIS was going to be "supporting" this connection once available and working with the downstream implementers.

pyrello:
> With Tippie, the webform conformed to the upstream, so my gut would say that we should design it how the upstream expects it and the forms would either be adjusted or re-created. It sounded like AIS was going to be "supporting" this connection once available and working with the downstream implementers.

I'm going ahead with the assumption that we'll use `phone` and `email` for now.

joewhitsitt:
Pulling to test

joewhitsitt:
Is it meant for webmasters to set the UUID? If so, I think the menu link permission needs to be adjusted.

joewhitsitt:
Should there be guidelines text on the handler form or the settings form to explain the prospector stuff more to the user or is the assumption that the webmaster coming across it has already been briefed about it?

"Data received and enqueued for processing." Should form submitters (anonymous visitors) see this message or something friendlier that the response from the upstream?

joewhitsitt (approved):
Made a couple of minor changes after testing. Everything else looks fine.

joewhitsitt:
Going to follow-up with putting the creds in the application environments' secrets file.

joewhitsitt:
Probably overkill and will need to be updated later but creds have been added to all uiowa* apps/all envs

PR #8158, opened 2024-10-22
Sprint 235 provisions

GaryRidgway:


pyrello (approved):
4 sites * 7 file additions per site + 2 aggregate file changes = 30 files changed

Note: It's `+ 2` now instead of `+ 1` because manifest.yml gets changed in addition to sites.php

PR #8160, opened 2024-10-22
[Search] Search toggle focus

### LINKED ISSUE #5372: As a site user, when I click on the magnifying glass to search the site, I don't want to have to touch the search area to type.
# Story

As a site user (desktop and mobile), when I click/touch the magnifying glass to search the site, I don't want to have to click/touch the search area also to type. I would like touching the magnifying to activate my keyboard.

image

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5372. 

# How to test

Reading through https://github.com/uiowa/uids/pull/241 it sounds like we removed the focus because the focus change was happening before the aria-expanded value on the toggle button is announced.  This pr adds a delay on the focus change, and we have created an aria-controls reference on the button to the aria-hidden element that toggles visibility. Finally we have added an aria-live polite announcement informing the user that the search form has opened/been closed.

1. Code review
2. Sync a site or go to [sandbox.dev.drupal.uiowa.edu/](https://sandbox.dev.drupal.uiowa.edu/)
3. Turn on Voiceover
4. Tab through to the search toggle on Chrome

**Testing that has been confirmed by TMW and user testing:**
Original test tasks:
Objective: Verify that the search functionality is accessible to screen reader users, including the automatic focus on the search input and clear feedback on the visibility of the search form.
Locate the search form button using your preferred screen reader navigation method.
Observe/Record: Are you able to locate a button that will open the search form?
Trigger the button.
Observe/Record: Are you able to understand what happens once the button is triggered. Is focus switched to the search form input and are you aware of that change?
Start typing a search query and then locate the search submit button. Do not submit the search query.
Observe/Record: Are you able to navigate the search form and locate the submit button?
Navigate away from the search form.
Observe/Record: Do you pass by a button to once again toggle the visibility of the search form?

**(Results of user’s experience)**
I was able to navigate to the search form.
When I did, it said “toggle search button collapsed.” Not sure why it’s collapsed.
I hit enter and that opened the search field.
I was able to note the focus change.
I easily entered data into the search field.
I tabbed once and found submit.
I navigated away from the search field by hitting escape.  I arrowed down and came to the search input field which contained the text of my previous search.


joewhitsitt:
Ben asked me to jog my memory a bit about this issue. aria-expanded is just an attribute that is read when the element has focus, not an "always announce this to the user" type thing. with the current PR, it does start to get read before switching focus over to the input. The audio portion gets cutoff because of the focus change.

Screenshot 2024-10-30 at 3 50 51 PM
Screenshot 2024-10-30 at 3 51 00 PM

The the previous examples that TM has shared with Accordions, the focus does not switch so the attribute is read.

To further sprinkle some UX enhancements to this, we can add aria-controls to show what is toggled in visibility with this button. we could have a more descriptive aria-label for the button to separate "search" from "search."

Finally, this might be an area that we try sprinkling in some more drupal.Announce() type messages.

joewhitsitt:
Thought about `this.searchButton.setAttribute('aria-label', isSearchOpen ? 'Open search form' : 'Close search form');` but because the button does have focus for a bit before switching, it starts to read "Close search.." and then switches.

joewhitsitt:
Re-deployed to DEV. Going to send TM an email early next week (11/5).

joewhitsitt:
Sent an email to TM via its-web

joewhitsitt (approved):
Biased at this point, but let send this out for live testing

bspeare:
Thanks @joewhitsitt!

Issue #8143, opened 2024-10-15
htaccess redirect for writinguniversity.uiowa.edu and www.writinguniversity.uiowa.edu


briand44:
These redirects currently live on the grad application. They want these redirects to remain once the grad app is gone.

Discussed in PL and since long term we would like the site to live at writinguniversity.uiowa.edu, we determined that it may make sense to redirect this using .htaccess rather than vulture so we don't need to update DNS multiple different times.

Michael mentioned no leadership capacity to have the conversations at this time to start moving to writinguniversity.uiowa.edu rather than writinguniversity.org.

```[tasklist]
### Tasks
- [x] cover writinguniversity.uiowa.edu and www.writinguniversity.uiowa.edu in uiowa01 ssl cert
- [x] add htaccess redirect that will redirect writinguniversity.uiowa.edu and www.writinguniversity.uiowa.edu to writinguniversity.org (might be able to copy from grad htaccess) - *note - currently if you go to writinguniversity.uiowa.edu/about it redirects to writinguniversity.org/about and not to the home page.
- [x] move www.writinguniversity.uiowa.edu and writinguniversity.uiowa.edu domains from grad app to uiowa01
```

joewhitsitt:
Noticed `humanities.uiowa.edu` is still covered and pointing at acquia but was deprovisioned. https://github.com/uiowa/uiowa/issues/5286
Noticed `coronavirus.uiowa.edu` and `www.coronavirus.uiowa.edu` are covered but was deprovisioned/redirected.

Going to clean up as this app already has a lot of SANs

joewhitsitt:
Sent an email to hostmaster to reclaim humanities.uiowa.edu

joewhitsitt:
Always forget we patch htaccess. Can't remember if we have docs on that...

Update: We do now! https://docs.uiowa.ddev.site/development-workflow/updating-patched-htaccessdevelopmentservicesrobotstxt-files

joewhitsitt:
Also, I think the old rule is covering writinguniversity.net. We don't need that anymore right?

```
RewriteCond %{HTTP_HOST} ^(www\.|)writinguniversity\.(net|uiowa\.edu)$ [NC]
RewriteRule ^(.*)$ https://www.writinguniversity.org/$1 [L,R=301]
```

```
RewriteCond %{HTTP_HOST} ^(www\.|)writinguniversity\.uiowa\.edu$ [NC]
[...]

PR #8170, opened 2024-10-25
Writinguniversity htaccess redirect

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/8143

# How to test

- Compare with acquia/uigraduatecollege-dev/docroot/.htaccess (removed the .net coverage)
- Check against https://htaccess.madewithlove.com/ though I think we got burned last time we tried to do one on the grad application (my guess was that there was a conflicting rule mucking things up).
- Deploy to PROD and do a quick domain switch to see?!

### DIFF
--- a/docroot/.htaccess
+++ b/docroot/.htaccess
@@ -2,6 +2,15 @@
 # Apache/PHP/Drupal settings:
 #
 
+# Block these IP addresses.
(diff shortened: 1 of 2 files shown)


bspeare (approved):
Looks correct!

PR #8172, opened 2024-10-25
[ Registrar ] Course deadlines query parameters.

### LINKED ISSUE #8157: Allow query parameters for Course Deadlines
# Story

As a member of the MAUI/MyUI team, I would like the ability to pass query parameters to auto-populate the form fields on the course deadlines page so that we can have a schedule view in MyUI that links directly to the deadlines without them having to manually select each dropdown.

# Additional Info

https://registrar.prod.drupal.uiowa.edu/calendars-and-deadlines/course-deadlines

> I work as an application developer on Becky Tjelmeland's MAUI/MyUI team and I'm working on a page in MyUI that needs to link over to the Registrar website's Course Deadlines page. Becky thought there would be a way to auto-populate each of the dropdowns on that page (Session, Department, Course, Section) via query parameters in a URL. I tried to pre-populate the Session and Department fields this way and it did not work: https://registrar.uiowa.edu/course-deadlines?session=20243&department=ABRD. Do you know if this is possible and how the URL can be updated to auto-populate these fields? I want to create a schedule view in MyUI where each course has a deadlines link which takes the student directly to the deadlines for the course on the Registrar website without them having to manually select from each dropdown.

GaryRidgway:
Closes #8157 

## To test
```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /calendars-and-deadlines/course-deadlines
```

1. add `?session=20243` to the URL and enter. Observe that the session field is filled.
2. Additionally, without removing the previous params, add `&department=NURS` to the URL and enter. Observe that the department field is filled.
3. Additionally, without removing the previous params, add `&course=3600` to the URL and enter. Observe that the course field is filled.
4. Additionally, without removing the previous params, add `&section=1018283` to the URL and enter. Observe that the section field is filled.
5. Now that you have one URL structure working, can you mess up what I've built?
6. Test other sessions and sections by getting their keys from the html select element for the session select field. Test the rest of the fields by using their indices, which are the same as their field values.
7. Try and find other bugs?

## To test part 2

1. Ensure that both the deadlines form and the correspondence form are not broken, and can both still use query parameters.
    - https://registrar.uiowa.ddev.site/correspondence
    - https://registrar.uiowa.ddev.site/calendars-and-deadlines/course-deadlines

### NOTE:
- The form is not going to change URL params when you change the form fields.
- If you put in incorrect data into the URL params, it may fill out the fields out of order and disable them incorrectly. This is not something we currently need to worry about. You may disregard.


joewhitsitt:
@GaryRidgway did I hear correctly that we have done stuff like this previously? Should `getFormValue()` be moved somewhere where it could be used again or is it too specific to course deadlines?

GaryRidgway:
> @GaryRidgway did I hear correctly that we have done stuff like this previously? Should `getFormValue()` be moved somewhere where it could be used again or is it too specific to course deadlines?

Currently it requires the class its implemented in to have the `$this->cachedFormState` and `$this->cachedParams` private variables defined, so I don't think its currently abstractable. if we could find a way to abstract those, then yes?

I guess we could make a new class that holds those values?

joewhitsitt:
Okay, without knowing about a future need, maybe we just sit on it for now, but i would think this could be the start of a FormHelpers trait and those would just be passed in instead of as private.

```php
<?php

use Symfony\Component\HttpFoundation\InputBag;
use Drupal\Core\Form\FormStateInterface;

trait FormHelpers {

  public function getFormValue(
    string $param_index,
    array $param_allowed,
    FormStateInterface $form_state,
    InputBag $params
  )
```

```
$session = $this->getFormValue('session', $this->sessionOptions(), $form_state, $params);
```

I can review this as-is though.

GaryRidgway:
> Okay, without knowing about a future need, maybe we just sit on it for now, but i would think this could be the start of a FormHelpers trait and those would just be passed in instead of as private.
> 
> ```
> <?php
> 
> use Symfony\Component\HttpFoundation\InputBag;
> use Drupal\Core\Form\FormStateInterface;
> 
> trait FormHelpers {
> 
>   public function getFormValue(
>     string $param_index,
>     array $param_allowed,
>     FormStateInterface $form_state,
>     InputBag $params
>   )
> ```
> 
> ```
> $session = $this->getFormValue('session', $this->sessionOptions(), $form_state, $params);
> ```
> 
> I can review this as-is though.

Yeah, I had this as the call in a previous iteration, but it was a long call signature and I was trying to slim it down. You are right though, this technically works fine. Not opposed to splitting it out if you are interested in starting that up

joewhitsitt (approved):
This works as expected. My nitpick is that it doesn't trigger the ajax callback so there is no info saying "returning information for ..."

I assume this was demonstrated at the meeting today and we are good from the customer sides?

e-marie-w:
The section param is not filling in for me. The testing example only has one section so it is auto-selected and works with only the session, department, and course params. Something with multiple section options, like https://registrar.uiowa.ddev.site/calendars-and-deadlines/course-deadlines?session=20243&department=ACCT&course=2100&section=0A04 does not work. I can investigate further.

GaryRidgway:
> The section param is not filling in for me. The testing example only has one section so it is auto-selected and works with only the session, department, and course params. Something with multiple section options, like https://registrar.uiowa.ddev.site/calendars-and-deadlines/course-deadlines?session=20243&department=ACCT&course=2100&section=0A04 does not work. I can investigate further.

Good catch. Looks like those are done in codes too, not the actual section number. using `section=1021855` works for me.

e-marie-w (approved):
New course deadlines query params work, and existing correspondence query params continue to work!

PR #8191, opened 2024-10-30
Back to top - accessible name

joewhitsitt:
# How to test

Fixes accessible name and the icon requiring alt text.

Sync sandbox and use the SiteImprove accessibility extension for chrome based browsers. Noticed the issues related to the back to top button are fixed locally compared to PROD.

Screenshot 2024-10-30 at 4 59 49 PM
Screenshot 2024-10-30 at 4 59 59 PM


e-marie-w:
So the SiteImprove extension flagging looked different for me than the screenshots, but given it was about the back to top button and that it didn't show up locally compared to prod, I feel comfortable approving.

e-marie-w (approved):

PR #8194, opened 2024-11-01
[Registrar] Remove 'last updated' final exam block

### LINKED ISSUE #8185: Remove last updated functionality from Final exam schedule
# Story

As a Registrar site owner, we no longer need the last updated date included in the Final Exam Schedule functionality.

# Additional Info

Screenshot 2024-10-29 at 2 54 35 PM

Screenshot 2024-10-29 at 2 55 18 PM

```[tasklist]
### Tasks
- [x] Remove last updated date from display
- [x] Remove last updated field from block config
- [x] Remove any other code related to the last updated functionality for Final Exam Schedules
- [x] Probably need to re-save the block once we remove this field
```

cory-skeers:
Relates to #8185 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /exams
```

See that the page loads fine, doesn't have a 'last updated' displayed.
Edit layout, edit the block and try adding a new final exams schedule block, and see there is no 'last updated' option.

Note: Existing blocks still have the value in their block config, and re-saving the block does not remove it. As there are very few instances of this block, I think we're fine to remove and re-add the block in prod after this code is released to clear it out, rather than the otherwise necessary update hook. The block loads and displays fine even with the unused value in place.

### DIFF
--- a/docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Form/FinalExamScheduleForm.php
+++ b/docroot/sites/registrar.uiowa.edu/modules/registrar_core/src/Form/FinalExamScheduleForm.php
@@ -77,11 +77,10 @@ class FinalExamScheduleForm extends FormBase {
     }
 
     $session_name = $session_info['session_name'];
-    $last_updated = date('F j, Y', strtotime($session_info['last_updated']));
 
     $form['final_exam']['intro'] = [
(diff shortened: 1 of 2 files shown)


pyrello:
Pulling to test.

pyrello (approved):
Works as expected.

PR #8199, opened 2024-11-01
Sticky indicated cards

### LINKED ISSUE #6695: Indication of sticky/featured content for cards
# Story

As a website visitor, I can tell what content has been marked as sticky/featured both visually and with a screen reader..

# Proposed solution

- [x] Be aware of where sticky is available
- [x] Implement new pre-title slot in card in uids for icon and pinned text in https://github.com/uiowa/uiowa/issues/6695#issuecomment-2389530290
- [x] Update the card teaser to indicate what content has been marked as sticky and is not following the normal sort.
- [x] Make sure non-visual users can get the same information when traversing the content.

joewhitsitt:
Resolves #6695

Screenshot 2024-11-01 at 3 50 56 PM

# How to test

- `ddev blt frontend` - ignore the depreciations/warnings. this is to account for uids_base's switch from node-sass to sass.
- Mark one or more content nodes as sticky.
- Using any view that has a sticky sort on it, see that the content is sorted as such and has the new UIDS indicator using the pre-title slot.
- Using VoiceOver, traverse the non-tab indexed content with ctrl+option+right arrow. VoiceOver should read out "Pinned content, custom sorted."
- Go in to that view and remove the sort sticky. Save the view, clear cache and reload the placed view. There should be no sticky indicators. Even though the content was marked as sticky globally, in this context it is not.
- Generally check other styles to make sure there are no regressions with the way the uids_base SCSS files are compiled.

# Before Merge

- update uids hash


docroot/modules/custom/uiowa_core/src/Entity/NodeBundleBase.php
```diff
@@ -71,6 +72,19 @@ public function buildCard(array &$build) {
     if (isset($build['#cache']['keys'])) {
       unset($build['#cache']['keys']);
     }
+
+    // Indicate sticky content if part of view with sticky sort applied.
+    $view = $this->values['view'] ?? NULL;
+    if ($view instanceof ViewExecutable) {
+      $sorts = $view->getDisplay()->getOption('sorts');
+      if (!empty($sorts)) {
+        if (array_key_exists('sticky', $sorts) && $this->isSticky()) {
```
GaryRidgway, line 81:
```suggestion
    $view = $this?->values['view'] ?? NULL;
    if ($view instanceof ViewExecutable) {
      $sorts = $view?->getDisplay()?->getOption('sorts');
      if (!empty($sorts)) {
        if (array_key_exists('sticky', $sorts) && $this?->isSticky()) {
```

I think that this can be changed for better coding safety?

Otherwise, I think the way you are checking for sticky seems good to me as well, as there is no way that a card will be sticky outside of a view, so checking there is good!

GaryRidgway: Definitely check me on my logic here though, perhaps I am remembering the safety question marks incorrectly

joewhitsitt: At the point this code is called I don't think those additional checks are necessary. The node object ($this) will at least have a `nid` in values and the getDisplay, getOption and isSticky are all safe core functions being called in the right context.

My question is more of should this logic be placed in NodeBundleBase as it is first time we are doing something like this (node card based change that is part of a view that isn't necessarily a listblock) or if it should live elsewhere.

pyrello: Using `?->` and `?? NULL` is redundant. `?->` assigns the value to `NULL` if it doesn't resolve to a value.

GaryRidgway: I think it should live there. I think it is a fine place. I can't off the top of my head think of a reason to put it anywhere else specific.

bspeare:
Pulling to test.

bspeare:
Sorry @joewhitsitt, I forgot to test for background colors and it looks like we need the pre-title set with opacity to make it adjust to the background colors: https://github.com/uiowa/uids/pull/941.

bspeare (approved):
This looks good, nice work @joewhitsitt!

joewhitsitt:
Thanks @bspeare

PR #8206, opened 2024-11-06
[Tables] Table header scope module

### LINKED ISSUE #8152: Probe table header scope issue
# Story

As a Digital Accessibility Coordinator, I would like to see SiteNow tables include the `scope` attribute on table headers without content editors having to manually apply the html.

# Additional Info

This may have been a regression introduced when upgrading to ckeditor 5.

A couple upstream issues:
https://www.drupal.org/project/drupal/issues/3427076
https://github.com/ckeditor/ckeditor5/issues/3175

A possible related issue in our backlog:
https://github.com/uiowa/uiowa/issues/7440

Should we probe this to see if there is anything we can do to help the upstream issues or to resolve this for SiteNow while we wait for solutions upstream?

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8152. 

# How to test

```
ddev composer install && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /table-without-scope
```
1. Confirm `scope="col"` is on each of the `<th>` tags.
2. Add a new table with a header row. 

Screenshot 2024-11-06 at 11 28 22 AM

2. Save the page and inspect the markup and confirm that `scope="col"` is on each of the `<th>` tags.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -190,6 +190,7 @@
         "drupal/stage_file_proxy": "^2.1",
         "drupal/superfish": "^1.6",
         "drupal/symfony_mailer": "^1.3",
+        "drupal/table_header_scope_attribute": "^1.0",
         "drupal/taxonomy_path_breadcrumb": "^1.0@beta",
         "drupal/theme_permission": "^2.0",
         "drupal/tvi": "^2.0@alpha",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f9512f2548e8bf2be9fe7749babe71c9",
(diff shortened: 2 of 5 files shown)


bspeare:
This is ready for review.  Should we create a separate issue for the `'allowed_sources' is not a supported key` work or do we think it should be included as part of adding this module?

joewhitsitt:
just curious on how you worked around the error on save? if this works, the allowed_sources and such can be a separate slice.

Sean raised some concerns about the UX with this solution in the issue. I take it you are fine with moving forward with the module's solution?

bspeare:
@joewhitsitt I just added the config manually which we have done at other times when we needed to update the formats.  I'm fine with the UX because it solves our accessibility problem.

GaryRidgway:
pulling to test

GaryRidgway (approved):
Re-tested this after the twig fix and it is still working

bspeare:
Thanks @GaryRidgway!

PR #8216, opened 2024-11-07
hotfix - update events_load to include a timeout and error handling

joewhitsitt:
# How to test

Temporarily change docroot/modules/custom/sitenow_events/sitenow_events.module#L764:

```php
      $request = \Drupal::httpClient()->get($endpoint, [
        'timeout' => 10.0,
      ]);
```
to
```php
      $request = \Drupal::httpClient()->get($endpoint, [
        'timeout' => 0.1,
      ]);
```

Make sure to clear caches `ddev drush @sitename.local cc bin uievents`

Try to load a page with an events block. See that error handling/message occur. Change back to 10.0 seconds and try again. It will either error again for a complex uncached queries or successfully pass and display events.

Connection issues (timeouts, network failures) via ConnectException.
HTTP-related errors (status codes or response-related issues) via RequestException.

I don't think we need the TransferException. "lower-level transport errors ... though these are rare in well-maintained HTTP requests"

You should also be able to load the field options within the block configuration under normal circumstances. If the timeout occurs, you won't be able to open and configure the block.

### DIFF
--- a/docroot/modules/custom/sitenow_events/modules/sitenow_events_paragraph/sitenow_events_paragraph.module
+++ b/docroot/modules/custom/sitenow_events/modules/sitenow_events_paragraph/sitenow_events_paragraph.module
@@ -138,8 +138,14 @@ function sitenow_events_paragraph_preprocess_paragraph(&$variables) {
           }
         }
         else {
+          if (isset($data['error'])) {
+            $message = '<div class="events-empty"><p>Unable to fetch events at this time.</p></div>';
+          }
+          else {
+            $message = '<div class="events-empty"><p>There are currently no events to display.</p></div>';
+          }
           $variables['list'] = [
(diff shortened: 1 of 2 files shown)


GaryRidgway:
pulling to test

joewhitsitt:
This doesn't do anything for the filters on the block configuration. `_sitenow_events_get_filter_field_options`

It would take some more digging to figure out how to gracefully fallback if no options can be fetched, but the current response `$data['error']` shouldn't mess with those currently.

GaryRidgway:
I tested that this would actually display events, and it does that. I am also able to configure the block, but I am unsure what I am missing from previous comments. either way, this seems like a good enough hotfix for now.

GaryRidgway (approved):
LGTM

PR #8242, opened 2024-11-13
[Brand icons] Brand icon media type

### LINKED ISSUE #8211: Make icon library available to SiteNow
[Grooming conversation was started here.](https://github.com/uiowa/uiowa/issues/6978#issuecomment-2460642513)
- How are we making these available to SiteNow?
    - Static job?
    - Copy paste?
    - Remote hosting?

**Proposed solution**:
- Talk with the team about a way of moving forward with importing/copying/remote hosting the icons
- Proceed to do so
- THIS DOES NOT INCLUDE IMPLEMENTING THEM IN ANYTHING

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8211. 

# How to test

```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /admin/content/media
```
1. Sync site
2. Uncomment update hook
3. `ddev drush @sandbox.local updb`
4.  Go to media page to confirm icons were imported, compare imported data to icons.json. 
5. Go to `/admin/structure/block-content/manage/uiowa_card/fields/block_content.uiowa_card.field_uiowa_card_image` and set "Brand Icon" as a media type. 
6. Create a card block and add an icon as an image. 

## Still to-do
- [ ] Remove Brand Icon from `/admin/content/media-grid` and `/admin/content/media views`


joewhitsitt:
Ben asked for a pre-review on this. Pulling down.

joewhitsitt:
Starting to lean towards keeping most of this in sitenow_media_wysiwyg were the media type is defined (aka move the import function from uiowa_core and whenever there is an update, call it from sitenow_media_wysiwyg.install). I can be persuaded on this though.

joewhitsitt:
Are you thinking the new media type is rolled out first and then a follow-up deployment to run the import? Otherwise I think we need to do some config importing as part of the update hook.

bspeare:
@joewhitsitt good question.  It seems like we could roll this out first since we aren't ready to attach it to the card yet.  What's your preference?

joewhitsitt:
>It seems like we could roll this out first since we aren't ready to attach it to the card yet. What's your preference?

#works4me

joewhitsitt:
i think this is looking pretty nifty! Seeing https://github.com/uiowa/uiowa/pull/8242/commits/7c20506420fca48da37adc09223c1b4356f49316 i wonder if we should start thinking about if/how different icon variants will be implemented and called. worse case it looks like more db rows and we can see about adding a filter to the selection method.

bspeare:
@joewhitsitt good point.  How would we want to prepare for that?

joewhitsitt:
it would be awesome if core caught up and allowed in-place overrides of the media entity for the media reference field (e.g. context specific alt text). https://www.drupal.org/project/drupal/issues/3023807

or should we be hand holding and changing the icon based on the parent background? might be tricky to get in spots..

just like the banner media we could expose a separate field when media type brand icon is selected? not sure how WYSIWYG would work, but there is context specific alt text overrides currently for images.

off hand i don't know what changes would need to be made to the json file if we went the more db rows route.

bspeare:
Thanks @joewhitsitt.  I think it sounds like the Brand team likes the idea of limiting the number of choices and only offering the two-color version.   That could change in the future, but for now I like the idea of offering one choice.

joewhitsitt:
Sounds good. Thanks for looking into this. None of the solutions seemed very palatable.

If things could just move to sitenow_media_wysiwyg and we hold off on the update hook for now, i'll approve this.

joewhitsitt (approved):
Looking good. Nice work!

bspeare:
Thanks @joewhitsitt!

PR #8251, opened 2024-11-15
Update find text

cory-skeers:
# How to test

```
ddev composer install
```

Sync a site and uli
See that the Find Text link is still in its usual spot with icon and all
Open Find Text link and see there are new "Workflow," "Status," and "Content Type" options
Run some searches, see that things work as expected


e-marie-w:
Find Text is in its usual spot and works the same as always. The URL does update with a query parameter, however if you search another term without first hitting Reset the query will not update to the second search. Not sure this needs changing but I wanted to make it visible. Changing what workflow to search works but has some interesting results I also wanted to make visible. Basically, the result will appear for every state before its current state. 

For example, I created 4 different pages with the same phrase on them. I saved one as draft, one as review, one published, and the last I published and then set to archived. If I search that phrase with workflow set to All, the four pages show up. The four pages also show up if I set workflow to Draft. If I set it to Review three show up - the Review, Published, and Archived pages. Published gets the Published and Archived pages, and Archived only shows the Archived page. When set to All, each result correctly identifies the current state of its page. When I choose Draft, though, all four pages are listed as in Draft status even though only one actually is. That part is especially confusing.

I think adding this to uiowa_core is fine.

joewhitsitt:
Could we use claro icons like we used stable in the past (currently needs to be fixed for brand since we switched to stable9)?

core/misc/icons/868686/magnifier.svg

docroot/sites/brand.uiowa.edu/modules/brand_core/css/toolbar-overrides.css

Screenshot 2024-12-09 at 3 15 14 PM

joewhitsitt:
I think there is also an issue between websites with different workflows. On brand, I don't see the published workflow state of the editorial workflow.

Brand
Screenshot 2024-12-09 at 3 14 06 PM

Default
Screenshot 2024-12-09 at 3 18 56 PM

joewhitsitt:
I wonder if we can "hide" this functionality and wait for upstream adjustments?

cory-skeers:
> I wonder if we can "hide" this functionality and wait for upstream adjustments?

Ugh. Either that, or the "upstream adjustments" may be a heavy-handed removal of some of the community contributions to the module.

joewhitsitt (approved):
Yes! Thanks for maintaining and updating this!

Issue #8218, opened 2024-11-07
Fix FilterIframe and text format configuration


joewhitsitt:
# Problem
Related to https://github.com/uiowa/uiowa/pull/8206

We get an allowed_sources error when trying to save filtered_html's configuration page. There is something wrong with how we are implementing this functionality.

docroot/modules/custom/uiowa_core/src/Plugin/Filter/FilterIframe.php

# Steps to reproduce
Try to save admin/config/content/formats/manage/filtered_html through the UI

# Proposed solution

- [x] Understand and fix what we are doing wrong with this type of text format filter. We still need to capture these allowed sources and let them render without data loss.
- [x] There is also an issue with `allowed_html` within the filtered_html and possibly other text formats where element attributes are not allowed since we don't have a CKEditor plugin that supports it. Navigate this new norm without content data loss.

joewhitsitt:
i think the issue stems from a missing or incomplete schema definition

joewhitsitt:
Found some links for reference but I think the main thing is that we aren't establishing schema for some of these things and there is more validation going on.

https://www.drupal.org/project/drupal/issues/3395562
https://www.drupal.org/node/3419181
https://www.drupal.org/project/drupal/issues/3401837

joewhitsitt:
The allowed_html issue is a regression https://www.drupal.org/project/drupal/issues/3410100. Going to update the PR to only focus on the FilterIFrame bit.

joewhitsitt:
For the allowed_html, we could try a module like https://www.drupal.org/project/ckeditor5_allowed_html or create a plugin ourselves that allows all of our existing attributes currently stored in config

PR #8257, opened 2024-11-18
Fix filter_iframe schema

joewhitsitt:
Relates to #8218
 

# How to test

On `main` try to save the admin/config/content/formats/manage/filtered_html configuration. You will get multiple errors, but one of them will be about allowed_sources not being a valid key.

Go to admin/reports/config-inspector/filter.format.filtered_html/list and see that filters.filter_iframe.settings.allowed_sources has some undefined columns and errors.

On `fixing_text_formats` re-sync a website and try saving the admin/config/content/formats/manage/filtered_html configuration again. You might still get multiple errors, but the allowed_sources one won't be one of them.

Go to admin/reports/config-inspector/filter.format.filtered_html/list and see that filters.filter_iframe.settings.allowed_sources no longer has issues.


cory-skeers (approved):
Can confirm, still get some errors in the `filtered_html` config page but unrelated to the allowed sources. Undefined problems were gone too.

joewhitsitt:
Yeah leaving the parent issue open for the allowed_html bug

PR #8258, opened 2024-11-19
Academic Dates Block - updating to filter on print and review dates

### LINKED ISSUE #7962: academic dates block should display dates that are "reviewed" and "printDate"
# Story

As a Registrar Office, we would like the academic dates block to display dates that are "reviewed" and "printDate" so that is aligns with the academic calendar.

# Additional Info

The Academic dates block component that we originally developed for Grad College as part of the uiowa_maui port should also be checking that dates are "reviewed" = TRUE and "printDate" = TRUE similarly to how the academic calendar is.

Josh brought this up as they are using this block on the [Commencement website](https://commencement.uiowa.edu) and noticed some dates shouldn't be displaying. They said this change should happen for the service knowing it will affect dates being displayed on Grad College and Housing websites.

Example of current academic dates block.
Image

How it should appear, similar to academic calendar.
Image

```[tasklist]
### Tasks
- [x] Get a list of sites using this feature
- [x] Communicate the upcoming changes to the sites with this feature enabled
- [x] update to only display dates where `reviewed = true` and `printDate = true`
- [x] Give Josh a heads up before this gets deployed
```

james917:
Fixes https://github.com/uiowa/uiowa/issues/7962

# To test

1.  Sync commencement and registrar
2.  Compare the filtered down academic calendar on register with the filtered commencement home page block (current session, graduation and commencement)
3. "on Oct 4 there should be one degree application deadline. Oct 5 there should be one degree application late fee rather than 4"

There isn't an easy date to point out that is not reviewed but isn't also printdate false when looking at the current session/category. If you were to remove the printdate argument and filter on student registration, there would be a couple dates where reviewed is null.

### DIFF
--- a/docroot/modules/custom/uiowa_maui/src/Form/AcademicDatesForm.php
+++ b/docroot/modules/custom/uiowa_maui/src/Form/AcademicDatesForm.php
@@ -112,12 +112,16 @@ class AcademicDatesForm extends FormBase {
       ],
     ];
 
-    $data = $this->maui->searchSessionDates($current, $category);
+    $data = $this->maui->searchSessionDates($current, $category, TRUE);
 
     if (!empty($data)) {
       $data = ((int) $limit_dates === 1) ? array_slice($data, 0, $items_to_display, TRUE) : $data;
 
       foreach ($data as $date) {
+        // Skip dates that are not reviewed.
+        if ($date->reviewed !== TRUE) {
+          continue;
+        }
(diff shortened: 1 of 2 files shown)


bspeare (approved):
Commencement now appears like the screenshot on https://github.com/uiowa/uiowa/issues/7962. 

Screenshot 2024-11-20 at 3 29 39 PM

PR #8261, opened 2024-11-19
[ ITS ] Retitle Article to News Article

### LINKED ISSUE #8244: ITS - Rename "Article" content type to "News Article"
# Story

Given the ITS website has a custom content type called "Support Article" they have found that the "Article" content type causes confusion among editors. They would like to rename the content type for their website to make it easier for editors.

# Proposed solution

- [ ] Change human readable labels for the article content type to "News Article" and package it up in the ITS website site split

GaryRidgway:
closes #8244 

<details><summary>Was getting the following error when trying to export, and saving the permissions page WSODed.</summary>
<p>

```
Adding non-existent permissions to a role is not allowed. The incorrect permis  
  sions are "access metatag advanced__content_language", "access metatag advance  
  d__news_keywords".        
```

</p>
</details> 

I deleted the errant entries by hand in a config file.

<details><summary>Now I'm seeing some odd warning messages when I export.</summary>
<p>

```  [notice] The FAQ bundle (entity type: paragraph) was deleted. As a result, the field_support_article_faqs entity reference revisions field (entity_type: node, bundle: support_article) no longer specifies a specific target bundle. The field will now accept any bundle and may need to be adjusted.
 [notice] The Status update bundle (entity type: paragraph) was deleted. As a result, the field_alert_updates entity reference revisions field (entity_type: node, bundle: alert) no longer specifies a specific target bundle. The field will now accept any bundle and may need to be adjusted.
 [warning] Entity form display 'node.alert.default': Component 'moderation_state' was disabled because its settings depend on removed dependencies.
                                                                                   
 [OK] Configuration successfully exported. 
```

</p>
</details>


joewhitsitt:
@GaryRidgway if it isn't too much trouble, i would see if you could `ddev blt di --site=default` and see if you can get that NRD change exported to the default. I don't think it is specific to ITS.

From the looks of it, the alert moderation changes make me feel uneasy, but comparing it with PROD, it seems to be okay.

To me it would make sense that article is being cleaned up to align closer with default except where ITS is custom.

Those paragraph bundle errors on export is also a concern but something we have seen on other websites. I don't think we have an issue to look into it, but I know I have seen it on Admissions.

config/sites/its.uiowa.edu/config_split.patch.core.entity_form_display.node.article.default.yml
GaryRidgway, line 1: https://github.com/uiowa/uiowa/blob/31b53d2ba769781c1eec2b3be8954d6d8a0ad4db/config/default/core.entity_form_display.node.article.default.yml#L24

It looks like the base article has the editorial workflow now, and that is why it is being removed from ITS here. 

Additionally, the base doesn't have draft anymore, so there is no need to remove it from the ITS split, so that portion is also being removed.

GaryRidgway:
> @GaryRidgway if it isn't too much trouble, i would see if you could `ddev blt di --site=default` and see if you can get that NRD change exported to the default. I don't think it is specific to ITS.
> 
> From the looks of it, the alert moderation changes make me feel uneasy, but comparing it with PROD, it seems to be okay.
> 
> To me it would make sense that article is being cleaned up to align closer with default except where ITS is custom.
> 
> Those paragraph bundle errors on export is also a concern but something we have seen on other websites. I don't think we have an issue to look into it, but I know I have seen it on Admissions.

So looking at the Default Article config file it seems that the NRD changes that are listed in the article split are not occurring, except that when I export them I am seeing that another entry for `only_drafts' is being created after saving the article content type.

PR #8264, opened 2024-11-20
[Events] Increase timeout

### LINKED ISSUE #8255: Intermittent issue with events block not being able to pull events
This was reported on https://uira.org.uiowa.edu/news-and-events/past-events-and-programs. I was able to observe the block displaying an error message of `Unable to fetch events at this time.`, which @joewhitsitt tells me is what happens when the timeout introduced in #8216 is triggered.

cory-skeers:
Resolves #8255 

# How to test
(note, there is some caching on the events side of things, so try to "save" some pages of the events block to test with. eg, don't click through all the pages initially, and go only far enough to have it error. Subsequent hits to that page of the block's events will likely load quickly and without error regardless of the local site's cache. If you run into the case that you need to continue testing, try editing the block and changing a filter, re-saving, and trying again. Alternatively, see the splunk link in the bottom notes and find another possibly troublesome page to test with)

With `main` checked out
```
ddev blt ds --site=uira.org.uiowa.edu && ddev drush @orguira.local en dblog && ddev drush @orguira.local uli /news-and-events/past-events-and-programs
```
Notice the events block doesn't load, but instead displays "Unable to fetch events at this time." (If it does load, try paging until you hit one that doesn't load.)
```
ddev drush @orguira.local watchdog-show --count=10
```
Verify that it was the expected `Connection timed out when reaching...` error.
```
git checkout sitenow_events_load_increase_timeout && git pull && ddev drush @orguira.local cr
```
Reload the page and page through the events block, and see that you can hit all the pages without hitting the timeout error. 

## Post-deployment
Keep an eye on logs, see if the number of events block related errors decrease, or if there is more work needed.

https://splunk.its.uiowa.edu/en-US/app/search/search?earliest=%40w0&latest=now&q=search%20index%3Duiowa-acquia%20content.uiowa.edu%20%22Operation%20timed%20out%22&sid=1732047035.263840&display.page.search.mode=smart&dispatch.sample_ratio=1

### DIFF
--- a/docroot/modules/custom/sitenow_events/sitenow_events.module
+++ b/docroot/modules/custom/sitenow_events/sitenow_events.module
@@ -760,7 +760,7 @@ function sitenow_events_load(
     try {
       // Set a timeout (e.g., 10 seconds).
       $request = \Drupal::httpClient()->get($endpoint, [
-        'timeout' => 10.0,
+        'timeout' => 20.0,
       ]);
       $data = json_decode($request->getBody()->getContents(), TRUE);
 


joewhitsitt:
pulling to test

joewhitsitt (approved):
Nicely laid out instructions and next action. Thanks for putting this together. Hopefully we can limp along a bit more.🤞

PR #8271, opened 2024-11-21
[ ITS ] Change "Article" content type label to "News Article"

### LINKED ISSUE #8244: ITS - Rename "Article" content type to "News Article"

GaryRidgway:
replaces https://github.com/uiowa/uiowa/pull/8261
closes https://github.com/uiowa/uiowa/issues/8244

## To test
- Pull branch
- Sync its
- Check that the content type is named "News article" and that the help text changes as well


cory-skeers (approved):
Pulled and got a clean sync with updated name and description! 🚀 

I did realize after that the "Articles" list block is maybe not clear, but the link in the block (leads to a filtered content page) works correctly and helps clarify. So all that to say...I don't think we need to worry about that here.

PR #8282, opened 2024-11-22
[Bug] Fix replicate event subscriber for multiple list blocks

### LINKED ISSUE #8279: Replicate homepage on default site isn't working
# Problem
When reviewing the core update PR (https://github.com/uiowa/uiowa/pull/8273), Sean noticed that replicate on the default content homepage is not working. Other nodes it is working. Confirmed it was the same on `main`

# Proposed Solution

cory-skeers:
Resolves #8279 

# How to test

```
ddev blt ds --site=default && ddev drush @default.local uli /node/1/replicate
```
See that the page is successfully replicated

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/51/replicate
```
See that the page is successfully replicated
Try to replicate https://sandbox.uiowa.ddev.site/node/41/replicate

Alternatively

- find or create a page that has at least 2 list blocks within a section, either in the same region (column) or across multiple regions
- `git checkout main`
- Attempt to replicate the page, and receive a WSOD
- `git checkout replicate-hotfix`
- Attempt to replicate the page, and SUCCESS

Further testing for good measure? Try or find different combos of 1, 2, or more list blocks with or with/out other blocks in the section and see that it still doesn't error when trying to replicate, and that blocks are replicated in the proper order


joewhitsitt:
pulling to test

joewhitsitt:
It worked on the default/1, but then i moved over to sandbox and tried both the articles and people lists page and got a similar uuid error.

https://sandbox.uiowa.ddev.site/node/51/replicate
https://sandbox.uiowa.ddev.site/node/41/replicate

Thinking it was gremlins or the size of the page, I re-synced the site and tried the sandbox homepage and got the same error.

cory-skeers:
> It worked on the default/1, but then i moved over to sandbox and tried both the articles and people lists page and got a similar uuid error.
> 
> https://sandbox.uiowa.ddev.site/node/51/replicate https://sandbox.uiowa.ddev.site/node/41/replicate
> 
> Thinking it was gremlins or the size of the page, I re-synced the site and tried the sandbox homepage and got the same error.

Getting the same error on sandbox.

Stepping through with xdebug, at the point the error occurs everything seems to be correct, where we're inserting a new component with a unique id after an existing component which has an id which is attached to the section. 
image
image
image

Need to do a bit more digging to understand this.

cory-skeers:
`$section->insertAfterComponent()` uses `$section->getComponentsByRegion()` rather than `$section->getComponents()`, so in addition to first or last of the section, we may run into OOO issues when it's the first or last of a region. There's probably a better way of handling both by getting components by region earlier in the process, so need to (continue to) dig more.

cory-skeers:
I think the additional checks now cover the edge cases of first and last of a section, as well as the various region-based issues (only or last of a region while _not_ the first/only/last of a section).

joewhitsitt:
pulling to test again

joewhitsitt (approved):
I didn't go debugging myself so I am not sure I understand the situations you described but I tried it out in several different situations and from what I can tell it worked and didn't cause a WSOD. IMO, worth sending out for more real-world testing. I look forward to another one of Cory's in-depth explainer demos!

PR #8333, opened 2024-12-12
[IWP] Removed the sticky from resource content type

bspeare:
# How to test

```
ddev blt ds --site=iwp.uiowa.edu    && ddev drush @iwp.local uli /admin/structure/types/manage/resource
```

1. Confirm that sticky is not set by default on content type


joewhitsitt (approved):
Reviewed, not tested.

Issue #8336, opened 2024-12-12
Caching issues with Emergency endpoint


briand44:
With a focus on Signage, I believe similar to the CORS issue we ran into previously that is the new endpoint that is the problem. We should have probably compared this more to the old website, but I think we need to do more to adjust the cache related on this endpoint so that consumers hitting it must re-fetch.

`curl -I https://emergency.uiowa.edu/api/active`

I think we should test whether setting cache headers through an EventSubscriber or RouteSubscriber for this path has any effect and can make it through Varnish.

`$response->headers->set('Cache-Control', 'no-cache');`

I don't know if we could also inspect the varnish vcl file for uiowa702 to see if there were any special rules done for the v1 website.

ITS-ETS (osc-help) has offered to setup signage devices to test whether it is device/browser configuration with the signage install profile.

For reference:
https://github.com/uiowa/uiowa/pull/7698/files

briand44:
>I don't know if we could also inspect the varnish vcl file for uiowa702 to see if there were any special rules done for the v1 website.

I'm pretty sure there were not.

briand44:
Grooming conversation... It was reported that the Hawk Alert was showing up on Signage and had to reload uiowa.edu multiple times before showing. Also, Hawk Alert still showing on sites hours after closed. Suspected issue with the Emergency endpoint. We could try specific cache headers for endpoint path to get around Varnish. Old Emergency site has been deprovisioned but we think there is a backup if we needed to take a look. Something with the endpoint may be telling requests not to fetch. Based on previous CORS issue, we think endpoint is behind Varnish. Can we exclude this from Varnish?  Cache busting query parameters? Lets try and solve this issue as part of this slice.

joewhitsitt:
Poking at this a bit more but hesitant to assign at this time as I will be out Mon/Tue of next week and this seems to be of high importance that it keeps moving.

From what I can tell the Purge/Acquia combo UI has changed quite a bit from D7 and doesn't provide any kind of mechanism to allow/deny list paths.

Since we want to first avoid a custom vcl file and we know from `curl -I https://emergency.uiowa.edu/api/active` that we get varnish cache hits, from what I've read we want to do some combination of the following:

- An EventSubscriber scoped to the path that sets `$response->headers->set('Cache-Control', 'max-age=30, public');`
  - My understanding is that this is the clearest message to Varnish not to cache this path too long.
[...]

PR #8340, opened 2024-12-13
Emergency website cache control

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/8336

# How to test

Using [https://emergency.dev.drupal.uiowa.edu/](https://emergency.dev.drupal.uiowa.edu/api/active) and [https://signage-stage.uiowa.edu/city-iowa-city/downtown-interchange-stop-0001](https://signage-stage.uiowa.edu/city-iowa-city/downtown-interchange-stop-0001) create hawk alerts and observe the behavior. Ideally alerts and replacement alerts should show up in less than or a little over a minute based on the various timers between the two systems. After an alert is marked as complete, it should fall off the sign in an hour. I've made no change to how signage fetches/displays these.

Then you `curl -I` the different endpoints (DEV/PROD), you should see a big difference in varnish cache age.

The hope is that cutting Varnish times lower will prevent signs from fetching stale information.


joewhitsitt:
So far, compared to PROD, this is limiting the max-age of the endpoint. Because Drupal-Cache is still a HIT, I think I will remove the attempt at gutting of those. If the node_list does trigger, it should clear this under the 30 sec max-age.

https://emergency.dev.drupal.uiowa.edu/api/active

joewhitsitt:
DEV:
```
age: 0
via: varnish
x-cache: MISS
```

PROD:
```
age: 96571
via: varnish
x-cache: HIT
```

joewhitsitt:
signage-stage hawk alert endpoint config has been switched to emergency.dev.drupal.uiowa.edu

Testing:

I opened a sign in an anonymous chrome-based browser and didn't touch it.
Created an alert on emergency.dev.drupal.uiowa.edu and set it to published. Less than a minute later It popped up on the sign without any interaction with the window.
I then went into the alert and marked it complete. This removes it from the endpoint, but I am still seeing the alert on the sign. A new alert created will replace it (less than or around a minute), but completing both alerts will keep the latest one showing for an hour. I _think_ this is working as it was originally designed. I don't love it though.

cory-skeers (approved):
Reviewed and tested with similar results. Let's get it up to prod.

PR #8350, opened 2024-12-19
emergency_website_cache_control_part_deux

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/8336

# How to test

Based on https://github.com/uiowa/uiowa/issues/8336#issuecomment-2556936671, I'd like to send this change out for live testing. I have deployed this out to develop, but unless your browser is still caching the old cache-control header, this near impossible to test.

Otherwise, I have fairly high confidence that this will resolve itself on Signage and SiteNow v2/v3 in less than a month as cache/varnish clears on websites and the previous max-age expires. Adding the must-revalidate though will help us in the future if there are further endpoint changes and we want consumers to get the latest.

My understanding is that this change will enforce stricter revalidation rules to reduce the risk of stale data on the consumer end. If it has the max-age, it will respect that. If it sees must-revalidate it will look at the rest of the response to see if there is a change (e.g. etag, last-updated, new content in the json array).

Screenshot 2024-12-20 at 6 30 40 AM
Screenshot 2024-12-20 at 10 46 36 AM

### DIFF
--- a/docroot/modules/custom/uiowa_alerts/es6/alerts.es6
+++ b/docroot/modules/custom/uiowa_alerts/es6/alerts.es6
@@ -13,6 +13,7 @@
         const updateAlerts = () => {
           // Get the alerts feed and track IDs as "new" alerts.
           $.ajax({
+            cache: false,
             url: settings.uiowaAlerts.source,
             dataType: "json",
             success: (response) => {
--- a/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/EventSubscriber/CacheControlSubscriber.php
+++ b/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/EventSubscriber/CacheControlSubscriber.php
@@ -33,7 +33,7 @@ class CacheControlSubscriber implements EventSubscriberInterface {
     // Check the path and modify headers accordingly.
     if (str_starts_with($request->getPathInfo(), '/api/active')) {
       // Reduce cache, leaving some DoS protection.
-      $response->headers->set('Cache-Control', 'max-age=30, public');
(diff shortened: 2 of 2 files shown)


joewhitsitt:
For this test, I've setup https://education.dev.drupal.uiowa.edu/ (ajax cache false) and https://admissions.stage.drupal.uiowa.edu/ (main) to fetch from https://emergency.dev.drupal.uiowa.edu/. I have cleared site and varnish cache for each site and then loaded up admissions and education within anonymous windows in both Chrome and Brave. Once the hawk alert is created, the setInterval of 30000 should update the tabs with the hawk alert within about 30 seconds.

Again Brave displayed the alert on education.dev. Chrome did not. Both had a console log errors saying that `https://emergency.dev.drupal.uiowa.edu/api/active?_=1734630769080` resulted in the Bad Request 400 - "The following query parameters violate the JSON:API spec: '_'."

Ugh: https://github.com/jquery/jquery/issues/3682 though supposedly core is using v3.7.1

Interestingly, admissions.stage is showing the alert in both Chrome and Brave... 🤔

joewhitsitt:
Reverted this branch's move into develop: https://github.com/uiowa/uiowa/commit/76824c05f84df23651ec3ff13ecb33f193b29e38

cory-skeers (approved):
Reviewed, not tested. Good with getting it up and testing in prod.

joewhitsitt:
Ah shoot. Forgot to remove the ajax cache false. Standby for follow up. My bad

joewhitsitt:
https://github.com/uiowa/uiowa/pull/8353

Issue #8358, opened 2024-12-23
Commencement ceremonies are being removed after they are past


briand44:
![Image](https://github.com/user-attachments/assets/a842e67a-a9d7-4527-8ed7-575c4fb51d07)

Notice in the above screenshot from Friday, Graduate Doctoral Commencement Ceremony, Graduate Masters Commencement Ceremony and College of Nursing Commencement Ceremony are all showing.

View https://commencement.uiowa.edu/ceremonies and notice that none of them are being displayed.

Looking at `/admin/content` I am not seeing these ceremonies. It seems like they are being deleted from the site. Commencement folks want them to continue to display so that folks can view the video of the ceremony after it has past.

There was also a College of Education ceremony that took place on Thurs, Dec 19th at 4pm that is not displaying.

I disabled the am and pm importers Sunday night in hopes that would keep the ceremonies from Sunday displaying and avoid even more data loss.

**Screenshot from Saturday**

![Image](https://github.com/user-attachments/assets/a9877cad-6865-4905-9793-f13fc801a55d)

**Screenshot from Sunday**

![Image](https://github.com/user-attachments/assets/f4621f20-8e8f-4c8c-ae11-dd3222404bf2)

cory-skeers:
Ran the events-importer locally, and it did delete 3 events.
<img width="899" alt="image" src="https://github.com/user-attachments/assets/a2800a6a-618f-4d9a-a663-0169954fb570" />
Need to familiarize with the code a bit to understand what needs to be changed to keep old content around. It's based on some of the other importers like buildings where deletions are intentional, so we may need to do new code to add the option. After that, I think we could do an import of old events to restore them.

briand44:
Thanks for taking a look. Even if we imported the old events, there would be some data loss as they add additional data after they are imported. I think Commencement would be ok with that though.

pyrello:
Taking a look at this.

pyrello:
Discussed this with @cory-skeers a little bit and we both thought that a change like in #8359 was the quickest, least painful route to providing an exception to the deletion assumption built into importers currently. Getting that deployed will at least allow us to turn the importers for commencement back on.

My assumption is that events that have been deleted will not be reimported because they have dropped off the feed that we are using to import events. We'll need an additional slice of work to reimport (probably as a one off update hook) events that were deleted.

briand44:
[...]

PR #8359, opened 2024-12-23
[Commencement] Skip deleting imported events

pyrello:
Related to #8358 

# How to test
- Check that this resolves the issue with events being deleted on the commencement site.
- Check that this change does not prevent other importers from deleting imported content. Spot checking (vs checking every importer) is fine.


cory-skeers:
Tested locally on facilities with an extra building added, and it was deleted as expected.
image
Testing locally on commencement, on main 3 events are deleted and (after re-syncing) on this branch 0 are deleted.
image

cory-skeers (approved):
Reviewed and spot tested, and works well for both the default deletion and overridden skipDelete options.

Issue #8355, opened 2024-12-20
Remove land-ack from the footer of SiteNow websites


briand44:
We've been asked to remove the land acknowledgement from the footer of all sites as well as any additional places this may be (brand site, UIDS, sitenow docs, dispatch templates).

# Additional Info

This needs to be done prior to Jan 1, 2025.

We wil want to let Jeneane know once this is complete.

Jeneane is working on the comms outreach to the relevant unit. 
```[tasklist]
### Proposed Solution
- [x] Remove land acknowledgement from UIDS footer - https://github.com/uiowa/uids/pull/962
- [x] Remove land acknowledgement from footer in uids_base - https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/uids/footer--socket-menu.twig - may need a cache clear - https://github.com/uiowa/uiowa/pull/8361
- [x] verify regents theme, healthcare split, etc are no longer showing the link - this can be tested once rolled out. Should all be handled by the template listed above
- [x] remove from brand site (Jess and Sondra gave us the ok to make changes) - https://brand.uiowa.edu/applying-brand/web#brand-footer and any other places on brand.uiowa.edu
- [x] Communicate with Dispatch folks about removing this from any templates
- [x] We may need to send comms to app-dev community. We would need to wait for Jeneane to share comms with us
- [x] Send comms to web community (forward Jeneane's message)
- [x] reach out to CLAS marketing/comms team and see if they can handle their sites or let us know if they need assistance - https://sustainability.uiowa.edu/
- [x] Viewbooks - exists on Law Viewbook, Vue project template maybe
- [x] Reach out to catalog.registrar.uiowa.edu about LandAck in their vended solution.
```

mjoneill:
Also, might have impact on guidance we offer on brand site, components in UIDS, sitenow docs, dispatch templates, elsewhere?

mjoneill:
Will also want to check with the Dispatch folks re: templates there.

briand44:
Comment from Michael...

CPH and Libraries have the landack on at least their primary sites…and will need to push a change.  I was expecting them to be covered by the planned comms already + I can followup individually

mjoneill:
Jeneane sent to campus communicators.

I forwarded to:

- App Devs listserv
- UI Web Community listserv

I also forwarded directly to:
- Dan M and Patrick R at CPH
- Ken C, Jose, and Anne B at Libraries
- marketing account at Iowa Law

[...]

PR #8361, opened 2024-12-23
Removes land acknowledgment

pyrello:
Related to #8355 

# How to test


cory-skeers (approved):
Synced a site on main, cached a page with the link. Pulled this, saw the link removed on uncached pages. Ran the update hook and saw it removed on the previously cached page. All checks out.

PR #8362, opened 2024-12-23
[Facilities] Energy dashboard links

### LINKED ISSUE #8260: Facilities Building Energy Dashboard links
# Story

As a Facilities site owner, we would like to output a link to the Building Energy Dashboard on each of our building pages.

# Additional Info

There should only be one link per building.
Each link will be custom to the building it appears on.
Not all buildings will have a link.

Facilities plans to add the link to the source data so we can pull it in and output it.

> The API has been updated with the energy dashboard links. Including an update from our IT team:
>  
> The new energyLink field is added to individual building api’s. When a building is not present on the energy dashboard then the value of energyLink is null.
> Example of a building api that has energy link: https://bizhub.facilities.uiowa.edu/bizhub/ext/building?bldgnumber=0001
> Example of a building api with energy link as null: https://bizhub.facilities.uiowa.edu/bizhub/ext/building?bldgnumber=0190
>  
> As mentioned below, our initial thoughts on placement for this resource is in that section of building resource buttons that includes accessibility map, lactation rooms, hearing loop systems, and gender inclusive restrooms.
> 

Previous info on placement:
>We’d like to have it go in a space between the sections containing “Contact and “Services” and the section with the “accessibility map, lactation rooms, etc.” Alternatively, it could go in the second section where the named building, area, year built, etc are listed, but I’d prefer a slightly less prominent spot.

```"energyLink": "https://pi-vision.facilities.uiowa.edu/PIVision/#/Displays/593/Building-Energy-Dashboard?asset=\\\\ITSNT2942\\UI-Energy\\Main Campus\\Old Capitol",```

```"energyLink": null,```

### PR DESCRIPTION
**Pull request by cory-skeers**
Resolves #8260 

# How to test

If you haven't, add the following to `docroot/sites/facilities.uiowa.edu/settings/local.settings.php`:

```
$config['uiowa_facilities.apis']['bizhub'] = [
  'auth' => [
    'user' => '',
    'pass' => '',
  ],
];
$config['uiowa_facilities.apis']['buildui'] = [
  'auth' => [
    'user' => '',
    'pass' => '',
  ],
];
```
and update with credentials from the password manager.

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local fm-buildings
```
See that a bunch of buildings have been updated.
Look at some buildings, check out that energy dashboard links are included (in button row)
Confirm buttons auto-adjust at different screen sizes. 
Inspect Energy Dashboard link to see it has an aria-label attached. Can also confirm via VoiceOver.
Check a building without an Energy Dashboard link, it should only be showing four buttons instead of five.
Code review - specifically, is `if ($energy_dashboard != NULL)` the best way?

### CODE REVIEW

**Comment by cory-skeers** [comment]
Currently the import part of this is working as expected, but display needs to be updated, likely with client feedback. It's currently set up as a link field, and however it should be displayed the default link text can be set. Alternatively, a plain text field could be swapped in, and making it a link and rendering can be done with some field processing.

**Comment by e-marie-w** [comment]
Everything is working as expected. I am going to see if I can get the link to appear in their preferred location.

**Comment by e-marie-w** [comment]
Screenshot 2024-12-27 at 11 35 51 AM

I got the link to appear in the button row. For now I decided to keep both locations of the link, as to cover both discussed location options.

**Comment by joewhitsitt** [comment]
I think having both for comparison is good, but I feel like we should do something to mask/beautify the long absolute URL in the top example. We might also want to add information to the link informing the user that it requires logging in since I doubt the apache login prompt is very accessible by itself.

Sans label, icon instead:

`<a href="https://..." aria-label="Energy Dashboard (Login Required)">Energy Dashboard</a> <span class="login-required">(Login Required)</span>`

The login-required class could be used to style the text smaller, etc. Not sure we have an established pattern for that already.

**Comment by e-marie-w** [comment]
> We might also want to add information to the link informing the user that it requires logging in since I doubt the apache login prompt is very accessible by itself.
> 
> Sans label, icon instead:
> 
> `<a href="https://..." aria-label="Energy Dashboard (Login Required)">Energy Dashboard</a> <span class="login-required">(Login Required)</span>`
> 
> The login-required class could be used to style the text smaller, etc. Not sure we have an established pattern for that already.

I haven't been able to add the aria-label to the top example. I believe we need to switch the field type away from link in order to do that. I did update the title display for that link, though, and was able to add the aria-label to the bottom example. I figure we can wait and see which one they prefer before changing the field type.

**Comment by joewhitsitt** [comment]
>I figure we can wait and see which one they prefer

I agree but I don't think it will mean a different type. Does something like `$variables['items'][0]['content']['#attributes']['aria-label'] = 'Energy Dashboard (Login Required)';` work?

File: docroot/sites/facilities.uiowa.edu/modules/facilities_core/src/Plugin/Block/AccessibilityLinks.php [hunk 1]
```diff
@@ -60,6 +62,9 @@ public function build() {
       $url = 'https://uiadmin.maps.arcgis.com/apps/webappviewer/index.html?id=' . $link['id'] . '&query=Buildings,BuildingNumber,' . $building_number;
       $list_markup .= '<a class="bttn bttn--transparent bttn--tertiary bttn--small" href="' . $url . '">' . $link['label'] . '<i class="fa-solid fa-' . $link['icon'] . ' "></i></a>';
     }
+    if ($energy_dashboard != NULL) {
```
cory-skeers:
```suggestion
    if (!empty($energy_dashboard)) {
```
If we don't have a value set, the `->getString()` should return an empty string, and `empty()` and `== NULL` evaluate the same way in this case, but I think the `empty()` makes it more clear when reading that an empty string is expected and not a true null value.

cory-skeers:
Noting because it was mentioned in the email discussion of the addition:

There are a couple screen sizes where text breaks differently leading to slightly different button heights
image
but the difference comes from keeping the widths and spacing equal, and is the existing behavior prior to the button being added.
image
It's more likely to happen with 5 buttons than 4, but I think it's just a product of responsive design and dynamic content, and looks good at mobile and most desktop sizes.

I've reviewed and tested this and it works and looks great! My only question is if there could be any visual indication of the login requirement outside of the aria label? I think it'd be a nice-to-have, but if there's not a great way with it being a button, I don't think it's necessary.

Otherwise, I'd be happy to approve from a technical standpoint, but I opened the PR and so cannot :(

e-marie-w:
> My only question is if there could be any visual indication of the login requirement outside of the aria label? I think it'd be a nice-to-have, but if there's not a great way with it being a button, I don't think it's necessary.

I agree it'd be nice-to-have. We could add just (Login) to the button, like this:
Screenshot 2025-01-03 at 3 12 10 PM

It takes up roughly the same amount of space as Gender Inclusive Restrooms. Adding 'Login Required' made the text break even on my large monitor.

docroot/sites/facilities.uiowa.edu/modules/facilities_core/src/Plugin/Block/AccessibilityLinks.php
e-marie-w: Ahh that makes sense, thank you for the explanation! I've updated the code to use `empty()`.

joewhitsitt (approved):
Reviewed, not tested. Approving based on Cory's and Marie's work/review. As discussed in standup, lets leave the login text off for now.

Issue #5427, opened 2022-07-05
Config Split export not working as expected


bspeare:
# Problem

Pretty much since we adopted Config Split 2.0, there have been little issues that we have had to work around that usually involve deleting files that get exported that we don't want or manually editing files to add or remove sections that should or shouldn't be included in a config file. This issue is a place to collect examples so we can start to figure out how to address these issues. 

## Issues
- default config exports want to remove the deny list of blocks from the page.default config (limiting use of slider, timeline).
- Sites with paragraph types get a concerning warning message about fields/types being removed.
- Changes to filtered_html configuration strips out attribute values we have set (e.g. table role, iframe loading/sandbox)
- Issues were encountered as part of https://github.com/uiowa/uiowa/pull/5625
  - For topic_page, views had to be added to the complete section to avoid both a complete file and a patch file for the same configuration.
- If partially splitting an existing config in a way that introduces a new dependency to the config where the new dependency is completely split, a complete split will be generated in addition to the partial split.
- #8137 

<details>
<summary>

## Previous issues

</summary>

- ~~Admissions/Grad admissions - `tvi`-related config not getting added to taxonomy form display, need to be added manually.~~
- ~~Admissions/Grad admissions -  taxonomy field image is showing up as a complete split item~~
- Admissions - site exports sometimes contain key changes to user.role config (changes from `-` to integer). Discarding changes, re-saving permissions page usually provides a clean export.
- ~~IISC - Viewer role permission and actions related to adding and removing the viewer role get exported and need to be removed.~~

</details>

# Proposed solution
- Collect issues in the list above
- Attempt to reproduce each one outside of SiteNow, documenting the steps to reproduce
- Open one or more issues (grouping together similar issues) in the Config Split issue queue

pyrello:
Related: https://www.drupal.org/project/config_split/issues/3310931

fenal69:
> # Problem
> Pretty much since we adopted Config Split 2.0, there have been little issues that we have had to work around that usually involve deleting files that get exported that we don't want or manually editing files to add or remove sections that should or shouldn't be included in a config file. This issue is a place to collect examples so we can start to figure out how to address these issues.
> 
> ## Issues
> * default config exports want to remove the deny list of blocks from the page.default config (limiting use of slider, timeline).
> * ~Admissions/Grad admissions - `tvi`-related config not getting added to taxonomy form display, need to be added manually.~
[...]

PR #8395, opened 2025-01-07
[Registrar] Additional context for featured content title

### LINKED ISSUE #8371: Additional context for featured content reference field
# Story

As a Registrar site owner using the featured content block, I would like to be able to see the entity type of the content that I am referencing so that I do not have to look up the node id to ensure I am selecting the right entity to reference since there are multiple pieces of content with the same title.

# Additional info

Registrar asked for the ability to have additional context when using the featured content block as they are running into contacts that have the same name as other content in their site.

# Proposed solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8371. 

# How to test

```
ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli /node/1/layout
```

1. Review code - Since this is a generic content reference view that other sites can potentially use, is there a better name for the view besides "Content reference"?
2. Add featured content block to the layout.
3. Confirm that featured content items list their content type after each title and that featured content block works as expected. 
4. You should be able to add Page, Person, Article, and Contact content types. 

## ITS

```
ddev blt ds --site=its.uiowa.edu && ddev drush @its.local uli /
```
- Test adding a service to a featured content block `Microsoft Teams` is one example. 
- I have tested this, but you could also test for data loss on an existing featured content block referencing a service. 
- You should be able to add Page, Person, Article, and Service content types. 

## Facilities

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli /test-page     
```
- Test adding an artwork to a featured content block `Small World` is one example. 
- I have tested this, but you could also test for data loss on an existing featured content block referencing an artwork.
- You should be able to add Page, Person, Article, and Artwork content types.  

## Featured content overrides
- [x] its.uiowa.edu
- [x] facilities.uiowa.edu
- [x] contact feature

### Outstanding issue
Running the config export command on any sites with the content_reference view partially split will result in exporting the entire view in addition to the patch file. 

- `ddev drush @facilities.local config-split:export site `
- `ddev drush @its.local config-split:export site   `
- `ddev drush @registrar.local config-split:export site  `


bspeare:
Moving back to draft for clarity on https://github.com/uiowa/uiowa/issues/8371#issuecomment-2581067620.

bspeare:
I'm encountering the issue where the contact feature is creating the necessary views patch file but also exporting the entire view as well.  Any suggestions on getting a clean export?

joewhitsitt:
>Any suggestions on getting a clean export?

If there was a sure way, we wouldn't have so many "fix config" issues. I can try pulling down tomorrow if you want and see what sticks but I won't pretend to understand the how.

pyrello:
> Running the config export command on any sites with the content_reference view partially split will result in exporting the entire view in addition to the patch file.

I reviewed this and I believe that the reason this is happening is because the view config is adding a dependency on another config that is completely split. My understanding of the process from years ago when I worked on it is that any config depending on another config (e.g. the `content_reference` view has dependencies on the content types it references) that must be completely split (because it is defining something new e.g. the `contact` content type), then that config must also be completely split. Most commonly we run into this where a config that should be partially split depends on a module that is enabled via the config split. I opened this issue regarding that case: https://www.drupal.org/project/config_split/issues/3349737.

Unless we want to write a patch for config split, our best bet is to manually remove the completely split files. It appears as if the partial split files are still being generated and everything works on the import side without them.

Additionally, I'm wondering if we could potentially use a `.gitignore` file in the config split folders to prevent these files from accidentally getting added to the project.

pyrello:
I tried out adding a `.gitignore` file to exclude the `views.view.content_reference.yml` file. This successfully prevented from being added via git. I'm wondering if this is something we want to start doing to prevent these files from getting added? The potential downside is that at some point we might forget it is there and drive ourselves crazy trying to get it to add the item in git.

bspeare:
@pyrello I like the idea of the `.gitignore`.  As long as we document the issue we are trying to prevent I think it will help reduce confusion when we test all the splits during Drupal updates.

joewhitsitt:
I think the documentation of this will be key. .gitignore allows comments, so having them all point to the upstream issue or an issue in our queue that links to the upstream issue would be helpful and prevent later issues. i'd rather the confusion be with git than continuing to have to determine what is going on with exports.

i think a problem with confusion will remain though as we will continue to drive ourselves crazy in trying to "fix" the problem ourselves of two files for the same config before even trying to commit it.

james917:
Currently testing this

james917 (approved):
Looks good

pyrello:
I'm going to add `.gitignore` files for the site splits in question here.

pyrello:
I added additional `.gitignore` files and included comments referencing #5427 (where I added a bullet point describing the issue we encountered here). I think this is good to go now.

PR #8397, opened 2025-01-08
[Banner] Hotfix

bspeare:
# How to test


joewhitsitt (approved):
that works. went in and re-saved the autocomplete result as well to make sure it processes both

bspeare:
Thanks @joewhitsitt!

PR #8398, opened 2025-01-08
Card Block - separate card switch with a break to avoid issues

joewhitsitt:
Tracking down what is probably an unrelated issue, but this was unneeded noise.

An error, "Warning: Undefined array key "override" in layout_builder_custom_form_alter() (line 264 of /mnt/www/html/uiowa/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module)" seems to occur on the card block related to this line `if ($form['settings']['override']['items_per_page']) {` which is triggered based on the switch case carry through from the card case statement.

This carry through is also causing the help text for the card media position to point to the content-lists link instead of the card block specific one.

I don't think this is the first time we have caused issues because of how this switch statement is setup.

# How to test

Within a card block you should see the following:

- Card library is loaded (image/video conditional logic)
- The image media position help text should link to https://sitenow.uiowa.edu/documentation/layout-builder/card-block
- Card image is stacked by default.
- Media size should show if format is circle or position is _not_ stacked


joewhitsitt:
If we would rather keep this flow-through and instead be more defensive with our coding inside, feel free to pitch that.

joewhitsitt:
https://splunk.its.uiowa.edu/en-US/app/search/search?earliest=%40d&latest=now&q=search%20index%3Duiowa-acquia%20inline_block%253Auiowa_card%20AND%20international.uiowa.edu&display.page.search.mode=smart&dispatch.sample_ratio=1&display.events.timelineEarliestTime=1736348400&display.events.timelineLatestTime=1736352000&sid=1736376093.1308341

james917:
Reviewing this task now

james917 (approved):
Tested it in Admissions added block with Joe looks good.

PR #8425, opened 2025-01-17
[Classrooms] Building update

### LINKED ISSUE #8373: Classrooms buildings update
# Story

As a Classrooms website owner, we would like to update our building list as there are some buildings that should no longer be listed on our University Classrooms search page or our Programmed Classrooms search page.

# Additional Info

The Cedar Rapids Center is now an inactive building as the University lease has expired. It should not be listed in our building lists.

Also looks like maybe the Birchwood Learning Center may get removed. Not sure about others.

Classrooms needs to remove associated rooms for the Cedar Rapids Center before we are good to update the building lists.

It looks like the Cedar Rapids Center is not listed in the following endpoints so if I am understanding this correctly, it should get removed when we run our command.

https://api.maui.uiowa.edu/maui/api/pub/facilityBuildingRoom/list?roomCategory=PROGRAMMED_CLASSROOM
https://api.maui.uiowa.edu/maui/api/pub/facilityBuildingRoom/list?roomCategory=UNIVERSITY_CLASSROOM

> Right now, entities referencing would need to be deleted and we would need to re-run the building config stuff for an updated list. We have this issue to address some of these pain points. https://github.com/uiowa/uiowa/issues/6557. If I recall, they add/delete the classrooms they want to show and we just keep them updated with upstream data. So they should be able to delete what is no longer needed. - @joewhitsitt

cory-skeers:
Resolves #8373 

# How to test

```
ddev blt ds --site=classrooms.uiowa.edu && ddev drush @classrooms.local uli /admin/structure/building 
```

Clean sync, and check that Cedar Rapids Center is not there.

Can also check the building dropdown list when adding a room or using the search feature at https://classrooms.uiowa.ddev.site/spaces/find-university-classroom

Note: Health Sciences Academic Building is a new one to the buildings endpoint used, and so has been added in addition to the Cedar Rapids Center being removed.

### DIFF
--- a/config/sites/classrooms.uiowa.edu/classrooms_core.building.cbcr.yml
+++ b/config/sites/classrooms.uiowa.edu/classrooms_core.building.cbcr.yml
@@ -1,8 +1,0 @@
-uuid: 49d7d90b-b94f-4adf-99e0-0f2a4749fba5
-langcode: en
-status: 1
-dependencies: {  }
-id: cbcr
-label: 'Cedar Rapids Center'
-description: null
-number: '0384'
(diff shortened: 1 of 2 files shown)


joewhitsitt (approved):
works as expected

Issue #8363, opened 2024-12-26
Remove parameters from hawk alert calls


joewhitsitt:
Follow up from https://github.com/uiowa/uiowa/issues/8336

The change to the cache headers occurred on 12/18/2024 so cached responses with a 32 day max-age should expire by 1/19/2025 and fetch the new 30 second max-age header.

Remove the endpoint additions to v2/v3, signage, healthcare signage.

joewhitsitt:
Created `remove_hawk_alert_param` branch on signage and hc signage

joewhitsitt:
btw. picking this up to fix the 400 bad request errors we are now getting. we weren't previously...

joewhitsitt:
Slack deploy request: https://iowaweb.slack.com/archives/C8SM82ZFZ/p1737482464431729

joewhitsitt:
I thought this got closed?

PR #8426, opened 2025-01-21
Hawk Alert - remove param it is now returning a 400 bad request

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/8363

# How to test

Any site, replace *.

`ddev blt ds --site=*`
`ddev drush @*.local cr`
Inspect the Network tab of the browser. Reload. See an /api/active 200. Compare it to the same URL currently in PROD.


pyrello:
Pulling to test.

pyrello:
Am I understanding the test instructions correctly that I'm supposed to be seeing it access the API without the `?include=id` parameter? If so, its not working for me currently.

joewhitsitt:
You don't see a 200 for /api/active?
Screenshot 2025-01-21 at 2 47 54 PM

If it is still fetching with /api/active?include=id then it is a cached JS file

pyrello:
> If it is still fetching with /api/active?include=id then it is a cached JS file

Is there something I need to do besides clearing cache to get it to reload the JS file?

joewhitsitt:
oh gosh. not sure. we are getting into the complexities of drupal/browser caching.. when you inspect the js file in sources, does it have the change to the file?

from signage.uiowa.edu (aggregated js and all)
Screenshot 2025-01-21 at 3 05 46 PM

pyrello:
I had to run `blt frontend` (I probably could have gotten away with a more targeted `yarn workspace ... build` command), which updated the generated JS. Probably a good reminder that our es6 no longer needs to be compiled and can be run as is. We just need to change the process to reflect this.

pyrello (approved):
After regenerating frontend assets, worked as expected.

PR #8432, opened 2025-01-22
[inrc.law.uiowa.edu] Hide local events block if no results

### LINKED ISSUE #8406: Hide events block from display on person page if no items to display
# Story

As a SiteNow site owner using the local event feature, I would like to hide the events block from showing on person profile pages if there are no results so that it doesn't take up space on the page.

# Additional Info

LAWINRC reported that they have an events block displaying no results showing on all of their people pages and would like to hide it.

They are using the local events feature.

Image
```[tasklist]
### Tasks
- [x] Update Events heading to have gold underline
- [x] Hide if no results
- [x] Test other sites using local events feature
```

e-marie-w:
Closes #8406 

# How to test
- `ddev blt ds --site=inrc.law.uiowa.edu && ddev drush @lawinrc.local uli`
- check the Events block is not displaying on a person page if no results
- check it does display if results (may need to create a new event or add a presenter to an existing event)
- see that the Events heading is underlined now
- Repeat for the other sites using this feature split:
  - `ddev blt ds --site=iowasummerwritingfestival.uiowa.edu && ddev drush @iowasummerwritingfestival.local uli`
  - `ddev blt ds --site=afr.fo.uiowa.edu && ddev drush @foafr.local uli`- not currently adding presenters to events but should still test
  - `ddev blt ds --site=onuiowa.uiowa.edu && ddev drush @oniowa.local uli` - not currently adding presenters to events but should still test

Note: Admissions also has this feature enabled, however because they are not linking people and events currently, events will not display on a person's page at all. `ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli` if you want to check for yourself.


joewhitsitt:
Sorry, I might have misunderstood your Slack question. I think we want the header change for all sites using the event feature, but that Admissions' split of this (not accepting the change) continues.

https://github.com/uiowa/uiowa/pull/8432/commits/335a733f15249ff0eb2d81c8a1fd786ed54cf832

e-marie-w:
> Sorry, I might have misunderstood your Slack question. I think we want the header change for all sites using the event feature, but that Admissions' split of this (not accepting the change) continues.
> 
> [335a733](https://github.com/uiowa/uiowa/commit/335a733f15249ff0eb2d81c8a1fd786ed54cf832)

I am not sure this change is even needed now, it turns out they are both using `<h2>`, the Authored by just adds a class="h3" and my eyes might be playing tricks on me in regard to them looking different. (Since the change they look unequal in the opposite way than before.)

One headline is being set at the view level, and the other is current being set at the block level, on the person content type's display layout. Maybe I think I'm seeing a difference because I know that? 

I'm fine with making the change for the event feature, at this point I just don't think it's needed and we should remove my last commit.

bspeare (approved):
This looks good! I updated the authored by heading to render the same heading as the events block.

PR #8441, opened 2025-01-23
Establish MAUI AIS RFI Middleware Remote Post Handler

### LINKED ISSUE #8288: MAUI RFI Gateway for Undergraduate Forms
# Story

As a site owner with Undergraduate RFI webforms, we would like the ability to pass collected information off to MAUI through AIS middleware, so that information can automatically be collected for further use in marketing and recruitment efforts.

# Additional Info

This should be similar to https://github.com/uiowa/uiowa/issues/8213 in that there should be a separate handler not tied to the advanced webform feature.

This handler should post to the same endpoint and use the same credentials.

It should allow referral codes to be passed as well.

For Undergraduate forms we need to pass the following info in addition to the endpoint and credentials:

- Client Key - for Undergraduate forms this is `maui`
  - This should be passed through the handler at a form level
- referralCodes - string
- allow webmaster to choose additional form fields to be passed in similar to the default remote post handler. New fields should not default to being passed through.

# Proposed Solution

- [x] Create new handler using existing base class or abstract shared items to new base class
- [x] Pass client key through handler `maui`
- [ ]  allow referralCodes to be passed - discussed just having webform creator use hidden field for now.
- [x] Allow webmasters the ability to select which fields from the webform should be submitted with the remote post handler similar to the default remote post handler that webform provides.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/8288

# How to test

```
ddev blt ds --site education.uiowa.edu && 
ddev drush @education.local en dblog && 
ddev drush @education.local uli admin/structure/webform/manage/ces_interest_form/handlers
```

```
ddev drush @education.local config:get sitenow_webform_ais_rfi.settings middleware.endpoint_url
```

This should return a blank or "" value.

```
ddev drush @education.local watchdog:tail
```
- The last line of the above output will start outputting the logs in the terminal.
- Click the login link to go to the site. Enable the previously disabled `AIS RFI Prospector` handler. Now edit the handler.
- Confirm that you see the following warning:
```
The AIS RFI Middleware authentication credentials are missing. Please contact the SiteNow team for assistance.
```
- Go ahead and save the handler without a UUID value. Visit the form and submit it (or use the "Test" tab).
- Check the log output in the terminal and confirm you see the following message:
```
AIS RFI Middleware configuration is incomplete. No data was sent to the middleware.
```
- Add the following to the `docroot/sites/education.uiowa.edu/settings/local.settings.php` and replacing the `username` and `password` with the ones stored in esaccess.
```php
$config['sitenow_webform_ais_rfi.settings']['middleware']['auth']['user'] = 'username';
$config['sitenow_webform_ais_rfi.settings']['middleware']['auth']['pass'] = 'password';
```
- Edit the handler (don't change anything though) and confirm the credentials warning on the form is no longer showing.
- Submit the test form again and confirm that you see the following message in the logs:
```
AIS RFI Middleware: Interaction UUID is missing. No data was sent to the middleware..
```
- Edit the handler and add the interaction UUID (there is an email in the web team inbox that contains the UUID or you can ping @joewhitsitt for it). Submit the form again and confirm you see the following message:
```
AIS RFI Middleware: Success: Data received and enqueued for processing.
```

Now add the AIS RFI Maui handler. It should look similar to the other handler but instead of a UUID, it has a Referral Code field. Save without adding a Referral code. Submit the test form. Check the log output.

```
AIS RFI Middleware: refCode is missing. No data was sent to the middleware.
```

Go back in and add a string as a referral code value. AFAIK it could be anything right now. Before you test the form again, set a breakpoint at https://github.com/uiowa/uiowa/blob/1e177878645ec55b7d06b0c0b6891b4d5ec7946d/docroot/modules/custom/sitenow_webform_ais_rfi/src/Plugin/WebformHandler/AisRfiMiddlewareBaseWebformHandler.php#L135 and turn on xdebug.

Test the form and see that the breakpoint gets triggered more than once for each handler and that `$data` includes all of the default values, the handler included elements, the clientKey and either the refCode or the siteInteractionUuid based on which handler it stops on. You will also note that despite not having a config value for `endpoint_url` these are being sent to https://test.its.uiowa.edu because you are running from a local environment.


cory-skeers:
Pulling down to test...may have some questions as I go through.

joewhitsitt:
Sounds good. The short of it is that what was working was moved around to make room for a small new thing built the same way.

docroot/modules/custom/sitenow_webform_ais_rfi/src/Plugin/WebformHandler/AisRfiMiddlewareMauiRemotePostWebform.php
```diff
@@ -0,0 +1,85 @@
...
+    $ref_code = $this->configuration['ref_code'];
+    if (!$ref_code) {
+      // Log that the refCode is missing.
+      $this->getLogger()->error('AIS RFI Middleware: refCode is missing. No data was sent to the middleware.');
+      return [];
+    }
+    $data['refCode'] = $ref_code;
+
+    return $data;
+  }
```
cory-skeers, line 76: I keep going back and forth on whether we should move the `ref_code` check to happen prior to fetching data with the `parent::getRequestData()`. In this case right now, if we don't have a ref code (or prospector id in the other class) there's no reason to spend time getting the data and no other processes/logging that happens there.

Buuuut, I think I'd still lean toward keeping as it is here and calling the parent first in case the base class ever gets updated with something we wouldn't want to miss. And it's not like it's that time-consuming or costly.

cory-skeers (approved):
Reviewed and tested, and the changes check out 👍

docroot/modules/custom/sitenow_webform_ais_rfi/src/Plugin/WebformHandler/AisRfiMiddlewareMauiRemotePostWebform.php
joewhitsitt: Good points. So far there isn't any data that needs to come first, but I've heard mention that the refCode could possibly be altered based on form values or URL parameters in the future.

Ultimately the siteInteractionUuid will always be static so that one could flip.

My guess is that we are still learning what this will actually be, but are being tasked to show it as functional first. Further optimizations like could happen once we know for sure what we are supposed to be passing.

joewhitsitt: There are still a few travis builds before this will get merged. Do you want to flip either or both for now @cory-skeers ?

cory-skeers: I think given the uncertainty there, let's leave it as is.

PR #8449, opened 2025-01-27
Grad - Scholar updates

### LINKED ISSUE #8414: Scholars block component
# Story

As a Grad college site owner, I would like to have a scholars block where we can filter the block by year so we can add the current years scholars to one page and add the previous years scholars to a separate page while having control over additional page content and url paths.

# Additional Info

- Want to only show current scholars and past scholars (previous 1 year).
- Year dropdown can be reduced - current year and previous two years
- new year of scholars gets displayed starting in June of that year. So 2025 scholars will get displayed in June 2025.
- scholars profile pages url alias pattern should use `gsb` in place of `dei`.

```[tasklist]
### Tasks
- [x] Create scholars block component with backend year filter that displays scholar teasers like existing page https://grad.uiowa.edu/dei/srop/past-scholars/2024
- [x] Remove past scholars view page
- [x] limit number of years on content type and block component to current minus 2 (possibly switch to taxonomy if needed).
- [x] update scholars url alias pattern to use `gsb` in place of `dei`
- [x] get feedback from Grad before deploying to prod
```

pyrello:
Resolves #8414 

# How to test

```
ddev blt ds --site grad.uiowa.edu && ddev drush @grad.local uli dei/srop/past-scholars
```
- Confirm views page was removed.
- https://grad.uiowa.ddev.site/node/add/scholar
- Confirm the year drop-down shows `current - 3` years.
- Save and confirm that automatically generated URL includes `gsb` instead of `dei`.
- Add a page, add a block, More..., add the SROP Scholars block
- Confirm the year drop-down shows `current - 3` years. Select 2024 or 2023.
- Confirm the items per page has a default value.
- Save block and save the page. Confirm the scholars listed are from the year you selected.


pyrello:
Going to wait to put this up for review until it is demoed for GC.

pyrello:
> Going to wait to put this up for review until it is demoed for GC.

Demoed. Making these changes:
- Updating the path for scholars to remove the `past-`.
- Adding a default items per page.
- Changing the number of years includes to current - 3.

cory-skeers (approved):
Reviewed and tested, and looks good!

For the change to the URL, was there talk of doing a bulk save to give existing scholar content the updated path? It's something they could do as long as they're aware the change will only impact new content (or old content after it has been saved).

briand44:
I believe I had a conversation with them in my original meeting and told them they may need to re-save each scholar to get the updated url. It did not come up in todays meeting. I am pretty confident they were fine with manually saving. Let's not worry about it as part of this work. If it comes to it, I'll resave them myself. There is only 31 total right now and only 12 current. I will make sure to mention this when we communicate when the deployment happens.

PR #8459, opened 2025-01-29
Adjust commencement config to ignore default_session

### LINKED ISSUE #8431: Default session for Ceremonies page not sticking
> Our ceremonies page doesn't seem to be sticking to the default session I've selected in the taxonomy. It keeps going back to fall '24.
> 
> I made sure to flip last night to spring '25 before logging off and it resorted back to fall '24 this morning.

https://commencement.uiowa.edu/ceremonies

joewhitsitt:
Resolves #8431 

# How to test

- Don't check out the feature branch at first. This will validate the behavior we are seeing.
- Check https://commencement.uiowa.edu/taxonomy/term/216/edit and confirm that it (Spring 2025) is set as the default session.
```
ddev blt ds --site commencement.uiowa.edu && ddev drush @commencement.local uli taxonomy/term/216/edit
```
- After syncing, Spring 2025 will no longer be set as the default session. It will be back to Fall 2024.
- Check out `commencement_ignore_config_3`.
```
ddev blt ds --site commencement.uiowa.edu && ddev drush @commencement.local uli taxonomy/term/216/edit
```
- Confirm that Spring 2025 is still the default session.


pyrello:
This did not work for me:

Screenshot 2025-01-29 at 2 43 11 PM

pyrello:
I think we need someone else to test this. I rejected this approach because it wasn't working for me. But it sounds like @joewhitsitt had it work for him, so maybe I'm just experiencing local gremlins.

joewhitsitt:
Depending on when you did, I _just_ changed the value again in PROD after Alan's deployment.

cory-skeers:
I tried this out and the default value was maintained with this branch.

pyrello:
> Depending on when you did, I _just_ changed the value again in PROD after Alan's deployment.

It's hard to say, I just tested again and I just checked the prod version to confirm its what is expected before running the command. I'm still seeing the same result that I posted above.

pyrello:
> I tried this out and the default value was maintained with this branch.

@cory-skeers is the tie-breaker, so I guess I have gremlins. I'm good with trying this out. If for some reason it doesn't work, its no worse than the current situation.

cory-skeers (approved):
Mixed results on whether this worked, but approving so we can try it out live. If it doesn't, no harm done.

joewhitsitt:
@briand44 will probably need to have one of us confirm post deployment before announcing to Josh that we fixed it. 🤞

Issue #8435, opened 2025-01-23
Test new event API endpoints


briand44:
# Story

As a web team, we would like to start testing the new event API endpoints so we can provide feedback if any changes are necessary in order for us to integrate with the new event system prior to March 1st.

# Additional Info

AIS has said they will have the new endpoints available around 1/27th.

These endpoints should be nearly identical to our current endpoints except for the url and possibly some additional data.

We need to make sure the endpoints contain everything we need to handle all of our event integrations (SiteNow v2/v3, SiteNow v1, Signage, etc.)? https://github.com/uiowa/uiowa/issues/8313 may contain some useful information about where we are currently integrating with content.uiowa.edu.

# Proposed Solution

Using the information from https://github.com/uiowa/uiowa/issues/8313, attempt to wire up to replace existing endpoint variables (e.g. https://{new endpoint} replaces https://content.uiowa.edu) and see if our existing functionality still works as expected, taking note of what breaks, what is not represented so that we can pass that back up to AIS for refinement.

Events block should pull in the filter dropdowns. We should be able to get past and upcoming. Events with recurring dates should be represented as individual instances. Images should still pull over. Event descriptions should handle special characters.

Though D7 is still involved, our main focus should be the functionality used in SiteNow v2/v3 and the sites that extend off it (e.g. commencement) which is our most current implementation.

Check to see if one of our endpoints/display ids is not represented (e.g. event_instances_api with the events (upcoming) and past args) and if not, make it known that this is our greatest need for the switch.

If we are given access to create/edit test events, we should see that our integrations still properly fetch that new or updated information.

briand44:
Grooming conversation... We may need to be clear that we need the event_instance endpoint if that is not provided along with others. Looking at new endpoints and comparing to existing. Making sure event instances array will work with how we currently use it.  Check date formats, etc.

Blocked until we get new endpoints, especially event_instance endpoint.

joewhitsitt:
Just met with Beth to clear some stuff up about event_instances_api, though there might be more questions coming. They said to hold on testing that, but the other endpoints should be ready to test (now includes paging).

They are going to update the items per page default to 10.

The event ids are going to change and the drupal id to events.uiowa.edu id routing will discontinue. We will need to update our 

`https://events.uiowa.edu/event/drupal/155421 => https://events.uiowa.edu/90113` links to just `https://events.uiowa.edu/90113` but that ID will be included now since it is the same system and not drupal/java.

[...]

PR #8461, opened 2025-01-29
Compatibility tweaks for new UI Events API

### LINKED ISSUE #8460: Update SiteNow v2/v3 event integration to work with new endpoints
# Story

As a web team that has campus events integrations, we would like to update our SiteNow v2/v3 integrations to work with the new endpoints so that SiteNow v2/v3 integrations continue to work when the new events system launches.

# Additional info

Discussed scoping this to include all v2/v3 integrations (both sitenow_events and uiowa_events according to https://github.com/uiowa/uiowa/issues/8313).

Sounds like we may need to make some changes in addition to the change in endpoint.

- nesting has changed so we will need to look a level deeper under `event` - https://github.com/uiowa/uiowa/issues/8435#issuecomment-2620252758
- Changes to id's. Beth checking with the AIS team about the ID's. - https://github.com/uiowa/uiowa/issues/8435#issuecomment-2622450002

We'll want to create and test the PR but hold to merge to coordinate with AIS when they launch the new system. We may need to change the endpoint from test to prod prior to merging as well.

As part of testing, Joe has some of this locally and can create a PR.

Grooming conversation - https://github.com/uiowa/uiowa/issues/8445#issuecomment-2622632576

joewhitsitt:
Relates to #8435 
Resolves #8460 

# How to test

## Commencement Event IDs replacement and importer

- `ddev blt ds --site=commencement.uiowa.edu`
- No errors on import, update hooks.
- `ddev drush @commencement.local import-events`
- Ideally, no new events should be created and only minor updates if any.
- Login and edit one of the recently update event nodes. See the event ID in the imported details section. Take note of that ID. Now go to the revisions tab of that node. See that the revision message includes the ID you just noted as well as the original event ID.
- Bring up the old and new ID in their respective endpoints: api/v1/node/{old/new id}.json
- The event data should match pretty much to a T but the new endpoint has an additional layer of nesting.

## Events Blocks (v3)

- No errors rendering or configuring Events blocks.
- Events blocks still pull in events and though the events might not match, they should pull in similar events and in the expected order.
- https://sandbox.prod.drupal.uiowa.edu/events-api-testing (hancher events)
- Try combinations like past (change sort), upcoming (change sort), date range (change sort).

## Events Paragraphs (v2)

- `ddev blt ds --site=hr.uiowa.edu`
- No errors on import, update hooks.
- Events blocks still pull in events and though the events might not match, they should pull in similar events and in the expected order.

## Single Event Page

- Go to https://commencement.uiowa.ddev.site/event/31084/0 and see that the event renders and is the same data as https://sandbox.prod.drupal.uiowa.edu/event/164331/0

# Deployment notes

Captured on #8605


joewhitsitt:
In putting https://github.com/uiowa/uiowa/pull/8461/commits/e59ab2054f06895f7fffd5d2ccb22c37ebe293f7 together, I am think it is hitting an already modified revision or something repeatedly. at that point the new value is in place and it gets compared to the old ids. if it is matched then it gets updated to a different new id...

I wonder if our helper function is up-to-date with how Drupal core handles things like tempstore. I will try to debug a bit more to track down which instances are called again and in what context.

left the files blank for now because I don't know how they will be structured and I've been told the IDs will be changing again once they sort out an issue with departments.

joewhitsitt:
Leaving as draft as this is not good to merge, but hoping for technical review on the approach.

joewhitsitt:
Initial testing of the latest build is looking good! Might need to programmatically clear any cached requests based on the old (events blocks). After a `cr` it seemed to pull things in.

>Commencement event content sync completed. 0 events were created, 76 updated, 0 deleted, 0 skipped. That is neat.

This means that the id update hook matched, the query filtered correctly and didn't create a bunch of new events.

Looking at what was updated, it might just be a recent content change to these that PROD hasn't imported yet. Will check tomorrow to understand it better.

Screenshot 2025-02-19 at 5 17 42 PM

joewhitsitt:
Experiencing some funkiness. I feel like content hub is changing the taxonomies ahead of re-launch and that events-test doesn't have the updates yet.

https://github.com/uiowa/uiowa/issues/8460#issuecomment-2689228382

joewhitsitt:
Deploying this out to DEV for testing on commencement, education, sandbox

joewhitsitt:
Heads up @e-marie-w, others. I am going to update the commencement IDs. They changed while I was out on vacation.

GaryRidgway:
just to ensure transparency, I am also testing this

e-marie-w:
For media, there's an original image field and then different sizes - thumbnail, medium, large, etc. Noticing that at the new endpoint the images are identical when clicked as there's no suffix of sizing info added to the URL like at the old endpoint. Not sure it's something we control but making it visible.

Additionally, at this point is it possible for the new endpoint to have data not at the old point? Found an event (old ID 146461, new ID 29186) that does not have address data at the old endpoint but does at the new.

e-marie-w:
For the Events Blocks (V3) we are adding 12:00 AM the next day as the end date for events with no set end date. It is only happening on the page with the events listblock, the individual page for a given event is fine.

GaryRidgway:
I looked at all the examples and found them satisfactory, but I'm glad Marie and I both tested because it seems that she found some things I didn't!

joewhitsitt:
>For media, there's an original image field and then different sizes - thumbnail, medium, large, etc. Noticing that at the new endpoint the images are identical when clicked as there's no suffix of sizing info added to the URL like at the old endpoint. Not sure it's something we control but making it visible.

Good eye. They are only providing the "original image" now (duplicating it to fill the old sizes), but have added the width and height information for lazy loading. https://github.com/uiowa/uiowa/issues/8435#issuecomment-2622871372

>Additionally, at this point is it possible for the new endpoint to have data not at the old point? Found an event (old ID 146461, new ID 29186) that does not have address data at the old endpoint but does at the new.

Whoa. That looks like bad migration data. I can point that out to Beth.

>For the Events Blocks (V3) we are adding 12:00 AM the next day as the end date for events with no set end date. It is only happening on the page with the events listblock, the individual page for a given event is fine.

I can take a closer look at that.

joewhitsitt:
> > Additionally, at this point is it possible for the new endpoint to have data not at the old point? Found an event (old ID 146461, new ID 29186) that does not have address data at the old endpoint but does at the new.
> 
> Whoa. That looks like bad migration data. I can point that out to Beth.

Beth found the issue and is testing a fix for this but this won't be available for us to test ahead of launch. "Thanks for pointing it out"

joewhitsitt:
> > For the Events Blocks (V3) we are adding 12:00 AM the next day as the end date for events with no set end date. It is only happening on the page with the events listblock, the individual page for a given event is fine.
> 
> I can take a closer look at that.

This is exposing a difference with how event instances are provided by the API. I am going to see how we can tweak `sitenow_events_format_instance` so that it will work when end is empty and last_date doesn't include a time

e-marie-w:
> > > For the Events Blocks (V3) we are adding 12:00 AM the next day as the end date for events with no set end date. It is only happening on the page with the events listblock, the individual page for a given event is fine.
> > 
> > 
> > I can take a closer look at that.
> 
> This is exposing a difference with how event instances are provided by the API. I am going to see how we can tweak `sitenow_events_format_instance` so that it will work when end is empty and last_date doesn't include a time

Thanks for looking into this and finding that! I was just about to comment that for an event with no end date in a series, the end date becomes 12 am of the last instance instead.
Screenshot 2025-03-28 at 2 04 48 PM

joewhitsitt:
@e-marie-w just pushed a fix. will you be willing to check those instances again?

e-marie-w:
> @e-marie-w just pushed a fix. will you be willing to check those instances again?

The changes worked, I'm no longer seeing any end dates!

joewhitsitt:
I am going switch the endpoints back to prepare for merge. Testing can continue, but you shouldn't pull anymore

GaryRidgway:
> > @e-marie-w just pushed a fix. will you be willing to check those instances again?
> 
> The changes worked, I'm no longer seeing any end dates!

With this, I am comfortable saying I approve!

GaryRidgway (approved):

Issue #8430, opened 2025-01-22
Map not rendering on Commencement


briand44:
>It seems like the map widget on the venues has dropped off on Chrome and Firefox. I usually work in Edge and it's working fine there so I hadn't noticed it:

https://commencement.uiowa.edu/venue/carver-hawkeye-arena#map

briand44:
Grooming conversation... Check for any console errors as a first step.

pyrello:
Console error:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'get')
    at loadMap
```
Line of code in question (aggregated):
```
map_container.data('leaflet', new Drupal.Leaflet(L.DomUtil.get(mapid),mapid,data.map));
```

bspeare:
I looked at this a bit on Friday:

- Disabling lazy loading made the map load for authenticated users.
- The map does not load for anonymous users. 
- The issue is similar to this one https://www.drupal.org/project/leaflet/issues/3172474. 
- I tried updating Leaflet, Geofield, and Drupal versions and had the same results but different console errors.
- Tried to recreate the issue on Simplytest.me and I could not recreate.

james917:
I am reviewing this now.

pyrello:
> Console error:
> 
> ```
> Uncaught TypeError: Cannot read properties of undefined (reading 'get')
>     at loadMap
> ```
> 
> Line of code in question (aggregated):
> 
> ```
> map_container.data('leaflet', new Drupal.Leaflet(L.DomUtil.get(mapid),mapid,data.map));
> ```

To make it obvious in case it isn't, the above error is saying that `L.DomUtil` is undefined. This line of code is from our custom implementation, so it is something we wrote.

pyrello:
I deleted my last comment because it was inaccurate.

james917:
After reviewing some of the comments Ben had left I noticed a couple other things in the console I am also getting the same errors but I noticed accordion js errors were showing up also 
`accordion.js?sqrn5t:121 Uncaught SyntaxError: Unexpected token 'export' (at accordion.js?sqrn5t:121:1)Understand this errorAI
sitenow.behaviors.js?sqrn5t:10 This is a Sitenow v3 site. For more information, please visit https://sitenow.uiowa.edu.
drupal.js?v=10.3.9:64 Uncaught TypeError: Cannot read properties of undefined (reading 'get')
[...]

PR #8467, opened 2025-01-30
[Bug] Temporary workaround to get maps loading on commencement

bspeare:
Temporary fix related to https://github.com/uiowa/uiowa/issues/8430. 

Possible related issues:
- https://www.drupal.org/project/drupal/issues/3294720
- https://www.drupal.org/project/leaflet/issues/3501647

# How to test

```
ddev blt ds --site=commencement.uiowa.edu && ddev drush @commencement.local uli /venue/carver-hawkeye-arena
```
1. Confirm you can see the map logged in and logged out in many browsers. 
2. https://commencement.dev.drupal.uiowa.edu/venue/carver-hawkeye-arena


joewhitsitt:
If this is temporary, can we pepper it with some more comments saying as such?

cory-skeers (approved):
Tested this out and the workaround is working well for me. 

As I understand it, there's a timing issue with how/when things get loaded that is difficult to track down and address, and this more or less forces leaflet to attach. Especially with the added comments tying everything to the related issue and the possibility of it to be fixed with a Core update, I think this is a good temporary way to get it working.

bspeare:
Thanks @cory-skeers!

PR #8482, opened 2025-02-04
Update google_tag

### LINKED ISSUE #7538: Upgrade google_tag
Need to be careful with this update:

>stopped all data going into Google Analytics. Starting with release 2.0.3 of that module, there is a feature to opt-in to Google's consent mode, which is enabled by default.
https://www.drupal.org/project/google_tag/issues/3180185

>With that feature enabled, we get zero data in GA, presumably because we are not using a consent management platform (CMP) or otherwise sending any consent "signals" to Google.

>There is also a bug that makes the button not show that it's disabled. Here's a link to the patch to fix that little issue: https://www.drupal.org/project/google_tag/issues/3424623

^^ is from 90+ days ago and is no longer in the DrupalEDU Slack.

# Proposed Solution

- [ ] Developer choice on trying to upgrade to 2.x or just to 8.x-1.8
- [ ] Need to setup a test example where both google_analytics and google_tag are working at the same time as expected based on the concerns above.
- [ ] Decide if we need to hide new options/configuration from editors.

cory-skeers:
Resolves #7538 

# How to test
Successful build

```
git checkout update-google_tag &&
ddev composer install
```
Masquerade as a `webmaster` and check you have access to the GTM settings.

Sync a site and check that the GTM settings are retained, with the addition of "hostname" (under the Advanced tab of GTM settings), and that hostname is set to `www.googletagmanager.com`

Additionally, changes between versions can be reviewed here:
https://git.drupalcode.org/project/google_tag/-/compare/8.x-1.6...8.x-1.8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,11 +47,7 @@
     },
     "config": {
         "audit": {
-            "abandoned": "ignore",
-            "ignore": {
-                "SA-CONTRIB-2025-011": "Temporarily ignore.",
-                "SA-CONTRIB-2025-012": "Temporarily ignore."
-            }
+            "abandoned": "ignore"
         },
         "sort-packages": true,
         "platform": {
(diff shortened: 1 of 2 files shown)


cory-skeers:
This change, as opposed to updating to 2.x, seems rather low-impact, and I think we can move forward with it. Created a follow-up issue #8485 for later followup, but some testing surrounding settings and another set of eyes on the 1.6 to 1.8 change updates is likely good to get this into prod.

james917 (approved):
Looks good Approved

cory-skeers:
Thanks @james917!

Issue #8480, opened 2025-02-03
Investigate service interruption notifications


briand44:
# Story

As a web team, we would like to understand why we have recently been seeing an increase in service interruption notifications from Acquia so that we can avoid any future downtime.

# Additional Info

We have been getting these notifications daily for the last week, sometimes multiple per day.

The majority have been on uiowa01 and uiowa02 but have also seen them on uiowa03 and uiowa04.

https://acquia.my.site.com/s/case/500Pb00000cYk0vIAC/service-interruption-detected-on-your-production-environment
https://acquia.my.site.com/s/case/500Pb00000cVRKiIAO/service-interruption-detected-on-your-production-environment
https://acquia.my.site.com/s/case/500Pb00000cORKYIA4/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000cNf9TIAS/service-interruption-detected-on-your-production-environment
https://acquia.my.site.com/s/case/500Pb00000cIuM1IAK/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000cCHgvIAG/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000bwXksIAE/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000bmuFXIAY/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000beOo1IAE/notification-brief-service-interruption-detected
https://acquia.my.site.com/s/case/500Pb00000beIibIAE/notification-brief-service-interruption-detected

> - Your application is uncached or isn’t cached sufficiently to withstand spikes in traffic. See performance improvements with Varnish cache (https://docs.acquia.com/cloud-platform/performance/varnish) and download your access.log (https://docs.acquia.com/cloud-platform/monitor/logs/apache-access) files to find patterns of uncached traffic.
> - Cache was recently cleared (https://docs.acquia.com/cloud-platform/performance/varnish/purging), such as after a code deployment, which caused a temporary spike in uncached traffic.
> - Cron jobs are running too frequently. Learn how to adjust the frequency or disable cron jobs (https://docs.acquia.com/cloud-platform/manage/cron).
> - Slow queries caused the database to become overwhelmed (https://docs.acquia.com/cloud-platform/monitor/slow-query).

Are these just related to traffic spikes? Is it legit traffic?

# Proposed solution

joewhitsitt:
[...]

PR #8487, opened 2025-02-05
Ban 2 more IP addresses suspected of malicious intent

pyrello:
Related to #8480 

# How to test

- Code review.


joewhitsitt (approved):
@pyrello and I talked about the existing instance and it was largely a guess and check push that we then saw the related traffic go down. Assuming this is the correct format, I am find with pushing this out to see what happens.

PR #8527, opened 2025-02-18
[Accordion] Accordion hashes in url 

### LINKED ISSUE #8454: Adjust accordion js to work with uids_base
# Problem
During the recent accordion testing, there was a console error related to the accordion JS.

```
Uncaught SyntaxError: Unexpected token 'export' (at accordion.js?sqiel8:121:1)
```

https://github.com/uiowa/uiowa/pull/8405#issuecomment-2610189111

# Proposed solution
- [x] Implement a similar solution as we did for [click-a11y-init.js](https://github.com/uiowa/uiowa/pull/6450)
- [x] Confirm that this resolves the issue with the hash fragment not getting populated or triggering the browser to scroll to the linked element.
- [ ] Follow-up with Ben L when this is fixed.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8454. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli  /accordion-1
```
1. Confirm that hash changes in the browser each time you click on an accordion item
2. Confirm pasting that hash into a browser window opens the accordion item directly
3. Test with single and multi select options
4. Test with custom IDs and generated IDs
5. Confirm the console error is gone

### DIFF
--- a/docroot/themes/custom/uids_base/assets/js/accordion-init.js
+++ b/docroot/themes/custom/uids_base/assets/js/accordion-init.js
@@ -1,0 +1,16 @@
+import { applyAccordion } from '../../uids/assets/js/accordion.js';
+
+(function ($, Drupal) {
+  "use strict";
+
+  Drupal.behaviors.uidsAccordion = {
+    attach: function (context, settings) {
+      applyAccordion('.accordion:not([data-accordion-processed])', context);
+
+      const accordions = context.querySelectorAll('.accordion:not([data-accordion-processed])');
+      accordions.forEach(accordion => {
+        accordion.setAttribute('data-accordion-processed', 'true');
+      });
+    }
+  };
+})(jQuery, Drupal);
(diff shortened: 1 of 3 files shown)


joewhitsitt:
Not sure if I will be able to pull down and test right away. I wonder if that changes anything with the aria-owns stuff because I think it was expecting the id from aria owns to be on the element with the role tab. Not 100% there though. That might have been the original reasoning why it ended up there, though it could have just been a mixup on my part.

Can you ease my concerns here?

bspeare:
@joewhitsitt the hash does not work when the id is set on the `<summary>`, and the UIDS example sets the id on `<details>` which is why the hash works in UIDS.  Do we need to redo the work in UIDS so that aria owns is on the element with the role tab?

joewhitsitt:
I think it is kind of a grey area because we are using all of this out of their normal usage.. aria-owns is not meant for tablist but because we do have it there, I think it expects the ids on the tab element/labelling element that is then read. 

Screenshot 2025-02-19 at 9 30 30 AM

I noticed here, the reason the anchor works for UIDS is because there is a different id than what aria-owns uses. It would be nice reduce how many ids we need to account for.

bspeare:
Closing for https://github.com/uiowa/uiowa/pull/8530.

PR #8530, opened 2025-02-19
Accordion JS fixes

### LINKED ISSUE #8454: Adjust accordion js to work with uids_base

pyrello:
Resolves #8454 

## How to test
- Set up your local.blt.yml config to include the list of sites to sync below.
```
ddev blt frontend && ddev blt dsa
```

### Test that the regression is fixed.
- Confirm no console errors
- Confirm that the anchor link fragment gets set in the URL (e.g. `#anchor`) when toggling an accordion item open and that it matches the ID of the `<summary>` element for that accordion item. 
- For collection items (on the sandbox site below), ensure that the `<summary>` ID matches the Unique ID field set for each collection item.
- Confirm that navigating to an anchor link for an accordion item causes the accordion item to toggle open (not to be confused with a collection item having its default state set to open) and the viewport slides to the top of that element, if possible.
- Confirm that a collection item that is set to be open by default displays as open by default (without being triggered by an anchor link).
- Confirm that the summary ID is the only HTML ID shown for the elements in an accordion item.
- Where multiselect or single select is configurable, it is working as expected, otherwise it should be single unless otherwise declared.
- Run a11y checkers against the pages containing accordions.
- Test background combos
- `aria-owns` information should be present for each accordion and the IDs match the IDs on the `<summary>` element for each accordion item.

### Sites to test
- sandbox.uiowa.edu
  - https://sandbox.uiowa.ddev.site/really-cool-accordions-copy
- admissions.uiowa.edu
  - https://admissions.uiowa.ddev.site/transfer-plans
  - https://admissions.uiowa.ddev.site/cost-aid/first-year-resident-scholarships
- housing.uiowa.edu
  - https://housing.uiowa.ddev.site/residence-halls/burge 
- commencement.uiowa.edu
  - https://commencement.uiowa.ddev.site/ceremony/fall-2023/college-liberal-arts-and-sciences-commencement-ceremony
  - https://commencement.uiowa.ddev.site/venue/carver-hawkeye-arena
- v2
  - https://hawkeyemarchingband.uiowa.ddev.site/drumhawk
- uipress.uiowa.edu
  - https://uipress.uiowa.ddev.site/books/23-woodcock-22-years
- inrc.law.uiowa.edu
  - https://lawinrc.uiowa.ddev.site/grants/bechtel-marie-h-charitable-remainder-uni-trust
- its.uiowa.edu
  - https://its.uiowa.ddev.site/services/microsoft-365-email-calendaring/connect-microsoft-365-email-and-calendar-your-mobile


joewhitsitt:
I think my comments from https://github.com/uiowa/uiowa/pull/8527 still apply with the work so far here.

pyrello:
> I think my comments from #8527 still apply with the work so far here.

Pulling those over to make parsing them easier in this context.

> I wonder if that changes anything with the aria-owns stuff because I think it was expecting the id from aria owns to be on the element with the role tab. Not 100% there though. That might have been the original reasoning why it ended up there, though it could have just been a mixup on my part.
> 
> Can you ease my concerns here?

From: https://github.com/uiowa/uiowa/pull/8527#issuecomment-2668869719

> @joewhitsitt the hash does not work when the id is set on the `<summary>`, and the UIDS example sets the id on `<details>` which is why the hash works in UIDS. Do we need to redo the work in UIDS so that aria owns is on the element with the role tab?

From: https://github.com/uiowa/uiowa/pull/8527#issuecomment-2668961815

> I think it is kind of a grey area because we are using all of this out of their normal usage.. aria-owns is not meant for tablist but because we do have it there, I think it expects the ids on the tab element/labelling element that is then read.
> 
> image
> 
> I noticed here, the reason the anchor works for UIDS is because there is a different id than what aria-owns uses. It would be nice reduce how many ids we need to account for.

From: https://github.com/uiowa/uiowa/pull/8527#issuecomment-2669065160

pyrello:
The accordion JS in UIDS is currently looking at the ID for the `<details>` element when reading or setting the URL hash fragment. There is no reason it needs to be like that. Maybe it would make the most sense to update that JS to read/write the ID from/to the summary element, since that element also needs to have an ID to work with `aria-owns`. It also works as the element to have the anchor attached to because it is the beginning of the details element and works with the scroll to behavior. If we leave it pulling from the `<details>` element, then we'll to define and implement to IDs for each accordion item, one on the `<details>` element for the hash fragment functionality, and one on the `<summary>` element for `aria-owns`.

joewhitsitt:
I followed the first part and think that works. You lost me on the second half, but I just don't think I understand with the wording as it is.

pyrello:
> I followed the first part and think that works. You lost me on the second half, but I just don't think I understand with the wording as it is.

I created https://github.com/uiowa/uids/pull/975 to switch the logic in UIDS. I think this allows us to only use/generate one unique ID per accordion item. The previous approach actually required two (and we couldn't just reuse one for both because HTML IDs should be unique).

docroot/themes/custom/uids_base/uids_base.theme
```diff
@@ -1094,18 +1094,18 @@ function uids_base_preprocess_block__inline_block__uiowa_collection(&$variables)
...
-    else {
-      // Fall back to the delta-based ID.
-      $variables['aria_owns'][] = $accordion_item_id . '-' . $delta;
     }
+    $variables['aria_owns'][] = $unique_id;
   }
-
+  // Convert this to a comma-separated string as that is how it will be printed
+  // in the template.
+  $variables['aria_owns'] = implode(', ', $variables['aria_owns']);
```
cory-skeers:
```suggestion
  $variables['aria_owns'] = implode(' ', $variables['aria_owns']);
```

Should this be space-separated instead of comma? The template was previously constructing it using `aria-owns="{{ _context['aria_owns']|join(', ') }}` which this matches, but the documentation I'm seeing has it as space-separated (eg https://www.w3.org/TR/wai-aria/states_and_properties#aria-owns). Not sure if there was a reason we'd had it commas previously.

pyrello: Good catch! Did you happen to check if Lighthouse flagged it with spaces vs commas? If it doesn't I'm good with making this change.

cory-skeers: I'm not seeing any difference with Lighthouse. With the Siteimprove extension, it flags the `aria-owns` on a `tablist` as incompatible with the commas but not with the spaces...which is good? According to the w3.org specs, tablist does support `aria-owns`, so maybe this is SI mislabeling what it's complaining about.

cory-skeers:
Noticing the ids aren't matching up for the commencement venue

Local
image

Prod
image

It looks like the `uids.theme` logic is setting the accordion item id after the commencement-specific processing. Working on an update for it.

cory-skeers (approved):
Original testing all checked out, and got a fix in for the regression I'd found in commencement (https://github.com/uiowa/uiowa/pull/8530#issuecomment-2683304652). Might be good to have another set of eyes check over my changes, but otherwise I think this is good to go.

pyrello:
I re-checked this after @cory-skeers updates. I found an issue with the v2 case, pushed a fix, and Cory re-checked it. We're good to go.

PR #8540, opened 2025-02-25
Added default hostname for google tag containers

### LINKED ISSUE #8520: Set default hostname for google tag containers
Follow up from #7538, there is now a required hostname field that is expecting a default value of `www.googletagmanager.com` but there is only placeholder on the input. On save, it is being validated but not focused because the input is under the advanced tab and it not visible.

e-marie-w:
Closes #8520 

# How to test
- sync a site `ddev blt ds --site=sandbox.uiowa.edu`
- visit the google tag manager settings page - `ddev drush @sandbox.local uli admin/config/system/google-tag/settings`
- scroll down to the Advanced section, the Hostname field should now have `www.googletagmanager.com` as actual highlightable text, not placeholder text that disappears when you click in the box.

I don't know why the file change shows that end line being removed and re-added, if you check https://github.com/uiowa/uiowa/blob/main/config/default/google_tag.settings.yml you can see the 404 is currently listed there. Likewise if i undo everything locally it's still there.


joewhitsitt:
It looks like the line number changes and the blank line at the end is removed. I don't think it is anything to worry about.

joewhitsitt (approved):

PR #8554, opened 2025-02-28
Custom 403

joewhitsitt:
I wish there was a better way to find the nid from the path, but my searches only come up with different versions of https://drupal.stackexchange.com/questions/197157/how-can-i-get-the-node-id-from-a-path-alias

I also don't love that I am doing similar code to handle the block preprocess. If we moved it to sitenow.profile it could call a trait right? otherwise, maybe it just moves to a helper function or a service.

# How to test

`ddev blt ds --site=education.uiowa.edu`

https://education.uiowa.ddev.site/node/6876 (unavailable, no login link)
https://education.uiowa.ddev.site/shaw-play (unavailable, no login link)
https://education.uiowa.ddev.site/admin/config (access denied, login link)

## Intranet
`ddev blt ds --site=connect.studentlife.uiowa.edu`

https://studentlifeconnect.uiowa.ddev.site/staff-news (normal custom intranet login required, login link)
https://studentlifeconnect.uiowa.ddev.site/admin/structure (normal custom intranet access denied, login link)


joewhitsitt:
Looking to get feedback on approach. I realize this isn't quite how we groomed it.

>I am finding that it is a bit more involved on immediate glance. page title is current served from a separate block. overriding the template is overriding the entire output (e.g. header and footer) and whatever I do I need to make sure intranet can "undo."

pyrello:
> I also don't love that I am doing similar code to handle the block preprocess. If we moved it to sitenow.profile it could call a trait right? otherwise, maybe it just moves to a helper function or a service.

This is just the regex matching to see if the route includes `node`, right? I'm not sure there is enough there that is similar enough to warrant worrying about a trait or helper function. I think if we find ourselves investing in this again, then we can revisit whether to build out more structure to support it. Not clear on the broader use case yet beyond alleviating unfounded suspicions that might lead to accusations.

> Looking to get feedback on approach. I realize this isn't quite how we groomed it.

I'm cool with this approach. It would be nice to add a comment to indicate why we are doing this (not the exact reason, but something that will help us remember why the code block is there).

pyrello:
Pulling to test.

pyrello:
I think we should change the text to "This page is unavailable and may be under review."

I mentioned this when we were discussing and grooming this issue. The rationale is that indicating that all unpublished content is under review potentially makes it look like we have a lot more content that is being reviewed. This seems undesirable since content under review might be something that is considered "out of compliance".

joewhitsitt:
Sounds good. I had misplaced that other language

pyrello (approved):
Tested. Looks good.

PR #8584, opened 2025-03-06
Updated archive/archived to unpublish/unpublished

### LINKED ISSUE #8574: Update Archive/Archived terminology
From MO:

> How big of a story is it to change "Archive/Archived" to "Unpublish/Unpublished" in all the places in the admin interface?
> 
> Is there an important distinction in our implementation between the two?  Feels like "archive" might suggest something more and different than "unpublish", so I'd like to make sure I'm aware of any nuances to the current terminology that I might not be picking up on.

pyrello:
Resolves #8574 

# How to test
```
```
- Observe and confirm label changes.


joewhitsitt:
pulling to test

joewhitsitt (approved):
Looks fine. James from ITS is usually in our sprint demos and I think we could point out to what extent we went in this slice. I'd say for the [Draft workflow](https://github.com/uiowa/uiowa/blob/main/config/default/workflows.workflow.draft.yml), we added that to the default thinking it could potentially be used by more websites and if so I think we would want to be consistent and bring ITS along with that change.

For ITS Alerts, yes, they had very specific instructions for the terminology because of their training docs.

For Brand Lockup, I agree that it probably makes sense in this case to keep the archived terminology.

james917:
I am reviewing this now

james917 (approved):
Looks good

PR #8588, opened 2025-03-06
AIS RFI webform MAUI session options

### LINKED ISSUE #8479: Pre-defined set of sessions to use in MAUI RFI handler
# Story

As a webmaster using the MAUI RFI handler, I would like to be able to add a select list of sessions from MAUI so that we are not burdened with the maintenance of manually creating and maintaining these options.

# Additional Info

Scott mentioned that 1 year back and 4 years forward would likely handle most all use cases.

Planned Enrollment Term
prospect.plannedEnrollmentSession
dropdown with -1 year and +4 years for fall, sping, summer *year*

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/8479

Screenshot 2025-03-07 at 11 58 51 AM
Screenshot 2025-03-07 at 11 59 00 AM
Screenshot 2025-03-07 at 11 59 41 AM
Screenshot 2025-03-07 at 11 59 50 AM

# How to test

`blt ds --site=education.uiowa.edu`

- No import errors.
- Go to a webform an add an "Options" element (e.g. select, checkboxes, etc.)
- Select from the options dropdown the AIS RFI MAUI Sessions, save the element. Go to the Test tab.
- See the options which excludes Winter sessions.


joewhitsitt:
We will want to wait to merge until the meeting (next week?) to get any customer feedback.

config/default/config_split.config_split.sitenow_webform_ais_rfi.yml
```diff
@@ -12,7 +12,9 @@ storage: folder
 folder: ../config/features/sitenow_webform_ais_rfi
 module:
   sitenow_webform_ais_rfi: 0
+  uiowa_maui: 0
```
pyrello, line 15: Is this here because the module is required? If so, it might be better to make it a dependency of `sitenow_webform_ais_rfi`, which will also ensure it is enabled

joewhitsitt: yes, the module is required because it leverages the API calls in that module. I added the module to the info file, but for existing sites with the feature on but uiowa_maui disabled, it didn't enable it.

i can remove it from here but i think an update hook would be needed instead for education, others? that have this feature turned on already

joewhitsitt: i don't know, i think it needs to stay in the split. otherwise it is trying to disable it since it isn't in core.extension default. The update hook fires and enables the module, 

```php
/**
 * Install uiowa_maui for existing websites with this feature enabled.
 */
function sitenow_webform_ais_rfi_update_10001() {
  // This module is now a dependency.
  \Drupal::service('module_installer')->install(['uiowa_maui']);
}
```

But then I get:

```
  Unable to uninstall the <em class="placeholder">uiowa_maui</em> module since the <em class=  
  "placeholder">SiteNow Webform AIS RFI</em> module is installed.  
```

joewhitsitt: For example: https://github.com/uiowa/uiowa/blob/cca204e01569afa23cb142cad6124f2348478164/config/default/config_split.config_split.thesis_defense.yml#L14

pyrello: Nevermind then. Not trying to blow this up. Just noting that while this is an established pattern, it creates complications. If we ever introduce config associated with the `uiowa_maui` module, then this will force that config to be completely split for this feature split. We have previously moved low cost dependencies into our default config to avoid just this type of thing. I was just trying to think about it in a future context.

joewhitsitt: I keep having to remind myself that uiowa_maui isn't turned on by default. I think we are getting closer. Maybe it could be on but the academic dates block has a separate permission or toggle?

pyrello: https://github.com/uiowa/uiowa/issues/8479#issuecomment-2675530401 seems to suggest that these options should be available for all sites?

joewhitsitt: Before Brian left, we were talking about this and how their specific requirements is reducing the usefulness of it for all. https://github.com/uiowa/uiowa/issues/8479#issuecomment-2707012788

joewhitsitt: Got more confirmation of this today from Ali that the needs right now are very undergrad specific and would not be particularly useful for others at this time.

I'd consider this an early prototype version. Depending on where it goes it could be updated for the websites using it or it could become deprecated and forms are updated with a replacement.

GaryRidgway:
I can pull and review this

GaryRidgway:
I have talked to Joe a little about some things I am seeing, but one thing that I don't think is as quick a fix for me is that there are some defaults here that are a bit confusing

<details><summary>

~~I am seeing these as the baseline defaults for the options elements, which seem a bit random?~~

</summary>
<p>

Screenshot 2025-03-12 at 11 59 19 AM


</p>
</details>

<details><summary>

~~And I say "Random" specifically because on refresh, they look like this.~~

</summary>
<p>

Screenshot 2025-03-12 at 12 01 27 PM


</p>
</details> 

UPDATE: apparently the test tab pre-populates random answers, and so would be the reason for my confusion. DISREGARD.

GaryRidgway (approved):
After the year correction, this all looks good to me!

joewhitsitt:
Thanks for making that change @GaryRidgway

PR #8589, opened 2025-03-07
[Hotfix] Add existence/type check for collections block

cory-skeers:
# How to test

On `main`
```
ddev blt ds --site=k12-cancer.sites.uiowa.edu && ddev drush @sitesstem-k12-cancer.local uli /she-oncology
```
Attempt to edit the layout and get a WSOD
Check out this branch and attempt to edit the layout
success

Alternatively, edit a layout on any site, add a new collection, and do not save the page
On `main`, return to the layout and get a WSOD
On this branch, return to the layout and see that you can edit it

Consider https://github.com/uiowa/uiowa/pull/8589#issuecomment-2706933744

### DIFF
--- a/docroot/themes/custom/uids_base/uids_base.theme
+++ b/docroot/themes/custom/uids_base/uids_base.theme
@@ -1245,11 +1245,13 @@ function uids_base_preprocess_paragraph__uiowa_collection_item(&$variables) {
   if (!$admin_context->isAdminRoute()) {
 
     $block = $paragraph->getParentEntity();
-    $unique_id = 'accordion-item-' . $block->id() . '-' . $variables['delta'];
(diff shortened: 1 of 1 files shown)


cory-skeers:
Relates to Cherwell incident 2470241

cory-skeers:
This approach means the accordion heading won't work correctly when they are first added to a layout in terms of `aria-owns` property, but avoids the WSOD.

The problem comes from the block not actually existing yet, so not having an `id` to construct in normal circumstances, which means the above won't ever be an issue on a published, publicly visible version of the page, but another option would be to have a fallback option for properly constructing the accordion item id for use with `aria-owns`.

joewhitsitt (approved):
Gah! How did we miss this earlier?! Thanks for the quick workaround @cory-skeers

PR #8592, opened 2025-03-12
Ignore field teaser in Find Text

### LINKED ISSUE #8594: Teaser results in Find Text
Results potentially populate, but the field is not used anywhere in V3, and also not viewable now.

The table can be easily ignored in the Find Text settings as one option, or we can run an update hook on V3 sites to clear out this data. If the former, we'll need to split the settings for the V2 feature to not ignore this field.

> I am leaning towards the ignore and v2 split and we just document it as part of the wind down task to remove the field for good. @joewhitsitt

cory-skeers:
Resolves #8594 

Field Teasers are still showing up in Find Text results, though the field is no longer used nor accessible by webmasters in V3

# How to test

On main, sync a site and do a Find Text search for something that will match a field teaser (pharmacy.uiowa.edu, search for "diversity"), and see a "Teaser" result
Check out this branch, do a config import
Repeat the search, and no longer see a "Teaser" result

Sync a V2 site and perform a Find Text search for something that will match a field teaser, and see that a "Teaser" result still appears
(Example: `ddev blt ds --site=hawkeyemarchingband.uiowa.edu && ddev drush @hawkeyemarchingband.local uli /node/1/edit`
Add a unique text string to Content Description, search for it in Find Text, and see that you get a result)

### DIFF
--- a/config/default/find_text.settings.yml
+++ b/config/default/find_text.settings.yml
@@ -50,4 +50,5 @@ tables_to_skip:
   - block_content_field_data
   - node_revision
   - paragraphs_item_field_data
+  - node__field_teaser
 save_as_csv: 0
--- a/config/features/sitenow_v2/config_split.patch.find_text.settings.yml
+++ b/config/features/sitenow_v2/config_split.patch.find_text.settings.yml
@@ -1,0 +1,4 @@
+adding: {  }
+removing:
+  tables_to_skip:
+    - node__field_teaser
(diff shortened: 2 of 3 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
Checks out. A bit of a fright with that permission change but confirmed that they don't see any of the region settings functionality currently in PROD and that HR synced as expected.

PR #8615, opened 2025-03-25
Diversity deprovision/htaccess

joewhitsitt:
# How to test


joewhitsitt:
my guess is the deprovision will be added to this and deployed at the same time correct @briand44? so holding on this for now

briand44:
I had the htaccess changes separate from the deprovision on the card but thought the team might want to do those two things together to save from having an additional deployment. I think it is fine to do them together, if we have issues with htaccess redirects we will just have to work through them without having a site to fallback on.

joewhitsitt:
yeah, the additional deployment is what i'd like to avoid

GaryRidgway (approved):
Reviewed, not tested.

PR #8623, opened 2025-03-26
Updating temporary_maximum_age for files

### LINKED ISSUE #8587: Lower time to delete file garbage collection process
# Story

As an editor, after I delete a media entity from the backend, more often than not, I do not want to wait six hours for the related file to be deleted off the server.

# Background

I think we took the current limit as it is and were waiting for lower options to be added by Drupal Core and they haven't come.

Though the safety window to recover a deleted file is nice, editors have been trained to understand that with all other content on SiteNow, once you hit delete it is gone for good.

# Proposed Solution

First try changing the configuration in /default. admin/config/media/file-system

1. sync a site `ddev blt di --site=default && ddev drush @default.local uli /admin/config/media/file-system`
2. Export change `ddev drush @default.local cex`
3. Check exported changes

OR

Instead of a form_alter to rewrite available options for the configuration form that is only visible to developers. Set a lower limit in global.settings.php

`$config['system.file']['temporary_maximum_age'] = 3300;`

At 55 minutes (3300 seconds), this will cause a file to delete within the hourly cron window.

james917:
# How to test

Run 
```
ddev blt di --site=default && ddev drush @default.local uli /media/add/image
```

1. Add image then add it to a page
2. Then delete the image
3. Run cron /admin/config/system/cron an hour later and see if image is deleted from local sites/default/files/2025-03


bspeare:
I tested this twice and it deleted the file within an hour.  Would love it if someone else could confirm!

joewhitsitt:
This worked for me as well. The file_managed table removes the row and the file is no longer in the site's files directory. There is the issue that if the file system page is saved, the value is stored as 0 = Never which will persist until the next deployment/cim

```
ddev drush @sandbox.local config:get system.file temporary_maximum_age
'system.file:temporary_maximum_age': 0
```

Not a major concern as non-admins don't have access to this page but not nothing either.

joewhitsitt:
@james917 if you add uiowa/developer as a reviewer to this PR, I will approve it.

james917:
@joewhitsitt I fixed it and added uiowa/developer

joewhitsitt (approved):

PR #8646, opened 2025-04-03
Maintenance mode - ignore changes to the message

joewhitsitt:
# How to test

Check out this branch `ignore_maintenance_message`.
Sync a site and change the message at admin/config/development/maintenance.
Run a `cim` and the message should not change. "There are no changes to import."
Viewing the website as anonymous should show your custom message, not the config default.


e-marie-w (approved):
Followed testing instructions, it works as expected!

PR #8648, opened 2025-04-04
Updating Panapto to get title video for iframe.

### LINKED ISSUE #6705: As an editor, I would like to have a unique Title field of an iframe embed so I don't get accessibility errors
# Story

When embedding a UICapture panopto video on a page the iframe has a Title of `title="Embedded content from uicapture.hosted.panopto.com"`. 

Screen Shot 2023-07-27 at 3 50 01 PM

When you have more than one UICapture embed on one page I get a error from SiteImprove saying "There are multiple inline frames on this page with the same text alternative. Every inline frame (iframe) needs a unique, descriptive text alternative." - SiteImprove

# Proposed solution

- https://sandbox.prod.drupal.uiowa.edu/media-panopto
- Title is being set by custom module: https://github.com/uiowa/uiowa/pull/5276. 
- Video is being constructed at https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/Field/FieldFormatter/PanoptoUrlFormatter.php. 
- https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/media/Source/Panopto.php
- https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/Validation/Constraint/PanoptoURLConstraintValidator.php

james917:
# How to test

Add a video, this is a test [video](https://uicapture.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=7cfdf7ca-2723-4d22-a63e-ada9016ff970)

Save the video add it to a page and see if title has the title of the Panopto video or the default verbage "Embedded content from uicapture.hosted.panopto.com"

Sync sandbox and re-save placed UI Capture videos there. Inspect to see the title attribute has been replaced with the video title.

### DIFF
--- a/docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
+++ b/docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
@@ -52,26 +52,28 @@ function sitenow_media_wysiwyg_media_presave($entity): void
     if (!$entity->hasField($source_field) || $entity->get($source_field)->isEmpty()) {
       return;
     }
-
     // Get the source field value.
     $source_field_value = $entity->get($source_field)->getValue();
     $parsed = UrlHelper::parse($source_field_value[0]['uri']);
     $id = $parsed['query']['id'];
-    $url_response = \Drupal::httpClient()->get(
-      Panopto::BASE_URL . "/Panopto/Pages/Embed.aspx?id=" . $id
-    );
 
-    $html = $url_response->getBody();
-
-    $document = Html::load($html);
-
-    $title = $document->getElementsByTagName('title')->item(0)?->textContent;
-    // Trim up the title.
-    if ($title) {
-      // Trim any extra whitespace
-      $title = trim($title);
-      $source_field_value[0]['title'] = $title;
-      $entity->set('field_media_panopto_url', $source_field_value);
+    try {
+      $url_response = \Drupal::httpClient()->get(
+        Panopto::BASE_URL . "/Panopto/Pages/Embed.aspx?id=" . $id
+      );
+      $html = $url_response->getBody();
+      $document = Html::load($html);
+      $title = $document->getElementsByTagName('title')->item(0)?->textContent;
+      // Trim up the title.
+      if ($title) {
+        // Trim any extra whitespace
+        $title = trim($title);
+        $source_field_value[0]['title'] = $title;
+        $entity->set('field_media_panopto_url', $source_field_value);
+      }
+    }
+    catch (\Exception $e) {
+      \Drupal::logger('sitenow_media')->error('Failed to retrieve Panopto video information: @error', ['@error' => $e->getMessage()]);
     }
   }
 }
--- a/docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
+++ b/docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
@@ -36,6 +36,18 @@ function sitenow_media_wysiwyg_inline_entity_form_entity_form_alter(&$entity_for
 }
 
 /**
+ * Implements hook_entity_presave().
+ */
+
+function sitenow_media_wysiwyg_entity_presave(EntityInterface $entity) {
+  // If unique id, clean it and save it as value.
+  if ($entity->getEntityTypeId() === 'paragraph' && $entity->hasField('field_uip_id') && !$entity->get('field_uip_id')->isEmpty()) {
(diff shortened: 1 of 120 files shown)


joewhitsitt:
just pulled down and it only does it on media add. i think we will want to give existing content a chance to update to use the video title so we should think about lifting and shifting what you have to somewhere else.

james917:
@joewhitsitt  I am looking into it now

james917:
@joewhitsitt  I made the updates last night they should be ready for review. Please let me know if you would like to go over it I added a lot of comments in the code so hopefully you understand kind of my thought process on it

docroot/modules/custom/sitenow_paragraphs/sitenow_paragraphs.module
joewhitsitt, line 1: Changes to this file shouldn't even exist. This was only supposed to serve as an example of hook_entity_presave

docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/media/Source/Panopto.php
joewhitsitt, line 1: We want to move what we did here out of this file and into sitenow_media_wysiwyg.module

docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
```diff
@@ -35,6 +35,57 @@ function sitenow_media_wysiwyg_inline_entity_form_entity_form_alter(&$entity_for
...
 }
 
+/**
+ * Implements hook_entity_presave().
+ *
+ * Ensures Panopto media entities have their titles set from the video's title.
+ */
+function sitenow_media_wysiwyg_entity_presave($entity) {
+  // Only process media entities.
+  if ($entity->getEntityTypeId() !== 'media') {
```
joewhitsitt: We could switch to sitenow_media_wysiwyg_media_presave (hook_ENTITY_TYPE_presave) and avoid this check.

docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
```diff
@@ -35,6 +35,57 @@ function sitenow_media_wysiwyg_inline_entity_form_entity_form_alter(&$entity_for
...
+
+  // Only proceed if we have a source field value.
+  if (!$entity->hasField($source_field) || $entity->get($source_field)->isEmpty()) {
+    return;
+  }
+
+  // Get the source field value.
+  $source_field_value = $entity->get($source_field)->getValue();
+
+  // If we have a title in the source field, use it for the media entity name.
```
joewhitsitt: This is not what we want to do. We want to set the source field's `title` value with the html title we get from the Guzzle request. Since this fires for both initial save and updates after, might as well check for a updated value from the source (changed video title) and use it. The media name name is set in the Panopto file and we want to keep that there.

docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
james917: resolved

james917: Switched to entity type.

joewhitsitt:
pulling to take a look

joewhitsitt:
I think this looks good except we should probably wrap this in a try/catch (log error and move on) statement to avoid a error screen if the upstream source is down/unavailable.

@james917 do you want to try to get that implemented?

james917:
Updated the try/catch ready for review

joewhitsitt:
@james917 just pointing out that Travis identified some [coding standards issues](https://app.travis-ci.com/github/uiowa/uiowa/builds/274814700#L1639):

```
FILE: .../docroot/modules/custom/sitenow_media_wysiwyg/sitenow_media_wysiwyg.module
--------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 45 | ERROR | [x] Opening brace should be on the same line as the declaration
 67 | ERROR | [x] Inline comments must end in full-stops, exclamation marks,
    |       |     question marks, colons, or closing parentheses
```

I'll test the try/catch though

joewhitsitt (approved):
Thanks for making those coding standards changes @james917.

I just tested the try catch and it works as expected. Made a few other tweaks, and added some unrelated items.

Resolves https://github.com/uiowa/uiowa/issues/6705

PR #8652, opened 2025-04-07
Signage Proof of concept

### LINKED ISSUE #8640: Proof of Concept of Signage sign/slide using layout builder
# Story

As a web team, we would like to put together a proof of concept of a Signage sign/slide using layout builder so that we can better understand what is all involved here and if there are any areas of complexity.

# Additional Info

Grooming conversation... 

Area of complexity - Proof of concept of ability to create a sign with a slides using layout builder implementation within SiteNow.

What does it look like to build a sign/slide in layout builder

slide/sign content type that uses layout builder
ability to display slideshow in a layout

GaryRidgway:
#### Things I'm skipping over in the name of speed.
Slide content type
- LB page should remove the "add section" buttons if there is already a section.

Sign content type
- LB page should remove the "add section" buttons if there is already something like, 3 sections??
- figure out how to display things in a slideshow box that have slides with different block sizes.
- Want a local action that allows you to go to the display path.

Slideshow block
- Only allow on sign content type.
- Currently the node titles are just being hidden in SCSS. They should be either preprocessed out or removed in a template.

Display view
- Figure out a better way to make a view for this. When I first wrote the JS, it was so that any page could be a sign. Now, it seems potentially unnecessary and basic when we could maybe make a view mode or something?

# How to use

- Make some slides
- Make a sign
- Add a slideshow somewhere on a sign
- add the slides to the slideshow!
- OBSERVE


pyrello:
Noticing that this PR introduces some new files that use camel casing instead of dash casing. I don't have anything against camel casing file names, but most of the CSS and JS files in our platform use dash casing, so unless we decide as a team to pivot to that as our standard, I don't think we should be using both types. I know that this has happened before and so we're already in a situation where we have both, but I don't think we should encourage that going forward.

GaryRidgway:
> Noticing that this PR introduces some new files that use camel casing instead of dash casing. I don't have anything against camel casing file names, but most of the CSS and JS files in our platform use dash casing, so unless we decide as a team to pivot to that as our standard, I don't think we should be using both types. I know that this has happened before and so we're already in a situation where we have both, but I don't think we should encourage that going forward.

Totally fine, I think we should transition them to dash casing then.

pyrello:
I'm going to close this in favor of #8668 so the information related to the discovery and proof of concept that happened in this PR can be preserved.

Issue #4916, opened 2022-03-08
As a site owner, I can use the sitenow-events page with the updated event page design


bspeare:
# Story

As a site owner, I can use the `sitenow-events` page with the updated event page design.

We created a new page design for events in https://github.com/uiowa/uiowa/pull/2765 and it would be great to update the sitenow-events page to use the same design.

# Proposed solution

- Move sitenow-events into a template/design similar to https://github.com/uiowa/uiowa/blob/main/config/features/event/core.entity_view_display.node.event.default.yml.

bspeare:
Related issues:

- https://github.com/uiowa/uiowa/issues/4169
- https://github.com/uiowa/uiowa/issues/3274

joewhitsitt:
Is there a way this can be "used" but separate so that it doesn't muddy the line between local and dynamic events when we do https://github.com/uiowa/uiowa/issues/3458

bspeare:
@joewhitsitt do you mean a separate page design or are you referring to something else?

joewhitsitt:
meaning that eventually, uiowa_events (ui events) will be split out from sitenow_events (local) and so if we define a event template it would live in sitenow_events or uids_base and get called into uiowa_events.

PR #8660, opened 2025-04-08
Events single page multiple-dates dropdown color contrast fix

### LINKED ISSUE #5495: Events single page multiple-dates dropdown need more color contrast
# Problem

https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html

Past dates on uiowa events single pages have insufficient color contrast.
~https://education.uiowa.edu/event/92671/0~
https://sandbox.prod.drupal.uiowa.edu/event/31316/21

Text color
 #999999 / rgb(153,153,153)
Background color
 #ffffff / rgb(255,255,255)
Contrast ratio
2.85 out of a required 4.5

WAVE shows a couple more examples of contrast ratio issues on elements that are not visible on the page. Do we need to address those as part of this, separate or do we not need to worry about those?

# Proposed solution

- Change to an accessible gray on white value.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/5495. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /event/31316/21
```
1. Confirm that first date has more space between it and the remaining dates
2. Confirm new gray color passes AA validation: https://webaim.org/resources/contrastchecker/?fcolor=737373&bcolor=FFFFFF.

### DIFF
--- a/docroot/modules/custom/sitenow_events/css/single-event.css
+++ b/docroot/modules/custom/sitenow_events/css/single-event.css
@@ -50,10 +50,13 @@
   font-weight: bold;
 }
 .date-instance__past {
-  color: #999;
+  color: #737373;
 }
-.date-instance__next-upcoming {
-  font-weight: bold;
+.event-field.event-time .date-instance:nth-child(2) {
+  margin-top: 1rem;
+}
+.date-instance__next-upcoming .date-instance__date {
+  font-weight: 600;
 }
 
 .sitenow-event .svg-inline--fa {


pyrello:
Pulling to test.

docroot/modules/custom/sitenow_events/css/single-event.css
```diff
@@ -50,10 +50,13 @@
   font-weight: bold;
 }
 .date-instance__past {
-  color: #999;
+  color: #737373;
```
pyrello, line 53: Should we be using a variable for this?

bspeare: It's not configured to be compiled in a workspace. Maybe we wait for https://github.com/uiowa/uiowa/issues/4916 for that?

pyrello: Do we not just have a CSS variable for this value?

bspeare: Nope, we need to switch everything from SCSS variables to CSS variables.

bspeare: Added https://github.com/uiowa/uids/issues/985 as a reminder to start that process.

pyrello (approved):
Looks good. Thanks!

PR #8664, opened 2025-04-10
Adjust caching method on footer login link

joewhitsitt:
We went back and forth on this max-age=0 on the footer link with a note to revisit.

https://github.com/uiowa/uiowa/pull/6040#pullrequestreview-1232019447

The solution for now is to get caching back with cache contexts!

# How to test

On education.dev.drupal.uiowa.edu navigate around the website anonymously and see that the login link in the footer is updated with the destination parameter.

Login and change one of the page URL paths. As an anonymous user again. View the page and see the login link is accurate for the path.

Note that subsequent hits to the page without changes to the node are `x-cache` hit meaning not only drupal is cached but that varnish is serving from a cached version. `age` and `x-cache-hits` should also increment.

Screenshot 2025-04-11 at 10 44 04 AM

### DIFF
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -1130,9 +1130,6 @@ function uiowa_core_preprocess_block(&$variables) {
             ],
             'rel' => 'nofollow',
           ],
-          '#cache' => [
-            'max-age' => 0,
-          ],
         ];
       }
       break;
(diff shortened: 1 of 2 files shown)


joewhitsitt:
didn't work... https://education.dev.drupal.uiowa.edu/about/strategic-plan

joewhitsitt:
Oh heck yea! I had to clear site and varnish cache first. Now I am not sure which change it was but my guess is that it is the login block. system messages block is not in the DOM.

https://education.uiowa.edu/about/strategic-plan

Screenshot 2025-04-10 at 5 28 48 PM

From what I can tell the cache context seem to be working. I am getting expected destinations as I traverse the website and the login link disappears once I am authenticated.

james917 (approved):

joewhitsitt:
Thanks for reviewing @james917

PR #8677, opened 2025-04-16
Security/Travis fixes

joewhitsitt:
https://github.com/travis-ci/dpl/issues/1281#issuecomment-2101322835

https://git.drupalcode.org/project/stage_file_proxy/-/compare/3.1.3...3.1.5?from_project_id=8559

# How to test

- Review as desired but technically a guess and check to see if we can still build for a stage or prod deployment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,18 +42,24 @@ deploy:
 - provider: script
   script: "${BLT_TRAVIS_DIR}/scripts/deploy_branch"
   skip_cleanup: true
+  edge:
+    branch: v2.0.3-beta.4
   on:
     branch: develop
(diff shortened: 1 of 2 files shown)


joewhitsitt:
stage_file_proxy has an update hook but it is to update the origin field and remove a trailing slash. we don't store this as config and we don't use it in prod. generally i think we are okay with this update, but just to make Travis happy.

joewhitsitt:
Passed on develop https://app.travis-ci.com/github/uiowa/uiowa/builds/274921777

pyrello (approved):
Reviewed, not tested.

PR #8700, opened 2025-04-25
ITS - Only send ITS alert digest if there are alerts

### LINKED ISSUE #8689: ITS Alert Digest sends an empty email when there are no alerts
# Problem
On 4/23, there were no active or 7-day out planned alerts, but an email was still sent.

Image

# Proposed solution
- Check and adjust the logic to see if there is output beyond the related links before sending the email.

pyrello:
Resolves #8689 

# How to test
The testing instructions assume that there are currently no alerts. If a new alert gets created before this gets tested, then you'll be first confirming that the alert digest sent, logging in and removing the alert, then confirming that the alert digest didn't send.
```
ddev blt ds --site its.uiowa.edu && ddev drush @its.local its_core:alerts-digest && ddev drush @its.local uli node/add/alert
```
- Confirm that you see the `Alerts Digest - No items to send` message.
- Add an alert.
```
ddev drush @its.local its_core:alerts-digest
```
- Confirm you see the `Alerts Digest sent` messgae.
- You can also check Mailpit to see the email: https://uiowa.ddev.site:8026/

### DIFF
--- a/docroot/sites/its.uiowa.edu/modules/its_core/src/Commands/ITSCoreCommands.php
+++ b/docroot/sites/its.uiowa.edu/modules/its_core/src/Commands/ITSCoreCommands.php
@@ -47,22 +47,20 @@ class ITSCoreCommands extends DrushCommands {
     $this->accountSwitcher->switchTo(new UserSession(['uid' => 1]));
 
     $views = [];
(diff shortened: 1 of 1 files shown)


joewhitsitt (approved):
Checks out. Thanks.

PR #8714, opened 2025-04-30
hotfix - resave default content after issues with new site installs

joewhitsitt:
Two recent site installs (first since default content was updated) are erroring when trying to view node 1 or the layout.

`TypeError: Drupal\Core\Render\Element::children(): Argument #1 ($elements) must be of type array, null given, called in /var/www/html/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/SectionComponentSubscriber.php on line 183`

https://github.com/uiowa/uiowa/blob/ce574f8e9f4c74f673fccac4714e0e6c767be69a/docroot/modules/custom/layout_builder_custom/src/EventSubscriber/SectionComponentSubscriber.php#L181-L187

When debugging, it thinks the plugin id is card but the content is statistic which then fails when it assumes the card image exists.

Screenshot 2025-04-30 at 3 22 21 PM
Screenshot 2025-04-30 at 3 23 04 PM

# How to test

gut the files directory except for .htaccess - `docroot/sites/default/files`
```
ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default
```
Login and try to resave /node/1/layout

Node Revision Delete changelog: https://git.drupalcode.org/project/node_revision_delete/-/compare/2.0.0-rc1...2.0.0?from_project_id=49284

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -161,7 +161,7 @@
         "drupal/migrate_plus": "^6.0",
         "drupal/migrate_tools": "^6.0",
         "drupal/moderated_content_bulk_publish": "^2.0",
-        "drupal/node_revision_delete": "^2.0@alpha",
+        "drupal/node_revision_delete": "^2.0",
         "drupal/node_view_permissions": "^1.5",
         "drupal/override_node_options": "^2.4",
         "drupal/paragraphs": "^1.10",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "556ad8a88ebeda3e4363802af733fd6b",
+    "content-hash": "b015ced1afe836ffe2571a36fafbef0a",
     "packages": [
         {
             "name": "acquia/blt",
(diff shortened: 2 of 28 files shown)


joewhitsitt:
I think the true test will be on site install in prod as that is where the issue currently showed up.

bspeare:
How to bypass this error? `  [Drupal\Core\File\Exception\DirectoryNotReadyException]                                                
  The specified file 'profiles/custom/sitenow/content/file/UIowa.CRWC-40.jpg' could not be copied because the destination directory 'public://2022-06' is not properly configured. This may be caused by a problem with file or directory permissions.  `

joewhitsitt:
great question. perhaps your mutagen sync isn't as fast as the site install? also not sure what is up with percy and the about page.

bspeare:
That is weird.  I've seen Percy do that before.  I'll let someone else test this and see if they can sync this site.

pyrello:
Pulling to test.

pyrello:
I tested this and did ultimately get to a point where I was able to get the site install working and re-save the node 1 layout. However, I ran into an issue with deleting `docroot/sites/default/files` and wasn't able to complete the site install until I restored the directory and just removed its contents. I wasn't immediately able to re-save the node 1 layout because I was getting a DB error saying that the base table `queue` didn't exist. I re-saved the node, then went back to the layout and was able to re-save. This doesn't make me feel totally confident that it worked cleanly, but it seemed to be okay in the end.

pyrello:
I ran through the test instructions again. At the point of re-saving the layout, I hit this error again:

```
Drupal\Core\Entity\EntityStorageException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'uiowa.queue' doesn't exist: SELECT "q"."item_id" AS "item_id" FROM "queue" "q" WHERE ("name" = :db_condition_placeholder_0) AND ("data" = :db_condition_placeholder_1) AND ("expire" = :db_condition_placeholder_2); Array ( [:db_condition_placeholder_0] => node_revision_delete [:db_condition_placeholder_1] => i:1; [:db_condition_placeholder_2] => 0 ) in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).
```

Not sure if this is just an issue with my local or not. Not sure if its worth pushing to dev to test it or if we just roll it out and see if it works and fix it if it doesn't.

bspeare:
Looks like we lost that 6th screenshot for the "About" page snapshot from Percy in the grid update pr and this code adds it back.  I approved Percy for it.

joewhitsitt:
i will note that deleting the files directory completely removes the .htaccess file there and that trying to do a site install caused issues for me. i should have adjusted the wording to "gut the files directory except for .htaccess" but I hesitated.

Thanks @bspeare for probing the percy result.

@pyrello i am going to try to sync down and try to recreate that queue issue. never seen that before.

I was also going to consider re-exporting node/2 since the node file is modified and that that might trigger more block changes with `dcem`

joewhitsitt:
Looks like https://www.drupal.org/project/node_revision_delete/issues/3436432 this is the queue issue.

joewhitsitt:
@pyrello can you pull again with the latest changes and try again?

https://git.drupalcode.org/project/node_revision_delete/-/compare/2.0.0-rc1...2.0.0?from_project_id=49284

bspeare (approved):
I was able to login and save node 1.

PR #8722, opened 2025-05-01
Default Content - go back to old patch, re-export node 1 and 2

joewhitsitt:
Background: https://iowaweb.slack.com/archives/C0164FNN1PV/p1746117287134089

Attempt to go back to the patch that was working for the most part. I _didn't_ get the `Object support when dumping a YAML file has been disabled.` error this time.

# How to test

gut the files directory except for .htaccess - `docroot/sites/default/files`
```
ddev drush @default.local sql-drop -y && ddev drush @default.local cr && ddev blt di --site=default
```
Login and try to resave /node/1/layout

"_Everything in Its Right Place_" 🎶

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -261,7 +261,7 @@
                 "Changing an existing embedded media's alignment or alt data attributes does not get saved with CKEditor": "https://www.drupal.org/files/issues/2023-01-03/3330723-3.patch"
             },
             "drupal/default_content": {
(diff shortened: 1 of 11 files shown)


bspeare (approved):
I was able to resave without an issue.

joewhitsitt:
Thanks for the quick review @bspeare

PR #8730, opened 2025-05-01
[Decorative images] Added decorative image widget module

### LINKED ISSUE #8711: Implement ability to add decorative images
# Story

As a SiteNow user, I would like to have the ability to add a decorative image without alt text so that screen readers aren't forced to read alt text for images that only purpose is decorative.

# Additional info

As part of #8683, we determined that the [decorative image widget module](https://www.drupal.org/project/decorative_image_widget) seems to meet our needs and we would like to move forward with implementing that solution. https://github.com/uiowa/uiowa/issues/8683#issuecomment-2839441130

We should also include adding additional documentation on how to add a decorative image and override alt text per instance.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8711. 

# How to test

- Review draft of https://sitenow.uiowa.edu/node/326/latest.  Added a section under "Alternative text" on the decorative images. 

- Only intranet sites are splitting https://github.com/uiowa/uiowa/blob/main/config/features/sitenow_intranet/config_split.patch.field.storage.media.field_media_image.yml.  Can you think of anywhere else this is split?

```
ddev composer install && ddev blt ds --site=connect.studentlife.uiowa.edu && ddev drush @studentlifeconnect.local uli /media/add/image
```
1. Add an image. 
2. Confirm you can check the box to make decorative. 
3. Confirm you can override alt text on that image when embedding the image in a page by clicking on the "eye" icon when you click on the embedded image. 
4. Test other places you can think of like a text area block, etc. 
5. Try editing an existing image alt text by turning it decorative under ` /media/` and you will get an error.  I've tested this on `main` and get the same issue. `Source image at [path] not found while trying to generate derivative image at [path].` 

```
ddev composer install && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /media/add/image
```
1. Add an image. 
2. Confirm you can check the box to make decorative. 
3. Confirm you can override alt text on that image when embedding the image in a page by clicking on the "eye" icon when you click on the embedded image. 
4. Test other places you can think of like a text area block, etc. 
5. Try editing an existing image alt text by turning it decorative under ` /media/` and you will get an error.  I've tested this on `main` and get the same issue. `Source image at [path] not found while trying to generate derivative image at [path].`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -101,6 +101,7 @@
         "drupal/cshs": "^4.0",
         "drupal/ctools": "^4.0",
         "drupal/date_popup": "^2.0",
+        "drupal/decorative_image_widget": "^1.0",
         "drupal/default_content": "^2.0@alpha",
         "drupal/devel": "^5.1",
         "drupal/diff": "^2.0@beta",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "b015ced1afe836ffe2571a36fafbef0a",
+    "content-hash": "20d85226c11faf8f3e2eae1b8c223942",
     "packages": [
         {
             "name": "acquia/blt",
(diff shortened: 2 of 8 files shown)


joewhitsitt:
pulling

bspeare:
Updated testing instructions with draft of https://sitenow.uiowa.edu/node/326/latest.

joewhitsitt:
>Try editing an existing image alt text by turning it decorative under  /media/ and you will get an error. I've tested this on main and get the same issue. Source image at [path] not found while trying to generate derivative image at [path].

Good news is I don't think I am running into this error on sandbox. I've previously --sync-public-files which might help editing existing images. Ultimately I think it is unrelated to the module and something we have seen with image style generation, not the html markup.

joewhitsitt (approved):
Looks pretty solid. Also tested v2 section bg image and ran it through a couple other accessibility checks to make sure none of them took issue with the alt attribute being present but empty.

bspeare:
Thanks @joewhitsitt!

PR #8731, opened 2025-05-02
[Signage] Branded header

### LINKED ISSUE #8715: Implement branded header region for Signage Signs
# Story

As a Signage user, I would like a UI branded header region on my sign so that we meet University Brand standards.

# Additional Info

Grooming conversation - https://github.com/uiowa/uiowa/issues/8528#issuecomment-2842908320

>Header may set the stage for other theming work. Could handle sign header by adding fields for configurable elements of the header to the sign content type. Block Iowa, Sign Name, Date and Time possible parts of the header. Discussed holding off on date/time for now and focusing on understanding how we want to implement the Block Iowa and Sign Name and getting those rendering in a header. If we need ability to hide header that would be a separate slice.

In Scope
- Block Iowa
- Sign Name

Out of Scope
- Date/Time
- Ability to hide header

# Proposed solution
- Add sign name field to the sign content type.
- Display the block Iowa icon and the sign name field in Layout Builder in a section above the section that the user can add blocks to. This could be done in the content type template or possibly via some type of preprocessing in code.
- If using the content type template, ensure that the block Iowa and sign name field blocks or the section they are are are not configurable in Layout Builder.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8715.

# How to test

```
ddev blt frontend && ddev blt ds --site=default  && ddev drush @default.local config-split:activate signage -y && ddev drush @default.local uli  /node/add/sign  
```
1. Masquerade as user with only "Sign manager" permissions.
2. Add a new sign.
3. Add a slide.
4. Add a slideshow block to the slide layout.
5. Confirm that the branding bar renders and that you are unable to edit or move that bar in the layout.
6. Confirm that each layout section is set to "edge-to-edge", "Remove top padding", and a black background as a default.  Confirm the "Styles" option is not available and that existing layout builder styles are limited to single choices. 
7. Confirm that you can't "move" a block in the contextual links. 
8. Edit sign and confirm that checking the "Hide header" block hides the header.


pyrello:
Pulling to test.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -5,17 +5,114 @@
...
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
+use Drupal\layout_builder\Plugin\SectionStorage\OverridesSectionStorage;
 use Drupal\node\NodeInterface;
 
+/**
+ * Implements hook_module_implements_alter().
+ */
+function sitenow_signage_module_implements_alter(&$implementations, $hook) {
+  if ($hook == 'form_alter') {
```
pyrello, line 18: Can we add a comment here explaining what is happening and why? I'm not understanding it just by reading the code.

pyrello: I guess I'm surprised that `sitenow_signage` hooks don't fire after `layout_builder_custom`, since it comes later alphabetically.

pyrello (approved):
Works as expected. Nice work.

bspeare:
Thanks @pyrello!

PR #8733, opened 2025-05-02
Add HSTS header to .htaccess

### LINKED ISSUE #8674: Add the Strict-Transport-Security header to .htaccess
From our ticket with Acquia: https://acquia.my.site.com/s/case/500Pb00000maTavIAE/followup-on-courtesy-investigation-01169884

It is recommended that we add the STS header to the .htaccess file in addition to our http -> https forwarding.

Follow-up from https://github.com/uiowa/uiowa/issues/8606

Acquia docs:
https://acquia.my.site.com/s/article/360004119254-How-To-enable-HSTS-for-your-Drupal-site

e-marie-w:
Closes #8674 

# How to test
- to test locally: sync down a site & uli `ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli`
- to test on dev: visit a dev site, like https://sandbox.dev.drupal.uiowa.edu, and follow the same steps below
- Inspect the page and click the Network tab. 
- If testing on Brave: Click one of the local files. The Headers tab should be selected automatically, check that the request URL starts with the local or dev domain instead of the prod domain. You should see `Strict-Transport-Security: max-age=31536000; includeSubdomains;` listed under the Response Headers section.
- If testing on Google Chrome: Need to click Reload Page in the Network tab. Then click the local or dev domain, which should be first in the long list of options. See that `Strict-Transport-Security: max-age=31536000; includeSubdomains;` is set, again listed under the Response Headers section.
- If testing on Firefox: same steps as for Google Chrome, but you can select any file as long as the domain is the local or dev URL. 
- You can try testing on Safari if you want, it would not show me any header responses until it suddenly did, not sure what I did wrong or right there, but it's the same as the Firefox steps.
- Code review, I added this to an existing <IfModule mod_headers.c> section instead of creating a new one, I assume that's acceptable?
- Check https://safety.dev.drupal.uiowa.edu/crime-log
- Check https://researchdsp.dev.drupal.uiowa.edu/contract-metrics, https://researchdsp.dev.drupal.uiowa.edu

Since this is a .htaccess change, we should hold off on merging until we've picked a day/time to release this.


joewhitsitt:
Sorry to make you run the patch process again, but I'd like a more detailed comment. Something like:

`# Declare HTTP Strict Transport Security (STS) header as recommended by Acquia. https://acquia.my.site.com/s/article/360004119254-How-To-enable-HSTS-for-your-Drupal-site`

joewhitsitt:
>Since this is a .htaccess change, we should hold off on merging until we've picked a day/time to release this.

I don't think this needs to be a scheduled release. In the docs I mentioned this type of change requires more attention, but I stopped short of adding it to the section requiring a special release coordination. As varnish for paths expire, the htaccess file will be called again so this will roll out slower. I am also optimistic that we are better protected by varnish this time around since the footer was still max-age 0 when the misconfigured redirect was sent out.

joewhitsitt (approved):

PR #8734, opened 2025-05-02
[Icons] Updated brand icons version

bspeare:
Updating the icons that were changed in https://github.com/uiowa/brand-icons/pull/38. 

# How to test

- Confirm commit hash matches latest commit in https://github.com/uiowa/brand-icons. 
- Delete `assets/icons/` directory
```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /
```
- Add a card, add a icon, search for removed icons.

### DIFF
--- a/docroot/themes/custom/uids_base/package.json
+++ b/docroot/themes/custom/uids_base/package.json
@@ -4,7 +4,7 @@
   "license": "ISC",
   "dependencies": {
     "@uiowa/uids4": "https://github.com/uiowa/uids.git#2276de0",
-    "@uiowa/brand-icons": "https://github.com/uiowa/brand-icons.git#4befd4a",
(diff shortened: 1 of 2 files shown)


pyrello:
Pulling to test

pyrello (approved):
Works as advertised. @bspeare and I discussed the possibility that needing to manually delete the icons might be a problem for deployment. Then I remembered that the deployment artifact only starts with the repo, so there isn't the problem of existing assets needing to be cleaned up.

pyrello:
> Works as advertised. @bspeare and I discussed the possibility that needing to manually delete the icons might be a problem for deployment. Then I remembered that the deployment artifact only starts with the repo, so there isn't the problem of existing assets needing to be cleaned up.

99% sure this is the case. Deploying this is going to be an experiment ;)

PR #8745, opened 2025-05-08
Layout clean up

pyrello:
#6478 attempted to resolve the two column layout bug by introducing new classes that didn't rely on core. That did not fix the issue so this PR reverts those changes. Additionally, the labels for our custom layouts were updated to make it easier to tell the difference with core layouts in the UI.

# How to test

```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli
```
- Check existing content to make sure that 2 columns layouts are still working.
- Add a new section to content and observe the new labels.
- Confirm that `SiteNow - Two column` is an available option.


bspeare (approved):
This looks good.  I tested existing and new two column layouts and I did not notice any change in functionality.

PR #8750, opened 2025-05-08
Theme Signage slideshows

### LINKED ISSUE #8738: Theme slides so they display in a slideshow when displayed on a sign
# Story

As a Signage user, I would like slides that rotate based on a set duration so that multiple slides will display as a slideshow.

# Additional Info

Need to choose module/library to provide this functionality.

https://www.drupal.org/project/splide was mentioned as a possibility in https://github.com/uiowa/uiowa/issues/8528#issuecomment-2843160267 and https://github.com/uiowa/uiowa/issues/8528#issuecomment-2843300900.

What are we using in v2?

Currently, users can set the duration the slide is visible prior to showing the next slide. It would be great if we could continue to have this ability.

# Proposed solution

pyrello:
Resolves #8738. Also introduces layouts that are specific to signage to allow us to have different LB style settings for signage than for pages.

# How to test

```
ddev composer install && ddev blt frontend && ddev blt ds --site=default  && ddev drush @default.local config-split:activate signage -y && ddev drush @default.local uli  /node/add/slide  
```
- Add a couple slides.
- Add a sign: https://default.uiowa.ddev.site/node/add/sign
- Add a slideshow and reference the slides. Save the layout.
- Confirm that the slideshow slides fade between each other and loop back to the beginning.
- Confirm all layouts (1-, 2-, and 3-column) are working with the sign.
- Confirm that no layout builder styles show for sign layout builder sections and that the `Display options` header also does not show.
- Confirm that `layout_builder_custom.overrides.js` gets loaded for all LB instances (nodes: page AND slide; fragments).
- Confirm that functionality to set cookie for tray width is still functional.
- Compare all section LB styles between local and production to ensure that none of those options have been lost for the page layout builder sections. 
- Confirm that the page LB container style is set to `Normal (default)` by default.


bspeare:
Slide implementation looks good. I was able to place multiple slideshow blocks on a page without any issues.  I wonder if the slideshow needs to be navigated by keyboard? What extent does this need to pass accessibility checks?  There was one issue reported with Siteimprove: 

> Page region missing an accessible name Any page section with the role of region should also have an accessible name to identify it. 

It seems to want the splide wrapper to have something specific for each slideshow block on the page like:
`aria-label="Custom slideshow title for slideshow block 1"`

Unrelated: looks like we have some `<h1>` and `<h2>` issues with the sign page name if you run the AInspector widget on Firefox.  

Also unrelated, I did notice that `layout_builder_custom.overrides.js` is not loading so we are not getting the custom drag handle for the slideshow block.

pyrello:
> Also unrelated, I did notice that `layout_builder_custom.overrides.js` is not loading so we are not getting the custom drag handle for the slideshow block.

Looks like this is because we are scoping where it gets added to just pages:
https://github.com/uiowa/uiowa/blob/515c454ba6d6e824897b7939f5da284d01bdb3d1/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L567-L568

docroot/modules/custom/sitenow_signage/assets/js/slideshow.js
```diff
@@ -0,0 +1,27 @@
...
+          speed: 1500,
+          slideFocus: false,
+          arrows: false,
+          pagination: false,
+          rewind: true,
+        }).mount();
+      })
+    },
+  }
+})(jQuery, Drupal);
```
GaryRidgway:
```suggestion
(function (Drupal) {
  Drupal.behaviors.signageSlideshow = {
    attach: function (context, settings) {
      context.querySelectorAll('.signage-slideshow').forEach(function (element) {
        console.log('adding splide', element);
        // Initialize Splide with the settings.
        new Splide(element, {
          autoplay: true,
          interval: 5000,
          type: 'fade',
          pauseOnHover: false,
          drag: false,
          speed: 1500,
          slideFocus: false,
          arrows: false,
          pagination: false,
          rewind: true,
        }).mount();
      })
    },
  }
})(Drupal);
```


To reduce use of JQuery, I think we can remove this?

GaryRidgway:
> > Also unrelated, I did notice that `layout_builder_custom.overrides.js` is not loading so we are not getting the custom drag handle for the slideshow block.
> 
> Looks like this is because we are scoping where it gets added to just pages:
> 
> https://github.com/uiowa/uiowa/blob/515c454ba6d6e824897b7939f5da284d01bdb3d1/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L567-L568

Should we just let LB have it everywhere? What was the reason to just have it on pages?

GaryRidgway:
> Slide implementation looks good. I was able to place multiple slideshow blocks on a page without any issues. I wonder if the slideshow needs to be navigated by keyboard? What extent does this need to pass accessibility checks? There was one issue reported with Siteimprove:

This is a good question. I'm not sure what else could be done for describing a slideshow. Considering they are so flexible, and that some of the images in them might be marked as decorative, adding something like a description field feels a bit odd. I feel like the slideshow might change so much that the description field might get outdated quickly.

~~Had we committed to any specific accessibility goals for signs besides contrast?~~

pyrello:
> > > Also unrelated, I did notice that `layout_builder_custom.overrides.js` is not loading so we are not getting the custom drag handle for the slideshow block.
> > 
> > 
> > Looks like this is because we are scoping where it gets added to just pages:
> > https://github.com/uiowa/uiowa/blob/515c454ba6d6e824897b7939f5da284d01bdb3d1/docroot/modules/custom/layout_builder_custom/layout_builder_custom.module#L567-L568
> 
> Should we just let LB have it everywhere? What was the reason to just have it on pages?

Nope. Just how we figured out how to do it. In the form alter, it required us to include forms IDs in the switch statement to add places it gets loaded. I found a different way to do it in the last commit.

Also added bullet points for testing this.

pyrello:
> > Slide implementation looks good. I was able to place multiple slideshow blocks on a page without any issues. I wonder if the slideshow needs to be navigated by keyboard? What extent does this need to pass accessibility checks? There was one issue reported with Siteimprove:
> 
> This is a good question. I'm not sure what else could be done for describing a slideshow. Considering they are so flexible, and that some of the images in them might be marked as decorative, adding something like a description field feels a bit odd. I feel like the slideshow might change so much that the description field might get outdated quickly.
> 
> Had we committed to any specific accessibility goals for signs besides contrast?

I think we need to discuss what our accessibility targets are for signs. As I wrote to @bspeare in slack:

> My thoughts about the accessibility portion of this is that digital signage is not an interactive application. Its also not accessible to non-sighted people. So, the UI for managing it should be accessible as should all UIs in SiteNow. But I don't think that the sign itself needs to be because it is intended for an application that is fundamentally not going to be accessible.
>
> Probably this is a discussion for the team though...

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -104,6 +41,24 @@ function sitenow_signage_preprocess_block(&$variables) {
...
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function sitenow_signage_preprocess_field(&$variables) {
+  if ($variables['element']['#field_name'] == 'field_slideshow_slides') {
+    // Attach the splide library.
+    $variables['#attached']['library'][] = 'sitenow_signage/slideshow';
+    $test = 'thing';
```
bspeare: Can we remove this?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -104,6 +41,24 @@ function sitenow_signage_preprocess_block(&$variables) {
...
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function sitenow_signage_preprocess_field__field_slideshow_slides(&$variables) {
+  $test = 'thing';
```
bspeare: And this?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
pyrello: ha.. good catch!

bspeare:
> Confirm that functionality to set cookie for tray width is still functional.

I don't think the width setting here is working.  Comparing the cookie on prod vs local, on prod the `ui_off_canvas_width` value in the cookie updates as soon as you move the tray.   Locally, that value only updates once the block is updated.

GaryRidgway (approved):
Looks good to me! I tested all parameters defined and everything seems to be working well! nicely done.

PR #8752, opened 2025-05-08
Remove "Add Views EntityReference filter to be available for all entity reference fields" patch

### LINKED ISSUE #6039: Investigate whether Views EntityReference patch is necessary
# Problem

We added a patch for "Add Views EntityReference filter to be available for all entity reference fields" at some point, I believe when we were working on getting our custom config entity for person types working. However, the patch does not do anything for config entities. It seems possible that we don't actually need this patch. If that is the case, we should remove it. 

# Proposed solution

Look at git blame to see when the patch got added. Review the associated PR to determine whether the patch adds functionality that is necessary for the rest of it to work. Consider possible ways to test the effects of removing the patch (e.g. does this affect how entity reference field widgets function or the display renders?). Test removing the patch. If you're convinced it's not necessary, put up a PR removing it.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/6039

# How to test

```
ddev composer install && ddev blt frontend && ddev drush @admissions.local sql-drop -y && ddev drush @admissions.local cr && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli areas-of-study
```

Test to make sure the filtering still work as https://admissions.uiowa.edu/areas-of-study does


pyrello (approved):
Looks good. Thanks for taking care of this!

PR #8758, opened 2025-05-12
Drupal OffCanvas Overrides - Disappearing fix

### LINKED ISSUE #8647: Off-canvas tray disappearing for Windows users
# Problem
As reported by Lydia F. and Kyle H., the off-canvas tray disappearing for Windows users. 

# Steps to reproduce

1. Connect to VPER-OSC-100.iowa.uiowa.edu 
2. Open Chrome and edit https://uiowa.edu/node/296/layout
3. Click around and open a few blocks then click on the blockquote block like the in video below. 
4. I was not able to recreate this using Browserstack.

https://github.com/user-attachments/assets/2a1b1232-bbeb-4773-9afb-7cc9cdd208ec

Image

It would be nice to rule out https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js as it doesn't seem to be setting the default tray width as seen in the screenshot above. 

# Proposed solution

- ?

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/8647

Context: https://github.com/uiowa/uiowa/pull/8758#discussion_r2087339602

# How to test

Since we haven't been able to reproduce the issue outside of PROD/Windows, I think testing should instead be for everything else just to make sure there aren't regressions caused by this.

```
ddev blt ds --site=uiowa.edu
```

Login and interact with layout builder. There should be no issues with the offcanvas tray placement and it should respect custom widths beyond 300 if the cookie is present.


docroot/modules/custom/layout_builder_custom/js/layout_builder_custom.overrides.js
```diff
@@ -37,6 +37,18 @@
...
+            const resizeData = {
+              data: {
+                $element: $element,
+                settings: settings,
+              }
+            };
+            setTimeout(() => {
+              Drupal.offCanvas.resetSize(resizeData);
+            }, 0);
+
```
GaryRidgway: What is currently occurring is that the top displacement for the tray is being miscalculated by the layout builder module because the styles for it are being injected directly onto the element when its being resized. What should really be happening is that they should be using the `:root{}` css variables that `Drupal.displace()` is updating, [as they will always be the most updated versions of what it is trying to do.](https://www.drupal.org/node/3275981)

I propose that instead of doing a javascript change, we instead add a line to `docroot/modules/custom/layout_builder_custom/css/layout_builder_custom.overrides.css` to simply ensure we are always using that offset value. (I think we should add default values to both variable usages as fallbacks as well).

```diff
#drupal-off-canvas-wrapper {
  overflow: visible;
  padding: 0px;
  margin-right: -1px;
+  width: var(--off-canvas-width, 300) !important;
+  top: var(--drupal-displace-offset-top, 0px) !important;
}
```

joewhitsitt: Nice idea @GaryRidgway. I'd be fine with rolling with this instead.

joewhitsitt:
I was able to semi-confirm the style fix handles the situation in PROD/Windows.

pyrello:
Pulling to test in non-windows environment.

pyrello (approved):
Poked around and didn't find any issues with LB.

Issue #8665, opened 2025-04-10
Spring dependency update marathon


joewhitsitt:
# Problem

We have been building up quite a queue of dependency updates and this was also noted in our courtesy audit from Acquia.

https://github.com/uiowa/uiowa/issues?q=state%3Aopen%20label%3Adependencies%20

Let groom together a sizeable chunk to work on.

## Dependencies
- [x] #8556 @pyrello 
- [x] #8769 @pyrello 
- [x] #8765  @pyrello 
- [x] #8768 @joewhitsitt 
- [x] #8775 @bspeare
- [x] #8776 @bspeare
- [x] #8762 @bspeare 
- [x] #8770 @joewhitsitt 
- [x] #8200 @bspeare
- [x] #8772 @pyrello
- [x] #8777 @bspeare
- [x] #8728 @bspeare
- [x] #8782 @bspeare
- [x] #8786 @bspeare
- [x] #8788 @bspeare
- [x] #8802 @bspeare
- [x] #8810 @bspeare
- [x] #8811 @bspeare
- [x] #7424 @bspeare
- [x] #8812 @bspeare
- [x] #8198 @bspeare
- [x] #8817 @bspeare
- [x] #8823  @pyrello 
- [x] #8798 @bspeare
- [x] #8824 @bspeare
- [x] #8818 @bspeare
- [x] ~#8805~ #8849 @bspeare
- [x] #8837 @pyrello 
- [ ] #8840 @pyrello 
- [x] #8852 @bspeare
- [ ] #8854 @pyrello 
- [x] #8853 @pyrello

briand44:
Grooming conversation... google_tag? was called out in the Acquia courtesy investigation. Not sure if we are currently insecure with the module/version we are on. Dependabot doesn't pick up all dependency updates. We get notifications on the uiowa application from Acquia about module updates available. Go to latest version of updates if possible and do those first will likely make updating to core 10.4 easier. If we run into issues or updates that seem too large, we can carve off follow up slices.

joewhitsitt:
Here are the items from the Acquia courtesy audit:

Project	Current version	Latest version	Release Date	Changes	Community support
- [x] [Acquia Connector (acquia_connector)](https://www.drupal.org/project/acquia_connector)	4.0.5	[4.1.0](https://www.drupal.org/project/acquia_connector/releases/4.1.0)	2024-12-16	New features, Bug fixes	Actively maintained, Under active development
[...]

PR #8765, opened 2025-05-14
acquia_connector 4.1.0

pyrello:
Related to #8665 

# How to test

Honestly not sure how to test this. As far as I can tell, what this module is doing is sending information about each site to our AC account. Probably we can just approve it and see if anything breaks when it goes to PROD.

No DB updates or config changes.

Changelog: https://git.drupalcode.org/project/acquia_connector/-/compare/4.0.5...4.1.0


joewhitsitt (approved):
>Probably we can just approve it and see if anything breaks when it goes to PROD.

I agree.

PR #8767, opened 2025-05-14
[Bug fix] Added default container setting for our page layout

bspeare:
This pr sets the default layout to "Normal" instead of "Narrow" because the container size was either never set or lost in an update. 

# How to test

```
ddev blt ds --site=default  && ddev drush @default.local uli  /node/add/page
```
1. Add a new page with some content
2. Edit layout and confirm the layout with the sidebar navigation has the "Normal" container setting


pyrello:
Pulling to test.

pyrello (approved):
Tested. Works as expected.

PR #8769, opened 2025-05-14
Updated to pwiki_pro 1.3.2

pyrello:
# How to test

I think code review is fine. Here is the changelog: https://git.drupalcode.org/project/piwik_pro/-/compare/1.3.1...1.3.2


joewhitsitt (approved):
that is a doozy of a changelog!

briand44:
Just hopping in here to mention that my understanding is that CCOM is not utilizing this yet, so we likely do not need to have them test anything related to this upgrade.

PR #8778, opened 2025-05-16
dependabot.yml update

pyrello:
While trying to look into why so many updates were not getting flagged by dependabot, I came across https://mglaman.dev/blog/my-dependabot-configuration-drupal-and-other-php-projects and thought it would be good to update our on configuration.

# How to test

Code review. This is guess and check work.


bspeare (approved):
Thanks for putting this together!

PR #8785, opened 2025-05-19
Updated postcss for gulp to 10.0.0

bspeare:
Trying this instead of https://github.com/uiowa/uiowa/pull/7424 because that pr is quite old. 

# How to test


bspeare:
I'd approve this if I could: No Percy regressions, no `blt frontend` issues, spot checked a site and it looked good.

PR #8788, opened 2025-05-19
Updated simple sitemap to 4.2.2

bspeare:
# How to test

1. Code review config changes occurred after re-saving all the options at `/admin/config/search/simplesitemap`.  The same config changes occur on `main` so not new to `4.2.2`.   Do we want these changes?
```
ddev composer install &&  ddev blt di --site=default && ddev drush @default.local ssg && ddev drush @default.local uli /sitemap.xml
```


GaryRidgway:
>  Do we want these changes?

I'd say so? I think it makes sense to re-bundle if its in main anyways. I will approve if my words alone make you comfortable with the update.

GaryRidgway (approved):

PR #8798, opened 2025-05-20
Updated media_library_theme_reset to 2.0.0-beta1

### LINKED ISSUE #6933: Remove or replace deprecated Media Library Theme Reset module
Discovered deprecation of [Media Library Theme Reset](https://www.drupal.org/project/media_library_theme_reset) module while working on #6901. I was still able to update to a d10 compatible version, but the module will be unsupported (but still available) after November 2023.

The recommended replacement is [Layout Builder iFrame Modal](https://www.drupal.org/project/layout_builder_iframe_modal), however that module changes much more than just the media library theme. 

# Proposed solution
Check if module is still needed. If so, investigate possible replacements and create our own if necessary.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6933. 

# How to test

```
ddev blt frontend && ddev composer install && ddev blt ds --site=sandbox.uiowa.edu  && ddev drush @sandbox.local uli  /node/1/layout
```
1. This update adjusts the styles for the media library that you see in layout builder only: https://github.com/uiowa/uiowa/pull/8798#issuecomment-2895164521
2. Edit a text area and click on the media icon to upload or insert media
3. Test paging through the images
4. Test Grid/Table view
5. Test searching for images
6. Test uploading an image
7. Test Audio upload
8. Test Facebook upload
9. Test UICapture upload
10. Test Remote video upload
11. Test static map image upload
12. Test Twitter
13. Confirm that the first screen and the second screen after upload look the same for each option
14. Compare against prod experience using `seven` theme
15. Compare against node edit screen using actual Claro styles (It's not a perfect match but they do look close)


bspeare:
It looks like this update removes custom styles from the seven theme that were being used.  I'm going to see if we can get the claro styles working on this. 

Screenshot 2025-05-20 at 11 36 56 AM

bspeare:
The `2.0` version of this drops support for `seven`, which is what our modal was using on the frontend when viewed in layout builder.  As mentioned on https://www.drupal.org/project/media_library_theme_reset, "The module provides only the styles needed for media library to function and look reasonably correct.
Some themes will need fixes in order to display correctly. Ideally theme problems should be fixed in the theme."  From the screenshot above it does look reasonably correct, however, I have added styles to make it look more like the claro modal that users will see when using the modal on the node edit screen.

cory-skeers:
I've gone through this in its different combos, and mostly the differences are minor.

One part that stood out to me was in the Table view it has Iowa gold in the header now.
image

At first I thought it was also larger and made it where the far right columns (Author and Updated) weren't immediately visible. But with more poking on prod, it looks like that already sometimes happens. Here I am now able to scroll and see those columns, on prod, I am not able to scroll horizontally to see them, so looks like a net improvement.

cory-skeers (approved):
I think this is good to go. Looks to introduce a bit of Iowa branding into the media gallery that could be taken out or left as is.

bspeare:
Thanks @cory-skeers! I removed the branding from the tables to make them look more like the Claro tables. 

Screenshot 2025-06-02 at 10 02 45 AM

PR #8807, opened 2025-05-22
Signage Hawk Alerts Display

### LINKED ISSUE #8742: Update display of Hawk Alerts on Signage
# Stories

As a Campus Safety unit, we need to be able to show Hawk Alerts on Digital Signs so we can continue to meet compliance.

As a Digital Signage sign owner, we would prefer Hawk Alerts don't take over the majority of our sign display so that folks viewing the sign can still see the sign content.

~As a Digital Signage sign owner, we would like active Hawk Alerts to display for no more than one hour even if they are active for a longer period of time.~

# Additional Info

Emergency endpoint - `https://emergency.uiowa.edu/api/active`

Should be checking endpoint every 30 seconds

Existing signage Hawk Alert module - `/sites/all/modules/custom/uiowa_signage_modules/signage_hawkalert`

# Proposed solution

The alert region is already rendering for sign nodes, so we just need to alter the styling of the alerts so that they overlay the top of the sign (using the dimensions of the header as our starting point).

e-marie-w:
Closes #8742 

# How to test

- Sync a site, enable the signage split, and set the site to use the stage endpoint:
 ```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local config-split:activate signage -y && ddev drush @sandbox.local cset uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active && ddev drush @sandbox.local cr && ddev drush @sandbox.local uli
```
- Create a sign. See that the alert appears above the content when viewing a sign while logged in.
- Open the same sign in a private browser window. See that the alert is now located at the bottom of the page, over the content (including the back to top arrow if present)
- Back in your logged in tab, set a custom alert at https://sandbox.uiowa.ddev.site/admin/config/sitenow/uiowa-alerts. Note the changed text about custom alerts not rendering on signs.
- Save and check the sign to see the local alert appearing below the hawk alert, with spacing between them. Check the private tab sign (may need to hard reload the page) and see that the local alert does not appear on the sign.
- Bonus points: sync down a different site, do not turn on the signage split, and see that the description text for custom alerts at /admin/config/sitenow/uiowa-alerts does not mention anything about signs.

### CODE REVIEW

**Comment by e-marie-w** [comment]
Putting this back into draft while I figure out how to hide the local alerts.

File: docroot/modules/custom/sitenow_signage/sass/alertsblock.scss [hunk 1]
```diff
@@ -0,0 +1,15 @@
+.uiowa-alerts-wrapper {
```
pyrello: It doesn't seem like we actually need to have an SCSS file being compiled here. The extra complication doesn't seem worth it for the nesting on this very small file.

docroot/modules/custom/sitenow_signage/sitenow_signage.libraries.yml
```diff
@@ -24,3 +24,9 @@ slideshow:
     assets/js/slideshow.js: { preprocess: false }
   dependencies:
     - sitenow_signage/splide
+
+# Hawk Alerts
+alerts-block:
```
pyrello: I think we should not include the word `block` in the name. Since the CSS is not targeting a block nor are we using `HOOK_preprocess_block` to attach the library. Maybe `signage-alerts`?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -58,6 +67,9 @@ function sitenow_signage_preprocess_node(&$variables) {
   }
   if ($variables['node']->getType() == 'sign') {
     $variables['attributes']['class'][] = 'bg--black';
+    if ($variables['view_mode'] == 'full' && (\Drupal::currentUser()->isAnonymous())) {
+      $variables['#attached']['library'][] = 'sitenow_signage/alerts-block';
```
pyrello: Reminder to change this to whatever your library gets re-named to.

docroot/modules/custom/sitenow_signage/sass/alertsblock.scss
e-marie-w: I didn't know SCSS caused extra complication, I can switch it back to being CSS!

joewhitsitt:
pulling to test

joewhitsitt (approved):
Looks fine to me.

>see the local alert appearing below the hawk alert, with spacing between them

I remember there being some talk about the spacing. Leaving it in as-is? `.region-alert .alert`

e-marie-w:
> > see the local alert appearing below the hawk alert, with spacing between them
> 
> I remember there being some talk about the spacing. Leaving it in as-is? `.region-alert .alert`

I was going to for this, the alerts seemed to have top and bottom margins added so I just set the bottom margin to 0 along with the left and right, just in case they ever change.

PR #8820, opened 2025-05-27
People list block exposed filters

### LINKED ISSUE #8568: Exposed filters for People list block
# Story

As a SiteNow v3 site owner, I would like to be able to expose filters on my people list block so that users can search for a person or filter by person type, research areas or tags.

# Additional Info

Nursing has asked for this most recently.

Option to expose search field that searches first and last name (title?) and filters results.
Option to expose person type field (select list)
Option to expose research areas field (select list)
Option to expose tags as a filter - can this be a select list?

# Discussion

Nursing asked for an a-z filter on the people list block as well but this doesn't seem like it would be as valuable for the service as the others. Would like feedback from others about this when we groom.

# Proposed solution

- [x] Update views config to switch filter methods
- [x] Write update hook to convert existing filter values
- [x] Remove old filter values once confirmed the update hook is working as expected.
- [x] Update listblock lb config form person type input (confirm this is only being used for the listblock)
- [x] Convert exposed person type filter (string equals) to dropdown
- [x] Check feature and site splits for custom people block filters and adjust update hook.

cory-skeers:
Resolves #8568 
Touching https://github.com/uiowa/uiowa/issues/7101

# How to test

```
ddev composer install && ddev blt frontend
```

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli node/2876/layout
```

- Block displays should match PROD and block headlines. Tempstore block at the top should also match PROD.
- Configuring the block should reveal the "expose this block" and match the set filter values from PROD.
- The person type field should be a dropdown with the person types available on the website.
- The person status field should not allow exposing to users.
- By default, all filters should be disabled from being exposed.
- Only admins can see the person type status field (e.g. Former Staff)
- Once exposed, the filters should operate and look as expected.
- Confirm https://sandbox.uiowa.ddev.site/people displays the person type filter as a dropdown and works as expected. 
  - Other customizations should remain (title, header section).
  - Configure the sitenow people settings and turn off all of the filters. Visit the page after save. The exposed form should not be visible.
- https://sandbox.uiowa.ddev.site/people-directory should show similar results as PROD (logged in).

```
ddev blt ds --site=cs.uiowa.edu && ddev drush @cs.local uli people
```

- Confirm the person type dropdown works
- From the top menu, visit Faculty, Staff, etc. pages to see that they start with the same people.

```
ddev blt ds --site=default
```

Compare to https://default.prod.drupal.uiowa.edu/people to confirm that the person type filter is not exposed.


joewhitsitt:
I believe the person type status field is being converted correctly, but there is an issue with the form config display and how the value is saved going forward. There is more sitenow_people code still to revisit surrounding this.

joewhitsitt:
Taking a screenshot and database backup of sandbox.prod before the 10.4 deployment goes out. I have a feeling it breaks certain person type functionality and is why I started to have issues after 2pm when I merged main into this PR.

PROD:
Screenshot 2025-06-10 at 7 07 40 PM

Drupal 10.4 (main)
Screenshot 2025-06-10 at 7 12 48 PM
Screenshot 2025-06-10 at 7 13 10 PM

I think it wipes the value for person type where configured and fails to filter on the string where it is still called for render.

pyrello (approved):
Works as expected. Thanks for all the effort on this and for catching the issues with #8853

PR #8823, opened 2025-05-28
Update smart_date to 4.2.3

pyrello:
Related to #8665 

The PR removes patches for the following issues as they were fixed and included in the current version:
- https://www.drupal.org/project/smart_date/issues/3452796
- https://www.drupal.org/project/smart_date/issues/3463447

Changelog: https://git.drupalcode.org/project/smart_date/-/compare/4.1.3...4.2.3 (no update hook, some config changes)

It is possible to unset the option to have the end date hidden and I've made that change for all date widgets where it wasn't already unset.

There was also the introduction of a new option to have a drop-down of suggested durations for the time widget. I have not enabled this new option as the functionality is pretty limited and IMO not very useful yet.

# How to test
The changes seem to primarily affect how the field widget functions. I wouldn't expect the storage or display of the dates to be affected. I think testing can be primarily focused on confirming that the widget works as expected.

```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli /
```
- Confirm date range selection in Events block works as expected.

```
ddev drush @sandbox.local config-split:activate event -y
```
- Test When field on Event content type.
```
ddev drush @sandbox.local config-split:deactivate event -y && ddev drush @sandbox.local config-split:activate sitenow_alerts -y
```
- Test Dates affected field on Alert content type.
```
ddev drush @sandbox.local config-split:deactivate sitenow_alerts -y && ddev drush @sandbox.local config-split:activate thesis_defense -y
```
- Test the Defense Date field on the Thesis Defense content type.
```
```
- Test the When field on the Event content type. It should be disabled still.
```
ddev blt ds --site now.uiowa.edu
```
- Test the Original Published Date field on the Article content type.


pyrello:
I'm taking a look at the Registrar site customizations which include a smart date field addition to the media file entity type. We have some code that is setting the field to show/hide based on which File Archive Type (term reference field) is selected:
https://github.com/uiowa/uiowa/blob/e9fdb7961998f9b6a1c70ba43cb0a9254e6c9403/docroot/sites/registrar.uiowa.edu/modules/registrar_core/registrar_core.module#L86-L95
We have hard-coded IDs for terms and none of our hard-coded term IDs match the actual terms that exist now. So the field will never show. Wondering if we need to do something to fix this? Or maybe this functionality isn't necessary anymore?

briand44:
I reached out to Josh related to the above question since I wasn't seeing any use of that functionality on the new site.

> with the upcoming accessibility changes coming down the road we've kinda revisited our archiving policy and partnership with the Iowa Digital Library and don't think we'll be utilizing the archive module much (or at all) on the new site. We'll be displaying the current data set and maybe a year or two but not the extensive dropdown list you all built for us.
> 
> If it's causing issues please don't spend any time on fixing it now and if it needs to be temporarily decommissioned that shouldn't impact our launch. We may want to revisit at some point but do whatever you all think is best at this time.

pyrello:
> I reached out to Josh related to the above question since I wasn't seeing any use of that functionality on the new site.
> 
> > with the upcoming accessibility changes coming down the road we've kinda revisited our archiving policy and partnership with the Iowa Digital Library and don't think we'll be utilizing the archive module much (or at all) on the new site. We'll be displaying the current data set and maybe a year or two but not the extensive dropdown list you all built for us.
> > If it's causing issues please don't spend any time on fixing it now and if it needs to be temporarily decommissioned that shouldn't impact our launch. We may want to revisit at some point but do whatever you all think is best at this time.

Since it is basically non-functional at this point, I'm not going to include it in the smart_date updates. If we invest in fixing it at some point in the future, we'll want to re-save the form display settings too. If it ends up getting removed, then nothing needs to be done other than removing the field and custom code.

joewhitsitt:
pulling to test

joewhitsitt:
seems unrelated to the update but noticed that the events block date range defaults aren't "now" as configured. on glance, i am not sure where it is getting this date but it is the same as PROD. just taking a note for now.

Screenshot 2025-05-30 at 10 06 27 AM

joewhitsitt:
For ITS, do we need to revisit what we are doing here? End date and time display all the time now. docroot/sites/its.uiowa.edu/modules/its_core/js/smartdate-override.js

**Local**
Screenshot 2025-05-30 at 10 24 33 AM

**PROD**
Screenshot 2025-05-30 at 10 25 17 AM

We are still good for the output, but I know they were particular about the form.

**Ongoing**
Screenshot 2025-05-30 at 10 30 38 AM
**End date**
Screenshot 2025-05-30 at 10 31 01 AM

joewhitsitt (approved):
I think this is looking good, though I wonder if ITS notices and reaches out about the form ux diff.

Also tested grad, facilities directly.

briand44:
I think we should take a look at ITS and see if we can continue to support the current form overrides.

pyrello:
I'll take a look at the ITS stuff. Thanks for testing @joewhitsitt!

pyrello:
I put in place a JS-based solution to show/hide the end date correctly for ITS. I think there is a bug in the smart date JS but I wasn't able to figure out exactly what it was in a way that would let me report it. My solution copies code from the smart date JS because there is no way to access it directly from our custom JS :(

joewhitsitt:
✅ 👍 checked again. this solves the ITS issue I experienced and matches what is in prod in terms of ux.

Issue #8275, opened 2024-11-21
uiowa cron/log rotate


briand44:
This app has not yet been scheduled for upgrade. Once it is upgraded, we should update the cron job and add the schedule job to rotate the logs similar to the other apps that have already been upgraded.

>Remove cron jobs on DEV/STAGE. Copy paste the cron job from uiowa07 and update the cron job, modifying the filename. Copying the rotate /shared/logs job from uiowa07 and scheduling it weekly on sundays.

briand44:
blocked until this app is upgraded to Cloud Next

joewhitsitt:
>Your application will be proactively scheduled to upgrade to Cloud Next infrastructure by 31 May 2025.

🤷

briand44:
It has been scheduled for upgrade:
Maintenance Start: Thursday, May 29, 2025 - 03:01 UTC
Maintenance End: Thursday, May 29, 2025 - 11:00 UTC

joewhitsitt:
dev/stage cron was already disabled.

prod has been updated. also updated the brand and classroom logging.

when [#8825](https://github.com/uiowa/uiowa/pull/8825) is merged and deployed. all apps should be checked on to clean up directories

until then, I think this is done.

joewhitsitt:
i think the updated logrotate stuff will work, but I need to change each schedule job to write the `logrotate.status` file to a directory where we have permissions. i'll probably switch it to /shared/logs. apparently where it was writing to before got locked down and so our last successful rotation was end of April.

```
error: error creating stub state file /var/www/html/uiowa09.prod/logrotate.status: Read-only file system
```

pair that with ssh'n in to each app and cleaning things up.

joewhitsitt:
prod is done. I need to loop back through tomorrow for dev/stage (all apps) i think.

joewhitsitt:
`rotate /shared/logs` has been setup/reconfigured for all apps, all envs.

ssh'ing into each app/env to clean up the shared/logs directory and initially set the logrotate.status file.

joewhitsitt:
establishing an ssh to each env took the longest. this is done.

i will like to monitor the parallel logging to make sure the recent changes allow for multiple deployments in a day to be properly rotated.

PR #8825, opened 2025-05-28
Logrotate revisited

joewhitsitt:
After determining /shared/logs is _not_ shared by the apps, the app addition to the log name was causing issues with the generalized rotate configuration.

switching to daily even though the job only fires weekly to ensure they are rotated into daily files. multiple deployments in a day should get a .1.gz type add-on.

delaycompress - my understanding is that in those situations of timezone offsets and active file writing, the log rotation will still work.

# How to test

Like past situations, It is easiest to treat this like a guess and check in prod type change.

# Follow-up
- Clean up current mess of files that isn't being properly rotated for each app.
- Check back in a bit to see that it is working


joewhitsitt:
Would be nice to coordinate with #8275 😁

pyrello (approved):
Guess and check and approved!

PR #8837, opened 2025-06-02
Update to views_bulk_operations and moderated_content_bulk_publish

pyrello:
Related to #8665 

VBO changelog: https://git.drupalcode.org/project/views_bulk_operations/-/compare/4.2.5...4.3.4
MCBP changelog: https://git.drupalcode.org/project/moderated_content_bulk_publish/-/compare/2.0.29...3.0.0-beta17

# How to test

```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli admin/content
```
- Test that publish and unpublish bulk actions work as expected.

## ITS
```
ddev blt ds --site its.uiowa.edu && ddev drush @its.local uli admin/content
```
- Test that publish and unpublish bulk actions work as expected with additional content types.


pyrello:
The only place that I see we are using VBO is in our fragments list view. We do no have any of the VBO actions selected, so effectively it is not being used there either. 

If we are not actually using this anywhere, maybe we should remove it?

pyrello:
> The only place that I see we are using VBO is in our fragments list view. We do no have any of the VBO actions selected, so effectively it is not being used there either.
> 
> If we are not actually using this anywhere, maybe we should remove it?

Okay, I was wrong. It appears this is a dependency of `moderated_content_bulk_publish`.

joewhitsitt:
pulling to test

joewhitsitt (approved):
Checked that the scheduling and iowanow embargo functionality still works in rolling things back if publish is premature.

the config changes look right to me, but I was getting a diff after `cex` locally. no issues on import though.

PR #8852, opened 2025-06-03
Google tag manager 2.0.9

### LINKED ISSUE #7186: Google Tag Settings Default - Insert on 403/404
Back in https://github.com/uiowa/uiowa/blob/7580330b06b00096e2ba264041adacbe0d30effc/config/default/google_tag.settings.yml we set 404s to not insert the google tag container.

Background: https://iowaweb.slack.com/archives/C0164FNN1PV/p1701889194747959

We have had a request to change this default setting back so that 403/404 are not specified when creating new containers.

This is a configurable setting by webmasters, but can be missed easily when setting up a new container.

### LINKED ISSUE #8485: Upgrade google_tag to 2.x or switch to gtm
Relates to #7538 

There are two branches of `google_tag`, and we are currently on 1.8.x. 2.x is intended to be used _without_ `google_analytics`, and introduces potentially breaking changes in relation to how Google's consent mode is handled.

>stopped all data going into Google Analytics. Starting with release 2.0.3 of that module, there is a feature to opt-in to Google's consent mode, which is enabled by default.
https://www.drupal.org/project/google_tag/issues/3180185

>With that feature enabled, we get zero data in GA, presumably because we are not using a consent management platform (CMP) or otherwise sending any consent "signals" to Google.

>There is also a bug that makes the button not show that it's disabled. Here's a link to the patch to fix that little issue: https://www.drupal.org/project/google_tag/issues/3424623

As this branch has been in development, it has led to stories of complete stoppage of flow into GA and tracking platforms (https://www.drupal.org/project/google_tag/issues/3414894#comment-15924046, Drupal slack convos).

https://www.drupal.org/project/gtm has been mentioned as a simpler alternative, but would need full investigation and testing.

# Proposed solution

### LINKED ISSUE #8912: Can't remove GA property id anymore
# Problem

If a SiteNow webmaster has previously configured GA for their website and decide they want to remove it for some reason (switching to GTM for analytics) they are unable to because it is a required field.

Sounds like we had a patch we were applying at one point that fixed this issue.

For Journalism, I plan to check the checkbox under core settings so the tracking code won't load on the site. If we resolve this, we probably want to remove the property id from Journalism and enable GA loading through the core settings again so we don't need to be involved if they decide to switch back in the future.

# Steps to reproduce

Go to https://journalism.uiowa.edu/admin/config/services/google-analytics and try removing the web property id and save the page.

Image

# Proposed solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8485. 
Resolves https://github.com/uiowa/uiowa/issues/8912 (by removing)
Resolves https://github.com/uiowa/uiowa/issues/7186 (by no longer setting as default, just paths)

User Impact: As a website manager I can see that my google analytics configuration is migrated over and that google tag is now on and the interface I use to manage both google analytics and google tag manager.

# How to test

## Setup

```
ddev drush @sandbox.local sql-drop -y && ddev drush @sandbox.local cr && ddev drush sql-sync @sandbox.prod @sandbox.local -y
```

You should see pending updates:
`ddev drush @sandbox.local updatedb-status`

`ddev drush @sandbox.local config:set uiowa_core.settings uiowa_core.ga 0`
`ddev drush @sandbox.local config:set uiowa_core.settings uiowa_core.gtag 0`
`ddev drush @sandbox.local config:set google_tag.container.chat_bot_test_container status 0`

Verify:

`ddev drush @sandbox.local config:get uiowa_core.settings uiowa_core.ga`
`ddev drush @sandbox.local config:get google_tag.container.chat_bot_test_container status`
`ddev drush @sandbox.local config:get google_tag.container.gtm status`

Should be ga=0, chat_bot_test_container=false, gtm=true.

`ddev drush @sandbox.local updb -y`
`ddev drush @sandbox.local cim`
`ddev drush @sandbox.local cr`

### GA migration

Since GA was setup on prod but ga switched to zero, the GA is still migrated but disabled in the google_tag bucket.

`ddev drush @sandbox.local sql-query "SELECT name FROM config WHERE name LIKE 'google_tag.container.G-%';"`
`ddev drush @sandbox.local config:get google_tag.container.{MACHINE_NAME}`
`ddev drush @sandbox.local config:get google_analytics.settings account`

New "ga" container has real tag_container_ids/settings (e.g. dimensions_metrics: type: dimension name: custom-dimension value: dimension-value and pages includes /my-custom-path and anonymous is in the roles section) populated, raw status: true. google_analytics.settings:account is now ''.

### Existing tag, disabled before update

`ddev drush @sandbox.local config:get google_tag.container.chat_bot_test_container`

Should be status: false, but tag_container_ids, settings exist.

### Existing tag, enabled, blocked.

`ddev drush @sandbox.local config:get google_tag.container.gtm`

Raw status: true, tag_container_ids: [GTM-52S96F8] populated.

### Check backend UI:

`ddev drush @sandbox.local uli admin/config/services/google-tag/containers`

Should see three containers, one disabled, two inactive overwritten.

In an annoymous browser, go to https://sandbox.uiowa.ddev.site/ and see that there are no google scripts running (e.g. ctrl+f `gtag`) in the inspector.

`ddev drush @sandbox.local config:set uiowa_core.settings uiowa_core.gtag 1 && ddev drush @sandbox.local cr`

Refresh the anonymous browser and see instances of `gtag` in the inspector

Go back to https://sandbox.uiowa.ddev.site/admin/config/services/google-tag/containers and see that the two inactive overwritten are now active.

`ddev drush @sandbox.local config:set uiowa_core.settings uiowa_core.gtag 0`

Go back to https://sandbox.uiowa.ddev.site/admin/config/services/google-tag/containers and see at least one inactive (overwritten). Try to enable one of the disabled containers. It is switched to status true but it is still inactive (overwritten) based on the previous config:set of gtag.

## Fresh install

`ddev blt di --site=default`

Google tag and use collections should be on by default. As a webmaster (create and masquerade) you should be able to go to admin/config/services/google-tag/containers and configure a container.

See that path exclusions are populated by default.

## UI Changes

  - /admin/config/services — one "Google Tag" link, going straight to /admin/config/services/google-tag/containers.
  - Container edit form — tab reads "Tags", not "Additional Tags".
  - As a non-admin role, visiting /admin/config/services/google-tag directly should be access-denied.
  - Message displays above admin/config/services/google-tag/containers when gtag is 0. Can't change the "inactive (overwritten) message without extending the interface class.

## Inline Entity Form

- Just a RC to stable upgrade that fixes an issue with one of the google_tag update hooks.

# Before merge

1. Restore the non-prod check in GoogleTagOverride.php
2. Un-skip testContainersOverriddenInNonProd in GoogleTagOverrideTest.php
3. Coordinate scheduled deployment alongside https://github.com/uiowa/uiowa/issues/9974

# Follow-up

- https://github.com/uiowa/uiowa/issues/9980 Slice to uninstall google_analytics update sitenow.services.yml to point the service directly at GoogleTagOverride, drop the GoogleAnalyticsOverride shim class.

# Follow-up to the follow-up

- Remove google_analytics from composer.json. https://github.com/uiowa/uiowa/issues/9983
- Update \Uiowa\Blt\Plugin\Commands\ReportCommands::analytics or remove as part of #9973


joewhitsitt:
Deploying to dev for testing

joewhitsitt:
Small hiccup with the sandbox test deployment. The update hook was not aware of the the site and marked all the containers as inactive (migrated but not enabled). The analytics settings got migrated over. Since that was just a temporary piece, I am going to remove that and just leave the temp local testing piece in place for now.

bewoldt:
I was able to complete all the testing steps.
-----------
Notes
after installing a fresh install `ddev blt di --site=default` I created an account setup as webmaster and masqueraded

I was able to add a container but once I saved I am directed to /admin/config/services/google-tag with a message of 

Access Denied
Status message
The configuration options have been saved.

You are not authorized to access this page.

When I go back to Google Tag containers I see that the tag as added and can edit as normal. Saving again brings me back to the Google Tag container page as expected. I only get the Access Denied after initial creation.

-----------
notes

When I tried to edit the weight of a container tag it did not save the weight I entered.

-----------
notes

edge case but happened to me while testing
If i enter a tag ID and click save without clicking outside the input I receive an error. I'm guessing that clicking outside the input processes the id to check validation.

joewhitsitt:
Thanks for the testing @bewoldt. I was so focused on the migration and lock down that I didn't really test the form UI itself. I will take a look at these

joewhitsitt:
@bewoldt fixes have been made for the issues you detailed.

bewoldt:
I hadn't tried updating the weights when using multiple ids with one container. I see that works as expected. I still can't update weights while having multiple containers.

joewhitsitt:
Ah okay. I believe that column is a carry-over from the old version. The field is no longer visible on the form. We are using a patch just to add the multiple containers functionality back and there is [this issue](https://www.drupal.org/project/google_tag/issues/3420714) where we could get that working again, but would like for the multiple containers patch to merge in first.

I just set the weights on sandbox.prod and synced down. The weights are preserved (good if this is supported later), but I do not see them actually having an effect on which one is loaded into the DOM first. https://sandbox.prod.drupal.uiowa.edu/ (i would expect with a weight of 2 that 52S96F8 would be first in the DOM over NSSJ7CWL which has a weight of 5.

Screenshot 2026-07-20 at 3 13 01 PM

bewoldt (approved):
Approved.

joewhitsitt:
Thanks @bewoldt! I will uncomment the prod checks and get this ready for merge. setting as draft to avoid early merge.

joewhitsitt:
Just going to say this now. This is output _before_ update hooks are run and does not indicate failure.

>[error]  In order to use Google Tag 2.x, you must decide the upgrade path between Google Tag 1.x and Google Analytics. (Currently using Google Tag Google Tag 2.x is incompatible with Google Analytics while upgrading from 1.x.)

PR #8802, opened 2025-05-21
Updated fontawesome to 3.0.0

bspeare:
# How to test

This release is mostly for Drupal 11: https://www.drupal.org/project/fontawesome/releases/3.0.0. 
```
ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /icons
```
1. Test that icons are rendering correctly using the `<span>` tag on https://sandbox.uiowa.ddev.site/icons. 
2. Masquerade as webmaster and confirm that you can adjust icons on https://sandbox.uiowa.ddev.site/timeline-test and that you don't see advanced options. 
3. Test that you can adjust event block at https://sandbox.uiowa.ddev.site/event-block. 
4. Resave https://sandbox.uiowa.ddev.site/admin/config/content/fontawesome and confirm `drush @sandbox.local cex` contains no config changes.

PR #8863, opened 2025-06-04
[Hotfix] Added patch to get fontawesome working again

bspeare:
Missed this in https://github.com/uiowa/uiowa/pull/8802 because it is only happening with some icons. 

# How to test

```
ddev composer install && ddev blt ds --site=uiowa.edu && ddev drush @home.local uli /node/1/layout
```
1. Resave the event blocks on uiowa.edu and confirm that icons are rendering as expected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -277,6 +277,9 @@
             "drupal/field_delimiter": {
                 "Doesn't work in layout builder": "https://www.drupal.org/files/issues/2020-04-08/field_delimiter-3099580-7.patch"
             },
+            "drupal/fontawesome": {
+                "[3380581] Iconpicker does not render classes appropriately": "https://www.drupal.org/files/issues/2025-03-06/iconpicker-does-not-render_3380581_11.patch"
+            },
             "drupal/fragments": {
                 "remove_views_data_type": "https://www.drupal.org/files/issues/2022-08-24/3305851-remove_views_data_type-6.patch"
             },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "05424e3dc765d867df89938782fb2d88",
+    "content-hash": "d6981bf3047f0ce10766bd096d2d2dae",
     "packages": [
         {
             "name": "acquia/blt",


pyrello:
Pulling to test.

pyrello:
For this update to go into effect are folks going to need to re-save any entities with icon picker fields? Should we add an update hook to re-save the entities for them?

bspeare:
Good question.  This regression has only been out there for a week if a user created/edited a block or content type using this field so I'm wondering if we want to add the update hook?

bspeare:
Looks like the event block and the timeline block are the main blocks using this.  Classrooms taxonomy fields is using it as well: https://github.com/search?q=repo%3Auiowa%2Fuiowa+type%3A+fontawesome_icon&type=code.   Maybe we add an update hook for the two blocks?

pyrello:
> Looks like the event block and the timeline block are the main blocks using this. Classrooms taxonomy fields is using it as well: https://github.com/search?q=repo%3Auiowa%2Fuiowa+type%3A+fontawesome_icon&type=code. Maybe we add an update hook for the two blocks?

Trying to think through how we would handle this if we didn't include an update hook. Post on the SiteNow site? If we aren't getting a lot of reports, then maybe we shouldn't worry too much about it.

pyrello:
> > Looks like the event block and the timeline block are the main blocks using this. Classrooms taxonomy fields is using it as well: https://github.com/search?q=repo%3Auiowa%2Fuiowa+type%3A+fontawesome_icon&type=code. Maybe we add an update hook for the two blocks?
> 
> Trying to think through how we would handle this if we didn't include an update hook. Post on the SiteNow site? If we aren't getting a lot of reports, then maybe we shouldn't worry too much about it.

Actually, I kind of wonder if clearing cache would be enough... it doesn't actually make sense to me why it would be necessary to re-save.

pyrello:
> > > Looks like the event block and the timeline block are the main blocks using this. Classrooms taxonomy fields is using it as well: https://github.com/search?q=repo%3Auiowa%2Fuiowa+type%3A+fontawesome_icon&type=code. Maybe we add an update hook for the two blocks?
> > 
> > 
> > Trying to think through how we would handle this if we didn't include an update hook. Post on the SiteNow site? If we aren't getting a lot of reports, then maybe we shouldn't worry too much about it.
> 
> Actually, I kind of wonder if clearing cache would be enough... it doesn't actually make sense to me why it would be necessary to re-save.

Unfortunately, cache rebuild did not fix it. I'm still leaning towards not including an update hook.

pyrello (approved):
This doesn't fix the issue for instances of blocks that were created since the `fontawesome` update went out. But it prevents future instances and doesn't prevent us from doing an update hook to resolve it in the future.

PR #8870, opened 2025-06-05
Updated lb_direct_add to 2.0.1

pyrello:
Related to #8665 

# How to test
- Code review.
- Confirm that the direct add button works as expected.
- Confirm that roles that currently are able to access the direct add button can see the More options link.


joewhitsitt:
pulling to test

joewhitsitt (approved):
looks fine to me. ran it as webmaster, still had more access. confirmed the effects of the patch were still working.

Issue #8796, opened 2025-05-20
Override URL pattern for articles on iowaprotocols


briand44:
# Story

As a site owner for iowaprotocols website, we would like the url alias pattern for articles to not have the Year/Month so that we can more easily setup redirects.

# Additional Info

They would like the pattern to be `protocols/[node:title]`.

https://github.com/uiowa/uiowa/issues/8760#issuecomment-2887094858

Currently that config is ignored but if someone saves the articles configuration form, the alias pattern will update to include the Year/Month` again.

Site split or change to how the article config form works? Opt-out of `Year/Month` pattern?

# Proposed solution

- [x] Move forward with solution to allow a different pattern for this website.
- [x] Run `ddev drush @medicineiowaprotocols.prod entity:delete redirect` post alias update since redirects are automatically created.

joewhitsitt:
I talked briefly with Brian about this earlier while preparing to take on the migration tweaks task. For a site-split, it isn't just a simple config diff as this config is already ignored but overridden by the form on save. The config form now contains other items which the site might want to save (related articles, articles on people, etc.) so would need to go beyond just disabling the form on that site.

It seemed more likely if we were to continue to keep this functionality tied to the config form, that an opt-out separate year/month checkboxes might make more sense. This could be an admin only editable set of fields. A concern is that I believe breadcrumbs, the monthly archive block attempt to tie this all back together and so there could be dead end URL path slugs or unintentional behavior.

briand44:
Grooming conversation... The redirect rule is not in scope for this work. That will initially live on the D7 CCOM application and likely move when that site launches. Opt-out option on articles configuration page to opt out of the year/month pattern could add value for others in the future. This could be just an admin level setting. We have some theme settings like this currently. Help text would be useful here. Is there a use case where someone would want the year slug but not month? Let's start with one option to opt-out of having both Year/Month in the pattern. Would be good to check with Jeremy when we pick this up to see if there is interest in pairing.

cory-skeers:
Emailed the ccom team to let them know; this should be wrapped up now.

PR #8874, opened 2025-06-06
[Articles] Add option to remove year/month slug

cory-skeers:
Relates to #8796 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /admin/config/sitenow/uiowa-articles
```

Check out the new "remove slug" option and its help text.
When the new option is checked, see that Article paths include only the "path" variable, and no year/month slug (https://sandbox.uiowa.ddev.site/news)
Check that the option is not available when masquerading as a non-admin, but that you can still make other changes to the form
Compare the rearranged form with its current iteration in prod, and provide feedback/suggestions: 
 - Path has been moved up to Article Settings alongside this new option
 - Help texts have been updated for Path, Remove Year/Month Slug, and Enable Articles Listing options


bspeare:
Should this option be listed under "Article Settings" instead of "View Page Settings", or a separate section?  I found it a little confusing under the "View Page Settings" section.

bspeare (approved):
This works great! I tested:


- The new "remove slug" option and its help text
- Confirmed when option is checked, the Article paths include only the "path" variable, and no year/month slug (https://sandbox.uiowa.ddev.site/news)
- Confirmed that url redirects are made when slug is toggled on/off
- Confirmed that Article pattern changes at `/admin/config/search/path/patterns` based on toggle select
- Confirmed that the option is not available when masquerading as a non-admin, but that I could still make other changes to the form
- I like the rearranged form and the new help text added to "Enable article listing" checkbox

PR #8878, opened 2025-06-09
[UIDS] Table adjustments for use in accordions

### LINKED ISSUE #990: Unordered/ordered lists within tables in accordions have a different font size than other cells

### LINKED ISSUE #8809: Issue with edited table cell values having a `<p>` tag added to them
> https://transportation.uiowa.edu/cambus/routes/34-research-park#accordion-item-24426-0
>  
> Here is an example. I am not 100% sure the exact steps our team member took, but it sounds like they pasted new text and center aligned.
> 
> Image

I did a little digging. The difference in how the cells look seems to be because some of them have a class that centers the text, some do not, and that class is either applied to the `<td>` element or a `<p>` element inside it wrapping the text. When I tried using the center alignment button on a cell, I noticed it added a `<p class="text-align-center">` tag around the text.

bspeare:
Resolves https://github.com/uiowa/uids/issues/990. 
Resolves https://github.com/uiowa/uiowa/issues/8809. 

# How to test
- Review and approve https://github.com/uiowa/uids/pull/983. 

```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /list-item-line-height
```

```
ddev blt frontend && ddev blt ds --site=transportation.uiowa.edu && ddev drush @transportation.local uli /cambus/routes/34-research-park#accordion-item-24426-0
```

1. Open the accordion on the page and compare it to https://sandbox.prod.drupal.uiowa.edu/list-item-line-height and confirm that table is rendering all in the same font size. 
2. This pr also includes minor update to line-height requested from Brand team and decreases `<li>` line-height from 1.7 to 1.6.


joewhitsitt (approved):
noticeable improvement here!

PR #8887, opened 2025-06-11
Refresh signs

### LINKED ISSUE #8746: Sign refresh for slide scheduling
# Story

As a Signage user, I would like to be able to set my sign refresh interval so that my scheduled slides will update.

# Additional Info

Currently users can set a sign refresh interval (10, 30, 60 mins) so that their sign updates and shows the correct scheduled slides.

Sounds like there may be a refresh built into the toolbox as well?

Do we need this built into the signs as well?

May factor in to the Hawk Alert functionality as well #8742 

# Proposed solution

pyrello:
Resolves #8746 

# How to test
- Note: for testing, it might be worth adding the following options to [`sitenow_signage_reload_intervals_allowed_values()`](https://github.com/uiowa/uiowa/blob/6fc3247d61fea009a03a07f95e967f00c78d812d/docroot/modules/custom/sitenow_signage/sitenow_signage.module#L154-L162):
```
    30 => '30 seconds',
    10 => '10 seconds',
```
```
ddev blt frontend && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli node/add/sign
```
- Add a sign with 2 or more slides. Start with at least one unpublished slide and schedule publishing for the next minute. Select one of the reload intervals added earlier to avoid having to wait as long to see the slide reload.
- Confirm that unpublished slides show but have the word "Unpublished" in the upper left hand corner.
- Open your sign in an incognito browser so you can watch it refresh as a logged out user. Confirm that you are not seeing any unpublished slides logged out.
- Assuming a minute has passed, you can now run cron:
```
ddev drush @sandbox.local cron
```
- Your unpublished slide(s) should be published by cron. Observe that this change is reflected in both browser windows after the next reload. The previously unpublished slide will continue to show logged in but will no longer display "Unpublished" on it.
- Edit the sign to set a different reload interval (whichever one you didn't select earlier) and edit one or more of the slides to schedule unpublishing in the next minute.
- Confirm that the sign is now reloading at the new interval. Hint: You can see the reload interval in the browser's dev tools console.
- Assuming a minute has passed, run cron again.
- You slides scheduled to be unpublished should be processed by cron. Confirm that you are seeing the expected changes after the next reload in both browser windows.
- Code review.
- Make sure you didn't accidentally commit the changes above for testing ;)


joewhitsitt (approved):
Works as expected!

Issue #8917, opened 2025-06-18
Investigate using Cloudflare Turnstile Free version on paid campaign form


briand44:
# Story

As a web team, we would like to try out Cloudflare Turnstile free version on a form to help with spam especially on paid campaign forms so that we can prevent spam submissions from flooding our recruitment forms and muddying up our internal applications.

# Additional Info

https://developers.cloudflare.com/turnstile/

>Turnstile is WCAG 2.1 AA compliant.

# Proposed solution

- Create an account with its-web@uiowa.edu email and save password in our password manager
- Create and configure Turnstile widget using `uiowa.edu` hostname?
- Try implementing this widget on an existing form - https://uiowa.edu/form/rmi-campaign-rfi
- Monitor analytics and submissions to see if it is adding value

joewhitsitt:
https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/

https://developers.cloudflare.com/turnstile/get-started/server-side-validation/

joewhitsitt:
Huddled with Brian and Michael on this. With them we established an account and saved the details to esaccess. Michael chose invisible/no pre-clearance to start with. It looks like we can change that in the Cloudflare UI without changing the site/secret.

There is a module that integrates turnstile with our existing webform captcha solution so we are evaluating that first. the module depends on the `key` module and we experimented with how that stores the key information. By using the `config` option the key and secret are stored in plain text in the config table. For this pilot, there wasn't strong concerns there since it does not grant access to the cloudflare account and the combination is limited to allowed hostnames (e.g. uiowa.edu to start). Later on we might want to look into storing these in the secrets file similar to other creds.

With the module configuration in a guess and check phase, we decided to ignore the config in config.ignore default.

joewhitsitt:
this has been configured/setup on uiowa.edu, data is coming over to cloudflare. no wsod on admin/config once configured. and with initial testing it looks like site and varnish caches are both still HITs.

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/0b3dce5a-2e6e-4d9f-89a8-f15fc107e6f9" />
<img width="1566" alt="Image" src="https://github.com/user-attachments/assets/c03040c7-8dfb-44b6-8305-daff02972053" />

joewhitsitt:
Confirmed that webmasters are able to select the captcha element and the turnstile option.

PR #8918, opened 2025-06-18
initial cloudflare turnstile config for uiowa.edu

joewhitsitt:
# How to test

```
ddev blt ds --site=uiowa.edu
```

See that the cloudflare and key modules are enabled for the website but the configuration is unconfigured and ignored.

More background here: https://github.com/uiowa/uiowa/issues/8917#issuecomment-2985005659

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -195,6 +195,7 @@
         "drupal/table_header_scope_attribute": "^1.0",
         "drupal/taxonomy_path_breadcrumb": "^1.0@beta",
         "drupal/theme_permission": "^2.0",
+        "drupal/turnstile": "^1.1",
         "drupal/tvi": "^2.0@alpha",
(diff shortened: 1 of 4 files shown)


e-marie-w:
I found the key module fine and see that no keys are available. While searching for the Cloudflare module in the menu, though, I got the following error
Screenshot 2025-06-18 at 11 52 49 AM
So far this only happens when visiting https://home.uiowa.ddev.site/admin/config. I can visit all other menu items fine, and even submenu items under Config fine.

joewhitsitt:
hmm. i'll drop my db and take a look quick

e-marie-w (approved):
Approving, with caveat that we'll need to configure it right away to avoid the error with the config menu item.

PR #8930, opened 2025-06-20
[records.fo.uiowa.edu] Migration fo records

### LINKED ISSUE #8792: Create migration of records for records.fo.uiowa.edu
# Story

As a site owner for records.fo.uiowa.edu, we would like to have our records migrated from our existing site into our new site so that we don't have to manually create over 300 records.

# Additional Info

Structure/fields should be pretty much the same from old site to new.

We shouldn't need to create redirects as part of the migration since the base url is changing as well.

Migrate all record retention content from fmb.fo.uiowa.edu to records.fo.uiowa.edu. Published and unpublished should both come over and keep their current published state.

Should migrate taxonomy terms as part of this.

Setting up dev environment and running migration in dev can happen as a separate slice.

**Field mappings - D7 -> v3**
title -> title
field_retention_record_category ->  field_record_category - taxonomy - create as part of record migration
field_retention_description -> body - no summary - filtered html
field_retention_unit_criteria -> field_record_ad_unit_responsible - taxonomy - create as part of record migration
field_retention_unit_maintain -> field_record_files_maintained_by - taxonomy - multiple values - create as part of record migration
field_retention_year_updated -> field_record_year_updated
field_retention_guidelines -> field_record_ui_retention_guide - plain text - Boolean - single on/off
field_retention_confidential -> field_record_confidential
field_retention_vital -> field_record_vital

# Proposed solution

cory-skeers:
Resolves #8792 

# How to test

In DevDesktop, sync fmb.fo.uiowa.edu (uiowa703. Database only is needed, files not necessary)
```
ddev blt ds --site=records.fo.uiowa.edu && ddev drush @forecords.local en forecords_migrate -y && ddev drush @forecords.local uli /admin/config/sitenow/sitenow-migrate
```
Database Name: `fmb_fo_uiowa_edu`
Database User: `root`
Database User Password:
Database Host: `host.docker.internal`
Database Host Port: `33067`
Base URL: `https://fmb.fo.uiowa.edu`
Public Files Path:
Definitions: all blank

```
ddev drush @forecords.local ms
```
See that the migration shows up with 300+ records
```
ddev drush @forecords.local mim forecords_record
```
Watch the migration finish (no time to make a coffee for this one, unfortunately)
Spot check content
(You can use https://fmb.fo.uiowa.edu/records-management/filtered on the existing site, and create a page on the local site with the Site custom block: Records list for easier comparison)

Check that the number of records migrated matches the number of records on the existing site, including both published and unpublished (Cory had 308 when creating the migration, Brian recently saw a number closer to 320. Discrepancy might've been something like the source having been pulled from dev instead of prod, or something else, and it should be double-checked).


joewhitsitt:
pulling to test

joewhitsitt (approved):
>no time to make a coffee for this one, unfortunately

heck, i couldn't even take a sip of coffee in the time it took to finish. speedy.

looks good. spot checked some records. not necessarily migration related but did we want to keep the URL path pattern? records-management/animal-protocol (d7) vs records/animal-protocol (d10). i know the domain changes.

briand44:
>did we want to keep the URL path pattern?

Discussed this with them when we last met. They were fine with the new url pattern.

PR #8935, opened 2025-06-23
Signage - Bus schedules widget

### LINKED ISSUE #8922: Signage bus arrivals widget
# Story

As a signage feature user I would like to be able to display a bus schedule on my sign so that people can be able to view those schedules.

# Proposed solution

- Define a widget fragment type for the bus schedule widget.
- Hook into the fragment entity hooks to access the bongo API, return results, and render them on the fragment item.
- Add a field to the slide content type to reference the fragment entity type.

pyrello:
Resolves #8922 

# How to test

```
```

- Create a Sign Manager user and masquerade.
- As a sign manager, you can still create signs and slides without permission restrictions.
- On a slide you see the image and bus arrivals paragraph options.
- Creating a slide image, the output doesn't change.
- Creating a bus arrivals slide, you can select a stop and save. On render it displays cambus and coralville predictions in a table. The bus times should refresh without browser reload every minute (if the bus prediction upstream has updated).
- The table is in a clear format that blends nicely with the existing signage look and feel.
- You can have multiple bus stop widgets in the same slideshow without error.


joewhitsitt:
i was too quick to steer towards the existing api client stuff. to refresh the stop info, we need to do a JS fetch like the hawk alerts

joewhitsitt:
ruh roh, the label is being shared between slides

joewhitsitt:
Questions

- Okay with cambus/uiowa represented as white this time around and the light alert colors for danger and info for ic, coralville?
- Signage specific table class for cell borders, font-size adjustments?
- Fix *bg- inherit issue for border color in uids_base/uids? This was also a headache with accordions outside basic page.
  - `[class*=bg--] *, [class*=bg--] [class*=bg--] * {border-color: inherit;}`

GaryRidgway:
Change 0m to "now arriving" and set "1 minutes" to "1 minute"

e-marie-w:
Pulling down to test.

e-marie-w (approved):
I followed the testing instructions, everything works as expected and visually looks great!

PR #8947, opened 2025-06-26
Hotfix - add schema alter info for custom fields on basic site settings form

joewhitsitt:
This is preventing the save of the basic site settings form even if they are not trying to set or modify a parent organization.

Reported by more than one user in its-web

Referenced https://www.jaypan.com/tutorial/drupal-extending-core-configuration-extending-core-forms-and-overriding-core-routes as it appears our docroot/modules/custom/uiowa_core/config/schema/uiowa_core.schema.yml file is not correct or not being picked up.

# How to test

On `main`:

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli admin/config/system/site-information
```

Try to save the form. There is an error about an invalid parent key.

Checkout this branch `extend_system_site_schema`.

```
ddev drush @sandbox.local cr
```

Try to save the form. Success! Try entering parent org info. Save. Success!

### DIFF
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -1847,3 +1847,28 @@ function uiowa_core_clean_address_fields(array &$form, FormStateInterface $form_
     }
   }
 }
+
+/**
+ * Implements hook_config_schema_info_alter().
+ */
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Success!

I also checked the filter settings since that is also being extended for the allowed iFrame sources, but it looks like the schema there is working correctly from the `uiowa_core.schema`.

Issue #8894, opened 2025-06-12
Improvements to people list block filter functionality


briand44:
Noticing a few things with this:

- Don't have ability to reset filters
- taxonomy term id's display along with term name when a tag or research area is selected
- if research area exposed and using custom label, custom label not displayed
- Can tags and research areas be select lists instead of autocomplete?

From Joe:
>Can tags and research areas be select lists instead of autocomplete?

I know this was requested in the IS, but I'd like to push against this just because some of these lists are very large and it doesn't necessarily have a person content relationship.

# Proposed Solution

- [ ] Add reset button
- [ ] Display custom label for Research Areas filter if site configured to use a custom label
- [ ] Display research areas filter as a select list rather than autocomplete to align with People page

joewhitsitt:
Also the autocomplete (tid) seems to be a big long issue:

https://www.drupal.org/project/drupal/issues/2881892

Dropdown or chosen/select2 seem to be the least hacky solutions.

joewhitsitt:
oops. also recalling that on the form config we allow multiple values in the filters which is why i think page list block has autocomplete. we have resisted allowing multiple selections in several cases and used chosen or something similar to wrangle it.

briand44:
Planning conversation... When a default is set while also exposing, the reset functionality resets it to the default and not empty. Ok with moving forward with it as is and can have follow up if folks want it to behave differently.

bspeare:
Ran this command:
```
./vendor/bin/blt ume "sqlq \"SELECT DISTINCT entity_id FROM node__layout_builder__layout WHERE layout_builder__layout_section LIKE '%people_list_block%' AND layout_builder__layout_section LIKE '%filter-field_person_research_areas_target_id%' AND (layout_builder__layout_section LIKE '%field_person_research_areas_target_id%\\\"value\\\";a:2:%' OR layout_builder__layout_section LIKE '%field_person_research_areas_target_id%\\\"value\\\";a:3:%' OR layout_builder__layout_section LIKE '%field_person_research_areas_target_id%\\\"value\\\";a:4:%' OR layout_builder__layout_section LIKE '%field_person_research_areas_target_id%\\\"value\\\";a:5:%') ORDER BY entity_id;\""
```

Found three sites using multiple research areas: 

Executing on writinguniversity.org...
4286

Executing on archaeology.uiowa.edu...
246

Executing on music.uiowa.edu...
1781

# Next steps
[...]

PR #8948, opened 2025-06-26
Facilities - fix view config and sort artists by label

joewhitsitt:
Reported by Facilities in its-web

# How to test

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli art-campus/art-campus-gallery
```

- See that the buildings filter does not contain non-building content
- See that the artists filter is sorted alphabetically by option value
- You see a seemingly complete list of buildings (goes beyond C down through the rest of the alphabet).
- You don't notice any performance issues with loading the view compared to PROD.

### DIFF
--- a/config/sites/facilities.uiowa.edu/views.view.artwork.yml
+++ b/config/sites/facilities.uiowa.edu/views.view.artwork.yml
@@ -1042,9 +1042,9 @@ display:
             default_group_multiple: {  }
             group_items: {  }
           reduce_duplicates: false
-          handler: 'default:node'
+          sub_handler: 'default:node'
           widget: select
-          handler_settings:
+          sub_handler_settings:
             target_bundles:
               building: building
             sort:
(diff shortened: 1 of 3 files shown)


cory-skeers:
It looks like the Building filter might also be truncating the results? 

image

I think this is only the first 100 buildings alphabetically.

cory-skeers:
Ope.
image

joewhitsitt:
ha.. ugh. another potential issue for https://github.com/uiowa/uiowa/issues/8894

cory-skeers (approved):
> I'd be interested in trying [this](https://github.com/uiowa/uiowa/pull/8948/commits/1261f5798f1b37a502136f27a2dc603e27fbe1a7) scoped to one website. There are ~400 buildings and the limit is generally set at 500. I have a question out here: https://www.drupal.org/project/drupal/issues/3481590#comment-16168683 asking about the performance issues that led to the limit. The solution here is interesting, but it doesn't look ready to use yet. Putting back up for review.

Interested in that as a more permanent solution, but the fix here seems to be working well without any noticeable performance issues from my testing. Probably will be good to check it in production after this gets out to be sure, but I think this view is a good test case for it as a solution.

PR #8949, opened 2025-06-27
Signage - Robots

### LINKED ISSUE #8940: Don't allow signage content to be indexed by search engines
# Story

As a web team, we don't want Signage content to be indexed by search engines so that folks aren't visiting and interacting with signs/slides in a browser.

# Additional Info

>Currently signs and slides can be indexed. We want to add no-follow meta info to make sure that is not the case.

# Question

Are we scoping this to the Signage feature or including the standalone site as well?

# Proposed solution

cory-skeers:
Resolves #8940 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && 
ddev blt ds --site=default && 
ddev drush @default.local config-split:activate signage -y && 
ddev drush @default.local cim -y && 
ddev drush @sitessignage.local uli /admin/config/search/robotstxt && 
ddev drush @default.local uli /node/add
```

Check that the signage.sites.uiowa.edu robots configuration should disallow all agents. (Note: You can't go directly to `/robots.txt`, because locally that'll have environment-specific settings anyway).

For signage-enabled default, add a sign and a slide. 
For each, View Page Source and see that robots `nofollow` and `noindex` metatags are set.
While editing, see that you don't have a `metatags` field to edit away from these defaults. Also verify that the Sitemap option is uneditable, and defaulted to not include in the sitemap.

### DIFF
--- a/config/default/config_split.config_split.signage.yml
+++ b/config/default/config_split.config_split.signage.yml
@@ -20,6 +20,8 @@ complete_list:
   - field.field.node.slide.field_image
   - field.storage.node.field_sign_display_title
   - field.storage.node.field_sign_hide_header
+  - metatag.metatag_defaults.node__sign
+  - metatag.metatag_defaults.node__slide
   - node.type.sign
   - node.type.slide
   - pathauto.pattern.sign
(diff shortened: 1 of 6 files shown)


joewhitsitt (approved):
imports clean. works as expected for both feature and site. confirmed as a sign manager (backend) and anonymous visitor (frontend)

PR #8955, opened 2025-06-30
[Signage] Hide back to top button

### LINKED ISSUE #8941: Hide back to top on Signage content
# Story

As a web team, we would like to hide the back to top functionality on signage content since since it doesn't add any value and can cause confusion.

# Additional Info

>Currently sites that have the back to top button enabled will display that back to top button on signs. We don't want this to render on signs, so we should remove it.

# Questions

Should we be hiding on slides in addition to signs?

# Proposed solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8941. 

# How to test
```
 ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli /sign/brian-test-sign
```
1. Confirm back to top button is not on sign
2. Go to another page and confirm that the arrow is there

### DIFF
--- a/docroot/themes/custom/uids_base/templates/layout/page.html.twig
+++ b/docroot/themes/custom/uids_base/templates/layout/page.html.twig
@@ -81,7 +81,7 @@
 
   {{ page.content }}
 
-  {% if header_toppage and not_digital_sign %}
+  {% if header_toppage and show_back_top_nav %}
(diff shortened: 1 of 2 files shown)


cory-skeers (approved):
Looks good! 👍  I like the approach here and how we can easily add extra content types to it if needed in the future.

PR #8956, opened 2025-07-01
[People list block] Enhancements 

bspeare:
# How to test
```
ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli /people-list-block-conversion-test-page
```
1. Confirm that "Custom label" renders in place of "Research areas" in the exposed form. 
2. Confirm that reset button appears all the time. 
3. Edit block and confirm that "Custom label" appears under "Filters" instead of "Research areas".


joewhitsitt (approved):
Works as expected. Removed all of the exposed filters to confirm that the reset/exposed form did not persist.

Custom label was seen in both places

Issue #8896, opened 2025-06-12
Test/Move schedule cron job to on the hour for all apps


joewhitsitt:
Now that we are on ACN, I wonder if we still need to offset our cron jobs for each application.

This could allow us to do more on time scheduling and more frequent runs (every 30, 15 min?).

# Proposed Solution

- [x] Open a ticket with Acquia to ask if we can move all apps to on the hour or if there are still performance concerns with this on ACN (shared resources).
- [x] Change apps' scheduled `cron` jobs to run on the hour similar to the uiowa application.

<img width="307" height="176" alt="Image" src="https://github.com/user-attachments/assets/605a0d3a-ab89-4091-be46-d9d23ef2ff8c" />

briand44:
Grooming conversation.. Switch scheduled jobs in prod on Cloud UI (cron). Article documentation updates is tied to #8990 as well as other help text stuff. Webform documentation doesn't need changed.

joewhitsitt:
Updated https://docs.prod.drupal.uiowa.edu/acquia-cloud and https://sitenow.uiowa.edu/documentation/article-content-type

briand44:
>Updated uiowa01-uiowa09 prod.

Assuming `uiowa` was already running on the hour?

joewhitsitt:
I am going to close this for now. There have been any reported issues with this change. Granted it is the dead of summer so we will want to consider this if we start to get service disruption type items around the hour mark in the future.

PR #8990, opened 2025-07-08
Scheduler language stuff

### LINKED ISSUE #8954: Make scheduler language consistent
# Story

As a web team, we want the scheduling language to be consistent so that we don't cause confusion for folks scheduling slides.

# Additional info

>Weirdly, when creating a slide on its own (vs. via a sign), the language of the scheduler widget says "Archive on" instead of "Unpublish on". It says "Unpublish on" when creating a slide via a sign.

# Proposed solution

cory-skeers:
Resolves #8954 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu &&
ddev blt ds --site=facilities.uiowa.edu &&
ddev drush @sitessignage.local uli /node/add/slide &&
ddev drush @facilities.local uli /node/add/alert
```

Check the scheduler options on the following:
Article (full node, default)
Slide (full node, inline entity form when adding a new slide from a sign, signage.sites or signage feature)
Alert (full node, facilities.uiowa.edu)

See that the "publish" and/or "unpublish" language in the labels and help texts of each field are correct and accurate, and that help text explaining the one-hour window for cron is present in each.

Any other forms with scheduler options I missed?


docroot/modules/custom/uiowa_core/uiowa_core.module
```diff
@@ -1848,6 +1836,27 @@ function uiowa_core_clean_address_fields(array &$form, FormStateInterface $form_
...
+ *
+ * @param array $form
+ *   The form array.
+ */
+function uiowa_core_scheduler_help_text(array &$form): void {
+  $hide_classes = ['js', 'js-hide'];
+  foreach (['publish', 'unpublish'] as $state) {
+    if (isset($form["{$state}_on"])) {
+      // Update help text for scheduler on content types.
+      $additional_help_text = t('Note: this content will transition to a @stateed state within a one-hour window of the selected date and time.', [
```
joewhitsitt: Don't know if we want to hold the presses on this?

https://github.com/uiowa/uiowa/issues/8896#issuecomment-3057550626

cory-skeers: Possibly... I think we'll still want to have help text about it transitioning according to the cron schedule, but if we're able to schedule cron more predictably for all applications, we could provide more specific (and therefore more _help_ ful) help text.

joewhitsitt: Would need to update docs too https://sitenow.uiowa.edu/documentation/article-content-type

Sorry, getting excited.

cory-skeers: Ooh, good catch.

joewhitsitt (approved):
This has my approval, but I've encouraged the planning of https://github.com/uiowa/uiowa/issues/8896 so that we can make the hour window change before this is merged.

docroot/modules/custom/uiowa_core/uiowa_core.module
joewhitsitt:
```suggestion
      $additional_help_text = t('Note: this content will transition to a @stateed state at the next hour after the configured time. (e.g. 5:13 p.m. will transition at 6 p.m.)', [
```

Ugh I don't know if this is any better. Now I want to change the time input to be restricted to hours.

Update: Added a time example based on https://github.com/uiowa/uiowa/pull/8990#discussion_r2208335543

joewhitsitt: Added "... due to when server cron operations trigger" to the article content type docs. 🤷

bspeare: It still seems a little confusing but I don't have a solution to make it better.    I think your text does help bring clarity to the new schedule.   Would it help to provide a time example and when it would be transitioned?

pyrello: Personally, I think the update text works without the e.g. example in parentheses. This might be a little clearer:
"Note: this content will be published on the next hour after the configured time."

If we're going to keep the parentheses, I think they should go before the period at the end of the sentence.

joewhitsitt:
```suggestion
      $additional_help_text = t('Note: this content will be @stateed on the next hour after the configured time.', [
```

m-blaisdell: "Configured" may be a bit of a confusing/technical term for some editors. 

Maybe something like "Publishing/unpublishing occurs every hour on the hour, regardless of the minute you select. (e.g., 2:43 p.m. will go live at 3:00 p.m.)."

Or something like "Publishing/unpublishing occurs at the top of the hour after the time you select. (e.g., 2:43 p.m. will go live at 3:00 p.m.)."

joewhitsitt:
```suggestion
      $additional_help_text = t('Note: this content will be @stateed at the top of the hour after the time you select. (e.g., 2:43 p.m. will go live at 3:00 p.m.).', [
```
I feel like we would be calling ourselves out with the word `regardless`, like we acknowledge we have little control over the functionality (🤷 who knows when!), so here is an adjusted version of Maggie's second suggestion. Because the `@stateed` var is "published" or "unpublished" I had to tweak it a bit for it to work.

GaryRidgway:
I think linking it directly to the save button will make users understand the trigger that will schedule it. So, something like "Note: Upon save, this content will be @stateed on the next hour."

docroot/modules/custom/uiowa_core/uiowa_core.module
pyrello: Is it worth considering removing the minutes on this? Maybe rather than tying ourselves in knots trying to explain how its going to work, could we just align the input to match what will actually happen?

joewhitsitt: >could we just align the input to match what will actually happen?

i'd like to look at that for the long term, but I think this was the stop-gap without trying to change the contrib module.

joewhitsitt: there was three thumbs up on https://github.com/uiowa/uiowa/pull/8990#discussion_r2208778518 so i committed that version for now. i'll create a follow up slice to look into changing the input.

joewhitsitt: Next slice: https://github.com/uiowa/uiowa/issues/9021

PR #9006, opened 2025-07-11
[Signage] Image sizing

### LINKED ISSUE #8953: Sign image sizing
# Story

As a web team, we would prefer not to be cropping image slides so that we are not removing necessary portions of images.

# Additional Info

>Fill vh vw but keep the image aspect ratio (center horizontally aligned to top?), scale down to fit - Brand is still shipping 1920x1080 templates (16:9). I think it would be good to support both 16:9 and 4:3 with the header/no header options. Perhaps restrict what can be uploaded or referenced somehow?

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/8953. 

```
ddev blt frontend && ddev blt ds --site sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/add/sign
```
1. Code review and confirm new files are scoped to signage split.
2. Confirm "Horizontal" is the default.
3. Try uploading an image smaller than 2000 pixels wide and confirm that the image style scales and crops it to 2592 × 1458 pixels. 
4. Switch to "Vertical" orientation and confirm image is scaled and cropped to 1458 × 2592 pixels.


joewhitsitt:
pulling to test

config/features/signage/core.entity_view_display.node.sign.default.yml
```diff
@@ -83,7 +85,14 @@ id: node.sign.default
...
 mode: default
-content: {  }
+content:
+  field_sign_orientation:
+    type: list_default
+    label: above
+    settings: {  }
+    third_party_settings: {  }
+    weight: 1
+    region: content
```
pyrello: I'd move this field to the `hidden` section similar to the other fields. This has to be done manually on the file as the UI for changing this is hidden because LB is turned on for this content type.

bspeare:
Added a bug issue at https://github.com/uiowa/uiowa/issues/9032 and it appears there is a similar known issue at https://sitenow.uiowa.edu/known-issues#uiowa-issue-7108.  Seems like we could adjust the known issue problem to be "Issue: Image style (https://sitenow.uiowa.edu/documentation/media-management/images-and-files) processing issues when trying to upload images as content or through web forms."

joewhitsitt:
😮 i didn't realize this was the cause. thanks for the detailed response. using non-iphone images from the web, this is all working as expected. can you think of anything we should do as a stop-gap for this (help text, programmatic restrictions, limit to horizontal only)?

+1 to Sean's comment about tidying up display settings

joewhitsitt (approved):
synced the changes, clean import and the same export (unrelated role permissions stuff).

PR #9012, opened 2025-07-15
Adding patch in composer for ckeditor and styling for list styles.

### LINKED ISSUE #6283: More options for styling lists

james917:
# How to test
Checkout the pull request run 

```
 ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /test-ordered-lists-brian 
```

1. Compare the node edit preview of the lists on `/test-ordered-lists-brian` to the layout builder edit preview of the text area block on `/test-ordered-lists-brian`. 
2. Compare to the prod preview of the same text area block at https://sandbox.prod.drupal.uiowa.edu/test-ordered-lists-brian.
3. Confirm that the `ol` and `ul` list styles are rendering the same in each WYSIYWG and that you can use the new styles dropdown menu to change the list styles and the preview will change as you select a new list style. 
4. Test that the "disc", "square", and "circle" list styles are working.  Compare against https://sandbox.prod.drupal.uiowa.edu/test-ordered-lists-brian. 
5. Test that existing lists are unaffected: compare https://sandbox.uiowa.ddev.site/lists with https://sandbox.prod.drupal.uiowa.edu/lists.


package.json
```diff
@@ -23,5 +23,6 @@
     "@percy/cli": "^1.30.10",
     "requirejs": "^2.3.7"
   },
-  "devDependencies": {}
+  "devDependencies": {},
```
joewhitsitt: Remember, this should not be included @james917

GaryRidgway:
Gonna poke around and see what I can do

GaryRidgway:
We got the inherit working here, but now its just removing the formatting. We are a step closer!

GaryRidgway:
Looks like when compiling there it keeps the original selector, but ALSO adds the prepended selector.

For example, if you write `.ck.ck-content ol li` as your selector, then the compiled selector will be `.ck.ck-content ol li, #drupal-off-canvas [data-drupal-ck-style-fence] .ck.ck-content ol li`.

So ultimately, this may still need some tweaking to understand where to place it.

joewhitsitt:
Sounded like @james917 was able to see the effects of these changes. Pulling down as additional review since James is the PR author.

bspeare:
@joewhitsitt we can just go with `$is_layout_builder_edit` if there's too many checks there.   Was there something you wanted to use there?

joewhitsitt (approved):
I think this is working. Compared the style types prod/local diff in Firefox.

PR #9035, opened 2025-07-17
[Signage] Sign intervals

### LINKED ISSUE #8784: Ability to set slide duration
# Story

As a Signage content creator, I would like to be able to configure the duration a slide is displayed before the next slide shows so that our slides can be displayed for more or less time, depending on the amount of information on the slide.

# Additional Info

Current Signage solution defaults to 10 seconds. Tippie's Signage solution defaulted to 30 seconds I believe.

Should set a default (greater than 10 seconds) but allow it to be configurable per slide.

# Proposed solution

- Understand what happens if there is a set duration and only 1 published slide (Do we need a 0 option)?
- Add allowed options select list with intervals
- Pass the option from the select list to splidejs - https://splidejs.com/guides/options/#interval
- document or make sure it gets documented in #8969

cory-skeers:
Resolves #8784 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/add/sign
```
Add a sign with some slides.
When adding a new slide directly from the sign, it should default to 15 second duration
When creating a slide from /node/add/slide, it should also default to 15 second duration, but have a functioning select list to change the duration
When viewing a sign with slides that have different durations, each slide should be visible for its allotted duration
When viewing a sign with a single slide, it should not fade in and out, but just stay constant


bspeare:
Pulling to test.

bspeare:
Converting this back to draft as it is missing some config.

joewhitsitt (approved):
glad to have this working, the quick default durations were driving me nuts!

cory-skeers:
Thanks for the IEF fix @joewhitsitt!

PR #9038, opened 2025-07-18
[Signage site] Output groups for user

### LINKED ISSUE #9019: Ability to see groups that I am a member of
# Story

As a standalone Digital Signage user, I would like to be able to see a list of all the groups I am a member of so that I can quickly get to the appropriate group.

# Additional Info

Ideally, the groups would be output on my profile since that is where I am taken immediately after authenticating. They should link to the group page.

Some discussion on whether it would be better as a top-level (and easy to find) link called "My groups".
https://github.com/uiowa/uiowa/issues/8965#issuecomment-3084932831

>Provide an alternative UI for viewing groups to which the user belongs

From https://github.com/uiowa/uiowa/issues/8965#issuecomment-3084835288

# Proposed solution

e-marie-w:
Closes #9019 

# How to test

- `ddev composer install && ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli`
- visit https://sitessignage.uiowa.ddev.site/user/11, should see list of signage groups user is a member of. Group names should link to the group pages.
- visit https://sitessignage.uiowa.ddev.site/admin/people and click one of the users without the sign manager role. Their profile page should not display a block of groups. 
- Add a new or existing user to a new or existing signage group, check that the profile page updates accordingly


joewhitsitt:
@e-marie-w i think you are on the right track. if you can get the entity id (Signage Group node id) as the output, you can follow up with preprocess_views_view_field and change the markup after doing a node load lookup to get the entity label.

I think the simplest views_view_field example we have is `brand_core_preprocess_views_view_field` where it replaces the nothing field with a download link. https://brand.uiowa.edu/lockup-system

joewhitsitt (approved):
I am happy with these changes.

- It appears the commit has stayed pinned through a composer install despite there being more commits since. ✅ 
- The bulleted list is basic but clean and informative. ✅ 
- The view is accurate and doesn't have duplicate results. ✅ 
- The view does not show unless there is a group assigned to the user. ✅ 
- No config import errors ✅

pyrello:
@e-marie-w I notice that the version of OG was updated as part of this, but I'm not finding any comments on the PR or the issue explaining this. Can you add any context for this?

e-marie-w:
> @e-marie-w I notice that the version of OG was updated as part of this, but I'm not finding any comments on the PR or the issue explaining this. Can you add any context for this?

Yep! With the version of OG we were using, only certain fields were available when accessing OG Membership through a view. The group name was not one of these fields. A recent commit in the module makes this field available in the view.

I originally tried just using a patch, but that did not work. So, I updated the module and pinned the commit we need.

e-marie-w:
> Can you include the link to the issue or commit for this?

https://www.drupal.org/project/og/issues/3525064

PR #9055, opened 2025-07-23
Delete uiservicecenter.uiowa.edu multisite on uiowa

### LINKED ISSUE #9053: Deprovision uiservicecenter.uiowa.edu from uiowa
> I heard back from the library that they've already taken a snapshot of this site and the service center stakeholders don't believe we should redirect to any one existing site. Whenever you all have time and want to decommission that site we would appreciate it!
> 
> If you could just give me a little heads up so I can let them know when to expect it to go down.
> 
> Thanks again!
> 
> Josh 

Should be able to reclaim the CNAME with this one.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9053. 

# How to test


joewhitsitt (approved):
mentioned creating an SSL issue to clean up completely

PR #9072, opened 2025-08-04
Ckeditor <small> tag in toolbar

### LINKED ISSUE #9050: As a SiteNow editor, I can use the `<small>` tag in CKEditor to make text appear smaller than paragraph text
# Story

As a SiteNow editor, I can use the `<small>` tag in CKEditor to make text appear smaller than paragraph text and not have to use the `<sup>` or `<sub>` options.  

# Proposed solution

- See if we can add the `<small>` button to CKEditor: https://www.drupal.org/project/ckeditor_small_tag

Admissions and Policy would both have use for this.

cory-skeers:
Resolves #9050 

# How to test

```
ddev composer install && ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli
```
Edit a piece of content, and see that there is a `small` toolbar icon in the WYSIWYG.
Try it out, see that it properly adds a `<small>` tag to the content, saves fine, etc.


e-marie-w (approved):
The `small` option works great, awesome job! I tested on both new and existing content, and the Edit tab WYSIWYG as well as the WYSIWYG of a card in Layout Builder. I agree with the decision to leave it out of the Basic HTML format and add it to all others. On text already using the tag via source HTML the icon correctly highlights to show it is being applied.

PR #9081, opened 2025-08-06
Sanitize filenames

### LINKED ISSUE #8960: Filename sanitation (Transliterate)
# Story

As an editor, I would like my files to adhere to best practices as it relates to filenames without having to fix/correct the filenames myself. This will allow me to easily share direct links to files and redirect file URLs after the files have served their purpose.

# Problem

- Menu link media matching does not yet work. LinkIt maintainer says our half-baked solution is best until core comes along: https://www.drupal.org/project/linkit/issues/3393201#comment-16168706
- Copying an absolute/relative filename URL for media items is tricky enough before double encoding occurs after pasting the link into a field.
- After a file containing spaces is deleted and removed from the server, the redirect module makes it difficult to create a redirect for that file to point visitors to the correct destination.

# Proposed Solution

- [ ] Evaluate and implement the filename sanitizing functionality now provided by core https://www.drupal.org/node/2972665 (does not affect existing files) by modifying and exporting configuration from the default site.
- [ ] Confirm re-saving media or using the file replace (overwrite existing) are not affected by these changes.
- [ ] Optional deep dive to understand existing filename patterns used and how they will be affected by these changes for new files following the same patterns.
- [ ] Update media page of documentation to focus on the general media upload process as well as detail about which file name sanitization bits will be applied.

cory-skeers:
Docs updated separately.
Resolves #8960 

# How to test

1. Sync your site of choice, ideally with a file, image, or other media with spaces in the filename. Alternatively, while on `main`, upload a file with a filename which includes things like single or multiple spaces, multiple dashes, etc., then checkout this branch and run a `ddev drush @site.local cim -y`
2. Add a new media item and upload a file with spaces and/or multiple dashes, underscores, etc.
3. See that there is a message with the updated filename, that the spaces and such have been replaced with hyphens
4. Save
5. Check the site's files directory on your computer and see that it was properly added with the changes
6. Check that a file that existed prior to the change (ie from step 1) still has its same filename
7. Edit that file's media entity by changing alt text or something similar, and see that the filename has not changed
8. Edit that file and upload a new file with the "Overwrite existing file" option checked and save
9. Ignore the status message (relates to https://github.com/uiowa/uiowa/pull/9081#issuecomment-3168611778), and check that the filename still matches the original file (checking the filename listed on the media entity should be sufficient, but you can check the actual file directory on your computer as well to verify)
10. Edit that file again, upload a new file without the "Overwrite existing file" option and save
11. See the status message, and check that the filename now matches the status message's output

Docs:
Check out the proposed doc updates at https://sitenow.uiowa.edu/node/171/latest and https://sitenow.uiowa.edu/node/326/latest
Make edits, suggestions

### DIFF
--- a/config/default/file.settings.yml
+++ b/config/default/file.settings.yml
@@ -8,8 +8,8 @@ icon:
 make_unused_managed_files_temporary: true
 filename_sanitization:
   transliterate: false
-  replace_whitespace: false
-  replace_non_alphanumeric: false
-  deduplicate_separators: false
+  replace_whitespace: true
+  replace_non_alphanumeric: true
+  deduplicate_separators: true
   lowercase: false
   replacement_character: '-'
--- a/config/default/file.settings.yml
+++ b/config/default/file.settings.yml
@@ -7,7 +7,7 @@ icon:
   directory: core/modules/file/icons
 make_unused_managed_files_temporary: true
 filename_sanitization:
-  transliterate: false
+  transliterate: true
   replace_whitespace: true
   replace_non_alphanumeric: true
   deduplicate_separators: true


cory-skeers:
Testing more with existing files, there's this scenario:

1. A file with spaces, multiple separators, etc. already exists on the site before this change
2. After the change, a user replaces the file and uses the "Overwrite existing file" option
3. The replacement file _also_ has spaces, multiple separators, etc.

The file will be replaced using the existing filename as expected, but they will receive the message showing how the new filename would have been altered by sanitization, even though no file will actually be in the system with that altered filename.

So the filesystem behavior matches what I think we want to happen, but the message to the user doesn't match.

cory-skeers:
> Testing more with existing files, there's this scenario:
> 
> 1. A file with spaces, multiple separators, etc. already exists on the site before this change
> 2. After the change, a user replaces the file and uses the "Overwrite existing file" option
> 3. The replacement file _also_ has spaces, multiple separators, etc.
> 
> The file will be replaced using the existing filename as expected, but they will receive the message showing how the new filename would have been altered by sanitization, even though no file will actually be in the system with that altered filename.
> 
> So the filesystem behavior matches what I think we want to happen, but the message to the user doesn't match.

Checked the issue queue and haven't found anything related to this. Opened a new issue at https://www.drupal.org/project/drupal/issues/3540496

config/default/file.settings.yml
```diff
@@ -8,8 +8,8 @@ icon:
 make_unused_managed_files_temporary: true
 filename_sanitization:
   transliterate: false
```
cory-skeers:
```suggestion
  transliterate: true
```

cory-skeers: Transliterate is required for `replace_non_alphanumeric` to be respected.

joewhitsitt: > [off] which is the Drupal default approach

As observed through a core install or one of the changelog statements? I'd be fine with turning it on unless there is a strong recommendation not to.

cory-skeers: Tracked it down to https://api.drupal.org/api/drupal/core%21modules%21file%21src%21EventSubscriber%21FileEventSubscriber.php/class/FileEventSubscriber/11.x in core.

I don't think there's a strong recommendation not to, but that Drupal defaulting to off is that it's not useful or functional for all languages. In our case we at least have a few examples of where it'd come into play, and more if it also enables the alphanumeric.

Let's do it!

joewhitsitt (approved):
Though a valid example was mentioned in https://github.com/uiowa/uiowa/pull/9081#issuecomment-3164481811, I think the majority of files going forward will benefit from this and users can adapt to the new pattern if they don't like the result on initial upload.

joewhitsitt:
Docs also look good. Should this close https://github.com/uiowa/uiowa/issues/8960 on merge?

PR #9082, opened 2025-08-07
[Signage site] Add operations links to Signage Group content view

cory-skeers:
# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /node/51
```
See that the Operations (edit/delete) is available in the view
Masquerade as a user who is not a member of the group, and see that you don't have the edit/delete options available


joewhitsitt (approved):
Looks good. I wasn't able to review the previous slice, but I am seeing the AJAX operations (filtering) break the styles:

Screenshot 2025-08-11 at 3 06 40 PM

looks like anon only

PR #9094, opened 2025-08-13
Signage - Campus Events slide type

### LINKED ISSUE #9005: Ability to display Uiowa Events on sign
# Story

As a user of Digital Signage, I would like the ability to display events from the University Events System on my sign.

# Additional info

There are 100 signs currently displaying uiowa events. During the usability study, it was mentioned as a must have feature for that group.

With the Bus Arrivals going away, it would be great to have at least one not image slide available for initial user testing.

# Proposed solution

- [x] Create paragraph type for Events widget. Add fields for:
  - Department
  - Audiences
  - Types
  - Interests
  - Keywords
  - Location
- [x] Add new paragraph type as target bundle for `field_slide_widget` field on the Slide content type.
- [x] Adapt existing `sitenow_events_preprocess_block` logic to fetch events based on field inputs and return results.
- [x] Style the output to be appropriate for signage display. This may involve scaling up the size of each event listing to be easily readable. This may also involve centering the list vertically, so that a shorter list of results doesn't sit awkwardly at the top of the page.

## Considerations
- The Events block has a lot of display options and it is not our intention to implement any of them for the launch version of this feature. You'll need to choose how each event is styled. A good starting point for thinking about how they should look would be the events block on https://demo.sitenow.uiowa.edu/ (the first one).
- It is not clear that there would be appreciable value in giving customers the options to select the number of results since the widget will take a whole slide. So, we are going to not include that at this time.

GaryRidgway:
Resolves #9005 

## Narrative
This PR adds a slide type for displaying lists of events from the Campus Events system. PHP classes related to fetching events from the Campus Events API were moved to the `sitenow_events` module. The `uiowa_events` module was left in place to avoid possible downtime during deployment and will be removed in a subsequent update. Some signage-specific card styles were added to accommodate the need for larger text on signs. Some refactoring of the existing signage styles was done to remove unnecessary style rules and to improve visual consistency between different slide types. A new `slide__headline` class was added to style the headline at the top of the slide and can be used for future slide types.

## To test
```
ddev yarn workspace sitenow_signage run build \
&& ddev blt ds --site=signage.sites.uiowa.edu \
&& ddev drush @sitessignage.local uli
```
- Create an events slide on a sign.
- Test out filters.
- Cross reference this with https://events.uiowa.edu/ and see if it is getting the correct results.
- Ensure that a horizontal display has 3 results and a vertical display has 5.
- Combine with other types of slides in a sign to ensure that visually they look consistent with one another.

### Commencement - The API classes that commencement uses got moved as part of this PR.
```
ddev blt ds --site commencement.uiowa.edu \
&& ddev drush @commencement.local import-events
```
- Confirm that Commencement event import process still functions as expected.


GaryRidgway:
> Without moving stuff around myself, i think it is the file organization that is not getting auto discovered correctly. `docroot/modules/custom/sitenow_events/src/Services/Utilities.php` but the class is namespace `Drupal\sitenow_events;` and I think it should be namespace `Drupal\sitenow_events\Services;` if you continue that way or move to `docroot/modules/custom/sitenow_events/src/Utilities.php`. Might as well move this to `uiowa_events/src/Utilities.php` and then update `services.yml` and the call in sitenow_events to `$EU = \Drupal::service('uiowa_events.utilities');`

I can try this, but while trying to get this to work I was referencing:
- `docroot/core/modules/config/config.services.yml` - L4-5
- `docroot/core/modules/config/src/ConfigSubscriber.php` - L3
- `docroot/modules/custom/engineering_migrate/engineering_migrate.install` - L20

Those didn't seem to be scoped to the namespace of `Services`.  Is there a reason this one might differ from that?

joewhitsitt:
Yeah, I am thinking the /src/Services folder is messing things up. Your example, `ConfigSubscriber` is in the src/ root so I don't understand what you are trying to say.

GaryRidgway:
> Yeah, I am thinking the /src/Services folder is messing things up. Your example, `ConfigSubscriber` is in the src/ root so I don't understand what you are trying to say.

Sorry, edited my comment for context

GaryRidgway:
Some context for why the
```php
sitenow_signage_value_query_string()
```
function exists.

<details><summary>Without function</summary>
<p>

>```'display_id=events&filters[audiences][0][value]=1&filters[audiences][1][value]=3&filters[department][0][value]=27&filters[interests][0][value]=50&filters[interests][1][value]=53&filters[keywords][0][value]=64&filters[keywords][1][value]=65&filters[place][0][value]=2&filters[types][0][value]=8&filters[types][1][value]=21&filters[enddate][value][date]=01-01-2100&items_per_page=100'```

</p>
</details> 

<details><summary>With function</summary>
<p>

>```'display_id=events&filters[audiences]=1,3&filters[department]=27&filters[interests]=50,53&filters[keywords]=64,65&filters[place]=2&filters[types]=8,21&filters[enddate][value][date]=01-01-2100&items_per_page=100'```

</p>
</details>

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -338,34 +323,125 @@ function sitenow_signage_preprocess_page(&$variables) {
...
-      $sign_node->hasField('field_sign_orientation') &&
-      !$sign_node->get('field_sign_orientation')->isEmpty() &&
-      $sign_node->get('field_sign_orientation')->value === 'vertical') {
+    case 'slide_events':
+      $paragraph = $variables['elements']['#paragraph'];
+      $filters = [];
 
-      $view_builder = \Drupal::entityTypeManager()->getViewBuilder('media');
+      $headline = $paragraph?->field_headline?->getValue();
+      $headline = $headline[0]['value'] ?? '';
```
pyrello:
```suggestion
      $headline = $paragraph?->field_headline?->value ?? '';
```
I think these are equivalent

config/features/signage/core.entity_view_display.paragraph.slide_events.default.yml
```diff
@@ -0,0 +1,66 @@
...
+    type: list_default
+    label: above
+    settings: {  }
+    third_party_settings: {  }
+    weight: 2
+    region: content
+hidden:
+  entity_print_view_epub: true
+  entity_print_view_pdf: true
+  entity_print_view_word_docx: true
```
pyrello:
```suggestion
content: {  }
hidden:
  entity_print_view_epub: true
  entity_print_view_pdf: true
  entity_print_view_word_docx: true
  field_slide_events_audiences: true
  field_slide_events_department: true
  field_slide_events_interests: true
  field_slide_events_keywords: true
  field_slide_events_location: true
  field_slide_events_types: true
```

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -338,34 +323,124 @@ function sitenow_signage_preprocess_page(&$variables) {
...
+        if ($sign_node && $sign_node->bundle() == 'sign') {
+          $variables['#cache']['tags'][] = 'node:' . $sign_node->id();
+        }
+
+        // Check if we have a sign node with vertical orientation.
+        if ($sign_node &&
+          $sign_node->bundle() == 'sign' &&
+          $sign_node->hasField('field_sign_orientation') &&
+          !$sign_node->get('field_sign_orientation')->isEmpty() &&
+          $sign_node->get('field_sign_orientation')->value === 'vertical') {
```
pyrello: Seems like this could use the `sitenow_signage_get_orientation` helper function here.

pyrello: Also for what its worth, we typically write this type of accessor as `$sign_node->field_sign_orientation->value`.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -338,34 +323,124 @@ function sitenow_signage_preprocess_page(&$variables) {
...
-      $view_builder = \Drupal::entityTypeManager()->getViewBuilder('media');
+      $headline = $paragraph?->field_headline?->value ?? '';
 
-      foreach ($variables['content']['field_slide_image'] as $delta => &$item) {
-        if (is_numeric($delta)) {
-          $media_entity = $paragraph->get('field_slide_image')->get($delta)->entity;
-          if ($media_entity) {
-            $item = $view_builder->view($media_entity, 'vertical_9_16');
-          }
+      $filters['audiences'] = sitenow_signage_value_query_string($paragraph?->field_slide_events_audiences?->getValue());
```
pyrello:
```suggestion
      $filters['audiences'] = sitenow_signage_value_query_string($paragraph?->field_slide_events_audiences?->value);
```
Again, this is typically how we do this type of accessor unless there is a specific reason to use the `getValue()` method.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -338,34 +323,124 @@ function sitenow_signage_preprocess_page(&$variables) {
...
+          'items_per_page' => $amount,
+        ],
+      ];
+
+      $response = \Drupal::service('sitenow_events.content_hub_api_client')->getEventInstances($query);
+      if (!empty($response->events)) {
+        $rows = [];
+
+        foreach ($response->events as $key => $value) {
+          $event = sitenow_signage_object_to_array($value->event);
```
pyrello: You are transforming the data from an object to an array before using it. Is there a reason it is returned as an object in the first place?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -446,3 +681,68 @@ function sitenow_signage_outlook_resource_allowed_values() {
...
+ *
+ * @param mixed $obj
+ *   Any object.
+ *
+ * @return array|mixed
+ *   Returns an object that is being converted into an array
+ *   until the complete object is only an array, then it
+ *   returns an array.
+ */
+function sitenow_signage_object_to_array(mixed $obj): mixed {
```
pyrello: I'm a bit skeptical that this is necessary. Seems like something that might already be handled by a core Drupal utility or helper function.

GaryRidgway: Currently the thing we've used is `json_decode(json_encode($obj))`. However, I was reading online about how this worked and this was 1.5-2x faster that that. I'm okay with going back to the former though if we dont want the extra code.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
GaryRidgway: I wanted to implement it similar to `getEvents()`, and that one has the same return type. Keeping in mind the intent to eventually stop using the `sitenow_events_load()` function, I thought keeping it consistent in the API client seemed the most reasonable.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
GaryRidgway: The `->value` accessor only gets the first value of a multi option field. `->getValue()` here allows us to get an array of all the selected options.

GaryRidgway: <details><summary>Here is a screenshot of the debug evaluation for convenience.</summary>
<p>

Screenshot 2025-09-03 at 2 26 45 PM


</p>
</details>

docroot/modules/custom/sitenow_signage/sitenow_signage.module
GaryRidgway: That helper function only takes a paragraph, and is more intended to traverse up and get the orientation of a sign that contains the paragraph that we pass as an argument.

Perhaps it should be named something like, `sitenow_signage_get_paragraph_sign_orientation()`?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
pyrello: I'm not opposed to introducing a new helper function to handle this if there is not something that already exists that does the same thing. But let's not bury it in the signage module where it is unable to be used by the rest of the platform.

GaryRidgway: I propose moving it to `uiowa_core` then, if that's okay with you?

docroot/modules/custom/sitenow_signage/sitenow_signage.module
pyrello: `ContentHubApiClient` is definitely a work in progress. The `getEvents()` method doesn't represent anything definitive about how it should work. As your code suggests, returning a result set as an object doesn't work very well. Generally speaking, I think it would be ideal to have individual items returned as `stdClass` or some bespoke class meant to specifically represent it, while collections of items should be returned as arrays, unless we develop or implement a better collection class to serve that purpose.

GaryRidgway: I think that makes sense to me, I can format it as an array in the client first then!

docroot/modules/custom/sitenow_signage/sitenow_signage.module
pyrello: I'm good with this for now. Thinking long term it might be better as a trait.

GaryRidgway:
Feedback from EOD:

- Need to add back the headline, accidentally nuked it after importing.
- Cache busting, hourly, daily, check other slidetypes for inspiration.
- Icon fixing, need to be larger, centered.
- Start support article.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
pyrello: I'd have to look at this a bit more closely, but it seems like a very specific helper function. Maybe it could be broken into two pieces: one function to get the sign for a paragraph slide entity and another that gets the orientation for a sign entity.

GaryRidgway: Completely true. It was originally factored out of some code that Joe wrote for the outlook stuff that I also needed as well, so that is how it started. Breaking it into two seems like it would provide a lot of value!

docroot/themes/custom/uids_base/scss/components/uiowa-bar.scss
```diff
@@ -58,7 +58,7 @@
 
 // styles for html.html.twig
 
-.uiowa__container {
+.uiowa__container:not(:empty) {
```
bspeare: I think we can revert this because of https://github.com/uiowa/uiowa/pull/9134?

joewhitsitt:
still need to test commencement, but wanted to point these things out:

`Events`

Screenshot 2025-09-09 at 11 21 15 AM

changed to `Campus Events`?

Screenshot 2025-09-09 at 1 27 09 PM

---

Alignment issue with no results text

Screenshot 2025-09-09 at 1 23 21 PM

---

Maybe over time we would have a better idea, but I feel like we could squeeze a couple more events in. Not a blocker. Something to consider later.

Screenshot 2025-09-09 at 1 24 56 PM
Screenshot 2025-09-09 at 1 25 42 PM

briand44:
I am not up to speed on this thread so my comment can be disregarded but I did notice the other day that the support email for the events team was campus-events-support@uiowa.edu.

pyrello:
> Alignment issue with no results text

This has been addressed.

> Maybe over time we would have a better idea, but I feel like we could squeeze a couple more events in. Not a blocker. Something to consider later.

I think its possible that we may end up deciding to make the text even bigger to make it clearly visible, so I'm good with waiting on that to see before making adjustments.

joewhitsitt (approved):
Approving based on testing earlier and the recent commits. Re-tested the commencement import and found it working. Lookin' good.

For the record, I'd vote for `UI Campus Events` or `Campus Events` instead of just `Events` for the add button and paragraph wrapper label.

pyrello:
> For the record, I'd vote for `UI Campus Events` or `Campus Events` instead of just `Events` for the add button and paragraph wrapper label.

I went with Campus Events. I'm not a fan of prefixing things with `UI` because it feels redundant to me.

PR #9095, opened 2025-08-13
[Hotfix] OnIowa Events date issues

e-marie-w:
# How to test
```
ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli
```
- visit https://oniowa.uiowa.ddev.site/events
- check that the date filter stays a select list upon clicking an option instead of turning into a date picker
- check that switching from a select date back to 'Any date' does not return past events
- check that using other filters does not break anything


e-marie-w:
Sent email checking if they'd like the Reset button removed as part of this work, but I'm going to get this up for review in the meantime.

docroot/sites/oniowa.uiowa.edu/modules/oniowa_core/oniowa_core.module
```diff
@@ -9,12 +9,15 @@ use Drupal\Core\Form\FormStateInterface;
...
 /**
- * Implements hook_form_alter().
+ * Implements hook_form_FORM_ID_alter().
  */
-function oniowa_core_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+function oniowa_core_form_views_exposed_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   // Use the form['#id'] over $form_id because it is more specific.
-  switch ($form['#id']) {
-    case 'views-exposed-form-events-page-upcoming':
+  $foo = $form_id;
```
joewhitsitt: Remove the foo and bar and i'd call this good

e-marie-w: Good catch, thanks!

joewhitsitt (approved):

PR #9104, opened 2025-08-15
Signage - Outlook Events Listing

### LINKED ISSUE #5142: Signage - Show organizerName instead of event summary
# Story

As a signage editor, when placing an outlook events pane, I'd like to configure it to show the Organizer (organizerName) instead of the event title/summary regardless of whether or not the event is marked as private.

See its-web > Joe "Example private events for future Signage functionality" email thread

# Proposed solution

- [x] Add a checkbox that prints the organizerName instead of the event summary for all events if organizerName exists. Leave existing isPrivate functionality as-is.

### LINKED ISSUE #5183: Signage/Exchange enhancements
>Could I suggest two future enhancements?
It would be handy to be able to adjust the font size.  The current font seems a bit small, and I’d like to see what a larger font would look like.
When no events are listed, could the sign say something like “No more events today” instead of just being blank?

Requested by Mike Jenn.

### LINKED ISSUE #9008: Ability to display Outlook events on sign
# Story

As a user of Digital Signage, I would like the ability to display Outlook events on my sign.

# Additional Info

Example - https://signage-dev.uiowa.edu/hssb/hssb-schedule

Currently has ability to have title, choose room, number of items to display and whether or not to display end time and date (if different day)

# Proposed solution

joewhitsitt:
Resolves #9008 
Resolves #5183
Resolves #5142

# How to test

```
git checkout signage_outlook_events && git pull && ddev blt frontend
```

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli admin/config/sitenow/hours
```
- Start by adding Signage as the uiowa_hours group
- Go to /node/add/slide and you should see an Outlook Events slide type now.
- You should be able to add multiple resources and see the output. (carry over from current signage)
- Change the sign orientation and see the styles and number of items change. (new in favor of number of items)
- Change to toggle end times (carry over from current signage)
- If a end date is tomorrow or later, the date time format changes (carry over from current signage)
- Change to use organizer name regardless of whether or not the event is private. (new)
- In Chrome's Emulator, you can set the viewport to 1600x900 and 900x1600 to see that the number of events should fit and not overflow. Because of potential wrapping, the max of 10 (horizontal) and 15 (vertical) was my best guess for now.
- Slide view should be a normal table, Sign view, the table should be dark mode.

## Test for regressions

```
ddev blt ds --site=recserv.uiowa.edu
```

Check to make sure the hours blocks on the homepage work. Go to https://recserv.uiowa.ddev.site/campus-recreation-wellness-center and make sure the date picker works.


bspeare:
@joewhitsitt is there an opportunity for https://github.com/uiowa/uiowa/issues/7122 here?

joewhitsitt:
Could, but then it would be a slippery slope on where to stop. I was hoping to address the whole platform in a single PR. There is an example where it could be programmatically applied instead of relying on field formatters. https://github.com/uiowa/uiowa/pull/8537/commits/bcd69d41c889a859c9158642b12796fe70390098

I can adjust most of the format to be in better shape but the period between am/pm and the month abbreviation rules need that module.

bspeare:
I wonder if we want to do a quick BEM cleanup of the classes here?  I created some of this mess so I'd be happy to do a quick organization cleanup if you'd like the help. 

```
.signage {
  // Main signage. 
}

.signage--empty {
  // Empty state modifier
}

.signage--vertical {
  // Vertical orientation modifier 
}

.signage__table {
  // Table styles
}

.signage__slideshow {
// Slideshow element
}

```

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -34,6 +34,153 @@ function sitenow_signage_preprocess_field(&$variables) {
...
+            'signage-table',
+          ],
+        ],
+      ];
+      $markup['container']['table'] = [
+        '#theme' => 'table',
+        '#caption' => $headline,
+        '#header' => $header,
+        '#rows' => $rows,
+        '#empty' => t('No scheduled events for this resource.'),
```
bspeare: Should this be something more like "There are no scheduled events to display" or something?  The "resource" part is a little weird even though I know that is the terminology used.

bspeare:
Is `admin/config/sitenow/hours` for admins only?  The "Resource" select list shows all the rooms on the node/add/slide without configuring `admin/config/sitenow/hours` - is that correct?

bspeare:
For sites using the signage feature, it looks like they will have access to the hours block in layout builder (in addition to the logo block that we should remove as well).  Do we want to suppress that?  

Screenshot 2025-08-22 at 3 53 21 PM

joewhitsitt:
>Is admin/config/sitenow/hours for admins only?

Yes. People trying to place the block see a message that the group needs to be set but get an access denied if they go to it.

If this is the feature, the group "Signage" will need to be setup to properly use this functionality. The resources will come into the form, but the API call to fetch the data expects the group to be set.

>Do we want to suppress that?

I didn't think we needed to. Same will be for uiowa_maui and the presence of academic dates block for the astra room schedules.

docroot/modules/custom/sitenow_signage/sitenow_signage.module
joewhitsitt:
```suggestion
        '#empty' => t('There are no scheduled events to display.'),
```

bspeare:
If recserv.uiowa.edu wanted to use the signage feature, do they use the `FusionRecServ` group or do they need to use `Signage`?

bspeare:
Seeing a difference between prod and local for recserv.  Prod displays the closed information for "Fitness East" while local is not. 

Screenshot 2025-08-22 at 4 25 48 PM

Screenshot 2025-08-22 at 4 25 41 PM

docroot/modules/custom/sitenow_signage/sitenow_signage.module
```diff
@@ -34,6 +34,153 @@ function sitenow_signage_preprocess_field(&$variables) {
...
+          $start = strtotime($event['iso8601Start']);
+          $end = strtotime($event['iso8601End']);
+
+          // Skip events already ended.
+          if ($end < time()) {
+            continue;
+          }
+
+          // Format time range based on settings.
+          $formated_date = date('g:i a', $start);
```
bspeare: Should we change this to `$formatted_date` for correct spelling?

e-marie-w:
Can you remind me what the expected behavior is regarding reloading? It seemed like just reloading the page didn't update the results on the sign or slide pages and that I had to resave each to see today's results instead of yesterday's. But maybe that's only because cron doesn't run locally? I know some results will drop off in 30 minutes, I can try running cron then and seeing if the sign updates as expected.

e-marie-w (approved):
I followed testing instructions, the new slide type works and looks great for both sign orientations. No regressions on the recserv site for me. Awesome work!

joewhitsitt:
Thanks @e-marie-w!

PR #9107, opened 2025-08-18
[medicine.uiowa.edu] Remove redirects install module

### LINKED ISSUE #9093: Remove module or gut install hook
Follow up to https://github.com/uiowa/uiowa/issues/9039

From #9076 

>Seems like we will need a follow up slice to either remove the module or gut the install hook as the next deployment uninstalls the module making it possible to enable again and re-run.

cory-skeers:
Resolves #9093 

# How to test

Code review
```
ddev blt ds --site=medicine.uiowa.edu
```
Clean sync

The module has already been enabled and uninstalled, so there shouldn't be any interactions


pyrello (approved):
Reviewed and tested. Works as expected.

PR #9124, opened 2025-08-22
Signage video slide

### LINKED ISSUE #9004: Ability to display video on sign
# Story

As a user of the Digital Signage, I would like the ability to place a video on my sign.

# Additional Info

Examples:

- https://signage-dev.uiowa.edu/admissions/admissions-2
- https://signage-dev.uiowa.edu/hancher/performing-arts-iowa-ucc-signage

Support YouTube and Vimeo. UiCapture?

Current functionality has an option for YouTube playlist. Not sure YouTube playlists work in our current Remote Video media type?

# Proposed solution

pyrello:
Resolves #9004 

# How to test

Grab the API key from https://esaccess.iowa.uiowa.edu/ under Web Services > API Keys > YouTube.  In your `local.settings.php` file add this entry:
```
$config['sitenow_signage.settings']['youtube_api_key'] = '***';

```

1. Code review

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /node/add/slide
```
### Test adding a video on the slide content type
1. Test that you can add an image, outlook, room schedule and the duration field is visible. 
2. Test adding a Vimeo (https://vimeo.com/351594821) and Youtube slide (https://www.youtube.com/watch?v=KAwJ4fDdnGo).  
3. Confirm that you can add a slide and that the duration field is hidden and save the node. 
4. Comment out the library in `sitenow_signage_form_alter`, clear cache and edit the slide again.  Change the "Duration" to "Other" and confirm that there is a seconds value saved in that field. 

```
https://sitessignage.uiowa.ddev.site/node/add/sign
```
### Test with multiple slides and videos
1. Create a sign with multiple slides.  Add two Vimeo videos (https://vimeo.com/481386833 plus the previous) and two Youtube (https://www.youtube.com/watch?v=iYv2KBtE7e4 plus the previous) videos.   Add an image slide too. 
4. Test out that the duration field is hidden for videos but shows for all other slide types. 
5. Once the sign is saved, inspect the source for `data-splide-interval` and confirm that is matches the custom duration for each video. 
6. Confirm that the videos autoplay, loop, and are muted and only start playing when they are advanced to in the slideshow. 
7. Confirm that multiple loops of the slideshow behave the same way.

### Test IEF "Other" show/hide

1. Confirm you can select "Other" duration for slides for everything but videos.  Edit multiple IEF slides at once and confirm that the "Other" will show/hide the duration field on all but videos (it should be hidden for videos). 
2. In IEF, test (on a slide type that is not a video) that you can save the slide with an "other" value then edit the slide and select a predefined option and save the IEF and the node.  Edit the node again and switch back to "Other" and confirm that the value is empty for that slide. 
3. Confirm that if the "Other" field is selected and left empty that the default `1500` duration is set on the splide data attribute (dev tools inspect). 

### Test with single looping video
1. Test having only a Vimeo or Youtube video and confirm that it will loop over and over. 
2. Confirm that single video will just loop over and over without help from the duration field. 
3. Test with Youtube playlist (https://www.youtube.com/watch?v=we8DheSUU5o&list=PL7d8BMotflM2S6cSaJD8AiZUdwuZaV_to) and confirm that it will loop when finished (Playlist duration will be respected if it is in slideshow, it will only play the first video) . 
5. Confirm either single video doesn't do any "Splide" slideshow fading when it loops.
6. Test with campus cam: https://www.youtube.com/watch?v=SDQW6kWL1kY (campus cam in a slideshow with other slides will not work and get the default `1500` duration).

### Test permissions
1. Masquerade as a sign manager and confirm that you can add a remote video on a sign.


joewhitsitt:
I think this is clever, but is the juice worth the squeeze here? Not sure it is completely working as-is either.

`Error: Call to a member function getParent() on null in sitenow_signage_preprocess_media__remote_video() (line 48...` when adding a new video to a new slide.

- Do we get the same reusability benefit of these being media entities?
- If we want to keep it as a media entity and not build it off of a string like in D7, does it make sense to try and hijack the existing default view mode or should we consider a custom view mode and target it that way instead of referring item?
- Could the parsing be simplified if we just get the necessary id similar to Panopto and construct our own iFrame using `html_tag` type `iframe`?

joewhitsitt:
Controls are showing on load/start and then go away. Doesn't seem to be caused by mouse hover.

pyrello:
A couple things I noticed while testing this:
- YouTube playlists don't seem to work as expected. Or at least I couldn't figure out how to embed one. After thinking about this a bit, I'm wondering if we should explicitly exclude playlists from this functionality because figuring out and setting the duration is going to be a challenge.
- As the admin user 1, I was not able to access groups I was not a member of when setting a group via the audience field. Not a big deal, but may be something we want to address down the road.

joewhitsitt:
Personal opinion, but I feel like more should be done with the help text/fields for the different behaviors (e.g. 15 seconds is set for duration but if there is only one slide, it will continue to loop but as soon as a second slide is added, that is not the behavior anymore). 

In addition to having the user find the duration of their video we are asking them to convert it to seconds. Do we want to convert the input for them? 3.44 becomes 224? What about longer videos though? Should the input format change to allow 1:03:44 type input?

bspeare:
@joewhitsitt can you explain this in a little more detail?  

> Personal opinion, but I feel like more should be done with the help text/fields for the different behaviors (e.g. 15 seconds is set for duration but if there is only one slide, it will continue to loop but as soon as a second slide is added, that is not the behavior anymore).

joewhitsitt:
>Confirm that single video will just loop over and over without help from the duration field.

>Confirm that the first slide (and all slides) in the slideshow follows your custom slideshow duration for each slide.

>For video slides, please select the "Other" option and set the duration to match the length of your video.

Just with all of this, it is a little confusing that with a video item selected for the slide the duration field doesn't change. So if I skip over the help text and save the slide with the default 15 second duration, and only have one slide. The 15 seconds does not apply. Once I add another slide, it does apply but doesn't tell me that is the behavior.

Having help text explaining that the video selected will be muted and looped under specific conditions would be helpful.

Is it easy to figure out the duration of a playlist or should we encourage playlists be a single slide sign?

bspeare:
@joewhitsitt thanks for the further information.   I added some more detailed help text at https://github.com/uiowa/uiowa/pull/9124/commits/ad7950bc43cd59fdea27f2114fa94bf101b2f72a and https://github.com/uiowa/uiowa/pull/9124/commits/e113dacef29a40e008c44332a9b1b2ce31affe38.

joewhitsitt:
Pushing out to DEV for PO review

e-marie-w:
For anyone else testing the videos listed, it should take 9:52 to go through all the videos + a ten second image slide. 
Here's my breakdown if you want to do it in the same order and just let a stopwatch run through the loop
Start with image slide 0:00
First vimeo vid starts 0:10
Second vimeo vid starts 0:40
First Youtube vid starts 2:21
Second Youtube vid starts 8:04
Video ends, returning to image slide 9:52

joewhitsitt:
Screenshot 2025-09-09 at 8 53 49 AM

Thanks @bspeare. Hope we don't need it but it would be good to have just in case.

Now that I've seen the Other field persisting on a change, I can't unsee it. Everything else looks good though.

joewhitsitt (approved):
Just tried the latest commit and it worked. That plus everything previously, I think this is stable to ship

config/features/signage/field.field.node.slide.field_slide_widget.yml
```diff
@@ -39,10 +43,11 @@ settings:
         weight: -16
         enabled: true
       slide_image:
-        weight: -17
+        weight: -13
         enabled: true
       slide_video:
         weight: 7
+        weight: -17
```
pyrello: This seems incorrect.

bspeare: Should be fixed in https://github.com/uiowa/uiowa/pull/9124/commits/95e206b016aa2182eafa463da15ecf192e259b44.

pyrello: Nevermind

PR #9127, opened 2025-08-25
[Facilities] Sitenow Dispatch bug fix

### LINKED ISSUE #9117: Facilities Dispatch confirmation email bug
# Problem

When subscribing to the Facilities Alert Dispatch population through the Facilities website (https://facilities.uiowa.edu/alert-sign-up), the subscriber isn't being sent a confirmation email so they never receive any of the email notifications.

# Steps to reproduce

Go to https://facilities.uiowa.edu/alert-sign-up and sign up.

View https://apps.its.uiowa.edu/dispatch/populations/574308527/members and see that the subscribed column says pending and verify you did not receive any confirmation emails.

Image

# Additional Info

It doesn't appear this is happening on the SiteNow Announcement sign up form. The main difference is that we don't use any custom fields.

From Dispatch:

> We’re seeing a NullPointerException on our end related to the customFields in the Dispatch API.
> Feel free to send me the request body you're using when calling the API - I can review it and let you know if anything’s missing or if there’s a structural issue with the JSON.

# Proposed Solution

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9117. 

# How to test

```
ddev  blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli admin/config/sitenow/sitenow-dispatch
```
1. Grab dispatch key from https://facilities.uiowa.edu/admin/config/sitenow/sitenow-dispatch (unmask field using devtools)
2. Enter api key locally at https://facilities.uiowa.ddev.site/admin/config/sitenow/sitenow-dispatch. 
3. Go to https://facilities.uiowa.ddev.site/alert-sign-up to test
4. Test with a 10-digit phone number, Dispatch won't accept the submission if you use a number that doesn't follow the NANP Standard.  It must follow the NANP Standard (https://www.sent.dm/resources/pa), and the middle three digits cannot start with a 0 or 1 (https://www.quora.com/Why-cant-the-exchange-code-the-middle-three-digits-of-a-phone-number-start-with-0-or-1-For-example-why-would-234-056-7891-and-234-156-7890-not-be-valid-numbers).  
5. View submission at https://apps.its.uiowa.edu/dispatch/populations/574308527/members. 
6. Confirm you received an email.
7. Delete your email from https://apps.its.uiowa.edu/dispatch/populations/574308527/members and test again with options listed below.
8. Test the phone field with `123` or `319.056.7890` and confirm that the form validation displays an error message.
9. Test the phone field with `319.555.5555`, `(319)555-5555`  and confirm special characters are stripped out and submission goes through and you receive an email. 
10. You could also test again using different combinations fields, you just need to delete your name from https://apps.its.uiowa.edu/dispatch/populations/574308527/members.


joewhitsitt:
Pausing a moment. Should we consider using something like https://github.com/giggsey/libphonenumber-for-php so we don't have to be experts on phone number validation and formatting?

Also, I imagine we will need to use this again at some point in the future. Should we move this to something like docroot/modules/custom/uiowa_core/src/FormHelpersTrait.php and call it from there within `validateForm`?

bspeare:
@joewhitsitt great question, I'm open to that direction. I https://github.com/giggsey/libphonenumber-for-php the library to use or do we need to do more investigation there?

joewhitsitt:
I saw that mentioned in a couple articles that was either "use regex to do your own logic or offload it to this library"

https://robots.net/tech/how-to-validate-a-phone-number-in-php/
https://skynix.co/resources/guide-to-validating-phone-numbers-in-php
https://www.techboo.net/how-to-validate-phone-numbers-in-php

https://github.com/giggsey/libphonenumber-for-php-lite is the lite version of it that might be better since we don't need "geolocation, carrier information and short number info"

bspeare:
Thanks for the links.  The library seems focused on international validation: https://github.com/giggsey/libphonenumber-for-php-lite/blob/main/docs/PhoneNumberUtil.md.   Maybe we should do some minimal regex and update to a library if we find that we need a stronger solution?

joewhitsitt (approved):
Thanks again for offloading some of this and making it reusable. It works as expected and handles more situations where they deviate from the 10-digit format.

At first I thought the validation and formatting could happen in the same method, but I could see situations where we want to check and not immediately fix.

bspeare:
Thanks @joewhitsitt!

PR #9134, opened 2025-08-28
Add conditional on page_bottom wrapper in html template

### LINKED ISSUE #9123: Empty div on full screen sign causes scroll bar
# Problem
Empty div at the bottom of the sign causing a scroll bar for 16:9 display.

Image

# Steps to reproduce
Chrome emulator, view the sign as unauthenticated and adjust the resolution to 2560x1440

# Proposed solution

- [ ] Determine what is causing the present but empty div and Cut. It. Out.

joewhitsitt:
Fixes https://github.com/uiowa/uiowa/issues/9123

# How to test

- We don't use `page_bottom` for much but to test this, you can switch between an anonymous sign display and an authenticated backend path. For the anonymous sign display, you no longer see the empty div `<div class="uiowa__container"></div>` and in the backend you do. The tour module loads in page_bottom so that is why this template properly renders the page_bottom div when authenticated.
- Uses the same pattern as https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-page-title-hero.html.twig#L38
- This does cause a change to the Percy baseline will need approval to merge.

### DIFF
--- a/docroot/themes/custom/uids_base/templates/layout/html.html.twig
+++ b/docroot/themes/custom/uids_base/templates/layout/html.html.twig
@@ -57,9 +57,7 @@
 
 {{ page_top }}
 {{ page }}
-{% if page_bottom %}
-  <div {{ uiowa_attribute.addClass(header_classes) }}>{{ page_bottom }}</div>
-{% endif %}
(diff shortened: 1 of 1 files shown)


e-marie-w (approved):
Empty div is no longer displaying on signs for logged out users, awesome work!

PR #9137, opened 2025-08-29
[Signage] Signage groups list block

### LINKED ISSUE #9027: List of all Signage groups
# Story

As someone interested in the Signage v2 platform, I would like to see a list of all existing groups, so I can understand if a group already exists for my unit or if I need to be requesting a new group.

# Additional Info

Currently a user who is signed in to the Signage v2 service can see all the groups by going to the content overview but it could be helpful for folks who do not yet have access to be able to see all signage groups. This could also help reduce our support burden as we could share that with folks who have questions.

Current example - https://signage.uiowa.edu/all-groups

Do we just output as text, or do we link to the group page? It seems like group pages are available to view for non-authenticated users.

cory-skeers:
Resolves #9027 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /user/71
```
See that the list of user groups to which the user is assigned is sorted alphabetically, as compared to https://signage.sites.uiowa.edu/user/71

Edit or add a new page, and add the Signage Groups list block, and see that it functions as described in #9027

### DIFF
--- a/config/sites/signage.sites.uiowa.edu/config_split.config_split.site.yml
+++ b/config/sites/signage.sites.uiowa.edu/config_split.config_split.site.yml
@@ -23,6 +23,7 @@ complete_list:
   - field.field.node.slide.og_audience
   - node.type.signage_group
   - robotstxt.settings
+  - views.view.signage_groups
   - views.view.signage_group_content
   - views.view.user_signage_groups
 partial_list:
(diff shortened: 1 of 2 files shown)


e-marie-w (approved):
I followed the testing instructions, it looks and works as expected! I added some extra signage groups locally to double-check the pager and it works as designed. Alphabetizing the Signage groups on the user page is a nice addition. Awesome work!

PR #9068, opened 2025-07-30
Medicine htaccess redirects update

james917:
# How to test

This is available to test out in DEV:
https://medicine.dev.drupal.uiowa.edu/

> URLs starting “https://www.medicine.uiowa.edu/iowaprotocols\” need to redirect to the same path at “https://iowaprotocols.medicine.uiowa.edu”. As an example, the link needs to be changed from https://medicine.uiowa.edu/iowaprotocols to https://iowaprotocols.medicine.uiowa.edu.

So for testing on DEV, note that the redirect rule is formatted like several others in that it covers our internal *.drupal.uiowa.edu domains.

- https://medicine.dev.drupal.uiowa.edu does not redirect
- https://sandbox.dev.drupal.uiowa.edu/iowaprotocols/ does not redirect
- https://medicine.dev.drupal.uiowa.edu/iowaprotocols goes to https://iowaprotocols.medicine.uiowa.edu/
- https://medicine.dev.drupal.uiowa.edu/iowaprotocols/content/asdf goes to https://iowaprotocols.medicine.uiowa.edu/protocols/asdf
- anything else under /iowaprotocols/*  (e.g. https://medicine.dev.drupal.uiowa.edu/iowaprotocols/nursing-protocols) goes to https://iowaprotocols.medicine.uiowa.edu/protocols/*
- https://iihr.dev.drupal.uiowa.edu/igs/geosam/well/33775/logs should still redirect to https://igs.iihr.uiowa.edu/igs/geosam/well/33775/logs

PR #9141, opened 2025-09-02
removing core_htaccess path line from composer running ddev composer …

### LINKED ISSUE #9114: Redirect opsmanual.uiowa.edu to policy.uiowa.edu
# Story

As an owner of the Policy website, we would like to make sure traffic going to opsmanual.uiowa.edu will get to policy.uiowa.edu once the site is launched.

# Additional Information

They are targeting Sept 8/9/10th for launch.

- Is an .htaccess redirect that would keep the url fragment an option for this site?
Based on Parking Lot conversation, it sounded like the team supports an .htaccess redirect that would pass along the url path.

For example, https://opsmanual.uiowa.edu/about should redirect to https://policy.uiowa.edu/about.

Policy folks have aligned url paths where possible or setup redirects on new site in order to limit the number of page not founds.

Htaccess patching method - https://github.com/uiowa/uiowa?tab=readme-ov-file#updating-core-patched-files
>Changes to .htaccess and other global files should be reviewed with extra attention as they can affect all websites across the SiteNow platform.

https://htaccess.madewithlove.com/ - used in past for testing. Last time there were issues with some browsers. Safari seemed to work but Chrome was having issues. - https://github.com/madewithlove/htaccess.madewithlove.com/issues/95

# Proposed solution

- [x] Establish htaccess redirect rule
- [x] Add opsmanual.uiowa.edu domain to uiowa04 once policy.uiowa.edu has gone live.

james917:
…install.

# How to test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -387,7 +387,6 @@
             "blt internal:create-project:init-repo"
         ],
         "post-drupal-scaffold-cmd": [
-            "patch -p1 < patches/core_htaccess.patch",
             "patch -p1 < patches/development_services.patch"
         ],
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
--- a/docroot/.htaccess
+++ b/docroot/.htaccess
@@ -2,20 +2,6 @@
 # Apache/PHP/Drupal settings:
 #
 
-# Block these IP addresses.
-# https://docs.acquia.com/cloud-platform/arch/security/restrict/#blocking-by-ip-with-mod-rewrite-in-htaccess
-<ifmodule mod_setenvif.c>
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -387,7 +387,6 @@
             "blt internal:create-project:init-repo"
         ],
         "post-drupal-scaffold-cmd": [
-            "patch -p1 < patches/core_htaccess.patch",
```
pyrello, line 390: This doesn't seem right.

joewhitsitt:
yep. this needs to be redone. output should be similar to https://github.com/uiowa/uiowa/pull/9068/files (no change to composer.json, only our addition to .htaccess, and a complete diff of all custom htaccess rules in the .patch file.

james917:
Incorrectly ran ddev commands

PR #9149, opened 2025-09-04
Aggregator/RSS Feed Block - Register cache tags for proper invalidation

### LINKED ISSUE #9139: RSS Feed block caching not bubbling up with aggregator updates
# Problem
It appears that our recent footer cache fix has revealed that the RSS Feed block does not properly get invalidated when aggregator updates or adds new items.

# Steps to reproduce
As anonymous visitor to a page with an RSS block on it, if the underlying aggregator feeds are updated (delete items). Nothing tells the block and then page to clear site/varnish.

# Workaround

Re-save the page the RSS Feed block is placed on.

# Proposed solution

- [x] Obtain cache tags for the aggregator content in the feed and implement as part of the block render.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9139

# How to test

- On `main`
- `ddev blt ds --site=engineering.uiowa.edu`
- As an anonymous user, go to https://engineering.uiowa.ddev.site/about/employment-opportunities (at the time of writing this, there should be five items listed)
- Login and go to https://engineering.uiowa.ddev.site/admin/config/services/aggregator, click the feed operations button and select delete items.
- As an anonymous user, reload https://engineering.uiowa.ddev.site/about/employment-opportunities
- Because cache did not invalidate, it should still show five items.
- Switch your branch to `aggregator_cache_fix`
- `ddev drush @engineering.local cr` and reload the anonymous page. This should remove the five items but also establish the cache tag.
- On https://engineering.uiowa.ddev.site/admin/config/services/aggregator, click the feed operations button and select update items. The same five items should import.
- As an anonymous user, reload https://engineering.uiowa.ddev.site/about/employment-opportunities
- It should go from no items back to five items.

### DIFF
--- a/docroot/modules/custom/uiowa_aggregator/uiowa_aggregator.module
+++ b/docroot/modules/custom/uiowa_aggregator/uiowa_aggregator.module
@@ -5,6 +5,8 @@
  * Primary module hooks for uiowa_aggregator module.
  */
 
+use Drupal\aggregator\Entity\Feed;
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
On `main` deleting all items in the feed did update properly for anonymous, but I was able to recreate the behavior of the feed not updating with newly updated items. After recreating the issue that way, the cache tag update here fixed it.

Nice fix!

PR #9155, opened 2025-09-05
[Signage] Additional info in slide IEF

### LINKED ISSUE #9145: Display additional slide info in IEF on signs
# Story

As a Signage v2 sign manager, I would like to see additional slide info at a glance while editing a sign so that I can quickly see which slides exist on my sign, their duration and if they are still scheduled to display.

# Additional Info

We had some feedback asking for this from an early tester.

Displaying additional information through inline entity form could also help ensure folks chose the correct existing slide.

Current Signage example:

Image

Signage v2 example:

Image

It would be great to include slide type, thumbnail if applicable (image/video?), scheduling info, duration.

I believe we have used `hook_inline_entity_form_table_fields_alter()` to do this in D7 and looks like we are using it for articles/people.
https://github.com/uiowa/uiowa/blob/ff77148902e7882e21d1de0fbd24d8c9a9e28295/docroot/modules/custom/sitenow_articles/sitenow_articles.module#L639
https://github.com/uiowa/uiowa/blob/147808ed3e1d510570917d1a4a260770a3d227c9/docroot/modules/custom/sitenow_people/sitenow_people.module#L452

# Proposed Solution

cory-skeers:
Resolves #9145 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli
```
Edit a sign or create a new sign
Check out the new additions/changes to the IEF display of slides attached to the sign
Test adding/removing/editing slides
Test with a variety of slide types, included or empty publication data, etc.
Create and edit a slide and see that the form is unchanged
Consider other adjustments we may want to do, including https://github.com/uiowa/uiowa/pull/9155/#issuecomment-3292895241 (get the dropdown actions option to work style-wise, or other adjustments)


cory-skeers:
Current iteration:

image

I'm undecided on the image. Currently it includes a thumbnail if it's an image slide, otherwise lists the slide type's label. I did it to see if I could, not sure yet how much value it adds over just listing it as "image".

I'm also going to look at maybe including the status as well as scheduling all in the same column, with some bolding for the label and see how that looks.

joewhitsitt:
what about putting the scheduling info into a details element?

bspeare:
@cory-skeers can everything be smallified?

cory-skeers:
Needs a bit of cleanup (like getting the duration key and not value), but example of the full small:

image

cory-skeers:
Opening this up for review; any suggestions are welcome and I can either get them in place or help point where they can go.

pyrello:
> Updated small example, with pushed changes: image

A few nits:
- Personally I'd prefer `Schedule` to `Publishing Schedule`.
- I also think we could change `Publish on` to `Publish` and the same with `Unpublish on`
- Is it hard to adjust the date format to something like `mm/dd/yy hh:mm am`?

bspeare:
I know that I had suggested going `small` here but having some second thoughts.  Our use of the tag doesn't really meet the use case for https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/small and it is probably not a great move for accessibility.  I think part of the reason why it doesn't look great is because of Claro, I just tested this on Gin and it looks a lot better.  Maybe we back out of the `<small>` changes for now?  

Screenshot 2025-09-15 at 2 36 15 PM

cory-skeers:
> I know that I had suggested going `small` here but having some second thoughts. Our use of the tag doesn't really meet the use case for https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/small and it is probably not a great move for accessibility. I think part of the reason why it doesn't look great is because of Claro, I just tested this on Gin (with small tags removed) and it looks a lot better. Maybe we back out of the `<small>` changes for now?

Gin looks pretty nice!

Fair point on the purpose of the `<small>` tag. I tried taking them out, but did leave the `button--small` class on the buttons. I think we could still do either that or the dropdown patch that still helps some.
image

joewhitsitt:
pulling to test

joewhitsitt (approved):
I've been testing this off and on and the recent changes seem to be a happy compromise of info for now. Thanks for trying all the things on this and getting it to here @cory-skeers.

PR #9160, opened 2025-09-08
[Bug] Image caption stretch fix

### LINKED ISSUE #9150: Caption on image component moves further below from its image in the grid when columns grow
# Problem
When using the "Media format" setting in the image component set to "Widescreen (16:9)," "Ultrawide (21:9)," "Circle," "No crop," or "Square"  in a two column section, if the column opposite of its placement starts growing in height, the caption starts drifting lower and lower as more content on the left is added. See video below:

https://github.com/user-attachments/assets/5466202f-62d0-4853-a00e-911e2e3055b1

Interestingly, the image caption seems to be presented completely differently when "No Crop - fit height" is selected and doesn't exhibit this issue:
Image

"Circle" sized images don't seem to exhibit this issue either.

# Steps to reproduce

1. Make a two column section
2. Add a text area on the left column with a lot of text
3. Add image component to the right column with an uploaded image and set its "Media format" to be  "Widescreen (16:9)," "Ultrawide (21:9)," "Circle," "No crop," or "Square" and add a caption in the right column
4. Note the caption being far below the image.

# Proposed solution
Investigate the use of  `display: table-caption;` for this sort of caption, since removing that style and reverting it to `display: block` seems to fix this. Since this isn't really a table caption, it may not be necessary. Alternatively, investigate why the "No Crop - fit height" option is displaying completely differently.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9150. 

# How to test

```
ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /image-block-0
```
1. Compare https://sandbox.uiowa.ddev.site/image-block-0 against https://sandbox.prod.drupal.uiowa.edu/image-block-0 and confirm that image caption stays below the image in all of the examples.


james917 (approved):
Approved

bspeare:
Thanks @james917!

PR #9162, opened 2025-09-08
Signage - add entity browser permission

joewhitsitt:
https://iowaweb.slack.com/archives/C0164FNN1PV/p1757354300638369

Screenshot 2025-09-08 at 12 57 40 PM

# How to test

As a sign manager user, you can add/edit a sign and select the existing slide button that shows the entity browser and not an access denied message.


bspeare (approved):
This fix works as described.  I was able to access the existing slides browser while masquerading as a sign manager.

PR #9170, opened 2025-09-10
[Bug] File overwrite issue

### LINKED ISSUE #6817: Overwrite original file checkbox is not overwriting old file
# Problem
From its-web:
> When I choose ‘replace file” what I expect is that the filename will remain the same, but the contents will be replaced with the file I’m choosing to upload...but it doesn’t seem to actually work that way.

I was able to duplicate with an interesting caveat: at first it did overwrite and replace correctly. Once the file was used on a page, however, I could not override the original file. I deleted every usage of the file, including past revisions, but even with the file not in use it would no longer override.

Note: I only tested it with pdf files.

# Steps to reproduce
Upload a file or find an existing one. Use the file somewhere on the site. Mark the 'overwrite original file' checkbox on file and attempt to upload a new file. Save and view the file. It will still show your old file instead of the new one.

# Proposed solution
- [ ] Using the patch htaccess method, lower the ExpiresDefault from 1 year to 10 min and commit the changes to a draft PR.
- [ ] Checkout the branch on DEV and deploy.
- [ ] Setup test situations between PROD and DEV with a pdf and image to see if the file replace functionality works better for browsers that have already download previous version. Varnish invalidation should happen on the file paths through sitenow_media_wysiwyg_media_update()
- [ ] Repeatedly ping the file to understand if the varnish caching layer is in place.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/6817 

# How to test

Looking for code review. We previously tested 10 minutes in DEV.


joewhitsitt:
whoa. I think we need to set a FilesMatch around this so that it doesn't set this header on every path. get file types from each local media entity (image, video, audio, document)?

GaryRidgway (approved):
Code has been reviewed

Issue #1245, opened 2020-03-24
Add "_blank" target option to external links in menus


pyrello:
Currently, there is no way to have external links open in a new browser tab/window in SiteNow. I think this would be the desired behavior for all the external links on uiowa.edu.

joewhitsitt:
https://www.drupal.org/project/menu_link_attributes

pyrello:
When we discussed this a while back, I think that @mjoneill was indicating that having all external links open in a new page is no longer considered best practice.

joewhitsitt:
This module provides attributes on a case-by-case basis. If that is no longer desired as well, I think this issue can be closed.

cory-skeers:
For reference: https://www.w3.org/WAI/WCAG21/Techniques/general/G200

Generally, they recommend against it unless there is a context and/or security reason for doing so, and to ensure there are indicators prior to the link to being clicked. Looks like the menu_link_attributes would work fine for this, if there are enough use-cases to match the above guidelines.

pyrello:
I'm going to close this for now. If we get requests for this functionality, we revisit this issue.

PR #9178, opened 2025-09-16
Signage - Warn user about the use of `prefers-reduced-motion` with signage.

### LINKED ISSUE #9169: Improve prefers reduced motion experience on Signage
# Story

As a Signage viewer, I would like to be able to view all of the slides on a sign while still respecting my prefers reduced motion setting.

# Additional Info

Currently if the prefers reduced motion is on, slides do not rotate.

Provide fading effects or other less complex animations.

>We probably need override this https://github.com/uiowa/uids/blob/4.x/src/scss/uids-core.scss#L35 for signage.

https://splidejs.com/guides/options/#reducedmotion

Update docs https://its.uiowa.edu/services/digital-signage/sign-slideshow-not-rotating-between-slides to align with the solution here.

# Proposed solution

pyrello:
Resolves #9169 

# How to test
```
ddev blt ds --site signage.sites.uiowa.edu \
&& ddev drush @sitessignage.local cset uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active \
&& ddev drush @sitessignage.local cr \
&& ddev drush @sitessignage.local uli sign/osc-sign
```
- Code review.
- Confirm you can see the new warning message on the sign node page styled like an alert warning and in the correct position on the page.
- Edit the sign node and confirm that you see the new warning message in addition to the message about the sign node being saved.
- Confirm that you cannot see the warning message while logged out.

### Logged in
- A user should be able to see all the regular site regions when viewing a sign.
- Hawk Alerts display in the normal location
- Signs do not reload.

### Logged out
- While viewing a sign node, confirm that all website regions except the region containing hawk alerts are removed from the rendered page, consistent with how the sign should look on a signage display.
- Hawk Alerts get displayed consistent with the outcome from #8807.
- Signs are reloaded at the appropriate interval.
- Adding `?signage-display` to the URL has no effect on the rendered output for sign nodes.


pyrello:
I'm wondering if the way through this is just to add the warning for everyone and not use the JS messages API, leave the 10.3 workaround in place, and then create a follow-up to figure out how to best get back to a state where we don't need the workaround

joewhitsitt:
pulling to test.

one thing I noticed though is the target blank on the link. don't we want to avoid doing that as it is no longer considered the best practice? https://github.com/uiowa/uiowa/issues/1245

joewhitsitt:
Hmm. I am not seeing the hawk alerts loaded in the source or on top of the video. 
https://sitessignage.uiowa.ddev.site/sign/osc-sign

Screenshot 2025-09-19 at 11 54 32 AM


When I refresh the logged in version, the reduce motion message disappears.
Screenshot 2025-09-19 at 11 53 55 AM

pyrello:
Setting back to draft to investigate issues that @joewhitsitt found while testing.

pyrello:
> Setting back to draft to investigate issues that @joewhitsitt found while testing.

I believe that the issues identified previously have been fixed. I'm not 100% on setting the `$variables['#cache']['max-age'] = 0` to get the message to show on multiple loads, but its probably okay since this is a temporary solution and it shouldn't be there that long.

joewhitsitt (approved):
Yes, I want to move away from that max-age=0 but it seems to do the trick for both issues I experienced. All else looks good.

PR #9179, opened 2025-09-16
Slide hotfix

GaryRidgway:
Adding sign checks so that slide pages don't whitescreen.

## To test
- Go to some slides
- See that they work on their own.
- Got to some slides on signs where they should be affected by the code change
- ensure they are still affected on signs


pyrello:
The idea I had for handling this differently was by registering and using an entity bundle class. We previously used this for adding card rendering to various node teasers. The nice thing about it is that it encapsulates the logic in a file that is all logic relevant to the node bundle type. Good future enhancement.

joewhitsitt (approved):
does the job

PR #9193, opened 2025-09-22
Policy pdf download

### LINKED ISSUE #9002: Ability to create PDF archive of entire Policy Manual on a yearly basis
# Story

As an owner of the Policy Manual website, we need the ability to create a PDF of the entire Policy Manual each year because it is a requirement of the Iowa Administrative Code and also the General Counsel's Office frequently needs archived policies for lawsuits, etc.

# Additional Info

Currently they use a Firefox browser extension to convert everything on http://opsmanual.uiowa.edu/printable to a pdf.

>Yes, (usually) once a year I use the firefox page, save each "page" as a pdf, and then use acrobat to combine and then and do some minor customizing (footers, page #s, etc).

# Proposed solution

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9002

# How to test

```
ddev blt ds --site=policy.uiowa.edu && ddev drush @policy.local uli --uid=6 admin/config
```

- As the webmaster you logged in as, you should see [Download Policy Manual](https://policy.uiowa.ddev.site/admin/pdf-content) among the other SiteNow config links.
- On first run since the PDF does not yet exist, clicking the link takes you right into the batch process which after loading the nodes and combining them as html takes a bit longer (at 100%) to finish and takes you do another route where you can download the file. Click download and inspect the PDF.
- The outline of the document should match that of the main menu system under the Table of Contents section https://policy.uiowa.ddev.site/admin/structure/menu/manage/main
- The menu parent (e.g. II. Community Policies) is added where available and in the instances where it would be "Table of Contents," "Policy Manual" is added instead.
- The styles should provide a clear output of content, hiding certain elements common to policy and adding in Zilla Slab and Roboto fonts.
- Additionally, the beginning of each "page," commonly chapters in this situation should be a page break resulting in about 935 pages in the PDF ending with `Appendix: Organizational Charts`
- Back on the website, you should be able to click the breadcrumbs to return back to the main status page where is shows that the existing file is ready for downloaded. You can re-download the file here to test.
- You can also visit the main menu and disable a parent within the Table of Contents tree. On output that page along with the nested children should not be part of the output.

```
Testing /var/www/html/docroot/sites/policy.uiowa.edu/modules/policy_core/src/tests/Kernel
Time: 00:00.640, Memory: 4.00 MB
OK (1 test, 14 assertions)
Process finished with exit code 0
```


bspeare:
@joewhitsitt would it worth going back to the unstyled version that you started with and sprinkle in some fonts?  That version looked a lot cleaner and might remove the headache of trying to use the uids_base theme.

bspeare:
@joewhitsitt should I expect the PDF to look like the linked example in https://github.com/uiowa/uiowa/pull/9193#issuecomment-3335499661 or did it change after that?

joewhitsitt:
...shoooot...

```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/vendor/dompdf/dompdf/src/Css/Style.php on line 1116

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 81920 bytes) in /var/www/html/vendor/composer/ClassLoader.php on line 576
```

joewhitsitt:
Bummer. I am now seeing what Ben is seeing locally in DEV.

joewhitsitt:
rubbing eyes to see

I believe fonts are working now!

https://policy.dev.drupal.uiowa.edu/sites/policy.uiowa.edu/files/exports/policy-manual.pdf

docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php
```diff
@@ -166,27 +166,30 @@ public static function batchFinished($success, $results, $operations): void {
...
     }
 
-    $print_styles = '';
-
-    // Include local fonts.
-    $path_resolver = \Drupal::service('extension.path.resolver');
-    $path = $path_resolver->getPath('module', 'policy_core');
-    $font_css = DRUPAL_ROOT . '/' . $path . '/css/pdf.css';
-    $print_styles .= '<style>' . file_get_contents($font_css) . '</style>';
+    $fonts = '<link rel="preconnect" href="https://fonts.googleapis.com"><link rel="preconnect" href="https://fonts.gstatic.com" crossorigin><link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100..900;1,100..900&family=Zilla+Slab:ital,wght@0,300;0,400;0,500;0,600;0,700;1,300;1,400;1,500;1,600;1,700&display=swap" rel="stylesheet">';
```
joewhitsitt: @bspeare I copied this straight from Google Fonts. I doubt we need all of these weights right?

bspeare: Maybe?  You can use https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/uids_base.libraries.yml#L8 as a reference.

bspeare:
@joewhitsitt I'm noticing that the block headlines aren't being included here from https://policy.uiowa.edu/governance/university-iowa, is that intentional?
**Web page:**
Screenshot 2025-10-03 at 9 16 36 AM

**PDF:**
Screenshot 2025-10-03 at 9 16 29 AM

joewhitsitt:
welp. that was it for the most part. just need to implement a solution cleanly/correctly without touching anything else.

Screenshot 2025-10-03 at 11 57 28 AM

docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php
```diff
@@ -0,0 +1,331 @@
...
+    $context['results']['temp_files'][] = $temp_file;
+    $context['message'] = t('Processed @count nodes in this batch.', ['@count' => count($nodes_chunk)]);
+  }
+
+  /**
+   * Batch finished callback.
+   */
+  public static function batchFinished($success, $results, $operations): void {
+    // Increase the memory limit temporarily for PDF generation.
+    ini_set('memory_limit', '512M');
```
bspeare: Will we need this for prod?  What is our current `memory_limit`?

bspeare (approved):
Nice work! This is now functioning as expected.  

A few notes:
- Some areas are still using `sub` and `sub` which will be need to be changed to `small` to avoid running together. 

Screenshot 2025-10-03 at 1 07 26 PM

- Some links are broken and we might want to consider adding something like https://github.com/uiowa/uiowa/blob/main/docroot/sites/admissions.uiowa.edu/modules/admissions_core/sass/pdf.scss#L83-L90. 
- Parentheses are not included in the PDF.  Hopefully that won't be an issue.

docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php
briand44: 256M - https://sandbox.prod.drupal.uiowa.edu/admin/reports/status

joewhitsitt: Good question. Ran into the issue in DEV (128). Maybe it won't be an issue in prod. I'd be willing to take this out until we know it is a problem there.

joewhitsitt:
>Some areas are still using sub and sub which will be need to be changed to small to avoid running together.

You mean that should be a content change and not something to address here?

We can add more CSS, it was just that pulling in the complete set of global-styling in uids_base has some assumptions or is expecting certain parent classes.

bspeare:
> > Some areas are still using sub and sub which will be need to be changed to small to avoid running together.
> 
> You mean that should be a content change and not something to address here?

- Just to note for Eileen. 

> We can add more CSS, it was just that pulling in the complete set of global-styling in uids_base has some assumptions or is expecting certain parent classes.

- I don't think we should add global styling.  Just wanted to note about the links, parentheses, and sup/sup.

bspeare:
@joewhitsitt yes push that up and I will test it.

joewhitsitt:
>What do you think about that last commit?

Nice idea. I like it.

Issue #8421, opened 2025-01-17
Superfish menu block still renders when all of the menu items are not enabled


bspeare:
# Problem
Brandon shared this report: I came across an issue with the site https://mlk.uiowa.edu/. Because this is a one page site, it doesn’t have any main navigation items. I’m getting an accessibility error in Siteimprove because the superfish unordered list is empty. [https://help.siteimprove.com/support/solutions/articles/80001051793-sia-r68-container-element-is-empty-explained](https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fhelp.siteimprove.com%2Fsupport%2Fsolutions%2Farticles%2F80001051793-sia-r68-container-element-is-empty-explained&data=05%7C02%7Cits-web%40uiowa.edu%7C131ee0e7d5654d7c7dc208dd3672c25f%7C1bc445959aba4fc3b8ec7b94a5586fdc%7C1%7C0%7C638726588697158957%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=qpGK14wMre%2Fo0hkpO5OE%2FBZ%2BlmgBzvWiVLCdheyFvZg%3D&reserved=0)

# Steps to reproduce

- Go to https://mlk.uiowa.edu/. 

# Proposed solution

quamsta:
@bspeare 

This also may be occurring when there are unpublished pages in a menu as well. We are working on moving some content under "Education -> Fellowships" in https://pathology.medicine.uiowa.edu that isn't published yet and I noticed the same SiteImprove errors for an empty container

<img width="1606" alt="Image" src="https://github.com/user-attachments/assets/e2f51f0d-3c8f-4f96-9676-93526b71dc3a" />

joewhitsitt:
I think this is intersecting with a couple other issues. https://github.com/uiowa/uiowa/issues/7970 and https://github.com/uiowa/uiowa/issues/3406

Ultimately, if we continue the whack-a-mole with superfish, let me resurface this one: https://github.com/uiowa/uiowa/issues/4973

bspeare:
The empty nav can also been seen at https://signage.sites.uiowa.edu/ with the gray border appearing under the gold bar. 

[...]

PR #9195, opened 2025-09-23
[Bug] Fix for empty ul rendering for superfish

bspeare:
Related to https://github.com/uiowa/uiowa/issues/8421

# How to test superfish bug with empty `<ul>` rendering 

```
ddev blt ds --site=mlk.uiowa.edu && ddev drush @mlk.local uli /
```

1. Review code changes at https://github.com/uiowa/uiowa/pull/9195/commits/eff0998d0ea314dcc0c9517661b2c5e074eb72d2
2. Compare source code to prod, do you see the `<nav class="nav--horizontal" aria-label="Primary menu">` rendering?
3. Go to https://mlk.uiowa.ddev.site/admin/structure/menu/manage/main and mess around with publishing, unpublishing, and disabling menu items. 
   * If the menu contains only a disabled parent item with child pages that are published or enabled, there should be nothing rendered for the nav. 
5. Test as authenticated and anonymous and compare what you see rendering on the page. 
6. Test with different menu styles set at https://mlk.uiowa.ddev.site/admin/appearance/settings/uids_base. 

## Todo
- [x] I need to add an upstream bug issue to https://www.drupal.org/project/superfish - https://www.drupal.org/project/superfish/issues/3548286


e-marie-w (approved):
I tried to test most combos - both display inline and display below the IOWA bar, and then all three header navigation styles for both display options. Also briefly tested the back to top button, different branding options, and the sticky header for the toggle navigation option.

Consistently, if the menu items were disabled, they would not appear for authenticated and anonymous users. Also, the `<nav>` element would not render, as desired. If menu items were enabled but unpublished, they and the `<nav>` element appear for authenticated users but not anonymous, as expected. The previous two results remain the same even if a child page of a menu item is enabled and published.

I also did a code review, it looks good to me. Awesome work.

PR #9203, opened 2025-09-24
Immuno Grad view update

cory-skeers:
# How to test

```
ddev blt ds --site=immuno.grad.uiowa.edu && ddev drush @gradimmuno.local uli /admin/structure/views/view/research_areas
```
See that you get a clean sync
Log in and try to save the view--don't get an error
View the site and see that the People > Faculty > Research link is not in the main menu


GaryRidgway (approved):
works as described!

PR #9204, opened 2025-09-24
[Messages] Enable JS messages API in uids_base

### LINKED ISSUE #9185: Enable JS Messages API
# Motivation

As part of the 10.3 upgrades, we put in place [a fix for some system messages not getting the correct styling](https://github.com/uiowa/uiowa/commit/8d3b8e122ca0b438b287342904f168b652229dc5). Part of what happened with 10.3 is that the message system was reworked so that a greater number of messages get generated through the JS messages API that was added back in 8.8. The reason this was a problem for us is that for JS messages to be styled correctly, you have to do the theme implementation that applies the correct markup.

This is a follow-up issue to #9169, where it was discovered.

Drupal issue with workaround comment: https://www.drupal.org/project/drupal/issues/3456176#comment-15738200

# Proposed solution

- Implement JS message theming in UIDS Base
- Remove workaround add in 10.3 upgrade
- Remove static prefers-reduced-motion warning added in #9178 
- Add a dynamic warning message based on whether the user has `prefers-reduced-motion: reduced` set.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9185. 

**Here's my understanding of the Drupal message rendering system:**

Drupal has two different ways it renders status messages. The first is server-side rendering using our Twig template (`status-messages.html.twig`), which happens during the initial page build. The second is client-side rendering using JavaScript (the `Drupal.theme.message` function), which creates messages dynamically through AJAX operations like form submissions, BigPipe placeholders, or JavaScript-triggered events. In Drupal 10.2 and earlier, BigPipe handled messages specially and they were rendered through the Twig template. Starting in Drupal 10.3, a core change ([#3379885](https://www.drupal.org/project/drupal/issues/3379885)) made BigPipe use the JavaScript `MessageCommand` system instead, which means many more messages are now rendered client-side via JavaScript. This exposed a long-standing issue: themes that don't implement `Drupal.theme.message` in JavaScript will show unstyled messages. Both logged-in and logged-out users can receive JavaScript-rendered messages, but we see it more with logged-in users because they trigger more dynamic operations and have permissions to see detailed error messages with backtraces.

To get consistent message styling, we need three pieces: (1) our Twig template for server-rendered messages, (2) a JavaScript theme override (`Drupal.theme.message`) that produces identical HTML for JavaScript-rendered messages, and (3) a pre-render callback (via `hook_element_info_alter()`) that ensures our JavaScript library loads on every page and sets up the fallback placeholder container (`[data-drupal-messages-fallback]`) where dynamic messages get inserted. This is the Olivero approach. The pre-render callback doesn't convert messages—it ensures the JavaScript infrastructure is available so messages get our UIDS alert styling regardless of rendering method. The workarounds in this thread (like patching BigPipe or using preprocess hooks) are temporary fixes. The real solution being worked on is in issue [#3396318](https://www.drupal.org/project/drupal/issues/3396318), which would make JavaScript messages automatically use the theme's Twig template for rendering, eliminating the need for themes to implement `Drupal.theme.message` at all. Drupal 11.2 reduced how often BigPipe triggers JavaScript messages, making the problem less visible but not fixing the underlying issue.

**This pull request adds:**
1. A JavaScript theme override (`Drupal.theme.message`) that produces identical HTML for JavaScript-rendered messages.
2. A pre-render callback (via `hook_element_info_alter()`) that ensures our JavaScript library loads on every page and sets up the fallback placeholder container.
3. Updates the HawkAlert js to use the new JavaScript theme override (`Drupal.theme.message`). 

# How to test alerts
```
ddev blt frontend && ddev blt ds --site=default && ddev drush @default.local uli
```

- Code review
- Add this to `uids_base_preprocess_page`:
```
  \Drupal::messenger()->addStatus('This is a success message');
  \Drupal::messenger()->addWarning('This is a warning message');
  \Drupal::messenger()->addError('This is an error message');
```
- Confirm messages are dismissible
- Confirm messages display the same whether logged in or in a private browser
- Edit a page, make page updates and verify that messages display as expected. 
- Break a function somewhere and verify that the error message is formatted as expected (replace the current `superfish.theme.inc` with https://git.drupalcode.org/project/superfish/-/blob/832a1266590d4e5d0f7b9eb4d20596d12b612bb7/superfish.theme.inc if you need an example).
- Compare with prod/local and look compare markup and aria notifications.

# How to test prefers reduced motion warning

```
ddev blt frontend && ddev blt ds --site signage.sites.uiowa.edu \
&& ddev drush @sitessignage.local cset uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active \
&& ddev drush @sitessignage.local cr \
&& ddev drush @sitessignage.local uli sign/osc-sign
```

- Code review.
- In your mac system settings, search for "motion" and check "Reduce motion" in the Display settings. 
- Confirm you can see the new warning message on the sign node page styled like an alert warning and in the correct position on the page.
- Edit the sign node and confirm that you see the new warning message in addition to the message about the sign node being saved.
- Confirm that you cannot see the warning message while logged out.
- Turn off the "Reduce motion" in your Display settings and confirm that you do not see the message while logged in. 

# How to test hawk alerts

```
ddev blt frontend && ddev drush @default.local cset uiowa_alerts.settings hawk_alert.source https://emergency.stage.drupal.uiowa.edu/api/active && ddev drush @default.local cr
```
1. Code review
4. Compare the markup against `main` and note any differences. 
5. See if you can test getting the message to show up in the `drupal-live-announce` div. 

# Test other Drupal messages
```
ddev blt frontend && ddev blt ds --site=now.uiowa.edu && ddev drush @now.local uli
```
1. Try to submit webform and view errors https://now.uiowa.ddev.site/form/share-your-news-test-1
2. Try to submit invalid value at https://now.uiowa.ddev.site/experts-list - **~_Update: Looks like dismissible never worked here (check prod), I will see if I can fix this._~**


docroot/modules/custom/uiowa_alerts/es6/alerts.es6
```diff
@@ -82,16 +91,16 @@ Drupal.behaviors.uiowaAlerts = {
...
           existing.push(existingAlert.getAttribute('data-drupal-message-id'));
         });
 
         return existing;
       }
 
-      // Takes a JSON item and creates the markup for an alert.
-      // Returns a string of HTML.
-      function alertMarkup(responseJSONItem) {
+      // Takes a JSON item and creates just the CONTENT for a hawk alert.
```
pyrello:
```suggestion
      // Takes a JSON item and creates just the content for a hawk alert.
```

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
```diff
@@ -58,15 +58,24 @@ Drupal.behaviors.uiowaAlerts = {
...
+          // Get the hawk alert content.
+          const alertContent = hawkAlertContent(item);
 
-          // And add it to the Drupal messages section with its unique ID and message type.
-          messages.add(alert, {
+          // Add the content to Drupal's message system with warning type.
+          messages.add(alertContent, {
             id: id,
-            type: 'warning'
+            type: 'warning',
```
pyrello:
```suggestion
            type: 'error',
```
I think this would allow removing the code below that swaps out the alert classes.

docroot/themes/custom/uids_base/assets/js/message.theme.js
```diff
@@ -0,0 +1,93 @@
...
+    let icon = '';
+    if (type === 'status') {
+      icon = 'fa-check';
+    } else if (type === 'warning') {
+      icon = 'fa-exclamation-triangle';
+    } else if (type === 'info') {
+      icon = 'fa-info';
+    } else if (type === 'error') {
+      icon = 'fa-exclamation';
+    }
```
pyrello: It seems like this could be combined with the section above that sets alert classes based on `type`.

docroot/themes/custom/uids_base/templates/misc/status-messages.html.twig
```diff
@@ -41,11 +41,13 @@
     {% endif %}
   {% endset %}
 
+  {% set message_attributes = create_attribute() %}
   {% include '@uids_base/uids/alert.html.twig' with {
-    'attributes': attributes,
+    'attributes': message_attributes,
```
pyrello: Is there a reason to create a new variable here rather than just adjusting the existing variable or resetting it with 
```
{% set attributes = create_attribute() %}
```

docroot/themes/custom/uids_base/uids_base.libraries.yml
```diff
@@ -72,6 +72,11 @@ virtual_tour:
 status-messages:
   js:
     assets/js/uids.status-messages.js: { preprocess: false }
+drupal.message:
```
pyrello: Just curious, is it recommended to name this `drupal.message` because it is a library to override another library? If not, I'd be inclined to not include the `drupal.` part.

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
bspeare: I have it set as `warning` because I think we want the Drupal status message type to be `warning` but the styling is `error` which is why it is set like this.   If you look at the current version, `warning` is being set but the markup is for `error`: https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_alerts/es6/alerts.es6#L67.

docroot/themes/custom/uids_base/assets/js/message.theme.js
bspeare: Good suggestion, thanks!

docroot/themes/custom/uids_base/uids_base.libraries.yml
bspeare: I think we can remove it.  Thanks!

docroot/themes/custom/uids_base/templates/misc/status-messages.html.twig
bspeare: Good question.  I lost track of what I was encountering there but changing it to your suggestion seems to work fine.

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
bspeare: Maybe we should be going with the status type `error` here or switch to `warning`?  It is a little weird to to present the design of `error` with the Drupal message semantics of `warning`.

Current behavior uses `alert--danger` class:
```
<div class="messages messages--warning" role="alert" data-drupal-message-id="hawk-alert-2025-05-28T19:22:47+00:00" data-drupal-message-type="warning" aria-label="Warning message">
```

```
<div class="alert alert--icon alert--vertically-centered alert--danger" role="alert" aria-label="warning message" data-drupal-message-id="hawk-alert-2025-05-28T19:22:47+00:00" data-drupal-message-type="warning">
```
If we change this to `type:error`:
```
<div class="alert alert--icon alert--vertically-centered alert--danger" role="alert" aria-label="error message" data-drupal-message-id="hawk-alert-2025-05-28T19:22:47+00:00" data-drupal-message-type="error">
```

docroot/modules/custom/sitenow_signage/js/reduced-motion-warning.js
```diff
@@ -0,0 +1,18 @@
...
+ * @file
+ * JavaScript for the detecting if prefers-reduced-motion is enabled.
+ */
+
+(function (Drupal, once) {
+  Drupal.behaviors.signageReducedMotionWarning = {
+    attach: function (context, settings) {
+      once('reduced-motion-warning', 'html', context).forEach(() => {
+        const isReduced = window.matchMedia(`(prefers-reduced-motion: reduce)`) === true || window.matchMedia(`(prefers-reduced-motion: reduce)`).matches === true;
+        if (isReduced) {
```
pyrello, line 11: Seems like maybe we don't actually need to set a `const` here. The value of the of `isReduced` could be put directly inside the `if` condition. Not a strong opinion if you think that having a variable set make this more readable.

bspeare: I do like the format currently to make it easier to read but I can change it if you'd like!

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
pyrello: I think we want the error styling. The error semantics seem off, but I'm not sure that matters very much. Possibly we may want to adjust our `message.theme.js` script to allow separate styling and semantic cues. However, I'd be inclined to just change this to `error` for now and see if anything bubbles up.

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
```diff
@@ -82,16 +91,16 @@ Drupal.behaviors.uiowaAlerts = {
...
         });
 
         return existing;
       }
 
-      // Takes a JSON item and creates the markup for an alert.
-      // Returns a string of HTML.
-      function alertMarkup(responseJSONItem) {
+      // Takes a JSON item and creates just the content for a hawk alert.
+      // Returns a string of HTML content (no alert wrapper).
```
pyrello: The changes to this comment makes sense in the context of this PR but I think seems weird on its own.

docroot/modules/custom/uiowa_alerts/es6/alerts.es6
bspeare: I like that approach.  We could definitely adjust `message.theme.js`, but it would be nice to hold off until it is necessary.

docroot/modules/custom/sitenow_signage/js/reduced-motion-warning.js
pyrello: No, let's leave it.

pyrello:
Pulling to test since I've already been reviewing the code changes.

pyrello:
> Have you run `blt frontend`? Maybe we need a `cr` update hook with this pull request too?

I followed the instructions that included both `blt frontend` and `drush cr` so I don't think that's it. I can try running through them again.

pyrello:
> Try to submit invalid value at https://now.uiowa.ddev.site/experts-list 

This looks like it renders in a different area than the other messages. I'm just checking to make sure that it still renders correctly?

bspeare:
> > Try to submit invalid value at https://now.uiowa.ddev.site/experts-list
> 
> This looks like it renders in a different area than the other messages. I'm just checking to make sure that it still renders correctly?

Yes, the prod version can't be dismissed but this version should work.

pyrello (approved):
Reviewed and tested. Nice update. Thanks!

PR #9221, opened 2025-09-29
[Signage] Sign group login update

### LINKED ISSUE #9191: Admin Log In link on group page
# Story

As a Signage user, I would like to see a log in link on my group page so that it is less confusing on how to log in or where to go to do that from this page.

# Additional Info

Some feedback around this:
https://github.com/uiowa/uiowa/issues/9116#issuecomment-3307439989
https://github.com/uiowa/uiowa/issues/9116#issuecomment-3308195274

There currently is not a log in link on the group page since it is rendering using the admin theme.

Image

The log in button/link should pass the destination parameter similar to the footer link so folks will return to the group page and can add signs and slides to their group once they authenticate.

cory-skeers:
Resolves #9191 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /admin/people
```
Masquerade as a Sign manager
Click on the masqueradee's profile to see the signage groups to which they belong
Click on one of the signage groups, see that you can see the page as normal, with an admin theme
Open the signage group in incognito or as logged out, and see that you get an access denied that is not the admin theme, and that the "admin login" button correctly contains the destination parameter for the signage group


bspeare (approved):
All testing steps worked as expected.  Nice work!

PR #9222, opened 2025-09-29
Sprint 282

### LINKED ISSUE #9218: Deprovision of iyws.clas.uiowa.edu on uiowa01
Do not _reclaim_ the domain. It is pointed at csi-012.its.uiowa.edu as a redirect.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9218

# How to test

3 provisions, 1 deprovision. 
Check there are no more instances of robinson.lab..uiowa.edu or robinson_lab__uiowa_edu.


e-marie-w:
> We can correct the PR files before merge using the Github UI. The database just needs to be deleted and recreated as it is `robinson_lab__uiowa_edu`. I can assist if you want.

Some assistance would be great, especially with the database part. I'm going to provision another site and then I'll reach out about doing this.

joewhitsitt:
the directory name too

joewhitsitt:
>There is an uncommitted local settings file that still has the extra underscore in the database. I doubt we'll need to work with this site locally, but wanted to capture this info just in case.

That should just be your local and anyone else that pull this branch down. I believe re-running blt bis would fix it.

bspeare (approved):

PR #9228, opened 2025-10-01
Mazevo slide type.

### LINKED ISSUE #9188: Signage slide type to display Mazevo event schedule for rooms
# Story

As the Signage Group Manager for IMU, I need the ability to display Mazevo event schedules for multiple rooms on an individual slide so that folks entering the building know which room to go to for their event.

# Additional Info

Mazevo is the replacement for the EMS system that was previously used. They don't have any plans to move away from this system. There is some locations and features that aren't supported by Outlook.

The old Signage system currently has Mazevo integration and it is needed on the new system to successfully transition their existing signs.

Example of existing sign - https://signage.uiowa.edu/imu/east-lobby-information

**Configuration Options**

- Ability to Enter Headline
- Choose building
- Rooms
- ~Statuses~ - can pass in confirmed and not make this setting configurable.
- Types
- Option to display end time
- Option to show room column

Image

**Output**

- Headline or generic "Event Schedule"
- Time - start and end if configured to display end time
- Meeting - `eventName`
- Room - if configured to display
- Number of items determined by orientation (Outlook may be showing 10 for horizontal, 15 for vertical?)

Image

D7 code in `docroot/sites/all/modules/custom/uiowa_signage_modules/uiowa_signage_panes/plugins/content_types/mazevo_events.inc`

We should add documentation for this feature as part of this slice.

Was mentioned they had previously asked for a feature enhancement to display blocks of time. So display events for all rooms selected for the chosen time blocks. Not requirement of the initial slice of functionality but something to consider as a way to display all rooms events on an individual sign since there is a limit to number of items (previously rotated through all). Probably more conversation necessary if we were to invest here. https://github.com/uiowa/uiowa/issues/6322

GaryRidgway:
# How to test
- Ping me for an api key setup walkthrough.

## Mazevo
- do all the stuff to set up signage.sites.
- try making a Mazevo slide!
- Does the slide respect the room column boolean?
- Does the slide respect the End time boolean?
- Are you getting results with the options you can select?
- Please try and break things.

## Other API stuff.
We modified the `ApiClientBase.php` file, so we need to test some other places it is used.
- Check a Maui call and ensure it still works
- Check a Bizhub call and ensure it still works
- Check a Content Hub call and ensure it still works


docroot/modules/custom/sitenow_signage/src/MazevoApiClient.php
```diff
@@ -0,0 +1,99 @@
...
+
+use Drupal\Core\Cache\CacheBackendInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\uiowa_core\ApiAuthKeyTrait;
+use Drupal\uiowa_core\ApiClientBase;
+use GuzzleHttp\ClientInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * The Content Hub API service.
```
joewhitsitt:
```suggestion
 * The Mazevo API service.
```

docroot/modules/custom/sitenow_signage/src/MazevoApiClient.php
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public function basePath(): string {
+    return 'https://api-east.mymazevo.com/api/';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getCacheIdBase() {
+    return 'sitenow_signage_api';
```
joewhitsitt: should this be mazevo specific or are we going to use the same cache for everything?

docroot/modules/custom/sitenow_signage/src/MazevoApiClientInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Drupal\uiowa_core\ApiClientInterface;
+
+/**
+ * A Content Hub API client interface.
+ */
+interface MazevoApiClientInterface extends ApiClientInterface {
+
+  /**
+   * Get all buildings.
```
joewhitsitt:
```suggestion
   * Get events.
```

docroot/modules/custom/sitenow_signage/src/MazevoApiClientInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Drupal\sitenow_signage;
+
+use Drupal\uiowa_core\ApiClientInterface;
+
+/**
+ * A Content Hub API client interface.
```
joewhitsitt:
```suggestion
 * A Mazevo API client interface.
```

docroot/modules/custom/sitenow_signage/src/MazevoApiClient.php
GaryRidgway: Hmmm, I'm not sure. Most of the caching stuff I have been making has been copies of your work, so I don't have as much knowledge on why or why not we would choose one way or the other. If you were making the decision, what would you pick?

joewhitsitt: uh. i believe each cacheid has been specific to the api service (e.g. uiowa_maui, uiowa_hours:request, etc) meaning that this is MazevoApiClient and not a SignageApiClient. from a debugging standpoint that has been easier to track down.

GaryRidgway:
> something doesn't seem right. i only selected one room and one event type.
> 
> Screenshot 2025-10-07 at 2 28 47 PM Screenshot 2025-10-07 at 2 29 03 PM

Hmmm, not sure what's going on here. I will run some debug stuff

GaryRidgway:
@joewhitsitt Fixed! [It was an error in the way that I was accessing values from the filters.](https://github.com/uiowa/uiowa/pull/9228/commits/5ec91c6020a0e4e8d97fe65e1f9b7abd5fbb0be3#diff-473de3f533af1566ffcc06ea25caad37f1ebab58d5d456a541b8c18a3d39d337R409-R410)

joewhitsitt:
Nice fix. I still have concerns about using preprocess_field here compared to the other slide types.

>This gets confusing in preprocess_field. I know I used it for outlook and astra, but I think this should be done similar to events since there are multiple fields. For outlook I was looping over field items (multiple foreach ($variables['items'] as $item) {) instead of re-fetching the field values $paragraph?->get('field_slide_mazevo_rooms')?->value as you have. The other thing is that the rooms field is not required so it doesn't run through preprocess_field if empty and then doesn't display anything.

config/sites/signage.sites.uiowa.edu/field.field.paragraph.slide_mazevo_events.field_slide_mazevo_rooms.yml
```diff
@@ -0,0 +1,21 @@
...
+    - field.storage.paragraph.field_slide_mazevo_rooms
+    - paragraphs.paragraphs_type.slide_mazevo_events
+  module:
+    - options
+id: paragraph.slide_mazevo_events.field_slide_mazevo_rooms
+field_name: field_slide_mazevo_rooms
+entity_type: paragraph
+bundle: slide_mazevo_events
+label: Rooms
+description: 'Select which rooms you would like to display events for.'
```
joewhitsitt:
```suggestion
description: 'Select which rooms you would like to display events for. Leave blank to display events from all rooms.'
```

Since it isn't a required field. 

I prefer "space" over "room" but that is being discussed elsewhere.

joewhitsitt: >Space may be a more appropriate term but Signage v1 has always referred to them as rooms and nobody has asked for a change so I would lean towards staying with rooms until we are told otherwise. - bdickey

GaryRidgway:
@joewhitsitt the captions stuff should be pushed now!

joewhitsitt (approved):

Issue #9194, opened 2025-09-22
Find slide by id and slide sharing guidance


briand44:
# Story

As a Signage user, I would like some guidance on the recommended way of sharing content so I can bring more visibility to the information I want to share without requiring other users to duplicate images and/or slides.

<details>

<summary>Original feedback</summary>

>Question related to slide usage in the content inventory on the new signage system. 

>I just had someone in our department ask if a slide I created recently to promote "Switching to Duo Mobile" was picked up by anyone to display and wanted to keep their supervisor in the loop on where it was possibly being displayed. I know it wasn’t possible in the old system but assumed we could figure it out with the “Usage” tab similarly to how you can see what content is being displayed where on our regular Drupal sites. I didn’t see anything in the documentation at first glance in the support articles but did a little investigating and think there’s a way but wanted to reach out and see if this is the only way to figure this out. 

>From my searching–attached a screenshot of what I looked up and found–it looks like Tippie and Journalism may have picked up the slide in the new system–I had to take an additional step and search the authors hawkIDs to figure out where it may be displaying. However, if someone uploaded the slide on their own from where I shared them initially in the Digital Signage Teams channel they could have named the slide something completely different and it may be unable to be found and track down who is posting it. 

>In the past when we’ve emailed our slides to the listserv our only confirmation that it was getting picked up by a sign owner was if they reached out and requested a different dimension to be created for their specific display, but that didn’t always happen. I think it would be a nice feature that users could utilize for instances like needing to know how widespread across campus that the signage is actually getting displayed. Especially for when you have events and service promotions that you’re wanting to keep tabs on when deciding if more promotion is necessary.

>Thanks!
Jalessa Hohensee (she/her/hers)
</details>

# Additional Info

- Support article on the Signage service page (https://its.uiowa.edu/services/digital-signage) with recommendations for sharing Signage content.
  - Recommend sharing media (images)
  - Can share slides but should mention concerns about scheduling information, duration, updates, etc.
  - Promote Teams channel for communicating about your content you want shared and what information to provide
  - Let folks know they can use the usage tab on media and slides to see if others are using that specific content.

briand44:
## Grooming
 
**Quadrant:** 



 **Discussion:**
  - expose nid on slide node page (it exists at the end of the url alias path.)
  - make sure that when selecting a slide, you can find the slide by searching nid - fine to include in output as well
  - may want to adjust alias path so it is easier to explain how to find node id `slide/nid/title`?
[...]

PR #9229, opened 2025-10-01
Include slide nid and image mid for UX

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/9194

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli --uid=61 node/add/sign
```

- Add an existing slide and see that the entity browser doesn't link the node titles and has a ({NID}) appended to it. Filter on a number and see items returned back that partially match.
- While on the sign content type, note there is a alias path field visible in the sidebar.
- Navigate to add or edit a slide. You don't see the alias path field.
- With an image slide type, when selecting existing media, you can now filter by the media id and see the media id ({MID}) appended to the media name on the table display.
- Save the slide and see the new URL pattern.
- Navigate to your group member add page as see that the form alters are still working: https://sitessignage.uiowa.ddev.site/group/node/116/admin/members/add (hidden admin role, "Select if the member should have additional permissions within the group." help text)
- Navigate to https://sitessignage.uiowa.ddev.site/admin/content/media and see that the media has the mid treatment here too!
- Authors of media can get the media id from the URL similar to the slide and share with others.


james917:
I had an idea for the first step - Add an existing slide and see that the entity browser doesn't link the node titles and has a ({NID}) appended to it. Filter on a number and see items returned back that partially match.

Maybe create a new collumn for the {{NID}} to appear separate and distinct from the title. Maybe this can something down the road until we get feedback

joewhitsitt:
@james917 i went ahead and made that change for both slides and media:
Screenshot 2025-10-03 at 1 07 35 PM
Screenshot 2025-10-03 at 1 07 57 PM
Screenshot 2025-10-03 at 1 08 19 PM

james917:
It looks awesome 
Screenshot 2025-10-03 at 11 01 17 AM

james917 (approved):
Looks great approved

PR #9255, opened 2025-10-09
Signage - Search + group names link on all groups block

pyrello:
Adds linking from group names to the group pages. Adds a search filter.

# How to test

```
ddev blt ds --site signage.sites.uiowa.edu
```
- Visit https://sitessignage.uiowa.ddev.site/groups
- Confirm group names link to group pages. Group pages will still show access denied.
- Test search works.

### DIFF
--- a/config/sites/signage.sites.uiowa.edu/views.view.signage_groups.yml
+++ b/config/sites/signage.sites.uiowa.edu/views.view.signage_groups.yml
@@ -76,7 +76,7 @@ display:
           click_sort_column: value
           type: string
           settings:
-            link_to_entity: false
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Nice quick addition! Tested and works well. I added a quick No Results text--feel free to adjust the wording before pushing up.

I'd wondered for a bit if the contact email should be included in the search combine filter, but am leaning toward no. It's part of the table, so I don't think it would be out of place to include it, but at the same time, it's likely the unknown thing people are using the table to find, so searching by it to find groups seems odd.

PR #9260, opened 2025-10-09
Policy dompdf performance

### LINKED ISSUE #9002: Ability to create PDF archive of entire Policy Manual on a yearly basis

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9002

# How to test

- Code review

```
ddev drush @policy.dev uli --uid=6 admin/config
```

- As the webmaster you logged in as, you should see [Download Policy Manual](https://policy.dev.drupal.uiowa.edu/admin/pdf-content) among the other SiteNow config links. Click it and click the regenerate button.
- Await successful completion and download the PDF. Compare output to the menu structure ensuring every sections and chapters go in order. Starts with Governance and ends with Appendix. Chapters increment as expected in between.
- Font styles should be the same as before. ZillaSlab for the top headings and Roboto for the text.

The real test will be in PROD where this failed previously, but if the batch process does what it says, we shouldn't run out of time/memory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -211,6 +211,7 @@
         "drupal/workbench_email": "^3.0",
         "drush/drush": "^12.4.4",
         "giggsey/libphonenumber-for-php-lite": "^9.0",
+        "iio/libmergepdf": "^4.0",
         "kartsims/easysvg": "2.5",
         "npm-asset/accessible-menu": "4.x-dev#2481e91",
         "npm-asset/jquery-collapse": "^1.1",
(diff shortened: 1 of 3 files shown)


GaryRidgway (approved):
I have gone through and spot checked very many places. All chapters seem to be in order and the generator worked on dev! I think this can be approved.

PR #9268, opened 2025-10-13
Main navigation menu a11y update

### LINKED ISSUE #9254: Main menu a11y issues
# Issue

Container element is empty

Some roles are designed to contain other roles. This element has this type of role, but it doesn't contain any [required-owned elements](https://www.w3.org/TR/wai-aria-1.1/#mustContain).

Seems to be related to the main menu on SiteNow sites. Not seeing a `role="list"` though.

https://my2.siteimprove.com/Inspector/998093/A11y/Page?pageId=24733541800&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r68/failed/cXpidjV2ZEc=;OTNiODc2ZDg=

### LINKED ISSUE #9257: Role not inside the required context
# Issue

Role not inside the required context

This page element has a role that only makes sense in the context of another role.

https://my2.siteimprove.com/Inspector/998093/A11y/Page?pageId=24733541800&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r42/failed/RnRkMjdHNzk=;YWEyOGI1MGU=

May also be related to #9254

cory-skeers:
Resolves #9254 
Resolves #9257 

# How to test

Sync a site on `main` and use the Siteimprove browser extension scanner
See two issues related to the menu:

- "Role not inside the required context"
- "Container element is empty"

`git checkout superfish-role-update` and clear cache
Run the scan again, and see the two issues are no longer flagged.
Additionally, check my understanding/conclusion from https://github.com/uiowa/uiowa/pull/9268#issuecomment-3398655797. I think the lowest footprint approach here is to add the `role="menu"` in to account for the included `role="menuitem"`s, with the intention that we will be moving to a different library and structure for menus in the future, so further reliance or overriding of superfish-isms would be nice to avoid.
Another option would be to remove the `role="menuitem"` instead, but we would need to account for how `role="menu"` gets added to items which have submenus, or risk breaking other portions of the menu's aria attributes.

Note:
This fix mostly emulates the update in https://git.drupalcode.org/project/superfish/-/compare/8.x-1.8...8.x-1.9?from_project_id=6566 to `templates/superfish.html.twig` which also includes an aria label `aria-label="{{ 'Menu'|t }}"`. I've left this off in lieu of the `aria-label` we have in the parent `nav` element.


docroot/themes/custom/uids_base/templates/navigation/superfish.html.twig
```diff
@@ -18,7 +18,7 @@
 
 <nav role="navigation" aria-label="Site main navigation">
```
cory-skeers:
```suggestion
<nav aria-label="Site main">
```

cory-skeers: `role="navigation"` is assumed within a `<nav>` element. Similarly, screen readers will read the label along with the element, so including "navigation" in the label will be read as "Site main navigation navigation".

docroot/themes/custom/uids_base/templates/navigation/superfish.html.twig
```diff
@@ -16,9 +16,9 @@
  */
 #}
 
-<nav role="navigation" aria-label="Site main navigation">
+<nav aria-label="Site main">
```
pyrello: Sorry to nitpick this and maybe you're reading indicated this was the right way to go. I think the removal of the word navigation from the aria-label seems incorrect. This take stems from reading through the W3C disclosure menu example: https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation-hybrid/

In that example, they just put the name of the organization/website in the aria-label, which is an odd example I'll admit, but that makes me think that the label isn't necessarily interpreted as navigation-related. Since we have multiple menus on many pages, it seems appropriate to specify that it is navigation in the label.

If this change is based on some other specific guidance, feel free to ignore this.

joewhitsitt: >A brief description of the purpose of the navigation, omitting the term "navigation", as the screen reader will read both the role and the contents of the label.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role

pyrello: Maybe this should just be `Main`? The word `Site` doesn't really seem to add anything context-wise that is valuable.

cory-skeers:
```suggestion
<nav aria-label="Main">
```

GaryRidgway (approved):
I completely agree with the statements posed about reworking the NAV structure, but I also agree that it's probably not in this issue that we should be doing that. I think that for now the adjustments made makes sense and solve the immediate accessibility issue.

PR #9271, opened 2025-10-13
[Banner] Gradient midpoint and block form adjustments

### LINKED ISSUE #9230: implement banner adjustments into SiteNow
# Story

As a SiteNow user, I would like to use the recent options added to UIDS on my SiteNow site so I have more control on how gradients are used on my banners.

# Additional Info

UIDS changes from https://github.com/uiowa/uids/issues/987

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9230. 

# How to test

```
ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /
```
## Test default settings
1. Test that when a new banner block is added that all of the "Default" values are set and the banner block saves them as expected.  Compare with prod form to confirm previous default values. 
2. Test that when a block is created and saved without an image or background selected, when the block is edited again the "Color / pattern" is selected with a "Black" background color. 
3. Test existing blocks and make sure that all of their options remain selected and unchanged.

## Test update hook
1. Test that the update hook makes the proper selection for "Image / Video" and "Color / Pattern" for existing banner blocks. Look at https://sandbox.uiowa.ddev.site/node/916/layout and make sure the colors are selected. 
```
ddev composer install && ddev blt frontend && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/3141/layout
```
## Test new gradient midpoint checkbox 
```
ddev composer install && ddev blt frontend && ddev blt ds --site=tippie.uiowa.edu && ddev drush @tippie.local uli /
```
- Edit the Tippie home page banners
- Check the "Customize gradient midpoint" checkbox and adjust the gradient midpoint.  Save the page and confirm that the block gradient midpoint as changed. 

```
ddev composer install && ddev blt frontend && ddev blt ds --site=tippie.uiowa.edu && ddev drush @tippie.local uli /node/1/layout
```
- Adjust the gradient midpoint on the top to bottom, bottom to top, and left to right gradients
- Confirm block form values save and maintain selection after saving and editing again

## Test "Gradient options" show/hide states are working
- The "Customize gradient midpoint" checkbox should only appear if anything other than "No gradient" is selected in the "Media overlay" select. 
- Confirm that checking the "Customize gradient midpoint" checkbox reveals the radio percentage options. 
- Confirm that defaults are selected based on the "Media overlay" select.
- Double check that "Enable autoplay for video" checkbox shows for video upload.

## Test heading sizes update
- Compare prod https://tippie.uiowa.edu/brooke-wang-story-phd with local version and confirm font size scales down better than current version.  Review screenshots https://github.com/uiowa/uids/issues/987 and test other banners in small layouts. 

## Test that JS and form_alters are unsetting values
- Switch back and forth between Image and Color background.  Confirm values are unset after block save and just switching back and forth after uploading an image/video.  Test other options and confirm that values are unset as expected. 

## Layout builder style patch
- Confirm that we want to add this patch https://www.drupal.org/project/layout_builder_styles/issues/3290930. 

### Todo
- [x] P2LB updates? P2LB seems good, the only issue is if background color sections were converted to banners and they appear to be converted to text areas instead. 
- [x] Remove layout builder style for midpoint percentage
- [x] Create midpoint percentage field on the banner block with 20-70% options using input range slider for HTML element called `field_styles_gradient_midpoint`
- [x] See if we want https://www.drupal.org/project/range_slider/issues/3516294 for `field_styles_gradient_midpoint`
- [x] Update banner block twig template to render this field as an inline style like in https://github.com/uiowa/uids/pull/1010.
- [x] Update hook for new image/video color/pattern radio option.
- [x] Rearrange form into some details elements and remove borders from some fieldsets and adjust spacing to make each section look distinct possibly using prefix and suffix div wrappers to create a faux fieldset look.
- [x] Update banner-block-form.js to work with field rearrangements
- [x] Test if autoplay setting is broken


docroot/modules/custom/layout_builder_custom/layout_builder_custom.module
```diff
@@ -537,6 +527,24 @@ function layout_builder_custom_form_alter(&$form, FormStateInterface $form_state
...
+                $form[$container_name][$key_duplicate]['#options'][$settings['default']] .= ' (default)';
+              }
+
+              // Set layout builder style default settings.
+              if ($form_id === 'layout_builder_add_block') {
+                $form[$container_name][$key_duplicate]['#default_value'] = [$settings['default']];
+              }
+            }
+          }
+        }
```
pyrello: It seems like this is necessary because the form alter method is creating the duplicates before defaults are getting set? Potentially we could adjust when the form alter happens to avoid this being necessary.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,548 @@
...
+      $form['gradient_options']['layout_builder_style_banner_gradient_midpoint_duplicate']['#weight'] = 4;
+    }
+
+    // Layout group.
+    $form['layout_group_heading'] = [
+      '#type' => 'html_tag',
+      '#tag' => 'h3',
+      '#value' => t('Layout'),
+      '#weight' => 94,
+      '#attributes' => ['class' => ['layout-builder-style-heading']],
```
pyrello: 'layout-builder-style-heading' could probably be replaced with something related to the `off-canvas` since that is actually what we are adapting to? Also, if there is a BEM-ish class name we could use, that would be good too.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,548 @@
...
+    // Check if gradient midpoint field has a value.
+    $has_midpoint_value = !empty($form['layout_builder_style_banner_gradient_midpoint']['#default_value']);
+
+    $form['layout_builder_style_adjust_gradient_midpoint'] = [
+      '#type' => 'checkbox',
+      '#title' => t('Adjust gradient midpoint'),
+      '#default_value' => $has_midpoint_value,
+      '#weight' => 94,
+      '#attributes' => [
+        'class' => ['adjust-gradient-checkbox'],
```
pyrello: It would be nice to see this as a BEM-ish style class name instead of something this bespoke.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,548 @@
...
+
+    if (is_array($selectedStyles)) {
+      foreach ($selectedStyles as $selectedStyle) {
+        if (array_key_exists($selectedStyle, $color_pattern_options)) {
+          $default_bg_type = 'color-pattern';
+          $default_bg_value = $selectedStyle;
+          break;
+        }
+      }
+    }
```
pyrello: I'm not sure if some of this is new, but I think that some/all of this can be removed in favor of just removing the `#access=>FALSE` constraint on the LB styles background form element.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
bspeare: Added `off-canvas` class.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,613 @@
...
+   * Validates the banner block form.
+   *
+   * @param array $form
+   *   The form array.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public static function validateForm(array &$form, FormStateInterface $form_state) {
+    // Sync duplicated fields to original fields.
+    $field_mappings = [
```
pyrello: It seems like this entire list is just mapping a value to the same value but  with `_duplicate` added to the end. I'm wondering if we need this list to be a key/value list or if it could just be a list of the current keys and the sync function "maps" it by adding the `_duplicate` part?

docroot/modules/custom/layout_builder_custom/css/layout_builder_custom.overrides.css
```diff
@@ -127,6 +127,99 @@
...
+  padding-inline-start: 0;
+  margin-block-start: 0;
+}
+
+#drupal-off-canvas-wrapper fieldset:not([data-drupal-ck-style-fence]) {
+  margin-block-end: 0;
+}
+
+/* Block form group container - creates visual grouping */
+#drupal-off-canvas-wrapper .block-form-group__container {
```
pyrello: Wondering if we want to change the naming of this `block-form-group__container` class to be something more generic for off-canvas? It seems like more of an off-canvas forms thing than block form specific.

bspeare: Does `off-canvas-group` work as the base?  Feel free to suggest something if you have some ideas.

pyrello: If you think it will have some utility outside of form styling, I think that works. Otherwise, maybe `off-canvas-form-group`?

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,755 @@
...
+    // =========================================================================
+    // CREATE ALL FORM GROUPS/CONTAINERS
+    // =========================================================================
+    // Block heading.
+    $form['banner_group_heading'] = [
+      '#type' => 'html_tag',
+      '#tag' => 'h2',
+      '#value' => t('Banner'),
+      '#weight' => -20,
+      '#attributes' => ['class' => ['heading-a']],
```
pyrello, line 59: What is this `heading-a` class?

bspeare: It's a class from the `off-canvas` styles.  https://git.drupalcode.org/project/drupal/-/blob/10.0.x/core/misc/dialog/off-canvas/css/base.css

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
```diff
@@ -0,0 +1,733 @@
...
+    $element['background_type'] = [
+      '#type' => 'radios',
+      '#title' => t('Background type'),
+      '#options' => [
+        'media' => t('Image / Video'),
+        'color-pattern' => t('Color / Pattern'),
+      ],
+      '#default_value' => $default_bg_type,
+      '#weight' => 10,
+    ];
```
pyrello, line 548: This section seems like it could be in the `formAlter` method. Any reason you chose to put it here?

bspeare: I don't recall the reason so feel free to move it!

pyrello:
This is ready for testing/review again.

docroot/modules/custom/layout_builder_custom/src/BannerBlockFormHandler.php
joewhitsitt: @pyrello did you decide to leave this? is this resolved?

joewhitsitt (approved):
After previous testing and review of recent changes, this looks good. Feels monumental. Great work!

I'll create an issue for https://github.com/uiowa/uiowa/pull/9271#issuecomment-3481678555 if someone else can confirm.

PR #9282, opened 2025-10-17
Splide/Signage Reduced Motion Override

joewhitsitt:
# How to test

On `main`

```
ddev blt ds --site=signage.sites.uiowa.edu
```

As an anonymous user visiting https://sitessignage.uiowa.ddev.site/sign/master-mcu-onstage, open up Chrome's inspector tools, go to More Tools -> Rendering and turn on the prefers reduce motion option. Reload the page and wait. The slides do not transition.

On `splide_reduced_motion_override`

`ddev blt frontend && ddev drush @sitessignage.local cr`

Reload the page. It should transition the slides.

### DIFF
--- a/docroot/modules/custom/sitenow_signage/js/slideshow.js
+++ b/docroot/modules/custom/sitenow_signage/js/slideshow.js
@@ -7,6 +7,23 @@
   Drupal.behaviors.signageSlideshow = {
     attach: function (context, settings) {
       context.querySelectorAll('.signage__slideshow').forEach(function (element) {
+
+        // Override prefers-reduced-motion for Splide.
+        const originalMatchMedia = window.matchMedia;
+        window.matchMedia = function (query) {
+          if (query === '(prefers-reduced-motion: reduce)') {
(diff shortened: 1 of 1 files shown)


joewhitsitt:
The PO has requested this get tested to understand if this is suitable for rolling out if necessary.

cory-skeers (approved):
Reviewed and tested, and I think this works as a path forward for this.

Re:
> Also, was wondering if
> 
> https://github.com/uiowa/uiowa/blob/fc48a08880fc76dcdd71556aff0ca8bebc127203/docroot/modules/custom/sitenow_signage/js/reduced-motion-warning.js#L10
> 
> is redundant in what it is checking. They look identical.

I think it should be specifically using the `window.matchMedia(`(prefers-reduced-motion: reduce)`).matches === true` with the `matches` property, but I think checking the `matchMedia` will also resolve the same way...So I think it can be reduced to just use the second half as the "most accurate" check. Not 100% sure on my reasoning there though.

PR #9284, opened 2025-10-17
[Signage] Mazevo event status

### LINKED ISSUE #9277: Mazevo feedback/questions
> First of all, thank you and your team for pushing this out so quickly!
>  
> I was able to play around with this yesterday and I have some questions.
>  
> Is it possible to add “Status” as a selectable field in addition to “Rooms” and “Event Type”? I had an event that was marked as private, but it still showed on the screen because every status looks to be pulled from the website. If it’s too much trouble to make a field with selectable “Status” options, can we just have the Confirmed and Confirmed-Set Change options be active on the back end? Those are the only two statuses that are active in the current Drupal system.
>  
> Since we need to make slides for groupings of rooms can you tell me how many rows of events can fit on a single slide? This will help to determine if we need 2, 3, or 4 rooms per slide. It also looks like as event end times were passed those items were removed from the slide making more room for upcoming events, is that accurate?

cory-skeers:
Resolves #9277 

# How to test

Set up a local Mazevo api key
```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /slide/7551
```
Compare to https://signage.sites.uiowa.edu/slide/7551
Might need to pull up a different Mazevo slide and/or use `xdebug` to check results.
If you remove the statusIds addition in `sitenow_signage.module`
```
      $request_body['statusIds'] = [
        335,
        468,
      ];
```
you can emulate what `prod` is pulling in, and using `xdebug` you can see the statuses of the pulled-in events.


joewhitsitt:
pulling to test

joewhitsitt (approved):
By manipulating the end date and number of items, i was able to see a difference when the status ids were filtering and when they weren't.

checked mazevo to confirm the "confirmed" ids:

Screenshot 2025-10-20 at 3 25 14 PM
Screenshot 2025-10-20 at 3 25 21 PM

cory-skeers:
Thanks @joewhitsitt!

briand44:
thank you! I reached out this morning to get test examples but never got a response, sorry.

PR #9292, opened 2025-10-23
Adjust body summary field help text

### LINKED ISSUE #8593: Update help text on the summary field
We get questions regularly by people expecting the field to use trimmed body copy in the article teasers like on /news, but the behavior described by the help text ("Leave blank to use trimmed value of full text as the summary.") only applies to the content's meta description.

Additionally, the help text may suggest that it is okay, if not desirable, to leave blank and let it trim the body text, when ideally a unique, crafted summary should be written. Different help text could help encourage this, as well as provide some guidance on what makes a good summary and meta description.

cory-skeers:
Resolves #8593 

# How to test

Sync a site(s)
Log in and edit a node
If the content type uses the body field with summary, see the updated help text
If the content type does not use the body field with summary, see that the form is unchanged, doesn't error, etc.

Any feedback or edits for the help text?


e-marie-w (approved):
I checked the default content types on sandbox, the help text updated as expected. I also synced admissions and checked the several additional content types there. Again, the help text was changed as expected when used and there were no errors when not used. There was a warning about undefined array key weight for the scholarship content type, but I confirmed that is the behavior on the main branch as well so it's not related to this. 

I like the new help text, it clarifies the purpose and uses of the summary well in my opinion.

cory-skeers:
Thank you @e-marie-w!

Issue #9283, opened 2025-10-17
Policy PDF download/generation issues


briand44:
# Problem
The Policy Manual PDF download seems to be missing portions of the manual.

<details>
<summary>Report from customer</summary>

> unfortunately I ran into a significant amount of missing text. The first time I downloaded I noticed that between pages 223 and 224 it skipped from Part II chapter 14 to Part III chapter 36; also between pages 279 and 280 it skipped from Part V chap 7 to chap 28.
> 
> I generated a new download and it started with Part II chap 15 through chap 34, then skipped to Part III chap 16 ...
> 
> So, some glitches. In general, each page has the necessary info — both Part and Chapter numbers and titles (& subtitles). And there didn't seem to be any missing pages/subchapters/text within the downloaded chapters.

</details>

# Steps to reproduce
Download the manual at https://policy.uiowa.edu/admin/pdf-content/download and see that it starts with II. Community Policies - Chapter 15. Searching for I. Governance, turns up nothing.

joewhitsitt:
Out of curiosity, I went looking in Splunk and didn't see any logging related to `PDF generation failed for chunk`
https://github.com/uiowa/uiowa/blob/f44c24ad0cf4839eb81db380fe6b8ed6eb81b5ad/docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php#L189

briand44:
## Grooming
 
**Quadrant:** 

Complex - not sure what the problem is, results seem to differ from version to version.

 **Discussion:**
 - I would like to add more logging (notices) that just log all the things. Batch number, tmp file location, etc.
   - I want to see paths and batch ids in the logs and when they are cleaned up that they are cleaned up and not reused
 - Early exit if one of the steps fail could be baked in
 - We didn't notice any issues testing this prior to rolling out to prod
 - 900+ pages is the expected. One version I downloaded was only 400 something
 - if we notice any issues while adding additional logging we can debug and try to resolve otherwise additional logging may allow us to get additional info if this is only happening in production.
 - Check if file exists before showing download button.
 - Could it be a similar issue to https://github.com/uiowa/uiowa/pull/7631?

**Proposed solution or definition of done:** 
- Test to see if related to https://github.com/uiowa/uiowa/pull/7631?
- If we can't reproduce locally, improved logging in production to better understand what is failing.
- If we can reproduce locally, a fix that allows the PDF to generate in full and in order.

joewhitsitt:
[...]

PR #9320, opened 2025-10-28
policy pdf - re-link to private dir, sprinkle more logging

joewhitsitt:
Relates to #9283

This re-wires a connection to the private files directory which was lost when I switched the batching to include the dompdf part. The private files directory seems more stable than the default temp directory across environments and allows easier monitoring during the process.

# How to test

```
ddev blt ds --site=policy.uiowa.edu && ddev drush @policy.local en dblog && ddev drush @policy.local uli admin/pdf-content && ddev drush @policy.local watchdog:tail
```

Before you click the generate button, navigate to uiowa/files-private. As the generation process runs, you may see the creation of policy.uiowa.edu and within that a policy_pdf_temp directory. If you run the generation process a few times you will see this directory created and filled with temp files before being deleted (directory will remain).

In your terminal, you will see logging that mentions batches 0-8 being created and then merged together.

The resulting pdf should be be ~3 MB and 985 pages. We wanted to at least get the logging out to PROD as that is where the problem is occurring.

### DIFF
--- a/docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php
+++ b/docroot/sites/policy.uiowa.edu/modules/policy_core/src/Controller/PDFContentController.php
@@ -29,7 +29,7 @@ class PDFContentController extends ControllerBase implements ContainerInjectionI
    */
   const EXPORT_DIR = 'public://exports';
   const EXPORT_FILE = 'policy-manual.pdf';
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Looks good! It'll be nice to have the extra logging in case it runs into further issues.

joewhitsitt:
Thanks @cory-skeers. fingers-crossed but i think the webform temp directory would be a good next experiment if this doesn't work.

PR #9334, opened 2025-10-30
remove text-transform from ckeditor's version of the table caption

joewhitsitt:
# How to test

```
ddev blt frontend && ddev blt ds --site=dentistry.uiowa.edu && ddev drush @dentistry.local uli node/1346/layout
```

See that the table captions are not all caps.


bspeare (approved):
Looks good!

PR #9349, opened 2025-11-05
Facilities coordinators list fix.

### LINKED ISSUE #9301: Hidden element has focusable content
On the facilities [Building Coordinator List](https://facilities.uiowa.edu/building-coordinators-network/building-coordinator-list), the hidden width measurer implemented to not have the table jump around when the details are opened is throwing errors.

It seems as though `aria-hidden='true'` is not enough for screen readers, and the content after it is still navigable.

https://my2.siteimprove.com/Inspector/1005109/A11Y/Page?pageId=65055813775&impmd=6F4B1BAAA780A4D7A3CA00D93433A366&issue=cantTell+failed&wcag=twopointtwo&decision=dismissed+falsePositive+none#/sia-r17/failed/R1BLaHRXN3A=;MWQxNDI3NDE=

# Proposed solution

add the `visibility: hidden;` rule to the `.width-measurer ` scss rules. This removed it from being an issue when I ran an SI extension scan.

GaryRidgway:
resolves #9301

# How to test
1. sync facilities
2. [open this page](https://facilities.uiowa.ddev.site/building-coordinators-network/building-coordinator-list)
3. Observe that now the contact and managers columns have fixed widths (in inspector)
4. Click a bunch of details dropdowns and see that there is no column shifting
5. check that the original siteimprove issue is not happening anymore.


docroot/sites/facilities.uiowa.edu/modules/facilities_core/css/coordinators.css
```diff
@@ -0,0 +1,6 @@
+.view-display-id-block_building_coordinators .views-field-nothing {
```
joewhitsitt, line 2: For this CSS, can we add a comment for the what/why. nothing and nothing-1 doesn't really reveal much to a developer without having to look through views config to understand what it represents in the view and why.

joewhitsitt:
pulling to test

docroot/sites/facilities.uiowa.edu/modules/facilities_core/css/coordinators.css
GaryRidgway: Yeah, definitely can. I think thats a good idea. Do you think there would be any issues with me going back and editing those custom fields to be named differently?

joewhitsitt: they are "views fields" that i don't think we control the machine name for. a quick comment seems easier

GaryRidgway: sounds good!

joewhitsitt (approved):
Does the job, responds well and resolves the siteimprove error (confirmed) in the browser extension.

Thanks for the adjustments on this @GaryRidgway

PR #9357, opened 2025-11-07
Vertical video heading

### LINKED ISSUE #8515: Content missing after heading - SiteImprove - Vertical Video - uiowa.edu
# Problem 

Vertical video on uiowa.edu is getting flagged in Siteimprove for content missing after heading.

The vertical video used to have content associated with the heading.  Now it longer does and we should remove the heading tag on the vertical video overlay. 

This is a "S" Best Practice issue in Siteimprove. 

Example - https://my2.siteimprove.com/Inspector/995976/A11y/Page?pageId=30184895912&impmd=264ED37B65FA6186129334A508391970#/sia-r78/failed/VWdGUGc0MGc=;MjQ5NmNiNTE=

# Solution

- ?

cory-skeers:
Resolves #8515 

# How to test

```
ddev blt ds --site=uiowa.edu
```
Compare the vertical videos on https://home.uiowa.ddev.site/ to https://uiowa.edu/
Use the Siteimprove browser extension, and see that you get "Content missing after heading" flagged for at least one of the vertical videos on the live site, but not the local

View the Accessibility DOM, user Voiceover, or other preferred tools to check for any differences in navigation of the page

### DIFF
--- a/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-vertical-video.html.twig
+++ b/docroot/themes/custom/uids_base/templates/block/block--inline-block--uiowa-vertical-video.html.twig
@@ -58,9 +58,9 @@
     </div>
     {% if video_title is not empty %}
       <div class="highlight__wrapper">
-        <h3 class="highlight__title">
+        <span class="highlight__title">
           <span class="headline__heading">{{ video_title }}</span>
-        </h3>
+        </span>
       </div>
     {% endif %}
   </div>


cory-skeers:
Noting here that the video text usually gets skipped over by my navigating, but this behavior was the same with the previous heading. The video controls are all available, and so works accessibly as expected by my understanding.

With that, the text seems like visual candy rather than of structural importance, so I think this is fine or preferred, and also seems like a `span` to contain it feels correct to me.

joewhitsitt (approved):
Removes the siteimprove error

PR #9359, opened 2025-11-07
Signage - Add text slide

### LINKED ISSUE #9326: Ability to add text slide
# Story

As a Sign Manager, I would like to be able to add a slide of text so that I can quickly make updates to a schedule that is changing daily without having to create a new image slide every time.

# Additional Info

<details>
<summary>Feedback</summary>
>Am I able to enter text directly onto the sign, like we could with the old Drupal? Iatl is under construction, and I need to post a schedule that could change daily. So I don’t want to have to create a new slide every time. Is there a way to do this? I saw the Global text, but wasn’t sure what that is.
</details>

This is functionality that was available in D7 Signage.

Would we need to limit the amount of text that can be added or can we leave that up to the Sign Manager to use appropriate amount of text?

Would there need to be a headline for a text slide?

Filtered html?

pyrello:
Resolves #9326 

# How to test
```
ddev blt frontend && ddev blt ds --site signage.sites.uiowa.edu && ddev drush @sitessignage.local uli
```
- Masquerade as a sign manager
- Test out the new text card functionality
- Formatting should only allow bold, italic, and unordered lists
- Headline should use underline variant (`.headline--underline`)
- Check the other slide types that include a headline to see that they are all using the underline headline variant
- Outlook events and Astra room schedule slides should have a statically defined table caption that is visually hidden. They should display any user defined headline using the headline field and the underline variant of such
- For the Astra room schedule slide, if no headline is supplied, it should display "Room schedule" as the headline (consistent with the current behavior)


bspeare:
Should we require the headline or add some spacing if the headline is not provided?

Screenshot 2025-11-14 at 2 03 36 PM

pyrello:
> I'd probably vote to make it required otherwise users might create blank slides to create space between slides.

I chatted with @briand44 about this. While neither of us had a strong feeling about making the headline required and are probably actually inclined the other way, making it required now vs. in the future is easier. Also, it will encourage people to give us feedback about the slide if they don't like having to have a required headline.

Unrelated: I think I might have changed the functionality on the slides where I adjusted the headline to show independent of the caption. Previously we had a fallback headline that would show if the user didn't make one of their own. I think I broke that with the changes I made. I'm going to do a bit more digging and probably will need to push a fix.

cory-skeers (approved):
Reviewed and tested, and I think this works really well. Worth getting out there with feedback to see if we need to address alignment or other options, but imo this offers plenty of flexibility to fit most needs.

Opening a follow-up issue re: headlines sounds good to me. The adjustments to styling, moving table captions up to a headline, etc that has been done here all looks good.

I wanted to note in case others run into this or have thoughts, especially if we look at the headline field in another issue: I started to confuse "Title" and "Headline" during testing and occasionally edited the title when I'd intended to adjust the headline. But I think for other slide types there either isn't a headline, or it's more likely to have a different title and headline for the difference between organization and display. I think this makes sense to remain consistent with the same field, and to offer different title/headline for this slide as a result, and that the confusion was more that I was re-editing many slide headlines at once.

PR #9360, opened 2025-11-10
Signage js scheduling investigation.

### LINKED ISSUE #9348: Investigate an alternative to scheduler in Campus Signage
# Problem

Coming from https://github.com/uiowa/uiowa/issues/9325, the current scheduler solution is clunky and needs refinement. Alan previously pitched using JS to store scheduled times with the slide output to determine whether to show or hide published/unpublish. Could lengthen but not eliminate the sign reload interval.

# Proposed Solution

- [ ] Investigate removing scheduler from signage.sites and replacing it with a JS driven scheduling system in which content is published but not shown to the user until the specified time.

GaryRidgway:
# How to test

### DIFF
--- a/config/sites/signage.sites.uiowa.edu/core.entity_form_display.node.slide.default.yml
+++ b/config/sites/signage.sites.uiowa.edu/core.entity_form_display.node.slide.default.yml
@@ -6,11 +6,14 @@ dependencies:
     - field.field.node.slide.field_slide_description
     - field.field.node.slide.field_slide_duration
     - field.field.node.slide.field_slide_duration_other
+    - field.field.node.slide.field_slide_publish
+    - field.field.node.slide.field_slide_unpublish
     - field.field.node.slide.field_slide_widget
     - field.field.node.slide.og_audience
     - node.type.slide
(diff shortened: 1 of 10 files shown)


GaryRidgway:
If I would code this from the beginning again, I would make an object to hold the information of the slides, and just have flags that can be set in that. I think that would be much neater than the `dormant` array

GaryRidgway:
Notes that I have for working on a writeup:

- We should be worrying about the "unpublished" css we have.
    - Should we even allow people to publish or unpublish?
- Abstract helper functions for platform as a whole in JS
    - Namespace functions and utilities etc under drupal or new sitenow object namespace
- Make class/object for Slide
    - Can hold all the stuff!
- Checkbox to show you want to schedule stuff.
- Scheduling terminology still seems like the correct terminology
- weekday checkboxes?
- clear button?
- Should we be calling these terms “Display after” or “Hide after”, re-using Drupalisms can come with a cost.

<details><summary>I made a mockup for a potential form structure.</summary>
<p>


mockup

</p>
</details>

PR #9364, opened 2025-11-11
Remove access to block content overview

pyrello:
This change removes access to the block content overview page. It appears that when the block content overview was split out from the block types in Structure, we didn't catch that it was accessible to users.

# How to test
```
ddev blt ds --site hr.uiowa.edu && ddev drush @hr.local uli admin/people
```
- Masquerade as @mkyoder or another webmaster.
- https://hr.uiowa.ddev.site/admin/content/block should not be allowed anymore.
- Check that you can still edit the footer contact block via the contextual link and/or the link on the basic site settings config page.


cory-skeers (approved):
Reviewed and tested 👍

PR #9376, opened 2025-11-17
[hotfix] block library access - add these permissions back

joewhitsitt:
Rolling back https://github.com/uiowa/uiowa/pull/9364

# How to test

On `main`

```
ddev blt di --site=default && ddev drush @default.local uli admin/people
```

- Create and masquerade as an editor
- Go to node/1/layout and try to add a media image to an image block. see the console error similar to the screenshot below.

Checkout `hotfix_restore_block_library_access` and run `ddev drush @default.local cim`
Refresh the page and try to add the same image as before.

### DIFF
--- a/config/default/user.role.editor.yml
+++ b/config/default/user.role.editor.yml
@@ -75,6 +75,7 @@ weight: -6
 is_admin: false
 permissions:
   - 'access administration pages'
+  - 'access block library'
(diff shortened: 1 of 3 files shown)


pyrello (approved):
Works as expected. Thanks for the quick fix!

PR #9380, opened 2025-11-18
Smart trim for CLAS Policy book pages

pyrello:
This is a fix for how book page teasers are being trimmed on the CLAS Policy website.

# How to test

```
ddev blt ds --site policy.clas.uiowa.edu
```
- Check https://claspolicy.uiowa.ddev.site/book-tags/salary and make sure there are no hidden links using the SI browser scan tool.
- Check other taxonomy term pages to ensure this generally looks okay.


cory-skeers (approved):
Since it cuts off without an ellipsis, it's hard to tell that with where it happens to cut off in the Applicability section here, it should also include "Accountants"

image

An ellipsis in the suffix might help, but the alternative is that the summary field can also still be used to curate the display here. This fixes the direct issue so I'm happy to 👍 and leave it to your judgment based on the issue reporting whether you want to slip in an extra '...' suffix before merging or not.

PR #9384, opened 2025-11-19
Border bottom fix

### LINKED ISSUE #9363: Vertical secondary menu items don't have border-bottom
# Problem

Image

Noticed on https://pathology.medicine.uiowa.edu/our-people/administrative-services that we were missing `border-bottom` from our menu items. Should look like:

Image

I dug a bit deeper and noticed the following line on `main-menu.scss`:

https://github.com/uiowa/uiowa/blob/cf79a6234713eb8829507c4aade75f62977b71a7/docroot/themes/custom/uids_base/scss/components/menus/main-menu.scss#L65

[In UIDS' variables we have:](https://github.com/uiowa/uids/blob/9857bb4cff0682393573f8a8a7425d7803b6435b/src/scss/uids-core.scss#L20)

`--uids-grey-light: #CACACA;`

Which is presumably what we use as a basis for the SASS variable `$grey-light` and which won't fit into an `rgba()` property due to its hex form. It also won't fall back to the previous line:

`border-bottom: 1px solid rgba(0, 0, 0, .125);` 

....since technically `border-bottom: 1px solid rgba(variables.$grey-light, .5);`isn't invalid CSS,  but rather`$grey-light` is not a valid color for `rgba()`, thus leaving us with no border.

# Steps to reproduce
Make a secondary vertical menu anywhere on any site. Note the missing borders.

# Proposed solution
Consider converting `grey-light` and/or other variables to RGB, like `202, 202, 202` in order to change their alpha property with 
`rgba(variables.$grey-light, .5);`

Changing the property to: `border-bottom: 1px solid rgba(202, 202, 202, .5);` in my browser dev tools seems to solve the issue:
Image

GaryRidgway:
tracks #9363

uses https://github.com/uiowa/uids/pull/1017

## To test
1. `git pull && ddev composer install && ddev blt frontend && ddev blt ds --site=pathology.medicine.uiowa.edu && ddev drush @medicinepathology.local uli our-people/administrative-services`
2. Check the borders on the vertical menu. do they work?
3. Are these things using the new border variable?
4. check other components that use the new border variable. Are they still functioning correctly? (Sandbox may be a good place to look at things like this.)


joewhitsitt (approved):
fixes menu and components are receiving the variable

PR #9385, opened 2025-11-20
[Charts] Removed access to data collector color picker

bspeare:
Removes access to a color picker that we missed in testing:

Screenshot 2025-11-20 at 10 29 16 AM

# How to test

```
ddev blt ds --site=uiowa.edu && ddev drush @home.local uli /node/411/layout
```
1. Edit chart block and confirm that you can no longer access the color options displayed in the screenshot above and on https://uiowa.edu/node/411/layout.


joewhitsitt (approved):
checks out. yeah i just thought it was a weird horizontal rule to indicate the header.

PR #9393, opened 2025-11-21
[Facilities] Adjust Building alerts

### LINKED ISSUE #9314: Headings are not structured - Facilities Buildings
# Motivation

SiteImprove is reporting Headings are not structured accessibility best practice issue on Facilities building pages when an alert is being displayed on the building page.

https://my2.siteimprove.com/Inspector/1005109/A11y/Page?pageId=65056259075&impmd=264ED37B65FA6186129334A508391970#/sia-r53/failed/S0YwYWpOeFU=;MDU0MGRlYTI=

# Additional Info

The alert title is being output as a h3 and the previous heading is an h1.

cory-skeers:
Resolves #9314 

# How to test

~ddev blt ds --site=facilities.uiowa.edu && ddev blt ds --site=its.uiowa.edu && ddev drush @facilities.local uli /building/0021 && ddev drush @its.local uli /node/10871/edit~
~On Facilities, compare https://facilities.uiowa.ddev.site/building/0021 to https://facilities.uiowa.edu/building/0021~
~There shouldn't be any visual change to the alert~
~Inspecting the alert heading should reveal an `h2` rather than an `h3`~
~Using the Siteimprove browser extension on the local page should not contain a "Headings are not structured" error~

~On ITS, adjust the start time of the provided alert, another alert, and/or create a new alert that is active~
~Go to the affected service page~
~Inspect the alert heading(s)--they should be set to `h3`, consistent with the `h2` "Alerts" heading above them.~

```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli /building/0021
```
Inspect the alert, and see that the alert has an `h3` nested under a visually hidden `h2`
Using the Siteimprove browser extension on the local page should not contain a "Headings are not structured" error
Check a building that doesn't have an active alert and confirm that there is no hidden `h2` "Building announcements" present


joewhitsitt:
Should there be a visually hidden headline here instead? "Building alerts" or "Building announcements" so the user understands the context of the area they navigate through?

cory-skeers:
> Should there be a visually hidden headline here instead? "Building alerts" or "Building announcements" so the user understands the context of the area they navigate through?

Hrmm...

If we went that route I think we could use the same views prerender here and add markup to include a visually hidden headline. That way it only gets added if there are results, to match the visual behavior and to keep from introducing a 'headline with no content' problem.

ITS handles it a bit differently by just always having it present with a headline and 'no results' text, but the alerts are a little different in nature.

Thinking through which approach would be better. I think I'm loosely on the fence, but could be tipped if anyone has a stronger preference.

joewhitsitt:
This too was working @cory-skeers but I think I found even _another_ way closer to how other blocks would be setup. In display settings for the node, we can add a headline to the block and visually hide it. Then remove the header content you had and set the view to "hide if empty."
Screenshot 2025-11-25 at 10 42 51 AM
Screenshot 2025-11-25 at 10 42 29 AM

Screenshot 2025-11-25 at 10 49 00 AM
Screenshot 2025-11-25 at 10 50 10 AM

Care if I push that up?

joewhitsitt (approved):
Approved but wouldn't mind some confirmation.

PR #9396, opened 2025-11-21
Adding arrow key controls to slides.

### LINKED ISSUE #9327: Ability to navigate slides using keyboard
# Story

As a Sign manager, I would like the ability to navigate through slides using the left/right arrow keys when viewing a sign so that I can review all the slides without waiting for the specified duration.

# Additional Info

<details>
<summary>Feedback</summary>

> Is it possible to add left/right arrow keyboard navigation when previewing/viewing the slideshow to go to the previous/next slide without waiting for the specified duration? I believe the old system had this. Came in handy to make sure a specific slide looks correct in a "live-like" environment without physically going to a sign.

https://github.com/uiowa/uiowa/issues/9116#issuecomment-3266984289

>When viewing the slides, is there a way to advance them faster (e.g. arrow key, but that doesn’t work).

https://github.com/uiowa/uiowa/issues/9116#issuecomment-3451821330

>Are we able to add controllers of the slideshow if we wanted?

https://github.com/uiowa/uiowa/issues/9116#issuecomment-3463125209

</details>

This functionality was available in D7 Signage.

From Ben:
>For the keyboard navigation, it looks like we might need to plan work to implement this and adjust our slideshow.js based on https://splidejs.com/guides/options/#keyboard.

https://github.com/uiowa/uiowa/issues/9116#issuecomment-3271150378

GaryRidgway:
tracks #9327

## To test
1. Sync down signage!
2. Make or go to a sign.
3. Ensure focus is on the slideshow by clicking it
4. use keyboard right arrow and left arrow to see if you can navigate between slides.
5. Run an accessibility scan of the page to ensure nothing is bonkers broken.
6. Review the [marquee role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/marquee_role) and offer opinion on whether you think this is correct or not.
    7. Siteimprove was flagging this for not having the correct role beforehand, so that is why I tried to change it to the correct role.[ I did research on live region roles to try and fix this.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles#4._live_region_roles)

### DIFF
--- a/docroot/modules/custom/sitenow_signage/js/slideshow.js
+++ b/docroot/modules/custom/sitenow_signage/js/slideshow.js
@@ -39,10 +39,9 @@
           drag: false,
           speed: 1500,
           slideFocus: false,
-          arrows: false,
+          arrows: true,
           pagination: false,
           rewind: true,
-          keyboard: 'focused'
         });
 
         // Set up initial video states before mounting.
@@ -67,30 +66,6 @@
         });
 
         splide.mount();
-
-        // The controller has functions to go forward and back in the carousel.
-        const controller = splide.Components.Controller;
-
-        ['keydown'].forEach(function(event) {
-          window.addEventListener(event, function(event) {
-            const key = event.key;
-
-            const callback = {
-              "ArrowLeft"  : this.leftHandler,
-              "ArrowRight" : this.rightHandler,
-            }[key]
-            callback?.(event);
-
-            this.leftHandler = function(event) {
-              const prev = controller.getPrev();
-              controller.go(prev);
-            }
-            this.rightHandler = function(event) {
-              const next = controller.getNext();
-              controller.go(next);
-            }
-          });
-        });
       });
 
       /**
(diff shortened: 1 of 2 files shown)


GaryRidgway:
> I can't answer whether marquee is the best fit for this, but it looks like from the same link you provided, that it needs to be labelled then and possibly set with aria-live off if we keep it.
> 
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/marquee_role#associated_wai-aria_roles_states_and_properties
> 
> Splide set `region` as the default and suggests `group` with a label for decorative carousels. https://splidejs.com/guides/accessibility/#landmark
> 
> My own searches for non-interactive carousel/slideshow haven't been returning any clear signal for aria role.

Understood, I will change it to that!

docroot/modules/custom/sitenow_signage/js/slideshow.js
```diff
@@ -8,6 +8,10 @@
     attach: function (context, settings) {
       context.querySelectorAll('.signage__slideshow').forEach(function (element) {
 
+
+
+
+
```
pyrello: Unnecessary blank space

joewhitsitt (approved):
Works as groomed. Of the accessibility issues, keyboard navigation isn't one of them (checked siteimprove and ainspector).

Issue #7176, opened 2023-12-05
Service review of Facebook Embedded Posts


joewhitsitt:
- Need to do another app review to maintain Facebook oEmbed (Jan 4, 2024). My understanding is that this is also required for Instagram
- media_entity_facebook has a new method that doesn't require the API creds:

>Since 4.0.0, you can now use [Embedded Posts](https://developers.facebook.com/docs/plugins/embedded-posts/) to generate the embedded content (see /admin/config/media/facebook-settings). It can be used without app review for everyone.

- See if either module has usage. See if Instagram can leverage the same embedded posts functionality.

joewhitsitt:
App review was approved but now we are in violation of not being tied to a Meta Business Account. We have until Jan 3, 2025 to respond/resolve.

Instagram does not appear to have the same embedded post functionality at this time.

Just from checking twitter usage, I suspect Facebook and Instagram will have similar numbers: https://github.com/uiowa/uiowa/issues/2147#issuecomment-1858418173

joewhitsitt:
For facebook we have switched to Embedded Posts so the API/App Review issue is not affecting this content.

For Instagram, ran a similar report to what I previously did for Twitter. Again these are instances in the database, not necessarily usage.

academics.athletics.uiowa.edu,1
admissions.uiowa.edu,4
aiaa.org.uiowa.edu,1
alphachisigma-alphatheta.org.uiowa.edu,1
baran.lab.uiowa.edu,1
campuscouncil.stanleymuseum.uiowa.edu,1
careers.uiowa.edu,1
dasblatt.sites.uiowa.edu,1
dentistry.uiowa.edu,1
diversity.uiowa.edu,37
education.uiowa.edu,2
engineering.uiowa.edu,4
facilities.uiowa.edu,2
firstgen.uiowa.edu,2
fys.uiowa.edu,2
gpsg.uiowa.edu,1
hawkeyemarchingband.uiowa.edu,10
honors.uiowa.edu,12
indianstudentalliance.org.uiowa.edu,3
international.uiowa.edu,1
intranet.writersworkshop.uiowa.edu,1
law.uiowa.edu,6
ncec.sites.uiowa.edu,2
our.research.uiowa.edu,4
paleorepository.sites.uiowa.edu,6
pentacrestmuseums.uiowa.edu,25
physics.uiowa.edu,5
psp.org.uiowa.edu,2
recserv.uiowa.edu,1
robotics.org.uiowa.edu,2
[...]

PR #9432, opened 2025-12-09
media_entity_instagram - delete existing entities and footprint

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/7176 and https://github.com/orgs/uiowa/projects/20?pane=issue&itemId=49617004

# How to test

As an authenticated administrator, visit the following pages to see the oembed error and notice there is no instagram embeds visible to a visitor:

https://pentacrestmuseums.uiowa.edu/news/2022/02/iowa-winter-birds
https://pentacrestmuseums.uiowa.edu/news/2022/09/museum-natural-history-muse-mfa-grad-student-mariceliz-pagan-gomez
https://pentacrestmuseums.uiowa.edu/salt-and-pepper-looking-love-and-familial-history

```
ddev blt ds --site=pentacrestmuseums.uiowa.edu
```

See that 20 some instagram entities are deleted in an update hook and that the module and related config are uninstalled/removed without errors.

Poke around the website and see that Instagram is not part of the tours, not part of the media add/interfaces interfaces and not an option to filter on.

# Follow-up

- [ ] Remove from composer.json
- [ ] Remove facebook api creds from secrets files on the remote apps/envs as facebook doesn't us the API anymore to embed posts.
- [ ] Delete oembed read app from facebook developers account.
- [x] Update docs.prod.drupal.uiowa.edu, sitenow.uiowa.edu/documentation
  - With removal in mind, maybe note that it was previously a thing and they were removed. There might be content on your website that lack context now that the embed was fully removed. We have the list down below of usage to help folks that become concerned. 
    - https://sitenow.uiowa.edu/documentation/media/social-media-embed
    - https://sitenow.uiowa.edu/known-issues#instagram-embeds
    - https://docs.prod.drupal.uiowa.edu/miscellaneous/facebookinstagram-graph-api-oembed
    - https://docs.prod.drupal.uiowa.edu/miscellaneous/embedded-content-decision-tree


joewhitsitt:
>My initial reaction to this is that we should not remove these. While its not ideal to have them there, they are fairly innocuous. A content editor can either remove them or replace them with something.

Ope. Just saw this. That is how I was feeling about these initially, but tracking this down after the module has been removed is going to be difficult and could be embarrassing on certain content if they aren't aware of it. If we leave these, I think we should do a communication out to folks that still have them placed like we have done in the past for other disruptions. Doing an update hook feels easier than a communication plan though so I've started to look into this.

joewhitsitt:
Yeah I hinted at `<!--Broken Instagram Post-->` so that it could be found via the Find Text functionality. I will pause again for team discussion before going that far and just focus on locating and removing for now.

bspeare (approved):
This looks good after testing this twice with a few sites.  Thanks for removing the placeholders!

PR #9433, opened 2025-12-09
CKEditor Allowed Sources - opt-out of restrictions and re-save config

### LINKED ISSUE #8218: Fix FilterIframe and text format configuration

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/8218

As mentioned in standup, this temporary workaround is safe in my opinion since the elements and attributes are tracked in config and can't be changed willy nilly.

The risk I think is if a plugin allows/disallows a certain attribute that we then don't update properly. In that case, I think it would mean our attribute might not have the effect we are expecting.

# How to test

- Review that the only changes to the allowed_html config are re-ordering of existing attributes
- Go to https://sandbox.uiowa.ddev.site/admin/config/content/formats/manage/filtered_html and see that you can save the form and it doesn't bark at you about an element/combination that shouldn't be in that input.
- Go to https://sandbox.uiowa.ddev.site/kitchen-sink, make an edit to the body and see no regression in HTML compared to prod.


bspeare:
pulling to test.

config/default/editor.editor.filtered_html.yml
```diff
@@ -6,7 +6,6 @@ dependencies:
     - filter.format.filtered_html
   module:
     - ckeditor5
-    - ckeditor_small_tag
```
bspeare, line 9: What's the story with this removal here?

config/default/filter.format.filtered_html.yml
```diff
@@ -135,6 +135,7 @@ filters:
     weight: -50
     settings:
       title: true
+      media_substitution: metadata
```
bspeare, line 138: Any idea what this does?

bspeare (approved):
This appears to work as expected.  Thanks for finding this fix!

config/default/editor.editor.filtered_html.yml
joewhitsitt: I am guessing it was manually added in https://github.com/uiowa/uiowa/pull/9072 but not actually needed? This is the result of a proper config export and as far as I know, the small functionality wasn't affected but might be good to test. It can be added back in but it will be one of those pieces that will show up again I think as a removal.

bspeare: Sounds good. The small tag/button does seem to work without it.

config/default/filter.format.filtered_html.yml
joewhitsitt: ~i _think_ that traces back to https://www.drupal.org/project/linkit_media_library/issues/3299988~

🤷

PR #9428, opened 2025-12-08
upgrade samlauth with dependencies, externalauth

joewhitsitt:
This jumped up in the queue for reasons.

This is the config that got emptied out. I don't think it is necessary for our setup and is not currently set that way in [prod](https://sandbox.prod.drupal.uiowa.edu/admin/config/people/saml). Just got moved around a bit:

Screenshot 2025-12-09 at 10 07 16 AM

# How to test

Trusting pairs could coordinate the addition/removal from https://iam.uiowa.edu/access/ for testing different stories.

- In the DEV environment (e.g. facilities.dev.drupal.uiowa.edu), try logging into a website using your hawkid.
- If it is a new to you website, see that your developer hawkid is assigned the administrator role.
- Logout and log back in using `uli` and remove the administrator role and any other roles from your hawkid. Logout.
- Log back in as your hawkid and see that your role is re-assigned.
- Logout and log back in using `uli` and block access to your account. Logout.
- Test with a non-admin, non-webmaster user.

## Stories from https://github.com/uiowa/uiowa/pull/3687

- As a user logging in to a site with a role mapping, my account is created and granted the appropriate role(s) upon log in.
- As a user who has already logged in to a site via a role mapping, my role(s) is revoked if the role mapping changes to an attribute or value that I do not have.
- As a user logging in to a site without a role mapping or an existing account, I receive an error during log in and my account is not created.
- As a user logging in to a site where a manually created account exists with my HawkID, I can log in successfully and my email is synced correctly.
- As a user logging in to a site where a manually created account exists with my HawkID but has been blocked, I cannot log in.

PR #9435, opened 2025-12-10
fix onelogin/php-saml versioning

joewhitsitt:
The previous [PR](https://github.com/uiowa/uiowa/pull/9428) fixed the dependency of onelogin/php-saml, robrichards/xmlseclibs. Since then, it appears onelogin/php-saml has updated and is also part of the dependency tree for samlauth.

# How to test

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -15118,21 +15118,21 @@
         },
         {
             "name": "onelogin/php-saml",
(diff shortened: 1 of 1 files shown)


bspeare (approved):
Thank you!

PR #9445, opened 2025-12-15
logrotate - fix rules for deployment logs

joewhitsitt:
On this edition of my never ending quest to get log rotation correct...

- The pattern is match and "rotated" but is not cleaned up resulting in a bunch of files that end with `*.log.1`
- This changes the rule from rotating to just cleaning up after 14 days. I think.

Will need to visit each /shared/logs directory and clean up the existing mess.

# How to test

Just review for testing over time like before.


bspeare (approved):

joewhitsitt:
i've gone through the apps/envs to clean up old logging attempts

used `find . -type f -name "*.log.1" -delete` to clean up the bad files.

Issue #8203, opened 2024-11-04
Blockquote documentation page?


cory-skeers:
We currently don't have a basic documentation page to link to for the blockquote component

# Proposed solution
Add a blockquote documentation page to sitenow.uiowa.edu
Potentially link to it on https://sitenow.uiowa.edu/news/2024/10/letting-others-share-iowa-story-harnessing-power-blockquote-component

bewoldt:
@briand44 I added a page with documentation for the Blockquote block. https://sitenow.uiowa.edu/documentation/page-content-type/layout-builder/blockquote

briand44:
Thanks @bewoldt! Closing this as complete.

PR #9466, opened 2025-12-19
[Blockquote] Removed limit from blockquote content

### LINKED ISSUE #9351: Block quote character limit
Came from Brynne in CLAS, discussed with Maggie B:

- The component itself doesn't provide a character limit to the editor.
- Nor does it alert the editor when/if the characters are too long, making it very easy to save the component and not even realize the quote is being cut off on the page.
- Current character count cuts off half-way on a 1 column default setting - does that visually make sense?

Image

# Proposed solution

Look into the blockquote block form UX and update
 - Help text to specify character limit?
 - Cut off at the character limit or alert the user that text will be truncated?
 - Would it be possible to dynamically show the current character count?
 -  Does the current character limit need to be adjusted/removed?

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9351. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /blockquotes
```
1. Confirm "Unlimited" blockquote displays all of the text compared to https://sandbox.prod.drupal.uiowa.edu/blockquotes.


joewhitsitt:
pulling to test

joewhitsitt (approved):
checks out. also ran a cex to make sure there weren't other config diffs with this change.

no documentation to update, because we never wrote any 😬 https://github.com/uiowa/uiowa/issues/8203

PR #9468, opened 2025-12-20
[Special characters] Dialog bug

### LINKED ISSUE #9451: Special characters button not visible in Layout Builder
When adding/editing the Text block component in Layout mode, the special characters button does not function. The menu of special character options does not open or appear. I see that this button does function correctly when in Edit mode; however, we do not use this mode to add content to our sites anymore.

Tamara E from journalism.uiowa.edu; email in its-web

# Proposed solution

Investigate root cause.
Adjust z-index of the character options window?

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9451. 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /node/656/layout
```
1. Confirm you can use the special characters button while editing a text area.  Revert the CSS change and confirm that the dialog is not visible.


joewhitsitt:
pulling to test

joewhitsitt (approved):
It works, though I think sometimes I need to click it twice to show (no console error). Better than not working at all. Noticed the tray has a bit of a snap couple pixel adjustment each time the tray is dragged.

Thanks for the write up in https://github.com/uiowa/uiowa/issues/9451#issuecomment-3716252855

PR #9469, opened 2025-12-22
media_entity_instagram removal - reorder when the entities are deleted

joewhitsitt:
Delete the instagram entities (if any) after the usage lookup but before the early exit if there is no usage.

# How to test

Run on these sites. Some have entities but no usage. Some have entities and usage. One is a v2 site with entities.

```
ddev blt ds --site=admissions.uiowa.edu && ddev blt ds --site=advisingcenter.uiowa.edu && ddev blt ds --site=sandbox.uiowa.edu && ddev blt ds --site=hawkeyemarchingband.uiowa.edu
```

Should be no import errors. Run `ddev blt ds --site=admissions.uiowa.edu` on `main` and you should get an error.

### DIFF
--- a/docroot/profiles/custom/sitenow/sitenow.install
+++ b/docroot/profiles/custom/sitenow/sitenow.install
@@ -4521,6 +4521,19 @@ function sitenow_update_10031() {
     ->execute()
     ->fetchAll();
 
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Clears up the errors. Thanks!

Issue #9437, opened 2025-12-11
Generate list of admin routes


briand44:
Spreadsheet of custom admin routes: https://iowa.sharepoint.com/:x:/r/sites/CSI-Drupal2/Shared%20Documents/a11y/routes-custom.xlsx?d=wdc668ce4b78245f48757a96ea585d91a&csf=1&web=1&e=q1dAYA

---

# Motivation

We need to be compliant with the DOJ accessibility requirement by April 24, 2026. This includes administrative pages and pages behind authentication. We discussed identifying and creating a list of all admin routes as a first step.

# Additional Info

Some points from our accessibility conversation on Dec 2nd, 2025:

- Desire to switch focus to admin side of SiteNow while we make public facing compliance issues easier to identify in Siteimprove
- node/add, node/edit for content types, content overview, add/edit media, add/edit taxonomy, config pages heavily used
- Might be able to do a programmatic inventory by scanning module .routing.yml files and matching against /admin/*
- identify Drupal core related issues, find any related existing issues, report issues if don't find existing

Grooming conversation from https://github.com/uiowa/uiowa/issues/9413#issuecomment-3638619254

# Grooming Conversation

**Quadrant:** 
- Complicated

**Discussion:**

- Pull a list of all routes are in routing.yml files (as described above) and use that for one-off Siteimprove scans of those pages and report findings and collect them for a future issue. 
- Investigation.
- The list will tell us how many points we need to examine
- Then each route will need to be examined

**Proposed solution or definition of done:**

- Generate list of admin routes by scanning module .routing.yml files and matching against /admin/*
- Provide additional details of what is at each item at the list (type of form, etc).
- Developer choice on data delivery method for report.

pyrello:
```
find ./docroot/modules/custom -type f -name "*.routing.yml"
./docroot/modules/custom/sitenow_pages/sitenow_pages.routing.yml
./docroot/modules/custom/sitenow_dispatch/sitenow_dispatch.routing.yml
./docroot/modules/custom/sitenow_articles/sitenow_articles.routing.yml
./docroot/modules/custom/uiowa_hours/uiowa_hours.routing.yml
./docroot/modules/custom/sitenow_people/sitenow_people.routing.yml
./docroot/modules/custom/sitenow_events/sitenow_events.routing.yml
./docroot/modules/custom/uiowa_area_of_study/uiowa_area_of_study.routing.yml
[...]

PR #9478, opened 2025-12-29
Robo infrastructure + example command

pyrello:
This came out of #9437. It includes updates to allow us to start writing and executing Robo commands in a class file based organization similar to what we are currently using for BLT. It includes the command I wrote to generate a custom route list as an example to show how it works and provide a basis for future experimentation.

# How to test

- Code review.
```
ddev composer dumpautoload && ddev robo routes:custom
```
- Above command should output CSV formatted list.


pyrello:
> Is RoboFile needed?

I guess not. In a more basic setup (where we are not using separate class files for commands), you can just put all your commands in RoboFile.

joewhitsitt (approved):
Thanks for exploring this and getting a foundation started.

PR #9487, opened 2025-12-31
Disable security check for time being

pyrello:
Builds are failing and the process of updating composer for our project is less than straightforward. This disables the security check to allow for a release.

# How to test

- Code review + build passes

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -30,6 +30,9 @@ drush:
   aliases:
     ci: self
   default_alias: self
+disable-targets:
+  tests:
+    security-drupal: true
 tests:
   phpunit:
     - config: '${repo.root}'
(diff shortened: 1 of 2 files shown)


joewhitsitt (approved):
I think this is fine to disable all temporarily, but I am curious if it is possible to ignore specific advisories like we learned how to do in the past.

https://github.com/uiowa/uiowa/blob/b753e0ee522b5b51c3106824323a61ecc7e6dd6c/composer.json#L51

Not sure if composer can ignore advisories about composer 🤷

pyrello:
> I think this is fine to disable all temporarily, but I am curious if it is possible to ignore specific advisories like we learned how to do in the past.
> 
> https://github.com/uiowa/uiowa/blob/b753e0ee522b5b51c3106824323a61ecc7e6dd6c/composer.json#L51
> 
> Not sure if composer can ignore advisories about composer 🤷

I had forgotten this technique. Thanks for the reminder!

Issue #4973, opened 2022-03-15
Consider moving away from Superfish


joewhitsitt:
Follow-up from https://github.com/uiowa/uiowa/issues/3757

# Story

As a developer and a steward of UIDS on campus, I don't want our platform and design library to rely on Superfish. There are several issues with it.

# Proposed Solution

- Investigate lightweight and accessible alternatives to implement for our navigation and to back port to UIDS for community use.

richardbporter:
I looked into writing a kind-of superfish a11y test for https://github.com/uiowa/uiowa/pull/4956#issuecomment-1102675655 but was getting many schema errors. I looked at the superfish module for a test example and noticed there are 0 tests and no config schema.

joewhitsitt:
From https://github.com/uiowa/uiowa/pull/9268 I looked back on the ten-year old suggestion that was just sprinkled in and released. Seems like it would be better to fork these templates and remove menu and menuitem after recently watching the https://practical-accessibility.today/ video about the misuse of these roles.

https://git.drupalcode.org/project/superfish/-/commit/db46ecd0ed7e2254ea4ae7dc6ff44fc29e039b00

PR #9515, opened 2026-01-09
A11y tweaks for superfish and offcanvas menus

### LINKED ISSUE #9475: Container element is empty (menu)
# Motivation

Siteimprove is reporting a WCAG 2.1 A accessibility issue for Container element is empty.

Some roles are designed to contain other roles. This element has this type of role, but it doesn't contain any [required-owned elements](https://www.w3.org/TR/wai-aria-1.1/#mustContain).

# Additional Info

We addressed this issue previously as part of https://github.com/uiowa/uiowa/issues/9254

Being reported on 58 SiteNow sites.

This appears to be related to the menu system and specifically when there are child menu items.

Obermann Center example - https://my2.siteimprove.com/Inspector/1006434/A11y/Page?pageId=33370252600&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r68/failed/L3BXV0p0eEY=;MTMwNzBiMjI=

SHL example - https://my2.siteimprove.com/Inspector/1218676/A11y/Page?pageId=77670586839&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r68/failed/bmdFTiszUms=;Y2UxYmI4Y2E=

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9475

Forking the superfish templates to undo a change to the the module https://github.com/uiowa/uiowa/issues/4973#issuecomment-3513999394

https://github.com/uiowa/uiowa/issues/9475#issuecomment-3739314701

# How to test

First, confirm with the SiteImprove extension that shl.uiowa.edu has ~7 container element is empty (role=menu) instances.

```
ddev blt ds --site=shl.uiowa.edu
```

Without authenticating to reduce noise, go to https://shl.uiowa.ddev.site/ and run the Siteimprove widget. There should be no container element is empty issues.

Obermann has content issues as noise but can be checked as well.

Now login and change the menu to horizontal and toggle.

```
ddev drush @shl.local uli
```

For the toggle menu, open it up. Scan. Expand a menu. Scan. There should be no additional issues for each scan.

No ariahaspopup issue because I added role=button to the markup for the div.

**Bonus points**: Interact with the menus using VoiceOver and note that it is still keyboard operable and that links and number items is still indicated.

### DIFF
--- a/docroot/themes/custom/uids_base/templates/navigation/superfish-menu-items.html.twig
+++ b/docroot/themes/custom/uids_base/templates/navigation/superfish-menu-items.html.twig
@@ -1,0 +1,78 @@
+{#
+/**
+ * @file
+ * Default theme implementation of Superfish menu items.
+ *
+ * Available variables:
+ * - html_id: Unique menu item identifier.
+ * - item_class: Menu item classes.
+ * - link: Link element.
+ * - link_menuparent: Link element, when a menu parent.
+ * - children: Menu item children.
+ * - multicolumn_wrapper: Whether the menu item contains a column.
+ * - multicolumn_column: Whether the menu item contains a column.
+ * - multicolumn_content: Whether the menu item contains a column.
+ *
+ * @see template_preprocess_superfish_menu_items()
+ *
+ * @ingroup themeable
+ */
+#}
+{% set classes = [] %}
+{% for item in menu_items %}
+
+<li{{ item.attributes }}>
+
+  {% if item.multicolumn_column %}
+  <div class="sf-multicolumn-column">
+    {% endif %}
+
+    {# Parent items get a toggleable element #}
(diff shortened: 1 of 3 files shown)


cory-skeers (approved):
> Ah okay. I noticed that on the parent top-level `<ul>`, not each `<li>` item within that it revealed 8 items. I got ahead of myself.
> 
> That said, I am not sure how to bring that back in this "temp" superfish/siteimprove relationship. I can't find anything that says a semantic `<li>` announces which item it is within a total number of items.

Ah, gotcha. I went back and confirmed that it does still list the quantity on the parent. I think this works as a way to get things working, and getting a proper structure is worth the slight change in VO behavior.

Issue #9460, opened 2025-12-18
Color contrast does not meet minimum requirement - SiteImprove - Search label


joewhitsitt:
# Problem

I believe SiteImprove has issue with a `display: none` span that has the same color as the background (mobile).

# Background

https://my2.siteimprove.com/Inspector/1009549/A11Y/Page?pageId=47390440940&impmd=8BF541F76C11646F59D0764F0BD0CFC8&issue=failed&decision=dismissed+falsePositive+none&comp=ada#/sia-r69/failed/YnYxbmNFajU=;MzgyOGMxMWU=

<img width="812" height="848" alt="Image" src="https://github.com/user-attachments/assets/038f52f9-665f-4197-9e7d-e0e136c8739d" />

briand44:
Still not finding this in the cross site accessibility report when filtering by A and AA conformance. Went through 190 of 288 color contrast issues. Sometime I get frustrated with Siteimprove.

joewhitsitt:
Boom! 367 instances on Provost down to 0 after yesterday's deployment. Not sure how long it will take for the dashboards to stop showing it, but when you click in, it is no longer returning as an issue.

<img width="349" height="370" alt="Image" src="https://github.com/user-attachments/assets/8829f116-01c8-41c8-97df-0975fff19bc7" />

pyrello:
I'm going to be looking into this issue next as part of #9587. Re-opening until we get to the bottom of it.

PR #9519, opened 2026-01-13
Search label - Color contrast

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/9460

Going to leave the issue open after merge and deployment to check in on the fix in PROD.

# How to test

On `main`

```
ddev blt frontend && ddev blt ds --site=provost.uiowa.edu
```

Shrink viewport to mobile and scan using the SiteImprove widget. If you can't recreate the issue locally, just note there is no difference in how the search label's visibility changes compared to main when you checkout `search_toggle_color_contrast`.

On `search_toggle_color_contrast`

```
ddev blt frontend && ddev drush @provost.local cr
```

Shrink viewport to mobile and scan using the SiteImprove widget. If you had reproduced the issue locally, this _should_ fix it.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/components/search-toggle.scss
+++ b/docroot/themes/custom/uids_base/scss/components/search-toggle.scss
@@ -98,12 +98,10 @@ $search-img-path: '../../images/search-01.svg';
   }
 
   span {
-    //@include utilities.element-invisible;
-    display: none;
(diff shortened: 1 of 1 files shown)


bspeare (approved):
The widget wasn't giving me an error previously, but this implements the element-invisible mixins and works as expected.

Issue #9217, opened 2025-09-29
launch writing.uiowa.edu and deploy htaccess redirects


briand44:
# Story

As a member of the new Office of Writing and Communication, we would like to redirect writinguniversity.uiowa.edu and writinguniversity.org to writing.uiowa.edu as part of that site launch.

# Additional Info

>Regarding the launch date, a few days before the start of the semester would be great. But the main goals is to be able to share the site with colleges, departments, and students once the new term begins. If we get things done/approved before then, we are happy to lauch it early.

- currently have redirects related to these sites in .htaccess
- these should be updated so the following domains redirect to writing.uiowa.edu:
  - www.writinguniversity.org
  - writinguniversity.org
  - www.writinguniversity.uiowa.edu
  - writinguniversity.uiowa.edu

# Proposed Solution

- [x] Update `.htaccess` redirects using patch method
  - www.writinguniversity.org and writinguniversity.org should redirect to writing.uiowa.edu and pass along path/query parameters as well
  - www.writinguniversity.uiowa.edu and writinguniversity.uiowa.edu should redirect to writing.uiowa.edu and pass along path/query parameters as well
    - They would like to keep the old site available for a short time after launch so we shouldn't redirect the `.prod` domain
  - remove writing.uiowa.edu redirect
- [x] Deploy `.htaccess` changes as part of launch
- [x] Verify redirects working as expected (redirects from https://github.com/uiowa/uiowa/issues/9425 should now land on appropriate page of writing.uiowa.edu as well)
- [x] Additional launch tasks (xml sitemap, remove writing.prod.drupal.uiowa.edu, comms, etc.)
- [x] Add writinguniversity.prod.drupal.uiowa.edu so old site can still be reached
- [x] Create reminder to revisit and see if we can remove internal prod domain
- [x] Any related Siteimprove updates

briand44:
I have questions out regarding the launch date and the redirects for this. Will update once I get a response.

bspeare:
Looks like they want to launch the site on Tuesday (1/20).  Let's discuss this in standup tomorrow.

PR #9524, opened 2026-01-16
Writing redirects htaccess update

joewhitsitt:
Relates to #9217

# How to test

Deployed out to DEV but there is little value for it other than to confirm that a redirection is still taking place. Everything is being redirected to https://writinguniversity.uiowa.edu/about-us/writing-iowa currently because of the existing rules in PROD.

https://htaccess.madewithlove.com?share=f40030c5-f1b1-42a0-9843-43b7db4de309

### DIFF
--- a/docroot/.htaccess
+++ b/docroot/.htaccess
@@ -2,20 +2,6 @@
 # Apache/PHP/Drupal settings:
 #
 
-# Block these IP addresses.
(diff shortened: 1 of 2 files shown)


bspeare (approved):
This looks like what we need and properly scoped.  Tested with https://htaccess.madewithlove.com/.

Issue #9482, opened 2025-12-31
Investigate bandwidth usage across sites


briand44:
# Motivation

WAF costs are related to bandwidth. Acquia is reporting our monthly bandwidth at around 35TB. They are saying that is a lot higher than most. Most are closer to ~2TB/month.

We should investigate bandwidth across sites and understand if there are individual sites that are high bandwidth users or if there are platform improvements that could be made to help address this.

# Additional Info

We had ISPO provide some initial information from Splunk:

https://iowa.sharepoint.com/:x:/r/sites/CSI-Drupal2/_layouts/15/Doc.aspx?sourcedoc=%7B758B7A71-CDC8-4FC1-BD45-EF08A46FDF84%7D&file=acquia-traffic-by-domains-90d.csv&action=default&mobileredirect=true

Chart from Michael:

<img width="771" height="589" alt="Image" src="https://github.com/user-attachments/assets/6db99235-1bc4-4441-a5f2-4b186d241581" />

Joe had mentioned the following issues as potential improvements related to this at one point:
https://github.com/uiowa/uiowa/issues/9342#issuecomment-3666507144
https://github.com/uiowa/uiowa/issues/4651#issuecomment-2887692041

joewhitsitt:
>We had ISPO provide some initial information from Splunk

Curiously waiting for what it reveals, but my surface level guess is that others are using CDNs and don't have the kind of cache expiration that we currently have (hourly for all files and site cache clears on deployment). I wonder how far back the data goes and whether we see a bump after we deployed https://github.com/uiowa/uiowa/pull/9170 (September)

It would be interesting to see the intersection of bandwidth/views/visits as it relates to the [emergency website endpoint](#9358). Legit traffic triggering a 5kb json response/requests per hour (1300+ websites where all loaded/viewed pages ping with JS every 30 seconds) adds up even if it doesn't count towards Views/Visits.

mjoneill:
Is uiowa.edu front page really loading all these videos (uiowa.edu):

<img width="1006" height="812" alt="Image" src="https://github.com/user-attachments/assets/0cac1bf9-d011-49a3-9c52-1ed678a2f624" />

joewhitsitt:
Another tell would be the percentage of apache requests vs varnish responses by analyzing splunk

joewhitsitt:
Capturing this Claude CoWork analysis of the "March" bandwidth data report.

# The caching change is working
[...]

PR #9528, opened 2026-01-20
WebP revisited

### LINKED ISSUE #8961: Support Next Gen Image Formats
# Story

As a website owner chasing page speed scores, I want the images displayed on my website to be served in a next gen format in an effort to reduce load times.

Splitting off from https://github.com/uiowa/uiowa/issues/5432 as that is more about the consideration of whether or not we continue to deliver images using responsive image. Recent adjustments to Drupal Core now support native lazy loading, responsive images and next gen formats so that we don't have to choose an either/or.

# Background

>AVIF support added with fallback to WebP
>Drupal 11.2 now supports AVIF in our image toolkit. AVIF offers better compression and image quality than WebP, especially for high-resolution images and HDR content. However, not all servers support conversion to AVIF. For that reason, a fallback mechanism was added to convert to WebP when AVIF support is not available. - https://www.drupal.org/blog/drupal-11-2-0

https://acquia.my.site.com/s/article/4402063525527-Improving-site-performance-with-WebP-images-on-Acquia-Cloud

#5432
#8868 

# Proposed Solution

- [ ] Investigate removing ImageMagick in order to support next gen formats confirming that image processing doesn't choke on large images and that iPhone image rotation handling still works. https://github.com/uiowa/uiowa/pull/7218#issuecomment-1863511354
- [ ] Enable next gen images through Drupal core and make sure it is working with our Acquia remotes.
- [ ] Confirm lazy loading, responsive images, and image style derivatives are all working based on our past woes https://github.com/uiowa/uiowa/issues/1476, https://github.com/uiowa/uiowa/issues/5121, https://github.com/uiowa/uiowa/issues/1406
- [ ] Update media view mode formatters to use next gen.

joewhitsitt:
Resolves #8961 
Relates to #9482
There might be some steps upstream at Acquia that we need to do first. This PR rips the band-aid off, but maybe we start with one site/feature to make sure things are working before rolling it out to all?

https://docs.acquia.com/acquia-cloud-platform/help/94381-how-enable-support-webp-images-acquia-cloud

~~Went with the contrib module instead of the core because of responsive images (`<picture>` and available fallback jpeg support), but I think we are only doing the former of that because our fallback image styles are also being converted to webp, we would have to create additional image styles to maintain a jpeg fallback. not sure if a fallback would be needed though https://caniuse.com/webp~~

~~**Update**: realizing that I might have mixed both the module and core functionality. the module doesn't mention converting each image style.~~

~~**Update2**: despite the default of 75% and the calls from Google Lighthouse to go lower, everything I have been reading recommends 80% or higher. We could consider that for our default Imagemagick setting (currently 90!) https://github.com/uiowa/uiowa/blob/main/config/default/imagemagick.settings.yml#L3~~

# How to test

`--sync-public-files` is recommended for your `blt ds`

Requires a `ddev restart` and stage_file_proxy needs to be uninstalled.

Optional: `drush image-flush` _all_ will regenerate existing, otherwise, newly uploaded or focal point changes should trigger a webp version.

```
ddev blt ds --site=admissions.uiowa.edu --sync-public-files && ddev drush @admissions.local pmu stage_file_proxy
```
or

```
ddev blt ds --site=uiowa.edu --sync-public-files && ddev drush @home.local pmu stage_file_proxy
```

In chromium devtools > network tab. record and refresh. sort size column and see that images compared to PROD are much smaller in size with no visible change in quality.


quamsta:
too early to ask for [AVIF](https://caniuse.com/avif)  ? : )

PR #9530, opened 2026-01-21
Htaccess file cache reset to a year with exceptions

### LINKED ISSUE #9341: Revisit the replacing file and overwrite checkbox?
## Issue

The file replacement checkbox and its intended use continues to be a source of confusion.

> Coud the issue or recommended procedure for replacing documents and the overwrite checkbox be revisited or reviewed after this afternoon’s demo?
I remember that Joe covered it in a recent demo, but I’m unclear on the details, effects of using or not using the checkbox and what happens when the checkbox is used or not used.  There was also a component of whether or not a user cache refresh would triggered.
Anyway, I was trying to go over it earlier this week with a content manager, and I clearly did not take careful enough notes.
Appreciate you considering.
Thanks so much,
~Viki

## Possible steps?

Re-visit documentation and add clarification. Maybe include more "use this in this situation" type examples
Possibly can switch the checkbox to off by default, as more often than not, this is probably the right choice to use

### LINKED ISSUE #9514: Increase cache lifetime for css/js and other types of files
# Motivation

As part of https://github.com/uiowa/uiowa/pull/9170, we reduced the cache lifetime of files in order to serve replaced documents quicker. We should revisit and consider increasing the cache lifetime for files that are not user managed documents. This came up as part of our review of best practices for loading and aggregating JS files (https://github.com/uiowa/uiowa/issues/9342#issuecomment-3666507144). This also may help with performance and with issues related to bandwidth (https://github.com/uiowa/uiowa/issues/9482).

# Additional Info

Slack conversation - https://iowaweb.slack.com/archives/C0164FNN1PV/p1761858921954739

>our site files directory serves both, files/2025-10/..., files/css, and files/js. it would be easier to match on those two frontend asset directories instead of all the different file types and date based directories. we might have to test which is supported (location or file type).  I'd love not to maintain a separate list of upload file types as regex (pdf|doc|docx|...) in our htaccess file. we started to go that route at one point and just went for everything to start the experiment.

- css
- js
- gz
- font
- images?

Update `.htaccess` using patch method

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9514
Resolves https://github.com/uiowa/uiowa/issues/9341

# How to test

On `main`

Sync a website with the `--sync-public-files` flag. Run a Google Lighthouse scan on the front page and see that it complains about inefficient cache lifetimes.

Screenshot 2026-01-21 at 10 46 28 AM

On `htaccess_file_cache_reset`

- `ddev restart`
- Go to the application of Chrome's DevTools and delete site data.
- `ddev drush @sitename.local cr`
- Re-run the Google lighthouse performance report and see that cache lifetimes are in the passed criteria section.

Screenshot 2026-01-21 at 10 49 51 AM

- Go to a media document (e.g. PDF) and load the direct URL in an anonymous browser session to establish a cached local version.
- `curl -i https://link-to-your-file.pdf` to see the cache headers of `cache-control: max-age=3600`
- Replace the PDF file and note that the checkbox to overwrite existing is unchecked by default. For this test, go ahead and check it to overwrite the file with your replacement PDF.
- I am still a bit confused by this though because I am seeing my anonymous browser fetch the replacement right away. That and we should have some invalidation in place still with `sitenow_media_wysiwyg_media_update`
- Reload the anonymous browser tab after an hour (or less) has passed to see your replacement file.
- Review the latest draft to the replace/caching docs https://sitenow.uiowa.edu/node/171/latest


GaryRidgway (approved):
Went through and tested all of this and it worked as described. Joe And I chatted a bit about how the real test will be when this goes live, but so far things look good. Well done!

PR #9536, opened 2026-01-23
Decorative banner videos

### LINKED ISSUE #9168: Siteimprove - Potential Issue - Mark banner videos as decorative
# Problem

Similar to the decorative images checkbox that was recently added to images, SiteImprove is flagging banner videos as potential issues for not having captions, etc.

## Flagged potential issues:
- Video missing captions
- Video without audio missing an accessible alternative

# Proposed Solution

Previous attempts and research indicate that the solution is likely going to be adding an `aria-hidden` attribute on the video element.

Do the above in a way that doesn't trigger a separate "Hidden element has focusable content". Previously, this might have happened because `aria-hidden` was applied to the Media entity wrapper div rather than the `<video>` element. It might have also been because there is a visible video control on banner images.

When this is done, update this draft article (or delete it and create a new one): https://webcommunity.sites.uiowa.edu/banner-video-flagged-missing-captions-or-accessible-alternatives

Revert #9577.

Optional: Add a description field to the video to allow users to provide a description that would be displayed appropriately as the text alternative and would suppress the `aria-hidden` behavior.

pyrello:
Resolves #9168 

# How to test

```
```
- Find or create a banner block with video media.
- Inspect the video element markup — the media wrapper should have `aria-hidden="true"`.
- Verify the video still plays inline as expected.
- Run a Siteimprove browser scan and confirm the video is no longer flagged as an accessibility issue.


joewhitsitt:
>Should we push out to test in PROD?

https://iowaweb.slack.com/archives/C8SM82ZFZ/p1769704444781169

GaryRidgway (approved):

PR #9537, opened 2026-01-23
Admissions - remove split for events view, set start in config

### LINKED ISSUE #9508: Remove calendar display from Admissions and tweak filters
# Motivation

As part of https://github.com/uiowa/uiowa/issues/9365, we reached out to Admissions to understand if they could get by without a calendar display because of existing accessibility issues. They are fine with removing this display as they don't have events far enough out in the future to justify the need for a calendar display. They would like some tweaks to the filters for the upcoming list display of events though.

# Additional Info

- would like to have the date filters enabled on that display (https://admissions.uiowa.edu/events). This is already a configuration option available
- would like the start date filter to default to `today` so that only upcoming events are displayed by default
- would like the end date filter to display right after the start date filter
- title search may be more appropriate first
- remove calendar display

Admissions doesn't need to review the changes if we can accommodate the above. They would just appreciate a heads up on when the changes will be deployed.

They are interested in understanding if we do end up with an accessible calendar display though.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9508

Admissions:

Screenshot 2026-01-27 at 11 18 42 AM

OnIowa:

Intradent:

# How to test

Check the following sites for regressions caused by setting the date filter similar to the events list block to `now` or from the URL.

```
ddev blt ds --site=admissions.uiowa.edu && ddev drush @admissions.local uli admin/config/sitenow/sitenow-events
```

No import errors. Configure the date and category filters to show. Save. Go to /events and see that the sidebar and calendar tab are gone compared to admissions.uiowa.edu/events. Defaults to today and still allows category filtering. Clicking the reset button will remove today's date from the filter and show all events because there is no value. On reload the input will be filled in again and filtered appropriately.

Scott has approved the changes in DEV.

For the intradent, it actually fixes a bug that was showing events from 2022 onward. Communication with Viki took place and they are happy and expecting the fix.

For oniowa, the filter was changed to a dropdown and that functionality still works as expected. You might need to create your own (e.g. Feb 3) as several of their events just ended. Confirm that it defaults to any date just like PROD https://oniowa.uiowa.edu/events

Screenshot 2026-01-27 at 11 34 07 AM

Current websites with the feature enabled:

- afr.fo.uiowa.edu
- intradent.dentistry.uiowa.edu
- inrc.law.uiowa.edu
- admissions.uiowa.edu
- oniowa.uiowa.edu


GaryRidgway (approved):
Tested on the listed sites and looks good!

PR #9554, opened 2026-01-29
[Charts] Added ability to add pie chart

bspeare:
Pillar project team has requested to use the pie chart. 

# How to test
**uiowa.edu (Has existing chart page)**
```
 ddev composer install && ddev blt ds --site=uiowa.edu && ddev drush @home.local uli /node/431/layout
```
1. Add a chart block, select "pie" option 
2. Upload a file below
3. Confirm that minimum and maximum colors (in screenshot below) are not allowed to be selected.
4. Adjust "Categories are identified by" to "first row" to get it to accurately display as a pie chart
5. Confirm you can add prefix and suffix and they will display in the tooltip
6. Confirm comma is added to dollar values
7. Check existing chart like https://uiowa.edu/explore/biomedical-engineering-degree/what-are-the-different-types-of-biomedical-engineering and see that prefix/suffix in tooltip still works
Screenshot 2026-01-29 at 4 35 09 PM

## CSV
Test using this file below:

- [pie.csv](https://github.com/user-attachments/files/24949625/pie.csv)

### DIFF
--- a/docroot/modules/custom/sitenow_charts/sitenow_charts.module
+++ b/docroot/modules/custom/sitenow_charts/sitenow_charts.module
@@ -54,6 +54,22 @@ function sitenow_charts_chart_definition_alter(array &$definition, array $elemen
     }
     $definition['yAxis'][0]['labels']['format'] = $format;
   }
+
+  // Handle pie charts separately because they use yaxis element but not yAxis definition.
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Works well!

If we opened up charts to a wider group of users, there's a lot of adjustments we'd want to make to the form, but for the narrow use, this is really slick.

PR #9556, opened 2026-01-30
[CKEditor] Interactive element size adjustments

### LINKED ISSUE #9504: Interactive element does not meet minimum size nor spacing - Siteimprove - CKEditor
From #9462

Interactive element does not meet minimum size nor spacing - 2.5.8 Target Size (Minimum) WCAG 2.1 AA

Generally there are buttons on the WYSIWYG toolbar (lists, etc.) that have drop buttons that do not provide the necessary spacing/size. Additional the about text format link below the editor is too small.

cory-skeers:
Resolves #9504 

# How to test

```
ddev blt frontend
```
Sync a site of your choice.
Edit a node, block, or anything else with a formatted text field.
Compare against a prod version, and see if you notice the differences (there are visual differences, but hopefully they are not jarring)
Run the Siteimprove browser extension
See that the "About text formats" link is not flagged as "Interactive element does not meet enhanced size (AAA)".
When a Filtered HTML (or above) format is selected, see that there are not a slew of "Interactive element does not meet minimum size nor spacing (AA)" targeting the dropdown arrow (for instance, next to the bulleted list toolbar buttons)
See that the "Note: This content will be published..." text is not flagged for "Line height is below minimum value (AAA)" on content with scheduling options (eg Article)

### DIFF
--- a/docroot/modules/custom/uiowa_core/uiowa_core.module
+++ b/docroot/modules/custom/uiowa_core/uiowa_core.module
@@ -2087,7 +2087,7 @@ function uiowa_core_scheduler_help_text(array &$form): void {
       $additional_help_text = t('Note: this content will be @stateed at the top of the hour after the time you select. (e.g., 2:43 p.m. will go live at 3:00 p.m.).', [
         '@state' => $state,
       ]);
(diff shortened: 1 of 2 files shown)


GaryRidgway (approved):
Checked all things described and this looks good!

Issue #9526, opened 2026-01-20
Low dependency updates continued and composer security update


briand44:
Here is an updated list with just lows remaining after https://github.com/uiowa/uiowa/issues/9486. We also discussed doing a composer security update as a follow up to https://github.com/uiowa/uiowa/pull/9487.

| Module Name | Current Version | Latest Version | Notes |
|---|---|---|---|
| [drupal/coder](https://git.drupalcode.org/project/coder/-/compare/8.3.28...8.3.31?from_project_id=1261) | 8.3.28 | 8.3.31 | we might have some coding standards to fix though |
| [drupal/imagecache_external](https://git.drupalcode.org/project/imagecache_external/-/compare/3.0.5...3.0.7?from_project_id=49732) | 3.0.5 | 3.0.7 | config update though |
| [drupal/inline_entity_form](https://git.drupalcode.org/project/inline_entity_form/-/compare/3.0.0-rc19...3.0.0-rc21?from_project_id=9858) | 3.0.0-rc19 | 3.0.0-rc21 | make sure default content, replicate, workflows aren't affected by access checks |
| [drupal/js_cookie](https://git.drupalcode.org/project/js_cookie/-/compare/1.0.1...1.0.2?from_project_id=119542) | 1.0.1 | 1.0.2 |  |
| [drupal/jsonapi_extras](https://git.drupalcode.org/project/jsonapi_extras/-/compare/8.x-3.24...8.x-3.27?from_project_id=59525) | 3.24.0 | 3.27.0 | update hook, confirm all is okay with hawk alerts endpoint |
| [drupal/jsonapi_include](https://git.drupalcode.org/project/jsonapi_include/-/compare/8.x-1.7...2.0.0?from_project_id=26968) | 1.7.0 | 2.0.0 | would like to confirm legacy code removal" does not affect hawk alerts endpoint" |
| [drupal/mailsystem](https://git.drupalcode.org/project/mailsystem/-/compare/8.x-4.4...8.x-4.5?from_project_id=9138) | 4.4.0 | 4.5.0 |  |
| [drupal/masquerade](https://git.drupalcode.org/project/masquerade/-/compare/8.x-2.0-rc4...8.x-2.0?from_project_id=51488) | 2.0.0-rc4 | 2.0.0 | config update. imo, it should not update the last active timestamp |
[...]

PR #9565, opened 2026-02-02
Update uiowa_auth, samlauth. adjust accessible-menu declaration

joewhitsitt:
# How to test

- Check passing build out to DEV https://app.travis-ci.com/github/uiowa/uiowa/builds/277384582
- Note that UIHC has already been running with an update to drupal/samlauth for a while now and this just bumps it a patch version.
- Test authentication to DEV websites (admissions.dev, sandbox.dev, etc.)
- Optional: Run tests locally in PHPStorm docroot/modules/uiowa/uiowa_auth/tests
- Run through these test instructions related to accessible-menu but it should be the same working commit as before. - https://github.com/uiowa/uiowa/pull/4956

```
ddev composer install && ddev blt frontend && ddev blt ds --site=tippie.uiowa.edu && ddev drush @tippie.local uli iowa-mba/curriculum-iowa-mba
```

PR #9573, opened 2026-02-04
DDEV Composer Update

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/9526

# How to test

## Locally on your host (e.g. mac):

- Self-service make me an admin
- `sudo composer self-update`
- `composer --version` _should be 2.9.5 or higher_

## DDEV

I think it is just `ddev restart` but I `brew upgrade ddev` just to do it all (careful though, the mutagen sync is awful).

- `ddev composer --version` _should match composer_version in .ddev/config.yaml_

## Acquia

Acquia manages composer upstream so if you ssh in you can see that it is also running above 2.9.3.

- `ddev drush @uiowa.prod ssh`
- `composer --version`

## Travis

We are using BLT Drupal Test as part of our build which has an old copy of composer/composer defined. We cannot update with dependencies because it is a deprecated/archived project.


joewhitsitt:
Not urgent/important. Whenever.

GaryRidgway (approved):
It looked ugly for me, but it worked! Took long though 😬

joewhitsitt:
I don't think it is a big deal with this merging into main. next time folks restart it should just update. The problem is that composer.lock has differences between composer/composer versions. so we could see unintended diffs between commits. https://github.com/uiowa/uiowa/pull/9565/changes/e98e07249d1bc42d4d2aa8399b4698966bb5f3c2#diff-f37acfaa6b11f575a9a6f41a75fa73a61d0f8ebc2c9b8cddc215d8aca10e44f5L24450-L24456

PR #9578, opened 2026-02-05
[charts] Pie chart CSS

bspeare:
This sets a min-height for the pie chart so it is more visible on mobile. 

# How to test
```
 ddev blt frontend && ddev blt ds --site=uiowa.edu && ddev drush @home.local uli     /explore/pre-med-programs
```
1. View the pie chart with the mobile emulator.  
2. Remove the CSS and confirm that the pie chart renders extremely small on mobile. 
3. Confirm that if right or left legend is selected, it will be forced to stack for smaller devices.


joewhitsitt (approved):
Lookin' good.

PR #9581, opened 2026-02-06
Piwikpro for more of sitenow beyond ccom

### LINKED ISSUE #9557: Enable Piwik on two CLAS sites
# Story

As a CLAS marketing team, we would like to enable Piwik on a couple sites that have clinical pages so we can have analytics and remain in compliance.

# Additional Info

- Department of Communication Sciences and Disorders (csd.uiowa.edu): - higher priority
- Department of Psychological and Brain Sciences (psychology.uiowa.edu):

They plan to use Piwik Pro across all pages on these two sites rather than Piwik Pro on clinical pages and GA on other pages.

We will need to update the core settings to disable Google Analytics and campus-wide GTM container.

Can we provide Webmaster permissions to configure Piwik Pro? I believe for CCOM sites only admins have permissions to configure.

We should communicate to them that if they have GA through GTM that they should remove/deactivate that.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9557

I made some guesses as to how all this should work. This results in no change to the ccom_core websites with a minimum one-step (optional two-step) process for other websites.

# How to test

```
ddev drush @medicinepathology.local cr && ddev blt ds --site=pathology.medicine.uiowa.edu && ddev drush @medicinepathology.local uli admin/config/services/piwik-pro
```

See that `sitenow_piwikpro` is installed during config import.

piwik_pro module is still enabled and configuration is in place for a ccom split site. Masquerade as a webmaster to make sure you can't access admin/config/services/piwik-pro

**Because the class file is moved, there is a window where ccom_core sites could error before they receive the `cr` on deployment, so this should probably be deployed off-peak time.**

See that piwik_pro assets aren't being loaded in Dev Tools Sources tab despite having credentials. Xdebug from PiwikProOverride to see it step through and evaluate that it is not PROD.

```
ddev blt ds --site=default && ddev drush @default.local en sitenow_piwikpro && ddev drush @default.local config-split:activate piwikpro && ddev drush @default.local uli --uid=11 admin/config/services/piwik-pro
```

Ensures the uiowa_core settings are unchecked (module install) and the webmaster permission (split activate) is imported.

Go to https://default.uiowa.ddev.site/admin/config/sitenow/uiowa-core to see both Google Tag Manager and Campus GTM are unchecked.

Just the piwikpro split can be activated, but the automated uncheck of the uiowa_core settings does not occur.

### DIFF
--- a/config/default/config_ignore.settings.yml
+++ b/config/default/config_ignore.settings.yml
@@ -19,6 +19,7 @@ ignored_config_entities:
   - 'config_split.config_split.grad_person_extended:status'
   - 'config_split.config_split.p2lb:status'
   - 'config_split.config_split.periodical:status'
+  - 'config_split.config_split.piwikpro:status'
   - 'config_split.config_split.profiles:status'
   - 'config_split.config_split.scholarship:status'
   - 'config_split.config_split.signage:status'
--- a/config/default/config_split.config_split.piwikpro.yml
+++ b/config/default/config_split.config_split.piwikpro.yml
@@ -1,19 +1,19 @@
-uuid: 9c723b89-48f9-4562-a000-1a89b5797ec8
+uuid: 1ebdf029-6047-4db5-be22-85b9140821e3
 langcode: en
 status: false
 dependencies: {  }
-id: ccom
-label: CCOM
+id: piwikpro
+label: PiwikPro
 description: ''
 weight: 80
 stackable: true
 no_patching: false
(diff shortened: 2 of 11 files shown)


GaryRidgway:
pulling to test

GaryRidgway (approved):
After chatting with Joe and doing a fair amount of testing, this all seems to be an order!

PR #9583, opened 2026-02-09
Empty H2 heading fix.

### LINKED ISSUE #9571: Empty h2 heading - SiteNow Dispatch Forms
# Motivation

We were notified of an accessibility compliance issue related to the ability to have an empty heading on SiteNow Dispatch Forms. We should address this issue at the platform level so we can meet compliance and avoid this issue surfacing again in the future.

<details>
  <summary>Original email notification from its-web</summary>

> Hi OSC web friends,
I’m seeing a A issue for “Empty headings” on https://writersworkshop.uiowa.edu/summer between the form introductory text and the form itself. When I look at the form and the text area component, I’m not seeing any empty H2s in the source.
How can I fix this? All advice is welcome - I tried deleting a space in the text area component like I have with past results like this and it didn’t change the page scan.

- Reported by Brynne in CLAS
</details>

# Additional Info

Example - https://writersworkshop.uiowa.edu/summer
Siteimprove browser extension is reporting "Empty Headings"

A 1.3.1 Info and Relationships
AA 2.4.6 Headings and Labels

From Cory:

>I poked at it, and it looks like there's no check if the headline field is empty, and ends up rendering an `h2` with no content in that case.

Image

>"Workaround" would be to add a heading, but in this case, the heading (and intro text) has been added already as a Text Area above it.

GaryRidgway:
closes #9571 

As a portion of the Definition of done for this issue, here are my findings of where else we may have issues with this:

- docroot/modules/custom/sitenow_dispatch/src/Plugin/Block/ThankYouBlock.php - LN:66
- docroot/modules/custom/uiowa_hours/src/Plugin/Block/HoursBlock.php - LN:101
- docroot/modules/custom/uiowa_maui/src/Plugin/Block/AcademicDatesBlock.php - Potentially both LN:300 & LN:303

# How to test
```

ddev blt ds --site=writersworkshop.uiowa.edu && ddev drush @writersworkshop.local uli /admin/config/sitenow/sitenow-dispatch

```

- Navigate to https://writersworkshop.uiowa.edu/summer
- Check the with the siteimprove extension with the AA filter to see that there is an issue with empty Headlines.
- Message me on slack for an api key to put in this field and when you have it, place it in this field and save.
- Navigate to https://writersworkshop.uiowa.ddev.site/summer
- Check the with the siteimprove extension with the AA filter to see that there are not any issues.
- Also check that adding a Headline to the form block doesn't break it.
- Try and break other things?


docroot/modules/custom/sitenow_dispatch/src/Plugin/Block/SignUpFormBlock.php
```diff
@@ -11,6 +11,7 @@
 use Drupal\sitenow_dispatch\DispatchApiClientInterface;
 use Drupal\uiowa_core\HeadlineHelper;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use function PHPUnit\Framework\isEmpty;
```
cory-skeers:
```suggestion
```

This looks like an extra add-in that wasn't used. I've had this happen with the phpstorm autocomplete before, where it suggests `isEmpty()` when `empty()` was intended.

GaryRidgway: Ah dang! thanks for that catch!!

joewhitsitt (approved):
works. thanks much!

Issue #9400, opened 2025-11-25
Remove text on Facilities alert form


briand44:
# Story

As a Facilities Strategic Communications team, we no longer need text on our alert form to remind collaborators to notify us when they create an alert for review since email notifications are configured to automatically send now.

# Additional Info

The text was added as part of https://github.com/uiowa/uiowa/issues/9059.

It is no longer needed.

Remove custom code that provides message on alert form - https://github.com/uiowa/uiowa/blob/9f44e5cbbac1ea0c1c98e1ea3359f6cf531c3f11/docroot/sites/facilities.uiowa.edu/modules/facilities_core/facilities_core.module#L104-L109

PR #9584, opened 2026-02-09
[Facilities] Remove alert help text (deploy on/after Feb 11)

cory-skeers:
Relates to #9400 

# How to test

```
ddev blt ds --site=facilities.uiowa.edu &&
ddev drush @facilities.local uli /node/add/alert
```
Scroll to the bottom, and see that the help message is no longer there.
Edit an existing alert, and see that there is no help message.

The help message in question:
image


joewhitsitt (approved):
I can confirm the text is no longer there. Masqueraded as a collaborator just in case there was another thing populating it.

I will let you determine when to merge it.

Issue #9570, opened 2026-02-03
Remaining low dependency updates


briand44:
Previous slice https://github.com/uiowa/uiowa/issues/9526 
| Module Name | Current Version | Latest Version | Notes |
|---|---|---|---|
| ~[drupal/masquerade](https://git.drupalcode.org/project/masquerade/-/compare/8.x-2.0-rc4...8.x-2.0?from_project_id=51488)~ | ~2.0.0-rc4~ | ~2.0.0~ | ~config update. imo, it should not update the last active timestamp~ |
| [drupal/media_entity_file_replace](https://git.drupalcode.org/project/media_entity_file_replace/-/compare/8.x-1.1...8.x-1.3?from_project_id=61974) | 1.1.0 | 1.3.0 | should just test the functionality. maybe increase scope to include https://github.com/uiowa/uiowa/issues/9341? |
| [drupal/menu_block](https://git.drupalcode.org/project/menu_block/-/compare/8.x-1.11...8.x-1.14?from_project_id=3161) | 1.11.0 | 1.14.0 | review the Display block if block content is empty" addition" |
| [drupal/menu_breadcrumb](https://git.drupalcode.org/project/menu_breadcrumb/-/compare/2.0.0-alpha0...2.0.1?from_project_id=59384) | 2.0.0-alpha0 | 2.0.1 |  |
| [drupal/moderated_content_bulk_publish](https://git.drupalcode.org/project/moderated_content_bulk_publish/-/compare/3.0.0-beta17...3.0.0-rc4?from_project_id=60824) | 3.0.0-beta17 | 3.0.0-rc4 |  |
| [drupal/node_revision_delete](https://git.drupalcode.org/project/node_revision_delete/-/compare/2.0.0...2.0.1?from_project_id=49284) | 2.0.0 | 2.0.1 |  |
| [drupal/pathauto](https://git.drupalcode.org/project/pathauto/-/compare/8.x-1.13...8.x-1.14?from_project_id=57542) | 1.13.0 | 1.14.0 |  |
| [drupal/purge](https://git.drupalcode.org/project/purge/-/compare/8.x-3.5...8.x-3.6?from_project_id=56717) | 3.5.0 | 3.6.0 |  |
| [drupal/purge_queuer_coretags](https://git.drupalcode.org/project/purge/-/compare/8.x-3.5...8.x-3.6?from_project_id=56717) | 3.5.0 | 3.6.0 |  |
[...]

PR #9598, opened 2026-02-13
Dependency updates cont.

cory-skeers:
Relates to #9570 

# How to test

## Masquerade

Sync a site and uli
Masquerade as another user
Note that the user's "last accessed" time is not updated
Make an update, create content, try to access routes, etc., and see that it successfully treats you as the masqueraded user.

## Views Bulk Operations and Moderated Content Bulk Publish

`ddev blt ds --site=now.uiowa.edu && ddev drush @now.local uli /admin/content`
Select multiple pieces of content and publish or unpublish
See that it works *except* if one of the pieces was under embargo, in which case it should fail gracefully (publish all that it could, no WSOD, and the embargoed piece remains unpublished). 

## Node revision delete

Code change review: https://git.drupalcode.org/project/node_revision_delete/-/compare/2.0.0...2.0.2

## Purge

Code change review: https://git.drupalcode.org/project/purge/-/compare/8.x-3.5...8.x-3.6

## Redirect

Sync a site, uli and visit /admin/config/search/redirect
See the view loads fine, with an additional 'enabled' column
Add a new redirect
Note the match redirect text describing wildcards
Create a redirect, possibly using a wildcard for the match redirect patch
See that it saves; test the new redirect and see that it works


bspeare (approved):
In testing these updates the modules appeared to work as expected!

cory-skeers:
Thanks @bspeare!

PR #9604, opened 2026-02-16
Facilities utility alerts feature

### LINKED ISSUE #9452: Facilities utility alerts
# Story

As a Facilities site owner, we would like to display utility alerts on our website so they can be visible to visitors of our site in addition to existing email, text and phone notifications to subscribed users.

# Additional Info

- API: http://utils-srv.nexus.facilities.uiowa.edu/utilities/alerts/api/report?days=14 - (The days attribute controls how long back we want to go)
  - credentials in our password manager
  - 30 second cache acceptable
- They have updated their alerts page in preparation for this data - https://facilities.uiowa.edu/fmalerts
- Similar to the Access and Construction alerts (with fewer categories)
  - Should only be two categories (information, emergency)
- Will need no results text
- Also provided a screenshot similar to ITS alerts but open to our expertise on design

Image

# Question

From Facilities:
>Can they invoke the API every time the page is requested? That way, whenever a user visits, they see the latest information

>the API is super quick and almost ready to share with you. He said load time is not an issue since this is a real-time alert—we just need the page to fetch new data from API every single time.

> Alerts emitted from the FM-Alerts application are available immediately as the API is invoked (real-time). Introducing any form of caching in front of the API layer—as described below—could create situations where the webpage displays stale or lagging alerts relative to the latest notifications generated by the system. As Rakesh has already highlighted, our expectation is that the Alerts API should be executed on every page load, ensuring the UI always reflects the most current alert chronology. Besides, there is no fixed time or frequency when the alters may be triggered, it is ad-hoc and mostly during odd hours.
> That said, if there are frontend or architectural constraints that require mandatory caching on the website, we recommend configuring the lowest viable TTL to preserve data freshness as much as possible.

# Proposed solution

pyrello:
Resolves #9452 

# How to test

- The API token must be configured locally. Add the following line to `docroot/sites/facilities.uiowa.edu/settings/local.settings.php`:
```php
$config['uiowa_facilities.apis']['utility_alerts']['auth']['key'] = 'YOUR_API_TOKEN';
```
- The token is in password manager or you can ping Sean for it.
```
ddev blt ds --site=facilities.uiowa.edu && ddev drush @facilities.local uli /
```
- The utility alerts block needs to be placed via Layout Builder
- Verify alerts load asynchronously into the block container
- Confirm that the block cache is not busted by the 30 second cache lifetime of the data.
- Confirm each alert displays as a card with a date, alert type badge, and description.
- To test the empty state, temporarily change `getAlerts(30)` to `getAlerts(7)` in `UtilityAlertsController.php` to narrow the date range so no alerts are returned.
- Verify the "No utility alerts at this time." message displays when no alerts are returned.

## Note

- The default for the number of days to fetch alerts for needs to be set to 14 before this can be merged.
- This will be tested in dev before getting merged.


docroot/modules/custom/uiowa_facilities/src/UtilityAlertsApiClientInterface.php
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * Get utility alerts.
+   *
+   * @param int $days
+   *   The number of days of alerts to retrieve.
+   *
+   * @return array|false
+   *   The alerts data or FALSE on failure.
+   */
+  public function getAlerts(int $days = 14): array|false;
```
pyrello: Does it seem right to set a default for the number of days in this interface?

docroot/sites/facilities.uiowa.edu/modules/facilities_core/js/utility-alerts.js
```diff
@@ -0,0 +1,24 @@
+(function (Drupal, drupalSettings, once) {
+  Drupal.behaviors.utilityAlerts = {
+    attach: function (context) {
+      once('utility-alerts', '.utility-alerts-container', context).forEach(function (container) {
+        var url = drupalSettings.facilities_core.utilityAlertsUrl;
+
+        fetch(url)
+          .then(function (response) {
+            if (!response.ok) {
+              throw new Error('Network response was not ok');
```
pyrello: Need a better error message.

pyrello:
> Looks like we are hardcoding h2s. Might we add the headline field here so that child headings can adjust to fit the nesting?

Adjusted in most recent commit. It was an interesting problem to solve. Have to set the block heading setting in drupalSettings, read it in the JS, and pass it to the controller. 

If we think we are going to be building more blocks like this in the future, might want to make some of this reusable.

GaryRidgway:
I'll pick this up for review!

GaryRidgway (approved):
Everything looks good to me! The only issue I had was testing the cacheing, but working with Joe in webteam allowed me to come to the conclusion that I had enough info to feel good about this.

APPROVED!

pyrello:
> I got this setup on DEV only to discover that (for reasons I don't understand) the formatting looks like @joewhitsitt's screenshot above: [#9604 (comment)](https://github.com/uiowa/uiowa/pull/9604#issuecomment-3917295226)

This was because the `card` library wasn't getting attached. It "worked" on test pages that already had other cards on them and looked off when tested in isolation.

pyrello:
This was approved by the customer.

PR #9606, opened 2026-02-16
Fontawesome version sync

### LINKED ISSUE #9616: Old `fa` markup for svg replaced fontawesome icons causes missing icons
# Problem

When we upgraded the fontawesome library to 6.7.2 when we switched to CDN, it caused some of our fontawesome icons in the social media menu to render as missing svgs. We were hardcoding the `fa` class in the markup which can cause a mistmatch/missing icon.

Inline icons that aren't converted to svgs are not affected at this time, but their classes should update as they are revisited.

<details>
<summary>Customer provided details</summary>

> Yes, I made sure to save the footer and tried again, but still no luck.
> Not sure if it matters, but since you mentioned permissions, I believe I have all possible permissions, as far as I can tell:
> Image
> 
> Also, I noticed that the issue doesn't seem to apply to other parts of the site, like text blocks.
> For example, on [this page](https://wics.org.uiowa.edu/uicc-2026/uicc-2026-parking), you should be able to see a few location-arrow icons.
> 
> I looked into that a bit more, and I actually think I’ve found the problem!
> 
> When the page embeds the footer icons, the HTML it includes looks like this:
> `<span role="presentation" class="fa fa-solid fa-building-columns" aria-hidden="true"></span>`
> 
> However, in a text block, for example, it looks like this:
> ```
> <span class="fontawesome-icon-inline">
> <span class="fa-solid fa-location-arrow" role="presentation">&nbsp;</span>
> </span>
> ```
> 
> They’re a bit different, but I noticed the footer icons have the “fa” class while the others don’t.
> In the DevTools of my browser, if I modify the network request for the page to eliminate this class from the footer icons, the icons actually load!
> The branding icons aren’t impacted either, they still work as well.
> 
> This doesn’t appear to be something I can fix on my own though, since the class is injected by SiteNow.
> For example, here’s what I put in our footer, which resulted in the first bit of HTML above:
> Image
> You’ll note the “fa” class isn't there.
> 
> Maybe getting rid of this class could be a solution?
> It seems to not have been used since Font Awesome v4 (released in 2014!)
> 
> The root cause seems to me to be that SiteNow is loading Font Awesome twice, possibly by accident: once through the site theme (through a CSS file, which maps icons to special characters, rendered via a font), and once through the more modern Font Awesome JavaScript library, which renders things differently (as an SVG).
> This seems to be a common issue, and it’s mentioned on Font Awesome’s troubleshooting page:
> Image
> 
> I’ve added an example item in our social media footer for testing purposes, which you can use as a reference if you’d like.
> 
> Apologies for the delay in getting back to you!
> Hope this is helpful!
> Thanks again,
> - Claire

</details>

https://github.com/uiowa/uiowa/pull/9606

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9616

Using https://libraries.io/bower/Font-Awesome/6.7.2 for the vanilla icon picker in the social media menu form which matches the CDN delivered via the fontawesome module https://github.com/uiowa/uiowa/blob/main/config/default/fontawesome.settings.yml#L7

The presence of the old `fa` class on SVG converted icons cause problems with the lookup.

# How to test

```
ddev composer install && ddev blt frontend && ddev blt ds --site=wics.org.uiowa.edu
```

It should load the building icon at the end (in addition to the castle)

Login and try icons that we purposed updated to support (x and tiktok)

```
```

Test event blocks, timelines located on this website. Creating and reviewing existing.

https://sandbox.prod.drupal.uiowa.edu/timeline
https://sandbox.prod.drupal.uiowa.edu/event-combinations
https://sandbox.prod.drupal.uiowa.edu/icons - there are missing icons but they should all be PRO icons that we don't support and shouldn't be available for selection again.

Our help text reads:
"If an icon below is displayed with a question mark, it is likely a Font Awesome Pro icon, unavailable with the free version of Font Awesome."

Make sure this non-svg version still renders on https://registrar.uiowa.ddev.site/course-deadlines:

```
ddev blt ds --site=registrar.uiowa.edu
```

Use the following to filter to a course that features the info icon.

Screenshot 2026-02-16 at 4 01 26 PM

# Follow-up

Contact Claire S. from its-web to confirm fix.


bspeare (approved):
The icons are working as expected now.  Thanks for figuring out this fix!

Issue #8181, opened 2024-10-29
BLT end-of-life


briand44:
# Story

As a web team using BLT, we would like to begin transitioning off BLT since it has reached end-of-life.

# Additional Info

https://github.com/acquia/blt/issues/4736

>BLT will not be updated to support Drupal 11 – so, if and when you plan to upgrade to Drupal 11, Acquia’s BLT will be a blocker if it’s still in your project.

>Acquia will end support of BLT at the end of 2024 (December 31, 2024) and publicly archive the repository at that time

https://dev.acquia.com/tutorial/you-dont-need-blt-acquia-cloud

Multisite specific recommendation: https://github.com/acquia/drupal-recommended-settings. We cannot use cloud actions to replace cloud hooks.

# Questions

* Curious about our thinking re: adopting GrumPHP as part of our replacement strategy
* Curious about moving off travis (paid) in favor of github actions (free) given that we will be touching a lot of this at the same time.

# Proposed Solution

joewhitsitt:
I think initial slices should be to move our helper functions like `umc` and `umd` out of BLT to the point where we only use BLT for deployment.

GaryRidgway:
Grooming conversation...
- Remember us saying that we might potentially keep BLT, but seems like there are other options that are okay.
- BLT is large, perhaps postpone grooming on this?
- Should we be making a slice to probe where BLT is in our codebase to understand the scope of removing it.

pyrello:
> https://dev.acquia.com/tutorial/you-dont-need-blt-acquia-cloud

This is useful for thinking about how to go about this switch. I was thinking that we would need to replace all the BLT commands with our own bespoke Robo implementations, so it is a relief to know we could potentially use composer scripts for some of it.

I'm noting that they don't provide a recommendation for replacing one of our most used commands `blt ds` (or `blt dsa`), so we would still need to figure out what that looks like.

We have a fair amount of customization and a whole set of custom commands (like the config split re-exporter). Even with the above instructions, I think there is a fair amount of complexity here until we identify everything that needs to be accounted for  and replaced, and how we are going to do it in each case.

I'm cool with an initial slice to switch an existing command or two over to a new implementation to learn something, but I feel like at some point we are going to need a full audit of what needs switched to really understand how much work this is.

joewhitsitt:
[...]

PR #9607, opened 2026-02-16
Replace BLT settings includes with project-owned settings

pyrello:
Related to #8181 

# How to test


joewhitsitt:
I think it would be good to get blt umc up to date with this too so if this merges, we can still do provisions properly in the short term.

pyrello:
Closing this because #9858 will use a different approach.

Issue #9549, opened 2026-01-28
Add data attributes to banner link field


briand44:
# Story

As a content editor managing campaigns across our website, I want links to have data attributes for tracking instead of relying on URL parameters or div IDs, so that I can update or rearrange page layouts without breaking our conversion tracking.

# Additional Info

Epic - https://github.com/uiowa/uiowa/issues/9459

Previous slice to address menu links - https://github.com/uiowa/uiowa/issues/9531
PR - https://github.com/uiowa/uiowa/pull/9543

Grooming conversation - https://github.com/uiowa/uiowa/issues/9459#issuecomment-3813505298

Some potentially useful info from epic:

When tracking is enabled, render these attributes:

```html
<!-- Button component -->
<a href="/apply"
   class="btn btn--primary"
   data-sn-event="apply_now_click"
   data-sn-event-type="click"
   data-sn-event-component="button"
   data-sn-event-label="Apply Now">
  Apply Now
</a>

<!-- Link component -->
<a href="/programs"
   data-sn-event="view_programs"
   data-sn-event-type="click"
   data-sn-event-component="link"
   data-sn-event-label="View all programs">
  View all programs
</a>
```

**Attribute Contract:**
- `data-sn-event` - webmaster's event name (required)
- `data-sn-event-type` - always "click" (platform-defined)
- `data-sn-event-component` - SiteNow component type: "button", "link", etc. (platform-derived)
- `data-sn-event-label` - webmaster's label or link text (platform-handled)

**Key principle:** Platform defines keys, webmasters only provide values.

cory-skeers:
Draft updates are available at https://sitenow.uiowa.edu/node/926/latest and https://sitenow.uiowa.edu/node/281/latest. Pretty minimal updates; changes to the conversion tracking page to not be menu-specific, and an addition to the banner page to link to the conversion tracking page.

PR #9621, opened 2026-02-20
Banner link attributes

cory-skeers:
Relates to #9549 

# How to test

```
ddev blt di --site=default
```
Check for a clean install for the config changes

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli /banner-buttons
```
Inspect a few banner buttons, and see that they don't have any of the `data-sn-*` attributes on them
Edit the layout
Edit existing banners and add new banners and use the new "attributes" option for one or more links
Inspect the markup on the page, and see the new `data-sn-*` attributes

1. `data-sn-component` should always be "button"
2. `data-sn-event-type` should always be "click"
3. `data-sn-event-name` should match what was entered on the banner link
4. `data-sn-event-label` should match what was entered or, if left blank, should match the link title

If there is no `data-sn-event-name` entered for the link, none of the `data-sn-*` attributes should be present. Try removing one from a link that had one, and see that the page markup updates

If the banner has a single button and a banner title, the link should be in the headline, and the attributes attached there
If the banner has a single button and no banner title, or if the banner has more than one button, the attributes should be attached to the button link(s)

## Tests
My preferred way to run these is by editing `blt/local.blt.yml` and adjusting the drupal test definitions. Something like:
```
# Configure Drupal tests to run on ddev
tests:
  drupal:
    web-driver: false
    sudo-run-tests: false
    mink-driver-args-webdriver: ''
    simpletest-db: ''
    simpletest-base-url: 'http://web'
    phpunit:
      -
        # The directory to scan for tests. Change to what you want to test.
        directory: '${docroot}/modules/custom/uiowa_core'
        # Use the core PHPunit config file.
        config: '${docroot}/core/phpunit.xml.dist'
        filter: BannerLinkAttributesTest
```
Then you can run `ddev blt tests:drupal:run`
There are two new tests:
in `layout_builder_custom`: `BannerBlockFormHandlerTest`
and in `uiowa_core`: `BannerLinkAttributesTest`

The first tests the validation handling in the banner block form, the second tests the actual rendered markup (and through that, the form validation handling).

# Follow-up
Part of #9549 also includes documentation updates that are to come


joewhitsitt (approved):
Looks good. Manual and automating testing passed. Out of curiousity I tried removing the schema update and saw that it caused the uiowa_core test to fail. Nice find and addition. 👍 

Default content didn't seem to mind that the new fields were not set. I was able to save the existing banner.

Replicate, replicated the values.

PR #9626, opened 2026-02-23
Adjust parallel deployment logging

joewhitsitt:
AI explained that the logrotate rule I had for parallel deployment logs was triggering but not finishing the job because of the datatime specific naming convention. We had the datetime specific convention to be able to grep for "error" afterward. The compromise is to log to a temp file for the grep so it is still deployment specific, then copy and append the result to a main log file that is then rotated. For someone sifting through the logs, clear start and end comments are included between deployments.

# How to test

- Code review, but needs to be deployed for it to be tested.

# Follow up

- Once again clean up past attempts at rotating the parallel deploy logs (all envs/apps)


cory-skeers (approved):
I think the updates look good, based on my understanding of the described problem. The switch from a datetime id to uniqid also seems good if it's only ever going to be used as a temporary file, and moving the datetime to the START and END appends should hopefully still keep handy searching options.

Godspeed to this next attempt 🚀

PR #9646, opened 2026-03-02
[Registrar] Updated correspondence cleanup with Dispatch VISIBLE+TAG filter

### LINKED ISSUE #9650: Registrar correspondence block displaying Dispatch campaigns that are marked as hidden
# Problem 
Image

An internal memo but was inadvertently attached and sent through a public facing Dispatch campaign. When they realized they went in and selected the "Hide Issue" flag in Dispatch which removed it from the public archives and thus giving us a broken link on the correspondence page.

# Proposed solution
- Check for `VISIBLE=true&TAG=registrar` and delete items that no longer have those attributes. We are currently only deleting if items age out only.

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9650. 

Fixes an issue for when the "Hide Issue" flag in Dispatch is set and removes it from the public archives and thus giving us a broken link on the correspondence page.

# How to test

```
 ddev blt ds --site=registrar.uiowa.edu && ddev drush @registrar.local uli  /correspondence
```
1. Search for [Accessible Content Project Resources](https://apps.its.uiowa.edu/dispatch/archive/accessible-content-resources-W8ZwwvE) and confirm the link is broken. 
2. Add Dispatch key from https://registrar.uiowa.edu/admin/config/sitenow/sitenow-dispatch to https://registrar.uiowa.ddev.site/admin/config/sitenow/sitenow-dispatch
```
ddev drush @registrar.local state:set registrar_core_daily_last 2000-01-01  && ddev drush @registrar.local  cron
```
1. Confirm broken link has been removed. 
2. You should see a message ` [notice] 17 archives removed because they are no longer visible in Dispatch.`
3. The 17 items are being removed because the sync now performs both inclusion and cleanup using the same Dispatch filter (VISIBLE=true&TAG=registrar). Those params already existed before, but previously the code only upserted returned rows and never deleted rows that dropped out of that filtered result set (except age-based pruning), so stale entries accumulated. With the new cleanup, Drupal now deletes correspondence_archives rows in the 2-year window when their URL is no longer returned by that filtered API query. That means items can be pruned even if their public archive URL still loads, if they are no longer in the filtered Dispatch set (for example, no longer tagged registrar); truly hidden items are removed for the same reason.

**Example pruned item(s):**
```
{
    "slug": "academic-calendar-replication-review-D0jBzLn",
    "title": "REMINDER: Review Your 2026 Academic Calendar Dates",
    "createdOn": "2025-07-24T11:32:23-0500",
    "hidden": false,
    "tags": [
        null
    ]
}
{
    "slug": "accessible-content-resources-W8ZwwvE",
    "title": "Accessible Content Project Resources",
    "createdOn": "2025-11-12T12:53:30-0600",
    "hidden": true,
    "tags": [
        null
    ]
}
```
**Example item with tags that is not pruned:**
```

{
    "slug": "textbook-assignment-reminder--instructors--summer-GYvyyRL",
    "title": "Summer 2026 Textbook Assignment Reminder - Instructors",
    "createdOn": "2026-02-23T13:01:50-0600",
    "hidden": false,
    "tags": [
        "faculty/staff",
        "registrar"
    ]
}
```

### DIFF
--- a/docroot/sites/registrar.uiowa.edu/modules/registrar_core/registrar_core.module
+++ b/docroot/sites/registrar.uiowa.edu/modules/registrar_core/registrar_core.module
@@ -141,8 +141,11 @@ function registrar_core_post_submit(&$form, FormStateInterface $form_state) {
 
 /**
  * Retrieve correspondence data from Dispatch.
+ *
+ * @param int $timestamp
+ *   The UNIX timestamp after which records should be retrieved.
  */
-function registrar_core_fetch_correspondence_from_dispatch() {
+function registrar_core_fetch_correspondence_from_dispatch(int $timestamp) {
   $logger = \Drupal::logger('registrar_core');
   $logger->notice('Fetching correspondence archives.');
 
@@ -151,8 +154,6 @@ function registrar_core_fetch_correspondence_from_dispatch() {
   // so stale local rows can be removed.
   $visible_archive_urls = [];
   $database = \Drupal::database();
-  // Two-year retention window (63,072,000 seconds).
-  $cutoff = time() - 63072000;
   $endpoint = 'https://apps.its.uiowa.edu/dispatch/api/v1/';
   $dispatch_params = [
     'VISIBLE' => 'true',
@@ -173,7 +174,7 @@ function registrar_core_fetch_correspondence_from_dispatch() {
     // Fetched archives are in descending chronological order
     // and we're only keeping ones that are younger than the cutoff,
     // so once we've hit one that is older, we know we're done processing.
-    if (strtotime($archive->createdOn) < $cutoff) {
+    if (strtotime($archive->createdOn) < $timestamp) {
       break;
     }
     $key = basename($archive_url);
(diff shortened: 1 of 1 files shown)


docroot/sites/registrar.uiowa.edu/modules/registrar_core/registrar_core.module
```diff
@@ -166,11 +172,13 @@ function registrar_core_fetch_correspondence_from_dispatch() {
     // and we're only keeping ones that are 2 years or younger,
     // so once we've hit one that is older than 2 years old
     // (63072000 seconds) we know we're done processing.
```
cory-skeers:
```suggestion
// Fetched archives are in descending chronological order
// and we're only keeping ones that are younger than the cutoff,
// so once we've hit one that is older, we know we're done processing.
```
The cutoff of 2 years and its time in seconds is defined higher up. Changing this comment to only reference the "cutoff" and not a specific time amount can help keep it from getting out of sync, if the cutoff timeframe gets changed above. (Eg, the window gets changed to one year, but this comment doesn't get updated)

cory-skeers (approved):
Nice updates! Solves the immediate issue and aligns things more accurately with the upstream.

I made one more adjustment while thinking about the cutoff time and its possibility for being changed in the future. The fetching and pruning time should always be the same for how this is set up, and now it's set in a single spot for both. Functionally the update shouldn't change how anything worked.

PR #9649, opened 2026-03-02
Banner videos -- aria-hidden

### LINKED ISSUE #9168: Siteimprove - Potential Issue - Mark banner videos as decorative

cory-skeers:
Resolves #9168 

# How to test

```
ddev blt ds --site=medicine.uiowa.edu
```
Visit https://medicine.uiowa.ddev.site/
Siteimprove browser extension comes up with only AAA issues unrelated to the banner video
Inspect the video--see that the video itself is `aria-hidden="true"`
Inspect the video controls, and see that they are not hidden (and/or tab through the page, and see that the button is focusable)

Any feedback on the reasoning or understanding here: https://github.com/uiowa/uiowa/pull/9649#issuecomment-3991594694

### DIFF
--- a/docroot/themes/custom/uids_base/assets/video/decorative-video-captions.vtt
+++ b/docroot/themes/custom/uids_base/assets/video/decorative-video-captions.vtt
@@ -1,4 +1,0 @@
-WEBVTT
-
-00:00:00.000 --> 00:00:01.000
-[background video, no audio]
--- a/docroot/themes/custom/uids_base/templates/field/file-video.html.twig
+++ b/docroot/themes/custom/uids_base/templates/field/file-video.html.twig
@@ -31,7 +31,4 @@
   {% for file in files %}
     <source {{ file.source_attributes }} />
   {% endfor %}
-  {% if attributes.storage['data-decorative'] %}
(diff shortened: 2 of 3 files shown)


joewhitsitt (approved):
At EoD in my head i equated aria-hidden as not playing or being in the DOM at all and thus no concern to the user. Thanks for the explanation and thoughts. I agree with this approach.

PR #9657, opened 2026-03-04
[Commencement] Non-venue addresses for events

### LINKED ISSUE #9602: Commencement: import off-campus event locations
# Story

As a Commencement site owner, I would like to import custom locations from events.uiowa.edu into our Commencement ceremony nodes so that visitors can have location information for off-campus ceremonies.

# Additional Info

<details>
<summary>Original notification from Josh H.</summary>

>Good morning! It seems that some events that we have synced between the events hub and the commencement site aren't bringing over locations if they are not in the dropdown campus building list:
> 
> https://events.uiowa.edu/event/35374
> 
> https://commencement.uiowa.edu/ceremony/spring-2026/sioux-city-masters-social-work-graduation-celebration
> 
> I'm hoping you can take a look and see if there is something going on? I did make an edit on that event hoping to trigger an update in the next sync but someone also mentioned something so thought I would send a note.
> 
> Let me know if you need additional info or details.
> 
> Many thanks!
> 
> Joshua H.

</details>

Venue mapping happening here:
https://github.com/uiowa/uiowa/blob/129e08d1f04d4951bb30095ba4d6d2002b19eb03/docroot/sites/commencement.uiowa.edu/modules/commencement_core/src/EventsProcessor.php#L104-L108

- Create fields on the event content type to store Company, Address, City, State, Zip, Country
  - Won't need to output Country
  - events.uiowa.edu only allows United States for Country
  - should not be able to edit this info
- Map custom address from events.uiowa.edu to these new fields
- Output Company in teaser, not linked if there is no venue but is a custom location
- Output full address, without country on node display if there is no venue but is a custom location
- Commencement site owner can make changes to events.uiowa.edu to force update once new fields are mapped

cory-skeers:
Resolves #9602 

# How to test

```
ddev blt ds --site=commencement.uiowa.edu && ddev drush @commencement.local import:events && ddev drush @commencement.local uli /ceremony/university-college-commencement-ceremony
```
Compare page to prod; there should be no changes.
Go to /admin/content
View one of the most recent Events (location field should be updated, so most recent ones should be good test cases)
See that it has a venue with a name and address
View the event on the the Events site, and check that the name/address match
Edit the Event as a webmaster
See that the Venue Name and Venue Address fields are in the "imported" section, and not editable


cory-skeers:
Updated off-campus venue display:
image

I updated which field/s the address was pulled from so we could control how it gets put together, including adding a line break

GaryRidgway (approved):

PR #9659, opened 2026-03-05
[sitenow_media_wysiwyg] Set Panopto media name from video title (#5029)

### LINKED ISSUE #5029: Get a better default Panopto media name.
docroot/modules/custom/sitenow_media_wysiwyg/src/Plugin/media/Source/Panopto.php:138

# Problem
When Panopto media is created, it gets a hash value for the media name. This can make it difficult to identify media in the media library. It'd be nice if the name of the item in UICapture could somehow be used.

# Proposed solution
Leverage the API or another mechanism to get a better name. Details TBD.

jeremyhastings:
Fetches the title from the Panopto embed page to use as the default media name, falling back to the machine-generated name on failure.  Fixes #5029.

# How to test

Find a WYSIWYG, use UICapture, see a name that is not a unique hash.


joewhitsitt (approved):
Nice! Much needed improvement. I think it is good that it only handles media on creation. Thanks for 255 safeguard.

Image

PR #9662, opened 2026-03-05
Add aria-atomic to CKEditor 5 live regions

### LINKED ISSUE #9506: Field input error is not announced in full - Siteimprove - CKEditor live region
From #9462

Field input error is not announced in full
3.3.1 Error Identification A
4.1.3 Status Messages AA

CKEditor live announce region have an aria-live assertive value. There seems to be two of these per page. We changed this to polite for uiowa_hours recently but since CKEditor also provides polite containers, I don' think that is the solution here. https://github.com/uiowa/uiowa/pull/9426/changes

Image

Unless, adding aria-atomic=true to this assertive divs makes SiteImprove happy, I think this is another situation where SiteImprove doesn't like the look of this but would be up to CKEditor to adjust.

pyrello:
Resolves #9506 

# How to test
- On `main`:
```
ddev blt ds && ddev drush @default.local uli admin/config/sitenow/uiowa-articles
```
- Visit the above page and run the Siteimprove browser plugin
- It is expected that you'll see two instances of `Field input error is not announced in full`
- On `ckeditor_live_regions_a11y`:
```
ddev blt ds && ddev drush @default.local uli admin/config/sitenow/uiowa-articles
```
- Visit the above page and run the Siteimprove browser plugin
- Confirm there are no instances of `Field input error is not announced in full`


pyrello:
Code has been adjusted to only attach `aria-atomic` to assertive live region.

joewhitsitt (approved):
Thanks for making the change. Confirmed.

Image

PR #9664, opened 2026-03-06
thesis defense - adjust filter labels

joewhitsitt:
>We are wondering if we could make a change to the search functionality. Could we change the label from “Name” to “Search by Name”? The same for “Date”, or could we change the “Apply” button to “Search”?

# How to test

```
ddev blt ds --site=grad.uiowa.edu && ddev drush @grad.local uli academics/thesis-defenses
```

See the filters have been updated. No config import errors.

### DIFF
--- a/config/features/thesis_defense/views.view.thesis_defenses.yml
+++ b/config/features/thesis_defense/views.view.thesis_defenses.yml
@@ -96,7 +96,7 @@ display:
       exposed_form:
         type: bef
         options:
-          submit_button: Apply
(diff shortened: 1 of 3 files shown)


joewhitsitt:
ack. there is a WSOD issue with this in PROD.

bspeare (approved):
Changes look good.  I removed CSS rule that was preventing spacing between the input fields and added heading classes to the dates for more visual distinction between the list items. Image

PR #9673, opened 2026-03-10
Dependency updates: editor_advanced_link and linkit

GaryRidgway:
# How to test

For editor_advanced_link:
- open a page and go to edit the layout
- add a text block
- for minimal and minimal + html formats:
    - create a link and set advanced values
        - title, aria-label, id, rel, target
    - ensure advanced is above the check and close buttons
    - once you set all values, save the link.
    - open the link again, check all your values work.
    - edit an existing link and see all values are maintained.
    - confirm no browser console errors.
- for filtered html and full html
    - do the above but there is also a class field in advanced.
- confirm on render all links render correctly.

For linkit:
- open a page and go to edit the layout
- add a text block
- For minimal, minimal+, filtered_html, and full_html
- Open the link dialog and confirm Linkit autocomplete appears.
- Search and insert each of these link types:
    - node
    - media
    - webform
    - front page
- save and ensure all values are maintained.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -292,9 +292,6 @@
             "drupal/draggableviews": {
                 "Completely hide the weights field for anonymous users": "https://www.drupal.org/files/issues/2021-08-31/3229357-7.patch"
             },
-            "drupal/editor_advanced_link": {
-                "[3371633] Fix for new window option not rendering under the\n advanced group on initial dialog open.": "patches/14.patch"
-            },
             "drupal/entity_reference_revisions": {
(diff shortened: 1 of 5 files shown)


GaryRidgway:
[This issue keeps occurring and is definitely a regression.](https://github.com/orgs/uiowa/projects/20/views/1?pane=issue&itemId=164063489&issue=uiowa%7Cuiowa%7C9674) unsure if this is a blocker or if we should roll a patch from what codex made or not until we get to 2.3.x?
[
Found this issue where people were experiencing similar things, but it doesn't seem like there's a patch that cleanly just applies onto our codebase.](https://www.drupal.org/node/3371633)

Issue #9529, opened 2026-01-20
Vertical video TypeError this.video_btn.id


joewhitsitt:
# Problem

Noticed a console error while analyzing page load.

```php
// Give each video button an id so that we can index them individually later.
        // As well, no two elements should have the same ID, so assigning them like this ensures that is the case.
        this.video_btn = this.element.querySelector('.video-controls .video-btn');
        this.video_btn.id = this.video_btn.id + '-' + index;
```

`Uncaught TypeError: Cannot read properties of null (reading 'id')
    at new Video`
`TypeError: null is not an object (evaluating 'this.video_btn.id')`

# Steps to reproduce
Inspect console for uiowa.edu

# Proposed solution

joewhitsitt:
This is affecting performance scans on the homepage

joewhitsitt:
I think this is related to the vertical videos on the page vs the banner video which is why i am not seeing this on other sites.

PR #9680, opened 2026-03-12
Hawk alert fetching improvements

joewhitsitt:
This PR has moved in a different direction after getting clarity surrounding the issue on 3/10. The homepage was affected by https://github.com/uiowa/uiowa/issues/9529 and that has been resolved and pushed. Therefore the concerns about Acquia not respecting the 30 max age cache header is not an issue. The no cache adjustment has been removed from the PR but the logging improvements, script syntax improvement and config clean up remains.

<details>
<summary>Old details</summary>
# Background

There was a severe weather alert that went out on 3/10. Initially it was bad text that was revised in the emergency website. Another alert went out later and was closed within five minutes of import. Between the two, folks noticed the hawk alert was on the emergency.uiowa.edu website served by views as expected. Seeing the alerts served by uiowa_alerts on other websites though was inconsistent. Notably the homepage's front page did not show the alert.

# In this PR

So to get a fresh set of "eyes" on this, I fed Claude chat all the things (the related files in the codebase, revision history, splunk logs scoped to /api/active during the window and even mentioned recent changes to the htaccess file just in case).

```
- 17731891050001393532335: 7:32–7:47 PM CDT (15 minutes) - ended up being revised by Hayley earlier from the original Xx placeholder text
- 17731917640003540431010: 8:17–8:22 PM CDT (5 minutes)
```

Claude points the finger at the XHR request to emergency's endpoint. Despite having a max-age of 30, it thought that Acquia varnish might have a minimum TTL floor (e.g. never cache for less than X seconds) and so websites that got a varnish response instead of an apache-request response might have seen stale data for the entire 15 min window that the first alert was up for. the later alert was only imported and open for about 5 min.

We originally had 30 seconds only for DoS protection, but we already get some protection for free from Drupal cache and the JSON API properly invalidates when content is inserted or updated. So, it suggested changing to `no cache` but is unsure if Acquia varnish will respect that either. If not, we might be asking Acquia to change the varnish rule for this endpoint.

I added more logging points just so it would be easier to debug issues in the future.

Claude also suggested a change to the hawkalert.sh script:

>Fixed sleep math from 60 - $(date +%s) + $start to 60 - ($(date +%s) - $start) — the original formula was algebraically wrong and would produce an incorrect sleep duration if the drush command ever took more
  than a second to complete.

`¯\_(ツ)_/¯` 

- Code review?
  - "The no-store response directive indicates that any caches of any kind (private or shared) should not store this response." - https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#no-store
- I am going to start a ticket with Acquia to see if I can get more info on this.
  - https://acquia.my.site.com/s/case/500Pb00001nxIm0IAE/is-there-a-minimum-ttl-for-varnish-dynamic-paths

</details>

# How to test

- Code review of the logging changes.


joewhitsitt:
I've updated the summary above to better reflect the state of the PR.

GaryRidgway (approved):
code REVIEWED

PR #9689, opened 2026-03-13
update robrichards/xmlseclibs to 3.1.5

joewhitsitt:
Builds are going to start failing and this is a [high severity](https://github.com/advisories/GHSA-4v26-v6cg-g6f9) item.

# How to test

required by onelogin/php-saml which is required by drupal/samlauth which is required by uiowa/uiowa_auth... but onelogin/php-saml already allows >=3.1.4 so we can just update it directly to resolve.

to be safe, this could be deployed out to DEV to make sure authentication still works.

## Test in DEV

Using your hawkid and password (not `uli`), confirm that you can login to https://sandbox.dev.drupal.uiowa.edu/ after [successful deployment](https://app.travis-ci.com/github/uiowa/uiowa/builds/277685306) to [uiowa02](https://cloud.acquia.com/a/applications/a912cf23-b421-4071-99a4-2f6f6edb97e3) (task: 354303012).

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -18296,16 +18296,16 @@
         },
         {
             "name": "robrichards/xmlseclibs",
-            "version": "3.1.4",
+            "version": "3.1.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/robrichards/xmlseclibs.git",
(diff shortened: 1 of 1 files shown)


pyrello (approved):
Tested on dev. Able to log in.

Issue #9665, opened 2026-03-06
Keyboard focus indicator is missing - menus


briand44:
# Motivation

Siteimprove is reporting "Keyboard focus indicator is missing". This is a WCAG 2.1 AA compliance issue related to some parent menu items. These menu items seem to be using `<nolink>`.

# Siteimprove info

**Technical doc**
https://alfa.siteimprove.com/rules/sia-r65

Siteimprove link to existing issue
- https://my2.siteimprove.com/Inspector/1223972/A11y/Page?pageId=83425661300&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r65/failed/Rkt6TDBGNDg=;MDUyMDBmYzM=
- https://my2.siteimprove.com/Inspector/1174032/A11y/Page?pageId=53272424901&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r65/failed/bU4zYmpLZUQ=;ZDM5ZjJlYjU=

Example
- https://obgyn.medicine.uiowa.edu/news (Siteimprove browser plugin not reporting)
- https://performingarts.uiowa.edu/about (Siteimprove browser plugin not reporting)

**Keyboard focus indicator is missing**
Issue
2.4.7: Focus Visible
This page element is not visibly highlighted when receiving focus from a keyboard.

The default focus indicator may have been removed in CSS.

**Note that this occurrence is based on a guided review.**

briand44:
Going to close. Feel free to re-open if there was other post prod testing necessary.

briand44:
Siteimprove is reporting this again on a couple sites:

- https://my2.siteimprove.com/Inspector/1009710/A11y/Page?pageId=26016388377&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r65/failed/b25XMEd6eUo=;YTYwYmEwNmE=
- https://my2.siteimprove.com/Inspector/1174032/A11y/Page?pageId=53272424250&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r65/failed/bU4zYmpLZUQ=;YTQwNDIyOWI=

Seems like when we addressed https://github.com/uiowa/uiowa/issues/9725, this became an issue again for top level main menu links using `<nolink>`.

briand44:
## Grooming
 
**Quadrant:** 

Complicated

 **Discussion:**
 - issue is with top menu nav links when using `<nolink>`
[...]

PR #9695, opened 2026-03-19
Menu focus indicator fix - hover background color for focus too

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/9665

Can't test SI locally yet as the browser extension doesn't pick this up, but https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/sia-r65/rule.ts reveals what SI checks for related to this. Ultimately a guess and check right now.

Here is what the menu looks like now for focus. I left the existing bold text for `nolink` items and matched the background color to other hover style definitions within docroot/themes/custom/uids_base/scss/components/menus

Screenshot 2026-03-19 at 11 57 02 AM
Screenshot 2026-03-19 at 11 57 18 AM
Screenshot 2026-03-19 at 11 57 07 AM

# How to test

```
ddev blt frontend && ddev blt ds --site=obgyn.medicine.uiowa.edu
```

Compare https://medicineobgyn.uiowa.ddev.site/ and https://obgyn.medicine.uiowa.edu/ keyboard focus styles for the horizontal menu. Trying to avoid radical design changes but provide a "toggle" to satisfy rule 65.

# Follow-up

Keeping the issue open, review the site improve scans after deployment to see if this resolves the issue.

### DIFF
--- a/docroot/themes/custom/uids_base/scss/components/menus/superfish/superfish.scss
+++ b/docroot/themes/custom/uids_base/scss/components/menus/superfish/superfish.scss
@@ -442,6 +442,7 @@ ul.sf-accordion ul:focus-within {
     text-decoration: underline;
     cursor: pointer;
     outline: none;
+    background: #e1e1e1;
   }
 
   &:hover {


pyrello (approved):
This seems like a good middle ground fix for the near term. Nice work!

PR #9696, opened 2026-03-19
BLT UMD/T - better error handling for file deletion logic

joewhitsitt:
[Marie encountered a failed file cleanup](https://iowaweb.slack.com/archives/C0164FNN1PV/p1773764267402779) during a deprovision (ssh known_hosts change) which made a bigger mess than I originally thought. With the change to the command to keep going and note that manual file deletion would be necessary, it deleted the database in the remotes and cleaned up the local files (aliases) that are used within the command. This prevented re-running the command after fixing the known_hosts. My local didn't have those git changes and after temporarily recreating the database, I was able to successfully run the command including the file deletion process.

This PR restores the exception for failures resulting from the environment lookup as well as a host key change, allowing the option to fix the option, try again later.

# How to test

- Code review unless you want to run a [deprovision](https://github.com/uiowa/uiowa/issues?q=state%3Aopen%20label%3Adeprovision) and trick the code into a failure situation.
- This same `deleteRemoteMultisiteFiles` is used for the `umt` command as well but I don't think that command as a whole has been tested in quite some time (post ACN migration).


blt/src/Blt/Plugin/Commands/MultisiteCommands.php
```diff
@@ -1198,14 +1199,19 @@ protected function deleteRemoteMultisiteFiles(string $id, string $app, string $e
...
+        $environments = new Environments($client);
+        $envs = $environments->getAll($uuid);
+        foreach ($envs as $environment) {
+          if ($environment->name === 'stage') {
+            $env_name = 'stage';
+            break;
+          }
         }
       }
+      catch (\Exception $e) {
```
pyrello, line 1212: Wondering if there is a more specific exception to catch on.

pyrello: I see we're using a generic exception handler above. So, not suggesting this should be a blocker. I think it is generally a best practice to have the exception handler be more specific to the case. But maybe that is just in cases where there are multiple exceptions and they need to be handled differently.

pyrello (approved):
Code is reviewed. Change makes sense.

blt/src/Blt/Plugin/Commands/MultisiteCommands.php
joewhitsitt: yeah, i wasn't sure what all exceptions we could encounter and thought best to just stop overall. as we deal with them, we could refine (if it is still in BLT at that point :) )

PR #9698, opened 2026-03-20
Robo Siteimprove Alfa Audit

joewhitsitt:
# Background

https://iowaweb.slack.com/archives/C0406A9VB5L/p1774023916630219

# How to test

_disclaimer: completely guess and check developed by claude ai_

Attempts to get it working with `ddev robo` have not worked so far. Despite ddev running node 24 or node 20, there still seems to be issues. I like the lightweight `npx` install, but I bet this could be a dev dependency and incorporated differently if we wanted to keep it.

Doesn't seem to return some rules like https://alfa.siteimprove.com/rules/sia-r115. Can't seem to get it to run on authenticated pages either.

`nvm use 20`

Examples:

`vendor/bin/robo si:audit https://home.ddev.site/`
`vendor/bin/robo si:audit https://home.ddev.site/ --rules=sia-r65`
`vendor/bin/robo si:audit https://home.ddev.site/ --all`


joewhitsitt:
If we want to move forward with this, it will need some refactoring as we decided to move away from Robo. https://github.com/uiowa/uiowa/issues/9895#issuecomment-4635421108

PR #9715, opened 2026-03-31
[Commencement] Off-campus venue names in teaser display

### LINKED ISSUE #9701: Output newly imported venue name on event teasers
# Motivation

Following up from https://github.com/uiowa/uiowa/issues/9602, we should be outputting the newly imported venue name field on event teasers if a value exists and there is no referenced venue.

# Additional Info

We currently output the referenced venue and room number on event teasers if they exist but if they don't exist and there is a venue name, that location does not get output on the teaser.

Example - https://commencement.uiowa.edu/ceremonies

Image

In the screenshot above, see that the Sioux City event does not have a location displayed. 

Image

The venue name from the screenshot above should be output in the teaser.

Example - https://commencement.uiowa.edu/ceremony/spring-2026/sioux-city-masters-social-work-graduation-celebration

cory-skeers:
Resolves #9701 

# How to test

```
ddev blt ds --site=commencement.uiowa.edu && ddev drush @commencement.local uli /ceremonies
```
Compare to https://commencement.uiowa.edu/ceremonies, and see that events with off-campus locations are displaying the location name.
Click into an event with an off-campus venue, and see that the full address appears under the venue name


joewhitsitt (approved):
woot!

Image

PR #9717, opened 2026-03-31
Updating OG to stable release.

### LINKED ISSUE #9611: Update orgranic groups to latest D10 compatible version
# Motivation

We would like to be on the latest compatible release so that we get any new features, fixes and are in better position for security releases and Drupal 11.

# Additional Info

**Module**
[drupal/og](https://www.drupal.org/project/og)

**Current Version**
dev-2.x578b4678

**Latest Version (at time of audit)**
dev-2.xb3b2e1b

**Notes**
- see if we can get to a stable release?
- OG finally has a stable release https://www.drupal.org/project/og/releases/2.0.1

GaryRidgway:
# How to test
- Checkout `main`
```
ddev composer install && ddev blt ds --site=signage.sites.uiowa.edu
```
- Checkout `update_organic_groups`
```
ddev composer install && ddev drush @sitessignage.local updb
```
- ensure that when you run a `ddev drush @sitessignage.local config-split: export site` there are no config differences.
- Check a signage group node, ensure that it renders correctly.
- View the group members list, ensure that it renders correctly
- Masquerade as a user that is _not_ a signage group manager.
    - Edit some content, a sign and add a new slide within
    - Ensure that when adding the signage group there are no errors
    - ensure that upon saving the new slide and the sign there are no errors
- Masquerade as a user that _is_ a signage group manager.
    - remove a member from a signage group
    - add a member from a signage group
    - check the removed person _cannot_ access the group's content.
    - check that the added person _can_ access the group's content.
- OG roles have changed pretty significantly, so we want to try to test that.
    - Remain masqueraded as a signage group manager
    - add another user to the group, assign them signage group manager as well
    - masquerade as that user and verify they can access the group and create other users as well.


joewhitsitt (approved):
Okay. So this module had a big change from a homegrown og_complex field for managing group assignment to an entity reference autocomplete field. It looks like the update hook to bulk switch and assign was successful and in spot checking the assignments look correct. The config exports because of this change look correct. I did get some config noise exporting the feature with `config_split.patch.core.entity_view_display.node.sign.default.yml` but thinking that should be in the site split and declaring that in the complete split, i didn't get the same export 🤷 

Went through the test instructions and everything looks good. I did notice the buttons to add sign/slide on a group display for a user with insufficient permissions for the group. they get access denied once clicking, but maybe something to hide in the future.

PR #9727, opened 2026-04-03
Skip cancelled Hawk Alerts

### LINKED ISSUE #9710: Don't import cancelled Hawk Alerts
# Story

As a Campus Safety representative, I would like to be able to cancel Hawk Alerts in Rave without a new alert being created on the Emergency website so people aren’t continuing to receive messages once the threat has subsided.

# Additional Info
 
<details>
<summary>Original report</summary
>I noticed one weird hiccup with the emergency site today during the test. A few minutes after I closed out the Hawk Alert on the website, hit save, checked to make sure it came down, it re-appeared on the site and had to be closed out again.  - Public Safety

> I noticed the CAP outbound feed currently shows a message type of “cancel” at 10:11AM.
> [https://content.getrave.com/cap/uiowa/channel1](https://content.getrave.com/cap/uiowa/channel1)
> 
> From what I can tell, this means the alert was cancelled at 10:11AM within Rave which triggered an update/cancel to the original CAP message sent at 10:00AM. I am guessing this may have caused the alert to publish again on the emergency website. - ITS

>We have started canceling the alert in Rave once the situation is resolved so people aren’t continuing to receive messages once the threat has subsided - Public Safety

>It also deactivates Alertus so people are getting those notifications once they login after there are no longer any active alerts.  Cleary this may be a problem with the website so let us know if need us to re-evaluate this process.  I amt thinking if we just cancel in Rave before we close out, that would fix it based on what you all have said-would that be accurate? - Public Safety
</details>

Campus Safety has started cancelling alerts in Rave so that Alertus notifications aren't being received after there are no longer any active alerts.

The problem with this is that when an alert in Rave is cancelled, it creates a new alert in the CAP endpoint and the Emergency website then creates new alert content.

Image

When an alert is cancelled, the endpoint contains `<msgType>Cancel</msgType>` compared to `<msgType>Alert</msgType>`. I am not aware of other potential types.

Once we have implemented this change we will need to let Campus Safety know, so they can resume their process of cancelling alerts.

cory-skeers:
Resolves #9710 

# How to test

```
ddev blt ds --site=emergency.uiowa.edu && ddev drush @emergency.local cset emergency_core.settings rave_endpoint 'https://content.getrave.com/cap/uiowatest' -y &&  ddev drush @emergency.local emergency_core:alerts_import
```
With the current status at the test endpoint, you should see the alert get skipped as it is a "Cancel" type.

In addition to code review, using `xdebug` you can locally alter the retrieved info and change the `msgType` to "Alert" and observe the alert get imported.

### DIFF
--- a/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/AlertsProcessor.php
+++ b/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/AlertsProcessor.php
@@ -98,7 +98,7 @@ class AlertsProcessor extends EntityProcessorBase {
     $this->processedRecords[] = $recordSyncKey;
 
     // Alert cancellations should be skipped so they don't
-    // double the existing alert.
+    // create duplicate alerts.
     if ($record['msgType'] === 'Cancel') {
       $this->skipped++;
       static::getLogger('emergency_core')->notice('Alert @identifier skipped (msgType "Cancel").', [
--- a/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/AlertsProcessor.php
+++ b/docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/AlertsProcessor.php
@@ -97,6 +97,16 @@ class AlertsProcessor extends EntityProcessorBase {
     $recordSyncKey = $record[$this->apiRecordSyncKey];
     $this->processedRecords[] = $recordSyncKey;
 
+    // Alert cancellations should be skipped so they don't
+    // double the existing alert.
+    if ($record['msgType'] === 'Cancel') {
+      $this->skipped++;
+      static::getLogger('emergency_core')->notice('Alert @identifier skipped (msgType "Cancel").', [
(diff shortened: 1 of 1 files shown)


docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/AlertsProcessor.php
```diff
@@ -97,6 +97,16 @@ public function process() {
     $recordSyncKey = $record[$this->apiRecordSyncKey];
     $this->processedRecords[] = $recordSyncKey;
 
+    // Alert cancellations should be skipped so they don't
+    // double the existing alert.
```
joewhitsitt:
```suggestion
    // create duplicate alerts.
```

joewhitsitt (approved):
Logic is clean, output was as expected.

Worth considering tests in the future. Cancel with no existing node (skipped, no node created), Cancel with existing node (skipped, existing node unchanged), Alert (node created). 👀

PR #9728, opened 2026-04-03
[Commencement] Bump up the items per page

### LINKED ISSUE #9694: Commencement - allow more than 100 events to be imported
> Good morning! Not sure if I did something wrong or tagged something incorrectly but I'm having trouble finding a ceremony from the events hub in the commencement site: https://events.uiowa.edu/event/36909
> 
> Looks like I created it on 03/09 with a bunch of others that did come over ok. 
> 
> I've since tried making some edits in the event hub hoping to force over an update but I don't see anything come over on the commencement side.
> 
> Usually when I make a change in the events hub, I'll see the event content type have an update status change but doesn't seem to be happening. I would expect to see an update to the engineering ceremony on the 11th or 12th from the change I made in the events hub on the 11th.
> 
> Hope you can take a look and see what's going on. 

No other events seem to be missing. Missing event is not noticeably different from the same event in a different semester, or even other events also in Fall 2029. I noticed the date is December 16, while all other events are a day or two before that, but I'm not sure how a one day difference could affect the importer - Marie

Update from Josh:

> Yes, as far as I can tell, this is the only event that hasn't been pulled into the commencement site. And yes, it should be like the others, I think all of the fall 2029 events were cloned from fall 2028 events in the events portal and all those came over fine.

cory-skeers:
Resolves #9694 

# How to test

On `main`

```
ddev blt ds --site=commencement.uiowa.edu && ddev drush @commencement.local import:events && ddev drush @commencement.local uli /admin/content
```
See that there are probably no new events created. Check out the admin content page, see that there are no unpublished events.

```
git checkout commencement-events-pager-bump && git pull && ddev drush @commencement.local import:events
```
See that there is at least one new event created. Check out the admin content page, and see that there is at least one new unpublished event

### DIFF
--- a/config/sites/commencement.uiowa.edu/views.view.program_archive.yml
+++ b/config/sites/commencement.uiowa.edu/views.view.program_archive.yml
@@ -367,8 +367,8 @@ display:
             default_group_multiple: {  }
             group_items: {  }
           reduce_duplicates: false
-          handler: 'default:media'
           widget: autocomplete
+          handler: 'default:media'
           handler_settings:
             target_bundles:
               file: file
--- a/docroot/sites/commencement.uiowa.edu/modules/commencement_core/src/EventsProcessor.php
+++ b/docroot/sites/commencement.uiowa.edu/modules/commencement_core/src/EventsProcessor.php
@@ -76,7 +76,7 @@ class EventsProcessor extends EntityProcessorBase {
             'department' => 329,
             'type' => 28,
           ],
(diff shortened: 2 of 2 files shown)


joewhitsitt:
pulling to test quick

joewhitsitt (approved):
After digging into this earlier, I can confirm that a new event is created as unpublished and that it is `36909` which was missing.

PR #9730, opened 2026-04-03
Limit menu bg color change to sf-depth-2+

### LINKED ISSUE #9725: Scope keyboard menu focus fix
# Problem
Follow up from https://github.com/uiowa/uiowa/issues/9665, the solution is not adequetly scoped and is causing unintentional background color for click/focus for top-level menu items, mega menu items.

#9665 represented a short term fix for the siteimprove accessibility issue as we continue to develop menu styling in UIDS.

# Steps to reproduce
Click on top-level or mega menu items.

Image
Image

# Proposed solution
- Scope the focus rule to just the level of navigation that siteimprove is complaining about.
- Continue to refine menu styles in UIDS for consumption in uids_base.

cory-skeers:
Resolves #9725 

# How to test

```
ddev blt frontend && ddev blt ds --site=docs.uiowa.edu
```
Visit https://docs.uiowa.ddev.site/ and tab into the main menu
See that the top level menu items do not get a background change, but submenu items (2nd as well as 3rd level, like under Miscellaneous) do
Condense the viewport size until the menu collapses. Tab through focus, and see that only the ordinary text styling edits are in place
Change the menu to the toggle version (in /admin/appearance/settings/uids_base) and tab through--see that only the ordinary text styling edits are in place.


joewhitsitt:
pulling to test

joewhitsitt (approved):
Thanks for fixing my oops. Looks better and still covers the original issue.

Also tested against the original issue https://github.com/uiowa/uiowa/pull/9695 (nolink).

I was trying to recall why the hardcoded hexcode was used over something like variables.$light and I think it was the point in which siteimprove scans and can't pick up the value. something that was revealed with https://github.com/uiowa/uiowa/pull/9698 that wasn't the case for the browser extension. I think it returned a `cantTell` instead of `passed`

Issue #9636, opened 2026-02-26
Usage table restricted access bug


GaryRidgway:
# Problem
When viewing certain documents in the usage table, you cannot see where they are referenced, and instead the title of the reference is `- Restricted access -` ([Example here](https://law.uiowa.edu/media/18726/edit/usage))

<details><summary>Extra context from user</summary>
<p>

A little extra info in case it ends up being useful, as I'm pretty sure this is a page-level thing impacting a few pages on our site:
I've had documents where some of the Usage information is normal, and other line items are "restricted"
The original screenshot came from a situation where a document was deleted (incorrectly assumed that "restricted access" note came from deleted/inactive webpages and deleting wouldn't cause an issue), finding a resulting broken link in SiteImprove, and then uploading a replacement doc in the broken link's place on the [identified page](https://law.uiowa.edu/consumer-information-aba-required-disclosures). The replacement doc's usage then showed a single "restricted" line from the page it was on.

And if not useful... sorry for making you read all of that! Thanks again for your time.

</p>
</details> 

<img width="1864" height="457" alt="Image" src="https://github.com/user-attachments/assets/2384906f-d0cb-4011-8e47-46a12cc82edb" />

# Steps to reproduce
Go to the above example, log in, be confused because you are an admin and should be able to see everything.

# Proposed solution
Check how this is being output and trace back to see what is happening?

PR #9755, opened 2026-04-13
upgrade entity_usage to beta28

### LINKED ISSUE #6509: Entity usage "published" status not accurate
# Story

As a site editor, I would like an accurate indication of page publication status on the media entity usage pages, so that I can use it to help moderate and clean up my content.

# Background

Brought up by Viki Mueller, when looking at entity usage pages, some pages are listed as "published" even if they are not currently published.

On initial investigation, it is accurate for some things like featured image fields, but not others. It _might_ only be an issue with media used in layout builder blocks.

Her use case: They'd like to remove media that is no longer used, and would like to include media used only in archived pages in that (eg remove media if it is only used on unpublished pages)

(Note, same page is listed as both published and unpublished, depending on the usage)
Image

# Proposed solution

Investigate for any upstream issues?
Homebrew fix to update column to use the parent (if present) entity status rather than direct reference?
Remove the column for now until upstream/homebrew fix is in place, to avoid confusion?

### LINKED ISSUE #9722: Update entity usage to latest D10 compatible release
# Motivation

We would like to be on the latest compatible release so that we get any new features, fixes and are in better position for security releases and Drupal 11.

# Additional Info

**Module**
[drupal/entity_usage](https://git.drupalcode.org/project/entity_usage/-/compare/8.x-2.0-beta16...8.x-2.0-beta25)

**Current Version**
2.0.0-beta16

**Latest Version (at time of audit)**
2.0.0-beta25

**Notes**
- Lots of extra checks in places which seem straightforward, but some major functional changes for some of the plugins like LayoutBuilder

joewhitsitt:
Changelog: https://git.drupalcode.org/project/entity_usage/-/compare/8.x-2.0-beta16...8.x-2.0-beta28?from_project_id=22086

Resolves https://github.com/uiowa/uiowa/issues/9722
Resolves https://github.com/uiowa/uiowa/issues/6509

# How to test

## Existing site

`ddev blt ds --site=sandbox.uiowa.edu`

No config import errors, database updates run successfully.
Re-save admin/config/entity-usage/settings and `drush cex`. Should be identical.
Inspecting the usage table for media or nodes should be more illustrative. Instead of just saying `Content`, it includes the bundle (e.g. Content: Article). Publish status is now consistent across entities and points at the parent entity not the media entity state and updates if the parent entity is unpublished, even if the usage is tied to a previous revision.

Same issue of --Restricted Access-- still occurs when the media entity is attached to a block that is no longer part of the node. Until the page is deleted, that block entity still exists and won't let go of the media usage. https://github.com/uiowa/uiowa/issues/9636

## New install usage

Tests concerns around https://github.com/uiowa/uiowa/issues/9722#issuecomment-4194616762

`ddev blt di --site=default && ddev drush @default.local uli node/4/usage`
See that Demo article is listed in the usage table.
Go to https://default.uiowa.ddev.site/media/8/edit/usage and see two rows of usage.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5780,38 +5780,39 @@
         },
         {
             "name": "drupal/entity_usage",
-            "version": "2.0.0-beta16",
+            "version": "2.0.0-beta28",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/entity_usage.git",
-                "reference": "8.x-2.0-beta16"
+                "reference": "8.x-2.0-beta28"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://ftp.drupal.org/files/projects/entity_usage-8.x-2.0-beta16.zip",
-                "reference": "8.x-2.0-beta16",
-                "shasum": "af6533149a0926d2d539f279e7e08073f8f08c55"
+                "url": "https://ftp.drupal.org/files/projects/entity_usage-8.x-2.0-beta28.zip",
+                "reference": "8.x-2.0-beta28",
+                "shasum": "7e97367d1c583a41dfab91226e4f688309ae2ce7"
             },
             "require": {
-                "drupal/core": "^10.2 || ^11"
+                "drupal/core": "^10.3 || ^11"
             },
             "require-dev": {
                 "drupal/block_field": "~1.0",
-                "drupal/ckeditor": "^1.0",
-                "drupal/dynamic_entity_reference": "~1.0 || ^2.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


cory-skeers (approved):
Reviewed and tested, and things look good. I really like the updates to the usage view!

I experimented with various deleting and updating, and the usage seemed to keep up and accurate. I also checked it with a quick p2lb page conversion, and the block usage update still looked to work as expected.

joewhitsitt:
Good call on testing p2lb! Thanks @cory-skeers

Issue #9693, opened 2026-03-18
Revisit Hawk Alerts to address recent concerns


joewhitsitt:
Follow-up from https://github.com/uiowa/uiowa/issues/1899

# Problem

Issue #1899 revealed concerns about how the hawk alert functionality triggers, where it is placed and how screen readers are notified.

# Proposed Solution

- Consult with TM and campus screen reader users on the use of `role=alert` - Draft email here: https://github.com/uiowa/uiowa/issues/1899#issuecomment-4076706235
- Consider whether alerts should be merged in with Drupal messages
- Consider changes to how alerts are initially fetched, displayed, updated, removed
- Revisit how they are displayed, especially if multiple while maintaining compliance

joewhitsitt:
T.M. had a couple initial thoughts today but wants to ask some users on campus and contacts elsewhere in higher ed and will be pointing them at admissions.stage for reference:

On load seems to be the main concern and thinks that role=alert for items that come in after the page is loaded should probably stay the same. Acknowledged the definition seems for more systematic, but is generally what is expected in practice.

Couple ideas that were presented on the spot:

- Slow the delivery a bit on load. Most screen reader users have the speed dialed up anyway so if the alerts (if multiple) would render to the DOM one at a time there would be less chance for interruption. Simplifying the alert that is announced would shorten things up.
- Consider role=status or similar for alerts on load, but have a single role=alert announcement that says something like "There are X number of alerts to be announced." Which is then followed by polite announcements.
- Refactor (out of drupal messages) into a separate aria-live region aria-atomic=true or wrapper div so that it is still assertive but reads out all of them in full starting with a count. Announce removals as well.
- Noted the duplicative "alert, Hawk Alert..." but understands the branding piece.

I told T.M. that we have "fixed" the SiteImprove compliance issue (so no immediate rush) but we were interested in getting the UX right while still meeting Public Safety's expectations.

joewhitsitt:
We are having to touch the "how should campus alerts be announced" piece of this as we replace a views block with a dynamic js block on the emergency.uiowa.edu website in https://github.com/uiowa/uiowa/pull/9767

The "no immediate rush" on this might have to be adjusted a bit.

GaryRidgway:
In trying to slim down some code written for https://github.com/uiowa/uiowa/pull/9767, I was chatting with Claude about trying to remove two separate systems of alerting. This was its response and it was _quite convincing._

<details><summary>Claude's argument:</summary>
<p>

[...]

PR #9767, opened 2026-04-17
Emergency refresh block

### LINKED ISSUE #9732: Page refreshes or redirects without warning (within 20 hours) - emergency.uiowa.edu
# Motivation

Siteimprove is reporting "Page refreshes or redirects without warning (within 20 hours)". This is a WCAG 2.2? A compliance issue related to the Emergency website.

# Siteimprove info

**Technical doc**
https://alfa.siteimprove.com/rules/sia-r9

Siteimprove link to existing issue
- https://my2.siteimprove.com/Inspector/940344/A11y/Page?pageId=47302581264&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r9/failed/QkVXdUg3NjY=;NjgzMjU4YjQ=

Example
- https://emergency.uiowa.edu/

**Page refreshes or redirects without warning (within 20 hours)**
Issue
2.2.1: Timing Adjustable
2.2.4: Interruptions
3.2.5: Change on Request
A <meta> element is being used to refresh or redirect visitors away from the page within 20 hours after content has loaded.

GaryRidgway:
# How to test
## Setup
1. Copy and paste the following command and follow the instructions in your terminal.
    i. This command will set the local variables for the hawk alert api for both emergency and sandbox to be your local emergency site.

```bash
git pull && ddev composer install && ddev blt ds --site=emergency.uiowa.edu && ddev blt ds --site=sandbox.uiowa.edu && ddev drush @emergency.local config:set uiowa_alerts.settings hawk_alert.source https://emergency.uiowa.ddev.site/api/active -y && ddev drush @sandbox.local config:set uiowa_alerts.settings hawk_alert.source https://emergency.uiowa.ddev.site/api/active -y && ddev drush @emergency.local cr && ddev drush @sandbox.local cr && echo "Click both of the login links below..." && printf '%*s\n' "${COLUMNS:-80}" '' | tr ' ' '─' && echo "vv Emergency site vv" && ddev drush @emergency.local uli && printf '%*s\n' "${COLUMNS:-80}" '' | tr ' ' '─' && echo "vv Sandbox site vv" && ddev drush @sandbox.local uli && printf '%*s\n' "${COLUMNS:-80}" '' | tr ' ' '─' && echo "Then click this link and remove the refresh value from SEO settings -> advanced -> refresh." && echo https://emergency.uiowa.ddev.site/node/1/edit && printf '%*s\n' "${COLUMNS:-80}" '' | tr ' ' '─' && echo "Then click this link to make a hawk alert." && echo https://emergency.uiowa.ddev.site/node/add/hawk_alert
```

4. remove the old `Hawk Alerts - Active` block from the homepage
5. add a new `Active Alerts Block` to the homepage.

## hawk alerts
_It might be useful to open two tabs side by side on your monitor to see the changes in real time without tab swapping_
1. _You can set the polling time lower for your convenience in the following files for emergency and sandbox respecvitely:_  
     1. `docroot/sites/emergency.uiowa.edu/modules/emergency_core/js/active-alerts.js:8`  
     2. `docroot/modules/custom/uiowa_alerts/es6/alerts.es6:13`  
4. Make an example hawk alert on your local site. Create a few updates on it with at least one picture or wacky thing in the wysiwyg. Ensure it is marked it as open, and publish it.
7. You should see that it now appears on the homepage of both:
    1. https://emergency.uiowa.ddev.site/
    2. https://sandbox.uiowa.ddev.site/
8. Change the hawk alert updates and observe that, **_without reloading the page_**, the updates get added after the pre-set time (set in code) has elapsed.
9. Note changes on sandbox too, and that they update without reloading the page.

## Siteimprove
11. Run a SiteImprove extension check and note that the "Page refreshes or redirects without warning (within 20 hours)" warning is gone.

## Voice over
13. Turn on Voice over for mac _(The mute button is your best friend when you are swapping contexts so you don't have to have VO yelling at you all the time.)_
14. On emergency  
    15. Note that hawk alerts only announce when added or updated, not _every_ time the endpoint is polled.
    16. Note that when an update is made, it announces it as an update to a hawk alert
    17. Note that when all hawk alerts are gone, there is an announcement that the campus status is back to normal.
18. On sandbox 
    15. Note that hawk alerts only announce when added, not _every_ time the endpoint is polled.
    16. Note that when an update is made, it does _not_ announce it.

## Bonus points
19. Try and break things.


docroot/sites/emergency.uiowa.edu/modules/emergency_core/emergency_core.routing.yml
```diff
@@ -0,0 +1,7 @@
+emergency_core.active_alerts:
+  path: '/api/active'
```
joewhitsitt: How does this not affect the current endpoint that lives there that all SiteNow websites and others consume?

https://emergency.uiowa.edu/api/active

docroot/sites/facilities.uiowa.edu/modules/facilities_core/js/utility-alerts.js
```diff
@@ -3,6 +3,7 @@
     attach: function (context) {
       once('utility-alerts', '.utility-alerts-container', context).forEach(function (container) {
         var url = drupalSettings.facilities_core.utilityAlertsUrl;
+        console.log(drupalSettings.facilities_core.utilityAlertsUrl);
```
joewhitsitt: is this needed?

GaryRidgway: oh definitely not, good catch.

joewhitsitt:
pulling to test

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
```diff
@@ -0,0 +1,186 @@
...
+      hour12: true,
+    });
+    let day = '';
+    const month = monthFormatter.format(date);
+    const time = timeFormatter
+      .format(date)
+      .replace('AM', 'a.m.')
+      .replace('PM', 'p.m.');
+
+    const is_emergency = window.location.host.startsWith('emergency');
```
joewhitsitt: Why did this site-specific code end up in this shared abstracted file? Shouldn't emergency specific js be in active-alerts? Was it just a late discovery?

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
```diff
@@ -0,0 +1,186 @@
...
+    let title = 'Hawk Alert';
+    let hawk_alert_body = '';
```
joewhitsitt: Aren't we trying to use camelCase for JS? `is_emergency` `hawk_alert_body`, `emergency_link`?

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
```diff
@@ -0,0 +1,186 @@
...
+    return `<p class="hawk-alert-updates-title"><small><strong>Situation update(s):</strong></small></p>`;
+  }
+
+  static async getSituationUpdates(item) {
+    if (item?.relationships?.field_hawk_alert_situation?.links?.related?.href !== undefined) {
+      const updates_url = item.relationships.field_hawk_alert_situation.links.related.href;
+      return await AlertsUtilities.fetchAlerts(updates_url);
+    }
+    else {
+      return '';
```
joewhitsitt: Should we try to return null here instead and clean up the `!response` checks?

docroot/sites/emergency.uiowa.edu/modules/emergency_core/emergency_core.libraries.yml
```diff
@@ -2,3 +2,13 @@ emergency-hawk-alerts:
   css:
     component:
       css/emergency-hawk-alerts.css: {}
+active_alerts:
+  js:
+    js/active-alerts.js: {}
+#  css:
+#    theme:
```
joewhitsitt: commented out css block. is it needed?

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/Plugin/Block/ActiveAlertsBlock.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+    parent::blockSubmit($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $config = $this->getConfiguration();
+    $source = \Drupal::config('uiowa_alerts.settings')->get('hawk_alert.source');
```
joewhitsitt: We should use dependency injection similar to the uiowa_alerts block

`$this->config->get('uiowa_alerts.settings');`

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/Plugin/Block/ActiveAlertsBlock.php
```diff
@@ -0,0 +1,105 @@
...
+
+    if (empty($config['headline'])) {
+      $child_heading_size = $config['child_heading_size'];
```
joewhitsitt: It doesn't look like child_heading_size is being used

joewhitsitt: or headingSize for that matter (hardcoded h2)

joewhitsitt:
I appreciate the attempt to give an "all the things" one liner, but I get the following when copy/pasta/enter:

`cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand cmdand dquote bquote> `

I'll step through and run these separately

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
```diff
@@ -0,0 +1,186 @@
+class AlertsUtilities {
```
joewhitsitt: New territory here with it being abstracted and shared, but should this still be wrapped even though it is intended to only be called as a dependency? Do we want this out a global window thing or contained/controlled a bit?
```
  (function () {
    class AlertsUtilities {                                                                                                                                                                                                      
      ...                    
    }
```

joewhitsitt:
Added drupal.announce to active-alerts because it was not firing. https://github.com/uiowa/uiowa/pull/9767/commits/ecd46a53c0a68df726babc162cad16f9be7c7948

Screenshot 2026-04-29 at 3 02 32 PM

joewhitsitt:
Can we tighten this up a bit?

Screenshot 2026-04-29 at 3 04 51 PM

joewhitsitt:
This might have an intersection with what we decide to do here https://github.com/uiowa/uiowa/issues/9693, but with VoiceOver I am not hearing the announcements. The messages are within a role=region but might need to be a live region or the messages themselves marked up as role=alert or role=status (depending on the issue above).

On load or fetch, I am not hearing `Active alerts have been loaded.`

@GaryRidgway did you test and hear this during development? I wonder if it being the same message (instead of reading the hawk alert) affects whether VoiceOver announces it.

Either way, it is announcing every fetch, not whether or not something is added or removed and that should probably change.

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/Plugin/Block/ActiveAlertsBlock.php
joewhitsitt: Looks like an empty heading possibility at that

Screenshot 2026-04-29 at 3 37 26 PM

Do we need the headline field on this?

GaryRidgway:
Thanks for the review @joewhitsitt! I will get on these ASAP.

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
GaryRidgway: This is generally the correct call, but since this is a class it has to be able to be accessed from somewhere else. I've instead added it to the drupal namespace which kind of cuts the middle.

GaryRidgway:
> Can we tighten this up a bit?
> 
> Screenshot 2026-04-29 at 3 04 51 PM

This was actually getting the _views_ styles which was making it smaller on emergency. I would have flagged that old style as a bug, but I'm not opposed to changing it.

GaryRidgway:
> This might have an intersection with what we decide to do here #9693, but with VoiceOver I am not hearing the announcements. The messages are within a role=region but might need to be a live region or the messages themselves marked up as role=alert or role=status (depending on the issue above).
> 
> On load or fetch, I am not hearing `Active alerts have been loaded.`
> 
> @GaryRidgway did you test and hear this during development? I wonder if it being the same message (instead of reading the hawk alert) affects whether VoiceOver announces it.
> 
> Either way, it is announcing every fetch, not whether or not something is added or removed and that should probably change.

Because of the way I changed it so that it didn't re-render all the time, I goofed on how it did the drupal announces. I will fix that.

joewhitsitt:
Giving this all another look

docroot/sites/emergency.uiowa.edu/modules/emergency_core/js/activeAlerts.js
```diff
@@ -0,0 +1,237 @@
...
+        // Cached state from the previous successful fetch — used to detect
+        // real content changes so screen readers only hear about diffs.
+        // Starts empty so first-load alerts announce as "new"; first load
+        // with no alerts stays silent (empty diff).
+        let prevAlerts = new Map();
+        let lastError = false;
+        updateActiveAlerts();
+
+        // Check for changes every 55 seconds.
+        setInterval(() => updateActiveAlerts(), 55000);
```
joewhitsitt: I might have promised that we could check more frequently with this setup now that we aren't reloading the whole page each time. 30 seconds?

GaryRidgway: works for me, no qualms! about twice as many calls to the api though, is that okay?

joewhitsitt: Every Sitenow webpage/sign is hitting the endpoint every 60 seconds via uiowa_alerts. changing one website from 55 to 30 is a drop in the bucket and will increase the chances of the hawk alert being shown before people click the link from a website/sms/app.

GaryRidgway: Donezo

joewhitsitt:
@GaryRidgway, revisiting the announce stuff is the only hiccup I see. 

In https://github.com/uiowa/uiowa/pull/9684 we were getting double announcements so we set `announce: ''` knowing that by having the `role=alert` markup, it would be announced on load and as they are inserted until we could figure out https://github.com/uiowa/uiowa/issues/9693

For one, I don't think these are announced on load for uiowa_alerts or emergency and when they are announced the duplication occurs.

I've temporarily bumped up the interval further to try to listen as this changes think about how to announce this without affecting the new enhancements/sharing

docroot/sites/emergency.uiowa.edu/modules/emergency_core/src/Plugin/Block/ActiveAlertsBlock.php
joewhitsitt: The empty h2 issue is resolved, but the unused child_heading_size is still in here. Not a blocker, but not sure if it is necessary if we don't plan on adding a headline to this block.

joewhitsitt:
It sounded like Alan might test the recent additions tomorrow. From my perspective, I feel like this PR is in a decent spot given the attempt to share functionality with uiowa_alerts and compromise on not restructuring completely until we know more from https://github.com/uiowa/uiowa/issues/9693

This PR resolves https://github.com/uiowa/uiowa/issues/9732 without introducing regressions and layers in some nice improvements/additions.

joewhitsitt:
@GaryRidgway expressed interest in getting another dev to review the current state. I'll say the recent changes make sense to me and the PR as whole look good to me, but un-assigning for now.

pyrello:
> @GaryRidgway expressed interest in getting another dev to review the current state. I'll say the recent changes make sense to me and the PR as whole look good to me, but un-assigning for now.

I'll be the extra set of dev eyes on this. Pulling to test.

pyrello:
I had Claude do a review of the PR and it surfaced a concern about the block of JS that catches errors and displays an error message:
https://github.com/uiowa/uiowa/blob/76a89c044a013a14769e02c20b5c4502f447f325/docroot/sites/emergency.uiowa.edu/modules/emergency_core/js/activeAlerts.js#L49-L55

Specifically, if an error is caught, the message shown to the user is replaced by an error message. If this happens in response to a temporary network unavailability, this means that the Hawk Alert is not displayed even though it might still be valid. The suggestion is to leave the existing message in place until the connection is re-established. 

I thought this is worth bringing up although I'm not close enough to the work to say whether this makes sense in the current work or as a follow-up issue.

GaryRidgway:
> I had Claude do a review of the PR and it surfaced a concern about the block of JS that catches errors and displays an error message:
> 
> https://github.com/uiowa/uiowa/blob/76a89c044a013a14769e02c20b5c4502f447f325/docroot/sites/emergency.uiowa.edu/modules/emergency_core/js/activeAlerts.js#L49-L55
> 
> Specifically, if an error is caught, the message shown to the user is replaced by an error message. If this happens in response to a temporary network unavailability, this means that the Hawk Alert is not displayed even though it might still be valid. The suggestion is to leave the existing message in place until the connection is re-established.
> 
> I thought this is worth bringing up although I'm not close enough to the work to say whether this makes sense in the current work or as a follow-up issue.

Ah, great catch. So how we want this logic to work is:

- if we already have hawk alerts or the campus normal loaded
- do not output the "Unable to load hawk alerts"
- keep trying to ping the emergency site until we get a response
- see if we need to update the hawk alerts and do so if needed.

I can make that change if we would like

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
```diff
@@ -0,0 +1,225 @@
+// Shared alerts utilities.
+//
+// Provides per-instance polling state plus static helpers for fetching
+// the alerts feed, composing Hawk Alert markup, and formatting timestamps.
+((Drupal) => {
```
pyrello, line 5: A couple thoughts:
- I don't understand the choice to introduce a new `.es6` file. My understanding is that ES6 can be run directly in JS and doesn't need to be compiled.
- I know we're all over the place with JS file names, but this is even inconsistent with the other file being created in this PR (`activeAlerts.js`) as far as capitalization.

GaryRidgway: both of these 100%.

I just added the utilities file here because that made the most structural sense, and additionally I would love to remove all es6 from our codebase if I could, but i didn't want to scope creep more than I already have.

The file name is something that should be changed to match our camel case JS standards. agreed.

pyrello: No need to scope creep. You should be able to just add the new file as `.js`

pyrello:
> Ah, great catch. So how we want this logic to work is:
> 
>     * if we already have hawk alerts or the campus normal loaded
>     * do not output the "Unable to load hawk alerts"
>     * keep trying to ping the emergency site until we get a response
>     * see if we need to update the hawk alerts and do so if needed.
> 
> I can make that change if we would like

Claude suggested showing an error after N number of tries. That seems like overkill to me. I think erring on the side of showing an alert for too long is better than prematurely taking one down.

pyrello:
Just noting that I'm not a fan of the very long copy/paste block. I like copy/paste blocks as a time-saver for the reviewer and I encourage that. The problem I have with this is that its hard to see everything that is going to happen before you copy/paste it. Once I pasted it, I was surprised by how long it was. If something goes wrong during the process, it makes it hard to pick it up partway through.

My preference would be for several shorter blocks that get you to the same place. Especially with things like `git pull` and `ddev composer`, which only usually need to be run once or occasionally.

The instructions don't include running `ddev blt frontend` even though that is necessary for the new JS to get generated and picked up.

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.js
```diff
@@ -0,0 +1,225 @@
...
+
+      let title = customConfig?.title ?? 'Hawk Alert';
+
+      let hawkAlertBody =
+        customConfig?.body ??
+        `<span class="hawk-alert-body">${item.attributes.alert}</span>`;
+
+      let emergencyLink =
+        customConfig?.link ??
+        `<a class="hawk-alert-link alert-link" href="${item.attributes.more_info_link}">Visit ${item.attributes.more_info_link} for more information.</a></p>\n`;
```
pyrello: Seems to have an extra closing `</p>` tag.


```suggestion
        `<a class="hawk-alert-link alert-link" href="${item.attributes.more_info_link}">Visit ${item.attributes.more_info_link} for more information.</a>\n`;
```

docroot/modules/custom/uiowa_alerts/es6/AlertsUtilities.es6
GaryRidgway: Ope, just looked up some JS best practices and I remembered why I named it such. It is actually convention to capitalize the first letter of a javascript file that contains a class, which this is. I've changed it to a `.js` file still though.

GaryRidgway:
@pyrello Your qualms have been remedied!

GaryRidgway:
> Just noting that I'm not a fan of the very long copy/paste block. I like copy/paste blocks as a time-saver for the reviewer and I encourage that. The problem I have with this is that its hard to see everything that is going to happen before you copy/paste it. Once I pasted it, I was surprised by how long it was. If something goes wrong during the process, it makes it hard to pick it up partway through.
> 
> My preference would be for several shorter blocks that get you to the same place. Especially with things like `git pull` and `ddev composer`, which only usually need to be run once or occasionally.
> 
> The instructions don't include running `ddev blt frontend` even though that is necessary for the new JS to get generated and picked up.

Totally fair. Would you like me to split this one into chunks to make it clearer? or is this more so feedback for next time I create something? if it would help your testing, I'd be glad to split it (_marking tone, I am sincere about this question. Not trying to be rude._)

pyrello:
> 4. Change the hawk alert updates and observe that, without reloading the page, the updates get added after the pre-set time (set in code) has elapsed.
> 5. Note changes on sandbox too, and that they update without reloading the page.

I'm seeing the emergency site reflect situation updates when refreshed, but not Sandbox. Is that expected?

pyrello:
> Totally fair. Would you like me to split this one into chunks to make it clearer? or is this more so feedback for next time I create something? if it would help your testing, I'd be glad to split it (_marking tone, I am sincere about this question. Not trying to be rude._)

Just meant as feedback for future PRs. Thanks!

pyrello:
Switching `AlertUtilities` to use the `.js` extension partially addressed my concern about using es6 vs JS. However, that file still gets processed and built into a separate JS file. That is because the file is still located in `docroot/uiowa_alerts/es6` directory, which is the target of the `es6-compile` command.

I think it would be reasonable to switch back to using `.es6` to avoid establishing a new pattern for where JS files are located in this module as part of this unrelated work and to push that to a follow-up. I think there is already an open issue related to es6 => JS conversion.

Also noting that I requested the name of the file be changed to match the camelCase name of the other JS file being introduced in the same PR for at least that much consistency.

GaryRidgway:
> > 4. Change the hawk alert updates and observe that, without reloading the page, the updates get added after the pre-set time (set in code) has elapsed.
> > 5. Note changes on sandbox too, and that they update without reloading the page.
> 
> I'm seeing the emergency site reflect situation updates when refreshed, but not Sandbox. Is that expected?

It is expected, yes.

pyrello:
> Also noting that I requested the name of the file be changed to match the camelCase name of the other JS file being introduced in the same PR for at least that much consistency.

Withdrawing this request. Discussed with @GaryRidgway in Slack and this naming is actually consistent with current Drupal coding standards. Drupal essentially [outsourced their JS coding standards to Airbnb in D8.4](https://www.drupal.org/project/drupal/issues/2815077) and Airbnb's rules [23.6](https://github.com/airbnb/javascript#naming--filename-matches-export) + [23.8](https://github.com/airbnb/javascript#naming--PascalCase-singleton) indicate that Pascal casing is appropriate for files that export a Class as their default export.

We also discussed that it would be good to have a shortened guide on common coding style guide questions so we don't have this confusion in the future.

pyrello (approved):
Tested + reviewed. Looks good.

PR #9768, opened 2026-04-17
dependabot - group npm, symfony

joewhitsitt:
Went to a session at Drupalcon that talked about some project management type optimizations. They used Renovate to group dependencies, and I thought maybe dependabot could at least group our npm type things. Try symfony too, but those will typically end up in a core upgrade.

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups

# How to test

Figured we could merge and try and if we don't like it we can revert some or all of it.


pyrello (approved):
Reviewed, not tested.

Issue #9759, opened 2026-04-14
Identify inactive sites


briand44:
# Motivation

We have a policy related to inactive sites (https://sitenow.uiowa.edu/policies#inactive). We should be enforcing that policy in order to minimize overhead, maintenance and platform bloat. In order to enforce this policy, we need to be able to identify sites that have not had a user log in (excluding admins) over a specified time period.

Michael is wanting the number of inactive sites over the last 6 months and 1 year by April 21ish.

# Additional Info

- Identify SiteNow sites that have not had a user log in (excluding admins/user 1?) over a specified time period
  - Will need to report over last 6 months and last 1 year.
  - If we could also provide list of sites that have not had content updates over the specified time period, that would be a bonus.
- Would be great to be able to get/include additional information about the sites 
  - Application, url, SiteNow version, last login date, webmaster emails
- Previously discussed not using BLT but possibly Robo and/or [Drall](https://github.com/jigarius/drall) to run against all sites
- Previously used `drush user:list`
  - https://github.com/uiowa/uiowa/issues/3153#issuecomment-853280500
  - Supported by Rick P. - https://github.com/richardbporter/drush-users-commands
- Could https://github.com/uiowa/uiowa/blob/main/blt/src/Blt/Plugin/Commands/ReportCommands.php be used to get Michael the info quickly?
- Future slices to attempt communications with the webmasters of the sites and to deactivate/backup/remove sites no longer needed or that we don't get any response from.

<details>
<summary> Previous grooming conversation from #3455 </summary>

 **Discussion:**
 - Could be Drush command but Drush commands run on single sites
 - Possibly use Robo or other tool in order to run against all sites
 - previous Drush command supported by Rick
 - `blt/src/Blt/Plugin/Commands/ReportCommands.php`
   - gets lists of applications
   - runs commands
   - outputs csv
 - would be preferred to avoid BLT
 - conversation around starting fresh vs adapting/porting
 - shouldn't need D7 commands anymore
 - questions around how to measure inactivity
   - user log in?
   - content updates?

**Proposed solution or definition of done:** 
- Ability to use Robo to run command against all sites on all D10 applications
- Produce list of sites we determine are inactive based on user log in info, excluding admins
- Ability to pass in different lengths of time
</details>

briand44:
## Grooming
 
**Quadrant:** 

Complex

 **Discussion:**
[...]

PR #9770, opened 2026-04-21
Inactive websites robo command

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/issues/9759

A report was delivered here https://github.com/uiowa/uiowa/issues/9759#issuecomment-4292113201

# How to test

`ddev auth ssh`
`ddev ssh`
`./vendor/bin/robo uiowa:report:inactive` (defaults to all apps and 1 year and terminal output)

OR 

`./vendor/bin/robo uiowa:report:inactive --apps=uiowa01,uiowa02 --threshold="6 months" --export`

Check the domains report is still working as a lot of stuff was moved around and shared.

`./vendor/bin/robo uiowa:report:domains --application=uiowa02 --env=dev --export`

Last revision excludes revisions made by user 1 (database updates) but includes administrators (e.g. jwhitsit) whereas login is based on non-admins/user 1 and is when they logged in, not when they last accessed (authenticated session persisted). Days are rounded.


joewhitsitt:
in preparing for demo, i realized i broke the default 1 year threshold (no arg). passing the `--threshold="1 year"` still works. i'll try to fix here shortly.

pyrello:
Pulling to test

pyrello:
This seems wrong?
```
...
➜    Checking sung.lab.uiowa.edu...
➜    Checking uimobileclinic.org.uiowa.edu...
➜    Checking usthaiconsortium.conference.uiowa.edu...
➜  Here are the results.
+-------------+----------------------------------------------------+---------------------+------------------+------------------------+
| Application | URL                                                | Days Since Revision | Days Since Login | Login Inactive: 1 year |
+-------------+----------------------------------------------------+---------------------+------------------+------------------------+
| uiowa       | accreditation.uiowa.edu                            | N/A                 | N/A              | Error                  |
| uiowa       | ace.lab.uiowa.edu                                  | N/A                 | N/A              | Error                  |
| uiowa       | adsforruralamerica.uiowa.edu                       | N/A                 | N/A              | Error                  |
| uiowa       | advising.uiowa.edu                                 | N/A                 | N/A              | Error                  |
| uiowa       | afr.fo.uiowa.edu                                   | N/A                 | N/A              | Error                  |
| uiowa       | alphachisigma-alphatheta.org.uiowa.edu             | N/A                 | N/A              | Error                  |
| uiowa       | ap-purchasing.fo.uiowa.edu                         | N/A                 | N/A              | Error                  |
| uiowa       | asce.engineering.uiowa.edu                         | N/A                 | N/A              | Error                  |
| uiowa       | assessment.uiowa.edu                               | N/A                 | N/A              | Error                  |
| uiowa       | audit.org.uiowa.edu                                | N/A                 | N/A              | Error                  |
| uiowa       | autismcenter.centerforconferences.uiowa.edu        | N/A                 | N/A              | Error                  |
| uiowa       | bands.uiowa.edu                                    | N/A                 | N/A              | Error                  |
| uiowa       | billing.uiowa.edu                                  | N/A                 | N/A              | Error                  |
| uiowa       | biomost.engineering.uiowa.edu                      | N/A                 | N/A              | Error                  |
| uiowa       | brand.uiowa.edu                                    | N/A                 | N/A              | Error                  |
| uiowa       | budget.fo.uiowa.edu                                | N/A                 | N/A              | Error                  |
...
```

joewhitsitt:
Thanks for flagging this. I can pull and check this as it is not expected.

joewhitsitt:
@pyrello I think you ran without entering the container with `ddev ssh`. Running outside I got the same result. Added some guardrails similar to what we have in blt/src/Blt/Plugin/Commands/ValidateCommands.php

pyrello:
> @pyrello I think you ran without entering the container with `ddev ssh`. Running outside I got the same result. Added some guardrails similar to what we have in blt/src/Blt/Plugin/Commands/ValidateCommands.php

This is what I ran:
```
ddev ssh
sadamshiett@uiowa-web:/var/www/html$ ./vendor/bin/robo uiowa:report:inactive
➜  Fetching domains from Acquia Cloud API...
➜  Processing uiowa...
➜    Checking accreditation.uiowa.edu...
➜    Checking ace.lab.uiowa.edu...
...
```

joewhitsitt:
Hmm. What about `ddev auth ssh`? My guess is that connection is not being properly established. Thinking about how to exit early if that was the case.

pyrello:
> Hmm. What about `ddev auth ssh`? My guess is that connection is not being properly established. Thinking about how to exit early if that was the case.

Okay, I exited out of DDEV and tried to ssh into an Acquia application and was denied, so I think I did not run `ddev auth ssh`. I guess I would have expected an error message the first time it checked a domain?

joewhitsitt:
Yeah, still trying to think that part through as I am currently chalking up empty databases or not found drush aliases as errors in the output. A preflight ping might be necessary before each site is hit.

joewhitsitt:
Found some ssh-agent checks among the .ddev/providers stuff. Didn't like the idea of a separate drush ping and didn't want to unweave the null/error handling when a website is hit.

```
Starting uiowa... 
 Container ddev-ssh-agent Started  
ssh-agent container is running: If you want to add authentication to the ssh-agent container, run 'ddev auth ssh' to enable your keys. 
```

```
jwhitsit@VPER-OSC-194 uiowa % ddev ssh                                                                                                                                                                      [05/1/26 | 12:18:17]
jwhitsit@uiowa-web:/var/www/html$ ./vendor/bin/robo uiowa:report:inactive --apps=uiowa09
➜  [ERROR] No SSH keys loaded. Please 'ddev auth ssh' before running this command.
```

briand44:
I'm curious about the thinking around filtering out .prod.drupal domains? I could see value in understanding those that are inactive as well but haven't put much thought into this yet.

joewhitsitt:
> I'm curious about the thinking around filtering out .prod.drupal domains? I could see value in understanding those that are inactive as well but haven't put much thought into this yet.

@briand44 My honest answer to that question is that it is easier to get the drush alias if all of the domains are in the same format. 😁 

We also kind of determined that when auditing the internal domains earlier we kept telling folks "no rush, just wanted to make sure..." 

IMO, I saw identifying inactive was for live websites that are no longer in use or might be serving inaccurate/out-of-date content to the public.

briand44:
The internal prod domains don't need to be in scope as part of this slice since we did audit these recently. I do think we'll want to have some sort of process in the future to keep an eye on these as there are a pretty significant number.

>We also kind of determined that when auditing the internal domains earlier we kept telling folks "no rush, just wanted to make sure..."

Yeah, from my perspective it is no rush if they are working on them. If they haven't been logged into or updated in over a year, it would be interesting to understand.

joewhitsitt:
I am looking into this a bit, but I think it might be a follow-up slice unless you think it is the expected deliverable of https://github.com/uiowa/uiowa/issues/9759 to include internal domains.

Right now domains and application info are fetched from Acquia, internal domains, etc. are filtered out and we get the drush alias the same way the umc creates them. Instead it might make sense to loop through the codebase's drush/sites dir to get the drush alias for user/content data and then match to a result from the acquia domains call to determine if the website is at the internal domain still or if it has launched. application can either come from the domains call or the drush yaml.

I guess understanding if you want the current domain (internal prod) or the expected domain (final live domain) in the results steers how this is put together.

This would help reduce noise of htaccess redirect rows and duplicative www rows.

pyrello (approved):
Works as expected.

PR #9709, opened 2026-03-30
[p2lb] a11y fixes

bspeare:
Resolves https://github.com/uiowa/uiowa/issues/9500. 
Resolves https://github.com/uiowa/uiowa/issues/9711. 

# How to test https://github.com/uiowa/uiowa/issues/9500
```
 ddev blt sync --site hawkeyemarchingband.uiowa.edu --define drush.aliases.remote=hawkeyemarchingband.test && ddev drush @hawkeyemarchingband.local config-split:activate p2lb -y && ddev drush @hawkeyemarchingband.local uli admin/content/sitenow-converter
```

1. Test https://hawkeyemarchingband.uiowa.ddev.site/admin/content/sitenow-converter with Siteimprover converter and confirm that error about heading is gone. 
2. Test https://hawkeyemarchingband.uiowa.ddev.site/admin/content/sitenow-converter/converted and confirm that error message about empty table cell is gone.  Convert a page and confirm that actions checkbox appears in the table but "hides if empty" to avoid accessibility issue. 
3. Code review views changes. 

## How to test other places where "No data cells assigned to table header" occur in the platform

```
ddev drush @oniowa.local sql-drop -y && ddev blt ds --site=oniowa.uiowa.edu && ddev drush @oniowa.local uli
```
1. Go to https://oniowa.uiowa.ddev.site/admin/content/events and delete all events and confirm that there is not Siteimprove table error. Might want to edit the view to be able to delete 250+ events. 
2. Review my comment on https://github.com/uiowa/uiowa/issues/9500#issuecomment-4163440927 and chime in your thoughts on the core overview pages. 

# How to test https://github.com/uiowa/uiowa/issues/9711
```
 ddev blt sync --site admissions.uiowa.edu && ddev drush @admissions.local uli admin/content/
```
1. Spot check and run accessibility review on pager heading with Siteimprove plugin.  Compare to prod and confirm heading structure error is no longer there. 
3. Code review the rest of the changes.

PR #9785, opened 2026-04-28
Switch from fullcalendar_view to fullcalendar

### LINKED ISSUE #9365: Investigate Full Calendar accessibility compliance issues
# Motivation

SiteImprove is reporting a wcag 2.1 A issue with calendar displays.

>This page has a scrollable section that can't be operated by keyboard.

**Examples**

Admissions
https://my2.siteimprove.com/Inspector/998093/A11y/Page?pageId=71788717907&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r84/failed/bVNFOUdSMDc=;MTM2ZGVkNzg=

Grad
https://my2.siteimprove.com/Inspector/998426/A11y/Page?pageId=29973646729&impmd=264ED37B65FA6186129334A508391970&conf=a+aa#/sia-r84/failed/ZkQ5NTFybFk=;NDg3YjE5NzI=

Related issue #8497

ITS
https://my2.siteimprove.com/Inspector/994335/A11y/Page?pageId=28396872889&impmd=264ED37B65FA6186129334A508391970&conf=a#/sia-r84/failed

---

Also seeing another issue reported related to calendar displays: No data cells assigned to table header

>Every header cell in a table should have at least one data cell assigned to it.

https://my2.siteimprove.com/Inspector/998093/A11Y/Page?pageId=71788717907&impmd=264ED37B65FA6186129334A508391970&issue=failed&decision=dismissed+falsePositive+none&comp=ada#/sia-r46/failed/TGFBWVhmbzM=;ZDVhNmI2ZDY=

joewhitsitt:
Closes #9365

This PR switches from the `fullcalendar_view` module to the `fullcalendar` module based on the results of #9365 determining it was a better fit for our needs. This change requires a config switchover for the one remaining view we have that still uses Fullcalendar, the ITS alerts calendar. Customizations we had in place have been adapted.

Note: `fullcalendar_view` will be removed in a follow-up PR once this has deployed to avoid deployment downtime.

# How to test
- Code review
```
ddev yarn workspace uids_base run build &&
ddev blt ds --site its.uiowa.edu
```
- https://its.uiowa.ddev.site/alerts/calendar and confirm:
  - Calendar only looks back 90 days (3 months) and beyond that you are not able to click the back arrow.
  - When loaded in a mobile-sized viewport (below width of 768px), the display defaults to the list display.
  - In mobile the calendar display looks similar to production.
  - Sticky headers sit on top (z-indexed) of calendar events scrolling underneath them in list view - different than production
  - The calendar and list views generally look similar to https://its.uiowa.edu/alerts/calendar except where noted
- Look at the calendar events in the calendar view and confirm:
  - The time displays above the event text - different than production
  - Long text inside events wraps onto multiple lines.
  - There is a light amount of padding.
  - The time format matches production.


pyrello:
# Left to do (from POC)

- [x] Supplemental JS to change nav images from role=img to role=presentation - fixes Image missing a text alternative issues
- [x] Same white-space normal hack to display full pills of info
- [x] Flex styles to display full time
- [x] Horizontal scroll on mobile
- [x] Adjust the event split calendar display similarily.
- [x] views list display alternative
- [x] limit the number of past alerts, disable calendar nav back.

pyrello (commented):
Two small things worth picking up before this comes out of draft:

**1. Nullsafe access in `its_core_views_pre_render()`** — [docroot/sites/its.uiowa.edu/modules/its_core/its_core.module:518](https://github.com/uiowa/uiowa/blob/fullcalendar_swap/docroot/sites/its.uiowa.edu/modules/its_core/its_core.module#L518)

```php
$filter = $view->filter['field_alert_date_value'] ?? NULL;
$offset = $filter->options['value']['value'] ?? '-90 days';
```

If `$filter` is null, `$filter->options` throws a PHP 8 "attempt to read property on null" warning — the `??` only catches the inner array access. In practice the filter is always there on this display, so it's defensive, but the current shape reads as if it's null-safe when it isn't. Suggested:

```php
$offset = $filter?->options['value']['value'] ?? '-90 days';
```

**2. `composer.lock` `plugin-api-version` downgraded** from `2.9.0` → `2.6.0`. Probably an artifact of regenerating the lockfile on an older Composer; worth re-locking on a current environment so we don't accidentally roll back the platform plugin API for unrelated future installs.

(Confirmed staged-removal is intentional for `drupal/fullcalendar_view` — sequential per-site config import on the multisite means dropping the package in this PR would fatal every site that hasn't yet had its config cleared. So that one's not a comment, just noted for my own understanding.)

— Claude Opus 4.7 (assisting Sean)

pyrello:
This is mostly ready for review. I'm in the process of troubleshooting the split command in #9821 based on discovering last week it was not reporting all the sites with event split turned on. That will allow a complete check against sites that have the split enabled before we finalize the details.

pyrello:
Confirmed that none of the sites with the `event` split touch `views.view.event` config, so the display that was deleted there should be good to go.

pyrello:
Grooming feedback: 
- Time display does not match the format in production
- Mobile view needs another pass

pyrello:
> Grooming feedback:
> 
>     * Time display does not match the format in production
>     * Mobile view needs another pass

Done.

bewoldt:
@pyrello I can review this.

bewoldt:
Adding some css to move the Month to it's own line making the header easier to read on mobile. 

Before
Screen Shot 2026-05-19 at 13 50 39 PM

After
Screen Shot 2026-05-19 at 13 57 07 PM

@media (max-width: 768px) {
  .fc-header-toolbar.fc-toolbar {
    flex-wrap: wrap;
  }

  .fc-header-toolbar .fc-toolbar-chunk:nth-child(2) {
    order: -1;
    width: 100%;
    text-align: left;
    margin-bottom: 8px;
  }
}

bewoldt:
- [x] The Calendar only looks back 90 days 
- [x] The view defaults to the list view on 768px breakpoint.
- [x] The time displays above the event text - different than production
- [x] Long text inside events wraps onto multiple lines.
- [x] There is a light amount of padding.

## Notes
- The time varies slightly with production when on the List view.
  - production = 02:00 am - 02:00 pm. 
  - new = 2:00am - 2:00pm
- In the new version, the days of the Month are no longer clickable. In the production version, you could click on the day an the month and get a list of events for that day.
- The line-height for the text inside the event on the production version is 1.4. The new version is 1.6. The 1.6 give the text some space but takes more room.
- When using the Month view in the mobile breakpoint, the table doesn't stretch 100%
Screen Shot 2026-05-20 at 10 27 26 AM



## Accessibility
**color contrast issues**
- text using the gray background “Today” button and “List” button.
- the brand secondary colors have contast issues when using small text.

**ARIA attribute unsupported or prohibited**
- Siteimprove flags use of aria-labelledby on the anchors for days of the week and dates.

**Table header cell is missing a header role**

pyrello:
> The time varies slightly with production when on the List view.
>     * production = 02:00 am - 02:00 pm.
>     * new = 2:00am - 2:00pm

I was consulting the brand style guide to see where we are supposed to land on this.

> Do not use unnecessary colons and ciphers in figures indicating time of day.
> 
>     7 a.m. (Not: 7:00 a.m.)
> 

Source: https://brand.uiowa.edu/editorial-style-guide#accordion-item-44256-4

> The abbreviations “a.m.” and “p.m.” are lowercase with periods and no space between letters. There should be a single space between “a.m.” or “p.m.” and the numeral that precedes them.
> 
>     6 a.m.
>     10 p.m.
> 
> In display type, tabular material, or lists, “a.m.”/“p.m.” style may vary depending on graphic design. For example, small caps may be used, or the periods may be deleted. [UI]

Source: https://brand.uiowa.edu/editorial-style-guide#accordion-item-44331-3

The above confused me a bit, so I'm reaching out to the people in charge of the style guide to clarify. Specifically, the rules about omitting "unnecessary" colons and numbers from times and whether a.m./p.m. should always contain periods. The second one already has an answer: yes.

joewhitsitt:
>The above confused me a bit, so I'm reaching out to the people in charge of the style guide to clarify. Specifically, the rules about omitting "unnecessary" colons and numbers from times and whether a.m./p.m. should always contain periods. The second one already has an answer: yes.

I would stress too much about this as we need to address this for the whole platform: https://github.com/uiowa/uiowa/issues?q=is%3Aissue%20state%3Aopen%20ap%20style and it isn't very easy to do.

pyrello:
> > The above confused me a bit, so I'm reaching out to the people in charge of the style guide to clarify. Specifically, the rules about omitting "unnecessary" colons and numbers from times and whether a.m./p.m. should always contain periods. The second one already has an answer: yes.
> 
> I would stress too much about this as we need to address this for the whole platform: https://github.com/uiowa/uiowa/issues?q=is%3Aissue%20state%3Aopen%20ap%20style and it isn't very easy to do.

I don't think that this use's Drupal's date formats. I checked on it and have an answer about how it should look.

pyrello:
Updated:
- Calendar header (month name) is moved to its own line above toolbars for mobile views.
- Space added between time and `am`/`pm`.
- Day numbers are clickable and display a list of events for that day.
- Event text line height set to 1.4

No additional changes necessary to comply with style guide rules.

bewoldt:
> Updated:
> 
>     * Calendar header (month name) is moved to its own line above toolbars for mobile views.
> 
>     * Space added between time and `am`/`pm`.
> 
>     * Day numbers are clickable and display a list of events for that day.
> 
>     * Event text line height set to 1.4
> 
> 
> No additional changes necessary to comply with style guide rules.

I’ve reviewed the updates and verified that the changes have been implemented.

pyrello:
@joewhitsitt Was this an intentional change? https://github.com/uiowa/uiowa/pull/9785/changes#diff-72f52cd212430d5025a6dd79f4f288d9e56c5e1967e7a1ebd708d21b980d9958 (change to `config/sites/its.uiowa.edu/config_split.patch.views.view.content_reference.yml`)

joewhitsitt:
@pyrello not intentional. might have been a config blip related to the recent [pager a11y work Ben worked on](https://github.com/uiowa/uiowa/pull/9709/changes). didn't check the config output when I originally pushed.

pyrello:
The SCSS file was not using anything that strictly needed SASS so I took advantage and converted it to CSS. In doing so, also moved it to the ITS Core module so that all the dependencies are scoped to that one site.

joewhitsitt:
pulling to check few items

composer.json
```diff
@@ -309,8 +310,9 @@
             "drupal/fragments": {
                 "remove_views_data_type": "https://www.drupal.org/files/issues/2022-08-24/3305851-remove_views_data_type-6.patch"
             },
-            "drupal/fullcalendar_view": {
-                "only_rebuild_on_view_change": "patches/fullcalendar_only_rebuild_on_view_change.patch"
+            "drupal/fullcalendar": {
+                "Expose Calendar instance on DOM element for external JS access": "patches/fullcalendar_expose_calendar_instance.patch",
```
joewhitsitt, line 315: Bit of a nitpick. It seems we have been starting a convention of naming the patch file with the issue number and making the issue number part of the composer declaration. Can we keep that pattern going. This one in particular is hard to track down.

docroot/sites/its.uiowa.edu/modules/its_core/css/fullcalendar-overrides.css
```diff
@@ -0,0 +1,104 @@
...
+.fc-button-primary {
+  background-color: var(--brand-secondary) !important;
+}
+
+.fc-button-primary:hover {
+  background-color: var(--brand-secondary) !important;
+}
+
+.fc-button-primary:focus {
+  box-shadow: 0 0 0 0 var(--brand-secondary) !important;
```
joewhitsitt: was this a carry over from earlier? with all zeros, this is essentially invisible. is that intended? should it have some focus?

```suggestion
  box-shadow: 0 0 0 0.2rem var(--brand-secondary) !important;
```

composer.json
pyrello: That patch does not correspond to an issue. The Fullcalendar module JS runs in a way that doesn't allow us access to affect some of its internals. The patch adds a line that exposes the FC object so that we can act on it in JS.

joewhitsitt: maybe a comment in the patch file then to that effect? it caught my attention initially because it wasn't very clear on glance.

pyrello:
Both review items that @joewhitsitt posted are addressed.

cory-skeers (approved):
Approving based on other reviews 👍

joewhitsitt:
I can hold on further dependabot pulls until this is in a happy mergable state

PR #9796, opened 2026-05-01
[Signage] Add room number to events slides

### LINKED ISSUE #9677: Signage - Display room number on campus event listing slides
# Story

As a Sign Manager, I would like to output the room number along with the existing campus location when displaying campus events on our Signs so that visitors can find the appropriate room location for the event they are interested in.

# Additional Info

<details>
<summary>Original Request</summary>
> I hope you are doing well today. I have a small request regarding one of our signs that pulls from the event calendar. In the law building, we have a sign that lists our events for the week (see below). Would it be possible to have this sign also display the room number that is listed on the event calendar posting as well? It doesn’t really help with wayfinding for our visitors if it just says Boyd Law Building.
</details>

Request from Andrew S. and Katie L. in College of Law.

**Questions**
- Option to toggle or always display if provided?
  - Leaning towards always display if provided (Maybe prepend ` - [room_number]`?
- Do we need to do anything with custom addresses?
  - Leaning towards no until we get requests to output full address

cory-skeers:
Resolves #9677 

# How to test

```
ddev blt ds --site=signage.sites.uiowa.edu && ddev drush @sitessignage.local uli /node/2251
```

- Clean sync, and see that the sign doesn't include room numbers (default state for existing signs, and an empty field)
- Edit the sign, edit the slide, and check to "display room number"
- See that the room numbers are now displaying (non-default state, field to true)
- Repeat, and turn the "display room number" back off
- See that the room numbers no longer display (non-default state, field to false)
- Try adding a new events slide, with and without the room number, and see that it displays correctly
- Check an events block (such as at https://sandbox.uiowa.ddev.site/extra-events-testing), and see that the room number is not displaying

### DIFF
--- a/config/sites/signage.sites.uiowa.edu/core.entity_form_display.paragraph.slide_events.default.yml
+++ b/config/sites/signage.sites.uiowa.edu/core.entity_form_display.paragraph.slide_events.default.yml
@@ -9,6 +9,7 @@ dependencies:
     - field.field.paragraph.slide_events.field_slide_events_interests
     - field.field.paragraph.slide_events.field_slide_events_keywords
     - field.field.paragraph.slide_events.field_slide_events_location
+    - field.field.paragraph.slide_events.field_slide_events_room_num
     - field.field.paragraph.slide_events.field_slide_events_types
     - paragraphs.paragraphs_type.slide_events
 id: paragraph.slide_events.default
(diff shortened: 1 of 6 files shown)


cory-skeers:
I'm wondering if there might be a desire to include room numbers in other campus events feeds in SiteNow?

The way I have it currently, the logic for adding the room number is in the signage module and targets the events slide.

Another approach would be to add an option in the `sitenow_events_build_card` to do the location/room concatenation (default to no room if it's not marked true). 

I might pivot to that route just to make the option available regardless, and keep the logic consistent.

joewhitsitt:
pulling to test

joewhitsitt (approved):
>there might be a desire to include room numbers in other campus events feeds in SiteNow

I imagine there probably desire but I think we should hold on additional decorations (if possible) until we hear more about what might be coming down the pipeline from the campus events team. This PR does a good job of positioning to adopt that later on.

PR #9802, opened 2026-05-05
Add MathJax support

### LINKED ISSUE #9814: Allow MathJax in SiteNow
# Story

As a SiteNow editor, I would like the ability to add LaTeX code in the SiteNow WYSIWYG with MathJax module so we can output mathematical notation on our websites.

# Additional Info

As part of https://github.com/uiowa/uiowa/issues/9753, we investigated different options for displaying mathematical notation in SiteNow. Based on that investigation (https://github.com/uiowa/uiowa/issues/9753#issuecomment-4392542886), using the contrib project https://www.drupal.org/project/mathjax, seemed like the best path forward.

We demoed this to CLAS stakeholders and this seems to fit their needs. They would like us to give a quick demo to George M. and Cori L. once this is rolled out.

PR exists from the investigation - https://github.com/uiowa/uiowa/pull/9802

**Summary of next steps from the investigation**
There may be more cleanup and work to be done with the POC to get it ready for primetime, if that's the route we're going. There's also a need for documentation, either on its own, or possibly with the addition of help text which may be able to be included if its in its own "filtered html and also math" text format.

Likely next steps

- Install drupal/mathjax (default? split?)
- Set up a new text format with mathjax (copy of filtered_html? more or less features? which roles have access?)
- Move the mathjax filter above the faulty html filter (or remove the faulty html filter for this format). Evaluate other filters.
- Evaluate and decide whether to turn on/off the math settings (see below)
- Documentation? Help text?
- Further testing

cory-skeers:
# How to test

## Docs
- [Check out the docs I have made for this here.](https://sitenow.uiowa.edu/filtered-html-math-text-format)
- Are they helpful? Anything to change? I have tried to keep them short and poignant. 
- There is some debate on whether we should keep the reference or not, give your thoughts and remove it if it is out of scope.

## Code
- `ddev composer install`
- Sync a site
- Go into a WYSIWYG text editor and change the text format to "Filtered HTML + Math".
- Put in some LaTaX maths surrounded by `\(...\) ` (inline) or `\[...\]`  
    - If you would like some example text, check below for a copyable source HTML.
- Save and view the page
- DO YOU SEE MATH?
- ... Do you see any issues?
- Try to break it!

### Post Launch
- We will need to publish the page on the sitenow site under the WYSIWYG documentation menu item in the menu tree.

<details><summary><b>Copyable `source` HTML for testing.</b></summary>

```

<h2>
    MathJax Test Cases
</h2>
<h3>
    Inline delimiters
</h3>
<p>
    Dollar signs: The quadratic formula is $x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$ inline.
</p>
<p>
    Parens: The quadratic formula is \(x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}\) inline.
</p>
<h3>
    Display delimiters
</h3>
<p>
    Double dollar: $$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$
</p>
<p>
    Brackets: \[x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}\]
</p>
<h3>
    Long square roots (the overflow fix)
</h3>
<p>
    \[\sqrt{\frac{a^2 + b^2 + c^2 + d^2 + e^2 + f^2}{n-1}}\]
</p>
<p>
    \[\sqrt{\int_0^\infty e^{-x^2}\,dx}\]
</p>
<h3>
    Fractions and nested expressions
</h3>
<p>
    \[\frac{\frac{1}{x}+\frac{1}{y}}{y-z}\]
</p>
<h3>
    Summation and integrals
</h3>
<p>
    \[\sum_{n=1}^{\infty} \frac{1}{n^2} = \frac{\pi^2}{6}\]
</p>
<p>
    \[\int_{-\infty}^{\infty} e^{-x^2}\,dx = \sqrt{\pi}\]
</p>
<h3>
    Greek letters and operators
</h3>
<p>
    \[\alpha + \beta = \gamma, \quad \Delta E = mc^2, \quad \nabla \cdot \vec{B} = 0\]
</p>
<h3>
    Matrix
</h3>
<p>
    \[\begin{pmatrix} a &amp; b \\ c &amp; d \end{pmatrix} \begin{pmatrix} x \\ y \end{pmatrix} = \begin{pmatrix} ax+by \\ cx+dy \end{pmatrix}\]
</p>
<h3>
    Aligned equations
</h3>
<p>
    \[\begin{aligned} (a+b)^2 &amp;= a^2 + 2ab + b^2 \\ (a-b)^2 &amp;= a^2 - 2ab + b^2 \end{aligned}\]
</p>
<h3>
    Escaped dollar sign (should NOT render)
</h3>
<p>
    This costs \$5.00 and that costs \$10.00.
</p>
<h3>
    Mixed inline in a sentence
</h3>
<p>
    If $f(x) = x^2$ then $f'(x) = 2x$ and $f''(x) = 2$, so the function is convex for all $x \in \mathbb{R}$.
</p>

```

</details>


config/default/mathjax.settings.yml
```diff
@@ -0,0 +1,7 @@
+_core:
+  default_config_hash: UOFmZIwiZl6qJ50XXrC2r_F26ZTYR_3-d93LAnL8J4w
+use_cdn: 1
+cdn_url: 'https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS-MML_HTMLorMML'
+config_type: 0
+default_config_string: '{"tex2jax":{"inlineMath":[["$","$"],["\\\\(","\\\\)"]],"displayMath":[["$$","$$"],["\\\\[","\\\\]"]],"processEscapes":true},"showProcessingMessages":false,"messageStyle":"none","showMathMenu":false}'
```
cory-skeers:
```suggestion
default_config_string: '{"tex2jax":{"inlineMath":[["$","$"],["\\\\(","\\\\)"]],"displayMath":[["$$","$$"],["\\\\[","\\\\]"]],"processEscapes":true},"showProcessingMessages":false,"messageStyle":"none","showMathMenu":true}'
```

config/default/metatag.settings.yml
```diff
@@ -50,5 +50,5 @@ tag_trim_maxlength:
   metatag_maxlength_twitter_cards_title: null
   metatag_maxlength_twitter_cards_description: null
 tag_scroll_max_height: ''
-tag_trim_end: "|.,-:;/+&([{\"'"
```
joewhitsitt, line 53: I feel like this would break the regex. @cory-skeers do you recall how sensitive this new config is?

GaryRidgway: ~~its escaping that quote with the backslash i believe?~~

_(The above was unintentionally irrelevant.)_

GaryRidgway: This was automatically generated as well, I don't think I specifically made this change. It was made with the `ddev drush @default.local cex` command.

joewhitsitt: i understand it was auto exported, just that additional testing/understanding is now necessary to make sure there isn't a regression here.

GaryRidgway: for sure, just wanted to provide context.

cory-skeers: This would remove the `'` and `"` from being considered in the end trimming process, which would differ from the metatag default. It looks from the change like it's trying to clean things up by removing what could be considered unnecessary quotes (the value can probably be interpreted as a string without being designated as such), except that they're needed to be able to include `'` and the escaped `"`.

GaryRidgway: Okay, lets roll this one back then. Thanks for the catch @joewhitsitt, and the consult @cory-skeers !

joewhitsitt:
pulling to test

joewhitsitt (approved):
I took the _any_ site directive and tried hawkeyemarchingband (v2) in addition to admissions. My math skills are lacking but was able to put together some examples from my days with the gov (can't explain it though).

Image

My only thought would be to add an accessibility statement or something that points out it scans wcag 2.1 A/AA (siteimprove browser widget) compliant https://docs.mathjax.org/en/latest/basic/accessibility.html and that MathJax provides keyboard navigation that is a bit different. It is announced as an application. https://docs.mathjax.org/en/latest/basic/explorer-commands.html#explorer-keyboard-commands

joewhitsitt:
Mentioned the a11y doc addition after standup https://iowaweb.slack.com/archives/C0164FNN1PV/p1779977781976229

Went ahead and updated the draft page to include a statement, link to the keyboard docs, added the page to the menu system under the wysiwyg page.

PR #9826, opened 2026-05-13
[dependabot] avoid major version bumps similar to composer

joewhitsitt:
This should avoid the cssnano issue of upgrading beyond what our current node supports.

https://github.com/uiowa/uiowa/pull/9824


joewhitsitt:
I think some of the ungrouped items are security items

pyrello (approved):
Reviewed, not tested.

Issue #9746, opened 2026-04-10
Investigate standardizing heading functionality


briand44:
# Motivation

We are currently running an unsupported version of the heading project https://www.drupal.org/project/heading. We also need to prepare for an upgrade to Drupal 11, which our current version will not work with.

Based on conversations from https://github.com/uiowa/uiowa/issues/8866, we would like to investigate standardizing heading functionality around a single contrib project or standardize around our uiowa_headline feature.

# Additional Info

https://www.drupal.org/project/heading
https://www.drupal.org/project/heading_formatter
uiowa_headline - https://github.com/uiowa/uiowa/blob/main/docroot/modules/custom/uiowa_core/src/HeadlineHelper.php?

Some comments from #8866 
- https://github.com/uiowa/uiowa/issues/8866#issuecomment-3329516861
- https://github.com/uiowa/uiowa/issues/8866#issuecomment-3329963867

briand44:
## Grooming
 
**Quadrant:** 

Complex

 **Discussion:**
 - Consider replacing with https://www.drupal.org/project/heading_formatter
   - provides option to add classes to headings
 - What does this do vs uiowa_headline?
   - uiowa_headline handles children headings
 - Investigation into how to proceed?
   - heading_formatter vs heading vs uiowa_headline
   - can we standardize around something that can minimize our custom code footprint? 
   - if not, can we standardize around our custom code?
   - investigation - complex - evaluate and weigh features, scoped to investigation and plan for moving forward, size medium

**Proposed solution or definition of done:**
- review our custom code, and heading contrib projects to determine best way to standardize heading functionality across SiteNow
- Provide recommendation on how we should move forward

GaryRidgway:
Writing things down that I've found before I un-assign myself, considering my work in other places, I don't want to squat.

So far in my investigation it seems like heading_formatter may not be the way to go considering it is a field formatter, and we lose the ability to configure heading size per block.

From a small investigation with claude, it seems that perhaps a way that we could reduce the complexity of our code is by removing uiowa headline. We can use the heading module and then add custom fields in the headline helper for that.

it doesn't seem like we will be able to find a catch all easily with the proposed modules.

<details><summary>Here is a rundown of what was proposed by claude, interesting, but not picked apart.</summary>


Here's what Option C looks like concretely. Reading my own sketch back — this is a one-field migration that touches a focused set of files; the surface area is smaller than the earlier discussion implied.

## Scope

[...]

PR #9827, opened 2026-05-14
Update drupal/heading

### LINKED ISSUE #8866: Upgrade `drupal/heading` to 2.0.x
Changelog: https://git.drupalcode.org/project/heading/-/compare/8.x-1.7...2.0.0

Config schema is changing, so possibly an update hook will be necessary, possibly just to re-save entities where heading field is used.

# Proposed solution
- Audit usages of heading field.
- Test out upgrading module and determine whether entities with heading field need to be re-saved or require an update hook to modify existing config to match new schema.
- Write update hook if necessary.

cory-skeers:
Resolves #8866 
Relates to #9746

# How to test

```
git checkout update-drupal_heading &&
ddev composer install &&
ddev blt ds --site=sandbox.uiowa.edu &&
ddev drush @sandbox.local uli
```

Check for a clean sync
The blocks that use the heading field are: banner, card, cta, event, hero, statistic
For some or all of these, check adding a new block, editing an existing block, and see that everything saves correctly, the heading gets the right h level

```
ddev blt di --site=default
```
Check for a clean install

The heading text field character limit was bumped up from 128 to 256, but because it's being upped, there shouldn't be any regressions for existing content. Otherwise, there are schema updates, but it's completing formerly incomplete/improper schema, so no update hook needed.


GaryRidgway:
I can pull to test

GaryRidgway (approved):
Things look good to me!

Issue #9337, opened 2025-10-31
Ability to display Faculty jobs over 1 year old


briand44:
# Story

As a SiteNow webmaster, I would like to be able to display faculty jobs that were posted over a year ago so that we can list all available jobs on our website.

# Additional Info

Currently Feed items with post dates within 1 year are imported. - https://sitenow.uiowa.edu/documentation/layout-builder/rss-feeds

I believe the idea for jobs is that jobs should be updated if they are over a year old but apparently that is not an attainable option for Faculty jobs. They don't have the ability to revise the post date and would need to go through a whole process in order to remove a job listing and create a new listing which isn't feasable.

Example currently displaying 10 job listings - https://pediatrics.medicine.uiowa.edu/about-us/join-our-team/faculty-jobs

Example Feed url currently display more than 10 job listings - `https://hris.uiowa.edu/otac-rss/faculty.php?dept-code=17-3255`

Staff jobs feed documentation - not sure if this will be helpful at all.
- https://iowa.sharepoint.com/sites/CSI-Drupal2/Shared%20Documents/Forms/AllItems.aspx?viewid=bf54b9d9%2D0975%2D4fbc%2D9f1b%2D6377bdc2307c&id=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FGeneral%2FStaff%20Job%20RSS%20Feed%20Reference%2Epdf&parent=%2Fsites%2FCSI%2DDrupal2%2FShared%20Documents%2FGeneral

joewhitsitt:
The other recommendation when this came up earlier was to highlight these jobs outside of the feed on a page. 1 year is the configurable max for aggregator. We would be again into custom code territory with this, not saying we shouldn't do that, but I think this should be per site setting (site-split) and at that point maybe we need to create that jobs block after all.

<img width="456" height="511" alt="Image" src="https://github.com/user-attachments/assets/2127cb6f-2f7d-41ea-9c68-6681fbc65c54" />

quamsta:
Would love to see this-- I have a mixture of faculty positions [on this page ](https://pathology.medicine.uiowa.edu/about-us/join-our-team/faculty-positions)from a feed and one solitary position created over a year ago, manually made with a card component that I'd love to just have be part of the feed.

Out of scope for this issue, but it would also be nice to not have all-caps headings for these, but I'm guessing that's on the jobs.uiowa.edu end of things.

briand44:
## Grooming
 
[...]

PR #9835, opened 2026-05-19
Jobs feed

cory-skeers:
Relates to #9337 

# How to test

- Sync your site of choice and log in
- Edit a page in Layout Builder
- Try adding a new Job Postings block
- Try a non-hris feed (any RSS feed or anything else), and see that you can't save the block
- Try an HRIS feed (https://pediatrics.medicine.uiowa.edu/about-us/join-our-team/faculty-jobs uses https://hris.uiowa.edu/otac-rss/faculty.php?dept-code=17-3255 and https://hris.uiowa.edu/otac-rss/jobs.php?dept-code=17-3255)
- See that the jobs load in based on the feed you've used
- Test out the various options, see that they work
- Try out the full staff jobs feed (https://hris.uiowa.edu/otac-rss/jobs.php) with paged or unpaged, and see that it caps itself off and doesn't take a million years to load
- Review documentation at https://sitenow.uiowa.edu/documentation/page-content-type/layout-builder/job-postings

## Additional notes:

- Card should link to the actual job posting
- Default should be "show posted date" and "don't show description"
- The Faculty jobs are all caps from the source--that's expected
- The descriptions have a series of "This: Thing" in the beginning--that's expected


cory-skeers:
Bah. Worked off a dirty local branch

cory-skeers:
The way this is currently set up, the non-paged version has an upper limit on results (25), but the pager version does not. This means you can go to the generic `.../jobs.php` endpoint and pull in all jobs, 732 at the time of testing.

It actually works smoothly once it gets cached, but uncached page hits require a pretty significant upfront load.

We could 

- limit it to the `...faculty.php` endpoint (with no query parameters, currently 119, which is not insignificant, but operates pretty speedily), 
- require some sort of filtering on the feed URL,
- cap the paged version to a certain number of results as well,
- or something else

cory-skeers:
Ready for review, though some work on documentation still needs to be done.

joewhitsitt:
I've been a bit disconnected from the work thus far (out sick, loopy). At the time of grooming, I had not considered the potential of that many jobs listed at once. Thought that was a handshake issue only and that we were instead only dealing with a couple jobs that were posted over a year ago. I forget whether HR was planning to have a landing pages for these filtered routes for these someday that we could link off to. That would help steer my thoughts about this.

For aggregator we are limiting to a year (solves the handshake issue). Maybe for this block we limit the number displayed (50)

cory-skeers:
> Maybe for this block we limit the number displayed (50)

I like that as an option, and have included it, as well as updating the "# to display" field's help text to match.

Most uses of the block should be using a department filter and not run into the issue anyway, but I like having a bit of a cover just in case someone tries to use this for the full feed, or if a non-department filter gets used that ends up with more results.

briand44:
hmmm... My first instinct was to limit it to ...faculty.php as I was thinking of this specifically for faculty jobs as my understanding is those are the ones that have a difficult process to revise the post date.

cory-skeers:
> hmmm... My first instinct was to limit it to ...faculty.php as I was thinking of this specifically for faculty jobs as my understanding is those are the ones that have a difficult process to revise the post date.

It was slightly lighter to allow for all of the hris.uiowa.edu/otac-rss/X.php job feeds, but it'd be a pretty small change to limit it to `faculty.php` (and update help text). We're already validating each part prior to that, so it's just one more check.

We'd probably still want a limit just to cover our bases though.

If we limit to faculty, should we also update the block title to something like "Faculty Jobs?"

config/default/field.field.block_content.uiowa_jobs.field_uiowa_jobs_feed_url.yml
```diff
@@ -0,0 +1,21 @@
...
+    - block_content.type.uiowa_jobs
+    - field.storage.block_content.field_uiowa_jobs_feed_url
+_core:
+  default_config_hash: IRpf3Xmvg2RZqoJ0IJ2FwE7a4jA0mVZnEhIrxIbCfGE
+id: block_content.uiowa_jobs.field_uiowa_jobs_feed_url
+field_name: field_uiowa_jobs_feed_url
+entity_type: block_content
+bundle: uiowa_jobs
+label: 'Feed URL'
+description: 'RSS feed URL from hris.uiowa.edu (e.g., https://hris.uiowa.edu/otac-rss/faculty.php?dept-code=DEPT-NUMBER)'
```
cory-skeers:
```suggestion
description: 'RSS feed URL from hris.uiowa.edu (e.g., https://hris.uiowa.edu/otac-rss/faculty.php?dept-code=DEPT-NUMBER). Note: The format=json and format=xml options are not currently supported.'
```

joewhitsitt:
pulling to test

joewhitsitt:
- What does the align button to bottom style do?
- Should the HRIS validation happen on block submit? Placed a handshake feed and it just says unable to display on render.
- smart_trim should be added to the .info file as a dependency.
- Use `t()` for some of these strings? `Feed URL is not configured.` `No job postings are currently available.` `Posted {$formatted_date}` `Unable to load job postings. Please try again later.`
- not sure how to fix it off the top of my head but the pager system if multiple blocks is fragile. e.g. changing the page of block 1, resets the pager of block 2. - _realized we have similar code in sitenow_events..._
- Undefined var here:
Screenshot 2026-05-26 at 2 19 39 PM

docroot/modules/custom/uiowa_jobs/src/RssFeedParser.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $html
+   *   The HTML content.
+   * @param int $trim_length
+   *   The max length before trimming.
+   *
+   * @return string
+   *   Cleaned HTML.
+   */
+  protected function cleanHtml(string $html, int $trim_length): string {
```
joewhitsitt, line 189: Not sure if it is worth to use uiowa_core_html_cleanup/uiowa_core_html_empty_after_cleanup or if the allowed could be converted to an array `strip_tags($html, ['p', 'br', 'strong', 'b', 'em', 'ul', 'ol', 'li']);`

docroot/modules/custom/uiowa_jobs/uiowa_jobs.module
```diff
@@ -0,0 +1,257 @@
...
+  // Add validation to the feed URL field on uiowa_jobs block forms.
+  if ($form_id === 'layout_builder_add_block' || $form_id === 'layout_builder_update_block') {
+    if (isset($form['settings']['block_form']['field_uiowa_jobs_feed_url'])) {
+      // Add validation handler.
+      $form['settings']['block_form']['field_uiowa_jobs_feed_url']['widget'][0]['value']['#element_validate'][] = $validator;
+    }
+  }
+
+  // Also handle non-layout-builder forms.
+  if (strpos($form_id, 'block_content_uiowa_jobs_') === 0) {
```
joewhitsitt:
```suggestion
  if (str_starts_with($form_id, 'block_content_uiowa_jobs_')) {
```

IDE recommendation

joewhitsitt:
Curious if we can make use of our `time:hourly` cache tag instead of these time based rules?

cory-skeers:
> * What does the align button to bottom style do?

I don't think it does anything. I started with matching aggregator's allowed styles and whittled down, and probably missed removing this one.

> * Should the HRIS validation happen on block submit? Placed a handshake feed and it just says unable to display on render.
> * smart_trim should be added to the .info file as a dependency.
> * Use `t()` for some of these strings? `Feed URL is not configured.` `No job postings are currently available.` `Posted {$formatted_date}` `Unable to load job postings. Please try again later.`

Good notes, and all can/should be updated.

> * not sure how to fix it off the top of my head but the pager system if multiple blocks is fragile. e.g. changing the page of block 1, resets the pager of block 2.

Dang, didn't think of multiple blocks. I think it involves adding a unique id to the wrapper, but I'll need to do a bit of digging.

> * Undefined var here:

Missed removing from an earlier iteration. That line can be removed completely.

> Curious if we can make use of our time:hourly cache tag instead of these time based rules?

Yes! I completely forgot about that tag.

docroot/modules/custom/uiowa_jobs/uiowa_jobs.module
```diff
@@ -0,0 +1,257 @@
...
+      // Add validation handler.
+      $form['settings']['block_form']['field_uiowa_jobs_feed_url']['widget'][0]['value']['#element_validate'][] = $validator;
+    }
+  }
+
+  // Also handle non-layout-builder forms.
+  if (str_starts_with($form_id, 'block_content_uiowa_jobs_')) {
+    if (isset($form['field_uiowa_jobs_feed_url'])) {
+      $form['field_uiowa_jobs_feed_url']['widget'][0]['value']['#element_validate'][] = $validator;
+      $form['field_uiowa_jobs_feed_url']['widget'][0]['value']['#description'] = $help_text;
```
cory-skeers:
```suggestion
```

joewhitsitt:
> > * not sure how to fix it off the top of my head but the pager system if multiple blocks is fragile. e.g. changing the page of block 1, resets the pager of block 2.
> 
> Dang, didn't think of multiple blocks. I think it involves adding a unique id to the wrapper, but I'll need to do a bit of digging.

Just mentioned I saw similar code in sitenow_events. We haven't solved this and I don't know if it should be a blocker for this PR but perhaps a follow up issue to make both better in the future.

joewhitsitt:
All caps was brought up recently in design refresh. Wonder if they can adjust this upstream?

Screenshot 2026-05-26 at 3 08 41 PM

@briand44 is this something we could raise with HR?

config/default/layout_builder_styles.style.card_style_button_position.yml
```diff
@@ -10,5 +10,6 @@ group: default
 block_restrictions:
   - 'inline_block:uiowa_card'
   - 'inline_block:uiowa_aggregator'
+  - 'inline_block:uiowa_jobs'
```
cory-skeers:
```suggestion
```

briand44:
>All caps was brought up recently in design refresh. Wonder if they can adjust this upstream?

I reached out to see if this could be addressed in the feed. Will follow up once I hear back.

cory-skeers:
@joewhitsitt I believe each of the previous review notes have been addressed, other than the multiple pager issue.

- Validation moved to form submit to actually prevent including an unsupported feed
- `t()` wrapped user-facing strings
- switched to `time:hourly` cache tag
- Code cleanup

joewhitsitt:
pulling down again

briand44:
Just making it visible that the limit of 50 could become an issue for medicine.uiowa.edu in the future. Looks like there are 105 open faculty positions for that org currently https://jobs.uiowa.edu/jobSearch/faculty/searchResults.php?submit=Search+For+Openings&searchType=FACULTY&org=17. I think Max will plan to display 50 and link off and wait to see if it becomes an issue. From a department perspective (which is where this request was originating from), their largest department currently only has ~30 so should be good there.

So we are ok for now but could get requests to increase in the future.

joewhitsitt:
Wonder how much we can punt on this by indicating "showing only the first 50 results. view all at..." more link.

docroot/modules/custom/uiowa_jobs/uiowa_jobs.module
```diff
@@ -0,0 +1,272 @@
...
+        '#element' => $element,
+        '#quantity' => 3,
+      ];
+    }
+
+    // Add cache tags.
+    $cache_id = 'uiowa_jobs:feed:' . md5($feed_url);
+    $variables['#cache']['tags'] = Cache::mergeTags(
+      $variables['#cache']['tags'] ?? [],
+      ['uiowa_jobs:feed', $cache_id, 'time:hourly']
```
joewhitsitt:
```suggestion
      ['uiowa_jobs:feed', 'time:hourly']
```

I don't think the `$cache_id` works this way and would never be invalidating through this.

So this and the line 188 can be removed.

joewhitsitt: sorry i don't know how to do a multiline suggestion

joewhitsitt (approved):
Everything else looks fine to me. Thanks for making those adjustments.

docroot/modules/custom/uiowa_jobs/uiowa_jobs.module
```diff
@@ -0,0 +1,272 @@
...
+    if ($show_pager) {
+      $variables['content']['pager'] = [
+        '#type' => 'pager',
+        '#element' => $element,
+        '#quantity' => 3,
+      ];
+    }
+
+    // Add cache tags.
+    $cache_id = 'uiowa_jobs:feed:' . md5($feed_url);
```
cory-skeers:
```suggestion
```

docroot/modules/custom/uiowa_jobs/uiowa_jobs.module
cory-skeers: Thanks! I added a separate suggestion/commit for the line 188 removal.

PR #9840, opened 2026-05-19
Update setasign/fpdi

pyrello:
## Summary
- Bumps `setasign/fpdi` v2.6.4 → v2.6.7 (transitive via `iio/libmergepdf`).
- Three patch-level upstream releases; no API changes. PHP 8.3 fully supported.

## Test plan
- [ ] Composer install / autoload regenerate clean
- [ ] Regenerate the Policy Manual PDF on `policy.uiowa.edu` (admin → Policy Manual download → Regenerate) and confirm the merged PDF builds and opens

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -19743,27 +19743,27 @@
         },
         {
             "name": "setasign/fpdi",
-            "version": "v2.6.4",
+            "version": "v2.6.7",
             "source": {
                 "type": "git",
                 "url": "https://github.com/Setasign/FPDI.git",
-                "reference": "4b53852fde2734ec6a07e458a085db627c60eada"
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
Downloaded the current PDF from prod to compare against. The local version still generates as expected. Not seeing anything stand out in terms of changes in formatting/size, but the prod version was last created 2025-11-10 so there have been additions/edits since so the doc starts to not line up after page 27 and continues past the old version's 985 pages to 993. The last few pages are the same so it realigns at the end.

PR #9870, opened 2026-05-26
Remove ability to download pdf from content overview actions

### LINKED ISSUE #8774: 'Download PDF' visible but does not work
User in CLAS reported they tried to use the Download PDF option from the actions list on the content overview page. I did some testing, only admins can use it and I think it's because that option originates from the Entity Print module. Mike found a workaround for the user, I just wanted to check if this was an option we wanted to add in the future, or something to hide from users, or simply a wontfix. 

2025/05/15 SAH - Probably wontfix for now if it doesn't just work.

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/8774
User impact: Editors will no longer be teased with bulk operations they don't have permissions for.

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu && ddev drush @sandbox.local uli --uid=111 admin/content
```

As a webmaster, select a piece of content (checkbox) from the content overview and try to see the option to download PDF. you can't. Same for promoting to front page. you can't.

Where entity print is being used, it is still possible to print as PDF outside the content overview:

```
ddev blt ds --site=admissions.uiowa.edu
```

Go to /kiosk and try to print one of the areas of study.


GaryRidgway:
I can pull to test

GaryRidgway (approved):
Tested and works as instructions described!

Issue #9745, opened 2026-04-09
Filter errors that get sent to New Relic


joewhitsitt:
We tried to fix it ourselves unsuccessfully https://github.com/uiowa/uiowa/pull/8559

We had previously created an [acquia ticket](https://acquia.my.site.com/s/case/500Pb00000h2dxUIAQ/reduce-noise-sent-to-new-relic-apm) because #drupal-status was being hit hard with `Symfony\Component\HttpKernel\Exception\NotFoundHttpException` `error.class` messages among a few others. They acknowledged the issue at the time but didn't have a solution for us. Now there is one.

This support article has been updated with a possible solution to reduce noise in our errors inbox in new relic.
https://acquia.my.site.com/s/article/Upgrading-New-Relic-PHP-agent-version-for-Acquia-Cloud-Platform

# Proposed Solution

Try implementing this service in uiowa_core to see if it reduces the noise. Relates to https://github.com/uiowa/uiowa/issues/6242 and https://github.com/uiowa/uiowa/issues/7152

# Follow-up

Try turning our new relic to slack webhook so that we can see actual errors as they happen in #drupal-status

briand44:
## Grooming
 
**Quadrant:** 

Complex

 **Discussion:**
 - Way to filter out some of this common but unique items now
 - Would allow us to keep an eye on other issues (out of memory errors, etc.) and be more proactive
 - Slack webhook may have been written by Rick, likely have to disable and replace - can be future slice
 - Fire Slack webhook for a single application to monitor prior to enabling for all - can be future slice
 - Some information in our docs around setting up a new application
 - implement in uiowa_core
 - you can login to new relic and see the errors inbox currently.
   - Here is uiowa.prod’s error inbox: https://one.newrelic.com/nr1-core/errors-inbox/entity-inbox/MTY2MDk5MnxBUE18QVBQTElDQVRJT058NTU0NDczOTE?account=1660992&duration=1800000&filters=selectedInstance%20IN%20%28%29&state=7c14a799-3548-47f3-8eaf-6e4f2af15316
   - Each application has separate errors inbox  
 - Scope to filtering out HTTPException errors - webhook can come later after refinements
 - Further refining can happen in future slices

**Proposed solution or definition of done:**
- Filter HTTPException errors across all applications in New Relic
- Confirm less noise in each application errors inbox
- Document any additional refinements we may want in future

joewhitsitt:
[...]

PR #9886, opened 2026-05-29
[New Relic APM] Implement suggested filtering proposed by Acquia

joewhitsitt:
Relates https://github.com/uiowa/uiowa/issues/9745
User impact: Developer experience improvement allowing better use of the New Relic tool without all the noise

# How to test

- Because of `extension_loaded('newrelic')` this won't be testable locally and is something we will need to guess and check out in prod.
- Sync a website locally and navigate around the front and back to make sure routes don't error.
- Review implementation based on https://acquia.my.site.com/s/article/Upgrading-New-Relic-PHP-agent-version-for-Acquia-Cloud-Platform

# Follow-up

- Deploy to PROD and see if we get a noticable drop in noise to the New Relic errors inbox.
- See if other errors should be filtered out.
- Try sending slack notifications again to #drupal-status


cory-skeers (approved):
Checked around on a site locally and didn't run into any route issues.

Reviewed and this looks good--matches the Acquia-suggested implementation, and from my understanding of how this _should_ work, hopefully can get things cleaned up.

Issue #9858, opened 2026-05-22
Finalize BLT replacement (cleanup + replace BLT settings cascade)


pyrello:
# Motivation

All site settings.php files have a `require` line that references a BLT include file from the `vendor` directory. To remove BLT as a dependency, we need to replace that. As part of #9789, we discovered that there is an Acquia replacement: https://github.com/acquia/drupal-recommended-settings (DRS) which covers what BLT was doing for us with that settings.php `require`.

# Important note

DRS is not able to be installed alongside BLT. This is prevented in its composer.json file:
https://github.com/acquia/drupal-recommended-settings/blob/50b561bd97b68df4b351dc2a24dd950a288856b8/composer.json#L38-L40

This means that DRS has to be added and implemented in the same PR as the removal of BLT.

# Proposed solution
- Package with `acquia/blt` removal.
- Install `acquia/drupal-recommended-settings`.
- Migrate existing settings.php file to update the `require` line to whatever DRS needs.

pyrello:
Blocked until all other BLT replacement tasks are completed.

joewhitsitt:
Postponing refinement of this issue until more team members are back to make sure we are including all the remaining pieces with this slice.

pyrello:
Adding to the cleanup scope: a comment pass over the `sitenow/` code the earlier `./sn` migrations landed, against the criteria in #10055.

Those PRs established the code but not the comment conventions, so the same three problems #10055 was reviewed for are already on main.

**1. Backward-looking BLT references.** Stale now. `MultisiteExecuteCommand.php:20`, `SiteInstallCommand.php:21`, `:79`, `:86`, `:639`, `:735`, `SiteSyncCommand.php:150`, `:166`, `ClassifiesInstalls.php:116`.

```
git grep -nE '^ *(\*|//).*(\bum[dce]\b|BLT.s |Replaces |Carried over)' -- sitenow/src
```

Returns 9, all real.

**2. Scope, altitude, and encoded decision-making.** Not greppable, needs a read of each file. Examples:

* `SiteInstallCommand.php:26` pulls the per-state install behavior and its idempotence argument up into the class docblock.
* `MultisiteInstallCommand.php:26` justifies the install-state check by contrasting it with the approach not taken ("rather than asking only whether a config table exists"), and argues for per-site log files at class level.
* `SiteSyncCommand.php:18` lists the sync steps in the class docblock, where the numbered inline comments in the method already carry them.

[...]

PR #9913, opened 2026-06-09
Testing the things

cory-skeers:
# How to test

```
git pull &&
ddev composer install
```
If you haven't recently,
```
ddev blt bis
```
If you don't, you'll likely fail based on sites that have been deprovisioned but still exist on your local machine. After running the command, check for any leftover sites that still exist in your local directory (you can run a `git status` and they'll show up in the untracked files. `rm -rf docroot/sites/**thesite.url**` for each that's in there)

```
ddev ci
```
See it do go through the testing process.

Try breaking something small (like a `phpcs` error)
```
ddev ci --fast-fail
```
See it pop out early once it hits a failure

Try the individual tests:

- `ddev ci lint` (`ddev ci phpcs`)
- `ddev ci static-analysis` (`ddev ci phpstan`)
- `ddev ci test` (`ddev ci phpunit`)

See a successful Travis build.
See a successful Github Actions build.

## Before deployment

Make sure we're confident in the Travis CI update. If we want more testing, that can be removed for now and the scripts for local testing kept.
Also decide if we want the GitHub Actions portion. Similarly, that can be removed for now and added any time later.


.travis.yml
```diff
@@ -2,41 +2,72 @@ language: php
...
-- phpenv config-rm xdebug.ini
+  # Setup SSH keys for deployment
+  - openssl aes-256-cbc -K $encrypted_189e52c2c347_key -iv $encrypted_189e52c2c347_iv
+    -in deploy_key.enc -out deploy_key -d
+  - mv deploy_key $HOME/.ssh/id_rsa
+  - chmod 600 $HOME/.ssh/id_rsa
+
+  # PHP configuration
+  - phpenv config-rm xdebug.ini
+  - yes | pecl install memcache-8.0
```
cory-skeers:
```suggestion
```

cory-skeers:
I think the previous tests and files analyzed counts not matching what's current in production has been tracked down. Claude's initial analysis missed the `tests:deprecated` section, and so was seeing these changes as adding new coverage. After a newer comparison, I'm seeing the same coverage except for the added `sitenow/src` path coverage in this PR.

scripts/ci/run-all.sh
```diff
@@ -0,0 +1,117 @@
...
+
+set -e  # Exit on any error
+set -u  # Exit on undefined variable
+
+# Load shared color codes
+source "$(dirname "$0")/colors.sh"
+
+echo -e "${BLUE}╔════════════════════════════════════════════╗${NC}"
+echo -e "${BLUE}║   Running Full CI Pipeline Locally        ║${NC}"
+echo -e "${BLUE}╚════════════════════════════════════════════╝${NC}\n"
```
cory-skeers, line 16:
```suggestion
echo -e "${BLUE}╔════════════════════════════════════════════╗${NC}"
echo -e "${BLUE}║   Running Full CI Pipeline Locally         ║${NC}"
echo -e "${BLUE}╚════════════════════════════════════════════╝${NC}\n"
```

cory-skeers:
Not total parity going on here, as the Travis build is passing while others aren't (security vulnerability check missing).

cory-skeers:
> Not total parity going on here, as the Travis build is passing while others aren't (security vulnerability check missing).

Updated:
image

Adding a separate security script is technically unnecessary (essentially it's a wrapper for `composer audit`), but in the same way as calling the `blt tests:security-drupal` in the pipeline worked, it keeps these checks separate as their own thing and readily callable from other commands.

pyrello:
Going to start poking at this.

pyrello:
Running into a couple issues on the local runner:

> 2) Uiowa\Tests\PHPUnit\Unit\FilesystemTest::testMultisiteFiles
> Failed asserting that file "/var/www/html/docroot/../docroot/sites/alt.clas.uiowa.edu/blt.yml" exists.

Not seeing that sites dir. Not sure if its supposed to be there or not.

> There was 1 skipped test:
> 
> 1) Drupal\Tests\uiowa_search\FunctionalJavascript\SearchTest::testSearchInputVisibleAfterClickingSearchButton
> The test wasn't able to connect to your webdriver instance. For more information read core/tests/README.md.
> 
> The original message while starting Mink: Could not open connection: session not created: Chrome failed to start: exited normally.
>   (session not created: DevToolsActivePort file doesn't exist)
>   (The process started from chrome location /opt/google/chrome/chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
>   (Driver info: chromedriver=123.0.6312.58 (6b4b19e9dfbb93aa414dc045bd445287977d8d7a-refs/branch-heads/6312_46@{#3}),platform=Linux 6.8.0-64-generic x86_64)

It's been long enough that I've forgotten how to resolve a chromedriver failure like this.

Ran the following to get to this point:
```
git checkout testing-update
git pull
ddev composer install
ddev blt bis
ddev ci
```

cory-skeers:
> Running into a couple issues on the local runner:
> 
> > 2. Uiowa\Tests\PHPUnit\Unit\FilesystemTest::testMultisiteFiles
> >    Failed asserting that file "/var/www/html/docroot/../docroot/sites/alt.clas.uiowa.edu/blt.yml" exists.
> 
> Not seeing that sites dir. Not sure if its supposed to be there or not.

I think there's an extra step here to remove any site directory folders from your local after running the `ddev blt bis`. Looks like the testMultisiteFiles gets its list of "all sites" based on the file directory, so it'll still see `alt.clas.uiowa.edu` as a valid site and check that everything else is in place for it if it still exists in the directory.

pyrello:
Your commit flagging the uiowa_auth SSH→HTTPS switch said "CHECK THIS." Can you walk me through it? What drove the change?

cory-skeers:
> Your commit flagging the uiowa_auth SSH→HTTPS switch said "CHECK THIS." Can you walk me through it? What drove the change?

Good catch! I'd lost track of that.

The default setup for github actions works with https so it can use a temporary auth token. It can be changed to ssh with a dedicated key, but I think it needs to be consistent across everything in either https for all or ssh for all.

That commit changed the `uiowa_auth` designation in composer to use https instead of ssh as the only outlier that didn't fit the https access already and wasn't able to follow the rewrite rules in github actions because it's specified directly in composer.

There is probably a way to still handle it without changing it from ssh across the board, but I hadn't found a way. If we aren't implementing GHA right away, we can revert the `uiowa_auth` composer change here and tackle that problem later.

pyrello:
Overall this is a solid move toward provider-portable CI. The `scripts/ci/*.sh` + thin-caller pattern (Travis, GHA, and `ddev ci` all calling the same scripts) is the right shape, and it sets conventions the deploy rework (#9917) can reuse. Comments below, grouped by what I think should block merge vs. what can land and iterate.

**Before merge**

1. ~`composer.json` `uiowa_auth` SSH→HTTPS is a committed change, so it applies to every environment, not just GHA. I've asked Cory about the reasoning separately; holding on that answer.~ - This has been addressed.
2. `composer audit` is wired as a blocking gate. It fails on any advisory in any dependency, including transitive and dev, so the build can go red for reasons unrelated to a given PR. We should decide: blocking or advisory. 
3. Travis `install:` drops `setup_project` and `setup_environment`, but `deploy:` still uses BLT. Confirm deploy relies on nothing those steps produced.

**Non-blocking**

- `settings.php`: the CI fork around the BLT cascade reads as bridge code until #9858. Add a comment referencing #9858 so it's a clear delete-target, and confirm no production context exports `CI` (the GHA branch keys on it).
- GHA workflow: the auth + cache + install steps are duplicated across all four jobs. A composite action or reusable workflow would consolidate them.
- `run-all.sh`: step labels are inconsistent (`2/4` then `3/5` through `5/5`), and the `shift` calls inside `for arg in "$@"` are no-ops.
- `lint.sh` Twig check is a regex for nested delimiters, not real validation. Low value, and may mislead.
- `static-analysis.sh` runs PHPStan once per path with separate bootstraps. A single invocation would be faster.
- GHA `frontend` job uploads artifacts nothing consumes, and triggers on a `develop` branch I don't think exists.
- `phpunit.xml.dist` defaults now hardcode DDEV values. Fine given DDEV-required, just noting it.

**Nice catch**

`sitenow.profile`: the `moduleExists('uiowa_core')` guard in `sitenow_form_alter` fixes a latent install-time bug (the alter fires during `site:install` before `uiowa_core` is enabled), independent of CI.

---
Drafted by Claude on behalf of @pyrello.

pyrello:
> > Your commit flagging the uiowa_auth SSH→HTTPS switch said "CHECK THIS." Can you walk me through it? What drove the change?
> 
> Good catch! I'd lost track of that.
> 
> The default setup for github actions works with https so it can use a temporary auth token. It can be changed to ssh with a dedicated key, but I think it needs to be consistent across everything in either https for all or ssh for all.
> 
> That commit changed the `uiowa_auth` designation in composer to use https instead of ssh as the only outlier that didn't fit the https access already and wasn't able to follow the rewrite rules in github actions because it's specified directly in composer.
> 
> There is probably a way to still handle it without changing it from ssh across the board, but I hadn't found a way. If we aren't implementing GHA right away, we can revert the `uiowa_auth` composer change here and tackle that problem later.

Got it. I think the change makes sense as it aligns with how we are pulling other packages. Its a public repo, so we should just be able to access it without auth or a token, I would think.

pyrello:
Can you explain how we would use `--fast-fail` in practice?

cory-skeers:
> Can you explain how we would use `--fast-fail` in practice?

It was originally added thinking it might reorder things, or have further changes to the process to catch and fail early if there's an issue. But the ordering is already kind of there, so it ended up being smaller than thought.

Travis and GHA both exit early on failure, so I think the only remainder is if you're running the full `ci` locally, the `--fail-fast` option will end it early as soon as there's an error rather than continuing on. So if there's a PHPCS error it won't continue on with the PHPStan, and then Tests, like it would without that option.

pyrello:
> > Can you explain how we would use `--fast-fail` in practice?
> 
> It was originally added thinking it might reorder things, or have further changes to the process to catch and fail early if there's an issue. But the ordering is already kind of there, so it ended up being smaller than thought.
> 
> Travis and GHA both exit early on failure, so I think the only remainder is if you're running the full `ci` locally, the `--fail-fast` option will end it early as soon as there's an error rather than continuing on. So if there's a PHPCS error it won't continue on with the PHPStan, and then Tests, like it would without that option.

Got it. I guess I'm wondering if we don't just want the fast failure option all the time? I'm trying to imagine the scenario where you want to let it keep running.

joewhitsitt:
I remember when we started to work on php or drupal 10 we needed to update coding standards for things. We would get a failure of something to fix only to find out on the next run, it identified additional issues in the next directory. I could the see the use of that continuing on so we have a better picture of all the things we need to fix ahead of fixing them.

cory-skeers:
> > > Can you explain how we would use `--fast-fail` in practice?
> > 
> > 
> > It was originally added thinking it might reorder things, or have further changes to the process to catch and fail early if there's an issue. But the ordering is already kind of there, so it ended up being smaller than thought.
> > Travis and GHA both exit early on failure, so I think the only remainder is if you're running the full `ci` locally, the `--fail-fast` option will end it early as soon as there's an error rather than continuing on. So if there's a PHPCS error it won't continue on with the PHPStan, and then Tests, like it would without that option.
> 
> Got it. I guess I'm wondering if we don't just want the fast failure option all the time? I'm trying to imagine the scenario where you want to let it keep running.

Good point...I could see wanting to run other checks even if one fails, but that can still and probably would be done with the individual checks anyway. You don't lose that option with the full run exiting early. It'd just be possibly running several tests or selectively running scripts rather than `ddev ci` every time, which is like what we're doing now.

It'd also give that extra little parity with how external builds/tests get run, as they already "fail fast" if they hit an error.

cory-skeers:
> I remember when we started to work on php or drupal 10 we needed to update coding standards for things. We would get a failure of something to fix only to find out on the next run, it identified additional issues in the next directory. I could the see the use of that continuing on so we have a better picture of all the things we need to fix ahead of fixing them.

Not having it locally, I don't remember if the fail fast mode exits _immediately_, or at the end of each chunk of things...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,293 @@
...
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
+          extensions: gd, pdo_mysql
+          coverage: none
+
+      - name: Configure Git and Composer authentication
+        run: |
+          # Configure git to rewrite SSH URLs to HTTPS (must be done BEFORE cache)
```
pyrello:
```suggestion
          # Configure git to rewrite SSH URLs to HTTPS (must be done BEFORE cache).
          # Needed for a few packages that use SSH source for fallback.
          # Not used in the normal --prefer-dist path; this only guards the rare dist-download fallback
          # to `git clone` so CI doesn't fail on a missing SSH key for an unrelated dependency.
```
I had to have Claude explain this to me so I thought a more robust comment would be a good idea.

pyrello:
> > I remember when we started to work on php or drupal 10 we needed to update coding standards for things. We would get a failure of something to fix only to find out on the next run, it identified additional issues in the next directory. I could the see the use of that continuing on so we have a better picture of all the things we need to fix ahead of fixing them.
> 
> Not having it locally, I don't remember if the fail fast mode exits _immediately_, or at the end of each chunk of things...

Maybe the model here is to make the fail fast behavior the default? If we finding out from CI what we need to fix locally for a coding standards update, that is likely a workflow problem (we should probably be running the standard check locally to fix coding standards rather than triggering a 15-30 minute build process).

I guess it's a good question to be explicit about. What behavior do we want for our CI? Fail fast (we learn there is a problem quickly) or fail completely (we find out everything that failed)? My immediate lean is towards the former, but I'm going to sit with it until tomorrow.

pyrello:
> 2. `composer audit` is wired as a blocking gate. It fails on any advisory in any dependency, including transitive and dev, so the build can go red for reasons unrelated to a given PR. We should decide: blocking or advisory.

Re-reading this now, I'm realizing that this being a blocking gate is consistent with how we *usually* run CI. Security advisories can block the build. Currently we also have a mechanism for turning this off. Once we remove the BLT plumbing, we won't have that anymore. Is that going to be a problem?

cory-skeers:
>> 2. composer audit is wired as a blocking gate. It fails on any advisory in any dependency, including transitive and dev, so the build can go red for reasons unrelated to a given PR. We should decide: blocking or advisory.

> Re-reading this now, I'm realizing that this being a blocking gate is consistent with how we _usually_ run CI. Security advisories can block the build. Currently we also have a mechanism for turning this off. Once we remove the BLT plumbing, we won't have that anymore. Is that going to be a problem?

Ignoring specific, individual advisories in composer to pass the gate should still be an option here, and the full fallback of temporarily commenting out the security-check portion of the run-all command. The latter shouldn't ever be needed though, since the former should be respected.

pyrello (approved):
I'm satisfied with this. Nice work.

PR #9916, opened 2026-06-10
Sprint 315 provisions

pyrello:
Provisions for sprint 315.

# How to test

Code review.


e-marie-w (approved):

briand44:
just noting there are some deprovisions that could happen https://github.com/uiowa/uiowa/issues?q=state%3Aopen%20label%3Adeprovision. They don't need to happen but I currently am not prioritizing those. If I need to prioritize I can but was leaving those off the board with the thought that support checks as part of process.

PR #9920, opened 2026-06-11
Faster Split Check

GaryRidgway:
I worked with claude to batch and multiplex the split checks, reducing the amount of ssh handshakes that have to happen with aquia. this is just a POC.

Do not merge.


pyrello:
@GaryRidgway Is this still relevant after #9975?

PR #9871, opened 2026-05-26
Replace `uiowa:multisite:create` with Symfony Console `sn multisite:create` (#9789)

pyrello:
Relates to #9789

Replaces BLT's `uiowa:multisite:create` with a standalone host-side Symfony Console command, `multisite:create` (aliases `mc`, `umc`), at `sitenow/src/Command/MultisiteCreateCommand.php`, run via `./sn`
(direction set by #9895). Establishes the SiteNow CLI home (`Command/`, `Operation/`, `Plan/`, `Config/` under `sitenow/src/`, with the `./sn` entry point at the repo root) that later command ports will build on. 42 unit tests cover the pure logic.

The headline new capability is auto-selection. The command picks the target Acquia application automatically (formerly the developer chose interactively from a list). Auto-selection chooses the eligible application with the fewest sites that has SSL coverage for the host, breaking ties by application name. `--app=<name>` allows overriding auto-selection.

The command uses `sitenow/applications.yml` and the existing `manifest.yml` to enumerate applications and get site counts (a proxy for DB count). SSL coverage remains a live Acquia API call.

Existing tests were updated to work with the new command and new unit tests were added to pin required behavior.

Note: the old BLT `MultisiteCommands::create()` will be removed or marked deprecated in a follow-up PR once this has deployed.

## What changed since the Robo version of this branch
- The command moves `sitenow/src/Robo/Plugin/Commands/` → `sitenow/src/Command/` and shows as a rename with edits; the `Plan`/`Check`/`Config` layers and the unit tests carry forward largely unchanged.
- The Robo `Task/` classes are replaced by plain `sitenow/src/Operation/` classes; bundled Robo tasks become Symfony Filesystem + Process calls.
- New entry point `./sn` (repo root) replaces `./vendor/bin/robo` invocation.
- `--simulate` is gone (a Robo built-in); `--dry-run` is the preview.

# How to test
- Code review
- Valid example (`--dry-run` renders the plan and stops):
```
./sn mc test.sites.uiowa.edu --dry-run
```
- Explicit `--app` override:
```
./sn mc test.sites.uiowa.edu --app=uiowa03 --dry-run
```
- Reasoning reads `Explicitly specified via --app.`

## Validation behavior
The command runs preconditions before any work (environment, input, registry collisions, SSL coverage, git state). A FAIL stops the plan and shows the offending check inline; a WARN proceeds with confirmation or aborts under `--yes`.

- Unknown application (registry as allowlist):
```
[...]

PR #9926, opened 2026-06-15
Robo to Symfony console conversion

### LINKED ISSUE #9906: Refactor existing Robo commands using Symfony Console
# Motivation

After some investigation in https://github.com/uiowa/uiowa/issues/9895#issuecomment-4635421108, it was decided to use Symfony Console rather than Robo as our BLT replacement so we should replace any existing commands relying on Robo.

# Additional Info

- What all needs addressed as part of this slice?
  - ReportCommands were mentioned
  - What else?

GaryRidgway:
# How to test 
_Commands drafted by claude code, instructions drafted by me._

## Operator-facing changes from the old Robo commands:

- Commands are renamed: uiowa:report:domains/splits/inactive → report:domains/splits/inactive (aliases domains/splits/inactive), run via ./sn instead of ./vendor/bin/robo.
- report:domains drops the --debug option (it only added per-domain "found X" chatter; the table/CSV output covers the same ground).
- Progress output moved from stdout to stderr so stdout is clean for piping/redirecting the data.

## Tests - _(You can find the code for the tests in tests/phpunit/src/Unit)_

**ALL COMMANDS SHOULD BE RUN FROM UIOWA, NOT DOCROOT.**

1. Run unit tests with the following command:

```
ddev exec './vendor/bin/phpunit --filter "FleetDomains|ReportSplits|ReportInactive|CsvWriter" tests/phpunit/src/Unit/'
```

2. Ensure everything comes out green - (Ignore the HTML output error, that is from something out of scope.)
3. Run phpcs tests on this PR's files

```
ddev exec './vendor/bin/phpcs --standard=AcquiaDrupalStrict sitenow/src/Command sitenow/src/Report sitenow/src/Traits tests/phpunit/src/Unit/{FleetDomains,ReportSplits,ReportInactive,CsvWriter}Test.php'
```

4. If you see no output, that means that there were no errors or warnings.

## Command testing

_(You can check for output parity with these commands against the commands on main for bonus points)_

Progress lines (`Processing…`, `Checking…`, `Fetching…`) now print to **stderr** instead of stdout, keeping stdout clean for the data. They won't show up in a piped/redirected diff.

I recommend starting the slower commands in two terminals and then letting them run while running the faster commands.

### Slower testing

1. Run site `splits` detection across 09 for the `additional_person_types` split and see that it does not error. _(For some reason I had to run this twice, but it worked the second time — a single unresponsive site aborts the whole app and exits non-zero by design. re-running usually clears it.)_

```
ddev exec ./sn report:splits --apps=uiowa09 --split=additional_person_types
```

2. Run `inactive` site detection across 09 and see that it does not error, and outputs a table of sites that shows if they are active or not.
    1. **NOTE**: "Days Since…" columns naturally drift between runs as time passes, so its output won't be byte-identical run to run. Compare the status column, not the day counts if you are comparing against main.

```
ddev exec ./sn report:inactive --apps=uiowa09
```

### Faster testing

1. Run `list` and ensure it lists all the commands.

```
./sn list
```

2. Run `routes:custom` and see it output a list of our custom routes.

```
./sn routes:custom
```

3. Run `domains` on 02 and check that it produces a list of prod domains correctly.

```
ddev exec ./sn report:domains --apps=uiowa02
```

4. Run `domains` with non-prod flags on 02 and check that it produces a list of non-prod domains correctly.

```
ddev exec ./sn report:domains --apps=uiowa02 --env=dev,test
```

5. Run `domains` on 02 with an export flag and check that it produces a CSV of prod domains at the root of the application. _(If the CSV looks header-only in your editor, give it a moment or check inside the container. The file can take a second to sync to the host.)_

```
ddev exec ./sn report:domains --apps=uiowa02 --export
```

6. Try to run the `domains` command outside of ddev, and verify that it errors and tells you the right command to run in the container

```
./sn report:domains
```

7. Break things?


pyrello (commented):
_This review was written by Claude (Opus 4.8). @pyrello reviewed it before posting._

Strong PR. Clean decomposition, well-commented, well-tested. One correctness issue to fix before merge; the rest are minor or nits.

## Verification

Ran on the branch (HEAD `414c1bc13`): the unit suite `phpunit --filter "FleetDomains|ReportSplits|ReportInactive|CsvWriter"` passes (29 tests, 40 assertions), and `phpcs --standard=AcquiaDrupalStrict` on the new `Command`/`Report`/`Traits` files and the four test files is clean. These are pure-logic unit tests; they do not exercise `DrushRunner` or the live command paths, so finding #1 below is reasoned from the Symfony Process contract, not reproduced by a test. The live command-parity checks from the testing instructions are being run separately by @pyrello.

## Blast radius

No CI, cloud-hook, or cron references to the old `uiowa:report:*` commands or `ddev robo`; they are operator-run manual tools, so the command rename and robo-wrapper removal are safe. The `compile:theme` mention only ever lived in the deleted wrapper's own comment. The separate BLT `ReportCommands` (`analytics`/`last-login`/`v1-standard-list`) is untouched. Robo retirement is cleanly scoped: the whole `sitenow/src/Robo/` tree, `RoboFile.php`, and the `ddev robo` wrapper are gone, with no dangling `SiteNow\Robo` imports left behind. BLT's own `Acquia\Blt\Robo\*` and the transitive `consolidation/robo` correctly stay.

## Findings

**1. (Correctness, should fix) `DrushRunner::run()` lets a timeout crash the whole run.** The 120s timeout makes Symfony `Process::run()` throw `ProcessTimedOutException` rather than return a non-zero exit. Callers only handle the non-zero-exit path (-> `FALSE` -> graceful per-app abort), so an unresponsive site throws an uncaught exception and a fatal stack trace that halts the entire command, not the clean "drop this app, continue, exit non-zero" the code is built for. This looks like the cause of the re-run note in the testing instructions. Suggested fix: catch `ProcessTimedOutException` (and a launch `RuntimeException`) inside `run()` and return a non-zero exit plus a stderr message, so the existing `FALSE` handling takes over.

**2. (Minor) Error output goes to stdout, not stderr.** `requireDdev`, `requireSshAgent`, `requireAcquiaClient`, and the splits missing-manifest error use `$io->error()` (stdout) while progress correctly uses `$err`. Slightly at odds with the "keep stdout clean" design. Low impact, since these paths return `FAILURE` and emit no data.

**3. (Minor, comments) Two class docblocks describe the Robo code they replace.** `CsvWriter` ("Replaces `SiteNowCommandsTrait::initializeCsvExport()`...") and `DrushRunner` ("...instead of the report commands' former `exec()`/`shell_exec()` shell strings"). Once Robo is gone these become dangling references. Comments read best when they stay declarative and describe current state. Suggest trimming each to its present-tense behavior (keep "no shell interpolation", "stdout/stderr kept separate so callers don't parse Acquia chatter", "timestamped CSV at the repo root") and moving the migration rationale to the PR description.

**4. (Nit, comments) Uncommented magic number.** `ReportSplitsCommand::getSplitStatuses()` uses `substr($n, 26)`, where 26 is the length of the `config_split.config_split.` prefix. A short comment or a named const would make that self-explaining. Carried over from the Robo original; the rewrite is a good moment to fix it.

**5. (Note) Command rename and dropped `--debug`.** `uiowa:report:*` becomes `report:*`, and `domains` drops its `--debug` option. Reasonable for a fresh CLI; worth an explicit line in the PR body so the change in operator muscle-memory is documented.

**6. (Nit) `CsvWriter` reopens the file handle per row.** Parity with the old code, but a single lifetime handle would be cleaner now that it is a dedicated class.

## Worth acknowledging

- Shell-string `exec`/`shell_exec` to Symfony Process array args removes the escaping and injection fragility, and drops the brittle "string-match the word error" heuristic in favor of exit codes and separated stderr.
- The pure parsers are genuinely unit-tested: chatter-stripping, uid 1 skip, pre-2000 login noise, and the error/exit cases.
- Good `requireX()` precondition helpers cut real repetition, and the comments focus on the non-obvious "why".

GaryRidgway:
Thanks for the review! added these.

pyrello:
> 1. Run `list` and ensure it lists all the commands.
> 
> 
> ```
> ddev exec ./sn list
> ```

> 9\. Run `routes:custom` and see it output a list of our custom routes.
> 
> 
> ```
> ddev exec ./sn routes:custom
> ```

Neither of this should require `ddev exec` to run. `routes:custom` just loops through the codebase and parses YAML.

pyrello:
> > 1. Run `list` and ensure it lists all the commands.
> > 
> > ```
> > ddev exec ./sn list
> > ```
> 
> > 9. Run `routes:custom` and see it output a list of our custom routes.
> > ```
> > ddev exec ./sn routes:custom
> > ```
> 
> Neither of this should require `ddev exec` to run. `routes:custom` just loops through the codebase and parses YAML.

My comment update probably didn't come through as a notification, so noting that `./sn list` should also not require `ddev exec`.

GaryRidgway:
> My comment update probably didn't come through as a notification, so noting that `./sn list` should also not require `ddev exec`.

Thank you for checking this! I think it was just a typo on my part for that, they do still work without running ddev exec.

sitenow/README.md
```diff
@@ -1,18 +1,41 @@
...
+| `report:domains` | `domains` | DDEV |
+| `report:splits` | `splits` | DDEV + SSH agent |
+| `report:inactive` | `inactive` | DDEV + SSH agent |
+| `routes:custom` | — | host or DDEV |
+
+The report commands run inside DDEV because they use drush aliases and reach
+prod sites over your forwarded SSH agent. Load your keys first with
+`ddev auth ssh`. A command run in the wrong environment exits with an error
+naming the correct invocation.
+
```
pyrello: Noting that I didn't include a reports list in this README in #9871 because its a drift surface and the command are self documenting with `./sn list`

sitenow/README.md
```diff
@@ -1,18 +1,41 @@
...
 - `sn` (repository root): entry point; registers commands.
 - `sitenow/src/Command/`: command classes.
-- `sitenow/src/Plan/`, `sitenow/src/Operation/`, `sitenow/src/Config/`: validation, file and API operations, and the application registry the commands use.
+- `sitenow/src/Report/`: report helpers (fleet-domain iteration, the drush
+  runner, the CSV export writer).
+- `sitenow/src/Traits/`: cross-command helpers (Acquia/drush/environment
+  helpers, option parsing).
+- `sitenow/src/Plan/`, `sitenow/src/Operation/`, `sitenow/src/Config/`:
+  validation, file and API operations, and the application registry used by
+  `multisite:create`.
```
pyrello: Same concern as above. I know that the structure portion got created here, but I'm wondering if its actually valuable for a human. Since its a drift vector, an AI agent shouldn't trust it anyways and instead should rely on reading the code. All of this is discernible from reading the code/directory structure.

GaryRidgway: I think these are both good observations, and I will remove things to simplify. One thing I am thinking though, the table for the commands explicitly lists which ones need to be in a ddev container, and `list` does not. should we be adding that to `list` somehow?

pyrello: I agree with this idea!

pyrello:
> I think these are both good observations, and I will remove things to simplify. One thing I am thinking though, the table for the commands explicitly lists which ones need to be in a ddev container, and `list` does not. should we be adding that to `list` somehow?

@GaryRidgway Just wondering if you are planning a fix for this? I'm waiting to approve until all the code changes are done.

GaryRidgway:
> > I think these are both good observations, and I will remove things to simplify. One thing I am thinking though, the table for the commands explicitly lists which ones need to be in a ddev container, and `list` does not. should we be adding that to `list` somehow?
> 
> @GaryRidgway Just wondering if you are planning a fix for this? I'm waiting to approve until all the code changes are done.

I am yes, working on it now!

GaryRidgway:
@pyrello  try that out. Pretty nifty!

pyrello (approved):
Claude reviewed code changes. I tested functionality and reviewed code changes. This looks good. Nice work!

PR #9933, opened 2026-06-17
bump drupal core to 10.5.12

joewhitsitt:
Changelog: https://git.drupalcode.org/project/drupal/-/compare/10.5.11...10.5.12?from_project_id=59858

# How to test

Sync a site, poke around.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -121,8 +121,8 @@
         "drupal/config_ignore": "^3.3",
         "drupal/config_inspector": "^2.1",
         "drupal/config_split": "^2.0",
-        "drupal/core-composer-scaffold": "~10.5.9",
-        "drupal/core-recommended": "~10.5.9",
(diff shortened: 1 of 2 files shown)


joewhitsitt:
Synced admissions without errors.

Tried uploading a vimeo oembed just cuz i saw mentions in the changelog. worked.

Update: found out that we are no longer affected by the oembed stuff because we use providers.json (for youtube/vimeo) and have a separate mechanism for panopto/ui capture. we used to be affected when we were doing mediaspace/kaltura and were using that patch and then later on oembed_providers module.

cory-skeers (approved):
Reviewed the security vulnerabilities this should address and the changelog, and it seems straightforward.

Synced a site, poking around and testing things, and haven't come across any issues.

I think this is good to go 👍

PR #9941, opened 2026-06-19
temp guzzle workaround

joewhitsitt:
# How to test

- Passing Travis CI Build
- Test a page that pulls UI Events, including a single event instance.
- Test https://recserv.uiowa.edu to see if hours info comes over
- Test Public Safety Crime and Fire logs. Pick some dates beyond the cached default. https://safety.uiowa.edu
- Thank you form would also be in-scope but hopefully you don't need to set that up and can see that our typical usage is already compatible based on the three other use cases above.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,6 +93,9 @@
     "require": {
         "ext-json": "*",
         "ext-simplexml": "*",
+        "guzzlehttp/guzzle": "7.12.1 as 7.9.3",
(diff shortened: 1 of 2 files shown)


cory-skeers (approved):
Tested with the crime/fire logs, recserv hours, and UI events and all worked without a hitch.

Travis has completed and is happy.

🚀 and 🤞 the Thank You form doesn't have anything outstanding compared to the other uses.

PR #9948, opened 2026-06-22
Sprint 317

joewhitsitt:
Used the new `./sn mc` command even though the update hook hasn't been deployed and run on existing submissions.

# How to test


joewhitsitt:
there might be an immediate follow-up but don't want to hold up the bus any longer.

cory-skeers (approved):
👍

PR #9765, opened 2026-04-15
Drupal 10.5.9

joewhitsitt:
Fixes https://github.com/uiowa/uiowa/issues/138 (display text is now possible!)
Resolves https://github.com/uiowa/uiowa/issues/9674 (upgrading fixes the issues and maintains the effect of the patch that was removed, also removes a console error that was braking linkit)
Fixes https://github.com/uiowa/uiowa/issues/8347 (the link bubble closes when the media modal is opened)

I wanna say we were putting off a bump to 10.5+ just so we could only upgrade to Drupal 11 and not because of some technical limitation.

Some notable change log items that caught my attention though.

https://www.drupal.org/project/drupal/releases/10.5.0 (CKEditor, every table gets a table class)
https://www.drupal.org/project/drupal/releases/10.5.5 (Revision changes affecting moderation and json api)

# How to test

## Config sanity check

As far as I can tell, there are no config changes.

`ddev blt uiowa:usplits:site --host=careers.uiowa.edu`

## Patches updated

[3008292] ImageItem::getUploadValidators                                                                                                                                                                                       
  - First added: f863afc6d — 2022-04-08 — "set max image size for uploads going forward (#5089)"

[2230909] Simple decimals fail to pass validation                                                                                                                                                                              
  - First added: 93913c962 — 2024-02-01 — "[Facilities] Project data (#7285)"

PR #2983, opened 2021-02-01
Feature: Drupal 9 upgrade

pyrello:
# How to test

These are my best ideas about how to comprehensively test this. Please add additional items if there are things I have missed that you think are important.

* Exit VM.
* Ensure you have composer 2 installed on host machine.
* `composer install`
* `vagrant provision`
* `blt ds` to sync a site.

## Article and Person content types
* Test adding, editing, deleting.
* Test viewing revision diff, reverting to a previous revision.
* Test moderation workflow: set content item to draft, set to review, archive content.

## Roles/permissions
* Masquerade as viewer, editor, publisher, webmaster to check that permissions are as expected.

## V2 (Paragraphs)
* Add, edit, remove all the various basic paragraph block types.
* Test reverting to previous revision.

## V3 (Layout Builder)
* Add, edit, remove, move all the various basic block types.
* Test reverting to previous revision.

# Updates
* Resolves #2193. 
* Resolves #2592 
* Resolves #3030
* Resolves #931
* Created alternate version of views_tree to be compatible with D9. This is temporary until a merge request is accepted.

## Updated core patches
Updated “fix duplicate drag handles in claro” patch

Updated "[PP-1] Reverting entity revisions that contain custom blocks erroneously triggers EntityChangedConstraint” patch

## Layout Builder Styles
* Updated LB styles grouping patch
* Removed “Block theme suggestions not working for layout builder blocks using styles": "https://www.drupal.org/files/issues/2020-04-10/block-theme-suggestions-not-working-for-layout-builder-blocks-using-styles-3126609_0.patch"
* Re-exported config because of schema change.

## Composer update prior to switch to version `^9.0.0`
Prior to updating the required version of Drupal, I ran `composer update`, which triggered the following updates:

  - Upgrading behat/gherkin (v4.6.2 => v4.7.1)
  - Upgrading bower-asset/fontawesome (5.15.1 => 5.15.2)
  - Upgrading bower-asset/lazysizes (5.2.2 => 5.3.0)
  - Upgrading caxy/php-htmldiff (v0.1.9 => v0.1.10)
  - Upgrading chi-teck/drupal-code-generator (1.33.0 => 1.33.1)
  - Upgrading commerceguys/addressing (v1.1.0 => v1.1.1)
  - Upgrading composer/ca-bundle (1.2.8 => 1.2.9)
  - Upgrading composer/composer (1.10.15 => 1.10.20)
  - Upgrading composer/installers (v1.9.0 => v1.10.0)
[...]

PR #9954, opened 2026-06-25
Composer cleanup/ignore

joewhitsitt:
>Found 2 ignored security vulnerability advisories affecting 1 package

# How to test

- Passing builds
- Old security advisories no longer flagging.
- Current security advisories being ignored.
- See removed traces of blt-drupal-test which is no longer needed. See https://github.com/uiowa/uiowa/pull/9765/changes and https://github.com/uiowa/uiowa/pull/2983


GaryRidgway:
I will pull this to review

GaryRidgway (approved):
- Builds are passing, `composer install` runs green.
- `composer audit` shows the four old ones are gone and the two new ones are there.
- A `grep` shows `blt-drupal-test` is gonezo.
NICE

PR #9972, opened 2026-07-08
Research areas checkboxes

### LINKED ISSUE #8983: Render Research Areas exposed filter as a select list on people list block
# Story

As a SiteNow user, I would like my person list block to display the Research Areas filter as a select list rather than an autocomplete so users can be aware of the different types of research areas to filter by.

# Additional Info

We made some enhancements to these filters as part of https://github.com/uiowa/uiowa/issues/8894 but decided to hold on this because we were running into issues with BEF https://github.com/uiowa/uiowa/issues/8894#issuecomment-3029277383.

Also, some background info... Because the list block configuration has allowed users to pre-filter the results by multiple Research Areas, the exposed filter needs to allow multiple or we would be breaking existing functionality for the sites that are already pre-filtering by multiple Research Areas (I believe we identified 3 sites currently https://github.com/uiowa/uiowa/issues/8894#issuecomment-3028160991).

I am not aware of the need for users to select multiple here other than for the reasons mentioned above.

# Proposed solution

cory-skeers:
Resolves #8983 

# How to test

```
ddev blt ds --site=sandbox.uiowa.edu &&
ddev drush @sandbox.local uli /people-list-block-conversion-test-page
```
Clean sync and update hook
Compare the page to https://sandbox.prod.drupal.uiowa.edu/people-list-block-conversion-test-page; you should see similar filter types.
Test existing and new people list blocks:

- Various combinations of exposed, non-exposed filters
- Pre-filtering a block, and having that filter exposed and unexposed
- If no filters are exposed, the "Apply" and "Reset" buttons should not appear
- The Research Areas (or alternatively-named filter, if the site has changed it) should now be multi-select checkboxes
- Try "Filter Research Areas" turned on and off.
  - When turned on, only Research Areas which will have at least one result should appear in the options
  - When turned off, ALL Research Areas should appear as options
  - If multiple blocks, they should filter (or not) their options independently of other blocks on the page

Another good test page for combined filters is:
https://music.uiowa.ddev.site/people/faculty-area


config/default/views.view.people_list_block.yml
```diff
@@ -687,7 +688,7 @@ display:
         type: mini
         options:
           offset: 0
-          pagination_heading_level: h2
+          pagination_heading_level: h4
```
cory-skeers:
```suggestion
          pagination_heading_level: h2
```

bewoldt:
pulling down to test

joewhitsitt:
Adding @bewoldt as assigned to the PR while testing.

bewoldt:
Issues I've come across

**Duplicate people**
When selecting multiple research areas (exposed and unexposed), a person gets duplicated for each research area that they are in.

----------------------
**Back button**

- expose research areas but don’t preset research areas. Update and save page
- one live view, select a research area (internal medicine)
- click on person
- click back
- results show everyone(default) but the research areas dropdown is still set to internal medicine
----------------------
**Checked research areas come back (might work as expected)**

- expose research areas.
- preset research area to "duplicate bug test". Update and save page
- on live view, clear checkboxes
- click apply
- expanded research areas and see that checked research areas are selected.

----------------------
**Reset issue? Not an issue with your branch. Also present on production**

- Expose research areas
- Preset Person Type = Faculty but do not expose filter. Update and save page
- click reset
- The reset button resets all filters including ones not exposed so the person type Faculty is removed, showing everyone. Should the reset only work for exposed filters?

cory-skeers:
> Issues I've come across
> 
> **Duplicate people** When selecting multiple research areas (exposed and unexposed), a person gets duplicated for each research area that they are in.

I pushed up a change in https://github.com/uiowa/uiowa/pull/9972/commits/ebe906f8b25a7df22bd17a0fbecb0169acefe292 that should fix this.

> **Back button**
> 
> * expose research areas but don’t preset research areas. Update and save page
> * one live view, select a research area (internal medicine)
> * click on person
> * click back
> * results show everyone(default) but the research areas dropdown is still set to internal medicine

When I try this, after clicking back I'm seeing the research areas dropdown still set, and the results still filtered 

> **Checked research areas come back (might work as expected)**
> 
> * expose research areas.
> * preset research area to "duplicate bug test". Update and save page
> * on live view, clear checkboxes
> * click apply
> * expanded research areas and see that checked research areas are selected.

Hrm...It looks like if you clear it out and click apply, it gets re-added, but if you clear it out and add a different one, it stays cleared out. I think because of that, it can be treated as "works as expected." It's a little confusing and maybe not what I would choose, but it's also an odd case where I don't know that there is one way that for sure is how it should work.

> **Reset issue? Not an issue with your branch. Also present on production**
> 
> * Expose research areas
> * Preset Person Type = Faculty but do not expose filter. Update and save page
> * click reset
> * The reset button resets all filters including ones not exposed so the person type Faculty is removed, showing everyone. Should the reset only work for exposed filters?

Oof...I think it should absolutely only work for exposed filters, and that seems like a big bug.

Even if it's not being introduced as part of this, I might poke a bit and see if I can get a fix in. For IGHN people in International, that feels like it might be a very apparent bug.

bewoldt:
**Duplicate issue is fixed.**

**Checked research areas come back.** 
_"Hrm...It looks like if you clear it out and click apply, it gets re-added, but if you clear it out and add a different one, it stays cleared out. I think because of that, it can be treated as "works as expected." It's a little confusing and maybe not what I would choose, but it's also an odd case where I don't know that there is one way that for sure is how it should work."_
- That's what I thought too

**Back button issue.** Looks like this is a browser issue. Works correctly in Safari and Chrome. I still see it happening in Firefox every once in awhile but somehow fixes itself if you do it enough times. I saw it happen once using just the Person Type field so it might be a browser cache issue.

joewhitsitt:
@bewoldt do you approve? is this needing more folks to review?

bewoldt:
> @bewoldt do you approve? is this needing more folks to review?

I think another reviewer would be helpful.

cory-skeers:
> **Reset issue? Not an issue with your branch. Also present on production**
> 
> * Expose research areas
> * Preset Person Type = Faculty but do not expose filter. Update and save page
> * click reset
> * The reset button resets all filters including ones not exposed so the person type Faculty is removed, showing everyone. Should the reset only work for exposed filters?

This should now be fixed. The Reset button forces a page reload which should a) allow AJAX for all regular filtering and b) more closely match the Core Views experience without too much overhead.

GaryRidgway:
I will pull to review as a second reviewer
<details><summary>Checklist for me:</summary>
<p>





- [x] New blocks 
    - [x] Various combinations of exposed, non-exposed filters
    - [x] Pre-filtering a block, and having that filter exposed and unexposed
    - [x] If no filters are exposed, the "Apply" and "Reset" buttons should not appear
    - [x] The Research Areas (or alternatively-named filter, if the site has changed it) should now be multi-select checkboxes
    - [x] Try "Filter Research Areas" turned on and off.
      - [x] When turned on, only Research Areas which will have at least one result should appear in the options
      - [x] When turned off, ALL Research Areas should appear as options
      - [x] If multiple blocks, they should filter (or not) their options independently of other blocks on the page
- [x] Existing blocks 
    - [x] Various combinations of exposed, non-exposed filters
    - [x] Pre-filtering a block, and having that filter exposed and unexposed
    - [x] If no filters are exposed, the "Apply" and "Reset" buttons should not appear
    - [x] The Research Areas (or alternatively-named filter, if the site has changed it) should now be multi-select checkboxes
    - [x] Try "Filter Research Areas" turned on and off.
      - [x] When turned on, only Research Areas which will have at least one result should appear in the options
      - [x] When turned off, ALL Research Areas should appear as options
      - [x] If multiple blocks, they should filter (or not) their options independently of other blocks on the page

</p>
</details> 


Notes As I'm going:
___
**Case**: the view is filtered on a research area in the layout builder config backend, like `duplicate bug test`.
**Issue**: When I try to remove that filter in the exposed filter and hit apply, it is automatically added back when I click apply. By clicking apply, I am under the impression that the changes I made should persist.
**Solution**: hide the filters that are set in the backend if they are going to persist, or allow the removal of the filters upon apply. I'd lean towards the former, as there may be some ways a user wants to set up a view to only show people that intersect with a research area, but might have other research areas.

I'm noticing that this was also found by @bewoldt as well, but I think giving the user the choice to remove it is an anti-pattern. if it is set in the backend, it shouldn't be editable.

I am now, noticing that if you disable the default filter that should always be there, and then you enable another one, for example, politics, and then click apply. Then the default does not persist. That is a very confusing nuance for a user I think, especially if we are trying to keep that original filter on.
___

GaryRidgway (approved):
I see that there has been some conversation about the only issue I saw with the PR, but if we don't want to address that right now, then I would say this is good to go.

I'd love to see an issue to fix that if we don't resolve it though? No worries if not, just throwing it out there.

cory-skeers:
> I see that there has been some conversation about the only issue I saw with the PR, but if we don't want to address that right now, then I would say this is good to go.
> 
> I'd love to see an issue to fix that if we don't resolve it though? No worries if not, just throwing it out there.

Found the root of the issue, and poking at what a solution might look like. 

The submission handlers struggle to differentiate between a fully empty checkboxes field and a form that hasn't been interacted with at all. 

In the particular case where a default checked value is provided by the Views "exposed" options like we're doing in order to provide the defaults, the formstate is more-or-less viewed as the same as if it was all empty, the user clicked a checkbox, and then unclicked that same checkbox, and so didn't have a net interaction. The handlers then abort early, and that forces the form to re-set everything as its default.

If the user checks another box as well, then the form properly gets viewed as having been interacted with, and the full process ensues.

Edited to add: I've been testing it with checkboxes specifically, but it may be other filters as well, and not specifically limited by widget type.

cory-skeers:
An extra set of eyes reviewing this after https://github.com/uiowa/uiowa/pull/9972/commits/92351bada8caef19f72b9fdbfc003cf545c8cb26 would be much appreciated.

Claude assisted on the cleanup and comments, so the comments are verbose but describe the process fairly well otherwise.

It covers two things:

1. An "empty" checkboxes field shows as a non-interaction. So if a prefiltered value is added to an exposed Research Areas, unchecking that box and not adding another didn't work. That should now work.
2. The "Filter Research Areas" checkbox and Person Types values didn't persist consistently on submit (and technically paging, but that was a less apparent issue) because of AJAX. The values now carry through.

joewhitsitt:
i can take a look at this

docroot/modules/custom/sitenow_people/sitenow_people.module
```diff
@@ -435,10 +486,68 @@ function sitenow_people_entity_presave(EntityInterface $entity) {
...
+      $components = $section->getComponents();
+      // Skip empty sections.
+      if (empty($components)) {
+        continue;
+      }
+
+      foreach ($components as $component) {
+        $config = $component->get('configuration');
+        // Check if this is a people_list_block.
+        if (isset($config['id']) && strpos($config['id'], 'views_block:people_list_block') === 0) {
```
joewhitsitt:
```suggestion
        if (isset($config['id']) && str_starts_with($config['id'], 'views_block:people_list_block')) {
```

php 8.3 modernization so i am told. not sure what is more performant, but I think `str_starts_with` might be less error prone than `strpos` (sometimes `false`?).

docroot/modules/custom/sitenow_people/sitenow_people.module
```diff
@@ -889,16 +1058,280 @@ function sitenow_people_form_views_exposed_form_alter(&$form, FormStateInterface
...
+      ];
+
+      // Same carry-forward, for the same reason, for a block's fixed
+      // person-type restriction - see the fallback in
+      // sitenow_people_views_pre_build() that reads this back.
+      $person_type_filter = $block_config['exposed']['filter-field_person_types_target_id'] ?? NULL;
+      if ($person_type_filter && empty($person_type_filter['exposed']) && !empty($person_type_filter['value'])) {
+        $form['sitenow_people_person_type_value'] = [
+          '#type' => 'hidden',
+          '#value' => is_array($person_type_filter['value']) ? reset($person_type_filter['value']) : $person_type_filter['value'],
```
joewhitsitt, line 1179: >Potential issue with empty array in reset() call — If $person_type_filter['value'] is an empty array, reset([]) returns false. The hidden field would receive '#value' => false, which
  may not render as expected. Suggest: $person_type_filter['value'] ? reset($person_type_filter['value']) : NULL or a null coalesce pattern.

- Claude

joewhitsitt:
should all the display:none attributes also be sprinkled with disabled or will that affect the value handling?

joewhitsitt (approved):
Nice work on this @cory-skeers. I appreciate your attention to all the details/scenarios. It met my expectations during testing. Tried to trip things up as it related to https://github.com/uiowa/uiowa/pull/9972#issuecomment-5045961414. Left some suggestions in here, but they aren't blockers.

Would like to see some of this abstracted so that it can be used again (future slice).

docroot/modules/custom/sitenow_people/sitenow_people.module
cory-skeers: The `if()` above it checks for empty already, so we can't get into Claude's potential state.

cory-skeers:
> should all the display:none attributes also be sprinkled with disabled or will that affect the value handling?

There might be a way to make it work here since we're already doing some behind-the-scenes hand-waving, but I think disabled would hide it when we try to fetch it from the request level.

> Would like to see some of this abstracted so that it can be used again (future slice).

Agreed!

Thank you for the extra review and testing! I really appreciate it after this has iteratively updated over time.

Issue #9932, opened 2026-06-17
Replace existing deployment process


briand44:
# Motivation

As part of our work to move away from BLT, we did an investigation into replacing our current deployment process in #9917. We should move forward with implementing the new process based on the recommendation provided and agreed upon by the team.

# Grooming convo - from https://github.com/uiowa/uiowa/issues/8181#issuecomment-4734041099
 
**Quadrant:** 

Complex

 **Discussion:**
 - replacement of deployment makes sense as a next step.
 - some blt'isms will need to stay until we switch to drupal recommended settings
 - implement as one slice of work - 30/60/90 process useful
 - frontend is standalone, Sean to verify accounted for in process

**Proposed solution or definition of done:** 
- implementation of plan from https://github.com/uiowa/uiowa/issues/9917#issuecomment-4752239683 once feedback from team

pyrello:
# R60 Gate: Prove It

_Written with assistance from Claude (Opus 4.8)_

R30 landed on approach A: re-home the existing build-and-push deploy model off BLT, with the build as provider-independent shell and the Acquia-coupled steps as `sn` (Symfony Console) commands. R60 was to prove that approach with de-risking POCs and to answer the open questions. We went past POC: the full pipeline now runs in Travis and deployed the entire dev fleet.

## What we proved

The dev/test deploy path, end to end, off BLT, on live Acquia Cloud Next:

- **Build** (`scripts/deploy/build.sh`): assemble a clean copy of the working tree, compile front-end assets (`yarn`), install production deps (`composer install --no-dev`), strip cruft, stamp the version.
- **Distribute** (`./sn deploy:distribute`): commit the artifact and force-push `{branch}-build` to all 10 Acquia remotes.
- **Auto-deploy**: Acquia redeploys each dev environment on the push to its tracked branch.
- **Update** (`post-code-update` hook -> `./sn deploy:update` -> per-site `drush deploy`): fleet update with GNU `parallel`, `run_first` ordering, per-site guards, and pass/fail aggregation.

Validated twice on live Acquia: first a scoped single-app smoke test (uiowa09 dev), then the full fleet via Travis (CI green -> deploy stage -> all 10 apps). Multiple sites completed a full `drush deploy` (updatedb + config import + cache rebuild + deploy hooks); per-site failures were isolated and aggregated correctly.

## Open questions from R30, answered

[...]

PR #9975, opened 2026-07-09
Symfony console ume

### LINKED ISSUE #9963: Replace BLT UME
# Motivation

As part of our strategy to move away from BLT, we should implement a replacement for the BLT UME (uiowa:multisite:execute) command so that we have a way of running commands across all sites on the platform.

# Grooming convo from https://github.com/uiowa/uiowa/issues/8181#issuecomment-4858955051
 
**Quadrant:** 

Complex

 **Discussion:**
 - uiowa:multisite:execute?
 - ability to run commands against all sites
 - If it could handle escaping better this time around, I’ll buy the beers (From Joe)
 - https://github.com/jigarius/drall?
 - we may be able to accomplish this without any additional dependencies and that may be preferred based on being burnt by other dependencies recently
 - direct port vs something new
   - 30/60/90 check in

**Proposed solution or definition of done:**
- Investigate to understand if we want to use something existing like https://github.com/jigarius/drall or implement our own
- Replace BLT UME command with a non-BLT replacement that allows running commands across all sites
- Shouldn't have any loss of functionality from existing BLT UME command

GaryRidgway:
## To test

1. Pull branch, **ddev restart**, ddev composer install.
2. Run through the following commands:
    1. `ddev exec ./sn list` 
        1. Should show you the list of commands with the new one there. 
        2. Verify that it also shows the correct commands that require ddev.
    2. `ddev exec ./sn ume --help`
        1. Verify that it shows examples, arguments, and some gotchas.
    3. `ddev exec ./sn ume --dry-run -- cr`
        1. Verify that it lists 1,307 sites and their aliases and doesn't execute anything.
        2. Verify the header notes the `--ssh-options` string that gets appended to each command, and that the per-site lines are clean drush commands you can copy-paste.
    4. `ddev exec ./sn ume --apps=uiowa09 --exclude=accessibility.uiowa.edu --dry-run -- cr`
        1. Verify that the count goes down and accessibility is not listed because we excluded it.
    5. `ddev exec ./sn ume --apps=BOGUS --exclude=accessibility.uiowa.edu --dry-run -- cr`
        1. Verify that the command exist with error code 1.
    6. `ddev exec ./sn ume --apps=uiowa09 -- core:status`
        1. Answer the confirm prompt with 'n'. Command should abort.
        2. Answer the confirm prompy with 'y'. Command should execute.
    7. `ddev exec ./sn ume --apps=uiowa09 -y -- core:status`
        1. Verify that adding the `-y` to the command before the `--` made it so you didn't have to go through the confirm prompt.
    8. `ddev exec ./sn ume --apps=uiowa09 -v -y -- sql:query "SELECT COUNT(*) FROM node"`
        1. Verify that the command goes through and lists the amount of nodes each site has under its alias as they are processed. (This is an example of arguments being passed through to the command, something that UME couldn't do.)
    9. `ddev exec ./sn ume --apps=uiowa09 -y -- fake:command`
        1. Verify it aborts BEFORE running any site: the command is checked on one canary site first (`drush help` executes nothing), and with -y a failed check is a hard abort and asks you to run the command again without the -y if its intentional.
        2. Re-run without `-y` — verify you get "Run the fleet anyway? [no]", for the rare command that legitimately exists on only some sites (e.g. site-specific module commands like `registrar_core:final-exam-schedule`).

    10. `ddev blt ume`
        1. Verify that the command is not found, because we deleted the BLT version of it.
    11. `ddev exec ./sn ume --concurrency=0 --dry-run -- cr` (also try `--concurrency=abc`)
        1. Verify it errors ("must be a positive integer") and exits 1 instead of silently using the default.
    12. Multiplexing is scoped to fleet runs only (no more global SSH config file):
        1. `ddev drush @accessibility.prod status`, then `ddev exec 'ls ~/.ssh'` — verify no `cm-*` files (everyday drush never multiplexes).
        2. Start the command from step viii again, and then immediately run `ddev exec 'ls ~/.ssh'` — verify `cm-*` socket(s) exist. 
            1. Wait ~60s after the run finishes and `ls ~/.ssh` again — verify they're gone.
    13. `ddev exec vendor/bin/phpunit -c phpunit.xml.dist --filter 'FleetRunnerTest|ProcessPoolTest' tests/phpunit/src/Unit`
        1. Verify 25 tests pass (FleetRunner selection/jobs, ProcessPool scheduling/retry).

### Alternative feedback I would be interested in if you have the energy

- Exit codes are now 0/1/2 (Scripts will depend on these once we cut over — object now if you want plain 0/1.):  
    - 0 -> clean
    - 1  -> couldn't run (something goofed like ssh or something)
    - 2 -> completed-with-site-failures. 
- Does anyone have a problem with a confirmation prompt for the new command? (Old UME defaulted to yes and just ran with it.)
- Does anyone have an issue with the concurrency of 8 workers per app capped at 32? I did testing in #9920 and landed on this number, but open to opinions.
- Is there too much help text on the command?
- I added (ddev required) to the commands that need it to each command in the list and help text for them. Is that okay?
- Any other qualms?


pyrello:
Pulling to test.

pyrello:
## Code review: `sn ume`

Converts BLT `ume` to a Symfony Console `sn ume`, adds `FleetRunner` + `ProcessPool`, deletes the BLT command. Clean continuation of the console migration: argv-array passthrough, structured per-site results, bounded concurrency with a validated per-app cap.

**Must fix before merge**

- **Tests.** `FleetRunner` (`select`, `buildJobs`, `defaultConcurrency`) and `ProcessPool` (group-cap scheduling, retry merge) are pure logic, and `tests/phpunit/src/Unit/` is where that convention already lives. Same bar #9932 set.
- **Retry is indiscriminate.** `ProcessPool` re-runs every non-zero exit, so a deterministic failure (unknown command, wrong `--env`) runs the entire fleet a second time before reporting it. Retry should fire only when the failure looks transient (SSH, connection, bootstrap), so a systematic failure fails once.

**SSH multiplexing**

New config, added in this PR, applied by the SSH client to every command reaching an Acquia server from ddev across dev, test, and prod, not just ume. `report:splits`, `report:inactive`, and `blt ds` all ride the same per-app socket transparently.

- **Validation reach.** How much was this exercised beyond ume and the #9920 split-check? Those commands now inherit multiplexing they were never tested against.

**Worth flagging, your call**

- A malformed `blt/manifest.yml` throws an uncaught `ParseException`; only `RuntimeException` is caught, so the user gets a stack trace instead of a clean error.
- `--concurrency=0` silently falls back to the default rather than erroring.
- User abort (`n`) returns exit 1, the same code as "couldn't run," worth reconciling with the 0/1/2 contract you're testing.

**Fragmentation**

- `MultisiteExecuteCommand::failureReason()` and the error branch of `ReportSplitsCommand::parseSplitStatuses()` are the same failure-summary logic. Good candidate for a trait or small shared helper so there's one owner.
- `FleetRunner::select()` supersedes the inline manifest selection in `report:splits`. That's the right direction, generalizing bespoke code. It just needs a follow-up issue tracking the eventual `report:splits` migration.
- `ProcessPool` reproduces pieces of the existing `Report/DrushRunner` (the `{exit, output, error}` contract, timeout and launch handling). This is the one to account for explicitly: intentional or not? If not, work through the overlap and note what you would recommend, for example retiring `DrushRunner` when the report commands are next touched.

**Settled per #9963 and #9932, no action:** Slack-on-failure drop, custom-pool build, default-site and DB-guard behavior, confirm defaulting to no.

_Drafted by Claude (Opus 4.8) on behalf of @pyrello_

pyrello:
Setting to draft at @GaryRidgway's request

GaryRidgway:
I've been testing the multiplexing of drush commands this morning to ensure I completely understand the changes that our drush execution goes through with this, and it is as follows:
___
`.ddev/homeadditions/.ssh/config.d/acquia-multiplex.conf` multiplexes **every** SSH/drush
connection from ddev, which means that your first connection to an app server authenticates and stays open 10 minutes. Later commands reuse it (~0.4s instead of a full handshake). 

Thinking through the possibilities of what could break this, we tested the following:
_(A 'fleet run' in the following table is something that runs against the entire platforms set of sites.)_
| Scenario | Result |
|---|---|
| Manual drush while your own fleet run is going | Works, ~120ms |
| Two fleet runs on the same app at once | Both execute, 0 failures, ~3× slower |
| Connection killed ungracefully | Next command self-heals in ~1s |
| Laptop sleep | First command after wake stalls ~65s, then fine |
| VPN toggle | No effect — connection survived |
| 10 min idle | Cleans itself up |
| `ddev restart` | Clears all multiplexing state |

One thing worth knowing:

- Under heavy concurrent load you may see `mux_client_request_session: session request
  failed: Session open refused by peer` on stderr. It's harmless — ssh falls back to a
  direct connection and the command succeeds.

___

The only thing that may be a large enough change to note is the laptop sleep delay, but considering the upsides to the multiplexing speed and the inconsistent instances of laptop sleeping and the user logging back in to use the same drush session in between the 10m frame, I think it is acceptable.

GaryRidgway:
Going to move on to the other concerns in [this comment](https://github.com/uiowa/uiowa/pull/9975#issuecomment-4971090837) now.

pyrello:
> First command after wake stalls ~65s, then fine

> The only thing that may be a large enough change to note is the laptop sleep delay, but considering the upsides to the multiplexing speed and the inconsistent instances of laptop sleeping and the user logging back in to use the same drush session in between the 10m frame, I think it is acceptable.

Am I understanding this correctly? If you are connected and your laptop goes to sleep during the 10 minute connection window and you wake it back up, you encounter a ~65s delay in reconnecting? Or is it any time your laptop goes to sleep? Clarification of the exact conditions under which this happens would be great.

When you saw that its worth the tradeoffs, I want to be clear on what the tradeoffs are. Multiplexing speeds up commands that are run that can use concurrent requests (more than 1 thing happening at a time). That is a huge benefit when those commands are being run. But in my experience, those commands are only run occasionally. So, if there is a slowdown everytime you run a drush command and your laptop goes to sleep afterwards, I'm not sure I would say that it is obvious to me that the pros outweigh the cons. 

Another thing to consider: by changing the ssh settings to make certain commands we run occasionally faster is a big system change for a narrow target. I don't have an exact idea how it might work, but it introduces a new layer of complexity into daily operations. Commands that can't effectively use multiplexing, but are run every day run through this new setup. Ideally, there would be a way for us to only have multiplexing involved when it was specifically relevant.

GaryRidgway:
> Am I understanding this correctly? If you are connected and your laptop goes to sleep during the 10 minute connection window and you wake it back up, you encounter a ~65s delay in reconnecting? Or is it any time your laptop goes to sleep? Clarification of the exact conditions under which this happens would be great.

For clarity: You run some drush command in the previous 10 minutes, the laptop sleeps, and the next command after wake finds the stale socket and pays ~65s once while ssh times out and falls back. 

> Ideally, there would be a way for us to only have multiplexing involved when it was specifically relevant.

Agreed, and I just chatted with Claude about it and there is a way to do this. I will try an implementation now that removes the multiplexing structure from the base of ddev.

sitenow/src/Process/FleetRunner.php
```diff
@@ -26,6 +26,20 @@ class FleetRunner {
...
+   * Passed to each drush process via --ssh-options so that only fleet runs
+   * multiplex: everyday drush commands stay on stock SSH with no shared
+   * state. The first connection to an app server authenticates and becomes
+   * the master; the rest of the fleet rides it as sessions (sshd caps these
+   * around 10 — PER_APP_CAP stays under that, and over-cap requests fall
+   * back to a direct connection). The master self-closes 60 seconds after
+   * its last session ends. PasswordAuthentication=no is restated because
+   * this string replaces drush's default ssh.options, not appends to it.
+   */
+  const SSH_OPTIONS = '-o PasswordAuthentication=no -o ControlMaster=auto -o ControlPath=~/.ssh/cm-%C -o ControlPersist=60';
```
pyrello: `PasswordAuthentication=no` as an option inside this runner seems a little sus. Can you explain why this should be here?

GaryRidgway: Drush already does this with every command, if we didn't add it here we would actually be _removing_ it. There's a small docblock comment above that line that tries to explain it a bit. Long story short, acquia only uses public key auth, and since we already auth with SSH, the password stuff is irrelevant, and wouldn't work anyways, so its just always been a default argument added to ensure we don't deal with it.

GaryRidgway:
## Must fix

- **Tests** — Lots of tests have been added.
- **Indiscriminate retry** — A pass where *every* job failed is treated as systematic and
never retried. Partial failures still retry.

## SSH multiplexing
- Changed to be added to the commands as they are passed through `ume`, and no longer become the default for all drush commands.

## Worth flagging

- **Malformed manifest** — Already handled, though it's non-obvious: 
    - Symfony's
  `Yaml\Exception\ParseException` extends `Yaml\Exception\RuntimeException`, which extends
  `\RuntimeException`, so the existing catch renders a clean error. 
    - What *wasn't* handled was
  valid YAML with the wrong shape (a truncated scalar, an app whose value isn't a list) — those
  would have been a TypeError stack trace. `select()` now validates shape and throws the same
  clean `RuntimeException`; test-covered.
- **--concurrency** — Now errors (`exit 1`) instead of silently falling back on any value < 1, as does any non-numeric value.
- **Abort exit code** — Intentional: the contract is 0 = all succeeded, 1 = nothing ran,
  2 = partial. Declining the prompt is "nothing ran," so a script can't tell it apart from any
  other no-op — which is the point. Now documented on `EXITCODE_PARTIAL`.

## Fragmentation

- **`failureReason()` duplication / `report:splits` selection / `DrushRunner`** — All three are
  the same story: the report commands predate `FleetRunner` and should migrate onto it in future issues. It consolidates the failure-summary logic and retires
  `DrushRunner` (the `{exit, output, error}` contract overlap is intentional so that migration
  is mechanical — `DrushRunner` is blocking while the pool needs non-blocking start/poll, so one
  can't be built on the other in the meantime).

_Drafted by claude, Reduced and clarifieds significantly by me._

pyrello:
Two things.

Retracting my malformed-manifest point: `ParseException` already extends `\RuntimeException`, so it was caught, and your shape-validation fix covers the real gap. Ignore that one.

On `--ssh-options`: replacing drush's default also drops `-A` (agent forwarding), a uiowa default in `drush/drush.yml`, so a fleet `sql:sync` or `rsync` between environments would fail through ume but work via plain drush. Appending to the base instead of restating it keeps `-A` and can't drift:

```php
$base = Yaml::parseFile($repoRoot . '/drush/drush.yml')['ssh']['options']
  ?? '-o PasswordAuthentication=no';
$ssh_options = $base . ' -o ControlMaster=auto -o ControlPath=~/.ssh/cm-%C -o ControlPersist=60';
```

_Drafted by Claude (Opus 4.8) on behalf of @pyrello_

pyrello:
> `ddev exec ./sn ume --apps=uiowa09 -y -- fake:command`
> 
>     1. Verify that the sites run only once.
>     2. Verify that the output tells you what happened (That there is no command called `fake:command`)
>     3. Verify the whole command exist with exit code 2.

It seems odd to me that it even runs with a non-registered command.

GaryRidgway:
> On `--ssh-options`: replacing drush's default also drops `-A` (agent forwarding), a uiowa default in `drush/drush.yml`, so a fleet `sql:sync` or `rsync` between environments would fail through ume but work via plain drush. Appending to the base instead of restating it keeps `-A` and can't drift:

This has been fixed.

> > `ddev exec ./sn ume --apps=uiowa09 -y -- fake:command`  

> It seems odd to me that it even runs with a non-registered command.

I've added a pre-flight check and prompt for this to avoid running them.

pyrello:
> > > `ddev exec ./sn ume --apps=uiowa09 -y -- fake:command`
> 
> > It seems odd to me that it even runs with a non-registered command.
> 
> I've added a pre-flight check and prompt for this to avoid running them.

One other thing about this I find confusing is that it prints out the whole list of sites twice:

```
ddev exec ./sn ume --apps=uiowa09 -- fake:command
`fake:command` failed a preflight check on 3ddesignstudent.org.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)

 Run the fleet anyway? (yes/no) [no]:
 > yes

 Run `drush fake:command` on 87 sites across 1 app(s) (prod)? (yes/no) [no]:
 > yes

Running on 87 sites, 8 at a time...
✖ [1/87] 3ddesignstudent.org.uiowa.edu (exit 1)
✖ [2/87] accessibility.uiowa.edu (exit 1)
✖ [3/87] appliedai.org.uiowa.edu (exit 1)
✖ [4/87] bunsawat.lab.uiowa.edu (exit 1)
... (omitted for length)
✖ [84/87] tinelli.sites.uiowa.edu (exit 1)
✖ [85/87] uimobileclinic.org.uiowa.edu (exit 1)
✖ [86/87] usthaiconsortium.conference.uiowa.edu (exit 1)
✖ [87/87] yuanliao.sites.uiowa.edu (exit 1)

Finished in 44.1s: 0 succeeded, 87 failed.

[WARNING] 87 site(s) failed:
  uiowa09: 3ddesignstudent.org.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: accessibility.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: appliedai.org.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: bunsawat.lab.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  ... (omitted for length)
  uiowa09: thegammanupes1914.org.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: tinelli.sites.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: uimobileclinic.org.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: usthaiconsortium.conference.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
  uiowa09: yuanliao.sites.uiowa.edu — exit 1 (There are no commands defined in the "fake" namespace.)
Failed to execute command `./sn ume --apps=uiowa09 -- fake:command`: exit status 2
```

GaryRidgway:
> One other thing about this I find confusing is that it prints out the whole list of sites twice:

Ah, good call. I can ask it to change that to summarize and not list the same error for all sites.

pyrello (approved):
Tested. Looks good. Nice work.

PR #9982, opened 2026-07-15
User listing for all sites

### LINKED ISSUE #9810: Report that lists all users and all sites they have signed in to
# Story

As a member of OSC leadership, I would like to be able to view a report of all SiteNow users and the sites they have accessed over the last year.

# Additional Info

Report where leadership could search or filter by an email address and see what websites they have accessed over the last year.

Report doesn't need to be a live tool, we just need the ability to produce an updated version as needed.

Don't want a huge investment if it can be avoided here. Spreadsheet that requires some data cleanup likely could work.

Ability to see users email, role, website, last login.

pyrello:
Resolves #9810

## How to test

Runs from the host shell or inside the container; DDEV is not required. CI covers the unit logic. Smallest scope first.

```bash
git checkout 9810_user_site_report
ddev composer install
```

Needs SSH agent keys: check `ssh-add -l` on the host, or run `ddev auth ssh` in the container. A first host run against an app can fail with `Host key verification failed`; connect once to the host named in the error to accept its key, or run it in the container.

### Nothing queried

- `report:users` (alias `users`) is registered, and is the only fleet command with no `(ddev required)` note:

```bash
./sn list
```

- Help lists every option (`--apps`, `--exclude`, `--exclude-users`, `--threshold`, `--export`):

```bash
./sn report:users --help
```

- Both print one error and exit 1, before any site is queried:

```bash
./sn report:users --apps=uiowa09 --threshold="banana"
./sn report:users --apps=BOGUS
```

### One app, from the host

- One row per (user, site), Roles limited to viewer/editor/publisher/webmaster, ending with `Generated in ...`:

```bash
./sn report:users --apps=uiowa09
```

- CSV at the repository root: the four columns, no password, and no `Generated in ...` line:

```bash
./sn report:users --apps=uiowa09 --export
```

- Same or fewer rows than the 1-year default:

```bash
./sn report:users --apps=uiowa09 --threshold="6 months"
```

- The excluded user and site drop out (use values from the run above):

```bash
./sn report:users --apps=uiowa09 --exclude-users=<email> --exclude=<domain>
```

### Same command inside the container

- Identical output:

```bash
ddev exec ./sn report:users --apps=uiowa09
```

### Whole fleet (optional, slow)

- Unreachable sites warn at the end without failing the run:

```bash
./sn report:users --export
```

- Failures report drush's message, not a bare exit code. Mid-run lines say `attempt failed` because the pool retries; the end-of-run summary is the verdict.


GaryRidgway:
pulling to test

GaryRidgway:
An initial note, as I am running this I am finding that upon running `./sn report:users --apps=uiowa09 --threshold="banana"` I am getting the correct threshold error, but then running `./sn report:users --apps=BOGUS` it tells me I need to run `ddev auth ssh`. I do so, but it does not correct the error.

Edit: I ran `chmod 600 ~/.ssh/id_rsa_2 && ssh-add ~/.ssh/id_rsa_2` and it allowed me to run it.

pyrello:
Both fixed in 56c598b39, thanks.

The agent check ran before the selection was validated, so `--apps=BOGUS` reported a missing key instead of the unknown app. It now runs last, once the command is actually about to reach a site:

```
./sn report:users --apps=BOGUS
 [ERROR] Unknown application(s): BOGUS
```

The agent error itself only mentioned `ddev auth ssh`, which is no help on the host. It now names both:

```
 [ERROR] No SSH keys loaded. Run 'ssh-add' on the host, or 'ddev auth ssh' in the container, before running this command.
```

The key permissions part was local to your setup, but it is worth knowing that a host run also needs each app's host key in `known_hosts`, or sites on an app you have not connected to before fail with `Host key verification failed`. Noted in the testing steps.

_Drafted by Claude (Opus 4.8) on behalf of @pyrello_

pyrello:
@GaryRidgway Not sure if you saw the last comment.

GaryRidgway:
> @GaryRidgway Not sure if you saw the last comment.

I did, thank you!

GaryRidgway (approved):
I've tested everything and it works! I needed to fix local stuff, but otherwise its working great. the demo site failed twice,

```


✖ [134/1312] attempt failed: demo.sitenow.uiowa.edu — [preflight] The alias @sitenowdemo.prod could not be found. (exit 1)
✖ [9/338] attempt failed: demo.sitenow.uiowa.edu — [preflight] The alias @sitenowdemo.prod could not be found. (exit 1)

```

but that is just the demo site, so I'm not worried about it. and the csv all exported nice!

joewhitsitt:
> the demo site failed

this is occurring with the ume commands too (e.g. cron)

pyrello:
Thanks for flagging the demo site failure. It is a real bug, but not in this command: `demo.sitenow.uiowa.edu` is in the manifest while being served from `sites/default`, so no `@sitenowdemo.prod` alias exists and every fleet run over app `uiowa` fails for it. How the default site should fit into fleet tooling needs a team decision, so it is tracked in #10032 rather than patched here.

The `[preflight]` text is drush's own error label, surfaced verbatim now that failures report a reason instead of a bare exit code.

_Drafted by Claude (Opus 4.8) on behalf of @pyrello_

PR #10024, opened 2026-07-28
Run the report commands through FleetRunner

### LINKED ISSUE #9991: Migrate report commands to use new UME implementation
#9975 introduced `FleetRunner`/`ProcessPool` as the shared layer for running `drush` across
manifest-selected sites. The report commands (`report:splits`, `report:inactive`) predate it
and carry their own copies of what it generalized:

- `ReportSplitsCommand` selects sites from the manifest inline — superseded by
  `FleetRunner::select()`.
- `Report/DrushRunner` duplicates the `{exit, output, error}` contract and the
  timeout/launch-failure handling now in `ProcessPool`. The overlap is intentional so this
  migration is mechanical; `DrushRunner` is blocking while the pool needs non-blocking
  start/poll, so neither can be built on the other.
- `ReportSplitsCommand::parseSplitStatuses()`'s error branch and
  `MultisiteExecuteCommand::failureReason()` are the same failure-summary logic; consolidate
  into one shared owner during the migration.

Scope:
- [ ] `report:splits` and `report:inactive` should select sites via `FleetRunner::select()` and run
      drush via `FleetRunner::run()` (gaining pooled concurrency + scoped SSH multiplexing)
- [ ] Consolidate the failure-summary helper
- [ ] Delete `Report/DrushRunner`
- [ ] Per-site behavior contracts from #9977 (dead-site skip, per-app abort/discard) preserved?? - we weren't sure what this meant while grooming

Refs #9975, #9963.

pyrello:
Closes #9991. Builds on #9982, which merged first.

`report:splits` and `report:inactive` were converted to Symfony Console before #9975. This PR rewires them both to use `FleetRunner` and allows them to gain multiplexing.

- `Report/DrushRunner` is deleted.
- Both commands gained `--exclude`, allowing a domain to be excluded from the output.
- `report:domains` loses its DDEV gate too, and now rejects an unknown `--apps` value instead of printing an empty table.
- `report:inactive` now reads the manifest, one row per site, and no longer needs Acquia credentials. Any prod domain that is not a site's canonical domain had no alias and only ever produced `N/A` rows. Those rows are gone.

## How to test

Everything runs from the host; `ddev auth ssh` still needs to have happened.

Confirm none of the three report commands claim to need DDEV:

```
./sn list
```

Splits, from the host with no `ddev exec`:

```
./sn report:splits --apps=uiowa08
```

Inactive. Runs the fleet twice, so expect two progress headers and one row per site:

```
./sn report:inactive --apps=uiowa07
```

Unknown app aborts before anything runs, exit 1:

```
./sn report:splits --apps=BOGUS
```

Same for `report:domains`, which validates against the Acquia app list rather than the manifest:

```
./sn report:domains --apps=BOGUS
```

Excluded sites drop out of the count:

```
./sn report:inactive --apps=uiowa01 --exclude=uiowa.edu
```

Export flag sends output to a file:

```
./sn report:splits --apps=uiowa09 --export
```

---

*Drafted by Claude on behalf of @pyrello.*


pyrello:
Noting for a reviewer that this might currently error on `tung.lab.uiowa.edu` because #10029 hasn't deployed yet.

```
./sn report:splits --apps=uiowa09 --export
Reading split status on 88 sites...
✖ [82/88] attempt failed: tung.lab.uiowa.edu — [warning] Drush command terminated abnormally. (exit 1)
✖ [1/1] attempt failed: tung.lab.uiowa.edu — [warning] Drush command terminated abnormally. (exit 1)
```

cory-skeers (approved):
Reviewed and tested, and worked great! I had a few issues getting the ssh keys to work and the host properly added as a trusted source, but that should be on my end and not the updates.

One part that stood out to me was:
```
  public static function isReportable(object $application): bool {
    return $application->organization->name !== self::EXCLUDED_ORGANIZATION;
  }
```
I think it'd usually be useful to have this defined as `EXCLUDED_ORGANIZATIONS` and set up to have the option of none, one, or many. But the infrastructure isn't likely to change often, and even if it did, would be an straightforward change when an update would already be necessary.

So all that to say--something I'd thought about so wanted to note, but don't believe a change is needed.

PR #10036, opened 2026-07-30
Remove ddev gate and move errors to stderr.

### LINKED ISSUE #10028: Update site counts jobs (v2/v3)
# Motivation

The site counts scheduled jobs that run on each app are no longer working since they are using the old `blt ume` command. These have been disabled until we can get them updated. We should update these similar to the cron scheduled jobs so we don't have to manually get site counts.

# Additional Info

From Joe:

```
COUNT=$(/var/www/html/${AH_SITE_NAME}/vendor/bin/blt ume "cget config_split.config_split.sitenow_v2 status" | grep ": false" | wc -l); php -r "mail('osc-web@uiowa.edu', 'SiteNow v3 count', 'uiowa09: ' . $COUNT, 'From: noreply@uiowa.edu');"
```

I've disabled these jobs from triggering since they will fail in the current state. We could fetch these manually as a short term solution.

Cron scheduled jobs PR - https://github.com/uiowa/uiowa/pull/10020

GaryRidgway:
- Removes the ddev gate from unnecessary commands and moves errors to stderr
- Update site counts jobs (v2/v3) #10028 

### How to test
**1. Warm up the app's SSH connection (run once per app):**
`./sn me --apps=uiowa07 -y --concurrency=1 -- status`
Expected: N/N succeeded, 0 failed.

**2. v3 count:**
```
COUNT=$(./sn me --apps=uiowa07 -y -v -- config:get config_split.config_split.sitenow_v2 status 2>/dev/null | grep ": false" | wc -l)
echo "v3 count: $COUNT"
```
Expected: a number between 0 and the app's total site count. As of 8/04 should be 109

**3. v2 count:**
```
COUNT=$(./sn me --apps=uiowa07 -y -v -- config:get config_split.config_split.sitenow_v2 status 2>/dev/null | grep ": true" | wc -l)
echo "v2 count: $COUNT"
```
Expected: v2 count + v3 count should equal the app's total site count from step 1. If not, re-run before trusting either number.

### Other checks

**4. ddev gate is gone:**
`./sn me --apps=uiowa07 --dry-run -- cr`
Expected: dry-run job list prints, no DDEV error.

**5. Still works inside DDEV:**
`ddev exec ./sn me --apps=uiowa07 --dry-run -- cr`
Expected: same site list as step 5, using DDEV's own PHP/drush paths.

**6. Deprecation notices don't leak onto stdout:**
`./sn me --apps=uiowa07 --dry-run -- cr 1>/dev/null`
Expected: nothing prints (or only Deprecated: lines) — the dry-run job list must not appear.

**7. Automated tests:**
```
vendor/bin/phpunit tests/phpunit/src/Unit/MultisiteExecuteTest.php
vendor/bin/phpunit tests/phpunit/src/Unit/FleetRunnerTest.php
```
Expected: 12 tests pass for the first, 22 for the second (34 total).

---
### After Deployment
In Acquia, edit and re-enable the Scheduled Jobs for each application with the following

**site count V3**
```
COUNT=$(/var/www/html/${AH_SITE_NAME}/sn me -y -v -- config:get config_split.config_split.sitenow_v2 status | grep ": false" | wc -l); php -r "mail('osc-web@uiowa.edu', 'SiteNow v3 count', '${AH_SITE_GROUP}: ' . $COUNT, 'From: noreply@uiowa.edu');"
```

site count v2
```
COUNT=$(/var/www/html/${AH_SITE_NAME}/sn me -y -v -- config:get config_split.config_split.sitenow_v2 status | grep ": true" | wc -l); php -r "mail('osc-web@uiowa.edu', 'SiteNow v2 count', '${AH_SITE_GROUP}: ' . $COUNT, 'From: noreply@uiowa.edu');"
```


GaryRidgway:
@bewoldt what made the force push necessary here?

bewoldt:
> @bewoldt what made the force push necessary here?

From Claude
The remote remove-ddev-gate had a merge commit (70eac4845) merging main into the branch, plus Alan Way's two original commits. To resolve the conflicts (from main having advanced past that merge point via [#10024](https://github.com/uiowa/uiowa/pull/10024)), I rebased those two commits directly onto current main instead — which replays them as new commits with new SHAs (321beec82, ee5585c45), dropping the merge commit and the now-dead DrushRunner.php changes along the way.

Once local history no longer shared a tip with the remote's history, a normal git push is rejected — it only allows fast-forwards (adding commits on top of what's already there), and rebasing produces the opposite: different commits standing in for the old ones. Force-pushing was the only way to make the remote branch point at the rewritten history instead of the original.

GaryRidgway:
Ah, was that something that was done locally instead of here? My confusion came from what might be what is more commonplace on the team. I think I see people merging main into the branch in the GitHub web interface and then pulling it down instead of doing it locally. Or maybe I am still misunderstanding, lol. Either way, seems fine.

bewoldt:
@GaryRidgway Thanks. I'll do it that way in the future.

joewhitsitt:
pulling to test

joewhitsitt:
Visibility: --apps is pinned to the running app on AC, but --env isn't. It still defaults to PROD regardless of which environment this is running on, so the only thing stopping a DEV ume from targeting PROD is the absence of an ssh agent, not an explicit check. so an accidental safeguard.

joewhitsitt:
Updated the after deployment snippet for v3 to be grep false and labelled SiteNow v3

Screenshot 2026-08-11 at 10 59 43 AM

joewhitsitt (approved):
looks good to me. the comments (e.g. On Acquia Cloud...) are a bit dense for inline help but I am wary of simplifying if the wording as-is proves to be helpful to someone in the future.

joewhitsitt:
>The old blt ume streamed per-site drush output to stdout by default. The new sn me only outputs per-site results with -v, and pipes inside $() don't work in the Acquia scheduled job context (though the same command works fine interactively). The fix is to write sn output to a temp file first, then grep the file. no pipes.

The following does this and stays under the 255 character length.

v3
```
S=/var/www/html/$AH_SITE_NAME; $S/sn me -y -v -- cget config_split.config_split.sitenow_v2 status >/tmp/s 2>/dev/null; C=$(grep -c false /tmp/s); php -r "mail('osc-web@uiowa.edu','SiteNow v3 count','$AH_SITE_GROUP: '. $C,'From: noreply@uiowa.edu');"
```

v2
```
S=/var/www/html/$AH_SITE_NAME; $S/sn me -y -v -- cget config_split.config_split.sitenow_v2 status >/tmp/s 2>/dev/null; C=$(grep -c true /tmp/s); php -r "mail('osc-web@uiowa.edu','SiteNow v2 count','$AH_SITE_GROUP: '. $C,'From: noreply@uiowa.edu');"
```

Trying to run this on each application however is erroring with some preflight errors on some sites. We might have to switch to a script to workaround sites that produce issues.

uiowa
```
`config:get` failed a preflight check on [accel.education.uiowa.edu](http://accel.education.uiowa.edu/) — include() at /var/www/html/vendor/bin/drush:119 (exit 1)
```

uiowa04
```
`config:get` failed a preflight check on [accohpsychotherapy.org.uiowa.edu](http://accohpsychotherapy.org.uiowa.edu/) — include() at /var/www/html/vendor/bin/drush:119 (exit 1)
```

uiowa05
```
`config:get` failed a preflight check on [abboud.lab.uiowa.edu](http://abboud.lab.uiowa.edu/) — include() at /var/www/html/vendor/bin/drush:119 (exit 1)
```

Context: https://iowaweb.slack.com/archives/C0164FNN1PV/p1787231805925239

Issue #6739, opened 2023-08-03
Consider removing tours from SiteNow


joewhitsitt:
# Problem

They are becoming out of date and are under used. Core is giving it the boot. https://www.drupal.org/node/3376405

>The Tour module is slated for removal from core in Drupal 11, based on data that indicate that the module is not commonly used by the intended audience.

>To this end, the Tour module has been removed from the Standard profile in Drupal 10.2.0. This change does not affect existing sites, only new sites installing the Standard profile for the first time. Drupal 11.x sites may still enable the module, and it will be available as a contributed module for Drupal 11 sites, although it is recommended to consider alternatives for the long term.

# Proposed Solution

- [x] Remove tour config, disable the module
- [x] Replace with updated docs on sitenow.uiowa.edu
- [ ] Create a "quick start" training video that covers the same bits.

joewhitsitt:
With Signage being added to the SiteNow profile, more of this tour stuff will either be off limits or not applicable. I am in favor of removing it.

joewhitsitt:
I am not even seeing the "?" toolbar menu item show anymore where it is expected (e.g. content overview). I was looking at it because some google analytics docs will change as part of https://github.com/uiowa/uiowa/issues/8485 and these have not been kept up to date.

briand44:
## Grooming
 
**Quadrant:** 

Clear

 **Discussion:**
 - Haven't touched tours really since early v2 days
 - Add `?tour` to a path to activate (people config page, content overview, etc.)
 - May want to make sure we have documentation for information in tours that will be removed
 - Don't believe any training videos mentioning/demonstrating tours (Brian to check with Megan)
 - Tours based on selectors so maintenance can be an issue as selectors change, etc.
 - Currently part of core so should be able to remove config and disable in same PR
 - contrib tours will just go away once we disable the module
 - Not currently showing in toolbar as expected

**Proposed solution or definition of done:** 
- tour config removed
- tour module disabled
- docs updated mentioning tours (https://sitenow.uiowa.edu/documentation)
- docs added if needed to account for info in tours not already documented

pyrello:
Picking this up.

joewhitsitt:
was there anything identified that needed to be captured in the published docs?

I removed, "Also on the toolbar, depending on what page you are on, you will see a question mark icon on the far right. This indicates a click-through tour is available to assist you with managing your website." from https://sitenow.uiowa.edu/documentation

bewoldt:
[...]

PR #10048, opened 2026-08-04
Remove tours from SiteNow

pyrello:
Removes the Tour module and its config from SiteNow. Refs #6739. The documentation half of that issue is separate, so this does not close it.

Tip copy was extracted before deletion, so the documentation pass has a source.

The ITS `help_desk` role split also drops `access tour`, which is easy to miss among the deletions.

## How to test

- Check out this branch and import config:

```
git checkout 6739_remove_tours
ddev drush @sandbox.local cim
```

Expected: `Synchronized extensions: uninstall tour.` and a successful import.

- Confirm the module is gone:

```
ddev drush @sandbox.local pm:list --status=enabled --filter=tour
```

Expected: no rows.

- Log in and open the content overview page and add `?tour`:

```
ddev drush @sandbox.local uli /admin/content
```

Expected: no tour pops up.

---

*Drafted by Claude and heavily edited by @pyrello*


bewoldt:
pulling down to test.

bewoldt (approved):
Approved.

PR #10054, opened 2026-08-06
Bump php_codesniffer to 3.13.6

pyrello:
Bumps `squizlabs/php_codesniffer` to 3.13.6. Lock-only, since nothing constrains it directly in `composer.json`.

# How to test
All tests pass. Nothing in our code calls this so nothing to test.


e-marie-w (approved):
Thanks for putting this together!

PR #10055, opened 2026-08-06
Migrate blt umd

### LINKED ISSUE #10011: Replace blt umd command
# Motivation

As part of our strategy to move away from BLT, we should implement a replacement for the blt umd (uiowa:multisite:delete) command so that we continue remove sites no longer needed without the risks of doing this manually.

# Additional Info

- Replace `blt umd` command with a non-BLT replacement
- New command should account for everything existing command does
- Any cleanup related to `blt umd` that we are comfortable with

GaryRidgway:
Tracks #10011 

# How to test

## Setup

1. Pull the branch, `composer install`. 
2. `./sn list` 
    - Verify `multisite:delete` appears with aliases `md`, `umd`.
3. `./sn umd --help` 
    - Verify it documents that cloud comes down before the repo, that an already-absent resource is a WARN rather than a failure, and that a failed run is safe to run again.
 4. Verify that [sitesbar.prod.drupal.uiowa.edu](sitesbar.prod.drupal.uiowa.edu) exists and can be hit.

## It refuses what it should

5. `./sn umd bogus.uiowa.edu --dry-run`
    - `[FAIL] Site bogus.uiowa.edu is not in blt/manifest.yml. Nothing to delete.` exit `1`, no Actions block.
6. `./sn umd demo.sitenow.uiowa.edu --dry-run`
    - `[FAIL] demo.sitenow.uiowa.edu resolves to the shared docroot/sites/default directory, which this command will not delete.` exit `1`.
    - This host **is** in the manifest, and `sites.php` maps it to `default`. Without this guard the plan would target `docroot/sites/default`, the `uiowa` application's own database, and the default site's files on all three mounts.

## The plan, before touching anything

7. `./sn umd bar.sites.uiowa.edu --dry-run`
	-  Verify that you see the following:

```
  Application:   uiowa09
  Directory:     docroot/sites/bar.sites.uiowa.edu
  Database:      bar_sites_uiowa_edu
  Drush alias:   drush/sites/sitesbar.site.yml
  Cloud files:   dev, test, prod
  Cloud DB:      present
  Cloud domains: sitesbar.dev.drupal.uiowa.edu (dev), sitesbar.prod.drupal.uiowa.edu (prod), sitesbar.stage.drupal.uiowa.edu (stage)

  Validation: PASS
```

8. After this block, it will list 12 actions that a real run would take: 
	- three file deletes
	- the database
	- three domains
	- four repository removals
	- the commit. 
	- **Verify every cloud action precedes every repository action** — that ordering is what makes a failed run retryable.

9. Run it interactively and WITHOUT TYPING ANYTHING press Enter at the `Apply?` prompt.
    - `Aborted.`, exit `0`, `git status` still clean. The prompt defaults to no.

## The real run

10. `./sn umd bar.sites.uiowa.edu`, answer `yes`. 
    - Verify a line per action, `[OK] Done.`, a next-step pointing at `git push`, and a commit `Delete bar.sites.uiowa.edu multisite on uiowa09`.
11. `git show --stat HEAD`
    - Verify the site directory, `drush/sites/sitesbar.site.yml`, four `sites.php` aliases and the `manifest.yml` entry are all gone.
12. `./sn umd bar.sites.uiowa.edu --dry-run`
    - Verify that you see `[FAIL] … not in blt/manifest.yml`, exit `1`. It can't be run twice.

## The resume path

13. `git tag umd-deleted && git revert umd-deleted --no-edit`
	- The tag marks the delete commit from step 8 so the cleanup at the end needs no hash. 
14. `./sn umd bar.sites.uiowa.edu --yes`
    - Verify `Aborting: --yes was passed but validation has a WARN.` exit `1`
    - Unattended runs don't get to guess.
15. `./sn umd bar.sites.uiowa.edu`
    - Verify three WARNs (files, database, domains all absent), and that Actions contains only the repository steps. Answer `yes`; verify it finishes the job.
16.  `git reset --hard umd-deleted && git tag -d umd-deleted`
	- Leaves a single deletion commit and removes the tag.

## Removing `blt umd`

17. `./vendor/bin/blt list`
    - Exits `0`. No `uiowa:` commands remain, and no error or missing-class warning. BLT discovers commands by scanning the plugin directory, so a clean boot is what proves the deletions left nothing dangling.
18. `./vendor/bin/blt umd`
    - Verify that you see `Command "umd" is not defined.`
19. `./vendor/bin/blt drupal:sync:default:site`
    - Verify that you see `[error] This command is no longer supported. Use 'ddev sn ds SITE' instead.` exit `1`.
20. `git grep -n "MultisiteTrait\|AcquiaCloudApiTrait\|uiowa:multisite:delete" -- ':!vendor'`
    - _(You can use Shift+Q to exit.)_
    - Verify you only see a single line in the new command's docblock, [[MultisiteDeleteCommand.php:36](vscode-webview://0p79r9bq0rp57rvr41nkotflrihe6ri6ck6cditq1vvh8eb1vjpg/sitenow/src/Command/MultisiteDeleteCommand.php#L36)](vscode-webview://0p79r9bq0rp57rvr41nkotflrihe6ri6ck6cditq1vvh8eb1vjpg/sitenow/src/Command/MultisiteDeleteCommand.php#L36). No code references.

## Notes

`bar.sites.uiowa.edu` and `www.bar.sites.uiowa.edu` are also searched for as domains ([[MultisiteDeleteCommand.php:440](vscode-webview://0p79r9bq0rp57rvr41nkotflrihe6ri6ck6cditq1vvh8eb1vjpg/sitenow/src/Command/MultisiteDeleteCommand.php#L440)](vscode-webview://0p79r9bq0rp57rvr41nkotflrihe6ri6ck6cditq1vvh8eb1vjpg/sitenow/src/Command/MultisiteDeleteCommand.php#L440)) but aren't registered on Acquia, so nothing lists them and nothing tries to delete them. That branch is unexercised here.


pyrello:
Pulling to test.

GaryRidgway:
@pyrello When you delete the bar site during this set of steps, I encourage you to commit it to this branch.

pyrello:
```
./sn umd bogus.uiowa.edu --dry-run

─────────────────────────────────────────────────────────────────
  multisite:delete bogus.uiowa.edu
─────────────────────────────────────────────────────────────────

  Validation: FAIL
    ✗ [FAIL] Site bogus.uiowa.edu is not in blt/manifest.yml. Nothing to delete.
    ✗ [FAIL] Site directory docroot/sites/bogus.uiowa.edu does not exist, but bogus.uiowa.edu is in the manifest.
```
The second `[FAIL]` message contradicts the first about whether the site is in the manifest. Probably it shouldn't say anything about the manifest since that's not its purview?

pyrello:
```
./sn umd demo.sitenow.uiowa.edu --dry-run

─────────────────────────────────────────────────────────────────
  multisite:delete demo.sitenow.uiowa.edu
─────────────────────────────────────────────────────────────────

  Application:   uiowa
  Directory:     docroot/sites/default
  Database:      uiowa
  Drush alias:   drush/sites/sitenowdemo.site.yml

  Validation: FAIL
    ✗ [FAIL] demo.sitenow.uiowa.edu resolves to the shared docroot/sites/default directory, which this command will not delete. Remove its sites.php alias and manifest entry by hand.
    ✗ [FAIL] drush/sites/sitenowdemo.site.yml is missing a user for: dev, test, prod. The files mount path cannot be resolved.
```
What does `[FAIL] drush/sites/sitenowdemo.site.yml is missing a user for: dev, test, prod` mean?

pyrello:
This PR (and several before it) include comments that:
- Are not focused on the appropriate scope for the code being commented. Either commenting on other sections of code or the wrong altitude (docblock comments instead of inline or class docblocks instead of method docblocks). Either way, a surface for drift because they are less likely to be reviewed if the code changes.
- Are narrative or reactive in nature and reference the journey rather than the current state of the code.
- Are backward-looking and reference BLT as a rationale for how the code works when the whole point of #8181 is to move beyond BLT.

Examples:
`CloudDbDelete.php:14` defines current behavior by contrast with deleted code:
> Unlike umd, which fired the request and moved on, this returns only once the
> database is confirmed gone

`SitesPhpRemove.php:12`:
> a block whose formatting drifted is not silently left behind — the failure
> mode of BLT's umd, which replaced one literal block and did nothing if it had
> changed.

`MultisiteDeleteCommand.php:36`, matching `MultisiteExecuteCommand.php:20` and
`SiteInstallCommand.php:21`:
> Replaces the BLT `uiowa:multisite:delete` (umd) command.

`MultisiteDeleteCommand.php:426`:
> `www.<host>` is included because it is registered for many sites and is not
> derivable from the identifier, so umd never removed it.

pyrello:
While reviewing this PR, I noticed the proliferation of *7* new class files in the `SiteNow/Operation` namespace. These were atomic-ish classes mostly dedicated to a single operation. Meaning that, they potentially each had to duplicate logic for accessing the resource that they manipulated. This stems back to https://github.com/uiowa/uiowa/pull/9871 when the UMC command was originally being rewritten using Robo. These atomic classes were originally meant to extend `Robo/Task/BaseTask` and their structure made sense in that context. When that work pivoted towards Symfony Console, I thought I had refactored it in a way that was in keeping with how Drupal Core handled Symfony Console commands. But the `Tasks` got converted into `Operations` and they have been proliferating ever since. Rather than continue to accept this non-standard pattern that fragmented logic across many classes, I decided instead to clean up the original mess I made and help pivot instead of resource-based classes (e.g. Manifest.php vs ManifestUpdate.php), which is better aligned with how Drupal Core handles similar needs. This mostly works without expanding the scope of the PR, as new classes that would have been introduced in this PR instead land as methods on resource classes.

pyrello:
Second pass at `17732bbb3`. Suite green (104 tests, up from 81), `phpcs
--standard=phpcs.xml` clean. The probe rework in `0b71fb853` closes the one
finding that could lose data, and `ec7a681a4` fixes a `preg_replace` NULL
truncation I had missed.

**`Acquia/Mounts.php:165` reports successful deletes as failures.** The post-`rm`
probe throws if the directory is present, but Acquia reprovisions it within the
minute, so a delete that worked aborts the run with the database and domains
still live. It cannot tell reprovisioning from an `rm` that removed nothing, so
remove it and trust the exit status. `testDeleteIsConfirmedByProbingAgain` and
`testDeleteRefusesToReportSuccessWhileFilesRemain` go with it.

**Small fixes:**

* `MultisiteDeleteCommand.php:737` says files "cannot be recreated from the
  repository." True of the content, false of the directory, which is rebuilt
  from `sites.php`.
* `SitesPhpTest::testRemoveAliasesStripsTheSiteBlock` asserts only
  `assertStringNotContainsString('doomed', …)`, so it would pass on a truncated
  file, which is the bug `ec7a681a4` just fixed.
* `CloudApi::confirmAbsent()` and `awaitOperation()` are untested. Both are
  polling loops with timeout and terminal-status branches, and `confirmAbsent()`
  already takes an injected callable. `MountsTest`'s scripted-process seam is
  the template.
* `Config/Manifest::path()` and `::sites()` have no callers, while
  `SiteNowCommandsTrait::manifest()` at `:229` parses the file itself. Two
  readers of one file.

**The help text needs an editing pass.** Specifics:

* "come down" is doing the work of "deprovision" or "delete" in two places,
  `:108` ("The cloud comes down first") and `:126` ("What would come down?").
  It is a vague euphemism for the thing the command is named after.
* `:105` "its **directories**, drush alias, sites.php aliases and manifest
  entry" is plural where there is one site directory.
* `:108` "every cloud **delete** is confirmed gone" describes the resource,
  not the delete.
* `:111` attaches "the state a half-finished delete leaves behind" to "not a
  failure" with an em-dash, so it reads as though the failure is that state
  rather than the absent resource. The only em-dash in user-facing text here.
* `:115` "Files are deleted per environment, the site's whole directory rather
  than the contents of `files/`" joins two separate facts with a comma.
* `:123` "an Acquia SSH key has to be available **there** too" leaves "there"
  without a clear referent.

Happy to draft a rewrite if that is useful.

pyrello (approved):
This works as expected. Nice work, @GaryRidgway

PR #10068, opened 2026-08-12
bump diff to 2.1.1, externalauth to 2.0.13

joewhitsitt:
Diff changelog: https://git.drupalcode.org/project/diff/-/compare/2.0.0-beta4...2.1.1
Externalauth changelog: https://git.drupalcode.org/project/externalauth/-/compare/2.0.8...2.0.13

Datebase update:

```
externalauth   8104        hook_update_n   8104 - Make authmap authname and provider columns binary-safe. Converts both    
                                             columns from their previous case/accent-insensitive collations to binary-safe   
                                             equivalents so that byte-distinct authnames and provider values are matched     
                                             exactly by the database. NOTE: If your site has existing rows that differ only  
                                             in case or accents for the same provider (i.e. were inserted while a ci         
                                             collation was active and the unique key was not enforced), this update may      
                                             fail. Resolve duplicate rows manually before running this update.   
```

# How to test

- Diff interface still works as expected? That our additional block/entity detail is displayed within layout builder?
  - https://admissions.dev.drupal.uiowa.edu/node/151/revisions/view/66466/74181/split_fields, https://admissions.uiowa.edu/node/151/revisions/view/66466/74181/split_fields
- Deploy out to develop to make sure authentication still works as expected?
  - https://github.com/uiowa/uiowa/pull/10068#issuecomment-5273366880


joewhitsitt:
this has been deployed to develop for authentication testing and so far it seems to be working as expected

joewhitsitt:
Looking a bit more into externalauth's update and from what I understand (not much) i think we are good, samlauth has been the sole provider since we started, hawkids are all lowercase, no chance for case-sensitive collisions.

GaryRidgway:
Pulling to test

GaryRidgway (approved):
Compared dev to live and it seems good! approved.

PR #9889, opened 2026-06-02
Update scheduler and scheduler_content_moderation_integration

cory-skeers:
Resolves #9748 

# How to test

For the most part, these updates are bug fixes and code cleanup. No update hooks, no config changes.

Patch for scheduler_content_moderation_integration is still needed, and applies cleanly

```
ddev composer install
```
Sync some sites of your choice. Scheduler is used on default, alerts feature, and certain sites like signage

## Content moderation w/scheduling conflicts:
The primary conflict comes in when scheduler is set to un/publish a piece of content at a future date, but content moderation is attempting it to the same unpublished or published state (either through the node form or a programmatic save, such as using views bulk options).

For scheduled publish, any piece of content with scheduling can run into this issue.
To test:

- Sync a site
  - create a new piece of content, but don't save yet
    - set a future publish date
    - set it to "publish"
    - save, and see that it gets set to "draft" and gives a warning message
  - Edit an existing piece of content
    - set a future publish date
    - set it to "publish"
    - save, and see that it gets set to whatever state it had been in (draft, review) and gives a warning message
  - Go to the content overview
    - select a piece of content that has a future publish date (and others if you'd like)
    - Bulk action select "publish latest revision"
    - Apply to selected items
    - See a myriad of messages including a confusing error message...not ideal, but matches current prod behavior
    - See that the selected piece of content which has a future publish date is still unpublished

For scheduled unpublish, that is only used on its.uiowa.edu for _News Article_ content (other content like slides can be scheduled unpublish, but don't use content moderation, so don't have the same issues)

```
ddev blt ds --site=its.uiowa.edu
```
Repeat the above tests, but using a future unpublish date and setting to "unpublished"

## now.uiowa.edu embargo blocking

No changes seem to interact with this, but it's separate functionality.
Whether trying to publish directly or through scheduled publish, if there is an embargo message on a news article, it should _not_ be published.
Testing involves creating an article with an embargo message and scheduled publish time
After the scheduled time, run cron
You should see a log message along the lines of: "Publishing of "New test article" is prevented until after the content embargo has been lifted."

PR #10087, opened 2026-08-21
uiowa_core scheduler moderation fix

### LINKED ISSUE #9924: TypeError in uiowa_core_scheduler_moderation_validate
# Problem
```
Exception 'TypeError' with message 'strtotime(): Argument #1 ($datetime) must be of type string, array given' in /var/www/html/docroot/modules/custom/uiowa_core/uiowa_core.module:1272
                                               in strtotime called at /var/www/html/docroot/modules/custom/uiowa_core/uiowa_core.module (1272)
                                                                               /in uiowa_core_scheduler_moderation_validate called at ?
```

# Background

https://iowaweb.slack.com/archives/CSQA18BK4/p1781289450336479

# Steps to reproduce

# Proposed solution

joewhitsitt:
Resolves https://github.com/uiowa/uiowa/issues/9924

I believe this is related to a halo ticket (0244838) and the the following error occurs quite a bit:

```
TypeError: strtotime(): Argument #1 ($datetime) must be of type string, array given in strtotime() (line 1267 of /var/www/html/docroot/modules/custom/uiowa_core/uiowa_core.module)
```

[See splunk](https://splunk.its.uiowa.edu/en-US/app/search/search?earliest=-30d%40d&latest=now&q=search%20index%3Duiowa-acquia%20%22TypeError%3A%20strtotime()%3A%20Argument%20%231%20(%24datetime)%20must%20be%20of%20type%20string%2C%20array%20given%20in%20strtotime()%20(line%201267%20of%20%2Fvar%2Fwww%2Fhtml%2Fdocroot%2Fmodules%2Fcustom%2Fuiowa_core%2Fuiowa_core.module)%22&display.page.search.mode=smart&dispatch.sample_ratio=1&sid=1787342003.43899)

The scheduler datetime widget can hand back an array (`['date' => '', 'time' => '', 'object' => NULL]`) or a `DrupalDateTime` object instead of a string for `publish_on`/`unpublish_on`, and the old code passed that straight to `strtotime()`. Also switched off reading `publish_on`/`unpublish_on` from the entity, since during validation the entity can still hold a stale scheduled date after a user clears it on the form.

Couldn't get a reliable browser-level repro locally (tried masquerading as the reporting user on a synced copy of cyndari.lab.uiowa.edu, no luck), even though it's still firing in production against the same node. Added a unit test instead that calls the validate function directly with the exact shapes seen in the wild.

# How to test

```
ddev sn ds cyndari.lab.uiowa.edu && ddev drush @labcyndari.local uli
```

Edit node 151, leave "Publish on" blank, set moderation state to Published, save. Shouldn't crash — but see note above, this hasn't reproduced locally even on unpatched code.

To see the regression test actually catch the bug, run it against `main` first, then against this branch:

```
git checkout main
git checkout uiowa_core_scheduler_moderation_fix -- docroot/modules/custom/uiowa_core/tests/src/Unit/SchedulerModerationValidateTest.php
ddev exec -s web "cd docroot/core && ../../vendor/bin/phpunit --testsuite unit --filter SchedulerModerationValidateTest ../modules/custom/uiowa_core/tests/src/Unit/SchedulerModerationValidateTest.php"
```
fails: TypeError at uiowa_core.module:1267, matching the splunk log above

```
git checkout uiowa_core_scheduler_moderation_fix
ddev exec -s web "cd docroot/core && ../../vendor/bin/phpunit --testsuite unit --filter SchedulerModerationValidateTest ../modules/custom/uiowa_core/tests/src/Unit/SchedulerModerationValidateTest.php"
```
passes: 10/10

## Didn't break the original scheduler/moderation conflict behavior (#9889)

This function's actual job — falling back the moderation state when it conflicts with a scheduled date — is the thing #9889 added and tested manually. The test file now covers that too (future publish_on → falls back new content to draft, existing content to its prior state; future unpublish_on → same; past dates aren't a conflict), so it's covered by the same PHPUnit run above.

Still worth a manual pass per #9889's original steps, since only the node-form path is unit tested:
- Edit an existing published/in-review article, set a future publish date, set state to Published, save → should fall back to its prior state with a warning message.
- Same via the content overview bulk action ("Publish latest revision").


joewhitsitt:
accidentally had "auto mode" on in Claude when i asked about "some good tests from the last time this functionality (#9889) was touched to ensure no regressions."

cory-skeers (approved):
Reviewed and tested, and all looks good. Seems reasonable that this is the issue seen in the Halo ticket, and it at least seems to be cropping up.

Glad for the extra test coverage, however it came about! I had Claude take a pass at abstracting some of the comments away from references to specific issues/work, and describing each test in isolation (and then a pass myself at editing its edits...).

I went ahead and added the comment changes, but feel free to revert or edit further before merging.

joewhitsitt:
Thanks @cory-skeers 

🤞

Issue #10082, opened 2026-08-20
Site counts reporting zero


briand44:
https://iowaweb.slack.com/archives/C0164FNN1PV/p1787231805925239

More info:
https://github.com/uiowa/uiowa/pull/10036#issuecomment-5357954685

briand44:
## Grooming
 
**Quadrant:** 

Complicated

 **Discussion:**
 - subtle differences in how old `ume` worked vs `sn me`
 - Joe was able to get it to work with under the character limit
 - Trying to run in prod, was running into issues with any non-standard output
 - Some refining necessary
 - may need to move to a script, similar to cron job to give us more freedom with character limit
 - Some preflight issues on a few random sites
 - ddev gate worked/ was triggering

**Proposed solution or definition of done:**
- Command to get site counts has been refined or moved to a script and returns actual site counts for a given application
- Scheduled jobs have all been updated to use newly refined command or script
- Tested to confirm scheduled jobs run and accurately report site counts for a given application

cory-skeers:
#10092 has been deployed, and the jobs updated on Acquia. Should be able to see if it all works as expected tomorrow 🤞

PR #10092, opened 2026-08-25
Sn report counts command

cory-skeers:
Relates to #10082 

# How to test

Try out the new command:
`./sn report:count` alias `./sn count`
Couple options are available, like other sn commands:
`--apps=...` to limit to specific applications
`--totals-only` to get only the number of sites, with no v2/v3 checking (speedy; checks only against the manifest)

Verify accurate counts against recent counts, which can be found via the demo deck. v2 should largely have been unchanged, though the v3 counts may vary slightly since the counts were last updated


cory-skeers:
Untested, but I think with this the direct in Acquia scheduled jobs could be:

apps with v2/v3 mix:
```
O=$(/var/www/html/${AH_SITE_NAME}/sn count 2>&1);E=$?;S="SiteNow count";[ $E -eq 0 ]||S="$S FAILED";S="$S" B="$O" php -r "mail('osc-web@uiowa.edu',getenv('S'),getenv('B'),'From: noreply@uiowa.edu');"
```

apps with only v3:
```
O=$(/var/www/html/${AH_SITE_NAME}/sn count --totals-only 2>&1);S="SiteNow count" B="$O" php -r "mail('osc-web@uiowa.edu',getenv('S'),getenv('B'),'From: noreply@uiowa.edu');"
```

Short variable names make them a little less human-readable, but keeps them under the character limits. The v2/v3 mix will update the email subject line to help indicate failures, with what detail there is being included in the body. The v3-only one doesn't need to check for error, as it only deals with the manifest, and so will never 'fail' unless the whole thing is broken.

The `${AH_SITE_GROUP}` isn't used, but the body of the email uses the output from the command, which includes the application designation.

joewhitsitt:
I know it isn't in scope, but I was curious what v2 site is left on uiowa04

`uiowa04: 109 sites total (1 v2, 108 v3)` so close to just two apps with v2


```
      ->addOption('apps', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated app names to report on (e.g. uiowa02,uiowa03). Defaults to all apps; pinned to the running application on Acquia Cloud.', '')
      ->addOption('exclude', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated site domains to skip.', '')
```

The command already handles failed sites correctly without --exclude. A failed site goes into the failed list and is excluded from the v2/v3 breakdown automatically. The --exclude would let you skip a domain before the the command hits it, but since failures don't corrupt the count, what is the use-case to pre-exclude?

cory-skeers:
> ```
>       ->addOption('apps', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated app names to report on (e.g. uiowa02,uiowa03). Defaults to all apps; pinned to the running application on Acquia Cloud.', '')
>       ->addOption('exclude', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated site domains to skip.', '')
> ```
> 
> The command already handles failed sites correctly without --exclude. A failed site goes into the failed list and is excluded from the v2/v3 breakdown automatically. The --exclude would let you skip a domain before the the command hits it, but since failures don't corrupt the count, what is the use-case to pre-exclude?

Good question. I carried it over from the Report Splits command I was using as a reference. But since it's reporting on an aggregate count in this case, I don't think there is one--any site you'd exclude just means add one to the total count you get back. The main case would be if we expected it to fail, but like you point out, that gets handled already.

sitenow/src/Command/ReportSiteCountCommand.php
```diff
@@ -0,0 +1,239 @@
...
+    parent::__construct();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure(): void {
+    $this
+      ->addOption('apps', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated app names to report on (e.g. uiowa02,uiowa03). Defaults to all apps; pinned to the running application on Acquia Cloud.', '')
+      ->addOption('exclude', NULL, InputOption::VALUE_REQUIRED, 'Comma-separated site domains to skip.', '')
```
cory-skeers:
```suggestion
```

cory-skeers:
> I know it isn't in scope, but I was curious what v2 site is left on uiowa04
> 
> `uiowa04: 109 sites total (1 v2, 108 v3)` so close to just two apps with v2

Out of curiosity, I ran the splits report:

```
== sitenow_v2 ==
 ------------- ------------------------ 
  Application   Domain                  
 ------------- ------------------------ 
  uiowa04       our.research.uiowa.edu  
 ------------- ------------------------ 
```
So close!

joewhitsitt:
Ah. Forgot we had that command. Thanks!

I see that you decided to remove the exclude. Once the rest of that is cleaned up, I am happy to approve and put this to the test in PROD. The commands as written have several characters to spare

joewhitsitt (approved):
Updates look good. Thanks.

Issue #7031, opened 2023-10-23
Upgrade typhonius/acquia-php-sdk-v2 to 3.x


joewhitsitt:
It might be worth checking if we can make the jump to the newest version of this library. Here is full diff:
https://github.com/typhonius/acquia-php-sdk-v2/compare/2.0.15...3.0.9

There are a couple breaking changes in the 3.x branch, but it seems like those may not affect us or would be easy enough to update our code to fix: https://github.com/typhonius/acquia-php-sdk-v2/releases/tag/3.0.0

_Originally posted by @pyrello in https://github.com/uiowa/uiowa/issues/5738#issuecomment-1589443384_

PR #10106, opened 2026-09-01
Final BLT removal

### LINKED ISSUE #5359: Incorrect stage_file_proxy settings after site launch
# Problem

Our BLT setup is supposed to set the stage_file_proxy origin URL to the site's prod. ~~Instead, new sites seem to be getting `default.prod.drupal.uiowa.edu` always~~ - _this was fixed_

Once a website goes live, the stage_file_proxy settings are not updated and are are overwritten each time the site provision command is run.

# Workaround

- Manually override the local.settings.php associated with the site.

# Proposed Solution

- Detect that the website no longer has an internal production domain and switch to their live URL.

### LINKED ISSUE #9858: Finalize BLT replacement (cleanup + replace BLT settings cascade)

### LINKED ISSUE #10033: multisite:create --dry-run does not show the options that shape the site
## Problem

`sn multisite:create --dry-run` prints the plan, but not the option values that shape the site it would create. A dry run with `--site-name`, `--requester`, and `--split` renders the same plan as one without them:

```
  multisite:create testsite.org.uiowa.edu

  Application:   uiowa09
  Database:      testsite_org_uiowa_edu
  Reason:        Fewest sites (87) among eligible apps.

  Validation: PASS

  Actions:
    · Create cloud DB testsite_org_uiowa_edu on uiowa09
    · Copy docroot/sites/default → docroot/sites/testsite.org.uiowa.edu
    ...
    · Write docroot/sites/testsite.org.uiowa.edu/blt.yml
```

Those values land in the site's `blt.yml`, which a dry run does not write, so there is nothing to inspect either. The result is that a mistyped or dropped option cannot be caught before the real run.

Surfaced by @joewhitsitt while testing #10018, where a dry run could not confirm that `--site-name` was being passed.

## Open questions

- Which values the plan should show. Site name, requester, and splits are the ones that reach `blt.yml`; the app and database are already displayed.
- Whether to show them as their own lines, or as a preview of the `blt.yml` that would be written.
- Whether an option that will be ignored should say so, for example `--site-name` given with a value that ends up unused.

_Drafted by Claude (Opus 5) on behalf of @pyrello_

joewhitsitt:
Resolves #9858
Resolves #5359
Resolves #10033

Does not eliminate deprecation warnings such as those stemming from HtmlDiff or [typhonius/acquia-php-sdk](https://github.com/uiowa/uiowa/issues/7031)

# Summary

## Per-site files (+1,300 x 4)

- `settings.php` - Swapped the BLT settings include for DRS's, and normalized the trailing doc comment to DRS's exact wording so drush settings doesn't re-append a duplicate on every run. 
- `settings/default.local.settings.php` - Swapped Acquia\Blt\Robo\Common\EnvironmentDetector for DRS's own Acquia\Drupal\RecommendedSettings\Helpers\EnvironmentDetector (same methods) but the BLT class no longer exists. Also restores the `stage_file_proxy.settings.origin` line that BLT's `postSourceBuildSettings` used to write into `local.settings.php` on every regeneration (lost when `ReplaceCommands.php` was removed), now defaulting to the site's own live URL instead of its internal Acquia production domain - resolves #5359.
- `settings/default.includes.settings.php` - Updated the unused stub's cascade diagram/wording from blt.settings.php to acquia-recommended.settings.php. 
- `blt.yml to drs/config.yml` (renamed) - Consolidated the BLT-named per-site metadata file into DRS's own config file. DRS already reads drs/config.yml for the DB name so folding in the rest removes a stray "blt"-named file and a data duplication. The `uiowa.stage_file_proxy.origin` key some existing sites still carry here is no longer read by anything - superseded by the `settings/default.local.settings.php` change above.

## Everything else

  - `.ddev/commands/web/blt` - removed, no vendor/bin/blt left to shim.
  - `.travis.yml` - comment update, dropped stale "still using BLT command hook" note.
  - `README.md` - rewrote BLT-specific setup/workflow sections for DRS + ./sn.
  - `blt/.gitignore, blt/.schema_version, blt/blt.yml ` - removed with the rest of blt/.
  - `blt/src/Blt/Plugin/Commands/ReplaceCommands.php` - removed; guarded BLT-only Robo commands that can't run once BLT/Robo is gone.
  - `composer.json` - removed acquia/blt/acquia/blt-multisite, added acquia/drupal-recommended-settings + its operations override to use our own local-settings template.
  - `composer.lock` - regenerated to match.
  - `config/README.md, config/envs/{README,ci,dev,local,prod,stage}/README.md, config/features/README.md ` - swapped dead BLT doc links for a current Acquia config_split doc; one comment recredited to the deploy's config-import step instead of "BLT's configuration integrity checks."
  - `drs/config.yml` (new) - project-wide local DB credentials DRS needs; BLT's equivalent lived in blt/local.blt.yml.
  - `drush/Commands/UiowaCommands.php` - swapped ./vendor/bin/blt drupal:update for ./sn site:update in the post-sql:sync hook.
  - `hooks/{dev,prod,test,stage}/post-files-copy/*` - removed; BLT's post-files-copy hook was a no-op stub, never overridden here.
  - `patches/README.md` - reworded a sentence crediting BLT for triggering patches (it's cweagans/composer-patches itself).
  - `phpcs.xml, phpcs.xml.dist ` - ruleset renamed blt to uiowa.
  - `salt.txt` - reworded the placeholder text from "blt-expects..." to "drs-expects...".
  - `scripts/ci/lint.sh` - dropped blt/src from linted paths.
  - `sitenow/src/Command/MultisiteCreateCommand.php` - new sites now get one drs/config.yml instead of blt.yml, and local settings are generated via drush settings --uri= instead of blt:init:settings. Also dropped the now-unused stage_file_proxy.origin from the generated drs/config.yml (see the settings/default.local.settings.php note above). While touching this file: --dry-run now also shows site-name, requester, and config split when passed, since those land in drs/config.yml (which a dry run never writes) and previously had no way to be checked before the real run - resolves #10033.
  - `sitenow/src/Command/MultisiteDeleteCommand.php` - DB-name confirmation check now reads drs/config.yml instead of blt.yml.
  - `sitenow/src/Command/MultisiteInstallCommand.php` - docblock update (blt.yml to drs/config.yml).
  - `sitenow/src/Command/SiteInstallCommand.php` - siteConfig()/profile() now read drs/config.yml instead of blt.yml/blt/blt.yml.
  - `sitenow/src/Command/SiteSyncCommand.php` - auto-generates a site's local.settings.php before syncing if one doesn't exist yet, covering first-time syncs.
  - `sitenow/src/Traits/SiteNowCommandsTrait.php` - removed a stale `@todo Move out of BLT` comment.
  - `sitenow/src/Utility/Multisite.php` - updated an exception message from "...by BLT" to "...by DRS".
  - `tests/README.md` - rewrote BLT's local.blt.yml-driven test-selection instructions to plain vendor/bin/phpunit CLI usage.
  - `tests/phpunit/src/Unit/FilesystemTest.php` - updated assertions/fixtures for drs/config.yml and the new settings.php require line.
  - `tests/phpunit/src/Unit/MultisiteCreateTest.php` - variable rename ($blt to $drs); dropped the stage_file_proxy assertion and the now-removed prod_domain parameter, otherwise assertions/shape unchanged. Added coverage for the new dry-run summary rows (#10033).
  - `tests/phpunit/src/Unit/MultisiteDeleteTest.php` - updated fixtures to write drs/config.yml instead of blt.yml.
  - `tests/phpunit/src/Unit/SiteInstallTest.php` - updated fixtures/assertions for drs/config.yml (per-site and repo-root profile source).
  - `assets/appSearch.css, assets/appSearch.js, index.html` - needed to be updated to use new manifest file, but browser policy is preventing it from fetching/working currently. seldom used. removing for now.

# How to test

## [One-time steps](#one-time-steps)

_after pulling this branch (everyone needs to run this once)_

Existing `local.settings.php` files reference a BLT class that's now gone, which breaks even basic `drush` commands (`drush list`, `drush status`) until fixed. This one-liner patches just the broken `use` line - it does not touch any hand-edited customizations (aggregation toggles, `stage_file_proxy`, credentials) below it:

```
find docroot/sites -path '*/settings/local.settings.php' -exec perl -pi -e \
  's/use Acquia\\Blt\\Robo\\Common\\EnvironmentDetector;/use Acquia\\Drupal\\RecommendedSettings\\Helpers\\EnvironmentDetector;/' {} +
```

Existing `local.settings.php` files also either carry the old, stale `stage_file_proxy` origin (an internal Acquia domain, sometimes just wrong, e.g. `default.prod.drupal.uiowa.edu` on every site) or none at all. This corrects or adds it, defaulting each site to its own live URL:

```
find docroot/sites -path '*/settings/local.settings.php' | while read -r f; do
  site=$(echo "$f" | sed -E 's#docroot/sites/([^/]+)/settings/local\.settings\.php#\1#')
  if grep -q "stage_file_proxy.settings'\]\['origin'\]" "$f"; then
    perl -pi -e "s#\\\$config\['stage_file_proxy\.settings'\]\['origin'\] = '.*?';#\\\$config['stage_file_proxy.settings']['origin'] = 'https://$site';#" "$f"
  else
    printf "\n\$config['stage_file_proxy.settings']['origin'] = 'https://%s';\n" "$site" >> "$f"
  fi
done
```

Run both once, then `ddev drush status` should work again.

## Testing checklist

*1. Composer / bootstrap*

- `ddev composer install` - confirm no `acquia/blt` in the output, `acquia/drupal-recommended-settings` installs
- `ddev drush @default.local status` (or any site) - bootstraps and connects

*2. `multisite:create` (the big one)*

- `./sn umc <test-site>.uiowa.edu --dry-run --requester=<hawkid> --site-name="Test Site" --split=event` - review the plan; confirm it shows one `Write .../drs/config.yml` step, not `blt.yml`, and confirm the summary at the top shows `Site name`, `Requester`, and `Config split` rows with the values you passed (#10033)
- Run the same dry-run without those three options and confirm those rows disappear entirely rather than showing blank
- Run it for real against a throwaway site (`--no-commit --no-db` if you don't want a real cloud DB yet)
- Verify the new site's `settings.php` has the DRS require line, not a BLT one
- Verify `docroot/sites/<site>/drs/config.yml` exists with `project`, `drush`, `drupal.db.database`, and `uiowa` keys (the old `blt.yml` shape, same file now)
- Verify `local.settings.php` was generated with the correct DB name and DDEV host (`db`/`db`/`db`)
- Verify `sites.php` aliases and `sitenow/manifest.yml` got the new entries
- Discard everything this step created before moving on, so your tree is clean for the delete check below:
```
git checkout -- docroot/sites/sites.php sitenow/manifest.yml
rm -rf docroot/sites/<test-site>.uiowa.edu drush/sites/<test-site-id>.site.yml
```

*3. `site:install` / `multisite:install`*

Use an already-existing site here, not the throwaway one from step 2 (that one got discarded).

Test this in DDEV only - `site:install` can also run directly against a real Acquia Cloud environment (it only requires DDEV when it's not on Acquia), but that's not something to exercise as part of reviewing this PR.

Note: `--dry-run` on a not-yet-installed site only reports its classification (e.g. `absent`) and exits - it does not preview the `site-name`/`requester`/`config.split` values from `drs/config.yml`. Those are only read and applied for real, inside `reconcile()`. So there's no dry-run preview for this one; verify by running it for real and checking the result.

- `ddev sn site:install <test-site>.uiowa.edu --dry-run` - just confirms the site's classification (absent/partial/installed); not a preview of drs/config.yml values
- Run it for real, then `ddev drush @<test-site>.local uli` and open the link. Confirm, for whichever of these you set in drs/config.yml: the site name changed, the requester was added as a user, and the config split's content type/config landed

*4. `site:sync` (first-time vs. repeat)*

- Most sites already have a `local.settings.php` from past `blt bis` runs, so simulate a never-synced site: `rm docroot/sites/<site>/settings/local.settings.php`, then run `ddev sn site:sync <site>` - confirm the new "Generate local settings" step fires automatically, then the sync proceeds
- Run it again on a site you already have locally - confirm no regeneration step, normal sync behavior
- `ddev sn sync:all` with a couple of sites in `sitenow/local.sites.yml` - same check, since it just shells out to `site:sync` per site

*5. `multisite:delete`*

Dry-run only - `multisite:delete` (like `multisite:create`) refuses to run against a dirty working tree, which this branch's own tree usually is mid-review. A dry run for an existing site still needs a clean tree, so make sure step 2's discard actually happened first.

- `./sn umd <existing-site>.uiowa.edu --dry-run` - confirm the DB-name confirmation check reads `drs/config.yml` and passes

*6. Automated checks (should already be green, but worth a glance)*

- `ddev exec vendor/bin/phpunit tests/phpunit/src/Unit/` - 295 tests pass
- `ddev ci lint` - phpcs clean

# Follow-up

- Upon merge to `main` everyone pulling the project needs to run the "One-time step" above.
- Created follow-up https://github.com/uiowa/uiowa_auth/issues/17
- Team agreed on a scheduled deployment for this during issue refinement.
- Deploy `main` to `develop` to get envs in-sync and to confirm no major issues with dev/stage before deployment to prod.


joewhitsitt:
Screenshot 2026-09-02 at 2 31 32 PM

🫣 

<details>
Andre300 Yes GIF by Chris Piascik
</details>

GaryRidgway:
Pulling to test

GaryRidgway:
For testing section 3, it says use an existing site, but if I sync down an existing site would already have a db created that has been installed, yes?

Do I need to sync a site down, drop the db, and then try and install it?

joewhitsitt:
Of all the 1300 sites in the manifest, sync a site you haven't previously synced the db locally for. Some lab/org site, not sandbox/admissions/etc. By default the site exists in the repo only, the db on your machine has not been stubbed out and populated.

GaryRidgway:
For testing section 4, it says "sync a site that you haven't synced before, has no local.settings.php file". But ALL my sites in my `docroot/sites` have a local.settings.php file.

joewhitsitt:
Did you do the one-time step at the top of the instructions?

GaryRidgway:
> Did you do the one-time step at the top of the instructions?

I copied and ran it, but perhaps it did not work. ~~I'll try it again~~

Checked with Claude and it worked. Still the only two sites that don't have a `local.settings.php` for me are are the new `choirs.music.uiowa.edu` and `strings.studio.uiowa.edu` sites. I can try using those.

joewhitsitt:
okay. yeah sorry, I realized that you probably have one for every site since you last ran a umc/blt bis. if hadn't identified those two, it would be just to delete a local.settings.php file for a site you know you have never synced locally before.

GaryRidgway:
I am having trouble syncing a new site. Did some chatting with Claude and came to the conclusion that the grant for permissions to change DB stuff is done in `scripts/ci/setup.sh#70`, which is only ever ran if you have done the test suite, and not on `ddev restart` or `ddev sn`. 

Because this branch introduced the following structure in `drs/config.yml`:

```

drupal:
  db:
    username: db
    password: db
    host: db

```

It now actually CONSUMES the grant given in the test suite.

We should probably be checking/setting that permission in either `ddev restart` or `ddev sn` I think?

joewhitsitt:
Okay. I was hoping to get rid of root/root as BLT had it, but we might need to switch back. For some reason, I didn't encounter that

```
    username: root
    password: root
```

joewhitsitt:
We have been using a plain-jane ddev mysql since we adopted it back when we switched from drupalvm. So until we have a use-case for using db/db and overriding permissions to do what we need i think it is just easier to go back to root/root

GaryRidgway:
As I'm testing, and realizing that when someone else deletes a site and another person pulls it, they might still have leftover files in their local repo because it doesn't completely delete everything in a site directory. Since git allows for some of the ignored files to stay it's going to orphan some of those directories. When I ran step six and tried to get all the tests to pass I still had a few remaining sites left over and it failed on those. It will definitely fail again, unless we create a prune step in our composer install or our git pull or whatever.

GaryRidgway (approved):
After all the struggles I was having with my local, I was able to test this completely and it is working as intended!

joewhitsitt:
The site dir orphans from deprovisions isn't new but I totally hear you. I think in the past, a combination of `git clean -fd docroot/sites && ddev mutagen sync` has helped my local get up to speed with those types of remnants.

Issue #9613, opened 2026-02-17
`<p>` tags are stripped from image gallery captions


joewhitsitt:
# Problem
With the recent photoswipe update, we changed how captions were referenced and pulled in to the galleries. The functionality is using a heavy handed XSS::filter() which is stripping `<p>` tags.

# Proposed solution
Submit a patch to https://www.drupal.org/project/photoswipe/issues/3574103 and use locally.

briand44:
## Grooming
 
**Quadrant:** 

Complex

 **Discussion:**
 - Security update causes a regression here - https://iowaweb.slack.com/archives/C0164FNN1PV/p1788365626768279
 - Should take a different approach here
 - Change likely needs to happen higher in the processing before reaches the javascript
 - Currently global caption markup for image galleries is plain text, with this update the actual html tags would be rendered
 - Would be good to update issue for community visibility but use fix locally right away
 - Goal would be to allow basic html to be able to be used and rendered appropriately in these captions
 - Discussed ignoring security update for now until we can resolve this
 - Would want to remove the ignore in composer.json - https://github.com/uiowa/uiowa/pull/10111
 - Would also want update photoswipe to latest security release

**Proposed solution or definition of done:**
- Basic html rendered appropriately in image gallery captions
- Comment/Patch on drupal.org issue
- Security ignore in composer.json removed
- Photoswipe updated to latest security release + local fix/patch
- Tested with no known regressions

joewhitsitt:
welp. maintainers are working on it https://www.drupal.org/project/photoswipe/issues/3574103 so this seems to shift to review and implement

fixed/stable release: https://www.drupal.org/project/photoswipe/releases/5.0.10

PR #10111, opened 2026-09-02
ignore photoswipe temporarily in order to address regression

joewhitsitt:
Temporary ignore until we can focus on #9613

# To Test
Passing security build to get other builds running again.


GaryRidgway (approved):
passes security builds

PR #9443, opened 2025-12-15
Hotfix - Registrar final exam type mismatch

pyrello:
# How to test

```
git checkout main && ddev blt ds --site registrar.uiowa.edu && ddev drush @registrar.local registrar_core:final-exam-schedule
```
- https://registrar.uiowa.ddev.site/faculty-and-staff/exam-scheduling/final-exams/final-exam-schedule
- Search for "technology"
- Observe error
```
git checkout hotfix_registrar_error && ddev blt ds --site registrar.uiowa.edu && ddev drush @registrar.local registrar_core:final-exam-schedule
```
- https://registrar.uiowa.ddev.site/faculty-and-staff/exam-scheduling/final-exams/final-exam-schedule
- Search for "technology"
- Should work now.

PR #10138, opened 2026-09-22
Final exam schedule cache

### LINKED ISSUE #9444: Registrar - Improve logic for final exam searching to ensure cache is properly stored and accessed
# Motivation

We put a hotfix (#9443) in place to help fix https://registrar.uiowa.edu/faculty-and-staff/exam-scheduling/final-exams/final-exam-schedule during final exam week. But some issues were identified that should be given a better long term fix.

# Proposed solution

- Refactor the code that is involved in building the final exam form block to ensure more predictable outcomes.
- Ensure that the data structure returned from cache is the same as the data structure returned from `MauiApi::getFinalExamSchedule()`.
- Ensure that the cache ID used for cache the above data structure is consistent throughout.

bewoldt:
Resolves #9444 

Cache ownership for the final exam schedule now lives entirely in `MauiApi::getFinalExamSchedule()`. The cron warmer and the block form read and write the same shape under the same cache ID, replacing the `is_array()/isset()` guard the #9443 hotfix relied on.

The warmer force-refreshes on each run and overwrites the cache only on a genuine success, so a not-ready MAUI response can't wipe out the last good schedule. A non-success response is evicted rather than held for the full TTL.

A failed refresh still returns `_status: ok` when good cached data sits behind it, so nothing downstream distinguished "refreshed" from "served yesterday's copy" — and a 200 with an empty body logged nothing above info. `getFinalExamSchedule()` now warns at the point the refresh fails, so repeated failures surface before the 48-hour grace window runs out.

# How to test

Sync down the Registrar site `ddev sn ds registrar.uiowa.edu`

**1. Basic functionality** `ddev drush @registrar.local cache:rebuild`

Visit [https://registrar.uiowa.ddev.site/faculty-and-staff/exam-scheduling/final-exams/final-exam-schedule](https://registrar.uiowa.ddev.site/faculty-and-staff/exam-scheduling/final-exams/final-exam-schedule)

Confirm the full schedule table renders.
Search for "technology" (or any term) — confirm results filter correctly and no error appears.

**2. Cache is working (cold vs. warm)**
```
ddev drush @registrar.local cache:rebuild
ddev drush @registrar.local php:eval '
$maui = \Drupal::service("uiowa_maui.api");
$session_id = \Drupal::config("registrar_core.final_exam_schedule")->get("session_id");

$t0 = microtime(true);
$maui->getFinalExamSchedule($session_id);
echo "cold: " . round((microtime(true) - $t0) * 1000) . "ms\n";

$t1 = microtime(true);
$maui->getFinalExamSchedule($session_id);
echo "warm: " . round((microtime(true) - $t1) * 1000) . "ms\n";
'
```
Expect: cold call takes several hundred ms (a real MAUI round trip), warm call takes a few ms.

**3. Cron warmer runs clean**
`ddev drush @registrar.local registrar_core:final-exam-schedule`
Expect: no output/errors. (It only logs when MAUI returns something other than `ok`.)


cory-skeers (approved):
Reviewed and tested, and this looks to be a good fix.

I made a slight adjustment to the warning log with some comments--if a schedule isn't posted yet, it'll still drop the caching but won't log it as a failed fetch.

